PR #17, opened 2015-01-22
Linux build configuration.

jilott:


jaredbroad:
Thank you @jilott - it should build in linux without any changes (we run it on Travis CI which builds solutions from Mono)

jilott:
Yeah, I realized that too late. I didn't actually mean to submit this PR, and of course, I can't delete it. However, what I was trying to do was get it to build from the command line, which didn't work for me. I am pretty unfamiliar with Mono though. I was able to successfully build Lean once I added the NuGet package restore folder using Visual Studio.

PR #20, opened 2015-02-01
StandardDeviation and Bollingerbands indicators

mourednik:
* StandardDeviation indicator
* BollingerBands indicator
* Added en-US date format to ParseDate (for international users)


Indicators/QuantConnect.Indicators.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
mourednik: I don't know why this version number changed, or if it matters. I use MonoDevelop 5.7 on Linux.

jaredbroad:
Thanks for the awesome PR @mourednik!

If possible can you please update it to use a "BB" helper method so its inline with the TradeStation API. This will help keep it familiar for new people crossing platforms.

I was chatting with @mchandschuh and we're wondering if Standard Deviation should return 0 instead of input when its not ready - since returning input could have unexpected results (e.g. jumping from $100 price of stock to $0.1 deviation when its ready). Also its possible to return accurate deviations with 2 or more samples - perhaps it should return 0 for first sample, then actual deviations for 2+. The IsReady flag would still only be true with more than X-samples since it was a windowed indicator?

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -223,6 +223,38 @@ public RelativeStrengthIndex RSI(string symbol, int period, MovingAverageType mo
...
+        /// <summary>
+        /// Creates a new BollingerBands indicator which will compute the MiddleBand, UpperBand, LowerBand, and StandardDeviation
+        /// </summary>
+        /// <param name="symbol">The symbol whose BollingerBands we seek</param>
+        /// <param name="period">The period of the standard deviation and moving average (middle band)</param>
+        /// <param name="k">The number of standard deviations specifying the distance between the middle band and upper or lower bands</param>
+        /// <param name="movingAverageType">The type of moving average to be used</param>
+        /// <param name="resolution">The resolution</param>
+        /// <returns>A BollingerBands configured with the specied period</returns>
+        public BollingerBands BOL(string symbol, int period, int k, MovingAverageType movingAverageType = MovingAverageType.Simple, Resolution? resolution = null)
```
mchandschuh: I think we'd like to use BB for the short code helper function here, also, the ability type specify 'k' as a decimal would be nice too, this would allow me to use 1/2 a std deviation instead of full integer incremements

Indicators/BollingerBands.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System;
+
+namespace QuantConnect.Indicators
+{
+    public class BollingerBands : Indicator
```
mchandschuh: Please add some class level XML documentation

Indicators/BollingerBands.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public IndicatorBase<IndicatorDataPoint> LowerBand { get; private set; }
+
+        /// <summary>
+        /// Initializes a new instance of the BollingerBands class
+        /// </summary>
+        /// <param name="period">The period of the standard deviation and moving average (middle band)</param>
+        /// <param name="k">The number of standard deviations specifying the distance between the middle band and upper or lower bands</param>
+        /// <param name="movingAverageType">The type of moving average to be used</param>
+        public BollingerBands(int period, int k, MovingAverageType movingAverageType = MovingAverageType.Simple)
```
mchandschuh: Please update to use decimal instead of integer for the 'k' parameter

Indicators/BollingerBands.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the BollingerBands class
+        /// </summary>
+        /// <param name="name">The name of this indicator</param>
+        /// <param name="period">The period of the standard deviation and moving average (middle band)</param>
+        /// <param name="k">The number of standard deviations specifying the distance between the middle band and upper or lower bands</param>
+        /// <param name="movingAverageType">The type of moving average to be used</param>
+        public BollingerBands(String name, int period, int k, MovingAverageType movingAverageType = MovingAverageType.Simple)
```
mchandschuh: Please update to use decimal instead of integer for the 'k' parameter

Indicators/BollingerBands.cs
```diff
@@ -0,0 +1,84 @@
...
+            : base(name)
+        {
+            MovingAverageType = movingAverageType;
+            StandardDeviation = new StandardDeviation(name + "_StandardDeviation", period);
+            MiddleBand = movingAverageType.AsIndicator(name + "_MiddleBand", period);
+            var kConstant = new ConstantIndicator<IndicatorDataPoint>(k.ToString(), (decimal)k);
+            LowerBand = MiddleBand.Minus(StandardDeviation.Times(kConstant), name + "_LowerBand");
+            UpperBand = MiddleBand.Plus(StandardDeviation.Times(kConstant), name + "_UpperBand");
```
mchandschuh: I love how you did this!! I'm glad this stuff is being used and is helpful.

Indicators/QuantConnect.Indicators.csproj
mchandschuh: All of the other .csproj files use version 12, which equates to visual studio 2013.  I develop on windows. To prevent unnecessary code churn, please change it back to 12.0.

Indicators/StandardDeviation.cs
```diff
@@ -0,0 +1,72 @@
...
+*/
+
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using MathNetStatistics = MathNet.Numerics.Statistics.Statistics;
+
+namespace QuantConnect.Indicators
+{
+    public class StandardDeviation : WindowIndicator<IndicatorDataPoint>
```
mchandschuh: Please add class level XML documentaiton

Indicators/StandardDeviation.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint> window, IndicatorDataPoint input)
+        {
+            if (!IsReady) {
+                return input;
```
mchandschuh: I think we want to return 0 for the zeroth sample and the first sample,

```csharp
if (Samples < 2)
{
    return 0m;
}
```

for any other number of samples, we can use compute the standard deviation in the same manner as below. The `IsReady` flag will let consumers know when we have enough samples for the n period std deviation.

mchandschuh:
Thanks for the awesome pull request @mourednik! I've made a few review comments.  Let me know what you think. Once again, thanks for this awesome contribution!!

mourednik:
No problem. I will fix all issues shortly.

mourednik:
@jaredbroad OK. It shall return zero for initial sample, then standard deviation for 2+ samples. The isReady flag won't be set until it has buffered the full period.

PR #28, opened 2015-02-20
Flexible logging

bdilber:
I wanted to redirect the log output to separate files using log4net and could not find an elegant way of doing it without rewriting the Log class (the only way was redirecting and parsing the console output). 

I ended up extracting a simple `ILogHandler` interface and refactoring the existing code into a default `ConsoleLogHandler` implementation.  The logger should behave exactly the same way as before, but anyone who wants to redirect the log output can do so by setting the `LogHandler` property of the `Log` class to their implementation of `ILogHandler`. (This could also be read from config in Engine.cs for more flexibility.)

### DIFF
--- a/Algorithm/Examples/CustomData_Bitcoin.cs
+++ b/Algorithm/Examples/CustomData_Bitcoin.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Globalization;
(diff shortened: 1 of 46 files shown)


jaredbroad:
Awesome pull request, thank you @bdilber!

PR #36, opened 2015-02-27
Use market price for market orders when calculating required buying power.

bdilber:
Since market orders are created with no price, the `Order.Value` property always returns 0. `SecurityTransactionManager.GetOrderRequiredBuyingPower` method was using `Order.Value` and always returned 0 for market orders. This resulted in market orders always being filled until cash becomes negative during backtests.  I've modified this method to use the current market price to determine order value. (Maybe transaction fees should be added here as well. )

### DIFF
--- a/Common/Securities/SecurityTransactionManager.cs
+++ b/Common/Securities/SecurityTransactionManager.cs
@@ -329,9 +329,13 @@ namespace QuantConnect.Securities
         /// <returns>decimal cash required to purchase order</returns>
         private decimal GetOrderRequiredBuyingPower(Order order)
         {
-            try 
+            try
             {
-                return Math.Abs(order.Value) / _securities[order.Symbol].Leverage;    
+                //If this is a market order, use the current market price to calculate order value
+                var orderValue = order.Type == OrderType.Market
+                    ? order.Quantity*_securities[order.Symbol].Price
+                    : order.Value;
+                return Math.Abs(orderValue) / _securities[order.Symbol].Leverage;    
             } 
             catch(Exception err)
             {


bdilber:
Hi Jared,

The first thing I did when I noticed this was to set the order price in QCAlgorithm MarketOrder() method, but then I realised that the market order value needs to change as the market price changes. (The order could stay in submitted state for a while.) After that I decided to make the Order.Value property virtual, but then I realised that SecurityManager was not available in the Order class, so accessing the current market price within the Order class would have required a bigger change (possibly changing the constructor). After all that, I settled for the change I made in the SecurityTransactionManager class.

jaredbroad:
Ok interesting... Just curious about the use case - when does a market order take a while to fill? We model it at the moment so market order is instant (synchronous or one time loop), and limit orders can take time but have price fixed anyway. The exact fill price is set later with the OrderEvents which update the portfolio (but order.Price is just used for initial placement & our GUI update)

bdilber:
Ok, sorry, I'm still not fully familiar with how everything works, so I assumed there might be a time gap between order creation and processing. Now I realise it's handled synchronously during backtesting. This means the fill price will always be equal to the price when MarketOrder() is called, right?

jaredbroad:
No problem at all! Really appreciate your contributions! 

You are correct for modeled market orders. Typically in live trading this will be known within 5-20ms and will probably vary in price a little but it will be filled or cancelled by the order events themselves. 

In live trading we're actually going further to override the portfolio cash if the price ever differs from models (e.g. manual cash withdrawals).

bdilber:
Got it. Glad I could help. 

By the way, I think there might be another issue with the `GetSufficientCapitalForOrder` method: Leverage seems to be factored in both `GetOrderRequiredBuyingPower` AND  `portfolio.GetBuyingPower`, effectively taking the square of leverage. 

E.g. if you have $1,000 cash with 10 leverage and tried to execute a $100,000 buy order, `GetBuyingPower` would return 10,000, and  `GetOrderRequiredBuyingPower` would also return 10,000, allowing the order and giving you 100 leverage. 

Am I missing something?

jaredbroad:
Great catch, you're absolutely right! Can't believe I missed it.

Perhaps the best solution would be to make the GetOrderRequiredBuyingPower return the *real* buying power required? (it currently returns the *cash* required). Then we can compare order buying power to portfolio buying power

bdilber:
Hi Jared,

Sorry for the late reply. I had actually solved this by doing the opposite (I made GetBuyingPower return the current cash). I did it this way because I found another issue in GetBuyingPower and modified it to solve that issue as well. The second issue has to do with the way movements in the opposite direction are handled in GetBuyingPower:

`holdings * 2  + remainingBuyingPower` 

This causes problems when leverage is greater than 1 and holding P/L is non-zero. Because you end up dividing your margin P/L by leverage. Here is an example describing the problem:

>Let’s say you have $20,000 initial capital and 10 leverage. You use $10,000 to buy $100,000 worth of stock, with a remaining cash of $10,000.
>
>After a while, the price drops 5%, leaving you with a holding value of $95,000 and -5,000 of unrealised P/L. (assuming 0 fees.)
>
>Now you try to go short. The current calculated buying power is $95,000 x 2 + $100,000 = $290,000, which would mean that you could go short $195,000 after liquidating your $95.000 long position. In reality, you should be getting back $5,000 when you liquidate your long position (10k initial cost - 5k loss), and together with your existing 10k capital, you should only be able to go short $150,000 (10.000 x leverage + 5.000 x leverage).

I made the following modification to this method to fix the issue:

`(unrealisedProfit + InitialHoldingsValue/leverage) x 2 + Cash`

Where InitialHoldingsValue is calculated as `holdings.AveragePrice*holdings.AbsoluteQuantity`, and the output is the total unleveraged buying power. (We still need to add possible fees to this too btw.)

This brings me to another problem that I am having right now. Currently, this calculation (`holdings.AveragePrice*holdings.AbsoluteQuantity/leverage`) is the only way we can get the real opening cost of a holding. When dealing with a fixed margin product like futures, where there is a fixed margin requirement per contract instead of a fixed leverage factor, there is no way of calculating the current margin utilisation (because leverage keeps changing with price). I tried to implement fixed margin trading by overriding the Leverage property of my Security class, but this still results in errors when actual fills are calculated in SecurityPortfolioManager.ProcessFill, as it takes the *latest* leverage instead of the *initial* leverage.

I think the SecurityHolding class needs a Margin property that holds the initial cost of opening the position (currenPrice x quantity x leverage for fixed leverage instruments, quantity x initialMarginRequirement for fixed margin instruments.). Then the above expression becomes something like this:

`(unrealisedProfit + initialMargin) x 2 + Cash`

This way, actual margin P/L with fixed margin instruments can be calculated. It will be fully compatible with fixed leverage trading as well. The leverage property of a fixed margin instrument can be overridden to calculate leverage as `initialMarginRequirement/price`. Later, maintenance margin requirements can be added to the security class so a margin call can be realistically simulated. I am using Lean to do CME futures backtests right now, and margin call in CME happens when an account drops below maintenance margin.

If you wish, I can try to combine all of this in a pull request today. What do you think?

jaredbroad:
You're right about the bug in portfolio.GetBuyingPower()! It is dividing profit/loss by leverage, which would cause further inaccuracies. Your unrealized profit + initial holdings value is a good solution. I'm not 100% on the solution for second part because it mixes concepts (equity & futures) -- but lets brain storm :)

I'm relatively new to futures so please correct me if wrong; its a fixed margin requirement per contract held? And maintenance requirement means (Initial Margin + Loss > Maintenance)?

So I have $10k in account, and for 1 contract I need [$5060](http://www.tradestation.com/products/futures/margin-requirements) set aside which I can't trade with? When Lean makes purchases we subtract it from the cash; could we subtract the margin requirement from cash so its not available for other asset types? When you close and make a profit-loss is it added/subtracted in addition to the margin-repayment? I see on TradeStation's website margin requirement is 25% of initial for day-trading (like equities 4x leverage intraday); 

Perhaps we should move these methods into the Security types and make it a little more generic? e.g. Placing trades - equity.Holdings.GetCashImpact(order), futures.Holdings.GetCashImpact(order). The GetCashImpact could handle the buy & sell to generate positive and negative cash values. 

Inside the futures holdings system we use Leverage to set the Margin/Leverage to get the 25% for intraday? This way we could later stream line overnight reduction in leverage for both equities and forex (Day trading equities is 4x, overnight only 2x).

This would need creation of a "Futures" folder along with Futures.cs, FuturesExchange.cs FuturesHoldings.cs, etc to match the pattern of the other asset types. 

portfolio.GetBuyingPower() function would need to access the security specific handler as well? Maybe equity.Holdings.GetBuyingPower(order)? If we have $10k in account, and buy 1 contract, we don't have enough cash for a second contract correct? So future.Holdings.GetBuyingPower() would be $4940 going long (margin not quantity based) -- essentially very different math - Margin*NetContracts, etc..vs Price*NetQuantity?

Could we also have equity.Holdings.GetOrderProfitLoss(), futures.Holdings.GetOrderProfitLoss() - because futures profit loss is based on the number of items per contract I think? Whereas the price of the future is per item? i.e. 1000 bushels vs 100 barrels? ProfitLoss = ContractLotSize * deltaPrice * ContractHoldings

Later we could have configurable future.ContinuousFunction for customized stitching together contracts? (Like the equity.Model interface pattern with a default contract stitching function)

Feedback welcome! :) This is a new space for me. We've always wanted to add futures support though so this seems like a good time to do it properly,

(PS: I'm thinking ahead for Options and other assets down the line as well, maybe keeping it generic /asset type based would be safer for future proofing.)

(PPS: My understanding of buying power definition in equities is "cash x leverage"; but maybe to keep the function names (intent) correct, and compare apples to apples, we could rename both of the relevant functions? 
portfolio.GetBuyingPower -> portfolio.GetFreeCash()
transactions.GetOrderBuyingPower -> transactions / future.Holdings.GetOrderCashImpact()

bdilber:
Yes, you are right about the margin requirement for futures (at least for CME futures I think, I am no expert at this myself, I have only traded CME index futures so far). 

To recap, if we take the S&P E-mini: 

- Each contract is worth 50 x index, which puts it currently at around $105.000 right now.
- The initial margin requirement per contract is set at $5.060, currently defined as 110% of the maintenance margin, which is currently $4600. This what the exchange demands, the brokerage may ask for an additional margin.
- For each contract you buy/sell you need to set aside the initial margin requirement. This gives you a leverage of roughly 21 (~105k/5k). So If you wish to open a 5 contract position, you would set aside $5060 x 5 = $25.300, for an exposure of ~$525.000.
- Your position P/L is instantly reflected to your margin. If your margin goes below the maintenance margin, a margin call is issued. So, for the 5 contract position above, a margin call would be issued by the exchange if your margin drops below $4600 x 5 = $23.000. This corresponds to about a 2% change in the S&P 500 index.
- Tradestation gives you an additional 4x leverage by requiring only %25 of the initial margin if you day trade. So the total leverage is ~84 for day traders. Not all brokerages do this though. I believe Tradestation enforces this by requiring a stop order before the day ends.

I have already implemented the  Futures.cs, FuturesExchange.cs FuturesHoldings.cs, etc. classes. They seem to work fine except the ProcessFill issue I mentioned above. For the lot size part, I've overridden the Price property of the Futures class to multiply the base price with this multiplier, which is supplied to the constructor. I also have an InitialMarginRequirement property, which is also supplied to the constructor. The Leverage property is overridden to return Price/_initialMarginRequirement to calculate the current leverage on the fly. This works well when opening positions, but doesn't when liquidating them in ProcessFill, because the current leverage is used instead of the one when opening the position. 

This is why I think we need a mechanism to store the *cash impact* or *margin* of a position in the SecurityHolding class, so that it can be used when processing the fill instead of quantity x averagePrice / leverage. (GetCashImpact method for an order would work, but we also need to store the initial cash impact for a position somewhere)

Implementing options will be much more complicated probably, especially when writing options (I have 0 experience in options btw). But I think at the end of the day it will still boil down to calculating margin requirements for a position. 

PS: I think GetFreeCash() and GetOrderCashImpact() are better names, since "buying power" may be misinterpreted.

jaredbroad:
Interesting! Thank you for explaining. As a general principle I prefer if an implementation is a few more lines of code; but more "dedicated", and that everything explicitly means what it's name would suggest so people don't get confused in the future. I'd welcome your PR with this, but maybe we can tweak a few bits to meet those principles? Specifically --

- Agree lets store margin for the holdings we have, in a futures-holding class with property "Margin". When a user goes to place order and purchase the contract could we calculate the margin at runtime rather than fixing? We could set it in a Futures Transaction Model - with a new virtual method GetMargin(price) that others could override. I imagine this will change a lot over time (2008 margin vs today). Do you average margin's over multiple purchases?

- Can Future class Price property would be the cost for 1 unit of the asset? This way if a user imports oil prices as custom data it will work with the rest of Lean. I think people would expect to see the prices they know i.e. Oil is $45-$110 per barrel, and I'd expect the oil.Price property to reflect that.

- Lots size by constructor sounds good. Perhaps we could set it up with a helper method. I can imagine AddSecutity(SecurityType.Future, "ZS", 5000);!

To make these changes work, we'll need to shuffle around the cash impact, buying power to securities so they are generic going forward. Myself and Mike are probably in best place to make those architecture changes so will take a look at it this week. We'll have to rework ProcessFill to use securities implementations of the *concepts* instead of the explicit formulas. This way we won't need any futures specific terms in processing an order fill. It should be about cash changes, quantity updates and average price updates. Once we've setup the pattern, and rerun the tests hopefully it will be a piece of cake to import your Futures*.cs classes! :)

Please correct me if I'm wrong, but Leverage as a concept doesn't seem to apply to futures? Shouldn't it be removed entirely? Or set to 1 so it doesn't get in the way for the default case. And I don't think I understand why we need to store the initial cash impact? Does this work?
```
AverageMargin * ContractHoldings + (Price - AveragePrice)*LotSize = CashImpact?
```

bdilber:
Ok, I will keep experimenting with futures until you decide on the architecture changes then.

Here are my thoughts:

- I think leverage still applies to futures. It makes sense to think of margin as the source of leverage. This article helped me with this: http://www.investopedia.com/ask/answers/06/leverageandmargin.asp. So, l think overriding the Leverage property with Price/initialMarginRequirement makes sense.
- I also think margin applies to all securities, not just futures etc. Same margin calculation is done in all leveraged transactions even if the term margin is not used. It basically reflects how much cash was used to open a leveraged position. The only difference is in how the initialMargin is calculated: with some securities it is defined as a percentage, with others as a fixed amount per item. 
- I imagine margin as the cumulative cash used in a position, so I think something like AverageMargin is not necessary. At any given time for a holding, the total liquidation value can be calculated by taking margin + unrealizedProfit. (which is equivalent to the equation you wrote above)
- With a futures contract, the actual price per contract is itemPrice x lotSize, even though people are more used to talking in terms of per item price. I've overriden the Price property this way and preserved the base price in the futures class with another property: UnderlyingPrice. This was easier to implement since it required no change in the ProcessFill logic. Of course, if you move forward with the changes in ProcessFill you mention, the price can be kept as it is too.

glyphard:
@jaredbroad when you say 'LotSize' above, are you referring to the number of futures contracts bought/sold, or are you referring to the multiplier of the futures contract itself?  (e.g. ES has a multiplier of 50, so an order for 1 contract costs 50 * (Price(S&P 500))

The point earlier that price should be unit price for the future makes sense, to me, since that is how they are quoted by a broker.  However, the multipliers vary quite a bit from one contract to another so 'Multiplier' needs to be a property of a 'Future' class, and be either stored/loaded somewhere, or it needs to be a parameter that the dev is expected to know/input when using the class.

In regards to margining, it's a little more complex than just initial/maintenance.  Those values change over time.  Also, maintenance margin  =  initial * + abs(Mark2Market-losses)

The system that the CME uses for risk based margining involving futures and options is called SPAN.
Here are some details about it, as well as a link to their FTP where the daily parameter files for the past few years are stored.  This is how you can detect when the margin requirements changed at the exchange based on market conditions: 

http://www.cmegroup.com/confluence/display/pubspan/SPAN+Overview
http://www.cmegroup.com/clearing/risk-management/span-use.htmlftp://ftp.cmegroup.com/span/data/cme/

For doing basic backtesting on futures you do not need to dig into the details of span, just make a larger assumption for the initial/maintenance margin (e.g.  maintenance = 0.75* initial margin) instead of the more common maintenance margin = 0.5 * initial margin) during times of relatively high volatility.

That being said when working with any kind of derivative security whether it's futures, options, or credit derivative, it is critical to have two things:
1.  A link between the derivative security, and the underlying security that it is based on: otherwise you won't be able to sort out things like, bad quotes, arbitrage, roll yield, etc...  (I'm still quite a novice in LEAN, so i'm not sure how easy or difficult this linkage would be in this system)
2.  A pluggable margining system like Reg-T, SPAN, or Portfolio Margin that is assignable on a per asset class basis.  Correct me if i'm wrong, but LEAN currently has something like this baked into the <Security>TransationModel abstraction.  If so then that just needs some extension to accurately capture the margin regimes that are available on the various exchanges.

jaredbroad:
Sorry for the delay @glyphard this email slipped past. For "lot size" I meant "contract size" - like mentioned on (Quandl)[https://www.quandl.com/data/SCF/CME_CL1_ON-NYMEX-WTI-Crude-Oil-Futures-1-CL1-Unadjusted-Prices-Roll-on-Open-Interest-Switch-Continuous-Contract-History] - the contract size here is 1000 barrels; priced per barrel $55; so cost of holdings is $55,000?; and with assuming modeled margin $5500; dynamically calculated read-only leverage is 10. 

I think its fair to expect the developer to know the properties of the futures they need, it can be set when they do AddSecurity(). Maybe we can make an AddSecurity(Security obj) so we can pass fully formed futures into the system.

I think the Margin should be modeled - like the transaction models. e.g. Future.Holdings.MarginModel which defaults to something we make (e.g. 10th of cost of holdings + P&L as you say); but can be overridden so people can improve it. And perhaps in this model we can include the SPAN data? 

I don't think we have a pluggable margining system unfortunately; but it would be definitely worth the effort to make it pluggable. Do you think the dedicated asset holdings classes give enough flexibility for such a system? 

I feel like given the breadth of asset support needed its wise to take a minute and map out the proposed changes to ensure they'll elegantly support future needs. I'm just not an expert in those assets so will rely heavily on the community's expertise :)

glyphard:
No worries...  just wanted to clarify since lotsize refers to an order wheras contract size and multiplier are properties of a contract definition.

"I think its fair to expect the developer to know the properties of the futures they need"  That makes sense, but then makes it difficult to implement scanners/universe type strategies, if that matters.

While it may not make sense to link the derivative security to the underlying in AddSecurity() to use a risk based margining system you have to have a link between the derivative security, and the underlying somewhere.   If we implement a static margining system, like 10% of X + P/L that will break down quickly.  For example, if you trade a calender spread in futures say, -1 x 1month CL  +1 x 3month CL with CL trading at 55, that would mean 11k in margin.  The actual margin an exchange and thus a broker would allocate on this much lower, under 2500, which reflects the fact that the primary risk is the price distortion in the term structure of CL, and not the underlying of spot CL.

The thing that all of these margining systems have in common (SPAN, portfolio, even Reg-T (for options)), is that they bucket all securities by underlying, and then net out the P/L for each underlying based on a set of rules/inputs, and then use a percentage of the maximum loss for all of scenarios tested as the margin amount.  

Additionally some margining systems may allow correlations between underlyings, (portfolio margin is an example of this).  For example, this allows a portfolio that has options on the OEX(S&P100) , and options on the SPY (S&P 500 ETF) to have a correlation of say .7, meaning that if you have opposing exposure in both OEX and SPY, you will get lower margin requirements since those risks offset to roughly 70%

Finally for options a pricing model much be used by the margining system so that inputs other than price of the underlying can be 'shocked' (e.g. impliedVolatility) this is the only way to get back to P/Ls that can be netted out.

All that said, to extend LEAN to adequately handle exchange traded derivatives like futures, and options a margining system needs to have a structure something to the effect of:


List<Securities>  PortfolioHoldings; // (all securities held in a portfolio, derivative or not)

List<UnderlyingSecurity> Underlyings = GetUnderlyingSecuritiesForPorfolio()...
Dictionary<UnderlyingSecurity,List<DerivativeSecurities>>  UnderlyingsMap = GetUnderlying2DerivativeMap()...

enum ModelParameterInputType{
   UnderlyingPriceChange = 0,
   ImpliedVolatilityChange = 1,
...
}
class ModelParamChange{
  ModelParameterInputType InputType;
  ModelParamPercentChangeValue (number between 0 and 1)
}

decimal CalculateProjectedProfitAndLoss( ModelParamChange[] parametersToShock){
   decimal pnlChange = 0.0;
   foreach underlying
       foreach derivative
          pnlChange += model.getModelPnL(underlying, derivative, parametersToShock);

   return pnlChange;
}

decimal GetMarginRequirements(){
   inspect portfolio to decide what types of ModelParamChanges scenarios to generate
   decimal marginRequirements = 0.0;
   foreach scenario
      scenarioMargin = CalculateProjectedProfitAndLoss(scenario)
      marginRequirements = Max(scenarioMargin, marginRequirements)
  return marginRequirements   
}

There's a little more to it than this, but does this make sense so far?

To speed up backtesting, margining only needs to be calculated a few times a day like, on open, on close, and on trade, unless the account is close to overmargined(margin getting close to net-liquidating-value).

jaredbroad:
Hello all, I just wanted to ping this thread to let you know with commit #1f15eb2c0457cc825162e985a2906d86e9ce5fb6 we have designed a full PortfolioModel system, combined with a MarginModel system. These two work to move the ProcessFill() operation into the security; and moves the leverage calculations into a dedicated model so you can do powerful (i.e. futures) margin calculations which are separate from cash and can be dynamic (change over time).

We also used the new PortfolioModel system to create a special model for FOREX -- this enabled us to debit and credit the cashbook for the algorithm -- and maintain a *real* cashbook to properly model FOREX cash swap transactions! To power this we also import FX data streams to convert the cash to BaseCurrency (USD).

We're super happy with the new modelling framework. Hopefully it future-proofs QuantConnect to handle any asset class in the future. We added a bunch more tests to cover the sensitive areas (cash fills, expected total portfolio value) so please run tests to confirm behaving as expected.

Thank you all for your feedback here! It was super helpful for designing this new architecture! @bdilber  - if this architecture works please submit your futures system :)!

bdilber:
Great news! 

At first look, everything seems to be covered. I will start working on the futures implementation right away.

PR #37, opened 2015-02-28
Issue 7: CCI indicator

mattmast:
Here's the pull request for the CCI indicator.

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -213,13 +213,29 @@ namespace QuantConnect.Algorithm
         /// <param name="period">The period over which to compute the RSI</param>
         /// <param name="movingAverageType">The type of moving average to use in computing the average gain/loss values</param>
         /// <param name="resolution">The resolution</param>
-        /// <returns>The RelativeStrengthIndex indicator for the requested symbol over the speified period</returns>
+        /// <returns>The RelativeStrengthIndex indicator for the requested symbol over the specified period</returns>
         public RelativeStrengthIndex RSI(string symbol, int period, MovingAverageType movingAverageType = MovingAverageType.Simple, Resolution? resolution = null)
         {
             var name = CreateIndicatorName(symbol, "RSI" + period, resolution);
             var rsi = new RelativeStrengthIndex(name, period, movingAverageType);
             RegisterIndicator(symbol, rsi, resolution, x => x.Value);
             return rsi;
+        }
+
+        /// <summary>
+        /// Creates a new CommodityChannelIndex indicator. The indicator will be automatically
+        /// updated on the given resolution.
+        /// </summary>
+        /// <param name="symbol">The symbol whose CCI we want</param>
+        /// <param name="period">The period over which to compute the CCI</param>
+        /// <param name="movingAverageType">The type of moving average to use in computing the typical price averge</param>
+        /// <param name="resolution">The resolution</param>
+        /// <returns>The CommodityChannelIndex indicator for the requested symbol over the specified period</returns>
+        public CommodityChannelIndex CCI(string symbol, int period, MovingAverageType movingAverageType = MovingAverageType.Simple, Resolution? resolution = null) {
+            var name = CreateIndicatorName(symbol, "CCI" + period, resolution);
+            var cci = new CommodityChannelIndex(name, period, movingAverageType);
+            RegisterIndicator(symbol, cci, resolution);
+            return cci;
         }
 
         /// <summary>
--- a/Indicators/CommodityChannelIndex.cs
+++ b/Indicators/CommodityChannelIndex.cs
@@ -85,6 +85,9 @@ namespace QuantConnect.Indicators {
             _typicalPriceAverage.Update(input.Time, typicalPrice);
             _typicalPriceMeanDeviation.Update(input.Time, typicalPrice);
 
+            if (_typicalPriceMeanDeviation.Current == 0.0m) {
+                return 0m;
+            }
             return (typicalPrice - _typicalPriceAverage.Current) / (_k * _typicalPriceMeanDeviation.Current);
         }
     }
--- a/Indicators/CommodityChannelIndex.cs
+++ b/Indicators/CommodityChannelIndex.cs
@@ -31,8 +31,8 @@ namespace QuantConnect.Indicators {
     /// </summary>
     public class CommodityChannelIndex : TradeBarIndicator {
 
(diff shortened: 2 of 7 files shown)


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -213,7 +213,7 @@ public MomentumPercent MOMP(string symbol, int period, Resolution? resolution =
         /// <param name="period">The period over which to compute the RSI</param>
         /// <param name="movingAverageType">The type of moving average to use in computing the average gain/loss values</param>
         /// <param name="resolution">The resolution</param>
-        /// <returns>The RelativeStrengthIndex indicator for the requested symbol over the speified period</returns>
+        /// <returns>The RelativeStrengthIndex indicator for the requested symbol over the specified period</returns>
```
mchandschuh: thanks for fixing my spelling atrocities!

Indicators/CommodityChannelIndex.cs
```diff
@@ -0,0 +1,102 @@
...
+    ///
+    ///     There are four steps to calculating the Mean Deviation, first, subtract
+    ///     the most recent 20-period average of the typical price from each period's
+    ///     typical price. Second, take the absolute values of these numbers. Third,
+    ///     sum the absolute values. Fourth, divide by the total number of periods (20).
+    /// </summary>
+    public class CommodityChannelIndex : TradeBarIndicator {
+
+        /// <summary>This constanst is used to ensure that CCI values fall between +100 and -100, 70% to 80% of the time</summary>
+        private readonly decimal _k = 0.015m;
```
mchandschuh: nit - can we define this as a const?

Indicators/CommodityChannelIndex.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// Keep track of the simple moving average of the typical price
+        /// </summary>
+        private readonly IndicatorBase<IndicatorDataPoint> _typicalPriceAverage;
+
+        /// <summary>
+        /// Keep track of the mean absolute deviation of the typical price
+        /// </summary>
+        private readonly IndicatorBase<IndicatorDataPoint> _typicalPriceMeanDeviation;
```
mchandschuh: I think these indicators can be public get private set properties, this will allow these to be used in composition if a user desired.  Thoughts?

Indicators/CommodityChannelIndex.cs
```diff
@@ -0,0 +1,102 @@
...
+        protected override decimal ComputeNextValue(TradeBar input) {
+            decimal typicalPrice = (input.High + input.Low + input.Close) / 3.0m;
+
+            _typicalPriceAverage.Update(input.Time, typicalPrice);
+            _typicalPriceMeanDeviation.Update(input.Time, typicalPrice);
+
+            if (_typicalPriceMeanDeviation.Current == 0.0m) {
+                return 0.0m;
+            }
+            return (typicalPrice - _typicalPriceAverage.Current) / (_k * _typicalPriceMeanDeviation.Current);
```
mchandschuh: nice implementation!

Indicators/MeanAbsoluteDeviation.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace QuantConnect.Indicators {
+    /// <summary>
+    /// This indicator computes the n-period mean absolute deviation.
+    /// </summary>
+    public class MeanAbsoluteDeviation : WindowIndicator<IndicatorDataPoint> {
+        private readonly IndicatorBase<IndicatorDataPoint> _mean;
```
mchandschuh: I think this can be a public get private set property so other indicators can use it in composition. Thoughts?

Indicators/MeanAbsoluteDeviation.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Initializes a new instance of the MeanAbsoluteDeviation class with the specified period.
+        /// 
+        /// Evaluates the mean absolute deviation of samples in the lookback period. 
+        /// </summary>
+        /// <param name="name">The name of this indicator</param>
+        /// <param name="period">The sample size of the mean absoluate deviation</param>
+        public MeanAbsoluteDeviation(string name, int period)
+            : base(name, period) {
+            _mean = MovingAverageType.Simple.AsIndicator(string.Format("{0}_{1}", name, "Mean"), period);
```
mchandschuh: awesome!

Indicators/MeanAbsoluteDeviation.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <param name="window">The window for the input history</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint> window, IndicatorDataPoint input) {
+            _mean.Update(window[0]);
```
mchandschuh: window[0] == input, maybe I should at this to the documentation.  

The only difference i window[0] ends up having to do a little math, since the implementation of RollingWindow<T> uses a 'circular list' that keeps track of its head. It also ends up doing some bounds checking too.

mchandschuh:
Awesome code @mattmast !! Thank you so much for the contribution. I made a few minor comments, let me know what you think. Thanks!!

Indicators/CommodityChannelIndex.cs
mattmast: Yeah that makes sense.  I got the example looking at the ATR implementation which has the private _smoother.  Should that have the same private/public properties?  In general what's the deciding factor to make them private or public?

Indicators/MeanAbsoluteDeviation.cs
mattmast: Yeah I'll change it, I just missed the input parameter.

Indicators/CommodityChannelIndex.cs
mattmast: Yeah for sure.  It even has constant in the comment.... that shows my terrible spelling..... xD

mattmast:
No problem!  

Sorry for the duplicate commit messages.  I tried to rebase from upstream and messed it up.    I couldn't figure out how to remove them.   I have since gotten more familiar with git xD

Indicators/CommodityChannelIndex.cs
mchandschuh: Thanks for pointing this out.  I don't think the ATR example is a special case. I've opened an issue to update it.

PR #42, opened 2015-03-06
Fixes SecurityPortfolioManager.GetBuyingPower leverage bug

bdilber:
Hi, 
This fixes the buying power issue we talked about earlier. I renamed the methods GetFreeCash and GetOrderCashImpact as we discussed. 

(I also made the DataMinMaxPrices method in SecurityTransactionManager virtual)

### DIFF
--- a/Common/Securities/SecurityPortfolioManager.cs
+++ b/Common/Securities/SecurityPortfolioManager.cs
@@ -411,39 +411,46 @@ namespace QuantConnect.Securities
         ///     Similarly the desired trade direction can impact the buying power available
         /// </remarks>
         /// <returns>Decimal total buying power for this symbol</returns>
-        public decimal GetBuyingPower(string symbol, OrderDirection direction = OrderDirection.Hold) 
+        public decimal GetFreeCash(string symbol, OrderDirection direction = OrderDirection.Hold) 
         {
             //Each asset has different leverage values, so affects our cash position in different ways.
-            // Basically position affect on cash = holdings / leverage
-            var holdings = (Securities[symbol].Holdings.AbsoluteQuantity * Securities[symbol].Close);
-            var remainingBuyingPower = Cash * Securities[symbol].Leverage;
+            var holdings = Securities[symbol].Holdings;
+            var remainingCash = Cash;
+            var price = Securities[symbol].Price;
+            var leverage = Securities[symbol].Leverage;
 
-            if (direction == OrderDirection.Hold) return remainingBuyingPower;
+            if (direction == OrderDirection.Hold) return remainingCash;
             //Log.Debug("SecurityPortfolioManager.GetBuyingPower(): Direction: " + direction.ToString());
 
+            //If the order is in the same direction as holdings, our remaining cash is our cash
+            //In the opposite direction, our buying power is 2 x current value of assets + our cash
             if (Securities[symbol].Holdings.IsLong)
             {
                 switch (direction)
                 {
                     case OrderDirection.Buy:
-                        return remainingBuyingPower;
+                        return remainingCash;
                     case OrderDirection.Sell:
-                        return holdings * 2  + remainingBuyingPower;
+                        var profit = (price - holdings.AveragePrice) * holdings.AbsoluteQuantity;
+                        var assetCost = holdings.AveragePrice * holdings.AbsoluteQuantity / leverage;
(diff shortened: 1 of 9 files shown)


Common/Securities/SecurityPortfolioManager.cs
```diff
@@ -411,39 +411,49 @@ public void SetCash(decimal cash)
...
             if (Securities[symbol].Holdings.IsLong)
             {
                 switch (direction)
                 {
                     case OrderDirection.Buy:
-                        return remainingBuyingPower;
+                        return remainingCash;
                     case OrderDirection.Sell:
-                        return holdings * 2  + remainingBuyingPower;
+                        var profit = (price - holdings.AveragePrice) * holdings.AbsoluteQuantity;
```
jaredbroad: Excellent as usual @bdilber; perhaps to centralize the formulas we could use the holdings class here to get the unrealized profit?
https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityHolding.cs#L293

Common/Securities/SecurityPortfolioManager.cs
```diff
@@ -411,39 +411,49 @@ public void SetCash(decimal cash)
...
+                        var assetCost = holdings.AveragePrice * holdings.AbsoluteQuantity / leverage;
```
jaredbroad: With the same notion of centralizing formulas perhaps we could put this in the holdings class as a property? I use it in Portfolio.TotalUnleveredAbsoluteHoldingsCost -- this could be pulled out into the holdings classes themselves:
https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityPortfolioManager.cs#L251

bdilber:
Ok, I'm on it.

bdilber:
Ok, I've added Leverage as a property to SecurityHolding and modified the constructors of security classes to pass leverage to SecurityHolding. I've also added UnleveredHoldingsCost and  UnleveredAbsoluteHoldingsCost properties and modified SecurityPortfolioManager to use them instead of doing the calculation directly.

This solved the margin issue with my FuturesHolding implementation as well, all I had to do was to override UnleveredHoldingsCost to return maintenanceMarginRequirement * Quantity.

jaredbroad:
Awesome thank you! Almost there; since security.Leverage and security.SetLeverage() are well established we'll have to keep supporting it. This property is used elsewhere in the codebase(processFill and SetHoldings) and many user algorithms. So it means security.SetLeverage() would set the Leverage property in the security.Holdings class, and the security.Leverage would reference the security.Holdings.Leverage public property. 

Previous we thought the leverage was more of a "security" wide property and static/fixed. But now with futures it isn't quite as well defined :) So moving it to holdings is a step in the right direction.

jaredbroad:
Thanks @bdilber I'll do the PR now and fix the last step myself -- would like to deploy this fix tonight / ASAP since it affects many users algorithms in QuantConnect

bdilber:
Sorry, I was away for the weekend and couldn't reply. I just pulled the last version and everything seems to work fine. I will keep working with futures and do further tests with real numbers from my brokerage, to make sure all of the numbers (cash, fees, p/l, margin etc.) are calculated correctly.

jaredbroad:
Excellent thank you, perhaps we can create an example algorithm test which
asserts on failing? The CashTesting algorithm was meant for this case but
its still a manual process. We could mimic the cash test with manually
generated OrderEvents passed into the ProcessFill handler? I was terrible
at writing tests before but thankfully mhandschuh has shown me the light!

On Mon, Mar 9, 2015 at 6:33 AM, Baran Dilber <notifications@github.com>
wrote:

> Sorry, I was away for the weekend and couldn't reply. I just pulled the
> last version and everything seems to work fine. I will keep working with
> futures and do further tests with real numbers from my brokerage, to make
> sure all of the numbers (cash, fees, p/l, margin etc.) are calculated
> correctly.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/QuantConnect/Lean/pull/42#issuecomment-77823131>.
>



-- 
Jared Broad

www.quantconnect.com
Phone-USA: +1 917 310 1082
*Democratizing Finance, Empowering Individuals*
Facebook <http://www.facebook.com/QuantConnect> | Twitter
<https://twitter.com/#!/QuantConnect> | LinkedIn
<http://www.linkedin.com/company/quantconnect> | Skype: jaredbroad

PR #44, opened 2015-03-09
Fixed a few compilation warnings

ammachado:


jaredbroad:
Awesome thank you @ammachado! The compiler warnings bug me too, but its such a huge task! 40 more down, 600 to go... :)

PR #57, opened 2015-03-19
Minor change to allow config to be read from a custom json file

QANTau:
I would like to be able to read from a custom json file located in a path external to the project directory.  The intention is that the custom file could contain data that I never want to have in a git managed directory (usernames, passwords, etc).

### DIFF
--- a/Configuration/Config.cs
+++ b/Configuration/Config.cs
@@ -36,7 +36,7 @@ namespace QuantConnect.Configuration
         * CLASS VARIABLES
         *********************************************************/
         //Location of the configuration file.
-        private const string _config = "config.json";
+        private static string _config = "config.json";
 
         //Has the configuration been loaded from disk:
         private static bool _loaded;
@@ -83,13 +83,27 @@ namespace QuantConnect.Configuration
         }
         
         /// <summary>
+        /// Sets a custom file name
+        /// </summary>
+        /// <param name="customFileName"></param>
+        private static void SetCustomFileName(string customFileName)
+        {
+            _loaded = false;
+            _config = customFileName;
+        }
+
+        /// <summary>
         /// Get the matching config setting from the file searching for this key.
         /// </summary>
         /// <param name="key">String key value we're seaching for in the config file.</param>
         /// <param name="defaultValue"></param>
+        /// <param name="customFileName"></param>
         /// <returns>String value of the configuration setting or empty string if nothing found.</returns>
-        public static string Get(string key, string defaultValue = "")
+        public static string Get(string key, string defaultValue = "", string customFileName = "")
         {
+            if (!String.IsNullOrEmpty(customFileName))
+                SetCustomFileName(customFileName);
+
             var value = "";
             try
             {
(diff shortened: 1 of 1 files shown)


jaredbroad:
Thank you for PR @QANTau! We really appreciate community contributions! 

IMHO the right way to do this would be to add the config file to your git ignore to keep it out of the commits, or change the const config file name to a path outside of the git directory.

Is the submitted solution is for having two config files? (it is a method level argument which infers it would it change from request to request). In general this would increase entropy and might later cause bugs.

Also its slightly counter intuitive / non-deterministic:
```
Config.Get("my-password", "passwords.json");
Config.Get("algorithm"); //Not Found; because custom config file overrides the default config
```

I think it might be time we convert this to use the .NET app.config XML's since that is the "right" way to do configuration in .NET. If you're interested you could start this by making the Config implement IConfigurationHandler like @bdilber's ILogHandler abstracts the logging - then the current configuration class would be a "JsonConfigHandler" - https://github.com/QuantConnect/Lean/blob/master/Logging/Log.cs#L44

QANTau:
The intention was to be able to have a single configuration file that could be used for multiple purposes when running on a local instance (although no reason why this couldn't be used to work with multiple different config files when running in the cloud).  In my current tests, I have my QC login details (for Rest API Access) and broker credentials set in a central file so I don't need to have this duplicated across multiple projects.

I can adjust the code to ensure that the default config.json file name is reset after each 'custom file' read to resolve one of the issues (ie. by settings the _loaded flag to false after a 'custom file' read).

For me (personally), I think that the current mechanism of using json files is simple and I'm not sure if moving to XML adds much value.

Another option may be to create a CustomConfig class based on the current Config class to read from a non-standard file.

mchandschuh:
It seems the impetus for this change is to protect credentials from being
put into the git repository. It sounds like the easiest way to do this is
to just locally ignore that file using the gitignore. Is there a reason
this solution doesn't work?

Having multiple config files adds complexity to something that in my
opinion doesn't really need it.

As for needing to maintain multiple copies of the configuration, maybe we
can set those files up as links so only one file exists but multiple
projects reference it.

Configuration/Config.cs
```diff
@@ -83,13 +83,27 @@ private static void Initialize()
         }
         
         /// <summary>
+        /// Sets a custom file name
+        /// </summary>
+        /// <param name="customFileName"></param>
+        private static void SetCustomFileName(string customFileName)
+        {
+            _loaded = false;
```
mchandschuh: This is not thread-safe at all. Imagine the case where thread 1 comes in and requests the normal config, but is then suspended, thread 2 comes in for a custom file and sets the _config static field and is then suspended. Thread 1 resumes and reads value from the new value that thread 2 set.

mchandschuh:
Another option, why not use the regular configuration file to add an item that points to your configuration file. It seems this is 100% custom code that needs these values, so maybe an out of band solution is more appropriate.

QANTau:
No problems.  Agreed that this not thread-safe.  I currently have an entry in config.json that references an external file and was looking to use the existing methods to read the file.  I'll remove the pull request and just implement external config reader code.

PR #58, opened 2015-03-20
Some minor fixes/improvements to SubscirptionDataReader.GetReader

bdilber:
1. The `GetReader` method never checks if a file is already in the cache folder and always downloads the remote file, even if a copy exists in cache. I modified it to skip download if its cached already. (I considered this a bug, but maybe it was deliberate?)
2. The `GetExtension` method did not work if the data URL had a query string in it (e.g. “http://someurl.com/data_file.zip?somequerystr") . I modified it with a Regex pattern to correctly extract the extension in such a case. This is useful when retrieving data from S3 using the `GetPreSignedUrl` endpoint, as it returns such URLs.
3. The MD5 hashing used to generate filenames for the cache used the entire source, including the query string. This creates a problem again when using something like `GetPreSignedUrl` because Amazon changes the query string every time this method is called, and the file is never cached as a result. I modified this to strip the query string if a known extension is detected. Full URL is still used for custom data sources.

Note on file name hashing: The above method may still cause cache issues when the data source contains a known extension and a query string that changes the file contents. I think it may be best if we allowed an algorithm to provide the custom file hash along with the source, perhaps by supporting another `GetSource` signature that has the hash as an out parameter? In my use case for example, this would be useful, since I could easily use the ETAG property from S3 as the hash value. 

PS: I also added some holidays in Global.cs. National holidays on Sundays were missing. Sunday is a trading day for CME!


Engine/SubscriptionDataReader.cs
```diff
@@ -554,23 +558,29 @@ private SubscriptionStreamReader GetReader(string source)
                     // check if this is not a local uri then download it to the local cache
                     if (uri.IsAbsoluteUri && !uri.IsLoopback)
                     {
-                        try
+                        //Skip download if the file is already in cache
+                        if (!File.Exists(location))
```
mchandschuh: I suspect this is safe but would like @jaredbroad's opinion on this one.

jaredbroad: Its safe for QC financial data which is static and known but not for user's custom CSVs. They can update them on Dropbox then we need to re-download them to use the updated data. Often the files are changed daily, or if they have an error they can fix it and run the backtest again to make it work. Quandl's data updates daily as well.

mchandschuh: @bdilber I think we'll need to revert the change in behavior here since there's no way to know that the file we have is the correct file, or just has the same name. Also, the issue of holding possibly stale data comes into play.

bdilber: Ok,

What do you think about adding a `bool UseLocalCache` property to `SubscriptionDataConfig` and setting it to false by default? So the above line becomes something like:

`if (!_config.UseLocalCache || !File.Exists(location))`

This way, the default behavior would be preserved and the ability to use already downloaded files (if safe) would still be there.

mchandschuh: I think we may be able to accomplish the desired behavior without modifying the `SubscriptionDataConfig`.  Without knowing your exact use case, it seems like you're trying to download some data (possibly large?) only once, and then reference it from disk.  I think we may be able to put this logic in the `GetSource` method implemented by your custom data type.  The idea here would be that `GetSource` would check to see if your file exists, if not, then it would go and download it.  After the download completes, it can return the file path to the downloaded file.

I prefer a solution like this because it has the least impact on the system as a whole.  Does this solution fit your use case?

mchandschuh: Here's a [link](https://lean.quantconnect.com/docs#topic15.html) to the Lean docs on static vs dynamic data sources.

PR #59, opened 2015-03-22
The logic regarding the initial trade bar in TradeBarConsolidatorBase.cs:Update

willniu:
The AggregateBar() was always called before the OnDataConsolidated event was fired. This may result in an additional trade bar aggregated in TimeSpan-only mode.

The proposed logic is to aggregate the bars after firing the OnDataConsolidated event for TimeSpan-only consolidators, except when TimeSpan is TimeSpan.Zero.

Note that two test cases were updated to reflect the logic change, and an additional test case was written to verify the updated logic.

### DIFF
--- a/Tests/Common/Data/TradeBarConsolidatorTests.cs
+++ b/Tests/Common/Data/TradeBarConsolidatorTests.cs
@@ -310,7 +310,11 @@ namespace QuantConnect.Tests.Common.Data
         }
 
         [Test]
-        public void TimeSpanConsolidatorDelayedAggregationTest()
+        /// Testing the behavious where, the bar range is closed on the left and open on 
+        /// the right in time span mode: [T, T+TimeSpan).
+        /// For example, if time span is 1 minute, we have [10:00, 10:01): so data at 
+        /// 10:01 is not included in the bar starting at 10:00.
+        public void ClosedLeftOpenRightInTimeSpanModeTest()
         {
             // define a three minute consolidator 
             int timeSpanUnits = 3;
@@ -325,23 +329,24 @@ namespace QuantConnect.Tests.Common.Data
             var refDateTime = new DateTime(2014, 12, 1, 10, 00, 0);
 
             // loop for 3 times the timeSpanUnits + 1, so it would consolidate the bars 3 times
-            for (int i=1; i <= timeSpanUnits * 3 + 1 ; ++i) 
+            for (int i=0; i < 3*timeSpanUnits + 1 ; ++i) 
             {
-                refDateTime = refDateTime.AddMinutes(1);
                 consolidator.Update(new TradeBar { Time = refDateTime });
 
-                if (i <= timeSpanUnits)  // before initial consolidation happens
+                if (i < timeSpanUnits)  // before initial consolidation happens
                 {
                     Assert.IsNull(consolidated);
                 }
                 else 
                 {
                     Assert.IsNotNull(consolidated);
-                    if (i-1 % timeSpanUnits == 0) // i = 4, 7, 10
+                    if (i % timeSpanUnits == 0) // i = 3, 6, 9
                     {
-                        Assert.AreEqual(refDateTime.AddMinutes(-(timeSpanUnits-1)), consolidated.Time);
+                        Assert.AreEqual(refDateTime.AddMinutes(-timeSpanUnits), consolidated.Time);
                     }
                 }
+
+                refDateTime = refDateTime.AddMinutes(1);
             }
         }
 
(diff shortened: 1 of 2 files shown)


Tests/Common/Data/TradeBarConsolidatorTests.cs
```diff
@@ -130,11 +130,11 @@ public void OneMinuteAlwaysFiresEveryTimeOnMinueDataExceptFirstPoint()
 
             consolidator.Update(new TradeBar {Close = 2m, Time = reference.AddMinutes(1)});
             Assert.IsNotNull(consolidated);
-            Assert.AreEqual(2, consolidated.Close);
+            Assert.AreEqual(1, consolidated.Close);
 
             consolidator.Update(new TradeBar {Close = 3m, Time = reference.AddMinutes(2)});
             Assert.IsNotNull(consolidated);
-            Assert.AreEqual(3, consolidated.Close);
+            Assert.AreEqual(2, consolidated.Close);
```
mchandschuh: I'm not sure this is the behavior we want here. I agree that when it swallows the first point it isn't good, but it looks like now we're always one time unit in the past.  Before this change I would be getting the newest data immediately, but now I'm a time step behind.

I wonder if we can fix this.

willniu: It would only be a time step behind when it is TimeSpan-only mode. In
TimeSpan-only mode, we wouldn't know when the time span is fulfilled until
it's past. This is because TradeBar only has a Time, not duration. For
example, if the TimeSpan is 3 minutes, and we received 3 trade bars of
10:00:00, 10:01:00, 10:02:00, we wouldn't know whether the next trade bar
is going to be over 3 minutes. The next bars could be 10:02:01, 10:02:15,
etc. for all we know, and they should all be included in the consolidation,
as they are still within 3 minutes. Until a bar of 10:03:00 or later comes
in, we cannot tell whether the time span of 3 minutes are fulfilled.

The special case is when TimeSpan is TimeSpan.Zero, which in reality
doesn't really exist! Even an instant is not zero. But let's say,
TimeSpan.Zero exists. That just means we want to capture every single bar,
which is equivalent to maxCount = 1. In which case, we don't need to delay
the consolidation.

I hope that makes sense.

mchandschuh: OK, so with this we're saying that the bar's range is closed on the left and open on the right:
for example - [10:00 - 10:01)
So data at 10:01 is not included in the bar starting at 10:00.

I think this makes sense. Can we please update the comment in this same function to reflect the new behavior, and we'll also want to change the name of this test to match the behavior as well.

There are plans to add a Period property to TradeBar so we can distinguish how much time each bar represents, once that happens I think this case becomes trivial.

willniu: OK. See commit d4cc904. 

So, the Pull Request should include 3 commits:  0a2cab1, 60fb3ac and d4cc904.

WilliamNiuLibrary/HalfHourConsolidator.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Data.Consolidators
+{
+
+    public class HalfHourConsolidator : TradeBarConsolidator
```
mchandschuh: Is this supposed to be in this commit?

WilliamNiuLibrary/LookBackAlgorithm.cs
```diff
@@ -0,0 +1,117 @@
+﻿using System;
+using System.Linq;
+using QuantConnect.Securities;
+using QuantConnect.Data.Market;
+using QuantConnect.Data.Consolidators;
+
+namespace QuantConnect.Algorithm
+{
+    public class LookBackAlgorithm : QCAlgorithm
```
mchandschuh: Is this supposed to be in this commit?

WilliamNiuLibrary/LookBackAlgorithmHelper.cs
```diff
@@ -0,0 +1,210 @@
...
+using System.Collections.Generic;
+using QuantConnect.Securities;
+//using QuantConnect.Models;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Algorithm
+{
+
+    public class LookBackAlgorithmHelper
```
mchandschuh: Is this supposed to be in this commit?

mchandschuh:
It seems there are some files in here from your personal library. Please remove those files from the pull request.

willniu:
There you go. It should look better now. 
I'm still familiarising myself with Git/GitHub, in case it was not obvious enough ;)

mchandschuh:
Great contributions @willniu, thanks!

PR #60, opened 2015-04-02
BrokerageAuthentication Classes

QANTau:
As discussed last week, attempt #2 at building brokerage authentication code.

Note the initial seting of a values in OandaBrokerageAuthentication method; I'm assuming that we could display these on a drop-box in the web UI (instead of having the user enter the URL).

I'm aware that I haven't set intial values for the Oanda API anywhere yet.


Brokerages/BrokerageAuthentication.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace QuantConnect.Brokerages
+{
+    public abstract class BrokerageAuthentication
+    {
+        /// <summary>
+        /// Validate Brokerage Authentication Parameters
+        /// </summary>
+        /// <returns>true for OK (ie. no error)</returns>
+        public abstract bool Validate();
```
mchandschuh: We may need a way of communicating what failed validation. We could possible have this return a [ValidationResult](https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations.validationresult%28v=vs.110%29.aspx) instance instead of just true/false.

Brokerages/Brokerage.cs
```diff
@@ -57,6 +57,11 @@ public abstract class Brokerage : IBrokerage
         public abstract bool IsConnected { get; }
 
         /// <summary>
+        /// Broker Server/API Credentials 
+        /// </summary>
+        public abstract BrokerageAuthentication Credentials { get; }
```
mchandschuh: This will also need to be added to the IBrokerage interface.

QANTau:
Request cancelled (to the replaced).

PR #61, opened 2015-04-03
BrokerageAuthentication Classes

QANTau:
Changes based on feedback and tests.  Notes:
- Have a test application using the classes in QC, code is available at https://github.com/QANTau/Tools/blob/master/OandaBrokerage/Program.cs.
- The Oanda library is a modified version of their reference implementation.  Modifications are available at https://github.com/QANTau/rest-csharp-refimpl.

I looked into using ValidationResult during the Validation stage but have retained the bool return, along with adding a messages return using a StringBuilder.  ValidationResult looks very nice for checking defined objects, although I couldn't find any benefit when we are validating against entries in a Dictionary<string, string> object.

I've attached a couple of images with the output from my test application; I'm guessing that this is how the web ui would handle user input.

error
validated


jaredbroad:
Hey @QANTau! These changes are starting to take shape, nice work! I think you might be committing to this branch by mistake. I'll close this pull-request and when the implementation/branch is completed we can merge it in at that point :)

PR #67, opened 2015-04-22
Implementation of WILR indicator. Modifies MAX, MIN and AROON indicators to make WILR and AROON work.

AlexCatarino:


mchandschuh:
Hey @AlexCatarino, thanks for another contribution! It looks like the Travis CI build failed, please have a look into this: https://travis-ci.org/QuantConnect/Lean/builds/59566663.

Tests/Indicators/WilliamsPercentRTests.cs
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    Assert.AreEqual(expected, (double) ind.Current.Value, epsilon);
+                }
+                catch
+                {
+                    totalFailures++;
+                }
+            });
+            
+            Console.WriteLine("Willians%R failed {0} data points against {1} tests", totalFailures, sum);
```
mchandschuh: Does this produce totalFailures > 1 ? If so we should find out why.  The Aroon code that also did this was (I think) due to the bug in the max and min indicators. Ideally we would be able to remove try/catch from any assert code.

Indicators/QuantConnect.Indicators.csproj
```diff
@@ -45,6 +45,8 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="StochasticOscillator.cs" />
```
mchandschuh: From Travis CI: CSC: error CS2001: Source file `StochasticOscillator.cs' could not be found

Please limit pull requests to a single change. I think @kaffeebrauer is handling the stochastics already.

Indicators/AroonOscillator.cs
```diff
@@ -62,14 +62,14 @@ public AroonOscillator(int upPeriod, int downPeriod)
...
-            var max = new Maximum(name + "_Max", upPeriod);
+            var max = new Maximum(name + "_Max", upPeriod + 1);
             AroonUp = new FunctionalIndicator<IndicatorDataPoint>(name + "_AroonUp",
                 input => ComputeAroonUp(upPeriod, max, input), 
                 aroonUp => max.IsReady,
                 () => max.Reset()
                 );
 
-            var min = new Minimum(name + "_Min", downPeriod);
+            var min = new Minimum(name + "_Min", downPeriod + 1);
```
mchandschuh: do these changes permit the aroon tests to pass without the try/catch assertion code? If so we should clean that up.

mchandschuh:
Overall code looks great as usual, Looks like there's some lingering Stochastics code, please remove those files from the .csproj as someone else is working on that. Also a few comments regarding the try/catch pattern in the test assertion code. If we still need it, no worries, but ideally we could match without it.

As always, thanks for the awesome contributions!

AlexCatarino:
@mchandschuh, I made the corrections you have suggested. Travis CI pass them.

Indicators/WilliamsPercentR.cs
```diff
@@ -0,0 +1,94 @@
...
+    /// the past N days (for a given N). The value of this indicator fluctuats between -100 and 0. 
+    /// The symbol is said to be oversold when the oscillator is below -80%,
+    /// and overbought when the oscillator is above -20%. 
+    /// </summary>
+    public class WilliamsPercentR : TradeBarIndicator
+    {
+        /// <summary>
+        /// Gets the Max indicator
+        /// </summary>
+        public Maximum Max { get; private set; }
```
mchandschuh: Please spell out Maximum for the property name here

Indicators/WilliamsPercentR.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// Gets the Min indicator
+        /// </summary>
+        public Minimum Min { get; private set; }
```
mchandschuh: Please spell out Minimum for the property name here

Indicators/Minimum.cs
```diff
@@ -65,7 +65,7 @@ protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint>
                 return input.Value;
             }
 
-            if (PeriodsSinceMinimum >= Period)
+            if (PeriodsSinceMinimum >= Period - 1)
```
mchandschuh: Thank you soo much for fixing this!!

Indicators/Maximum.cs
```diff
@@ -65,7 +65,7 @@ protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint>
                 return input.Value;
             }
 
-            if (PeriodsSinceMaximum >= Period)
+            if (PeriodsSinceMaximum >= Period - 1)
```
mchandschuh: Thank you soo much for fixing this!!

mchandschuh:
Wonderful stuff @AlexCatarino, my final request is that you spell out the Maximum and Minimum property names, by convention we try not to use abbreviations in the indicators. After that I'll pull it in. As always, thanks for the awesome contributions!

Indicators/WilliamsPercentR.cs
AlexCatarino: You mean 
/// Gets the Maximum Indicator
/// Gets the Minimum Indicator
?

mchandschuh: Specifically the property name, but we can also update the comment as well, thanks!

PR #73, opened 2015-04-28
Added On Balance Volume Indicator

kaffeebrauer:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -375,6 +375,24 @@ public WilliamsPercentR WILR(string symbol, int period, Resolution? resolution =
             RegisterIndicator(symbol, williamspercentr, resolution, selector);
             return williamspercentr;
         }
+
+        /// <summary>
+        /// Creates a new On Balance Volume indicator. This will compute the
```
mchandschuh: Looks like we never got around to finishing up this doc.  This will compute the... (I'm not sure what to put, maybe steal some verbiage from [wikipedia's OBV article](http://en.wikipedia.org/wiki/On-balance_volume)

Indicators/OnBalanceVolume.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            OnBalanceVoulumeValues = new Identity("OBV");
+        }
+
+        /// <summary>
+        /// Gets a flag indicating when this indicator is ready and fully initialized
+        /// </summary>
+        public override bool IsReady
+        {
+            //get { return OnBalanceVoulumeValues.IsReady && OnBalanceVoulumeValues.Current.Value != 0; }
```
mchandschuh: We may want to remove this commented out code

Indicators/OnBalanceVolume.cs
```diff
@@ -0,0 +1,108 @@
...
+        public IndicatorBase<IndicatorDataPoint> OnBalanceVoulumeValues { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the Indicator class using the specified name.
+        /// </summary> 
+        /// <param name="name">The name of this indicator</param>
+        public OnBalanceVolume(string name)
+            : base(name)
+        {
+            OnBalanceVoulumeValues = new Identity("OBV");
```
mchandschuh: I love to see people digging into some of the more obscure classes in the indicator library, but I'm not sure we need to use the Identity indicator here. It appears as though the `OnBalanceVoulumeValues` will always have the same value as the instance of `OnBalanceVolume`.  I recommend removing the `OnBalanceVoulumeValues` and instead just using `this` in its place in this code file. External users would just use `Current.Value`.

Indicators/OnBalanceVolume.cs
```diff
@@ -0,0 +1,108 @@
...
+            //get { return OnBalanceVoulumeValues.IsReady && OnBalanceVoulumeValues.Current.Value != 0; }
+            get { return OnBalanceVoulumeValues.IsReady; }
+        }
+
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns> A new value for this indicator </returns>
+        protected override decimal ComputeNextValue(TradeBar input)
```
mchandschuh: Super clean code! Love it, thanks!

Tests/Indicators/OnBalanceVolumeTest.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        [Test]
+        public void ComparesAgainstExternalData()
+        {
+            var onBalanceVolumeIndicator = new OnBalanceVolume("OBV");
+
+            onBalanceVolumeIndicator.OnBalanceVoulumeValues.Update(
+                new DateTime(2013, 4, 30), Decimal.Parse("1.156486E+08", System.Globalization.NumberStyles.Float));
+
+            TestHelper.TestIndicator(onBalanceVolumeIndicator, "spy_with_obv.txt", "OBV",
```
mchandschuh: Nice usage of the custom asserts!

Tests/Indicators/OnBalanceVolumeTest.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            var onBalanceVolumeIndicator = new OnBalanceVolume("OBV");
+            foreach (var data in TestHelper.GetTradeBarStream("spy_with_obv.txt", false))
+            {
+                onBalanceVolumeIndicator.Update(data);
+            }
+
+            Assert.IsTrue(onBalanceVolumeIndicator.IsReady);
+
+            onBalanceVolumeIndicator.Reset();
```
mchandschuh: We need to add another line here that says `TestHelper.AssertIndicatorIsInDefaultState(onBalanceVolumeIndicator)`

mchandschuh:
Awesome contribution @kaffeebrauer.  I've made a few minor comments, all in all the code looks great. Let me know what you think. Thanks!

mchandschuh:
Looks great @kaffeebrauer, thanks for the awesome contribution!

PR #78, opened 2015-05-04
Fix DocString

dpallone:


jaredbroad:
Thanks @dpallone :) Good catch

PR #81, opened 2015-05-10
issue-79/Assure-Sell-quantity-is-negative

bizcad:
Change Buy and Sell to assure that the buy is positive and the sell is negative because they call Order and then MarketOrder which expects a buy operation to have positive quantity and sell to have negative quantity.

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -75,7 +75,7 @@ namespace QuantConnect.Algorithm
         /// <seealso cref="Order(string, double)"/>
         public int Buy(string symbol, int quantity)
         {
-            return Order(symbol, quantity);
+            return Order(symbol, Math.Abs(quantity));
         }
 
         /// <summary>
@@ -86,7 +86,7 @@ namespace QuantConnect.Algorithm
         /// <seealso cref="Order(string, double)"/>
         public int Buy(string symbol, double quantity)
         {
-            return Order(symbol, quantity);
+            return Order(symbol, Math.Abs(quantity));
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -162,7 +162,7 @@ public int Sell(string symbol, decimal quantity)
         /// <seealso cref="Order(string, double)"/>
         public int Order(string symbol, double quantity)
         {
-            return Order(symbol, (int) quantity);
+            return Order(symbol, (int) Math.Abs(quantity) * -1);
```
mchandschuh: I think we need to revert this, we want the order function to respect the input quantity

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -172,7 +172,7 @@ public int Order(string symbol, double quantity)
         /// <seealso cref="Order(string, double)"/>
         public int Order(string symbol, decimal quantity)
         {
-            return Order(symbol, (int) quantity);
+            return Order(symbol, (int) Math.Abs(quantity) * -1);
```
mchandschuh: I think we need to revert this, we want the order function to respect the input quantity

Common/Data/Market/DataDictionary.cs
```diff
@@ -12,10 +12,17 @@ public class DataDictionary<T> : IDictionary<string, T>
...
+        /// Empty Constructor
+        /// </summary>
         public DataDictionary()
         {
         }
 
+        /// <summary>
+        /// Constructor which sets teh Time to the frontier
+        /// </summary>
+        /// <param name="frontier"></param>
```
mchandschuh: Can we add comments for the parameters?  Adding them like this removes the compiler warning making it harder to find again.

mchandschuh:
Another great contribution @bizcad. I've made a few  comments, the most important one is it looks like we accidentally modified two `Order` functions. Please make the changes on the same branch and repush.

PR #82, opened 2015-05-19
Implemented the Average Directional Index Indicator with accompanying…

kaffeebrauer:
… test data and unit tests.

Implemented the Average Directional Index Indicator with accompanying
test data and unit tests.


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -417,13 +417,30 @@ public LinearWeightedMovingAverage LWMA(string symbol, int period, Resolution? r
         public OnBalanceVolume OBV(string symbol, Resolution? resolution = null,
             Func<BaseData, TradeBar> selector = null)
         {
-            string name = CreateIndicatorName(symbol, "OBV", resolution);
+            var name = CreateIndicatorName(symbol, "OBV", resolution);
             var onBalanceVolume = new OnBalanceVolume(name);
-            RegisterIndicator(symbol, onBalanceVolume, resolution);
+            RegisterIndicator(symbol, onBalanceVolume, resolution, selector);
```
mchandschuh: great fix! thanks!!

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -417,13 +417,30 @@ public LinearWeightedMovingAverage LWMA(string symbol, int period, Resolution? r
...
             return onBalanceVolume;
         }
 
         /// <summary>
+        /// Creates a new Average Directional Index indicator. 
+        /// The indicator will be automatically updated on the given resolution.
+        /// </summary>
+        /// <param name="symbol">The symbol whose Average Directional Index Average Directional Index we seek</param>
```
mchandschuh: looks like there's a little double speak double speak here :) - please remove the extra 'Average Direction Index'

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -417,13 +417,30 @@ public LinearWeightedMovingAverage LWMA(string symbol, int period, Resolution? r
...
+        /// <param name="resolution">The resolution.</param> 
+        /// <param name="period">The period over which to compute the Average Directional Index</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value).</param>
```
mchandschuh: ...'if null defaults to ( x => (TraderBar)x)'

Since the ADX accepts trade bar instances we don't default the selector to the .Value property of BaseData, but instead use a caste to TradeBar, the assumption here is that if you're using the ADX then you're receiving trade bar data

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    /// This indicator computes Average Directional Index which measures trend strength without regard to trend direction.
+    /// Firstly, it calculates the Directional Movement and the True Range value, and then the values are accumulated and smoothed
+    /// using a custom smoothing method proposed by Wilder. For an n period smoothing, 1/n of each period's value is added to the total period.
+    /// From these accumulated values we are therefore able to derived the 'Positive Directional Index' (+DI) and 'Negative Directional Index' (-DI)
+    /// which is used to calculate the Average Directional Index.
```
mchandschuh: Great documentation, thanks!!

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        private IndicatorBase<IndicatorDataPoint> SmoothedTrueRange { get; set; }
+
+        /// <summary>
+        /// Gets or sets the index of the Plus Directional Indicator
+        /// </summary>
+        /// <value>
+        /// The  index of the Plus Directional Indicator.
+        /// </value>
+        public IndicatorBase<IndicatorDataPoint> PositiveDirectionalIndex { get; set; }
```
mchandschuh: I think we may want this as a private set

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        /// <summary>
+        /// Gets or sets the index of the Minus Directional Indicator
+        /// </summary>
+        /// <value>
+        /// The index of the Minus Directional Indicator.
+        /// </value>
+        public IndicatorBase<IndicatorDataPoint> NegativeDirectionalIndex { get; set; }
```
mchandschuh: I think we may want this as a private set

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        /// <summary>
+        /// Computes the Plus Directional Indicator (+DI period).
+        /// </summary>
+        /// <returns></returns>
+        private decimal ComputePositiveDirectionalIndex()
+        {
+            if (SmoothedTrueRange == 0) return new decimal(0.0);
+
+            var positiveDirectionalIndex = (SmoothedDirectionalMovementPlus.Current.Value / TrueRange.Current.Value) * 100;
```
mchandschuh: Do we ever need to worry about TrueRange == 0m?

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+
+        /// <summary>
+        /// Computes the Minus Directional Indicator (-DI period).
+        /// </summary>
+        /// <returns></returns>
+        private decimal ComputeNegativeDirectionalIndex()
+        {
+            if (SmoothedTrueRange == 0) return new decimal(0.0);
+
+            var negativeDirectionalIndex = (SmoothedDirectionalMovementMinus.Current.Value / TrueRange.Current.Value) * 100;
```
mchandschuh: Do we ever need to worry about TrueRange == 0m? I notice we check the SmoothedTrueRange, maybe it should b checking just the TrueRange?

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+        {
+            decimal value;
+
+            if (Samples < period)
+            {
+                value = SmoothedTrueRange.Current + TrueRange.Current;
+            }
+            else
+            {
+                value = SmoothedTrueRange.Current - (SmoothedTrueRange.Current / period) + TrueRange.Current;
```
mchandschuh: Maybe our current implementation of Wilder's smoothing is incorrect. I think this is what he used to do, SMA until the period is full and then the kind of exponential corrector for each new point

Indicators/AverageDirectionalIndex.cs
```diff
@@ -0,0 +1,332 @@
...
+            DirectionalMovementPlus.Update(input);
+            DirectionalMovementMinus.Update(input);
+            SmoothedTrueRange.Update(Current);
+            SmoothedDirectionalMovementMinus.Update(Current);
+            SmoothedDirectionalMovementPlus.Update(Current);
+            PositiveDirectionalIndex.Update(Current);
+            NegativeDirectionalIndex.Update(Current);
+            var diff = Math.Abs(PositiveDirectionalIndex - NegativeDirectionalIndex);
+            var sum = PositiveDirectionalIndex + NegativeDirectionalIndex;
+            var value = Math.Round((sum == 0 ? 50 : ((_period - 1) * Current.Value + 100 * diff / sum) / _period), 5);
```
mchandschuh: Is rounding a part of the definition? I typically like to leave these unrounded, allows the caller to decide rounding behavior.

Indicators/QuantConnect.Indicators.csproj
```diff
@@ -11,6 +11,8 @@
     <AssemblyName>QuantConnect.Indicators</AssemblyName>
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
+    <RestorePackages>true</RestorePackages>
```
mchandschuh: What is this restore packages flag for? Nuget packages should have already been set to restore on build. Was this required?

Indicators/QuantConnect.Indicators.csproj
```diff
@@ -97,6 +100,13 @@
     <None Include="packages.config" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
+    <PropertyGroup>
+      <ErrorText>This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
```
mchandschuh: Was this nuget change required as well?

Tests/Indicators/AverageDirectionalIndexTests.cs
```diff
@@ -0,0 +1,36 @@
+﻿using NUnit.Framework;
```
mchandschuh: Please add the QuantConnect license banner

mchandschuh:
Awesome stuff @kaffeebrauer! I've made various comments, let me know what you think. I love to see people making good use of the Functional indicator class. My only real concern is the csproj changes relating to nuget, I'm not sure that they're required and I recall setting it up a certain way to keep travis CI happy.

Indicators/AverageDirectionalIndex.cs
kaffeebrauer: I think I'll only calculate this if there is a previous input

Indicators/AverageDirectionalIndex.cs
kaffeebrauer: I think I'll only calculate this if there is a previous input

Algorithm/QCAlgorithm.Indicators.cs
kaffeebrauer: Haha, double copy and paste exception thrown.

Tests/Indicators/AverageDirectionalIndexTests.cs
kaffeebrauer: Corrected, also added the banner to previous test classes that weren't included.

Indicators/QuantConnect.Indicators.csproj
kaffeebrauer: Probably shouldn't have done that and checked that in.

Indicators/AverageDirectionalIndex.cs
kaffeebrauer: I think it's not necessary.

Indicators/AverageDirectionalIndex.cs
mchandschuh: I believe you're correct, but imagine the case where we get a flat line stock, with open/high/low/close all the same value, would true range=0 in this case?

PR #86, opened 2015-05-22
Issue 85 - Add FisherTransform indicator

bizcad:
Added Fisher Transform Indicator


mchandschuh:
Hey Nick, can we split these indicators into separate pull requests? We like to limit pull requests to a single feature. Also, can we add some units tests for the new indicators? Thanks!

bizcad:
Inverse Fisher and Relative Vigor Index removed.  I will put them each in a separate fork/pull.

I tested  by comparing the result with an Excel Spreadsheet.  The spreadsheet was created by logging the OHLCV from the OnBar event handler in an algo and then calculating the Fisher using Excel and comparing it to the Current for the indicator. 

The algo is part of my development version of Lean.  This Lean is a repo in github/Bizcad called LeanWMA.  I use it to develop features so that I do not pollute the main QuantConnet/Lean repo with my test algos.

 I cannot think of a way of reproducing that in a test.  

Is there a way to import data from the Excel file or csv file and use the bars for the test?  Is there a way to make the Excel file accessible to you so you can see the computation?

mchandschuh:
I would recommend adding the .csv data to the QuantConnect.Tests\TestData folder. In that folder are several text files we use for validating indicator implementations. If you add your data there, then there are a few helper functions I've written to either read or auto-validate the indicator against the test data.

Check out [TestHelper.cs](https://github.com/QuantConnect/Lean/blob/master/Tests/Indicators/TestHelper.cs) for the helper functions I'm referencing. Then you can look at the indicator test classes who use these helper functions.

Ideally, if you can format your csv to follow the other test data files (Time, Open, High, Low, Close, Volume, [indicator output columns]) then you could use [this helper function](https://github.com/QuantConnect/Lean/blob/master/Tests/Indicators/TestHelper.cs#L58).

If you run into any issues just give me a shout!

bizcad:
That should work.  I will take a look at TestHelper.  Hold off on the merge until I get the tests  written please.

bizcad:
I had a problem with running the test, but I fixed it.  The TestHelper assumes that Close is going to be sent to the indicator and that the column will be named Close.  In the case of the Fisher Transform, whose purpose is to normalize the data, the input is an IndicatorDataPoint with some other indicator as the value.   The idea is to accentuate the peaks and valleys of the indicator to give a clearer trading signal.  

In my experimental development, I was using the RVI indicator and wanted to make the buy/sell points clearer.  However, that indicator is not available as yet, because it will be coming in my next couple of pull requests. So in my test I used the MACD as the input instead. 

So I added a method to TestHelper which takes an extra column as the parameter called, rather verbosely, customIndicatorFeedColumnName.  It becomes customColumnIndex locally to the function and is used as input instead of the "close".  The test then works.  

Oddly enough, the ResetsProperly still works even though it is calling for the Close column through GetTradeBarStream.  The code does not care what the values are when it is checking for a Reset.

I also added a couple of try/catch blocks to the TestIndicator function to try to pinpoint better where errors occur: either in the parsing of the date or the "Close".  These checks will do a better job of catching things like all comma lines or poorly formatted Dates.

<pre>
                decimal close;
                try
                {
                    close = decimal.Parse(parts[closeIndex], CultureInfo.InvariantCulture);
                }
                catch (Exception ex)
                {
                    throw new Exception(ex.Message);
                }
                DateTime date;
                try
                {
                    date = Time.ParseDate(parts[0]);
                }
                catch (Exception ex)
                {
                    throw new Exception(ex.Message);
                }
</pre>

I will push the changes along with the test as soon as I migrate from my development project to my origin and github/Bizcad.  That should move them into the PR.

I very much appreciate your writing that TestHelper class.  It really cut down the amount of work I had to do and made debugging the tests a lot clearer.  Fabulous work.  My hat is off to you!

bizcad:
A great question and one I had not considered before.  What is your preferred method of handling problems like a divide by zero error.  Naturally, code should be bug free and tested so that it does not happen.  That is the purpose of testing.  Since I usually code TDD, I usually put a lot of try/catches while I am developing so that my testing is more focused.

My answer to the first question is that the value add is that it is easier to track down a problem with a Date or value by throwing the exception.  Then I can use using the ReSharper test runner with a break point at the exception.  I could then inspect the local variables to see what the problem was.  Since the code is a unit test, it would not be run in production. Exception throws are an efficient way to see if there is a problem with your test data or code. (I had comma only rows and a divide by 0 error in my code).  I can remove them if you like, but I found them useful.  

I also wanted the stack trace to locate the line where the problem occurred.  If the throw is passed up the call chain it is not so easy to find out where it occurred because, as the stackoverflow says, the stack trace is destroyed resulting in the line referred to in the calling exception is the line which handles the exception in the lower level routine.  

As I said, it is your call.  I just found it easier to locate problems in the TestHelper class.

As to normal operation.  I have read that one should never use generic Exception, but specify what kind of Exception is being handled.  Likewise,  as stackoverflow points out, a straight throw without any added information about code specific problems and possibly diagnostic information is pretty useless.  I will be glad to remove the try/catch blocks in the Indicator.  They were entirely defensive coding and should never happen.  However, you bring up a really good question.

What is your preferred method of handing deal killing problems in production code?  Code does fail in production.  Application Insights might be a good way to handle production code fails.

I hated the way the backtest engine handled problems on the web site.  Here is why.  A couple of weeks ago, I copied the code for one of the sample algos and ran the backtest.  No code changes.  I just wanted to see what the output of the algo looked like so I could use it as a basis for further development.  The next day after making its way through the queue, the backtest engine just returned a message that the backtest failed.  There was no information about why.  I suspect the engine may have changed since the sample algo was written, but had no way to find out..  The positive spin is that the problem caused me to download the Lean project so that I could test my algos in debug mode in VS before I submitted them to the on line backtest engine using your historical data.

Once again I want to let you know that you are doing an awesome job.  I absolutely love your code reviews.  Sorry for being so verbose.

bizcad:
I pawed my way through the Travis log to see why the Travis CI build failed.  It appears the test data file was not in the Test/bin/Debug/TestData folder.  I think it is being excluded by git on my push to my origin/fork.  I do not know how do you get the files migrated from Tests/TestData to the bin folder on your end.

mchandschuh:
@bizcad - good discussion leads to gooder software :)

Apologies if the following seems disjointed :) Lots going on today.

### Divide by zero
In the indicator library I try to perform limit analysis to come up with a reasonable value in the divide by zero case. So what I'll do is simple check for zero. An example that stands out is the [RelativeStrengthIndex](https://github.com/QuantConnect/Lean/blob/master/Indicators/RelativeStrengthIndex.cs).

I prefer to allow the 'exceptional' case to throw, otherwise it will end up hiding until something more tragic happens.  I try to stay away from the [pokemon](http://c2.com/cgi/wiki?PokemonExceptionHandling) exception handling. If it is a case we can predict, such as divide by zero, my preference is to capture it and handle it accordingly. If it is truly 'unhandle-able' then throwing is the correct behavior.

### Value added
If you're using visual studio there's a super handy feature in the Debug -> Exceptions menu. It will open up a bunch of check boxes. Check the CLR Exceptions thrown to tell the debugger to break whenever there's an exception thrown. This is unbelievably useful when debugging code that should *never* throw exceptions, especially if sometimes they're 'caught'.

I guess what I'm missing (and maybe it's because I didn't experience what you did with the TestHelper) is if an exception is thrown (and un-caught) it should tell you the message and the full stack trace (line numbers included) in the output of the unit test. Re-wrapping the exception and re-throwing it as a new exception only (in my opinion) distorts the actual exception being thrown. However, you bring up a good point about the ease of use placing the breakpoint in the catch block.

I would prefer we remove the try/catch/throw new Exception pattern from test and indicator code.  Give the Debug -> Exceptions menu a try and let me know how you make out with that, I think it will provide the same behavior you seek.

### Runtime errors in production code
Very interesting topic here, and I'm not 100% as for where I stand. Typically my intuition is fail fast. If the software encounters some state that it doesn't expect I would prefer it self terminate and dump necessary information to the log so it can be reproduced and resolved. Now the other side of this argument is that we're live trading and would like to maintain a certain level and stability over the long run. So if the object encounters something that puts it into an invalid, unrecoverable state, termination is the only correct option (in my mind). If the instance is still in a valid state then it may be OK to log and swallow the error. I think it's very much dependent on the exact situation.

The engine is designed to return run time errors to the user's browser. I even spent some time last week implementing line numbers in the stack traces (not as easy as I thought!), but there were some severe performance implications. I'm thinking maybe we'll make a flag to allow users to specify for it to run in the debug configuration.  The correct way for algorithm components to convey that they've encountered an unhandled exception is to allow the exception to throw and the AlgorithmManager will catch it and terminate the algorithm properly.

### Final note
I want to be completely honest with you Nick, I love your level of activity in the LEAN community. Discussions like these help to propel the product towards wider acceptance and stability. We at QuantConnect greatly appreciate your time and contributions!

Indicators/InverseFisherTransform.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// It is useful because it normalizes the current input, such as price, to fit within the range of the 
+    /// High and Low of the data window.
+    /// 
+    /// For a discussion of how it is used in trading to clarify other indicators turning points
+    /// see http://www.tradingsystemlab.com/files/The%20Inverse%20Fisher%20Transform.pdf
+    /// 
+    /// The formula comes from one of the activation functions of a neural network 
+    /// 
+    /// </summary>
+    public class InverseFisherTransform : WindowIndicator<IndicatorDataPoint>
```
mchandschuh: I would prefer if we submitted indicators in their own pull requests.

Indicators/InverseFisherTransform.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// value1 is centered on its midpoint and then doubled so that value1 wil swing between -1 and +1.  
+        /// value1 is also smoothed with an exponential moving average whose alpha is 0.33.  
+        /// 
+        /// Since the smoothing may allow value1 to exceed the _period day price range, limits are introduced to 
+        /// preclude the transform from blowing up by having an input larger than unity.
+        /// </summary>
+        /// <param name="window">The IReadOnlyWindow of Indicator Data Points for the history of this indicator</param>
+        /// <param name="input">IndicatorDataPoint - the time and value of the next price</param>
+        /// <returns></returns>
+        protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint> window, IndicatorDataPoint input)
```
mchandschuh: do we use the 'window' parameter? If not maybe this class shouldn't derived from WindowIndicator

Indicators/InverseFisherTransform.cs
```diff
@@ -0,0 +1,124 @@
...
+                // get some local variables
+                var price = input.Value;
+                var minL = _minLow.Current.Value;
+                var maxH = _maxHigh.Current.Value;
+
+                // get the value1 from the last time this function was called
+                var v1 = value1[0].Value;
+
+                // compute the EMA of the price and the last value1
+                value1.Add(new IndicatorDataPoint(input.Time, .33m * 2m * ((price - minL) / (maxH - minL) - .5m) + .67m * v1));
```
mchandschuh: We maybe be able to utilize the ExponentialMovingAverage type instead of manually doing this here. There's an overload of the constructor that accepts the ema constant

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// is less than half wavelength of the dominant cycle.
+    /// 
+    /// If the dominant cycle period is not available, you can sum the RVI components
+    /// over a fixed default period.  A period of 8 or 10 is suggested because most 
+    /// cycles of interest are 16 or 20 bars long.
+    /// 
+    /// A crossing of the indicator with the indicator with a lag of 1 bar is a good way 
+    /// to create an unambiguous signal.
+    /// </summary>
+    public class RelativeVigorIndex : TradeBarIndicator
```
mchandschuh: I would prefer if we submit indicators in their own pull requests

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,133 @@
...
+    {
+        private RollingWindow<IndicatorDataPoint> value1;
+        private RollingWindow<IndicatorDataPoint> value2;
+        /// <summary>
+        /// 
+        /// </summary>
+        public RollingWindow<IndicatorDataPoint> RviWindow { get; set; }
```
mchandschuh: missing doc

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <summary>
+        /// 
+        /// </summary>
+        public RollingWindow<TradeBar> Bars { get; set; }
```
mchandschuh: missing doc

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            // Save the bar for summing
+            Bars.Add(input);
```
mchandschuh: Instead of manually doing this you may want to derive from WindowIndicator<TradeBar>

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,133 @@
...
+                value1.Add(new IndicatorDataPoint(input.EndTime, v1));
+
+                // Filter the High - Low with a four-bar symetical FIR filter before the terms are summed.
+                var v2 = ((Bars[0].High - Bars[0].Low) + 2 * (Bars[1].High - Bars[1].Low) +
+                          2 * (Bars[2].High - Bars[3].Low) + (Bars[3].High - Bars[3].Low) / 6);
+                value2.Add(new IndicatorDataPoint(input.EndTime, v2));
+
+                // The numerator and denominator are summed independently
+                decimal num = value1.Sum(point => point.Value);
+                decimal denom = value2.Sum(point => point.Value);
```
mchandschuh: It would be more performant if we used the Sum indicator type here so we don't need to recompute the whole thing each time.

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,133 @@
...
+                try
+                {
+                    // The RVI is the ratio of the numerator to the denominator.  Since
+                    //  they are lagged by the same amount, due to the filtering,
+                    //  the lag is removed by taking the ratio.
+                    RviWindow.Add(new IndicatorDataPoint(input.EndTime, num / denom));
+                }
+                catch (DivideByZeroException zex)
+                {
+                    throw new Exception(zex.Message + zex.StackTrace);
```
mchandschuh: catching and rethrowing exceptions adds little and hides the original stack trace. Please remove the try/catch/rethrow new exception.

Tests/Indicators/FisherTransformTests.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace QuantConnect.Tests.Indicators
+{
+    [TestFixture]
+    public class FisherTransformTests
+    {
+        [Test]
+        public void ComputesCorrectly()
+        {
+            var fish = new FisherTransform(8);
+            TestHelper.TestIndicator(fish, "spy_fisher_minutes.txt", "MacD", "ExcelFish", (ind, expected) => Assert.AreEqual(expected, (double)ind.Current.Value, 1e-3));
```
mchandschuh: awesome! thanks!! love tests!!

Tests/Indicators/TestHelper.cs
```diff
@@ -117,7 +116,63 @@ public static void TestIndicator(IndicatorBase<IndicatorDataPoint> indicator, st
...
+                            customColumnIndex = i;
+                        }
+                        if (parts[i].Trim() == targetColumn)
+                        {
+                            targetIndex = i;
+                        }
+                    }
+                    if (customColumnIndex * targetIndex < 0)
+                    {
+                        Assert.Fail("Didn't find one of 'Close' or '{0}' in the header: " + line, targetColumn);
```
mchandschuh: This doesn't appear to search for 'Close'

Tests/QuantConnect.Tests.csproj
```diff
@@ -222,6 +223,9 @@
     <Content Include="TestData\spy_bollinger_bands.txt">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
+    <Content Include="TestData\spy_fisher_minutes.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
mchandschuh: great stuff, this got travis-ci to pass, thank you.

PR #89, opened 2015-05-29
Questions about Roundup

cub-:
line 314, the equation divide d.Ticks and then multiple it again? Or should it just be the reverse of "RoundDown" method:
return time.AddTicks(time.Ticks / d.Ticks);
Could anyone look into it?

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -311,7 +311,8 @@ namespace QuantConnect
                 // divide by zero exception
                 return time;
             }
-            return new DateTime(((time.Ticks + d.Ticks - 1) / d.Ticks) * d.Ticks);
+            //return new DateTime(((time.Ticks + d.Ticks - 1) / d.Ticks) * d.Ticks);
+            return time.AddTicks(time.Ticks / d.Ticks);
         }
 
         /// <summary>


Common/Extensions.cs
```diff
@@ -311,7 +311,8 @@ public static DateTime RoundUp(this DateTime time, TimeSpan d)
                 // divide by zero exception
                 return time;
             }
-            return new DateTime(((time.Ticks + d.Ticks - 1) / d.Ticks) * d.Ticks);
+            //return new DateTime(((time.Ticks + d.Ticks - 1) / d.Ticks) * d.Ticks);
+            return time.AddTicks(time.Ticks / d.Ticks);
```
mchandschuh: is this intended to be a modulus operation instead of division?

mchandschuh:
I've made a comment, it looks like the math in your PR isn't exactly correct. I think you intend for a modulus operation vs division. Also, please include unit tests with pull requests, thanks for the contribution!

cub-:
That's right. It should be "%" instead of "/". 
Honestly, I haven't done any unit test before. But I'll read some examples of others' PR first and learn from it.

mchandschuh:
You can also just take a peak at the QuantConnect.Tests project, there are several examples of using nUnit in there. It's fairly easy to use.

Issue #90, opened 2015-05-31
TradeBarConsolidator should use rounded times


mchandschuh:
Currently the TradeBarConsolidator (and derived types) use the input data times directly, so it's common to get a daily bar that starts at 9:31am for equities. We should modify this to start a midnight like daily data is typically marked. Ideally this could be implemented generically by rounding the 'Time' on each bar, so 15 minute bars would be on 00/15/30/45, hour bars always on 00, ect..

Issue #91, opened 2015-06-01
Add public property WorkingBar to TradeBarConsolidatorBase<T>


StefanoRaggi:
Hi everyone, 
I am evaluating Lean for my algo development and so far I like it very much.

In my first intraday algorithm I am using the TickConsolidator to build minute bars from tick data and I need to access the current bar (in progress) from the OnData(ticks) method.
Unfortunately the current consolidating bar is accessible only in the DataConsolidated event which is fired only at the end of the chosen period. 

Adding the following in TradeBarConsolidatorBase.cs would be enough for me:

        /// <summary>
        /// Gets the current 'workingBar'.
        /// </summary>
        public TradeBar WorkingBar
        {
            get { return _workingBar; }
        }

Am I right or is there a better way and if ok do you need a pull request for this small change ?

Thanks
Stefano

jaredbroad:
Welcome @StefanoRaggi! Thank you for the suggestion, and I'm glad you're enjoying Lean!

That suggestion looks great; but perhaps the only caveat would be that it allows the user to modify the state of the working bar - and corrupt the potential output. Perhaps a better solution would be to return a clone of the working bar, making the Consolidator more robust.

Cloning a working bar is pretty simple with our underlying extensions: you just need to use baseData.Clone().

    /// <summary>
    /// Gets the current 'workingBar'.
    /// </summary>
    public TradeBar WorkingBar
    {
        get { return _workingBar.Clone(); }
    }

StefanoRaggi:
I agree 100% with your suggestion to clone the bar, I only needed read access to it anyways.

I am currently working on the Consolidator (I found a bug in tick aggregation due to issue #90, so will fix both at the same time).

Thanks
Stefano

PR #92, opened 2015-06-03
Fixed bug when aggregating ticks + fixed issues #90 and #91

StefanoRaggi:
With this update the TickConsolidator now aggregates ticks correctly (previously ticks were increasingly shifted forward - to the following bars).

### DIFF
--- a/Common/Data/Consolidators/BaseDataConsolidator.cs
+++ b/Common/Data/Consolidators/BaseDataConsolidator.cs
@@ -73,7 +73,7 @@ namespace QuantConnect.Data.Consolidators
                 workingBar = new TradeBar
                 {
                     Symbol = data.Symbol,
-                    Time = data.Time,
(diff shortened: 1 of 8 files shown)


Common/Data/Consolidators/BaseDataConsolidator.cs
```diff
@@ -73,7 +73,7 @@ protected override void AggregateBar(ref TradeBar workingBar, BaseData data)
                 workingBar = new TradeBar
                 {
                     Symbol = data.Symbol,
-                    Time = data.Time,
+                    Time = GetRoundedBarTime(data.Time),
```
mchandschuh: Absolutely wonderful, I was thinking of this same solution. Many thanks for the implementation!

mchandschuh:
Thanks for the amazing pull request @StefanoRaggi!!

PR #94, opened 2015-06-06
Replaced constant DataFolder with setting in config.json

StefanoRaggi:
This allows us to set the location of local data files (instead of using the fixed lean/data folder)

### DIFF
--- a/Algorithm/Examples/BubbleAlgorithm.cs
+++ b/Algorithm/Examples/BubbleAlgorithm.cs
@@ -235,11 +235,12 @@ namespace QuantConnect
         /// <summary>
         /// Return the URL string source of the file. This will be converted to a stream
         /// </summary>
-        /// <param name="datafeed">Type of datafeed we're reqesting - backtest or live</param>
         /// <param name="config">Configuration object</param>
         /// <param name="date">Date of this source file</param>
+        /// <param name="datafeed">Type of datafeed we're requesting - backtest or live</param>
+        /// <param name="dataFolder">The root directory of the data folder for this application</param>
         /// <returns>String URL of source file.</returns>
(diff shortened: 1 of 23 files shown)


mchandschuh:
Great catch with the failing tests and thanks for the fix!

@StefanoRaggi, while I agree with the proposed solution (pass dependencies/data into methods), it will end up breaking 1,000's of user algorithms. We must make considerations for existing user code when dealing with breaking changes like these. Since this change affects the BaseData.GetSource method, any user algorithm that derives from BaseData will be broken.

Maybe there's another way to achieve the end goal of this change?

mchandschuh:
Failing tests fixed in https://github.com/QuantConnect/Lean/commit/9227767437bdb312a309f06d4f019567e22a0e78

StefanoRaggi:
OK, I didn't think the impact of changing the signature of the GetSource method was so high. I guess I was wrong. No problem, I'll try to find another way.

mchandschuh:
I know it's similar to what was there currently, but possibly we can read in the configured value at application start and make that available in a similar manner as it is currently.

Were you trying to be able to read from multiple data directories? Or just one directory in a non-default location?

StefanoRaggi:
Just one directory read from the config.json would be fine.

Do you mean changing Constants.DataFolder from a constant to a property-get ?

mchandschuh:
Yes, I think that would satisfy both of our requirements

StefanoRaggi:
Change is here: https://github.com/StefanoRaggi/Lean/commit/360af61b241dd194125e901ec6a393116eac50d4

mchandschuh:
Code looks great, you can submit a PR or I can manually merge. Thanks!

StefanoRaggi:
New PR submitted.

PR #98, opened 2015-06-08
Live tick data support for IB

arcturius:
Hi, added support for live tick data. Tested with Equities, FX. Please test on your side as well. 

Not experienced in git though so if you see git typos/misuse feel free to help :) 

Pavel

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
  * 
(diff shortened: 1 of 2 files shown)


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1179,6 +1186,220 @@ private bool IsWithinScheduledServerResetTimes()
...
+        /// <summary>
+        /// IDataQueueHandler interface implementaion 
+        /// </summary>
+        /// 
+        public IEnumerable<Data.BaseData> GetNextTicks()
+        {
+            var ticks = new List<Tick>();
+            Tick tick;
+
+            while (_ticks.TryDequeue(out tick))
```
mchandschuh: we may want to limit the maximum number of ticks being dequeued to 100 or 1000, just so we can gaurantee that this function actually returns, I'm considering the case where we're able to push ticks in at a similar rate to them being pulled off, possibly occurring from subscribing to a myriad of symbols.

mchandschuh:
Awesome contribution @arcturius! One thought that came up is how we're emitting the IB ticks on every update. My concern is the volume may become misleading. I'm trying to think of a moving VWAP indicator and how it would respond to getting multiple ticks with the same volumes.

For example, IB sends updates for the bid, then we emit a tick. IB sends an update for the ask, we emit a tick. Both of these ticks will reflect the same volume.  Do you think this may be a problem?

If this may be an issue, does it make sense to emit on volume updates only?  If you think this is OK, just let me know. Once again, thanks for the great contribution!

arcturius:
Why both ticks should have the same volume? those are buyers and sellers. there is an update for price and size in IB API (no idea why). if you run the code, the ticks are generated one after another: bids, asks, trades.

mchandschuh:
My thought was that it looks like we emit a new 'tick' after each individual update streams in (price/volume), but I'll verify this thought.

Thanks for the response, I'll bring the code down and run it tonight to confirm or deny my thoughts. It was just a concern that came to mind while reading the code. It sounds like you've tested it thoroughly though and should work as advertised. Thanks for the due diligence!

mchandschuh:
This was merged in https://github.com/QuantConnect/Lean/commit/44fe66b881ea981e8467e95c0c2675b1a49706b3. with a minor change to only emit ticks when we receive volume data, without this change we were getting disjoint ticks (some with 0 volume, some duplicate price ticks, ect...)

PR #99, opened 2015-06-08
Constants.DataFolder is now a get property

StefanoRaggi:
that reads the path from config.json


mchandschuh:
Thanks for the great contribution!

PR #102, opened 2015-06-10
fix for ticks in IB connector

arcturius:
I think that previous merge was not good:

* in terms of ticks accumulating in the _ticks you were right - GetNextTicks could loop infitintely in the high load. it is fixed now.

* as for your modification of HandleTickPrice, it is not correct. Just imagine what would happen if as a market participant you are making the market and you have the standing limit buy order at 1,5 size 10 (that is bid in the market depth for other participants), and you then moved your order to 1.4. This case wont be reflected in your system at all since size has not changed. so the information is getting lost. This is fixed now.

I defined a valid tick in the system: the one that either is a nonzero trade, or ask or bid. I admit that your Tick class is designed to hold one of those, but not all at once.

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -1204,13 +1204,12 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
         /// 
         public IEnumerable<Data.BaseData> GetNextTicks()
         {
-            var ticks = new List<Tick>();
-            Tick tick;
-
-            while (_ticks.TryDequeue(out tick))
-                ticks.Add(tick);
-
-            return ticks;
+            lock (_ticks)
+            {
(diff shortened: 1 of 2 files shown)


mchandschuh:
I understand your concerns. It seems like we're using the 'Quantity' for multiple purposes. I think it would be best if we add a BidSize and AskSize field to the Tick class. Maybe longer term we can split tick into trade and quote since it has several mutually exclusive fields.

mchandschuh:
I went to add the bidsize/asksize we spoke about earlier today and my closer attention was brought to this 'bSend' flag in the HandleTickSize function. Could you briefly explain the logic here for me? Thanks!

arcturius:
Updated HadnletickSize code to support lastSize info from IB

arcturius:
Hello?

mchandschuh:
Thanks!

mchandschuh:
Opened #118 to look at adding bid/ask size

PR #112, opened 2015-06-13
Faster TradeBar Clone method

StefanoRaggi:
The BaseData.Clone() method uses Reflection for object cloning which is very slow. 

The TradeBar class does not have any reference type fields, so a fast shallow copy using MemberwiseClone() can be safely used.

This change alone gives a BenchmarkAlgo speed improvement (from 251k to 265k points/sec.)

In general, it would be better if possible to avoid using reflection at all.
The faster way of doing a deep copy of any BaseData-derived object (assuming it has reference type fields in it) is to call MemberwiseClone() first and then copy the reference type fields manually (in a Clone() override).

### DIFF
--- a/Common/Data/Market/TradeBar.cs
+++ b/Common/Data/Market/TradeBar.cs
@@ -349,5 +349,11 @@ namespace QuantConnect.Data.Market
             var source = Path.Combine(Constants.DataFolder, securityTypePath, resolutionPath, symbolPath, filename);
             return new SubscriptionDataSource(source, SubscriptionTransportMedium.LocalFile);
         }
+
+        public override BaseData Clone()
+        {
+            return (BaseData)MemberwiseClone();
+        }
+
     } // End Trade Bar Class
 }


mchandschuh:
This is good, maybe it can be implement on base data like this?  One thing to be tested is whether or not the reference string value is copied, it may need to be manually copied. If implemented on base data than all derived types, ticks included, can get the benefit and have the same behavior.

StefanoRaggi:
I thought about replacing the implementation directly in the BaseData class, but this would mean every complex class derived from it (for example DynamicData and its derivations), would risk making only a shallow clone instead of a deep one. 

For classes inside Lean, no problem, but what about user-defined classes ? They would need to override Clone() and implement it correctly, so this would be a breaking change. This is the reason why I only changed the TradeBar Clone.

As for strings, they are a reference type, but they are immutable, so it shouldn't be a problem.

StefanoRaggi:
Cloning strings:

class Person 
{
    public int Id;
    public string Name;
}

var a = new Person { Id = 1, Name = "xxx" ];
var b = a.Clone();
b.Name = "yyy";
Console.WriteLine(a.Name);    // prints "xxx"
Console.WriteLine(b.Name);    // prints "yyy"

AnObfuscator:
> As for strings, they are a reference type, but they are immutable, so it shouldn't be a problem.

Absolutely correct.

StefanoRaggi:
I forgot the most important line in the code above (the one with the Clone method). Added.

mchandschuh:
Great point about dynamic data, that will need it's own implementation.  I was thinking tick would also benefit from the speed improvement.  Maybe we can just do the same there as well if it isn't already.

jaredbroad:
Thank you @StefanoRaggi, nice commit!

StefanoRaggi:
@mchandschuh The Tick class already has an override for Clone and it uses the copy constructor.

I did a quick test calling Clone() 1 million times in a for loop on an instance of the Tick class:

Copy constructor: 35 ms
MemberwiseClone(): 55 ms
Reflection: 1025ms

The results confirm that reflection is slow! 

Btw, it seems that the reflection implementation called by BaseData.Clone does not do a deep clone, so we could use the copy constructor in all internal Lean classes needing Clone, except for the abstract DynamicData class that should use MemberwiseClone (inherited by all user derived classes).

When I have some time I'll try this change (maybe writing some unit tests first).

PR #113, opened 2015-06-13
Optimization in DataStream.GetData

StefanoRaggi:
This method is performance-critical, so even the simple caching of the EndTime property value can give a speed improvement.

Benchmark Algo: from 265k to 273k points/sec


StefanoRaggi:
Sorry for the double commit, I must have created the branch after the first commit. 
I have never had two PR open at the same time.

mchandschuh:
No worries, when creating a new branch for a project, make it from master to avoid this. 

'''
git checkout -b opt-datastream master
'''

StefanoRaggi:
I think I did that with Visual Studio, the problem is that I created the branch opt-datastream *after* the other commit on master.

mchandschuh:
Looks great @StefanoRaggi, thanks!

PR #115, opened 2015-06-18
Added Security.SetMarket

StefanoRaggi:
This enables users to download, convert and use data from providers currently unsupported by Lean (e.g. Oanda, Alpari, etc.)

The location of the custom data is, as usual, \data-folder\security-type\market\resolution\symbol\...

Usage example, in algorithm Initialize():

```
AddSecurity(SecurityType.Forex, "EURUSD", Resolution.Tick);
Securities["EURUSD"].SetMarket("Alpari");
```

### DIFF
--- a/Common/Securities/Security.cs
+++ b/Common/Securities/Security.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Text.RegularExpressions;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
 using QuantConnect.Securities.Equity;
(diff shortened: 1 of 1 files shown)


jaredbroad:
Cool thanks @StefanoRaggi  :) We're actually *just* about to clean that up now to merge "Country Code" and "Liquidity Provider" into one concept -- "Market". This way we can minimize the variables and switching and its a more generic idea.

Give me 30minutes and I'll commit the new Markets build.

StefanoRaggi:
@jaredbroad Ok. I'll wait and adapt the code to the new commits.

jaredbroad:
Thanks for the PR @StefanoRaggi, its a good addition

PR #116, opened 2015-06-19
Replaced "language" config setting with "algorithm-location"

StefanoRaggi:


mchandschuh:
Awesome stuff @StefanoRaggi! I like it!

PR #117, opened 2015-06-19
Removed 1024 MB ram limit for desktop Lean

StefanoRaggi:
On a local (desktop) installation of Lean, currently there is a limit of 1024 MB of ram usage.

With large backtests (more than 20-30K orders) this limit can be easily hit.

This change removes this limitation.

### DIFF
--- a/Engine/Engine.cs
+++ b/Engine/Engine.cs
@@ -84,7 +84,7 @@ namespace QuantConnect.Lean.Engine
         /// </summary>
         /// <param name="job">The algorithm job</param>
         /// <returns>The ram size in MB</returns>
-        public int GetRamAllocation(AlgorithmNodePacket job)
+        private int GetRamAllocation(AlgorithmNodePacket job)
         {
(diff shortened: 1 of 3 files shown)


jaredbroad:
I'm not keen on passing in isLocal to engine as we're trying to phase it out. A more generic solution might be passing in the maximum allowed ram, or the job to process. This will require a larger refactor so we'll take a look at it this week.

Thank you for highlighting the bug. It should definitely be unlimited ram allocation for local desktop backtesting and we'll fix it this week.

mchandschuh:
Another possibility is that we can add a different user plan that has no restrictions.

StefanoRaggi:
@mchandschuh Nice idea! That solves the problem very simply. We would also need to set the default to this new plan and add a config setting.
I'll do it later today (if you're ok with these changes), so we can close the issue.

mchandschuh:
@StefanoRaggi - I'll speak with Jared today to confirm this is what we want to do here.

jaredbroad:
Thanks @StefanoRaggi -- resolved with the new style of job architecture

PR #122, opened 2015-06-20
Thread safety in ConsoleResultHandler.Sample

StefanoRaggi:
The function is not thread-safe. Occasionally the following error occurs:

Engine.Run(): The process cannot access the file 'C:\...\Lean\Charts\...\...\Strategy Equity-Daily Performance.csv' because it is being used by another process.


mchandschuh:
Thanks for another great PR, thanks!

PR #129, opened 2015-06-28
BaseData.Bridge is now a bounded BlockingCollection

StefanoRaggi:
This update uses a [BlockingCollection\<T\>](https://msdn.microsoft.com/en-us/library/dd267312(v=vs.110).aspx) in place of a *ConcurrentQueue\<T\>* for the bridges (by default the BlockingCollection uses a ConcurrentQueue under the hood anyway).

The bounding option works well, so it solves the high RAM usage, with fast data feeds (as binary) or slow algorithms. As a by-product, we get a nice speed improvement as well.

### DIFF
--- a/Engine/DataStream.cs
+++ b/Engine/DataStream.cs
@@ -52,14 +52,13 @@ namespace QuantConnect.Lean.Engine
         {
             do
             {
-                TimeSlice timeSlice;
-                while (_feed.Bridge.TryDequeue(out timeSlice))
+                foreach (var timeSlice in _feed.Bridge.GetConsumingEnumerable())
                 {
                     AlgorithmTime = timeSlice.Time;
                     yield return timeSlice.Data;
                 }
-            }
-            while (!_feed.LoadingComplete && !_feed.Bridge.IsEmpty);
+            } 
+            while (!_feed.Bridge.IsCompleted);
 
             Log.Trace("DataStream.GetData(): All Streams Completed.");
         }
--- a/Engine/DataFeeds/DatabaseDataFeed.cs
+++ b/Engine/DataFeeds/DatabaseDataFeed.cs
@@ -69,7 +69,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// <summary>
         /// Cross-threading queue so the datafeed pushes data into the queue and the primary algorithm thread reads it out.
         /// </summary>
-        public ConcurrentQueue<TimeSlice> Bridge { get; private set; }
+        public BlockingCollection<TimeSlice> Bridge { get; private set; }
 
         /// <summary>
         /// End of Stream for Each Bridge:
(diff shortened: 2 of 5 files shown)


Engine/DataStream.cs
```diff
@@ -52,14 +52,13 @@ public DataStream(IDataFeed feed)
...
             {
-                TimeSlice timeSlice;
-                while (_feed.Bridge.TryDequeue(out timeSlice))
+                foreach (var timeSlice in _feed.Bridge.GetConsumingEnumerable())
                 {
                     AlgorithmTime = timeSlice.Time;
                     yield return timeSlice.Data;
                 }
-            }
-            while (!_feed.LoadingComplete && !_feed.Bridge.IsEmpty);
```
jaredbroad: Why don't we need loading completed anymore? What if data feed completes too quickly (e.g. daily data)

jaredbroad:
Looks really cool @StefanoRaggi, we'll do some testing once Mike gets back on land (he's literally fishing).

Engine/DataStream.cs
StefanoRaggi: The feed calls Bridge.CompleteAdding(); when finished.
The DataStream checks if finished with Bridge.IsCompleted.
So the bool LoadingComplete is not needed anymore.

jaredbroad: Great stuff!

On Sun, Jun 28, 2015 at 8:25 PM, Stefano Raggi <notifications@github.com>
wrote:

> In Engine/DataStream.cs
> <https://github.com/QuantConnect/Lean/pull/129#discussion_r33428951>:
>
> >                  {
> >                      AlgorithmTime = timeSlice.Time;
> >                      yield return timeSlice.Data;
> >                  }
> > -            }
> > -            while (!_feed.LoadingComplete && !_feed.Bridge.IsEmpty);
>
> The feed calls Bridge.CompleteAdding(); when finished.
> The DataStream checks if finished with Bridge.IsCompleted.
> So the bool LoadingComplete is not needed anymore.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/QuantConnect/Lean/pull/129/files#r33428951>.
>



-- 
Jared Broad

mchandschuh:
@StefanoRaggi -- Awesome stuff! This was the next change on my list, you beat me to it! :) Nice touch with the bounding of the blocking collection, I was unaware of that feature. Very cool, it will auto-block the data feed until there's room on the bridge.

Engine/DataStream.cs
```diff
@@ -52,14 +52,13 @@ public DataStream(IDataFeed feed)
...
+            } 
+            while (!_feed.Bridge.IsCompleted);
```
mchandschuh: do we still technically need this do/while loop here? My understanding is the GetConsumingEnumerable will never finish until IsCompleted=true

StefanoRaggi: @mchandschuh Of course you're right :) 
GetConsumingEnumerable already has a while (!IsCompleted) {...} internally.

So the external loop would never actually loop. It is useless.

PR #130, opened 2015-06-29
Fixed a couple of bugs

StefanoRaggi:
1. Missing copy of Value field in Tick copy constructor
2. Wrong initialization of Market field in FX internal currency subscriptions, when using a custom market set with Security.SetMarket()

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -138,6 +138,9 @@ namespace QuantConnect.Securities
                 }
             }
 
+            // get the market from the first existing subscription
+            string market = subscriptions.Subscriptions[0].Market;
+
             // if we've made it here we didn't find a subscription, so we'll need to add one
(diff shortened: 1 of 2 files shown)


Common/Securities/Cash.cs
```diff
@@ -138,6 +138,9 @@ public void EnsureCurrencyDataFeed(SecurityManager securities, SubscriptionManag
                 }
             }
 
+            // get the market from the first existing subscription
+            string market = subscriptions.Subscriptions[0].Market;
```
mchandschuh: maybe we should be grabbing the market from the first forex subscription? this could be an equity subscription which could cause issues.

StefanoRaggi:
Done.

mchandschuh:
Merged as of https://github.com/QuantConnect/Lean/commit/533a6ec2277c7959fe0b5c88e4446794c30f6f8d

PR #133, opened 2015-06-30
Misc fixes in ConsoleResultHandler

StefanoRaggi:
- corrected wrong date formatting in file name
- optimized thread loop
- moved file writing at the end (performance improvement)


mchandschuh:
Awesome stuff as usual @StefanoRaggi. It would also be cool to write the charts (could be aggregated in SampleRange) as well, but we can leave that for another PR :)

PR #134, opened 2015-07-01
Issue127 order requests

dpavlenkov:


dpavlenkov:
A heads up if you're pulling. Just fixed a race condition.

jaredbroad:
Thanks @dpavlenkov, thats a big PR! I'll do a deep review with Mike today and we'll get back to you with some feedback. From a quick skim my primary concern is breaking changes to the API. We have thousands of users coding to the current API. I'll try come up with some suggestions to keep it more backwards compatible.

dpavlenkov:
Basic QCAlgorithm methods for placing orders have same parameter
signatures. The returns have changed. Naturally, api is breaking, because
old is not very good. You used to get an int and try to guess whether it's
an error code or a valid order id. Now you get an OrderResponse, which
contains an OrderId.

I think most users just use SetHoldings or Market methods without checking
return value. More advanced users will appreciate the new features.

Common/Orders/MarketOrder.cs
```diff
@@ -42,6 +42,50 @@ public override decimal Value
...
+                Created = DateTime.Now,
+                Quantity = quantity ?? Quantity,
+                Tag = tag ?? Tag
+            };
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest SubmitRequest(string symbol, int quantity, string tag, SecurityType securityType, decimal price = 0, DateTime? time = null)
```
mchandschuh: Please rename this and the other `SubmitRequest` functions to `CreateSubmitRequest`

Common/Orders/MarketOrder.cs
```diff
@@ -42,6 +42,50 @@ public override decimal Value
         }
 
         /// <summary>
+        /// Create update request for pending orders. Null values will be ignored.
+        /// </summary>
+        public UpdateOrderRequest UpdateRequest(int? quantity = null, string tag = null)
```
mchandschuh: Please rename this and the other `UpdateRequest` functions to `CreateUpdateRequest`

Common/Orders/Order.cs
```diff
@@ -187,6 +205,83 @@ protected Order(string symbol, SecurityType type, int quantity, OrderType order,
         }
 
         /// <summary>
+        /// Create cancel request.
+        /// </summary>
+        /// <returns></returns>
+        public CancelOrderRequest CancelRequest()
```
mchandschuh: Please rename this to `CreateCancelRequest`

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -454,7 +496,7 @@ public List<int> Liquidate(string symbolToLiquidate = "")
                 }
 
                 //Liquidate at market price.
-                orderIdList.Add(Order(symbol, quantity));
+                orderIdList.Add(Order(symbol, quantity).OrderId);
```
mchandschuh: This could unintentionally add order IDs with a value of zero to this list. First we should check the order response to verify that it is processed:

```C#
                //Liquidate at market price.
                var orderResponse = Order(symbol, quantity);
                if (orderResponse.IsProcessed())
                {
                    orderIdList.Add(orderResponse.OrderId);
                }
```

Common/Orders/MarketOrder.cs
mchandschuh: Can we also move the SecurityType parameter to be the first parameter to match the ordering on other various API methods (see `AddSecurity`)

Common/Orders/LimitOrder.cs
```diff
@@ -37,6 +36,50 @@ public override decimal Value
...
+        {
+            base.ApplyUpdate(request);
+
+            LimitPrice = request.LimitPrice;
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest SubmitRequest(string symbol, int quantity, decimal limitPrice, string tag, SecurityType securityType, decimal price = 0, DateTime? time = null)
```
mchandschuh: Several of these `SubmitRequest` functions accept an optional `price` parameter but it never gets used. Let's remove it if it's unneeded.

Common/Orders/MarketOrder.cs
mchandschuh: Please make the time parameter required. This needs to be set as the algorithm time and I would recommend a new signature as follows:
```C#
SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal price, string tag = null)
```

Common/Orders/LimitOrder.cs
mchandschuh: Please make the time parameter required. This needs to be set as the algorithm time and I would recommend a new signature as follows:
```C#
SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal limitPrice, string tag = null)
```

Common/Orders/StopMarketOrder.cs
```diff
@@ -36,6 +36,62 @@ public override decimal Value
...
+        {
+            base.ApplyUpdate(request);
+
+            StopPrice = request.StopPrice;
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest SubmitRequest(string symbol, int quantity, decimal stopPrice, string tag, SecurityType securityType, decimal price = 0, DateTime? time = null)
```
mchandschuh: Please make the time parameter required. This needs to be set as the algorithm time and I would recommend a new signature as follows:
```C#
SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal stopPrice, string tag = null)
```

Common/Orders/StopLimitOrder.cs
```diff
@@ -46,6 +46,67 @@ public override decimal Value
...
+            base.ApplyUpdate(request);
+
+            LimitPrice = request.LimitPrice;
+            StopPrice = request.StopPrice;
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest SubmitRequest(string symbol, int quantity, decimal stopPrice, decimal limitPrice, string tag, SecurityType securityType, decimal price = 0, DateTime? time = null)
```
mchandschuh: Please make the time parameter required. This needs to be set as the algorithm time and I would recommend a new signature as follows:
```C#
SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal stopPrice, decimal limitPrice, string tag = null)
```

Common/Orders/MarketOnOpenOrder.cs
```diff
@@ -31,6 +31,49 @@ public override decimal Value
...
+                Created = DateTime.Now,
+                Quantity = quantity ?? Quantity,
+                Tag = tag ?? Tag
+            };
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest SubmitRequest(string symbol, int quantity, string tag, SecurityType securityType, decimal price = 0, DateTime? time = null)
```
mchandschuh: Please make the time parameter required. This needs to be set as the algorithm time and I would recommend a new signature as follows:
```C#
SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, string tag = null)
```

Common/Orders/MarketOnCloseOrder.cs
```diff
@@ -31,6 +31,48 @@ public override decimal Value
...
+                Created = DateTime.Now,
+                Quantity = quantity ?? Quantity,
+                Tag = tag ?? Tag
+            };
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest SubmitRequest(string symbol, int quantity, string tag, SecurityType securityType, decimal price = 0, DateTime? time = null)
```
mchandschuh: Please make the time parameter required. This needs to be set as the algorithm time and I would recommend a new signature as follows:
```C#
SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, string tag = null)
```

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -250,14 +243,14 @@ public void Scan()
                     catch (Exception err)
                     {
                         Log.Error("BacktestingBrokerage.Scan(): " + err.Message);
-                        _algorithm.Error(string.Format("Order Error: id: {0}, Transaction model failed to fill for order type: {1} with error: {2}", order.Id, order.Type, err.Message));
+                        _algorithm.Error(string.Format("BackTestingBrokerage.Scan(): Order Error: id: {0}, Transaction model failed to fill for order type: {1} with error: {2}", order.Id, order.Type, err.Message));
```
mchandschuh: Please remove the method names from these error messages. These messages are user facing.

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -250,14 +243,14 @@ public void Scan()
...
                     }
                 }
                 else
                 {
                     //Flag order as invalid and push off queue:
                     order.Status = OrderStatus.Invalid;
-                    _algorithm.Error(string.Format("Order Error: id: {0}, Insufficient buying power to complete order (Value:{1}).", order.Id, order.Value));
+                    _algorithm.Error(string.Format("BackTestingBrokerage.Scan(): Order Error: id: {0}, Insufficient buying power to complete order (Value:{1}).", order.Id, order.Value));
```
mchandschuh: Please remove the method names from these error messages. These messages are user facing.

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -420,11 +418,55 @@ private int PreOrderChecks(string symbol, int quantity, OrderType type)
...
+        {
+            return Transactions.UpdateOrder(order.UpdateRequest(quantity, limitPrice, tag));
+        }
+
+        public OrderResponse UpdateOrder(StopLimitOrder order, int? quantity = null, decimal? stopPrice = null, decimal? limitPrice = null, string tag = null)
+        {
+            return Transactions.UpdateOrder(order.UpdateRequest(quantity, stopPrice, limitPrice, tag));
+        }
+
+        public OrderResponse UpdateOrder(StopMarketOrder order, int? quantity = null, decimal? stopPrice = null, string tag = null)
```
mchandschuh: Let's modify the signatures of these `UpdateOrder` functions to move the price parameters first. In most cases, users will be updating the price of a limit/stop order vs the quantity of the order.
```C#
        public OrderResponse UpdateOrder(LimitOrder order, decimal? limitPrice = null, int? quantity = null, string tag = null)
        {
            return Transactions.UpdateOrder(order.UpdateRequest(quantity, limitPrice, tag));
        }

        public OrderResponse UpdateOrder(StopLimitOrder order, decimal? stopPrice = null, decimal? limitPrice = null, int? quantity = null, string tag = null)
        {
            return Transactions.UpdateOrder(order.UpdateRequest(quantity, stopPrice, limitPrice, tag));
        }

        public OrderResponse UpdateOrder(StopMarketOrder order, decimal? stopPrice = null, int? quantity = null, string tag = null)
        {
            return Transactions.UpdateOrder(order.UpdateRequest(quantity, stopPrice, tag));
        }
```

Common/Orders/OrderTypes.cs
```diff
@@ -139,4 +173,47 @@ public enum OrderStatus {
...
+        /// Test for Canceled, Filled, Invalid or PartiallyFilled orders.
+        /// </summary>
+        /// <param name="source">Order status</param>
+        /// <returns>True if completed status</returns>
+        public static bool IsCompleted(this OrderStatus source)
+        {
+            return source == OrderStatus.Canceled
+                || source == OrderStatus.Filled
+                || source == OrderStatus.Invalid
+                || source == OrderStatus.PartiallyFilled;
```
mchandschuh: I'm a little confused as to the difference in the method naming between `IsCompleted` and `IsFinal`. When I read `IsCompleted` it makes me think the same thing as `IsFinal`. Also, when I read `IsCompleted` it makes me think the order is completely finished, i.e., not partially filled, but completely filled or canceled/invalid.

Can you describe the reasoning for both? As it is currently defined, `IsCompleted` looks to be more of an `IsUpdateable`?

Also, maybe these two can be merged into a single `IsCompleted` that is the same definition as `IsFinal`?

Common/Orders/OrderResponse.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <summary>
+        /// Set response to processed state
+        /// </summary>
+        public void Processed()
+        {
+            this.Type = OrderResponseType.Processed;
+            this.ErrorCode = OrderResponseErrorCode.None;
+            this.ErrorMessage = "";
+        }
+
```
mchandschuh: In order to maintain backwards compatibility with existing user algorithms, can we add an implicit operator here that will return the OrderID for a successful response and the error code for an unsuccessful response. I've also requested that we define the `OrderResponseErrorCode` enum as negative values for backwards compatibility.

```C#
        public static implicit operator int(OrderResponse response)
        {
            if (response.Type == OrderResponseType.Processed)
            {
                return response.OrderId;
            }

            return (int) response.ErrorCode;
        }
```

Common/Orders/OrderResponse.cs
```diff
@@ -0,0 +1,114 @@
...
+        public bool IsError()
+        {
+            return Type == OrderResponseType.Error;
+        }
+
+        /// <summary>
+        /// Shortcut method for success check
+        /// </summary>
+        /// <returns></returns>
+        public bool IsProcessed()
```
mchandschuh: Let's update this to be a normal C# property.

```C#
public bool IsProcessed
{
   get { return Type == OrderResponseType.Processed; }
}
```

Common/Orders/OrderResponse.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            this.Id = request.Id;
+            this.OrderId = request.OrderId;
+        }
+
+        /// <summary>
+        /// Shortcut method for error check
+        /// </summary>
+        /// <returns></returns>
+        public bool IsError()
```
mchandschuh: Let's update this to be a normal C# property.

```C#
public bool IsError
{
   get { return Type == OrderResponseType.Error; }
}
```

Common/Orders/MarketOrder.cs
mchandschuh: Please update all the `UpdateRequest` functions to require a the time value as a parameter. This needs to be set to the algorithm time.

Common/Orders/Order.cs
```diff
@@ -187,6 +205,83 @@ protected Order(string symbol, SecurityType type, int quantity, OrderType order,
...
+            this.Time = request.Created;
+            Quantity = request.Quantity;
+            Tag = request.Tag;
+        }
+
+        /// <summary>
+        /// Copy order for update operations and messaging.
+        /// </summary>
+        /// <returns></returns>
+        public abstract Order Copy();
```
mchandschuh: Please update this method to be named `Clone` to match the pattern established with `BaseData`

Common/Orders/StopLimitOrder.cs
```diff
@@ -64,7 +125,7 @@ public StopLimitOrder()
         /// <param name="stopPrice">Price the order should be filled at if a limit order</param>
         /// <param name="tag">User defined data tag for this order</param>
         public StopLimitOrder(string symbol, int quantity, decimal stopPrice, decimal limitPrice, DateTime time, string tag = "", SecurityType type = SecurityType.Base) :
-            base(symbol, quantity, OrderType.StopLimit, time, 0, tag, type)
+            base(symbol, quantity, OrderType.StopLimit, time, limitPrice, tag, type)
```
mchandschuh: Why did we update this constructor to pass the `limitPrice` in as the `price` to the base constructor?  This will update the order's Price property which should be the fill price. Limit prices are always the same as the fill price for limit orders.

Common/Securities/MarginCallModel.cs
```diff
@@ -42,13 +42,46 @@ public MarginCallModel(SecurityPortfolioManager portfolio)
...
+        //        Portfolio.Transactions.WaitForOrder(order.Id);
+        //        executedOrders.Add(order);
+
+        //        // if our margin used is back under the portfolio value then we can stop liquidating
+        //        if (Portfolio.MarginRemaining >= 0)
+        //        {
+        //            break;
+        //        }
+        //    }
+        //    return executedOrders;
```
mchandschuh: Please delete this commented out code.

Common/Securities/ISecurityMarginModel.cs
```diff
@@ -72,5 +72,15 @@ public interface ISecurityMarginModel
         /// <param name="totalMargin">The totl margin used by the account in units of base currency</param>
         /// <returns>An order object representing a liquidation order to be executed to bring the account within margin requirements</returns>
         Order GenerateMarginCallOrder(Security security, decimal netLiquidationValue, decimal totalMargin);
+
```
mchandschuh: Please delete the now unused `GenerateMarginCallOrder` from the interface.

Common/Securities/SecurityMarginModel.cs
```diff
@@ -201,5 +201,45 @@ public virtual Order GenerateMarginCallOrder(Security security, decimal netLiqui
 
             return new MarketOrder(security.Symbol, quantity, security.Time, "Margin Call", security.Type);
         }
+
```
mchandschuh: Please delete the now obsolete `GenerateMarginCallOrder` function above these lines.

Common/Securities/SecurityMarginModel.cs
```diff
@@ -201,5 +201,45 @@ public virtual Order GenerateMarginCallOrder(Security security, decimal netLiqui
...
+        /// <summary>
+        /// Generates a new order for the specified security taking into account the total margin
+        /// used by the account. Returns null when no margin call is to be issued.
+        /// </summary>
+        /// <param name="security">The security to generate a margin call order for</param>
+        /// <param name="netLiquidationValue">The net liquidation value for the entire account</param>
+        /// <param name="totalMargin">The total margin used by the account in units of base currency</param>
+        /// <returns>An order object representing a liquidation order to be executed to bring the account within margin requirements</returns>
+        public virtual SubmitOrderRequest GenerateMarginCallSubmitOrderRequest(Security security, decimal netLiquidationValue, decimal totalMargin)
+        {
```
mchandschuh: It looks like we've correctly reproduced the implementation from above here, but it looks like we've forgotten to do the same for the `ForexMarginModel` class.  Here's the converted implementation for the `ForexMarginModel` class:

```C#
        public override SubmitOrderRequest GenerateMarginCallSubmitOrderRequest(Security security, decimal netLiquidationValue, decimal totalMargin)
        {
            if (totalMargin <= netLiquidationValue)
            {
                return null;
            }
            var forex = (Forex)security;

            // we haven't begun receiving data for this yet
            if (forex.Price == 0m || forex.QuoteCurrency.ConversionRate == 0m)
            {
                return null;
            }

            // compute the amount of quote currency we need to liquidate in order to get within margin requirements
            decimal delta = (totalMargin - netLiquidationValue) / forex.QuoteCurrency.ConversionRate;

            // compute the number of shares required for the order, rounding up
            int quantity = (int)Math.Round(delta / security.Price, MidpointRounding.AwayFromZero);

            // don't try and liquidate more share than we currently hold
            quantity = Math.Min((int)security.Holdings.AbsoluteQuantity, quantity);
            if (security.Holdings.IsLong)
            {
                // adjust to a sell for long positions
                quantity *= -1;
            }

            return MarketOrder.SubmitRequest(security.Symbol, quantity, "Margin Call", security.Type, security.Price, security.Time);
        }
```

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -322,22 +310,123 @@ public void Exit()
...
 
+        private Order GetOrderById(int id)
+        {
+            Order order = null;
+
+            if (_orders.TryGetValue(id, out order) == false)
+            {
+                var clientOrder = _algorithm.Transactions.GetOrderById(id);
+
+                if (order != null)
```
mchandschuh: I think this should be testing `clientOrder` for null, as the `order` is gauranteed to have a value via `_orders.TryGetValue(id, out order)`

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -322,22 +310,123 @@ public void Exit()
...
+
+            _orders[order.Id] = order;
+
+            // check to see if we have enough money to place the order
+            if (!_algorithm.Transactions.GetSufficientCapitalForOrder(_algorithm.Portfolio, order))
+            {
+                order.Status = OrderStatus.Invalid;
 
+                response.Error(OrderResponseErrorCode.InsufficientBuyingPower, string.Format("Order Error: id: {0}, Insufficient buying power to complete order (Value:{1}).", order.Id, order.Value));
+                _algorithm.Error("BrokerageTransactionHandler.HandleNewOrder(): " + response.ErrorMessage);
```
mchandschuh: Please remove method names from the algorithm's Error messages

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -322,22 +310,123 @@ public void Exit()
...
+        /// </summary>
+        /// <param name="request">Submit order request</param>
+        private void HandleNewOrder(SubmitOrderRequest request, OrderResponse response)
+        {
+            var order = GetOrderById(request.OrderId);
+
+            if (order != null)
+            {
+                response.Error(OrderResponseErrorCode.OrderAlreadyExists, String.Format("Cannot process submit request because order with id [{0}] already exists", request.OrderId));
+                _algorithm.Error("BrokerageTransactionHandler.HandleNewOrder(): " + response.ErrorMessage);
```
mchandschuh: Please remove method names from the algorithm's Error messages

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -133,52 +149,24 @@ public void Run()
...
-                Order order;
-                if (!_orderQueue.TryDequeue(out order))
-                {
-                    _algorithm.ProcessingOrder = false;
-                    Thread.Sleep(1);
-                    continue;
-                }
-
-                _algorithm.ProcessingOrder = true;
+                bool notIdle = false;
```
mchandschuh: Please use a positive sense for boolean variables and logic.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -87,17 +85,17 @@ public virtual void Initialize(IAlgorithm algorithm, IBrokerage brokerage, IResu
...
             _brokerage.SecurityHoldingUpdated += (sender, holding) =>
             {
-                HandleSecurityHoldingUpdated(holding);
+                Process(holding);
             };
 
             _brokerage.AccountChanged += (sender, account) =>
             {
-                HandleAccountChanged(account);
+                Process(account);
```
mchandschuh: Please inline these `Process` method implementations as all they're doing is calling `Enqueue` on the appropriate event queue.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -121,9 +123,23 @@ public virtual void Initialize(IAlgorithm algorithm, IBrokerage brokerage, IResu
...
         }
 
+        public bool HasPendingItems
+        {
+            get
+            {
+                return _orderRequestQueue.Count > 0
+                    || _orderEventQueue.Count > 0
+                    || _securityEventQueue.Count > 0
+                    || _accountEventQueue.Count > 0;
```
mchandschuh: As per the [MSDN documentation](https://msdn.microsoft.com/en-us/library/dd267339(v=vs.110).aspx) the preferred way to check if a `ConcurrentQueue` has any items is to use the `IsEmpty` property.

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -296,12 +400,25 @@ public void WaitForOrder(int orderId)
         /// <returns>List of open orders.</returns>
         public List<Order> GetOpenOrders()
         {
-            var openOrders = (from order in Orders.Values
+            var openOrders = (from order in _orders.Values
```
mchandschuh: Calling `Values` requires all locks within the collection, the preferred way is to just enumerate `_orders` as `KeyValuePairs` and then to access the individual `Value` properties on each `KeyValuePair`

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -296,12 +400,25 @@ public void WaitForOrder(int orderId)
...
             return openOrders;
+        }
+
+        /// <summary>
+        /// Get a list of all open orders.
+        /// </summary>
+        /// <returns>List of open orders.</returns>
+        public List<Order> GetOrders(Predicate<Order> filter = null)
+        {
+            var result = (from order in _orders.Values
```
mchandschuh: Calling `Values` requires all locks within the collection, the preferred way is to just enumerate `_orders` as `KeyValuePairs` and then to access the individual `Value` properties on each `KeyValuePair`

dpavlenkov:
This is great stuff. I'm working on all this.

jaredbroad:
Great work @dpavlenkov! The only design idea we were worried about was storing a clone of the orders in algorithm. This means the state will always be inconsistent between the algorithm and transaction handler. 

To fix this @mchandschuh will refactor the code on master to expose an "IOrderProvider" from the transaction handler, and send this to the algorithm from the SetupHandler. This way the algorithm can access to transaction handler's orders without needing its own copy.

We're going to rework this and publish it in the next 12 hours. It should tidy up a lot of the back and forth conversion you did and eliminate the need to clone orders each time.

Additionally the async code should probably be submitted in a separate pull request if possible. Its a nice idea but I'd like to put more thought into its implementation and pattern once this larger refactor is done.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
dpavlenkov: https://github.com/QuantConnect/Lean/pull/134/files#diff-93fd550d3a43e7fb3ded36524aa7aefbL257

Because of changes ot BrokerageSetupHandler, now it's possible for Transactions to have an order that BrokerageTransactionHandler doesn't

dpavlenkov:
@mchandschuh, I've addressed all comments, please review. @jaredbroad, it's perfectly fine to separate async code, plus it needs some more thinking. It's not fitting too well with the whole data flow.

dpavlenkov:
I've missed one comment, so here's another commit.

Algorithm.CSharp/OrderOperationsAlgorithm.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var security = Securities[_symbol];
+
+            if (!Portfolio.Invested)
+            {
+                submittedMarketOrder = null;
+
+                SetHoldings(_symbol, 1);
+
+                if (submittedMarketOrder != null)
```
mchandschuh: Won't `submittedMarketOrder`always be null here?

Algorithm.CSharp/OrderOperationsAlgorithm.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    var takeProfitLimitOrderId = ProcessAndLog(QuantConnect.Orders.LimitOrder.CreateSubmitRequest(security.Type, security.Symbol, -submittedMarketOrder.Quantity, security.Time, security.Price + 0.5m, tag: "Take Profit"));
+                    var stopLossStopMarketOrderId = ProcessAndLog(QuantConnect.Orders.StopMarketOrder.CreateSubmitRequest(security.Type, security.Symbol, -submittedMarketOrder.Quantity, security.Time, security.Price - 0.5m, tag: "Stop Loss"));
+                    Transactions.GetCompletedOrderAsync(takeProfitLimitOrderId).ContinueWith(t => { if (t.Result.Status == OrderStatus.Filled) Log(CancelOrder(stopLossStopMarketOrderId)); });
```
mchandschuh: Please remove the async component of this PR, we should have a dedicated design discussion for the async orders feature.

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -36,18 +36,10 @@ public partial class QCAlgorithm
         /// <summary>
         /// Wait semaphore to signal the algoritm is currently processing a synchronous order.
         /// </summary>
-        public bool ProcessingOrder
+        public bool ProcessingEvents
```
mchandschuh: Indeed this or `volatile` is a good way to do this, however, I recently pushed https://github.com/QuantConnect/Lean/commit/9c1b0f4616e599f77e33f30c3c1ee8f6b4ef3c74 which removes this entirely. Instead there's a `ManualResetEventSlim` that can be used to wait for things to finish.

Algorithm.CSharp/OrderOperationsAlgorithm.cs
dpavlenkov: How about I just remove this algorithm?

dpavlenkov: Removed. No big deal. Didn't like it much anyway.

dpavlenkov: Main reason for this algorithm was to test the order processing. It was not complete, nor to the point. But I can see you started adding tests on master branch, and that's great!

Common/Orders/MarketOnCloseOrder.cs
```diff
@@ -31,6 +31,48 @@ public override decimal Value
...
+                Quantity = quantity ?? Quantity,
+                Tag = tag ?? Tag
+            };
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal price = 0, string tag = null)
+        {
```
mchandschuh: Let's remove the unused optional price parameter here.

Common/Orders/MarketOnOpenOrder.cs
```diff
@@ -31,6 +31,49 @@ public override decimal Value
...
+                Quantity = quantity ?? Quantity,
+                Tag = tag ?? Tag
+            };
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal price = 0, string tag = null)
+        {
```
mchandschuh: Let's remove the unused optional price parameter here

Common/Orders/MarketOrder.cs
```diff
@@ -42,6 +42,49 @@ public override decimal Value
...
+                Quantity = quantity ?? Quantity,
+                Tag = tag ?? Tag
+            };
+        }
+
+        /// <summary>
+        /// Create submit request.
+        /// </summary>
+        public static SubmitOrderRequest CreateSubmitRequest(SecurityType securityType, string symbol, int quantity, DateTime time, decimal price = 0, string tag = null)
+        {
```
mchandschuh: Let's remove the unused optional price parameter from here

Brokerages/Backtesting/BacktestingBrokerage.cs
mchandschuh: It looks like this and the case below wasn't removed. Please remove the method names from algorithm error messages. These get sent directly to the user. The Class.MethodName(): convention is for system logging only.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
mchandschuh: See my recent change where I moved the storage of the order queue and orders to the transaction handler: https://github.com/QuantConnect/Lean/commit/9c1b0f4616e599f77e33f30c3c1ee8f6b4ef3c74

Common/Orders/MarketOrder.cs
dpavlenkov: Actually, I think instead I should use this parameter and set Price on SubmitRequest, because Order constructor reads SubmitRequest.Price.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
dpavlenkov: Ok, that's great. What should I do? I've only resently started using git.
Should I merge my branch with master?

Common/Orders/MarketOrder.cs
jaredbroad: Price parameter here is the *fill* price, so it can't be updated manually by the user or anyone.

mchandschuh:
Great stuff @dpavlenkov! We'll need to merge in my latest changes that move the order storage to the transaction handler. 

Also, it looks like we still have some async stuff in the SecurityTransactionManager.

Common/Orders/MarketOrder.cs
dpavlenkov: Should I just remove Price field from SubmitOrderRequest?

mchandschuh: The MarketOrder type requires the price, the others should have price set to zero. With that being said, @jaredbroad and I spoke about changing how this works in the future. Currently the price is needed for the `Value` property on `Order` during the call to `GetSufficientCapitalForOrder`. The new idea is to have an method such as:
```C#
public abstract decimal GetOrderValue(decimal marketPrice)
```
Which all orders will implement, for now let's just leave it alone though, we'll come back to this in due time.

dpavlenkov: Very well. So I should merge your upstream/master with my branch and see
what happens?

dpavlenkov:
Ok, I'll be working on merging tonight.

mchandschuh:
Thanks @dpavlenkov! I'll give it another sweep today.

dpavlenkov:
It's the slice stuff that fails.

mchandschuh:
@dpavlenkov Slice stuff fixed in https://github.com/QuantConnect/Lean/commit/3905276033b661b88a0a837b14da3645e18943ea

That was my fault :)

dpavlenkov:
QCAlgorithm.OnData(Slice slice) enters infinite loop when I debug RegressionAlgorithm

mchandschuh:
Thanks @dpavlenkov! You're two for two today and I'm 0 for two! Fix pushed in https://github.com/QuantConnect/Lean/commit/27cd4fbbe528e6aa56494dfc728ea6a9430ee4c3

mchandschuh:
Also, a note on `RegressionAlgorithm` I don't think all the data is in github to successfully run that algo. I'll have to write a new one in the same vein that uses only data available in github.

dpavlenkov:
I don't expect them to pass at this point, just to see that they do something.

dpavlenkov:
I just want to clarify that Submit, Update, Cancel order request methods on SecurityTransactionManager are blocking because they return Task.Result. Conceptually, me using TaskCompletionSource is the same as you looping and checking wheter the response has arrived while sleeping or spinning in the interim. However, my approach has better performance, because thread blocks; it doesn't have to constantly check for response, sleep or spin. If you can use Task.Wait, ManualResetEvent, Sleep, then why can't you use TaskCompletionSource? As far as I'm concerned, it's a signaling convenience. I'll remove the Task<OrderResponse> return, so users wouldn't be tempted to do crazy async stuff.

jaredbroad:
@dpavlenkov from my understanding its about *where* we do the synchronization; Market is the only type which needs instant sync and forcing all the other types to sync inline will hinder their performance -- especially if we're doing this with a real brokerage i.e. [BrokerageTransactionHandler] (https://github.com/dpavlenkov/Lean/blob/fb68aadcc021b8f871e493fa48bd6ee3a5a5b8bc/Engine/TransactionHandlers/BrokerageTransactionHandler.cs#L475). 

Imagine the case where you're doing an algorithm monitoring 400 symbols, and then submitting 400 limit orders... making it block the algorithm thread while you get OrderResponse would cause long delays -- lots of brokerages have a REST API, this is at least 50-100ms delay between order submissions = 400x100ms = 40sec delay. This is *why* Transaction Handler has its own thread and works asynchronously.

I agree thread sleeps are ugly, so we'll have to fix that separately. Complex order events is definitely an awesome idea -- but again lets chat about it separately and with collaboration in the design which will cover all the community use cases.

dpavlenkov:
I'm sorry, I thought you wanted blocking, hence my frustration. I'll fix
this.

mchandschuh:
Closing, this has already been committed, thanks @dpavlenkov!

PR #135, opened 2015-07-03
Fix AlgorithmRunner.RunLocalBacktest

dpavlenkov:
Remove unnecessary acknowledgement, dispose properly of loghandler et all.

### DIFF
--- a/Tests/AlgorithmRunner.cs
+++ b/Tests/AlgorithmRunner.cs
@@ -34,6 +34,8 @@ namespace QuantConnect.Tests
         {
             var statistics = new Dictionary<string, string>();
 
+            Composer.Instance.Reset();
+
             try
             {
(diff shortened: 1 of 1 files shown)


dpavlenkov:
I need to look more at this. It's still not right.

dpavlenkov:
I can run tests individually, but when more than one together, one completes and the other hangs. I can see that BrokerageTransactionHandler.Run() transaction thread is not running. So this needs more looking into.

dpavlenkov:
Now all tests run together.

mchandschuh:
Awesome stuff @dpavlenkov!!

Merged in https://github.com/QuantConnect/Lean/commit/78d21b0d75ea0c060d1f092153ea4e5b443d8ece

PR #142, opened 2015-07-18
Small fixes to transaction handler

dpavlenkov:
Update canceled order status before the order event.

Since we always maintain order reference in transaction handler and tickets, OrderTicket.SetOrder needs to be invoked just once, on order submission.

Internal order lookup should not clone.


dpavlenkov:
As it stands right now, canceled orders are not recorded in transaction handler, nor is the ticket status updated; that's the reason for the fix.

dpavlenkov:
Addressed your comments.

PR #147, opened 2015-07-23
change to use algorithm.Time in Algorithm Manager

bizcad:
Per issue 145, change TradeBars to use algorithm.Time instead of UTC. 

Mark DataDictionary<T> Time as obsolete.


jaredbroad:
Thank you @bizcad !

Issue #19, opened 2015-01-31
Add OANDA Brokerage


mourednik:
OANDA has a REST API with reference implementation in C#
https://github.com/oanda/rest-csharp-refimpl/
I would like to implement this feature. I will study it over the next few days.

mchandschuh:
Sounds great. The IBrokerage interface is fairly new and still may be in flux. If you have any questions let me know.

jilott:
@mourednik Have you made any progress on this? I'd probably be interested in helping, I've done some work with the Oanda API using Python. I'm pretty new to C# though.

mourednik:
@jilott I haven't. Unfortunately I don't think I will have time to work on this in the next few weeks.

jilott:
Well, it may be a race then.

mourednik:
Please go for it. If I do start working on it I will check with you first.

jaredbroad:
Awesome guys. I wrote a few more tutorials on how to do this today which I'll push soon.

There is one change we will make in the next few days - a BrokerageSetupHandler. This would allow an implementation of a IBrokerage to more or less setup an algorithm. At the moment the setup process isn't very generic and we don't have the IBrokerage SetupHandler committed.

Best would be to focus efforts on implementing the current IBrokerage class methods - PlaceOrder etc. And in the next few days we'll add a few more methods to the IBrokerage interface which will ask for e.g. "GetCash", "GetPositions", "GetTrades".

jaredbroad:
Closed in #78e12d7fe78cd87ff2e9ea67b71a7b965e64929c

PR #152, opened 2015-08-12
Add OANDA Brokerage

kaffeebrauer:
https://github.com/QuantConnect/Lean/issues/19


Brokerages/Backtesting/OandaBrokerageModel.cs
```diff
@@ -0,0 +1,64 @@
...
+        ///     order type, security type, and order size limits.
+        /// </summary>
+        /// <remarks>
+        ///     For example, a brokerage may have no connectivity at certain times, or an order rate/size limit
+        /// </remarks>
+        /// <param name="security">The security being ordered</param>
+        /// <param name="order">The order to be processed</param>
+        /// <param name="message">If this function returns false, a brokerage message detailing why the order may not be submitted</param>
+        /// <returns>True if the brokerage could process the order, false otherwise</returns>
+        public override bool CanSubmitOrder(Security security, Order order, out BrokerageMessageEvent message)
```
mchandschuh: I think we may want to override CanExecuteOrder for this behavior. [CanSubmitOrder](https://github.com/QuantConnect/Lean/blob/master/Engine/TransactionHandlers/BrokerageTransactionHandler.cs#L593) is called pre-submit to the brokerage, whereas [CanExecuteOrder](https://github.com/QuantConnect/Lean/blob/master/Brokerages/Backtesting/BacktestingBrokerage.cs#L221) is called each time an order is checked for a fill in backtesting.

Brokerages/Backtesting/OandaBrokerageModel.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            message = null;
+
+            if (order.DurationValue > DateTime.Now.AddMonths(3))
+            {
+                message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
+                    "Oanda does not support order expiration dates more than 3 months in the future."
+                  );
+                return false;
```
mchandschuh: I know it's not the best design, but for Oanda, after 3 months the order is dead, we may want to market is as invalid:
```c#
order.Status = OrderStatus.Invalid
```
@jaredbroad thoughts??

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        }
+        
+        private static string GetCommaSeparatedList(IEnumerable<string> items)
+        {
+            var result = new StringBuilder();
+            foreach (var item in items)
+            {
+                result.Append(item + ",");
+            }
+            return result.ToString().Trim(',');
```
mchandschuh: `string.Join(",", items)` may be more concise :)

mchandschuh:
Just started doing a little late night reading here :) didn't look to deeply at anything yet, first glance code structure/tidyness looks good :) I'll dig into it deeply tomorrow, thanks once again @kaffeebrauer for the awesome work here!

Brokerages/Backtesting/OandaBrokerageModel.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </remarks>
+        /// <param name="security">The security being ordered</param>
+        /// <param name="order">The order to be processed</param>
+        /// <param name="message">If this function returns false, a brokerage message detailing why the order may not be submitted</param>
+        /// <returns>True if the brokerage could process the order, false otherwise</returns>
+        public override bool CanSubmitOrder(Security security, Order order, out BrokerageMessageEvent message)
+        {
+            message = null;
+
+            if (order.DurationValue > DateTime.Now.AddMonths(3))
```
mchandschuh: Can't use DateTime.Now, maybe we could do something like this:
```c#
// verify the order is less than 3 months old
var currentTime = security.LocalTime.ConvertToUtc(security.Exchange.TimeZone);
if (order.Time.AddMonths(3) < currentTime)
{
    message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
        "Oanda does not support order expiration dates more than 3 months in the future."
        );
    return false;
}
return true;
```

Brokerages/Backtesting/OandaBrokerageModel.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
```
mchandschuh: I like to make these codes as specific as possible, so maybe something like: "OrderExpiredLimitReached" or something? I'm historically not the best with naming things :) so any suggestions are welcome.

Brokerages/Backtesting/OandaBrokerageModel.cs
mchandschuh: EDIT: This would need to be perform in `CanExecuteOrder`, not in `CanSubmitOrder` since we're trying to prevent us from modelling a fill if the order is too old, `CanSubmitOrder` only gets called before order placement to the brokerage implementation

Brokerages/Oanda/DataType/Account.cs
```diff
@@ -0,0 +1,161 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+﻿namespace QuantConnect.Brokerages.Oanda.DataType
+{
+    /// <summary>
+    /// Represents the Oanda Account.
+    /// </summary>
+    public class Account
```
mchandschuh: I know some may think this is nitpicky, but let's please use 4 spaces instead of tabs. If it's not already in CONTRIBUTING.md then I'll add it there now.

Was this code copied from somewhere? If it was we can leave it how it is but put a <remark /> at the top of the class with a link and/or description of the source.

Sometimes it's good to leave it exactly as copied so its easy to diff later and merge in bug fixes.

mchandschuh: Yup, it looks like the REST api is implemented [here](https://github.com/oanda/rest-csharp-refimpl)

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// </summary>
+        /// <value>
+        /// The oanda environment.
+        /// </value>
+        public static Environment OandaEnvironment { get; private set; }
+
+        private readonly object _lockAccessCredentials = new object();
+
+        //This should correlate to the Trades/Positions list in Oanda.
+        private readonly IHoldingsProvider _holdingsProvider;
```
mchandschuh: this appears to be unused. in the tradier impl we needed this since if an order will bring your holdings over the zero boundary it needs to be split into two orders. for example, I own 10 shares SPY and sell 20, tradier requires a sell to close the 10 and then a sell to open. does oanda have a similar requirement?

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+
+        //This should correlate to the Orders list in Oanda.
+        private readonly IOrderProvider _orderProvider;
+
+        private readonly ConcurrentDictionary<long, DataType.Order> _cachedOpenOrdersByOandaOrderId;
```
mchandschuh: this is populated but never accessed, maybe it's not needed

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// Refresh Token Access:
+        /// </summary>
+        public string RefreshToken { get; private set; }
+
+
+        /// <summary>
+        /// Creates a new Brokerage instance with the specified name
+        /// </summary>
+        /// <param name="name">The name of the brokerage</param>
+        public OandaBrokerage(string name, int userId) : base(name)
```
mchandschuh: this ctor can be removed

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// </summary>
+        /// <returns>The open orders returned from Oanda</returns>
+        public override List<Order> GetOpenOrders()
+        {
+            var oandaOrders = GetOrderList();
+
+            var orderList = oandaOrders.Select(ConvertOrder).ToList();
+            foreach (var openOrder in orderList)
+            {
+                _cachedOpenOrdersByOandaOrderId[openOrder.Id] = oandaOrders.FirstOrDefault(oo => oo.id == openOrder.Id);
```
mchandschuh: the 'oo' lambda parameter looks awesome, at first glance it made me think of the infinity symbol (∞)

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// <summary>
+        /// Converts the Oanda position into a QuantConnect holding.
+        /// </summary>
+        /// <param name="position">The position.</param>
+        /// <returns></returns>
+        protected Holding ConvertHolding(Position position)
+        {
+            return new Holding
+            {
+                Symbol = position.instrument,
```
mchandschuh: sometimes we need to map the symbols to the brokerage and from the brokerage back to the QC symbol, for example, IB specifies the euro/dollar pair as 'EUR.USD' but QC specifies it just as 'EURUSD' so we need to handle that. I think oanda uses '_' between the pairs from the symbol in the test class. I think we'll need to add the symbol mapping here.

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+                Type = InstrumentSecurityTypeMap[position.instrument],
+                AveragePrice = (decimal)position.avgPrice,
+                ConversionRate = 1.0m,
+                CurrencySymbol = "$",
```
mchandschuh: So what about forex pairs here? Imagine the position.instrument is EUR/GBP, that is a quote to buy euros using gb pounds. So we would need a ConversionRate from GBP -> USD for us to properly calculate the initial portfolio value before we start to receive live data, since the 'prices' are in GBP. IB does this, tradier didn't have to since they don't handle forex pairs

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// </summary>
+        /// <returns>The current cash balance for each currency available for trading</returns>
+        public override List<Cash> GetCashBalance()
+        {
+            var cash = new List<Cash>();
+            var requestString = EndpointResolver.ResolveEndpoint(OandaEnvironment, Server.Account) + "accounts/" + AccountId;
+            var accountResponse = MakeRequest<Account>(requestString);
+            //TODO figure how exchange rates work in Oanda. probably need to call http://developer.oanda.com/exchange-rates-api/v1/currencies/
+            cash.Add(new Cash(accountResponse.accountCurrency, accountResponse.balance.ToDecimal(), new decimal(1.0)));
+            return cash;
```
mchandschuh: I think we'll also need to query for any forex holdings, since forex trades are technically currency swaps. we model forex by our CashBook object, but we still provide what IB calls a 'virtual' forex position, so I think we would want to return forex from cash as well as from get account holdings as they're used to set different things. get account holdings could return all the virtual positions, but get cash balance would return the actual balance for each currency, this could be found by aggregating all the positions, but hopefully there's an api for it :)

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+            return result.ToString().Trim(',');
+        }
+
+
+        private Dictionary<string, SecurityType> MapInstrumentToSecurityType(List<Instrument> instruments)
+        {
+            var result = new Dictionary<string, SecurityType>();
+            foreach (var instrument in instruments)
+            {
+                var isForex = Regex.IsMatch(instrument.instrument, "[A-Z]{3}_[A-Z]{3}") && Regex.IsMatch(instrument.displayName, "[A-Z]{3}/[A-Z]{3}");
```
mchandschuh: I was never much good with regex without solid tests, I'll trust this is solid

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// <summary>
+        /// Gets the list of available tradable instruments/products from Oanda
+        /// </summary>
+        /// <returns></returns>
+        public List<Instrument> GetInstrumentsAsync(List<string> instrumentNames = null)
+        {
+            var requestString = EndpointResolver.ResolveEndpoint(OandaEnvironment, Server.Rates) + "instruments?accountId=" + AccountId;
+            if (instrumentNames != null)
+            {
+                var instrumentsParam = GetCommaSeparatedList(instrumentNames);
```
mchandschuh: GetCommaSeparatedList = `string.Join(",", instrumentNames);`

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        {
+            var requestParams = new Dictionary<string, string>
+            {
+                {"instrument", order.Symbol},
+                {"units", Convert.ToInt32(order.AbsoluteQuantity).ToString()}
+            };
+
+            PopulateOrderRequestParameters(order, requestParams);
+
+            Log.Trace(string.Format("OandaBrokerage.PlaceOrder(): {0} to {1} {2} units of {3}", order.Type, order.Direction, order.Quantity, order.Symbol));
```
mchandschuh: This log message can be simplified to just use the order.ToString() method

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        }
+
+
+        /// <summary>
+        /// Checks for fill events by registering to the event session to receive events.
+        /// </summary>
+        private void CheckForFills()
+        {
+            // reentrance gaurd
+            if (!Monitor.TryEnter(_fillLock))
```
mchandschuh: I don't think we need the `_fillLock` here, it was used in the tradier implementation because we had to poll their api. Oanda provides even streaming. I would recommend making the EventsSession instance a class wide variable to codify it's lifetime, as it's not immediately obvious that the object will live after this method due to the event handler binding.

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+                if (data.transaction.type == "ORDER_FILLED")
+                {
+                    var qcOrder = _orderProvider.GetOrderByBrokerageId(data.transaction.orderId);
+                    var fill = new OrderEvent(qcOrder, "Oanda Fill Event")
+                    {
+                        Status = OrderStatus.Filled,
+                        //    // this is guaranteed to be wrong in the event we have multiple fills within our polling interval,
+                        //    // we're able to partially cope with the fill quantity by diffing the previous info vs current info
+                        //    // but the fill price will always be the most recent fill, so if we have two fills with 1/10 of a second
+                        //    // we'll get the latter fill price, so for large orders this can lead to inconsistent state
```
mchandschuh: Since it is a streaming event API I'm not sure this same warning applies. We should be receiving every fill event in order

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+            {
+                var requestString = EndpointResolver.ResolveEndpoint(OandaEnvironment, Server.Account) + "accounts/" + AccountId + "/orders/" + orderId;
+                try
+                {
+                    MakeRequestWithBody<DataType.Order>(requestString, "PATCH", requestParams);
+                     
+                } 
+                catch (Exception)
+                {
+                }
```
mchandschuh: we should at least log or comment when this block is empty

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// <param name="instruments">the list of instruments to check</param>
+        /// <returns>List of Price objects with the current price for each instrument</returns>
+        public List<Price> GetRates(List<Instrument> instruments)
+        {
+            var requestBuilder = new StringBuilder(EndpointResolver.ResolveEndpoint(OandaEnvironment, Server.Rates) + "prices?instruments=");
+
+            foreach (var instrument in instruments)
+            {
+                requestBuilder.Append(instrument.instrument + ",");
+            }
```
mchandschuh: `string.Join(",", instruments.Select(i => i.instrument));`

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+                } 
+                catch (Exception)
+                {
+                }
+                
+            } 
+            else
+            {
+
+                OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Warning, "UpdateFailed", "Failed to update Oanda order id: " + orderId + "."));
```
mchandschuh: I think we'll need an OnOrderEvent here that let's the engine know that the order has the same old state... @jaredbroad  what should we do when an update fails? LEAN has already updated it's internal representation of the order, so if we want to keep the order alive we'll need to somehow tell LEAN that the order needs to be un-updated!

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Obtain the active open Trade List from Oanda.
+        /// </summary>
+        /// <param name="requestParams"></param>
+        /// <returns></returns>
+        public TradesResponse GetTradeList(Dictionary<string, string> requestParams = null)
```
mchandschuh: various API wrapping methods accepts a dictionary of parameters, it would be nice to define them all as formal parameters and make the dictionary internally in the method

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+            if (!order.BrokerId.Any())
+            {
+                Log.Trace("OandaBrokerage.CancelOrder(): Unable to cancel order without BrokerId.");
+                return false;
+            }
+
+            foreach (var orderId in order.BrokerId)
+            {
+                CancelOrderAsync(orderId);
+                OnOrderEvent(new OrderEvent(order, "Oanda Cancel Order Event") { Status = OrderStatus.Canceled });
```
mchandschuh: since the cancel is done async, I think the OnOrderEvent should be done in a ContinueWith block, we can change CancelOrderAsync to return the `Task<T>` instance and upon task completion we'll fire the event since it may have failed to be canceled.

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// Verify we have a user session; or refresh the access token.
+        /// </summary>
+        public bool RefreshSession()
+        {
+            var raw = "";
+            bool success;
+            lock (_lockAccessCredentials)
+            {
+                //Create the client for connection:
+                var client = new RestClient("https://www.quantconnect.com/terminal/");
```
mchandschuh: I'm open to suggestions on a different means of persisting access tokens. we did this for tradier, but it looks like we'll need to do this for almost everyone. also, it makes running locally difficult. no action required here for now, but something for those following to think about (and offer suggestions :) )

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+            if (IsConnected) return;
+            RefreshSession();
+        }
+
+        /// <summary>
+        /// Disconnects the client from the broker's remote servers
+        /// </summary>
+        public override void Disconnect()
+        {
+            throw new NotImplementedException();
```
mchandschuh: let's not throw, we call disconnect when cleaning up resource/disposing, if he has nothing to do here the // NOP is perfectly acceptable.

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// </summary>
+        /// <param name="accountId">the account IDs you want to stream on</param>
+        /// <returns>the WebResponse object that can be used to retrieve the events as they stream</returns>
+        public static async Task<WebResponse> StartEventsSession(List<int> accountId = null)
+        {
+            var requestString = EndpointResolver.ResolveEndpoint(OandaEnvironment, Server.StreamingEvents) + "events";
+
+            if (accountId != null && accountId.Count > 0)
+            {
+                var accountIds = accountId.Aggregate("", (current, account) => current + (account + ","));
```
mchandschuh: `string.Join(",", accountId);`

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+                accountIds = accountIds.Trim(',');
+                requestString += "?accountIds=" + WebUtility.UrlEncode(accountIds);
+            }
+
+            var request = WebRequest.CreateHttp(requestString);
```
mchandschuh: in all my years with C# (~10) I have not seen `WebRequest.CreateHttp` thank you for showing me this, I've always wondered why someone would design an API that requires casting! Typically I see:
```c#
var http = (HttpWebRequest)WebRequest.Create(url);
```

and always wondered but never investigated if it could be done without the cast, sweet! learn something new every day.

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+                return result;
+            }
+        }
+
+        /// <summary>
+        /// Retrieves the current non-zero positions for a given account
+        /// </summary>
+        /// <param name="accountId">positions will be retrieved for this account id</param>
+        /// <returns>List of Position objects with the details for each position (or empty list iff no positions)</returns>
+        public async Task<List<Position>> GetPositionsAsync(int accountId)
```
mchandschuh: this method looks like it will always run synchronously, there's no 'await' keywords in the method body. do we intended to call `MakeRequestAsync<T>` instead?

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+        /// <summary>
+        /// Helper function to create the parameter string out of a dictionary of parameters
+        /// </summary>
+        /// <param name="requestParams">the parameters to convert</param>
+        /// <returns>string containing all the parameters for use in requests</returns>
+        private static string CreateParamString(Dictionary<string, string> requestParams)
+        {
+            var requestBody = requestParams.Aggregate("", (current, pair) => current + (WebUtility.UrlEncode(pair.Key) + "=" + WebUtility.UrlEncode(pair.Value) + "&"));
+            requestBody = requestBody.Trim('&');
+            return requestBody;
```
mchandschuh: could use a one liner like this:
```c#
return string.Join(",", requestParams.Select(x => x.Key + "=" + x.Value).Select(WebUtility.UrlEncode));
```

Brokerages/Oanda/OandaBrokerage.cs
```diff
@@ -0,0 +1,1052 @@
...
+            }
+            qcOrder.Symbol = order.instrument;
+            qcOrder.Quantity = ConvertQuantity(order);
+            qcOrder.SecurityType = InstrumentSecurityTypeMap[order.instrument];
+            qcOrder.Status = OrderStatus.None;
+            qcOrder.BrokerId.Add(order.id);
+            qcOrder.Id = order.id;
+            qcOrder.Duration = OrderDuration.Specific;
+            qcOrder.DurationValue = XmlConvert.ToDateTime(order.expiry, XmlDateTimeSerializationMode.Utc);
+            qcOrder.Time = XmlConvert.ToDateTime(order.time, XmlDateTimeSerializationMode.Utc);
```
mchandschuh: great stuff, we'll need to discuss supporting this new DurationValue in backtesting, currently we only really support good til cancelled

Brokerages/Oanda/OandaBrokerageFactory.cs
```diff
@@ -0,0 +1,174 @@
...
+{
+    /// <summary>
+    /// Provides an implementations of IBrokerageFactory that produces a OandaBrokerageFactory
+    /// </summary>
+    public class OandaBrokerageFactory: BrokerageFactory
+    {
+
+        /// <summary>
+        /// Gets tradier values from configuration
+        /// </summary>
```
mchandschuh: a little left-over copy pasta from last night's dinner :D

Brokerages/Oanda/OandaBrokerageFactory.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <param name="job">The job packet to create the brokerage for</param>
+        /// <param name="algorithm">The algorithm instance</param>
+        /// <returns>A new brokerage instance</returns>
+        public override IBrokerage CreateBrokerage(LiveNodePacket job, IAlgorithm algorithm)
+        {
+            var errors = new List<string>();
+            var accountId = Read<int>(job.BrokerageData, "oanda-account-id", errors);
+            var accessToken = Read<string>(job.BrokerageData, "oanda-access-token", errors);
+            var refreshToken = Read<string>(job.BrokerageData, "oanda-refresh-token", errors);
```
mchandschuh: the refresh token is only being used for communication with /processOanda, I think this can be safely removed

Common/Orders/Order.cs
```diff
@@ -285,5 +285,9 @@ public static Order CreateOrder(SubmitOrderRequest request)
             }
             return order;
         }
+        /// <summary>
+        /// Order Expiry on a specific UTC time.
+        /// </summary>
+        public DateTime DurationValue;
```
mchandschuh: @jaredbroad -- thoughts on this?  I think implementing it wouldn't be too bad, each time we do a scan in `BacktestingBrokerage.Scan` we would just need to check that we haven't passed this time, this is a fairly general way of expressing expiry, maybe change the name, turn it into a property with internal set like others.

Common/Orders/OrderTypes.cs
```diff
@@ -68,6 +68,12 @@ public enum OrderDuration
         /// </summary>
         Day
         */
+
+        /// <summary>
+        /// Order valid until specific type.
+        /// </summary>
+        Specific
```
mchandschuh: this would need to be implemented in the engine. I'm thinking long term this enumeration is an API convenience thing, but the raw data would be a date to expire? thoughts?

QuantConnect.Lean.sln
```diff
@@ -51,11 +51,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "QuantConnect.Algorithm.Pyth
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "QuantConnect.Algorithm.Java", "Algorithm.Java\QuantConnect.Algorithm.Java.csproj", "{61097E9E-EB20-4677-B394-CE0682B9D610}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".nuget", ".nuget", "{93A63A28-9326-4E26-9930-0FDBC26B9A91}"
-	ProjectSection(SolutionItems) = preProject
-		.nuget\packages.config = .nuget\packages.config
-	EndProjectSection
-EndProject
 Global
```
mchandschuh: This section got removed... maybe this is why the travis build was failing?

Tests/Brokerages/Oanda/OandaBrokerageTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            return oandaBrokerage;
+        }
+
+        /// <summary>
+        ///     Gets the symbol to be traded, must be shortable
+        /// </summary>
+        protected override string Symbol
+        {
+            get { return "EUR_USD"; }
```
mchandschuh: the symbol should be 'EURUSD' like QC expresses it, we'll need to make changes to do the symbol mapping between QC and Oana

mchandschuh:
@kaffeebrauer - I've made a first pass here picking out things I could. Brokerage implementations are extremely complicated and have many concerns (sadly). I expect we'll do a few rounds of review here. 

Absolutely herculean effort! This looks much cleaner than the tradier implementation, very awesome stuff man!

Oh yea, one more note, I would prefer if contributors used `git rebase` instead of `git merge` as it keeps the history (more) linear. Have a peak at CONTRIBUTING.md for the steps used to update your local branch using the rebase option. I tend to do the following:
```bash
$ git fetch
$ git rebase origin/master
```

Common/Orders/OrderTypes.cs
jaredbroad: Agree is API convenience and each order should carry a time where it expires (DateTime.MaxValue). Perhaps better description of this order duration would be "Custom"

Common/Orders/Order.cs
jaredbroad: yes! #awesome! It should default to DateTime.MaxValue and we can scan it. Later we can actually implement the GTC / Day options! 

Potentially we should also include the Enum value here since if sending in *real* orders to brokerages you'd want to use their GTC / Day settings and not fake it with LEAN. 

To avoid scope creep I vote leaving it as "Expiry" (no enum) & addressing timed expiration scanning / sending-live-orders properly in a new issue.

Brokerages/Backtesting/OandaBrokerageModel.cs
kaffeebrauer: Agreed, I've moved this restriction to CanExecuteOrder(Security, Order).

Brokerages/Backtesting/OandaBrokerageModel.cs
kaffeebrauer: I've moved the duration limitation to CanExecuteOrder, but I am not too sure on marking the Order's property as Invalid we never pass the modified order.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Agreed, probably do not need an entire method for creating a comma delimited string.

Brokerages/Backtesting/OandaBrokerageModel.cs
kaffeebrauer: Agreed, I didn't consider the scenario when the system was backtesting.

Brokerages/Oanda/DataType/Account.cs
kaffeebrauer: Hi Mike, Yes this was a copied from https://github.com/oanda/rest-csharp-refimpl, I guess I'll removed the contributing section for those classes then?

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: No there isn't a requirement to close the 10 then open the other order. Their transaction handler is smart enough. I've removed this _holdingsProvider property in the Brokerage.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Yes, I've removed this line, could potentially need it in the future.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Agreed.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Yes it's misleading :). Nonetheless because cachedOpenOrdersByOandaOrderId isn't getting used, I've removed this line.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Added a helper method to remove the underscore if it's Forex.

One question though, how would you map other security types when we map it to QC? For example the S&P 200 ASX Index is mapped as 'AU200_AUD' and the UK 10 Year Bond is mapped as 'UK10YB_GBP'.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: In Oanda you can select the currency the account is represented in. It's usually in the following:

Australian Dollar (AUD)
Canadian Dollar (CAD)
Swiss Franc (CHF)
Euro (EUR)
British Pound (GBP)
Hong Kong Dollar (HKD)
Japanese Yen (JPY)
Singapore Dollar (SGD)
US Dollar (USD)

In QC is currently everything represented as USD?

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Hi Michael, When we call Oanda's Get Account Balance, It returns Current Cash in Balance + Current Trades Margins + Current Trade Gains/Loss.

When we query the Oanda Account (Oanda Currency Account) it only has one currency value. One user might have multiple Oanda Currency Subaccounts (for different currencies).

In this case we need to poll all their correspondence accounts (each one has a unique ID). There is where I'm not too certain how we want to store this in the configuration (perhaps make it comma delimited in the config.json?).

Tests/Brokerages/Oanda/OandaBrokerageTests.cs
kaffeebrauer: Done

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Done, was following the convention used in the Tradier brokerage implementation.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Agreed, the streaming event API doesn't have partial filled events based on my experience.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Good idea, I guess you have a strong dislike for foreach loops :)

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Agreed, this makes the code a bit more readable.

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: If I'm reading those securities correctly, it's read as the AU200 priced in AU, or the UK10YB priced in GBP. QC needs a little more engine work to support the different currencies on securities, so for now I'm not sure we can support. Eventually I imagine there will be a Security.CurrencySymbol property or something similar that would denote what currency is used to trade that security.

@jaredbroad - not sure if you have thoughts on this currently? I think for now we just can't support? Need more currency support for securities.

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: We use the USD conversion for computing the portfolio value, but we do store currencies in their native units, for example, if an account holds 10k CHF, we'll store it as 10k CHF, but we stil need the CHF->USD conversion for portfolio value.A little more complicated, if the account has a forex position, say 10k EUR/GBP, that's really the client sold GBP for EUR. We would need the EUR->GBP rate (the market price of the position) as well as the GBP->USD rate so we could compute portfolio value.

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: Ideally we can ask the Oanda API for these sub-accounts. If not, maybe we can talk to Oanda and see if they can add this feature. You're correct though, we'll need to poll each individual sub-account to get their holdings in those currencies, say (EUR,GBP,CHF,ect...) and return them via the GetCashBalances() call used to initialize the algorithm in live mode

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: Indeed, I think since then I've added the ToString override, thanks for updating :D

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: hehe, yea, I tend to be minimalist in some cases, I think it stems from my desire to have each method to have focus on its primary goal, so a setup step like this as a one liner fits that desire, but toe-may-toe-; toe-mah-toe :D

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: pinging @jaredbroad

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: I think this should still be updated to using a Continuation block and only doing it if the cancel succeeds, or maybe I'm misunderstanding the asyncness of this operation

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: bump - sometimes these can get lost in large reviews :D

Brokerages/Oanda/OandaBrokerage.cs
jaredbroad: What cases would an update fail? (e.g. just order is already cancelled, order is already filled?) or are there more sinister cases? Lets log them and find out, for now -- lets throw fatal error & terminate algorithm. 

I think if its order cancelled/filled the order state would probably reflect it (within a few milliseconds) and so its all handled.

If its a "500 server error / timeout" then we need to throw to keep algo state consistent.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: I've implement the the get cash balances method, however there could a bit of cost trying to use Oanda's Rates API (https://www.oanda.com/rates/#plans), perhaps we can convert the currency to the balance currency (USD) via a different provider?

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: This would happen that when the order in question is already filled/cancelled.

I've added an OnOrderEvent declaring the order as invalid back to QC with the message 'Order currently does not exist with order id: {0}.'

Would this suffice?

If it's a 500 error or timeout then the method will throw an exception before reaching that statement.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Just reviewing this part of the code, I originally intended it to async (following Oanda's own implementation). Since then, I've started to move to code base for Oanda to not run Async, this is probably one part I missed.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Agreed, removed the Exception thrown. Non-event streaming API of Oanda does not disconnect per se.

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Done. Looks shorter and less superfluous

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Casting just make you feel safer that's all :)

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Agreed!

Brokerages/Oanda/OandaBrokerage.cs
kaffeebrauer: Refactored this method to be not async,

Brokerages/Oanda/OandaBrokerageFactory.cs
kaffeebrauer: That's been microwaved and eaten...

Common/Orders/Order.cs
kaffeebrauer: With the three different variations of the Order constructors; because the Duration is set as OrderDuration.GTC, I'll set the duration value by default as DateTime.MaxValue as discussed.

Common/Orders/OrderTypes.cs
kaffeebrauer: Renamed this Enum as Custom.

Brokerages/Backtesting/OandaBrokerageModel.cs
mchandschuh: This is fine, my solution was certainly a hack :D this will be resolved when we further implement order durations

Brokerages/Oanda/DataType/Account.cs
mchandschuh: Yes, we'll need to include their original license, not our license on those files, and they should remain unmodified. thank you!

Brokerages/Backtesting/OandaBrokerageModel.cs
jaredbroad: Better late than never: Agree order duration is the real solution :)

Brokerages/Oanda/OandaBrokerage.cs
jaredbroad: Correct for now its best to not support. The Security base class multiple currency support is on the todo list for after universe and history functions.

Brokerages/Oanda/OandaBrokerage.cs
mchandschuh: That sounds good, thanks

mchandschuh:
Once again, awesome work @kaffeebrauer. I made some minor comments, what are your thoughts on the stability/readiness of this PR?

mchandschuh:
Great work @kaffeebrauer, I'll get this into an /oanda branch tonight/tomorrow morning and Jared and I will begin testing it in the cloud.

mchandschuh:
I'll be merging this into an /oanda branch today. I'll fix up the build.

mchandschuh:
Any further commits and be submitted as PRs to the /oanda branch

mchandschuh:
This was successfully merged in https://github.com/QuantConnect/Lean/commit/277f95dbe436a7d4b4bd9c07e1cc5319d9a83860.

Have a peak at the new [oanda branch](https://github.com/QuantConnect/Lean/commits/oanda)

Many thanks @kaffeebrauer!! Heroic contribution!

mchandschuh:
This was merged into master in https://github.com/QuantConnect/Lean/commit/78e12d7fe78cd87ff2e9ea67b71a7b965e64929c

Issue #96, opened 2015-06-07
Add Support for Algorithm Parameter Inputs (Configurable variables)


mattycourtney:
Enable finding the optimal set of parameters to trade an algorithm with

gururise:
I'm thinking parameter optimization could be easily supported by allowing optional parameters to be passed into the Engine instance when it first starts.  Once the backtest is complete, the results recorded, and another instance with new parameters is started.  This whole process would occur in a loop under some sort of user control.  By putting this process in a loop, the user could then either do some type of simple parameter optimization via brute-force, or utilize a GA or some other form of Evolutionary Programming.

mchandschuh:
I think the concept of parameter optimization is slightly outside of the scope of QuantConnect.Lean.Engine.csproj code.  I think what we really need are parameterized inputs for the algorithm and for a job to be able to accept/pass these parameterized inputs.  In this way, the infrastructure that runs Lean can perform the parameter optimization search, whether it be the QC cloud infrastructure or a local desktop Lean runner.

I think specifying the parameters in the algorithm code can be accomplished rather cleanly through the usage of C# attributes:
```c#
public class MyParameterizedAlgorithm : QCAlgorithm
{
    [IntParameter(Minimum: 3, Maximum: 50)]
    public int SimpleMovingAveragePeriod = 14;

    // indicator instantiated using a parameter
    public SimpleMovingAverage MySMA;

    public override void Initialize()
    {
        //...
        MySMA = SMA("SPY", SimpleMovingAveragePeriod);
    }
}
```

This would allow the infrastructure to inspect the algorithm to find parameters very easily, as well as display them to a desktop UI if needed.

Once parameterized inputs for an algorithm are implemented, then we could implement the optimization feature by running multiple lean instances and setting the algorithm parameters based on the optimization algorithm, be it brute force or some evolutionary algorithm.

Thoughts?

TangentLogic:
Don't know if anyone else is working on this; I'm working on an implementation (including a brute-forcer proof-of-concept).  I expect to submit a patch around the end of this week.

jaredbroad:
Awesome @TangentLogic! Look forward to seeing your design. Are you using C# attributes like demonstrated above?

From our past experiences its best to talk through the design publicly as early as possible to get the benefit of the "hive mind" :)

TangentLogic:
[...]

PR #153, opened 2015-08-14
#96-Parameterization

Vincent-Chin:
Unit tests done.  Note that the Engine.cs I've included is modified with rudimentary brute-forcing logic.  It doesn't work though (hangs on initializing the second iteration.)  I also never got around to parallelizing the loop with Parallel.ForEach.

### DIFF
--- a/Algorithm.CSharp/ParameterizedAlgorithm.cs
+++ b/Algorithm.CSharp/ParameterizedAlgorithm.cs
@@ -1,0 +1,127 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * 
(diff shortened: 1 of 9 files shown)


Vincent-Chin:
LoadParameters.cs in Tests is an extraneous file.  Didn't notice it during the commit.

jaredbroad:
This is a great first step @Vincent-Chin! There's some changes we'll need to make to properly use it in LEAN though. If you're open to making them we can merge in the PR.

1. The configuration parameter set must come through the Job packet, and, 
2. The job would have 1-set of parameters.
3. Revert Engine.cs & replace it will a simple way of setting parameters from the job packet.

This way LEAN is still a singular instance, running a configured parameter set. 

3. Later, we'll make the GUI spin up hundreds of LEAN instance objects each with different job packets specifying their configuration permutation.

mchandschuh:
I think it would also be beneficial to separate the concept of the parameter from the concept of the search method. For example, int parameter defines a min/max/step property. While the min/max do define the parameter, the step defines how we will search the parameter space. This is a concern best saved for optimization, so as part of this PR let's remove the Step properties.

jaredbroad:
Thanks for the PR @Vincent-Chin ! This is an awesome start but we'd like to design Parameterization with a more holistic approach in mind where parameters are passed through the jobs system, and its well integrated with the Engine class.

Lets keep this on reference and come back to it later once we start that project!

PR #155, opened 2015-08-25
created log return indicator and tests

chutes:
created log return indicator and tests


mchandschuh:
I can produce some  test data for you, I'll push it tonight.

chutes:
Thanks

mchandschuh:
https://github.com/QuantConnect/Lean/commit/52531aa216f5fe22e86b8a44f6831303c5e35d16 adds LOGR14 test data from freestockcharts.com using formula: LOG(c/c14)

Indicators/LogReturn.cs
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    ///     Represents the LogReturn indicator (LOGR)
```
mchandschuh: Can we add a little bit more here, maybe include the equation used and why someone would use it. Here's a good [read]https://quantivity.wordpress.com/2011/02/21/why-log-returns/) to start with.

mchandschuh:
@chutes - thanks for the great contribution! Code/doc/tests look great.

Issue #30, opened 2015-02-21
Create StatisticsResults Class, Refactor Statistics Class


jaredbroad:
Refactor the Statistics class to be an instance driven, with a constructor which takes the required information and creates an instance of Statistics. 

Statistics object should define all the currently generated statistics as public properties and include description parameters for the English text.

Benchmark should be separated out and implement an IBenchmark; be generated & cached in a local file once per day. 

It would require two new algorithm helper methods:
1. void SetBenchmark(IStatisticsBenchmark) - in the initialize set the IBenchmark to use for generating the statistics.
2. Statistics GetStatistics(TimeSpan period) - In the QCAlgorithm generate an instance of the statistics object for a specified timespan (default to all the algorithm run time).

Replace the statistics generation in Engine with a call to algorithm.GetStatistics, (and update the Result handlers to take a Statistics Object instead of a Dictionary<string,string>.

jaredbroad:
+ Note: Benchmark ideally would be customizable, numerous users requested custom benchmarks. This could be specific symbols, or functions which return a decimal value each day.

StefanoRaggi:
Here are my first thoughts on where to get started:

*Step 1*
- Create new Trade class (properties: Symbol, EntryTime, EntryPrice, Direction, Quantity, MAE, MFE, ExitTime, ExitPrice, NetProfit)
- Change type of SecurityTransactionManager.TransactionRecord from Dictionary\<DateTime,decimal\> to List\<Trade\>
- In BrokerageTransactionHandler.HandleOrderEvent when order filled, open new Trade and/or close open Trade (update _algorithm.Transactions.TransactionRecord)
- In Security.SetMarketPrice update MAE/MFE/NetProfit of current Trade (if open);  requires adding a reference to SecurityTransactionManager in Security class?

*Step 2*
- Create new AlgorithmPerformance class (properties: TotalTrades, AverageWin, AverageLoss, etc.)
- Constructor: AlgorithmPerformance(List\<Trade\> trades, Benchmark benchmark, decimal startingPortfolioValue, decimal totalFees)
- Method Generate(): calculates all values of all properties
- Benchmark could be optional (in this case relevant stats will not be calculated) 

Waiting for comments and suggestions...

jaredbroad:
I like the design using Trade objects. This will simplify creating rolling windows of statistics (e.g. Get all trades between X-Y, generate statistics).

[...]

PR #164, opened 2015-09-13
New Statistics + Rolling Statistics

StefanoRaggi:
Issue #30


Common/Orders/OrderEvent.cs
```diff
@@ -33,11 +33,21 @@ public class OrderEvent
         public Symbol Symbol;
 
         /// <summary>
+        /// The date and time of this event (UTC).
+        /// </summary>
+        public DateTime Time;
+
+        /// <summary>
```
mchandschuh: Can we rename this to UtcTime so it's more explicit

Common/Orders/OrderEvent.cs
```diff
@@ -111,6 +121,25 @@ public OrderEvent(Order order, string message = "")
...
         /// <summary>
+        /// Helper Constructor for TradeBuilder tests
+        /// </summary>
+        /// <param name="orderId">The Id of the order</param>
+        /// <param name="symbol">Asset Symbol</param>
+        /// <param name="time">Date/time of this event</param>
+        /// <param name="fillPrice">Fill price</param>
+        /// <param name="fillQuantity">Fill quantity</param>
+        /// <param name="orderFee">The order fee</param>
+        public OrderEvent(int orderId, string symbol, DateTime time, decimal fillPrice, int fillQuantity, decimal orderFee)
```
mchandschuh: Since we now require the time, I would prefer if we made it a required parameter on all constructors for the `OrderEvent` class. After that is added, this constructor can probably go away and we can define a helper method in the statistics test if need be, the reason being I don't want a developer to accidentally use this ctor which doesn't specify a status and it uses the `string symbol` instead of `Symbol symbol`

Common/Orders/OrderEvent.cs
```diff
@@ -123,6 +152,15 @@ public override string ToString()
...
                 : string.Format("OrderID: {0} Symbol: {1} Status: {2} Quantity: {3} FillPrice: {4}", OrderId, Symbol, Status, FillQuantity, FillPrice);
         }
+
+        /// <summary>
+        /// Returns a clone of the current object.
+        /// </summary>
+        /// <returns>The new clone object</returns>
+        public OrderEvent Clone()
+        {
+            return (OrderEvent)MemberwiseClone();
```
mchandschuh: This should be safe given the only reference type `Symbol` is immutable. Also, this is certainly not performance sensitive.

Common/Statistics/AlgorithmPerformance.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// </summary>
+        public PortfolioStatistics PortfolioStatistics { get; private set; }
+
+        /// <summary>
+        /// The list of closed trades
+        /// </summary>
+        public List<Trade> ClosedTrades { get; private set; }
+
+        /// <summary>
+        /// Initializes a new instance of the TradeStatistics class
```
mchandschuh: Looks like a lingering previous comment from a rename, recently I've been using:
```xml
 <see cref="AlgorithmPerformance"/>
```
in the comments so refactoring tools pick up the renames for me

Common/Statistics/StatisticsBuilder.cs
```diff
@@ -0,0 +1,304 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Helper class for rolling statistics
+        /// </summary>
+        private class PeriodRange
+        {
+            internal DateTime StartDate { get; set; }
+            internal DateTime EndDate { get; set; }
```
mchandschuh: I think this is a great type to have, we should keep this in mind if we need a similar structure in the future and elevate  it into the `QuantConnect` common namespace, for now this is a great place for it until we need to reuse a similar concept.

Common/Statistics/TradeBuilder.cs
```diff
@@ -0,0 +1,486 @@
...
+    public class TradeBuilder
+    {
+        /// <summary>
+        /// Helper class to manage pending trades and market price updates for a symbol
+        /// </summary>
+        private class Position 
+        {
+            internal List<Trade> PendingTrades { get; set; }
+            internal List<OrderEvent> PendingFills { get; set; }
+            public decimal TotalFees { get; set; }
```
mchandschuh: why is this one public? for a private class internal and public are equivalent

Common/Statistics/TradeBuilder.cs
```diff
@@ -0,0 +1,486 @@
...
+        /// <param name="conversionRate">The current market conversion rate into the account currency</param>
+        public void ProcessFill(OrderEvent fill, decimal conversionRate)
+        {
+            // If we have multiple fills per order, we assign the order fee only to its first fill
+            // to avoid counting the same order fee multiple times.
+            var orderFee = 0m;
+            if (!_ordersWithFeesAssigned.Contains(fill.OrderId))
+            {
+                orderFee = fill.OrderFee;
+                _ordersWithFeesAssigned.Add(fill.OrderId);
```
mchandschuh: Looks like this `HashSet<int>` will grow unbounded, I've contemplated the impact of this many times, but in other places I've used the `FixedSizeHashQueue<T>` which will remove items from the collection after it hits a certain limit.

Common/Orders/OrderEvent.cs
StefanoRaggi: The fields added to the OrderEvent class are actually two: UtcTime and OrderFee (both are required by the TradeBuilder). I agree this constructor is only used for testing purposes and is similar to another existing one (also used only in testing). Maybe I could add these two parameters to the existing and remove this one ? The other constructor OrderEvent(Order) could be left untouched (at least for now).

Common/Statistics/AlgorithmPerformance.cs
StefanoRaggi: Copy/paste... Good idea to use see cref.

Common/Statistics/TradeBuilder.cs
StefanoRaggi: No reason. Copy/paste again

Common/Statistics/TradeBuilder.cs
StefanoRaggi: I like the FixedSizeHashQueue class. Only in live mode, as Jared requested, TradeBuilder caps the number of trades to 10000. Normally two orders are needed to complete a trade, so shall we limit the OrderIds to 20000? Only in live mode?

Common/Extensions.cs
```diff
@@ -230,6 +230,15 @@ public static long ToInt64(this string str)
         }
 
         /// <summary>
+        /// Check if a number is NaN or equal to zero
+        /// </summary>
+        /// <param name="value">The double value to check</param>
+        public static bool IsNaNOrZero(this double value)
+        {
+            return double.IsNaN(value) || Math.Abs(value) < double.Epsilon;
```
mchandschuh: Is this check intended to include `double.Epsilon`? It looks like this is really doing:
```C#
return double.IsNan(value) || value == 0;
```
Since no positive double can exist between `double.Epsilon` and 0.

No change needed, the comments express this behavior clearly, just thinking out loud.

Engine/Results/IResultHandler.cs
```diff
@@ -197,7 +198,14 @@ bool IsActive
...
         /// </summary>
-        void SendFinalResult(AlgorithmNodePacket job, Dictionary<int, Order> orders, Dictionary<DateTime, decimal> profitLoss, Dictionary<string, Holding> holdings, Dictionary<string, string> statistics, Dictionary<string, string> banner);
+        /// <param name="job">Lean AlgorithmJob task</param>
+        /// <param name="orders">Collection of orders from the algorithm</param>
+        /// <param name="profitLoss">Collection of time-profit values for the algorithm</param>
+        /// <param name="holdings">Current holdings state for the algorithm</param>
+        /// <param name="statisticsResults">Statistics information for the algorithm (empty if not finished)</param>
+        /// <param name="statistics">Statistics information for the algorithm (empty if not finished)</param>
+        /// <param name="banner">Runtime statistics banner information</param>
+        void SendFinalResult(AlgorithmNodePacket job, Dictionary<int, Order> orders, Dictionary<DateTime, decimal> profitLoss, Dictionary<string, Holding> holdings, StatisticsResults statisticsResults, Dictionary<string, string> statistics, Dictionary<string, string> banner);
```
mchandschuh: I think we can remove the `Dictionary<string,string> statistics` parameter as it is the same as `statisticsResults.Summary` ?

Common/Statistics/TradeBuilder.cs
```diff
@@ -0,0 +1,486 @@
...
+
+            public Position()
+            {
+                PendingTrades = new List<Trade>();
+                PendingFills = new List<OrderEvent>();
+            }
+        }
+
+        private readonly List<Trade> _closedTrades = new List<Trade>();
+        private readonly Dictionary<string, Position> _positions = new Dictionary<string, Position>();
```
mchandschuh: I think we'll need to have this be `Dictionary<Symbol, Position>` so it properly handles mapping changes. Ideally, everywhere in the LEAN code base we'll refer to symbols using the `Symbol` type, only in Tests and the algorithm implementation will we allow the usage of the `string symbol` -- this is because eventually we'll move to a unique symbol id system that may not be symbol base at all (maybe an integer identifier)

Common/Statistics/TradeBuilder.cs
```diff
@@ -0,0 +1,486 @@
...
+        {
+            get { return _closedTrades; }
+        }
+
+        /// <summary>
+        /// Returns true if there is an open position for the symbol
+        /// </summary>
+        /// <param name="symbol">The symbol</param>
+        /// <returns>true if there is an open position for the symbol</returns>
+        public bool HasOpenPosition(string symbol)
```
mchandschuh: use `Symbol` instead of `string symbol`

Common/Statistics/TradeBuilder.cs
```diff
@@ -0,0 +1,486 @@
...
+
+            return position.PendingFills.Count > 0;
+        }
+
+        /// <summary>
+        /// Sets the current market price for the symbol
+        /// </summary>
+        /// <param name="symbol"></param>
+        /// <param name="price"></param>
+        public void SetMarketPrice(string symbol, decimal price)
```
mchandschuh: use `Symbol` instead of `string symbol`

mchandschuh: Another possibility is having the `TradeBuilder` hold onto `Security` objects, then he'll never need to have price data pushed to him, but rather he'll have it by virtue of the security instance. This will also help performance to remove the multiple dictionary lookups per time step per symbol.

Common/Orders/OrderEvent.cs
mchandschuh: Also, we need to verify this is being set by everyone who creates this. Recommend making it a required constructor parameter. I'm mostly thinking of the brokerage implementation who don't know to set this value.

Common/Orders/OrderEvent.cs
```diff
@@ -33,11 +33,21 @@ public class OrderEvent
...
+
+        /// <summary>
         /// Status message of the order.
         /// </summary>
         public OrderStatus Status;
 
         /// <summary>
+        /// The fee associated with the order (always positive value).
+        /// </summary>
+        public decimal OrderFee;
```
mchandschuh: Also, we need to verify this is being set by everyone who creates this. Recommend making it a required constructor parameter. I'm mostly thinking of the brokerage implementation who don't know to set this value. If the call site doesn't have the required info, they can pass zero, but at least it forces the question to all who instantiate this type.

Engine/Engine.cs
```diff
@@ -352,12 +355,20 @@ public void Run(AlgorithmNodePacket job, string assemblyPath)
...
-                                var benchmark = charts[benchmarkKey].Series[benchmarkKey].Values.ToDictionary(chartPoint => Time.UnixTimeStampToDateTime(chartPoint.x), chartPoint => chartPoint.y);
-                                statistics = Statistics.Statistics.Generate(equity, profitLoss, performance, benchmark,
-                                    _algorithmHandlers.Setup.StartingPortfolioValue, algorithm.Portfolio.TotalFees, numberOfTrades, 252);
+                                var profitLoss = new SortedDictionary<DateTime, decimal>(algorithm.Transactions.TransactionRecord);
+                                var totalTransactions = algorithm.Transactions.GetOrders(x => x.Status.IsFill()).Count();
+
+                                //var benchmark = charts[benchmarkKey].Series[benchmarkKey].Values.ToDictionary(chartPoint => Time.UnixTimeStampToDateTime(chartPoint.x), chartPoint => chartPoint.y);
+                                //statistics = Statistics.Statistics.Generate(equity, profitLoss, performance, benchmark,
+                                //    _algorithmHandlers.Setup.StartingPortfolioValue, algorithm.Portfolio.TotalFees, totalTransactions, 252);
```
mchandschuh: we can delete this commented out lines

Common/Statistics/StatisticsBuilder.cs
```diff
@@ -0,0 +1,304 @@
...
+            var equity = ChartPointToDictionary(pointsEquity);
+
+            var firstDate = equity.Keys.FirstOrDefault().Date;
+            var lastDate = equity.Keys.LastOrDefault().Date;
+
+            var totalPerformance = GetAlgorithmPerformance(firstDate, lastDate, trades, profitLoss, equity, pointsPerformance, pointsBenchmark, startingCapital);
+            var rollingPerformances = GetRollingPerformances(firstDate, lastDate, trades, profitLoss, equity, pointsPerformance, pointsBenchmark, startingCapital);
+            var summary = GetSummary(totalPerformance, totalFees, totalTransactions);
+
+            Results = new StatisticsResults(totalPerformance, rollingPerformances, summary);
```
mchandschuh: could this be just a static method that returns the `StatisticsResults` directly? Maybe `Statistics.Generate(...)` to match the previous, but now it returns the new fancy `StatisticsResults` class. This would allow for more natural call site invocations, currently its:
```C#
var builder = new StatisticsBuilders(...);
var results = builder.Results;
```
But could be written:
```C#
var results = Statistics.Generate(...);
```

Thoughts?

Tests/RegressionTests.cs
```diff
@@ -29,7 +29,7 @@ public void BasicTemplateAlgorithm()
                 {"Total Trades", "1"},
                 {"Average Win", "0%"},
                 {"Average Loss", "0%"},
-                {"Compounding Annual Return", "311.484%"},
+                {"Compounding Annual Return", "280.522%"},
```
mchandschuh: do we know why these were revised lower? just curious. I would have suspected our default statistics mode to match the old method.

Common/Extensions.cs
StefanoRaggi: Yes, I know, it was just to make R# happy :smile:

Engine/Results/IResultHandler.cs
StefanoRaggi: Yes, I left it there temporarily, just to compare results with the original Statistics code. The intention was to remove it in a later commit.

Common/Statistics/TradeBuilder.cs
StefanoRaggi: The TradeBuilder class was created before the Symbol class, but yes, I agree and will change it.

Common/Statistics/TradeBuilder.cs
StefanoRaggi: Ok for string -> Symbol. 

The problem with the Security object is that TradeBuilder needs to be notified of price changes to update MinPrice/MaxPrice for MAE/MFE calculation. Also, I didn't like adding the dependency.

Common/Statistics/StatisticsBuilder.cs
StefanoRaggi: We can make the class StatisticsBuilder static with only one public method:

`var results = StatisticsBuilder.Generate(...)`

Tests/RegressionTests.cs
StefanoRaggi: I ran the code on the master branch and the tests were failing, so I updated them until they passed, then I switched back to the statistics branch.
Maybe because I was working on Saturday/Sunday?

Common/Orders/OrderEvent.cs
StefanoRaggi: Maybe we should rename Order.Time to Order.UtcTime as well?

mchandschuh: Agree on the Order.Time, the only trouble there is it might be used in current algorithms, so we'll check to see how many users actually reference it. I just don't want to break existing code if possible. As for using UtcTime = order.Time, that isn't completely correct. order.Time is the time the order was placed in UTC, but orderEvent.UtcTime is the time the fill was produced, so I think we need to accept the time parameter explicitly. My concern is that I want new implementations to be forced to set this value correctly, vs assuming it will be set automatically.

Common/Orders/OrderEvent.cs
mchandschuh: If possible let's merge the two test helper constructors, that sounds great. As for the other `OrderEvent(Order)` constructor, I think we'll need to add a utcTime parameter and an orderFee parameter. I'm ok with these being addressed in a separate PR.

Common/Statistics/TradeBuilder.cs
mchandschuh: I think it's safe to do in live or backtest, but I may be mistaken. My thought is that this hash is really only used to prevent duplicating order fees, i.e. multiple fills that each have an order fee component. I think it's a safe assumption that a single order won't have two fills separated by 20k other orders. So I think we can always used the `FixedSizeHashQueue<T>` class, but either way is fine for me. My main concern here is certainly live mode stability.

Common/Extensions.cs
mchandschuh: #loveresharper

In other news, I try not to pollute our codebase with resharper hints, but there just for your information (and anyone following), there is a // Resharper Disable for floating point comparison with zero to remove the R# warning.

Engine/Results/IResultHandler.cs
mchandschuh: Sounds perfect, thanks.

Common/Statistics/TradeBuilder.cs
mchandschuh: Agreed on the extra dependency, when all is said and done we'll need to do some performance testing to confirm/deny the impact.

Common/Statistics/StatisticsBuilder.cs
mchandschuh: That sounds great

Tests/RegressionTests.cs
mchandschuh: Interesting, I can run the tests locally and they pass. Hopefully there's not a day of week dependency in the tests (I've done it accidentally before). Maybe it's an input data issue?

Common/Orders/OrderEvent.cs
StefanoRaggi: Oops. You're absolutely right: orderEvent.UtcTime is the time the event is generated (that can be a fill or not ?) and order.Time is the order placement time. I'll fix it immediately.
I suspected order.Time could be used in algorithm, so I didn't touch it.

Common/Orders/OrderEvent.cs
StefanoRaggi: I had already merged the two constructors in the last commit, they both have the two new parameters. I still have to add the utcTime parameter to the OrderEvent(Order) constructor.

Common/Statistics/TradeBuilder.cs
StefanoRaggi: I agree that fills for the same order should be close in time, I'll define a new smaller constant for the FixedSizeHashQueue size (maybe 1000), used both in live and backtesting.

Tests/RegressionTests.cs
StefanoRaggi: I ran the tests on the latest master branch using the json and default data files in Github. The only test still failing for me is the UniverseSelectionRegressionAlgorithm (the only one I didn't change), but this is because the data files required for it are missing.

StefanoRaggi:
The fixes for the second round of review are completed (last two commits)

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/fac6f46a9abca09eb8e6384af39271980bd118d1

jaredbroad:
Pretty epic PR @StefanoRaggi, great work.

PR #170, opened 2015-09-22
Issue 169 Rolling Window

bizcad:
See issue 169 for details. Fixes bug where RollingWindow[1] gets the wrong value when the Indicator is not ready.


mchandschuh:
Closed via #171

Issue #172, opened 2015-09-23
Setting DesktopResultHandler returns a ConsoleResultHandler


chrisdk2015:
On latest Mono/Ubuntu 14.04 here, the line 

Config.Set("result-handler", "QuantConnect.Lean.Engine.Results.DesktopResultHandler");

in LeanEngineWinForm.cs doesn't result in a DesktopResultHandler getting returned but a ConsoleResultHandler.

I couldn't figure out why my cast to a DesktopResultHandler failed but now I know.

If you look in PollingOnTick you can see the type.

StefanoRaggi:
The following line should be added in the `LeanEngineWinForm.Main()` function:

```
Config.Set("environment", "desktop");
```

chrisdk2015:
Yes, then it works. Thanks.

PR #173, opened 2015-09-23
Added environment setting in WinForms desktop app

StefanoRaggi:
It loaded ConsoleResultHandler instead of DesktopResultHandler (issue #172)


mchandschuh:
Nice, now this works out of this box. Thanks!

PR #182, opened 2015-10-05
Momersion indicator and test added.

Jay-Jay-D:
Jared and Michael, here is the Momersion indicator implementation and the respective test.

### DIFF
--- a/Indicators/Momersion.cs
+++ b/Indicators/Momersion.cs
@@ -1,0 +1,144 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


Indicators/Momersion.cs
```diff
@@ -0,0 +1,144 @@
...
+                Mc = _multipliedDiffWindow.Count(obs => obs > 0);
+                MRc = _multipliedDiffWindow.Count(obs => obs < 0);
+                momersion = 100m * Mc / (Mc + MRc);
+            }
+            else
+            {
+                momersion = 50m;
+            }
+
+            Current = new IndicatorDataPoint(input.Time, momersion);
```
mchandschuh: No need to set `Current` within the `ComputeNextValue` functions on indicators. They're designed such that you just return the new value and the base class will update the current

Tests/Indicators/MomersionTest.cs
```diff
@@ -0,0 +1,93 @@
+﻿using NUnit.Framework;
```
mchandschuh: Needs banner here

mchandschuh:
@Jay-Jay-D - awesome stuff! Will pull in later tonight, thanks!

Jay-Jay-D:
I'm so glad to help, thank you!

bizcad:
Hi JJ,

 

I hope you have a good trip.  Here is an interesting idea.

 

http://archive.aweber.com/awlist3824363/4iANX/h/How_to_trade_the_equity_curve.htm

 

Nick

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/9c2e8b03868d1724a9f1303a04a592b892465fc6

PR #185, opened 2015-10-08
Allow user to forcefully open the Market for equity and forex

desimonk:
Allow user to forcefully keep market open using config.json flag.


mchandschuh:
Hey, sorry it took a while to get to this. Could we change the configuration to be true/false values?

desimonk:
No problem.  I will send another pull request.

mchandschuh:
Pulled in #189

PR #187, opened 2015-10-15
FXCM Brokerage implementation

StefanoRaggi:


Brokerages/Fxcm/FxcmBrokerage.DataQueueHandler.cs
```diff
@@ -0,0 +1,121 @@
...
+using QuantConnect.Data.Market;
+using QuantConnect.Logging;
+using QuantConnect.Packets;
+
+namespace QuantConnect.Brokerages.Fxcm
+{
+    public partial class FxcmBrokerage
+    {
+        private readonly List<Tick> _ticks = new List<Tick>();
+        private readonly HashSet<string> _subscribedSymbols = new HashSet<string>();
```
mchandschuh: We should use the `Symbol` class instead of strings for everywhere in the engine, if possible.

mchandschuh: However, it looks like the `IDataQueueHandler.Subscribe` function needs to be updated to transport `Symbol` instances instead of strings, so for now, let's just modify this to be a `HashSet<Symbol>` and we'll add the signature change to the list.

Brokerages/Fxcm/FxcmBrokerage.DataQueueHandler.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using com.fxcm.fix;
+using com.fxcm.fix.pretrade;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Logging;
+using QuantConnect.Packets;
+
+namespace QuantConnect.Brokerages.Fxcm
```
mchandschuh: Maybe some doc at the top of the partials to describe what this part does

Brokerages/Fxcm/FxcmBrokerage.Messaging.cs
```diff
@@ -0,0 +1,443 @@
...
+                _mapRequestsToAutoResetEvents[_currentRequest] = autoResetEvent;
+            }
+            if (!autoResetEvent.WaitOne(ResponseTimeout))
+                throw new TimeoutException(string.Format("FxcmBrokerage.LoadOpenPositions(): Operation took longer than {0} seconds.", (decimal)ResponseTimeout / 1000));
+        }
+
+        /// <summary>
+        /// Gets the quotes for the symbol
+        /// </summary>
+        public List<MarketDataSnapshot> GetQuotes(List<string> symbols)
```
mchandschuh: From the code it looks like this is an FXCM specific symbol, USD/JPY, is this correct? If so, maybe we need to provide a distinction in parameter naming to highlight that there is a difference between fxcm and qc symbols.

Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -0,0 +1,364 @@
...
+    {
+        private readonly IOrderProvider _orderProvider;
+        private readonly IHoldingsProvider _holdingsProvider;
+        private readonly string _server;
+        private readonly string _terminal;
+        private readonly string _userName;
+        private readonly string _password;
+
+        private Thread _orderEventThread;
+        private volatile bool _orderEventThreadStopped;
```
mchandschuh: Where appropriate I prefer to use [CancellationTokenSource](https://msdn.microsoft.com/en-us/library/system.threading.cancellationtokensource(v=vs.110).aspx) instead of explicit usage of `volatile` memory barrier semantics

Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -0,0 +1,364 @@
...
+                    {
+                        Thread.Sleep(1);
+                        continue;
+                    }
+
+                    OnOrderEvent(orderEvent);
+                }
+            });
+            _orderEventThread.Start();
+            while (!_orderEventThread.IsAlive) {}
```
mchandschuh: We may want a `Thread.Sleep(1)` in here to allow other threads a change to do something while we're spinning here. Is this expected to be less than a 1 ms wait here?

Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -0,0 +1,364 @@
...
+            Log.TraceFormat("FxcmBrokerage.GetAccountHoldings()");
+
+            var holdings = _openPositions.Values.Select(ConvertHolding).ToList();
+
+            // Set MarketPrice in each Holding
+            var symbols = holdings.Select(x => ConvertSymbolToFxcmSymbol(x.Symbol)).ToList();
+            if (symbols.Count > 0)
+            {
+                var quotes = GetQuotes(symbols).ToDictionary(x => x.getInstrument().getSymbol());
+                foreach (var holding in holdings)
```
mchandschuh: IB has a tendency to send back 0 quantity holdings, maybe we should check for something similar here as well to prevent subscribing to data we don't need

Logging/Log.cs
```diff
@@ -140,6 +140,25 @@ public static void Error(string format, params object[] args)
         }
 
         /// <summary>
+        /// Log trace with format string
+        /// </summary>
+        public static void TraceFormat(string traceFormat, params object[] args)
```
mchandschuh: I like this, but I wonder if we can just have an overload of `Log.Trace`:
```C#
public static void Trace(string message) { .. }
public static void Trace(string format, params object[] args) { ... }
```

I know R# will say overload hidden by params, but in this case it would be OK?

Thoughts?

Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            Log.TraceFormat("FxcmBrokerage.Connect()");
+
+            // create new thread to fire order events in queue
+            _orderEventThread = new Thread(() =>
+            {
+                while (!_orderEventThreadStopped)
+                {
+                    OrderEvent orderEvent;
+                    if (_orderEventQueue.IsEmpty || !_orderEventQueue.TryDequeue(out orderEvent))
```
mchandschuh: I think it's best to remove the check on `_orderEventQueue.IsEmpty` - due to the concurrent nature, the value returned from that isn't completely reliable, best to just `TryDequeue` and if not, take a nap

Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -0,0 +1,364 @@
...
+            if (!IsConnected)
+                throw new InvalidOperationException("FxcmBrokerage.PlaceOrder(): Unable to place order while not connected.");
+
+            if (order.Direction != OrderDirection.Buy && order.Direction != OrderDirection.Sell)
+                throw new ArgumentException("FxcmBrokerage.PlaceOrder(): Invalid Order Direction");
+
+            var symbol = ConvertSymbolToFxcmSymbol(order.Symbol);
+            var orderSide = order.Direction == OrderDirection.Buy ? SideFactory.BUY : SideFactory.SELL;
+            var quantity = (double)order.AbsoluteQuantity;
+            var accountId = _accounts.Keys.First();
```
mchandschuh: What are the various accounts for? Do they each hold different cash types (EUR, USD, GBP, ect..) and if so, can you place an order for any fx pair from any account?

Logging/Log.cs
mchandschuh: In addition, I would like to see us not have to prepend ClassName.MethodName(): everywhere we do logging, the logging infrastructure could easily resolve this, see `Log.Error` -- these changes are certainly not required as part of this PR, just mentioning for posterity.

Common/Brokerages/FxcmBrokerageModel.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
+                    "This model does not support " + order.Type + " order type."
+                    );
+
+                return false;
+            }
+
+            // validate order quantity
+            if (order.Quantity % 1000 != 0)
```
mchandschuh: Very thorough preorder checks, I love it!

Common/Brokerages/FxcmTransactionModel.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            // From http://www.fxcm.com/forex/forex-pricing/ (on Oct 6th, 2015)
+            // Forex: $0.04 per side per 1k lot for EURUSD, GBPUSD, USDJPY, USDCHF, AUDUSD, EURJPY, GBPJPY
+            //        $0.06 per side per 1k lot for other instruments
+
+            // From https://www.fxcm.com/uk/markets/cfds/frequently-asked-questions/
+            // CFD: no commissions
+
+            if (security.Type != SecurityType.Forex)
+                return 0m;
```
mchandschuh: no action, but curious, should we throw since it's not supported?

Brokerages/Fxcm/FxcmBrokerage.Messaging.cs
StefanoRaggi: Correct. I'll rename the parameter to fxcmSymbols and do the same in all other places.

Brokerages/Fxcm/FxcmBrokerage.cs
StefanoRaggi: Absolutely not. I totally forgot the Thread.Sleep. Maybe I was in need of sleep when I wrote that code :smile:

Brokerages/Fxcm/FxcmBrokerage.cs
StefanoRaggi: I never saw that happening, but I'll add the check anyway.

Logging/Log.cs
StefanoRaggi: I just noticed the params overload was added two weeks ago, in the Toolbox import. :)
I'll just remove TraceFormat and use the new overload.

Brokerages/Fxcm/FxcmBrokerage.cs
StefanoRaggi: Totally agree.

Brokerages/Fxcm/FxcmBrokerage.cs
StefanoRaggi: This one was still on my todo list. I emailed fxcm asking if I could have multiple accounts for testing (the default when you open a new account is to have 1 login with 1 account). They gave me a new login id with 3 accounts, so now I can test trading with a specific account.
I would add the account id as a new parameter in config.json.
The account currency is the same for all accounts under the same login.

Common/Brokerages/FxcmTransactionModel.cs
StefanoRaggi: I added the check so GetOrderFee is CFD-ready. 
Maybe better temporarily remove CFD from FxcmBrokerageModel.CanSubmitOrder ?

Brokerages/Fxcm/FxcmBrokerage.cs
StefanoRaggi: I understand Fxcm multiple accounts are used primarily for managed accounts: [PAMM](https://en.wikipedia.org/wiki/Percent_allocation_management_module) or LAMM.

mchandschuh:
I've pulled this code down and have begun merging and will address my previous comment. Thanks for the amazing contribution @StefanoRaggi!

mchandschuh:
This was merged in https://github.com/QuantConnect/Lean/commit/f506ad503673e9c93ae7494af38d1f49f8456fbc

PR #188, opened 2015-10-16
Implements AlgoSeek Options converter

AlexCatarino:
This is the first version. In next versions, we will look for:
* Optimization; 
* Make use of an extended version of [LeanDataWriter](https://github.com/QuantConnect/Lean/blob/master/ToolBox/LeanDataWriter.cs).


ToolBox/AlgoSeekOptionsConverter/Program.cs
```diff
@@ -0,0 +1,407 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Tick class
+    /// </summary>
+    class Tick
```
jaredbroad: We can't use this name since it will conflict with the Tick type in LEAN

Data/market-hours/market-hours-database.csv
```diff
@@ -10,4 +10,5 @@
 #
 America/New_York,usa,,equity,-,-,-,-,4,9.5,16,20,4,9.5,16,20,4,9.5,16,20,4,9.5,16,20,4,9.5,16,20,-,-,-,-
 # fxcm data uses EST, does not use EDT, so use a fixed offset
-UTC-05,fxcm,,forex,17,17,24,24,+,+,+,+,+,+,+,+,+,+,+,+,+,+,+,+,0,0,17,17,-,-,-,-
\ No newline at end of file
+UTC-05,fxcm,,forex,17,17,24,24,+,+,+,+,+,+,+,+,+,+,+,+,+,+,+,+,0,0,17,17,-,-,-,-
+America/Sao_Paulo,bra,,equity,-,-,-,-,9.5,10,16.9166667,18,9.5,10,16.9166667,18,9.5,10,16.9166667,18,9.5,10,16.9166667,18,9.5,10,16.9166667,18,-,-,-,-
```
jaredbroad: Hey Alex, I think this should be moved to another pull-request. Do the SP market hours really close at 16.55?

ToolBox/AlgoSeekOptionsConverter/Program.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        public override string ToString()
+        {
+            return this.Underlying + " " + this.OptionType + " " + this.Expiration + " " + this.Strike + " " + this.Time.ToString("yyyyMMdd") + " " + this.CsvFileOutput();
+        }
+    }
+    /// <summary>
+    /// TradeBar class
+    /// </summary>
+    class TradeBar
```
jaredbroad: Same here (can't use this name), I think we should figure out the new name/design for the option-tradebar/quotebar types and update this writer to create those types directly. Will discuss in the options-issue,

Data/market-hours/market-hours-database.csv
AlexCatarino: Yep. It closes at 16:55. Crazy.
How can I undo this and move to another pull-request?

ToolBox/AlgoSeekOptionsConverter/Program.cs
AlexCatarino: I will rename this class.
I can only use Tick type from LEAN when we implement quantity for bid and ask.

Data/market-hours/market-hours-database.csv
mchandschuh: The easiest way since it's only a single file is probably to take the copy from master and submit a new commit with the version from master. after committing, you can repush to this pull request

AlexCatarino: OK!

mchandschuh:
looks like you deleted the market hours database :)

No worries, will pull down locally and fix. Thanks for the contribution @AlexCatarino!!

AlexCatarino:
I am such a noobie! :8ball:

PR #189, opened 2015-10-16
Force market open2

desimonk:
Changed "open"/"close" to "true"/"false".


mchandschuh:
Thanks for this change! Super helperful for debugging on weekends!

PR #191, opened 2015-10-23
LSMA indicator and test added.

Jay-Jay-D:
Momersion small changes and fixes:
    - Name changed to MomersionIndicator
    - Conditional to avoid division by zero and estimations with few real observations in case of forward filled data added.
MomersionTest update IPython notebook test address.


mchandschuh:
I will pull this one in, but in the future, please keep a pull request limited to a single change.

Also, I think we want to name it `LeastSquaresMovingAverage`

Jay-Jay-D:
I agree, is the correct name

I just realized that In Spanish squared and squares are the same word!

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/a4f44ada545b253c76b9554514cebc977033bcfd

PR #194, opened 2015-10-24
Changed Padded Performance log Error to Trace

StefanoRaggi:
ERROR:: StatisticsBuilder.EnsureSameLength(): Padded Performance

This message is now more frequent, due to rolling statistics, and is not really an error.


mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/11c6753297d026888f31e1d6cc127003c73c8dd0

Issue #118, opened 2015-06-19
Add BidSize and AskSize to Tick


mchandschuh:
If this data is available it would be nice to have a place to put it.  Currently InteractiveBrokersBrokerage could have access to bid/ask size data.

PR #195, opened 2015-10-26
Add BidSize and AskSize to Tick

AlexCatarino:
Working on the Options Project, I made this little change.
If this is merged, it would close issue #118.

### DIFF
--- a/Common/Data/BaseData.cs
+++ b/Common/Data/BaseData.cs
@@ -29,6 +29,7 @@ namespace QuantConnect.Data
         private Symbol _symbol = Symbol.Empty;
         private decimal _value;
         private bool _isFillForward;
+        private BaseContract _contract = new BaseContract();
 
         /// <summary>
         /// Market Data Type of this data - does it come in individual price packets or is it grouped into OHLC.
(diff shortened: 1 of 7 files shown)


mchandschuh:
@AlexCatarino - I can pull in the addition of bid/ask size to the `Tick` class, but the other change I think needs more thought, it will affect everything in our system. If possible, continue on the options stuff without considering the IContract concept. We'll need to fully flesh out the contract system before coding it up.

mchandschuh:
The first commit (with edits) was merged in https://github.com/QuantConnect/Lean/commit/942d3fc7723bf5c61af391bb32e2a82a452462f2

AlexCatarino:
Ok, but we need to define how IContract will be.
I used this additions to rewrite the AlgoSeek Options Converter and the LeanDataWriter.

mchandschuh:
Understandable, you can keep working with what you have, but I'll need more time to implement it properly throughout the entire code base since it can have far reaching implications, this is something I don't want to rush.

If required, make some temporary classes that represent the data you need and where you need it and we can refactor it later when a full-fledged contract system is implemented in Lean.

PR #197, opened 2015-10-26
Implements Bovespa Converter

AlexCatarino:
Implements a converter for files provide by Brazilian Stock Exchange at is ftp and http site.

I deleted InstrumentType.cs because it is unnecessary after the Toolbox has merged moved into Lean, since it only implements a enumeration that is identical to SecurityType.


ToolBox/BovespaConverter/config.json
```diff
@@ -0,0 +1,28 @@
+﻿{
+    //----------------------------------------
+    // Configuration for the Bovespa Converter
+    //----------------------------------------
+
+    // Input folder in which zip raw files are saved
+    "input-folder": "C:\\Users\\Alexandre\\Documents\\Data",
+
+    // Output folder in which files will be saved (usually the Lean data folder)
+    "output-folder": "C:\\Users\\Alexandre\\Documents\\GitHub\\Lean\\Data",
```
jaredbroad: Please make this a relative folder path

AlexCatarino: Other config.json files in the Toolbox have:
"output-folder": "C:\\Lean\\Data",
Is that alright?
If so, I could put
"input-folder": "C:\\Data",

jaredbroad: The input folder can be anything right - is it just a cache for ftp download? So to make it work in anyone's computer we could change it to "./cache" with a Directory.CreateDirectory call if it doesn't exist.

The output folder could be relative to the toolbox bin directory - "../../Data" so it will save data into the default data directory for LEAN.

Very cool how you made factor files by the way. We should probably extract that code into its own function -- e.g. move it into the LEAN data writer and make it accept generic input to generate the factors.

AlexCatarino: No, the input folder is the place the user saved the downloaded data. Since Bovespa delete the files to keep only two-years data, I would advise users to keep their copies and save them at folders "Equity", "Option" and "Future" inside "Data". I need to write a help file to explain that!

Ok, I will make 
"output-folder": "..\..\Data"

Thanks for the advise on factor files generation. I will look into it in future versions.

ToolBox/BovespaConverter/config.json
```diff
@@ -0,0 +1,32 @@
...
+
+    // Input folder in which zip raw files are saved
+    "input-folder": "C:\\Data",
+    // Save files that were downloaded from Bovespa ftp site under:
+    // "input-folder"\\Equity for equities
+    // "input-folder"\\Option for options
+    // "input-folder"\\Future for futures
+    
+    // Output folder in which files will be saved (usually the Lean data folder)
+    "output-folder": "...\\...\\Data",
```
jaredbroad: + Forward slashes so it works cross platform =)

mchandschuh:
Please review the changes from this pull: https://github.com/QuantConnect/Lean/pull/198. The goal here was to create an abstraction that all downloaders would follow. Do you think this Bovespa app should implement the `IDataDownloader` interface?

AlexCatarino:
I don't think so, @mchandschuh. At least not for tick or bar data.
First, Bovespa deletes files older than 2 years, so the user should keep them on disk.
Second, unlike Yahoo, Google, etc..., its files have all the symbols for a given period, so we would download the data, pick only one symbol defined by `<IDataDownloader.Get(...)>`, then download the same data for the second symbol and so on. In my code, I open the Bovespa file and all the symbols at once.
I could use the Downloader for constructing the factor files, but I would rather work on the tick data, since I have trade and quote data that now can be dealt efficiently with the new QuoteBar class.

jaredbroad:
Hey @AlexCatarino, going forward all toolbox projects will need to implement the IDataDownloader interface. This is so we can seamlessly integrate them with the new visual interface (coming soon!) :) 

Your IDataDownloader can throw an exception for requests older than 2 years because of the Bovespa limitations. IDataDownloader allows resolution to be specified so the class must return the resolution requested (tick, sec, min etc) or throw if not supported.

The wrapping project/Visual Interface will check if the converted file already exists in the output directory before asking to download it since this is a common behavior. 

We're also going to make an "DataDownloaderHistoryProvider" which will allow history requests via an IDataDownloader interface!

The only exception will be "Converter" projects which convert from X->LEAN format. Converter projects must be Converters *only* when there's no option to download the data from an external source. Your implementation downloads AND converts. 

If you're up for it --> please remove the batch processing and make a class which implements the IDD interface, and then a program which instantiates the class for all your tickers to download the data required. Otherwise please change it to be a pure converter project (directory input) without a downloading component.

AlexCatarino:
Ok @jaredbroad. I got it.
You can close this pull and I will make a new revamped Bovespa Converter.

PR #202, opened 2015-10-29
Creates IBar interface and Bar class that implement OHLC object.

AlexCatarino:
* Includes IBar to TradeBar object. 
* Introduce QuoteBar object to deal with Bid and Ask bars.


Common/Data/Market/IBar.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Low price of the bar during the time period.
+        /// </summary>
+        decimal Low { get; set; }
+
+        /// <summary>
+        /// Closing price of the bar. Defined as the price at Start Time + TimeSpan.
+        /// </summary>
+        decimal Close { get; set; }
```
mchandschuh: do we require `IBar` to have public setters? If not, it may be best for the abstraction to be get-only

Common/Data/Market/QuoteBar.cs
```diff
@@ -0,0 +1,420 @@
...
+
+        /// <summary>
+        /// Ask OHLC
+        /// </summary>
+        public Bar Ask { get; set; }
+
+        /// <summary>
+        /// Opening price of the bar: Defined as the price at the start of the time period.
+        /// </summary>
+        public decimal Open { get; set; }
```
mchandschuh: I think these should be computed properties, for example:
```C#
        public decimal Open
        {
            get
            {
                if (Bid != null && Ask != null)
                {
                    return (Bid.Open + Ask.Open)/2m;
                }
                if (Bid != null)
                {
                    return Bid.Open;
                }
                if (Ask != null)
                {
                    return Ask.Open;
                }
            }
        }
```

Thoughts? This way if the underling `Bid` or `Ask` properties change (they are mutable) then we won't ever get in an inconsistent state.

Common/Data/Market/QuoteBar.cs
```diff
@@ -0,0 +1,420 @@
...
+
+
+        /// <summary>
+        /// Update the quotebar - build the bar from this pricing information:
+        /// </summary>
+        /// <param name="lastTrade">This trade price</param>
+        /// <param name="bidPrice">Current bid price (not used) </param>
+        /// <param name="askPrice">Current asking price (not used) </param>
+        /// <param name="volume">Volume of this trade</param>
+        public override void Update(decimal lastTrade, decimal bidPrice, decimal askPrice, decimal volume)
```
mchandschuh: I think it may be best to use the other Update method, this would require changing the signature of the `BaseData.Update` method to accept the bid/ask sizes

Common/Data/Market/QuoteBar.cs
```diff
@@ -0,0 +1,420 @@
...
+                        return ParseForex<QuoteBar>(config, line, date);
+                }
+            }
+            catch (Exception err)
+            {
+                Log.Error("DataModels: QuoteBar(): Error Initializing - " + config.SecurityType + " - " + err.Message + " - " + line);
+            }
+
+            // if we couldn't parse it above return a default instance
+            return new QuoteBar { Symbol = config.Symbol, Period = config.Increment };
```
mchandschuh: Typical behavior is to return null when we were unable to parse, why return this new instance?

Common/Data/Market/QuoteBar.cs
AlexCatarino: Agree. I was thinking about that when I was writing the constructors.
That way, in the constructors I would need to define Bid and Ask first, the calculate the Mid, right?

Common/Data/Market/IBar.cs
AlexCatarino: We have setters on OHLC at TradeBar, so I didn't gave ant thoughts on that requirement.

mchandschuh: I would prefer we mark `IBar` properties as get-only. `TradeBar` can still expose its setters

Common/Data/Market/QuoteBar.cs
mchandschuh: We shouldn't need setters for the Open,High,Low,Close properties. The Bid and Ask can be settable.

Common/Global.cs
```diff
@@ -276,6 +276,8 @@ public enum MarketDataType
     {
         /// Base market data type
         Base,
+        /// QuoteBar market data type [Bid(OHLC), Ask(OHLC) and Mid(OHLC) summary bar]
+        QuoteBar,
```
mchandschuh: I often find it safer to add new enumeration items to the end, otherwise we're changing existing values, `MarketDataType.TradeBar` was 1, now it's 2

Common/Data/Market/QuoteBar.cs
```diff
@@ -0,0 +1,495 @@
...
+        /// <param name="bidlow">Decimal Low Price of this bid bar</param>
+        /// <param name="bidclose">Decimal Close price of this bid bar</param>
+        /// <param name="avgbidsize">Average bid size over period</param>
+        /// <param name="askopen">Decimal Opening Price of Ask</param>
+        /// <param name="askhigh">Decimal High Price of this ask bar</param>
+        /// <param name="asklow">Decimal Low Price of this ask bar</param>
+        /// <param name="askclose">Decimal Close price of this ask bar</param>
+        /// <param name="avgasksize">Average ask size over period</param>
+        /// <param name="period">The period of this bar, specify null for default of 1 minute</param>
+        public QuoteBar(DateTime time, Symbol symbol, decimal bidopen, decimal bidhigh, decimal bidlow, decimal bidclose, long avgbidsize, decimal askopen, decimal askhigh, decimal asklow, decimal askclose, long avgasksize, TimeSpan? period = null)
```
mchandschuh: These constructors have lots of parameters, maybe they should accept bid/ask as `IBar` objects?

mchandschuh:
Hey @AlexCatarino - great stuff! Just a few comments, let me know what you think! Thanks!

Common/Global.cs
AlexCatarino: Noobie stuff. : ) 
I will fix that.

Common/Data/Market/QuoteBar.cs
AlexCatarino: Well, yes. 
If I read an csv file with quote bars, I can make Bid/Ask `<IBar>` before I call the constructor...

AlexCatarino:
Your comments are welcome. I think that the QuoteBar object is far from being final, but it has the necessary structure for using it on the data converters.

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/a1837105a42ffb1f0cf404ccf0c485f29f6022fb

Thanks @AlexCatarino! More awesome stuff!

PR #206, opened 2015-11-05
Implements Bovespa Downloader

AlexCatarino:
The Bovespa Downloader downloads the files from the Bovespa FTP site or from my Dropbox (daily files) to user local file and process the data.


ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,449 @@
...
+                throw new ArgumentException("Invalid symbol requested: " + symbol);
+
+            if (type != SecurityType.Equity && type != SecurityType.Option && type != SecurityType.Future)
+                throw new NotSupportedException("SecurityType not available: " + type);
+
+            if (endUtc < startUtc)
+                throw new ArgumentException("The end date must be greater or equal to the start date.");
+
+            _inputDirectory = Directory.CreateDirectory(String.Format("{0}/{1}", Config.Get("input-directory", "./Data"), type)).FullName;
+            _ftpsite += (new Dictionary<SecurityType, string>() { { SecurityType.Future, "BMF" }, { SecurityType.Equity, "Bovespa-Vista" }, { SecurityType.Option, "Bovespa-Opcoes" } }[type]);
```
mchandschuh: Could we define this dictionary as a `static readonly` class variable? It looks like we're building the ftp site string, but the inline dictionary instantiation makes the line look a little confusing.

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,449 @@
...
+
+            #region For equity daily files, use equity daily data!
+            if (type == SecurityType.Equity && resolution == Resolution.Daily)
+            {
+                for (var year = startUtc.Year; year <= endUtc.Year; year++)
```
mchandschuh: After we complete this equity daily loop, is it intended to continue to the ListTickDataFiles method?

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,449 @@
...
+            if (!File.Exists(localfile))
+            {
+                Console.WriteLine("Do not have nor could download COTAHIST_A{0}.zip", year);
+            }
+            else
+            {
+                ZipFile zip;
+                using (var reader = Compression.Unzip(localfile, out zip))
+                {
+                    var bars = reader.ReadToEnd().Split(Environment.NewLine.ToCharArray(), StringSplitOptions.RemoveEmptyEntries)
```
mchandschuh: I'm always weary of calling `string.Split` on potentially large strings. [Here's](http://www.codeproject.com/Articles/176031/Out-of-Memory-Exception-A-simple-string-Split-can) a great, in-depth look into why.  If the string isn't that large then no need to change, more just for information :smile:

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,449 @@
...
+                Console.WriteLine("Do not have nor could download " + file);
+            }
+            else
+            {
+                ZipFile zip;
+                var isBid = localfile.Contains("OFER_CPA");
+                var isAsk = localfile.Contains("OFER_VDA");
+                var tickType = localfile.Contains("NEG") ? TickType.Trade : TickType.Quote;
+
+                using (var toplevelreader = Compression.Unzip(localfile, out zip))
```
mchandschuh: I think when we use this method we need to manually dispose of the `out zip` parameter.

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,449 @@
...
+        /// <summary>
+        /// Aggregates a list of ticks at the requested resolution
+        /// </summary>
+        /// <param name="symbol">Symbol</param>
+        /// <param name="ticks">Input tick data</param>
+        /// <param name="resolution">Output resolution</param>
+        /// <returns>Enumerable of trade bars or quote bars</returns>
+        private static IEnumerable<BaseData> AggregateTicks(Symbol symbol, IEnumerable<BaseData> data, Resolution resolution)
+        {
+            var interval = new Dictionary<Resolution, TimeSpan>()
```
mchandschuh: The resolution has a nifty [extension method](https://github.com/QuantConnect/Lean/blob/master/Common/Extensions.cs#L518) defined, `ToTimeSpan` that will do this for you :smile:

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,449 @@
...
+
+            var ticks = data.Cast<Tick>().OrderBy(t => t.Time).ToList();
+
+            if (ticks.All(t => t.TickType == TickType.Trade))
+            {
+                return
+                    (from t in ticks
+                     group t by t.Time.RoundDown(interval)
+                         into g
+                         select new TradeBar
```
mchandschuh: love the LINQ-iness!

mchandschuh:
Absolutely epic work @AlexCatarino! I have some minor comments, let me know what you think. Great code! Thanks!

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
AlexCatarino: Yes! If one doesn't have the daily data (we need to download it manually at the Bovespa site, because of a captcha), daily Lean data can be built from the tick data.

ToolBox/BovespaDownloader/BovespaDataDownloader.cs
```diff
@@ -0,0 +1,456 @@
...
+                         //    g.Where(t => t.BidSize > 0).Last().LastPrice
+                         //),
+                         //Ask = new Bar
+                         //(
+                         //    g.Where(t => t.AskSize > 0).First().LastPrice,
+                         //    g.Where(t => t.AskSize > 0).Max(t => t.LastPrice),
+                         //    g.Where(t => t.AskSize > 0).Min(t => t.LastPrice),
+                         //    g.Where(t => t.AskSize > 0).Last().LastPrice
+                         //),         
+                     });
```
AlexCatarino: Hey @mchandschuh, I don't know how to do this properly (so I commented out).
I have ticks with bid data (`<BidSize > 0 && AskSize = 0`>) and ask data (`<BidSize = 0 && AskSize > 0`>). I want to group then together by `<Time.RoundDown(...)`> then I need to group them by bid-only and ask-only list to make the calculations.
I don't think my solution is right, because the `<Where(...)`> can return a zero-length list and that would result into an exception when I try to access `<First()`> and`<Last()`>. And, of course, the several `<Where(...)`> calls are redundant and probably inefficient.

mchandschuh: Instead of using `First()` and `Last()` directly, you could use `FirstOrDefault()` and `LastOrDefault()`, or you could even do something like this:
```C#
Ask = new Bar
(
   g.Where(t => t.AskSize > 0).Select(t => t.LastPrice).FirstOrDefault),
   ...
)
```
This will produce zeros for the enumerables that are empty and avoid the exception.

Does this help?

AlexCatarino: I guess.
Would that be optimal? Or do you think there is a better way?
I will use this code snipped in the AlgoSeekOptionsConverter that I am about to revisit.

mchandschuh: If by optimal we're talking about performance, then not entirely. LINQ is a great language feature but does have some overhead given all the delegates/method calls. The `QuoteBar` class has an `Update` [method](https://github.com/QuantConnect/Lean/blob/master/Common/Data/Market/QuoteBar.cs#L223) that could be called in a loop. This would probably be the most efficient, straightforward way.
```C#
var resolutionIncrement = resolution.ToTimeSpan();
var grouped = ticks.GroupBy(t => t.Time.RoundDown(resolutionIncrement);
foreach (var g in grouped)
{
    // maybe we can make another constructor that just accepts time,symbol and resolution and
    // calls this other constructor with these values. this would make the aggregation case easier
    var quoteBar = new QuoteBar(t.Time, t.Symbol, null, 0, null, 0, resolutionIncrement);
    foreach (var tick in g)
    {
        quoteBar.Update(0, tick.BidPrice, tick.AskPrice, 0, tick.BidSize, tick.AskSize);
    }
    yield return quoteBar;
}
```

I just wrote this up in this window, so it may not compile :/ apologies, but I think it expresses the intent. The performance issue with using LINQ to do this aggregation is that we end up looping about 6x more then we need to, as each one of the `Max()`, `Min()`, and `Last()` calls will loop over the entire grouping.

I try to use the `Update` methods when possible since we can write a test against it, and if everyone uses it then we know everyone is aggregating with the same rules.

PR #209, opened 2015-11-12
Enables AvgBidSize and AvgAskSize calculation in QuoteBar Update method

AlexCatarino:


Common/Data/Market/QuoteBar.cs
```diff
@@ -226,9 +238,18 @@ public override void Update(decimal lastTrade, decimal bidPrice, decimal askPric
...
-            AvgBidSize += Convert.ToInt32(bidSize);
-            AvgAskSize += Convert.ToInt32(askSize);
+            if ((_updateBidCount == 0 && AvgBidSize > 0) || (_updateAskCount == 0 && AvgAskSize > 0))
+            {
+                throw new InvalidOperationException("Average bid/ask size cannot be greater then zero if update counter is zero.");
+            }
+            
+            if (bidSize > 0) _updateBidCount++;
+            _sumBidSize += Convert.ToInt32(bidSize);
+            AvgBidSize = _sumBidSize / _updateBidCount;
```
mchandschuh: This looks like it would throw an exception if the first update is ask only data. Should it be like this:
```C#
if (bidSize > 0) 
{
    _updateBidCount++;
    _sumBidSize += Convert.ToInt32(bidSize);
    AvgBidSize = _sumBidSize / _updateBidCount;
}
```

Likewise for the avg ask size calculation.

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/5176fd95ae730bee02a27db15fa2c8fcb5249c2b

PR #210, opened 2015-11-13
FXCM Brokerage - improved disconnection/reconnection logic

StefanoRaggi:


Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -128,6 +135,92 @@ public override void Connect()
...
+                        }
+                        else if (_connectionLost && elapsed <= TimeSpan.FromSeconds(5))
+                        {
+                            try
+                            {
+                                _gateway.relogin();
+                            }
+                            catch (Exception exception)
+                            {
+                                Log.Error(exception);
```
mchandschuh: If we throw an error during the call to `relogin` it looks like we'll swallow the exception and still set `_connectionLost=false` below. Maybe we could set `_connectionLost` to `false` inside the try after `_gateway.relogin()` and set it to `false` in the `catch` block

StefanoRaggi: You're right, although when it enters that if block, connection has been restored and I have never seen relogin() fail. I'll move the two rows inside the try block.

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/e2f9c527a63ec5becae594091fd265a694efd29d

PR #211, opened 2015-11-14
LeanDataWriter now preserves existing hourly/daily data

StefanoRaggi:


jaredbroad:
:D awesome love it.

ToolBox/LeanDataWriter.cs
```diff
@@ -104,28 +127,105 @@ public void Write(IEnumerable<BaseData> source)
...
+            var newRows = new SortedDictionary<DateTime, string>(source.ToDictionary(x => x.Time, GenerateFileLine));
+            SortedDictionary<DateTime, string> rows;
+
+            if (File.Exists(outputFile))
+            {
+                // If file exists, we load existing data and perform merge
+                rows = LoadHourlyOrDailyFile(outputFile);
+                foreach (var kvp in newRows)
+                {
+                    rows[kvp.Key] = kvp.Value;
```
mchandschuh: I think we want to load the hourly/daily file first, and then overly the newly created data. My thought is this, if the date in the old file is corrupt then it will never get replaced.

ToolBox/LeanDataWriter.cs
```diff
@@ -104,28 +127,105 @@ public void Write(IEnumerable<BaseData> source)
...
+            }
+
+            return rows;
+        }
+
+        /// <summary>
         /// Write this file to disk
         /// </summary>
-        private void WriteFile(string data, string filename, DateTime time)
+        private void WriteFile(string fileName, string data, DateTime time)
```
mchandschuh: thank you for swapping these to be correct :)

ToolBox/LeanDataWriter.cs
StefanoRaggi: I believe the code is already doing this: load the existing file first, then overwriting the dictionary entries with same date or adding new ones. 
If the existing file contains invalid timestamps, the load function would give an error because the key is a DateTime

mchandschuh: Ahh indeed, yes, you are correct. I read this loop incorrectly. Great stuff! Just checking to see if you're paying attention :smiley:

StefanoRaggi: :smile:

mchandschuh:
Yet another epic pull, thanks @StefanoRaggi!

Merged in https://github.com/QuantConnect/Lean/commit/e3ee93fc10a3dd5ee8275530c04df19c9164fb48

PR #215, opened 2015-11-21
FXCM Downloader

StefanoRaggi:


ToolBox/FxcmDownloader/FxcmDataDownloader.cs
```diff
@@ -0,0 +1,422 @@
...
+        /// <param name="startUtc">Start time of the data in UTC</param>
+        /// <param name="endUtc">End time of the data in UTC</param>
+        /// <returns>Enumerable of base data for this symbol</returns>
+        public IEnumerable<BaseData> Get(Symbol symbol, SecurityType type, Resolution resolution, DateTime startUtc, DateTime endUtc)
+        {
+            if (!_instruments.ContainsKey(symbol))
+                throw new ArgumentException("Invalid symbol requested: " + symbol);
+
+            if (resolution == Resolution.Tick)
+                throw new NotSupportedException("Resolution not available: " + resolution);
```
mchandschuh: Does their API not support tick data?

jaredbroad: FXCM supports tick, but Oanda doesn't. This might be copy-pasta from Oanda's implementation

StefanoRaggi: Merged in f93600809aa7a8a42269e574151c7d6cbc573ba4

ToolBox/FxcmDownloader/FxcmDataDownloader.cs
```diff
@@ -0,0 +1,422 @@
...
+            cal.setTime(javaDate);
+
+            // note that the Month component of java.util.Date  
+            // from 0-11 (i.e. Jan == 0)
+            return new DateTime(cal.get(Calendar.YEAR),
+                                cal.get(Calendar.MONTH) + 1,
+                                cal.get(Calendar.DAY_OF_MONTH),
+                                cal.get(Calendar.HOUR_OF_DAY),
+                                cal.get(Calendar.MINUTE),
+                                cal.get(Calendar.SECOND));
```
mchandschuh: @StefanoRaggi  - Maybe we need to use something like the following here:

```C#
return new DateTime(cal.get(Calendar.YEAR),
                    cal.get(Calendar.MONTH) + 1,
                    cal.get(Calendar.DAY_OF_MONTH),
                    cal.get(Calendar.HOUR_OF_DAY),
                    cal.get(Calendar.MINUTE),
                    cal.get(Calendar.SECOND))
                    .AddMilliseconds(cal.get(Calendar.MILLISECOND));
```

A similar change should be applied to ToJavaDataUtc

StefanoRaggi: @mchandschuh Thanks, note taken.

Another one here (to be moved and made public): https://github.com/QuantConnect/Lean/blob/master/Brokerages/Fxcm/FxcmBrokerage.Util.cs#L217

nsampietro:
Was this resolved?.  It's pending merge into master?.  Thanks.

StefanoRaggi:
@nsampietro I'm not sure which issue you are referring to. 
The downloader has been merged into master (although Tick data is not yet supported).

nsampietro:
Yes I was referring to Ticks data. Thanks for the answer Stefano. I made an attempt to solve this. It's on [this] (https://github.com/nsampietro/Lean/tree/master/ToolBox/FxcmDownloader)  commit. Feedback is welcome.

StefanoRaggi:
@nsampietro Thanks for the attempt to add Tick data support (it has been on our todo list for some time).

I had a look at the code and I think there are a few issues with FXCM requests that you may not be aware of:
1. There is a limit on the number of rows returned (600 for ticks, 300 for other resolutions)
2. When you request rows between two dates (or better say timestamps), if the number of rows is greater than the maximum allowed, you are returned the **last** ones (not the first ones).

So for Second resolution and above, I solved the problem by always requesting a range that will always have **exactly** 300 bars (or less if data is not available).

For Tick resolution it gets tricky: you cannot use the same "trick" because we don't know how many ticks there can be in a given period of time. Requests for tick data should go backwards instead of forward, setting the end time of the next request equal to the first time of the previous request minus one millisecond. Rows returned would then have to be inserted at the top of the result list instead of appending.

Please let me know if you have other questions on this.

nsampietro:
Aha!. So, what I need to do is to go backwards on the "End" date (and keeping the "Start" date fixed of course). I made the changes.

StefanoRaggi:
@nsampietro Good job! Please submit a Pull Request so we can do a review before merging.

Issue #190, opened 2015-10-23
Genetic optimization


chrisdk2015:
Created this issue for tracking the genetic optimization

https://github.com/chrisdk2015/LeanOptimization

It uses the GAF framework and is currently for running on the command line.

Howto parallize this would be nice to know. Maybe one needs to change the GAF framework.

bizcad:
@chrisdk2015 

I got the genetic algorithm running without blurring by on the console.   However, after 1723 trials, it blows up my Visual Studio.  I mean VS just quits with no notice or anything.  It has happened twice, once with console blur and once without.  It seems to happen in the same place but I cannot catch an Exception.  

I was wondering how many trials were needed to resolve the genetic algorithm.  And also did your crash like that.  Of course, if you were waiting 40 seconds between trials, it would have taken all night.  My current version did the 1723 trials in 1:09:22 or 2-3 seconds per trial and only shows the 1 line per trial in the console.  

I copied ConsoleResultsHandler and modified it into OptimizationResutsHandler.  I also commented out line 62 in JobQueue because it was Log.Trace("dodah") before the Log.Handler was set to a FileLogHandler.  

Here is a link to a dropbox folder with the log.txt and log.csv files for the 1723 trials.
https://www.dropbox.com/sh/zm19bgmuo1t5os0/AACt4pHp6zbE1KNmWxnhlO7wa?dl=0

I am pushing the complete project to https://github.com/bizcad/LeanOpto

On line 89 of Program, I set
<pre>
            Log.LogHandler = (ILogHandler)new FileLogHandler();
            Log.DebuggingEnabled = true;
            Log.DebuggingLevel = 2;
</pre>
so that Log output is only sent to a file.  Then I replace all the Console.WriteLines (except for 2) to Log.Trace, Debug or Error. 

I also commented the parts that refer to writing chart files.  We do not need them.  

Nick

chrisdk2015:
@bizcad 

I think there is a memory leak somewhere.

At least here when I run start running it the memory usage keeps going up. 

Maybe you can put the memory usage in the log file to see if that is the reason.

I tried to do a profiling here with mono but I just get segfaults so it is of no help. Maybe you could profile it in VS? I think the profile there is better.

With the OptimizationResultHandler it goes a bit faster.

For the termination criteria you could use the following

```
private static double oldFitness = Double.NaN;
				
public static bool Terminate(Population population, 
			int currentGeneration, long currentEvaluation)
		{
[...]

PR #216, opened 2015-11-25
Engine moved to a Class Library.

naicigam:
Now the Engine can be used from an custom VS project without having to open the entire solution and adding a new project.


Launcher/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Launcher
+{
+    class Program
+    {
+        /// <summary>
+        /// Primary Analysis Thread:
+        /// </summary>
+        public static void Main(string[] args)
+        {
+            QuantConnect.Lean.Engine.Engine.Launch();
```
jaredbroad: The Engine should take config and run as an instance - I'd recommend moving the entire of the Main() method into the Launcher Main(). The call to engine should look something like this:

```
//Get the configuration
var config = ...;
//Launch the engine
var engine = new Engine(config);
engine.Run();
```

naicigam: Great! I will do it that way.

naicigam: @jaredbroad Just started to make this change and I noticed that Config is static, so what would be the point of passing it to the Engine? In that case, it seems OK for the Engine to retrieve the values in the Constructor. Something like:

```
namespace Launcher
{
    class Program
    {
        public static void Main(string[] args)
        {
                var engine = new Engine();
                engine.Run();
        }
}
```

The other option is to get the configuration values in the Laucher.Main and passing LeanEngineSystemHandlers, LeanEngineAlgorithmHandlers and LiveMode to the Engine.
The problem I see with that is that the initialization code will have to be replicated in the application that uses the Engine DLL.

The third option involves many code changes: making the Config class non-static. But we will have to pass the Config object around to all the other classes that use it (ex. LiveTradingDataFeed).

What do you think?

jaredbroad: Engine should take its configuration items in its constructor. This is what I meant by the "var config ...." line. Its actually the LeanEngineSystemHandlers etc. Basically everything in the Main() method can be moved into the Launcher (i.e. not duplicated, but moved from the Engine.Main()).

bizcad:
@naicigam I like the idea of running Lean as a dll very much.  

In regards to pushing the Config into the dll, you might want to take a look at <a href="https://github.com/QuantConnect/Lean/issues/190"> issue #190 Genetic Optimization</a>.  @chrisdk2015 has a reference to running Lean from another project.  There he is passing parameters into the Config and setting LeanEnginexxxxxxHandlers.  He then runs Lean in a console with something like...

<code>
public decimal Run(int val)
        {
            Config.Set("EMA_VAR1", val.ToString(CultureInfo.InvariantCulture));
            LaunchLean();
            // Handle results
        }
</code>
and in the LaunchLean method...
<code>
        private void LaunchLean()
        {
            Config.Set("environment", "backtesting");
            string algorithm = "EMATest";  // tell Lean what algorithm to run.

            Config.Set("algorithm-type-name", algorithm);
            //string datapath = Config.Get("data-folder");
            _jobQueue = new JobQueue();
            _notify = new Messaging();
            _api = new Api();
            _resultshandler = new OptimizationResultHandler();
            //_resultshandler = new ConsoleResultHandler();
            _dataFeed = new FileSystemDataFeed();
            _setup = new ConsoleSetupHandler();
            _realTime = new BacktestingRealTimeHandler();
            _historyProvider = new SubscriptionDataReaderHistoryProvider();
            _transactions = new BacktestingTransactionHandler();
            Log.LogHandler = (ILogHandler)new FileLogHandler();
            Log.DebuggingEnabled = false;
            Log.DebuggingLevel = 1;

            var systemHandlers = new LeanEngineSystemHandlers(_jobQueue, _api, _notify);
            systemHandlers.Initialize();

            var algorithmHandlers = new LeanEngineAlgorithmHandlers(_resultshandler, _setup, _dataFeed, _transactions, _realTime, _historyProvider);
            string algorithmPath;
            
            AlgorithmNodePacket job = systemHandlers.JobQueue.NextJob(out algorithmPath);
            try
            {
                var _engine = new Engine(systemHandlers, algorithmHandlers, Config.GetBool("live-mode"));
                _engine.Run(job, algorithmPath);
            }
            finally
            {
                //Delete the message from the job queue:
                //systemHandlers.JobQueue.AcknowledgeJob(job);
                Log.Trace("Engine.Main(): Packet removed from queue: " + job.AlgorithmId);

                // clean up resources
                systemHandlers.Dispose();
                algorithmHandlers.Dispose();
                Log.LogHandler.Dispose();
            }

        }
</code>

You can find my version of the Optimization project in a <a href="https://github.com/bizcad/LeanOpto/blob/master/Optimization/Program.cs"> repo  on 
github.  </a>

On my local machine I have been using a version of his code to run various optimizations and have replaced ConsoleResultHandler with a custom OptimizationResultHandler.  I also have been running multiple algorithms and for various date ranges from a list in Program.cs like this...
<code>
        private static double RunAlgorithm(List<string> algos, Dictionary<string, DateRange> daysDictionary)
        {

            var sum_sharpe = 0.0;
            foreach (string s in algos)
            {
                foreach (string key in daysDictionary.Keys)
                {
                    var val = s;
                    var startDate = daysDictionary[key].startDate;
                    var endDate = daysDictionary[key].endDate;
                    AppDomain ad = null;
                    RunClass rc = CreateRunClassInAppDomain(ref ad);
                    Console.WriteLine("Running algorithm {0} for: {1} to {2}", val, startDate, endDate);

                    try
                    {
                        sum_sharpe += (double)rc.Run(val, startDate, endDate);
                    }
                    catch (Exception e)
                    {
                        Log.Error(e.Message + e.StackTrace);
                    }
                    AppDomain.Unload(ad);
          }
</code>
and 
<code>
        public decimal Run(string val, string startDate, string endDate)
        {
            // Set dates in Config.
            Config.Set("start-date", startDate);
            Config.Set("end-date", endDate);

            LaunchLean(val);   // pass in the algo name
...
       }
</code>
Then in the Initialize() of the algorithm...
<code>
            //Initialize dates
            sd = Config.Get("start-date");
            ed = Config.Get("end-date");

            _startDate = new DateTime(Convert.ToInt32(sd.Substring(0, 4)), Convert.ToInt32(sd.Substring(4, 2)), Convert.ToInt32(sd.Substring(6, 2)));
            _endDate = new DateTime(Convert.ToInt32(ed.Substring(0, 4)), Convert.ToInt32(ed.Substring(4, 2)), Convert.ToInt32(ed.Substring(6, 2)));


            SetStartDate(_startDate);
            SetEndDate(_endDate);
            SetCash(_portfolioAmount);
</code>

This code might give you some ideas on handling passing variables into the dll by making entries into Config.  

I hope this helps.  I am very much looking forward to your getting the Engine running as a dll.

Nick Stein aka bizcad

mchandschuh:
Extracting the `Config` static class can be handled in another issue. I've added a QuantConnect.Lean.Launcher csproj that will now act as the start up project for the engine in console mode. The QuantConnect.Lean.Engine has been converted into a dll project.

Here's the [commit](https://github.com/QuantConnect/Lean/commit/681cd32e328a621f7c73e0bababa40330185207f).

PR #218, opened 2015-11-25
FxcmBrokerage.ConvertSymbol exception with CFD symbols

StefanoRaggi:
The Fxcm brokerage implementation during logon receives all symbols (Forex and CFD) so this bug caused a timeout in the Connect method.

### DIFF
--- a/Brokerages/Fxcm/FxcmBrokerage.Util.cs
+++ b/Brokerages/Fxcm/FxcmBrokerage.Util.cs
@@ -126,9 +126,13 @@ namespace QuantConnect.Brokerages.Fxcm
         {
             var symbol = instrument.getSymbol().Replace("/", "");
             var securityType = GetSecurityType(instrument);
-            if (securityType == SecurityType.Forex)
+            switch (securityType)
             {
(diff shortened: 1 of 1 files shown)


mchandschuh:
Thanks for the fix! I've also added a helper method to reduce the switches on security types when creating `Symbol` instances. Have a peek at [this](https://github.com/QuantConnect/Lean/blob/master/Common/Symbol.cs#L44)

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/5e649baf70b3e744485ae3cf1e171f37d8154a50

StefanoRaggi:
@mchandschuh I used the Symbol.Create() method extensively in the new ISymbolMapper branch. I'll clean up brokerages and downloaders there.

mchandschuh:
@StefanoRaggi - you're always on top of it all! Awesome stuff. Yes, I think the Symbol.Create can cover most security types, may need overloads for creating options/futures due to needing to know extra info (expiry, strike, p/c,ect...)

PR #220, opened 2015-12-02
Symbol Mappers

StefanoRaggi:
- New ISymbolMapper interface
- FXCM and Oanda implementations
- Added SymbolMappers to FXCM and Oanda brokerages
- Added SymbolMappers to FXCM and Oanda downloaders
- Removed SecurityType parameter from IDataDownloader.Get() method
- Added Milliseconds to conversion functions between Java Date and .NET DateTime


Brokerages/Fxcm/FxcmBrokerage.Util.cs
```diff
@@ -224,7 +184,8 @@ private static DateTime FromJavaDate(java.util.Date javaDate)
                                 cal.get(java.util.Calendar.DAY_OF_MONTH),
                                 cal.get(java.util.Calendar.HOUR_OF_DAY),
                                 cal.get(java.util.Calendar.MINUTE),
-                                cal.get(java.util.Calendar.SECOND));
+                                cal.get(java.util.Calendar.SECOND),
+                                cal.get(java.util.Calendar.MILLISECOND));
```
mchandschuh: Please keep PRs to one issue. No worries on this pull, but this bug fix should have been in a separate PR.

Brokerages/ISymbolMapper.cs
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+namespace QuantConnect.Brokerages
+{
+    /// <summary>
+    /// Provides the mapping between Lean symbols and brokerage specific symbols.
+    /// </summary>
+    public interface ISymbolMapper
```
mchandschuh: I noticed there was no implementation for InteractiveBrokers. Any particular reason?

See `InteractiveBrokersBrokerage.MapSymbol` function

Brokerages/ISymbolMapper.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="leanSymbol">The Lean symbol</param>
+        /// <returns>The security type</returns>
+        SecurityType GetLeanSecurityType(string leanSymbol);
+
+        /// <summary>
+        /// Checks if the symbol is supported by the brokerage
+        /// </summary>
+        /// <param name="brokerageSymbol">The brokerage symbol</param>
+        /// <returns>True if the brokerage supports the symbol</returns>
+        bool IsKnownBrokerageSymbol(string brokerageSymbol);
```
mchandschuh: Does this assume that the symbol mapper implementations have a complete listing of all symbols? This seems unrealistic, especially for `InteractiveBrokersBrokerage` which will need to support all security types and even different markets.  Further I think it assumes that symbols are unique within a brokerage, but I fear this is not the case. Consider the equity symbol `AAPL` which can trade in various markets (USA, Mexico, ect..).

It appears that these functions are only used in downloaders and tests. Perhaps they can be concrete implementations on `FxcmSymbolMapper` and `OandaSymbolMapper` but not on `ISymbolMapper`.

Brokerages/ISymbolMapper.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Checks if the symbol is supported by the brokerage
+        /// </summary>
+        /// <param name="symbol">The Lean symbol</param>
+        /// <returns>True if the brokerage supports the symbol</returns>
+        bool IsKnownLeanSymbol(Symbol symbol);
```
mchandschuh: See comments for `IsKnownBrokerageSymbol`.  I would prefer that the `ISymbolMapper` just performs mapping, but not existence checks. Also, looking at the implementations, this appears to be more of a 'convert to brokerage symbol and see if it is known to the mapper' but not actually known to LEAN.

Brokerages/ISymbolMapper.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="symbol">A Lean symbol instance</param>
+        /// <returns>The brokerage symbol</returns>
+        string GetBrokerageSymbol(Symbol symbol);
+
+        /// <summary>
+        /// Converts a brokerage symbol to a Lean symbol instance
+        /// </summary>
+        /// <param name="brokerageSymbol">The brokerage symbol</param>
+        /// <returns>A new Lean Symbol instance</returns>
+        Symbol GetLeanSymbol(string brokerageSymbol);
```
mchandschuh: I think that this method may also need security type and market. I'm considering the case of IB, who supports trading just about anything. In this case,a single string is not enough to distinguish between various securities. Have a peek at the `Krs.Ats.IBNet.Contract` type. This is all the information that they use to define a single security.

PR #222, opened 2015-12-07
Changed BrokerageId from int to long

StefanoRaggi:
IOrderProvider.GetOrderByBrokerageId now accepts a long value (as required by Oanda and Tradier brokerages)


mchandschuh:
I think we wanted to change this value to a string to be more generic. There's an issue on Trello for it.

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/4558d085c6dbe24ec7f8ab00314b128f7dad2d46

PR #225, opened 2015-12-14
Oanda DataQueueHandler streams no data with more than one Subscribe request

StefanoRaggi:
Oanda only supports a maximum of two streaming rates sessions per token so we are reusing the same session for all subscriptions. In the case of multiple symbols to subscribe, on subsequent requests we have to close the previous session and start a new one, so requests must be synchronous.


mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/f2ecdf496b8aa0b4256c5e76970c76d977e68b49

PR #230, opened 2015-12-19
Fixed OnBalanceVolume indicator always returning zero

StefanoRaggi:


jaredbroad:
Looks good thank you @StefanoRaggi

PR #238, opened 2015-12-28
Fixed internal feed passed to algorithm.OnData

StefanoRaggi:


mchandschuh:
Didn't see this fix before I went and pushed a fix :)

Have a look at this commit: https://github.com/QuantConnect/Lean/commit/589a84224d0c435a55946a73c9c81791ea397ff0

When we add a security via AddSecurity, two important things need to happen. We add it to the user defined universe and we also add it to the Securities collection. This issue here was that the 'SPY' security did not exist in the Securities collection, so we were creating a new Security object for it in the `ApplyUniverseSelection` method.

PR #239, opened 2015-12-28
Fixed universe subscriptions with Forex

StefanoRaggi:


mchandschuh:
This looks great! Can we also add a regression test? If you don't have data available to define a good regression for this then I can make it.  Also, is there any mechanism by which we'll remove internal feeds that are no longer needed?

StefanoRaggi:
I'm not sure which data is needed for this regression test. 
Maybe I'll leave it to you :)

For the removal of internal feeds, I assume you mean adding the checks at each call of `UniverseSelection.ApplyUniverseSelection()`. This is not currently implemented, but I verified that all added internal feeds are correctly unsubscribed after algorithm termination. Do we want to add the feature to this PR ?

mchandschuh:
We can do that in another PR, I'll put something on Trello. It's almost like `EnsureCurrencyDataFeeds` should return a `SecurityChanges` instances, so it can say to add and/or remove currency feeds as needed. This may be important when considering varying resolutions as well, such as, I added EURGBP on a second resolution but I have EURUSD on minute and GBPUSD on daily. Just thinking out loud here on some of the potential concerns/gotchas. Maybe some of this gets easier when we allow multiple subscriptions per security instance.

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/a29b78cc0a2a4b2721d17b2d6299db8f474ff9e0

PR #249, opened 2016-01-07
Delayed order fills one time step

StefanoRaggi:
Market orders are filled at the close of the current bar, all other order types are filled on the next bar.

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -188,8 +188,8 @@ namespace QuantConnect.Algorithm
             //Add the order and create a new order Id.
             var ticket = Transactions.AddOrder(request);
 
-            // In live mode, wait for the order event to process, only if the exchange is open
-            if (!asynchronous && LiveMode)
+            // Wait for the order event to process, only if the exchange is open
+            if (!asynchronous)
             {
                 Transactions.WaitForOrder(ticket.OrderId);
             }
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -178,7 +178,7 @@ namespace QuantConnect.Algorithm
 
             var request = CreateSubmitOrderRequest(OrderType.Market, security, quantity, tag);
 
-            //Initalize the Market order parameters:
+            //Initialize the Market order parameters:
             var preOrderCheckResponse = PreOrderChecks(request);
             if (preOrderCheckResponse.IsError)
             {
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -188,8 +188,8 @@ public OrderTicket MarketOrder(Symbol symbol, int quantity, bool asynchronous =
             //Add the order and create a new order Id.
             var ticket = Transactions.AddOrder(request);
 
-            //Wait for the order event to process, only if the exchange is open
-            if (!asynchronous)
+            // In live mode, wait for the order event to process, only if the exchange is open
+            if (!asynchronous && LiveMode)
```
jaredbroad: We actually allow synchronous orders to fill in backtest as well. This is so people can say "MarketOrder()" ... and then have the portfolio filled on the next line of code. (This is what the 5 second timeout was for -- let people have a synchronous order which times out and continues)

I understand this will make delayed fill difficult :) So to keep this behavior -- the delayed time step should probably only apply to non-market orders?

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -214,6 +215,13 @@ public void Scan()
                         continue;
                     }
 
+                    // all order fills are processed on the next bar
+                    if (order.Time == Algorithm.UtcTime)
+                    {
```
mchandschuh: very smart to implement it like this, maybe just add `order.Type != OrderType.Market` to allow market orders to pass

Common/Orders/Fills/ImmediateFillModel.cs
```diff
@@ -43,8 +43,8 @@ public virtual OrderEvent MarketFill(Security asset, MarketOrder order)
             // make sure the exchange is open before filling
             if (!IsExchangeOpen(asset)) return fill;
 
-            //Order [fill]price for a market order model is the current security price
-            fill.FillPrice = asset.Price;
+            //Order [fill]price for a market order model is the security open price
+            fill.FillPrice = asset.Open;
```
mchandschuh: if we let market orders fill on the same bar we'll need to revert this line

Brokerages/Backtesting/BacktestingBrokerage.cs
StefanoRaggi: Ok, so we keep the old behavior only for market orders. For backtesting it wouldn't have made a big difference anyway (close of current bar vs open of next bar).

jaredbroad:
Thank you @StefanoRaggi!

PR #256, opened 2016-01-26
Fixes BacktestingResultHandler resample period

StefanoRaggi:
It was always set to the default minimum of 4 minutes


mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/b8120e689851bab8697aa863c4226a0c3c73c032

PR #262, opened 2016-02-03
Average, Median, Typical, Weighted indicator selectors

StefanoRaggi:


mchandschuh:
Cool stuff @StefanoRaggi - where are the definitions for these? Are they arbitrarily named or are these common names?

StefanoRaggi:
@mchandschuh These are 4 TA-Lib indicators, that were converted as selectors: http://ta-lib.org/function.html
Their original names are AVGPRICE, MEDPRICE, TYPPRICE, WCLPRICE.
NinjaTrader has the last three as well (named Median, Typical and Weighted).

Common/Field.cs
```diff
@@ -57,6 +57,38 @@ public static partial class Field
...
+        /// </summary>
+        public static Func<BaseData, decimal> Typical
+        {
+            get { return TradeBarPropertyOrValue(x => (x.High + x.Low + x.Close) / 3m); }
+        }
+
+        /// <summary>
+        /// Defines an average price that is equal to (H + L + 2*C) / 4
+        /// </summary>
+        public static Func<BaseData, decimal> Weighted
```
mchandschuh: Thoughts on naming this `WeightedClose` to closer match ninja?

StefanoRaggi: Ninja calls it Weighted, TA-lib WeightedClose, you choose :)

mchandschuh: Sounds good to me! Great stuff. I'm going to fix the travis build, looks like a test is failing and then merge this in

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/3e10a7d4819badfb29f853d28752b83f2952f420

PR #263, opened 2016-02-03
Balance Of Power indicator (BOP)

StefanoRaggi:


mchandschuh:
@StefanoRaggi - I forgot to mention that this needs a unit test to confirm the numbers from an external source

Issue #95, opened 2015-06-06
Custom Chart Line Colors + Scatter Plot Symbols


jaredbroad:
Make custom chart line colors configurable, and scatter plots accept custom symbols when creating the series object. https://github.com/QuantConnect/Lean/blob/master/Common/Charting.cs

Add a ScatterStyle enum; with the HighStock chart symbol object types. 
Add a SeriesColor enum; with 20 default colors in hexadecimal.

Make the series take a color parameter on construction - default to null/Automatic, if automatic when adding to chart use the (n)-index of the enum depending on the number of other series in the chart.

PR #266, opened 2016-02-04
Issue #95: Add custom chart line colors and scatter plot symbols

AlexCatarino:
public enum SeriesColor was added with the 16 basic colors. @jaredbroad wants 20, but I could not find a good 20-color palette. Suggestions?
The enum members are named after the color names: Black, White, Blue, Teal, etc. If those are converted to string: "Black", "White", etc; HighStock recognize the strings, therefore there is no need for hexadecimal. 

public enum ScatterMarkerSymbol was added with the highStock five predefined symbols: Circle, Square, Diamond, Triangle and TriangleDown. In this case, if these values are converted to strings, it won't work, because HighStock need them to be lowercase and TringleDown must be renamed to "triangle-down".

### DIFF
--- a/Common/Charting.cs
+++ b/Common/Charting.cs
@@ -135,6 +135,16 @@ namespace QuantConnect
         /// </summary>
         public SeriesType SeriesType = SeriesType.Line;
 
+        /// <summary>
+        /// Color the series 
+        /// </summary>
+        public SeriesColor SeriesColor = SeriesColor.Black;
+        
+        /// <summary>
+        /// Shape or symbol for the marker in a scatter plot
+        /// </summary>
+        public ScatterMarkerSymbol ScatterMarkerSymbol = ScatterMarkerSymbol.Circle;
+
         /// Get the index of the last fetch update request to only retrieve the "delta" of the previous request.
         private int _updatePosition;
 
@@ -153,6 +163,8 @@ namespace QuantConnect
             SeriesType = SeriesType.Line;
             Unit = "$";
             Index = 0;
+            SeriesColor = SeriesColor.Black;
+            ScatterMarkerSymbol = ScatterMarkerSymbol.Circle;
         }
 
         /// <summary>
@@ -166,6 +178,8 @@ namespace QuantConnect
             SeriesType = type;
             Index = 0;
             Unit = "$";
+            SeriesColor = SeriesColor.Black;
+            ScatterMarkerSymbol = ScatterMarkerSymbol.Circle;
         }
 
         /// <summary>
@@ -180,6 +194,8 @@ namespace QuantConnect
             SeriesType = type;
             Index = index;
             Unit = "$";
+            SeriesColor = SeriesColor.Black;
+            ScatterMarkerSymbol = ScatterMarkerSymbol.Circle;
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


Common/Charting.cs
```diff
@@ -334,4 +391,61 @@ public enum ChartType
         /// Stacked series on top of each other.
         Stacked
     }
+
+    /// <summary>
+    /// Color to the series object
+    /// Basic colors
+    /// </summary>
+    public enum SeriesColor
```
mchandschuh: I would prefer we use the build in `Color` struct for .NET. It is an open type and allows specifying individual RGB components as well as specifying colors directly, such as `Color.Blue`. In addition, the `Color` struct has some strange serialization rules. If it is a 'KnownColor' then it will serialize a string of the color's name, if it is not a known color then it will serialize comma separated RGB values. Let's not aim to support A (alpha) at the moment.

We can modify the serialization by creating a custom `JsonConverter`. For ease of use in the front-end, we desired the output of serialization to be of the form: `rgb(r,g,b)` so for a `Color` with `Red=123; Green=231; Blue=067` we expect an output of: `rgb(123,231,67)`. I would derive from the `TypeChangeJsonConverter<Color, string>` to make this a little simpler.

Common/Charting.cs
```diff
@@ -135,6 +135,16 @@ public class Series
         /// </summary>
         public SeriesType SeriesType = SeriesType.Line;
 
+        /// <summary>
+        /// Color the series 
+        /// </summary>
+        public SeriesColor SeriesColor = SeriesColor.Black;
```
mchandschuh: Let's rename this to simply `Color` as it is already a member of the `Series` class

Common/Charting.cs
```diff
@@ -334,4 +392,21 @@ public enum ChartType
         /// Stacked series on top of each other.
         Stacked
     }
+
+    /// <summary>
+    /// Shape or symbol for the marker in a scatter plot
+    /// </summary>
+    public enum ScatterMarkerSymbol
```
mchandschuh: I just re-read your comment on the serialization of these values and high charts. By default JSON.NET will serialize enumerations as integer values. Place `[JsonConverter(typeof(StringEnumConverter)]` attribute at the top to use the string values of the enumerations. You can then place an `[EnumMember(Value="my-serialized-enum-value")]` on each enum member to specify the string to be used (by default will use the enum code name).

AlexCatarino:
I had some crazy stuff going on.
When I run the strategy without breakpoints, everything works fine. I print the color to hexadecimal string conversion and all good.
By when I stop the code with breakpoints, I get:
System.NullReferenceException: Arg_NullReferenceException
   at Microsoft.Win32.SystemEvents.AddEventHandler(Object key, Delegate value)
   at Microsoft.Win32.SystemEvents.add_UserPreferenceChanging(UserPreferenceChangingEventHandler value).

I was banging my head against the wall to send the code without that but I couldn't. I hope it is just a MSVS problem.

Common/Charting.cs
```diff
@@ -135,6 +139,17 @@ public class Series
         /// </summary>
         public SeriesType SeriesType = SeriesType.Line;
 
+        /// <summary>
+        /// Color the series 
+        /// </summary>
+        [JsonConverter(typeof(ColorJsonConverter))]
+        public Color Color = Color.Black;
```
jaredbroad: Please make this nullable and default to null --- if default to black, and the user adds 10 series, they will all be black :)

mchandschuh: Agreed - null can be an indication for consumers of the chart data to autogen a color

PR #268, opened 2016-02-04
Adds online implementation of Standard Deviation indicator

StefanoRaggi:


Tests/Indicators/StandardDeviationTests.cs
```diff
@@ -34,7 +42,7 @@ public void ComputesCorrectly()
             Assert.AreEqual(0m, std.Current.Value);
 
             std.Update(reference.AddDays(2), -1m);
-            Assert.AreEqual(1m, std.Current.Value);
+            Assert.AreEqual(0m, std.Current.Value);
```
mchandschuh: this delta is from warm up semantics being a little different? This impl requires 3 points before emitting a real value?

PR #273, opened 2016-02-07
Adds DoubleExponentialMovingAverage indicator

StefanoRaggi:


mchandschuh:
Nice work @StefanoRaggi - can we also add this to the `MovingAverageType` enumeration, and then there's a `MovingAverageTypeExtensions` class that can be updated, this will allow use in other indicator constructors via the enumeration.

PR #275, opened 2016-02-07
Adds TriangularMovingAverage indicator

StefanoRaggi:


mchandschuh:
Nice work @StefanoRaggi - can we also add this to the `MovingAverageType` enumeration, and then there's a `MovingAverageTypeExtensions` class that can be updated, this will allow use in other indicator constructors via the enumeration.

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/1e29d525d12e2e27ecf1b733eee5088c8815ba85

PR #276, opened 2016-02-08
Issue108: Implement PatternDayTraderMarginModel and its Unit Test

AlexCatarino:


Common/QuantConnect.csproj
```diff
@@ -256,6 +256,7 @@
     <Compile Include="Securities\Equity\EquityMarginModel.cs" />
     <Compile Include="Securities\Equity\EquityPortfolioModel.cs" />
     <Compile Include="Securities\DelayedSettlementModel.cs" />
+    <Compile Include="Securities\Equity\PatternDayTraderMarginModel.cs" />
```
mchandschuh: We should name this file the same as the class name: `PatternDayTradingMarginModel`

Common/Securities/Equity/PatternDayTraderMarginModel.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace QuantConnect.Securities.Equity
+{
+    /// <summary>
+    /// Represents a simple margining model where margin/leverage depends on market state (open or close).
+    /// During regular market hours, leverage is 4x, otherwise 2x
+    /// </summary>
+    public class PatternDayTradingMarginModel : SecurityMarginModel
+    {
+        private decimal _openmarketleverage = 4.0m;
+        private decimal _closedmarketleverage = 2.0m;
```
mchandschuh: nit-picky, but please use camel casing for member variables: `_openMarketLeverage` and `_closedMarketLeverage`

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
```diff
@@ -0,0 +1,221 @@
...
+        private static Security CreateSecurity(DateTime newLocalTime)
+        {
+            var security = new Security(CreateUsEquitySecurityExchangeHours(), CreateTradeBarConfig());
+            security.Exchange.SetLocalDateTimeFrontier(newLocalTime);
+            security.SetLocalTimeKeeper(TimeKeeper.GetLocalTimeKeeper(TimeZones.NewYork));
+            security.SetMarketPrice(new IndicatorDataPoint(Symbols.SPY, newLocalTime, 100m));
+            return security;
+        }
+
+        private static SecurityExchangeHours CreateUsEquitySecurityExchangeHours()
```
mchandschuh: Please move private methods to the bottom of files, typically files should be laid out as follows:

private state/variables
public properties
constructors
public methods
protected/internal/private methods

Common/Securities/Equity/PatternDayTraderMarginModel.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PatternDayTradingMarginModel" />
+        /// </summary>
+        /// <remarks>
+        /// Set the base leverage to 4x (regular trading hours)
+        /// </remarks>
+        public PatternDayTradingMarginModel()
+            : base(4.0m)
```
mchandschuh: maybe this should be calling `: this(2.0m, 4.0m)` ?

Common/Securities/Equity/PatternDayTraderMarginModel.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <param name="order">The order to be executed</param>
+        /// <returns>The total margin in terms of the currency quoted in the order</returns>
+        public override decimal GetInitialMarginRequiredForOrder(Security security, Order order)
+        {
+            var closedopenratio = security.Exchange.ExchangeOpen ? 1 : _closedmarketleverage/_openmarketleverage;
+
+            var orderFees = security.FeeModel.GetOrderFee(security, order);
+
+            var price = order.Status.IsFill() ? order.Price : security.Price;
+            return order.GetValue(price) * InitialMarginRequirement / closedopenratio + orderFees;
```
mchandschuh: `order.GetValue(currentMarketPrice)` was refactored into `order.GetValue(Security` - you may need to merge in upstream/master

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
```diff
@@ -0,0 +1,221 @@
...
+        public void InitializationTests()
+        {
+            // No parameters initialization, used default PDT 4x leverage open market and 2x leverage otherwise
+            var model = new PatternDayTradingMarginModel();
+            var leverage = model.GetLeverage(CreateSecurity(Noon));
+
+            Assert.AreEqual(4.0m, leverage);
+
+            model = new PatternDayTradingMarginModel(2.0m, 5.0m);
+            leverage = model.GetLeverage(CreateSecurity(Noon));
```
mchandschuh: interestingly the implementation doesn't override the `GetLeverage` method

Common/Securities/Equity/PatternDayTraderMarginModel.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        /// Gets the total margin required to execute the specified order in units of the account currency including fees
+        /// </summary>
+        /// <remarks>
+        /// If we are in regular market hours, base 4x leverage is used, otherwise leverage is reduced to 2x
+        /// </remarks>
+        /// <param name="security">The security to compute initial margin for</param>
+        /// <param name="order">The order to be executed</param>
+        /// <returns>The total margin in terms of the currency quoted in the order</returns>
+        public override decimal GetInitialMarginRequiredForOrder(Security security, Order order)
```
jaredbroad: You should also update margin requirements in maintenance margin as well. When hold 4x leverage positions past 4pm the broker typically liquidates. This is handled in LEAN you just need to override the other Maintenance function too.

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
```diff
@@ -0,0 +1,221 @@
...
+            // Market is Closed on Tuesday, Feb, 16th 2016 at Midnight
+
+            var leverage = 2m;
+
+            var model = new PatternDayTradingMarginModel();
+            var security = CreateSecurity(MidNight);
+            var order = new MarketOrder(security.Symbol, 100, security.LocalTime, type: security.Type);
+
+            var expected = 100*100m/leverage + 1;
+            var actual = model.GetInitialMarginRequiredForOrder(security, order);
```
mchandschuh: I wonder what the result would be if I called `model.GetLeverage()` here -- I would expect to get 2x, correct?

Common/Securities/Equity/PatternDayTraderMarginModel.cs
```diff
@@ -0,0 +1,78 @@
...
+using QuantConnect.Orders;
+
+namespace QuantConnect.Securities.Equity
+{
+    /// <summary>
+    /// Represents a simple margining model where margin/leverage depends on market state (open or close).
+    /// During regular market hours, leverage is 4x, otherwise 2x
+    /// </summary>
+    public class PatternDayTradingMarginModel : SecurityMarginModel
+    {
```
mchandschuh: Are there more methods that need to be overridden for this implementation?

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
AlexCatarino: ReShaper did that! I used the clean code tool and it reshuffled the methods.

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
AlexCatarino: Yes, but I will add this test. Why not? It's just 2 lines of code.

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
mchandschuh: In resharper's settings you can set the order of the methods, by convention we push private methods to the bottom. This link should be able to guide you on setting up your local resharper to move private methods to the bottom, you should be able to use the graphic interface and not need to modify the XML directly. I'll look into getting some solution specific settings in so everyone is using the same resharper dotSettings file

Common/Securities/Equity/PatternDayTraderMarginModel.cs
AlexCatarino: I guess, I'm looking into it. After Jared's comment on maintenance margin.

Tests/Common/Securities/Equity/PatternDayTradingMarginModelTests.cs
AlexCatarino: That would be great.

Common/Securities/Equity/PatternDayTradingMarginModel.cs
```diff
@@ -0,0 +1,148 @@
...
+        public PatternDayTradingMarginModel()
+            : this(2.0m, 4.0m)
+        {
+        }
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PatternDayTradingMarginModel" />
+        /// </summary>
+        /// <param name="closedmarketleverage"></param>
+        /// <param name="openmarketleverage"></param>
```
mchandschuh: Prefer that we fill out these parameter descriptions -- lots of users and contributors rely on us having full source documentation

Common/Securities/Equity/PatternDayTradingMarginModel.cs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        /// <summary>
+        /// Gets the margin currently alloted to the specified holding
+        /// </summary>
+        /// <param name="security">The security to compute maintenance margin for</param>
+        /// <returns>The maintenance margin required for the </returns>
+        public override decimal GetMaintenanceMargin(Security security)
+        {
+            return base.GetMaintenanceMargin(security)*GetMarginCorrection(security);
```
mchandschuh: this looks great now, curious why the same pattern isn't used for `GetInitialMarginRequiredForOrder` or `GenerateMarginCallOrder` - it would be nice to minimize the duplicate between this class and the parent class

Common/Securities/Equity/PatternDayTradingMarginModel.cs
```diff
@@ -0,0 +1,148 @@
...
+            if (!security.Holdings.Invested)
+            {
+                return null;
+            }
+
+            // compute the value we need to liquidate in order to get within margin requirements
+            var delta = totalMargin - netLiquidationValue;
+
+            // compute the number of shares required for the order, rounding up
+            var quantity = (int)(Math.Round(delta / security.Price, MidpointRounding.AwayFromZero) / (MaintenanceMarginRequirement*GetMarginCorrection(security)));
```
mchandschuh: maybe even the base implementation could be reworked so that this method (on the base) uses GetMaintainenceMargin -- then we probably don't even need to override this method -- thoughts?

AlexCatarino: Yes! I didn't need to override `GetMarginRemaining` because it doesn't use `MaintenanceMarginRequirement` but it calls the security's `GetMarginRemaining` (`security.MarginModel.GetMaintenanceMargin(security)`).
If we do that on `SubmitOrderRequest`  and `GetInitialMarginRequiredForOrder` on the base, we don't override them. Oh, we need to include a `GetInitialMargin` too.
After this is done, we only need to override `GetMaintenanceMargin` and `GetInitialMargin`.

mchandschuh:
This was merged (with edits) in https://github.com/QuantConnect/Lean/commit/770dd714bdc22e9a4dd2a01cd77ea082f56734fc

PR #277, opened 2016-02-08
Adds TestIndicatorReset method in TestHelper class

StefanoRaggi:


Tests/Indicators/TestHelper.cs
```diff
@@ -199,6 +199,48 @@ public static void TestIndicator(IndicatorBase<TradeBar> indicator, string exter
...
+            foreach (var data in GetTradeBarStream(externalDataFilename, false))
+            {
+                indicator.Update(data);
+            }
+
+            Assert.IsTrue(indicator.IsReady);
+
+            indicator.Reset();
+
+            AssertIndicatorIsInDefaultState(indicator);
```
mchandschuh: we also want to verify that any 'sub-indicators' have also been reset. this is the case for indicators that make use of other indicators or have other indicators as properties/fields.

Maybe we can do something where we get all the indicator fields on the type and also assert that they're in the default state:
```C#
var fields = indicator.GetType.GetFields(BindingFlags.Instance|BindingFlags.Public|BindingFlags.NonPublic);
var indicatorFields = fields.Where(x => x.FieldType.IsSubclassOfGeneric(typeof(IndicatorBase<>)));
foreach (var subindicator in indicatorFields)
{
    // need to call the AssertIndicatorIsInDefaultState generic method at runtime
    // something like typeof(TestHelper).GetMethod("AssertIndicatorIsInDefaultState).MakeGenericMethod( {indicatorbase typearg})
}
```

The above was just quickly written but shows a possible solution

mchandschuh:
@StefanoRaggi - great stuff!

PR #278, opened 2016-02-08
Adding Expected Returns and Modifying LogReturn

ijebir:


jaredbroad:
Please use full english (no contractions) in the file names. We follow MS conventions in general for the project (variable names etc). You can read more about these conventions here:
https://msdn.microsoft.com/en-us/library/ms229045%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396

Please also submit the tests associated with the indicators. There are lots of examples in the /Tests/Indicators project.

Indicators/expReturn.cs
```diff
@@ -0,0 +1,56 @@
...
+        	/// <param name="input">The input given to the indicator</param>
+        	/// <param name="window">The window for the input history</param>
+        	/// <returns>A new value for this indicator</returns>
+		protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint> window, IndicatorDataPoint input)
+		{
+			_rollingSum += input.Value;
+			if(Samples < Period)
+				return 0m;
+			var meanValue = _rollingSum / Period;
+			var remvedValue = window[Period -1];
```
mchandschuh: Please set your text editor to use soft tabs (1 tab = 4 spaces). This is the convention we use in the rest of the code base.

mchandschuh:
Thanks for the contribution! Please add unit tests that verify the indicator's behavior against an external data source. You can find examples in QuantConnect.Tests\Indicators.

mchandschuh:
Is there a paper we can link to for the documentation of this indicator? I fear `ExpectedReturn` is not common enough for us to not have a good blurb or a link to a white paper that users can learn about it

Issue #70, opened 2015-04-23
Tests for Base Security Class (Security object)


jaredbroad:
Security class is fundamental entity representing an asset object in Lean. For the long term future of Lean we need solid tests which cover this object to ensure future changes don't break current behavior.

Following test-driven-design we must break testing units into their smallest components. For Security we need to test each property and method, and that changes to its internals, then make sure its still behaving.

To test Security we propose the following steps, first create a security test file in the tests project.

1. Init Security object with configuration object and update its holdings to a known state.

2. Test simple properties: e.g. Test security.Symbol is same as config.Symbol which constructed it. (Symbol, Type, Resolution, IsFillForward, IsExtendedMarketHours, SubscriptionDataConfig, MarginModel.Leverage == Leverage initializer, 

3. Test holdings properties behave as expected: HoldStock,Invested true when long or short, 

3. Test Methods:  Updating the security with price data causes the price property to change.

4. SetLeverage() command updates the security.Leverage value.

5. Test constructor - Verify the state of the internal initialization is as expected (e.g. Cache == typeof SecurityCache() class for each of the objects initialized in the constructor)

AnObfuscator:
Is anyone working on this? If not, I could take this on.

mchandschuh:
No one is currently working on this, please do!

PR #280, opened 2016-02-09
Add unit tests for Base Security Class (Security object) #70

AlexCatarino:


Tests/Common/Securities/SecurityTests.cs
```diff
@@ -0,0 +1,156 @@
...
+            var exchangeHours = SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork);
+            var config = CreateTradeBarConfig();
+            var security = new Security(exchangeHours, config);
+
+            Assert.AreEqual(config, security.SubscriptionDataConfig);
+            Assert.AreEqual(config.Symbol, security.Symbol);
+            Assert.AreEqual(config.SecurityType, security.Type);
+            Assert.AreEqual(config.Resolution, security.Resolution);
+            Assert.AreEqual(config.FillDataForward, security.IsFillDataForward);
+            Assert.AreEqual(exchangeHours, security.Exchange.Hours);
```
AlexCatarino: Instead of this test, my first option was:
`Assert.AreEqual(new SecurityExchange(exchangeHours), security.Exchange);`
It didn't pass and I don't know why.

mchandschuh: The `SecurityExchange` type uses reference equality, that is, there is no overridden `Equals` method, so to instances are equal if and only if they are in fact the same instance (same memory location)

AlexCatarino: Should I use `Assert.AreSame(new SecurityExchange(exchangeHours), security.Exchange);` instead? Or are we good?

mchandschuh: Asserting that the object instance you passed in is the same object instance that ends up being exposed is sufficient.

PR #284, opened 2016-02-10
Adds T3 Moving Average indicator

StefanoRaggi:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -752,6 +749,23 @@ public TriangularMovingAverage TRIMA(Symbol symbol, int period, Resolution? reso
...
         /// <summary>
+        /// Creates a new T3MovingAverage indicator.
+        /// </summary>
+        /// <param name="symbol">The symbol whose T3 we want</param>
+        /// <param name="period">The period over which to compute the T3</param>
+        /// <param name="volumeFactor">The volume factor to be used for the T3</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value)</param>
+        /// <returns>The T3MovingAverage indicator for the requested symbol over the specified period</returns>
+        public T3MovingAverage T3(Symbol symbol, int period, int volumeFactor, Resolution? resolution = null, Func<BaseData, decimal> selector = null)
```
mchandschuh: maybe we should define `int volumeFactor = 1,` as a default value?

Indicators/T3MovingAverage.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IndicatorDataPoint input)
+        {
+            _gd1.Update(input);
+
+            if (Samples <= 2 * (_period - 1))
```
mchandschuh: Is this really asking `if (_gd1.IsReady) { }` ? Same for the below `if` checks against the `Samples` property

Tests/Indicators/MovingAverageTypeExtensionsTests.cs
```diff
@@ -33,6 +33,21 @@ public void CreatesCorrectAveragingIndicator()
...
+
+            indicator = MovingAverageType.DoubleExponential.AsIndicator(1);
+            Assert.IsInstanceOf(typeof(DoubleExponentialMovingAverage), indicator);
+
+            indicator = MovingAverageType.TripleExponential.AsIndicator(1);
+            Assert.IsInstanceOf(typeof(TripleExponentialMovingAverage), indicator);
+
+            indicator = MovingAverageType.Triangular.AsIndicator(1);
+            Assert.IsInstanceOf(typeof(TriangularMovingAverage), indicator);
+
```
mchandschuh: Maybe we can automate this test by decorating the enum members with an attribute that points to the type they create, then this method could be written using reflection... not for now, but thoughts for later as we add to this enumeration

Algorithm/QCAlgorithm.Indicators.cs
StefanoRaggi: @mchandschuh If volume factor is 1, T3 is equivalent to DEMA(DEMA(DEMA(x))).
Looking at other implementations, the value is a parameter for T3.

Indicators/T3MovingAverage.cs
StefanoRaggi: @mchandschuh There is a different multiplier used in these checks (2,4) instead of 6

Tests/Indicators/MovingAverageTypeExtensionsTests.cs
StefanoRaggi: While I was duplicating lines, I was thinking the same thing :smile:

Indicators/T3MovingAverage.cs
mchandschuh: My point is that the following:
```C#
if (Samples <= 2 * (_period - 1))
{
}
```
is equivalent to the following:
```C#
if (!_gd1.IsReady)
{
}
```
I'm gathering this from `DoubleExponentialMovingAverage.IsReady` implementation looks the same.

Likewise, the next if check is checking:
`if (!_gd2.IsReady) { } `

StefanoRaggi: Sorry, I missed the __gd1_ part in the `IsReady` call in your first comment... :blush:  You're absolutely right, I'll make the change.

Indicators/MovingAverageTypeExtensions.cs
```diff
@@ -35,18 +35,28 @@ public static IndicatorBase<IndicatorDataPoint> AsIndicator(this MovingAverageTy
...
                     return new DoubleExponentialMovingAverage(period);
+
                 case MovingAverageType.TripleExponential:
                     return new TripleExponentialMovingAverage(period);
+
                 case MovingAverageType.Triangular:
                     return new TriangularMovingAverage(period);
+
+                case MovingAverageType.T3:
+                    return new T3MovingAverage(period, 1);
```
mchandschuh: Looking online it seems the normal default value for the volume factor is 0.7? Did TA-lib use a value of 1?

http://www.binarytribune.com/forex-trading-indicators/t3-moving-average-indicator/
http://www.multicharts.com/support/base/?action=article&id=1419

Algorithm/QCAlgorithm.Indicators.cs
mchandschuh: My reasoning behind providing a default value in this method is because many users may not know was a `volumeFactor` is, and so providing a default allows them to only specify a period. This default value should match the value used in the `AsIndicator` extension method as well.

Indicators/MovingAverageTypeExtensions.cs
StefanoRaggi: My mistake, default is 0.7 for TA-lib as well.

PR #285, opened 2016-02-10
Adds Normalized Average True Range indicator (NATR)

StefanoRaggi:


Indicators/NormalizedAverageTrueRange.cs
```diff
@@ -0,0 +1,102 @@
...
+                _atr.Update(input);
+                _lastAtrValue = _atr;
+            }
+            else
+            {
+                // next TrueRange values are smoothed using Wilder's approach
+                _lastAtrValue = (_lastAtrValue * (_period - 1) + _tr) / _period;
+            }
+
+            return _lastAtrValue / input.Close * 100;
```
mchandschuh: I think we need to worry about the div zero case here?

StefanoRaggi: @mchandschuh Yes, it definitely must be handled, do you suggest returning zero in this case ?

mchandschuh: It's tough. In the divide by zero cases I think we should be returning the result of the limit as the denominator goes to zero. The issue here is technically it goes to infinity, correct? I'm not sure what the right answer is here, I'm open to suggestions, but I feel that zero may be incorrect. @jaredbroad -- any thoughts on limits here?

StefanoRaggi: I checked the TA-lib source code, it returns zero if the denominator is zero.

mchandschuh: zero sounds good -- we can revisit this later

PR #289, opened 2016-02-13
Python algos in a file with the algo name like C#

AlexCatarino:
build.bat (batch.sh in Linux) includes the code inside the file with the desired algorithm (set in config.json by "algorithm-type-name") in the python DLL (QuantConnect.Algorithm.Python.dll) via IronPython.
In the AlgorithmFactory,  Loader.TryCreatePythonAlgorithm imports the module with the algorithm name instead of "main".


Algorithm.Python/build.sh
```diff
@@ -7,9 +7,12 @@ rm ../Launcher/bin/Debug/QuantConnect.Algorithm.Python.dll
 # Set the script variables: assuming installing the ./compiler/library/ and the caller is in ./compiler/bin/Debug/build.sh
 ipy="../../IronPython-2.7.5/ipy.exe"
 pyc="../../IronPython-2.7.5/Tools/Scripts/pyc.py"
- 
+
+# Get algorithm-type-name from config.json
+pyfile=$(grep -Po '(?<="algorithm-type-name": ")[^"]*' ../Launcher/config.json).py
```
mchandschuh: that's cool!

PR #292, opened 2016-02-16
Adds IndicatorBase<T>.ValidateAndComputeNextValue

StefanoRaggi:


Indicators/IndicatorResult.cs
```diff
@@ -0,0 +1,53 @@
...
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="IndicatorResult"/> class
+        /// </summary>
+        /// <param name="value">The value output by the indicator</param>
+        /// <param name="status">The status returned by the indicator</param>
+        public IndicatorResult(decimal value, IndicatorStatus status = IndicatorStatus.Success)
```
mchandschuh: may be nit picky - but I tend to avoid optional ctor args when it's something central to the type

Indicators/IndicatorBase.cs
```diff
@@ -91,11 +91,15 @@ public bool Update(T input)
...
-                Current = new IndicatorDataPoint(input.Time, nextValue);
 
-                // let others know we've produced a new data point
-                OnUpdated(Current);
+                var nextResult = ValidateAndComputeNextValue(input);
+                if (nextResult.Status == IndicatorStatus.Success)
+                {
+                    Current = new IndicatorDataPoint(input.Time, nextResult.Value);
+
+                    // let others know we've produced a new data point
```
mchandschuh: this is great for now, solves the immediate need. I'm thinking in the future we may want to have another event that can get called on an invalid result

Indicators/IndicatorResult.cs
StefanoRaggi: @mchandschuh No problem, I'll remove the optional, I thought that future usage inside indicators looked less verbose:

`return new IndicatorResult(value)`
vs
`return new IndicatorResult(value, IndicatorStatus.Success)`

mchandschuh: the change isn't required, now that you've written them both out I can see your point. we can go either way with this one, probably best to go the less verbose route for a user/API perspective as well

mchandschuh:
Merged in https://github.com/QuantConnect/Lean/commit/f35e7f9dbe1210d466e3cabf04e24b24ef400d23

PR #295, opened 2016-02-17
Fix for History<Tick>

jameschch:
Calls to History<Tick>() were returning nothing. Problem was related to unexpected behavior of dynamic type.


jameschch:
Yes that's right, just the first commit was intended.

mchandschuh:
In the future please issue pull requests from a dedicated branch to avoid mixing. Also, have a look at the [contributor's guide](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).

mchandschuh:
The heart of the 'one-tick' per symbol issue is the dictionary is symbol->T instead of symbol->List<T>, but maybe there's a way around this by type checking. The Tick is a special case and we do expect multiple of them. For other data types (TradeBar/Custom) it is safe to assume one data point per time step.

As for the if check comment - I think it may be better to check `if (typeof(T) == typeof(Tick))` instead of inspecting the internal data collection, since we're more concerned about what the caller requested then what's inside the collection currently, but I may be incorrect. Have you tried it out with tick and trade bar data inside?

jameschch:
Thanks, this explains a lot. I agree the suggested approach of `if (typeof(T) == typeof(Tick))` is a much better idea. I'll look closely at the Tick per timestep issue but in unit tests for Tick, Tradebar and Quandl I've not noticed any problems. I think this change is worth doing as its preferable to encapsulate as History<Tick> rather than Slices.

PR #296, opened 2016-02-18
Python Unit Test

AlexCatarino:
We extend algorithm unit tests to those written in Python using the existing framework.


Tests/RegressionTestsPython.cs
```diff
@@ -0,0 +1,294 @@
...
+                {"Profit-Loss Ratio", "0"},
+                {"Alpha", "0.752"},
+                {"Beta", "0.186"},
+                {"Annual Standard Deviation", "0.193"},
+                {"Annual Variance", "0.037"},
+                {"Information Ratio", "1.316"},
+                {"Tracking Error", "0.246"},
+                {"Treynor Ratio", "4.572"},
+                {"Total Fees", "$3.09"}
+            }, "Python");
```
jaredbroad: Perhaps its better to loop over the supported languages, and use the same result dictionary? The results should be the same for all languages.

foreach (supported language as language)
{
    RunRegressionTests (language);  
}

PR #298, opened 2016-02-19
Adds Heikin-Ashi indicator

StefanoRaggi:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -846,6 +846,23 @@ public NormalizedAverageTrueRange NATR(Symbol symbol, int period, Resolution? re
...
+
+        /// <summary>
+        /// Creates a new Heikin-Ashi indicator.
+        /// </summary>
+        /// <param name="symbol">The symbol whose Heikin-Ashi we want</param>
+        /// <param name="period">The period over which to compute the NATR</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value)</param>
+        /// <returns>The Heikin-Ashi indicator for the requested symbol over the specified period</returns>
+        public HeikinAshi HA(Symbol symbol, Resolution? resolution = null, Func<BaseData, TradeBar> selector = null)
```
mchandschuh: We may want to spell out `HeikinAshi` here, kind of like the ichomoku indicator. `HA` seems vague

Indicators/HeikinAshi.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <summary>
+        /// Gets the Heikin-Ashi Low
+        /// </summary>
+        public IndicatorBase<IndicatorDataPoint> Low { get; private set; }
+
+        /// <summary>
+        /// Gets the Heikin-Ashi Close
+        /// </summary>
+        public IndicatorBase<IndicatorDataPoint> Close { get; private set; }
+
```
mchandschuh: may want to provide a property for the current bar:
```
public TradeBar CurrentBar
{
    get { return new TradeBar{ OHLC };
}
```

Indicators/HeikinAshi.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+            else
+            {
+                Open.Update(new IndicatorDataPoint(input.Time, (Open + Close) / 2));
+                Close.Update(new IndicatorDataPoint(input.Time, (input.Open + input.High + input.Low + input.Close) / 4));
+                High.Update(new IndicatorDataPoint(input.Time, Math.Max(input.High, Math.Max(Open, Close))));
+                Low.Update(new IndicatorDataPoint(input.Time, Math.Min(input.Low, Math.Min(Open, Close))));
+            }
+
+            return Close;
```
mchandschuh: Maybe we could use the new validate and compute feature to prevent firing updated events before a bar is closed? thoughts?

StefanoRaggi: This indicator has no warmup, just the first bar calculated differently, so maybe no need to validate ?

Indicators/HeikinAshi.cs
StefanoRaggi: Do we also need a valid Symbol in the TradeBar ? I didn't set it in the OHLC.

Indicators/HeikinAshi.cs
mchandschuh: Maybe I understand heiken ashi incorrectly (most-probable :) ) - for each input bar, is there exactly one output bar, or does it sometimes take two input bars to make one output bar?

Indicators/HeikinAshi.cs
mchandschuh: it would be nice. most indicators don't set the symbol of their data points, so maybe it's OK to not populate the symbol? I'm just thinking of usability and imagine people would typically want to get a `TradeBar` instance out of this indicator

Indicators/HeikinAshi.cs
StefanoRaggi: To build each output bar it needs the current bar and the previous one.
In the link below, only for the first bar they use a different formula.

More info: http://stockcharts.com/school/doku.php?id=chart_school:chart_analysis:heikin_ashi

Indicators/HeikinAshi.cs
StefanoRaggi: From a usability standpoint, I thought that:

```
var ha = new HeikinAshi();
if (ha.Open > ha.Close)
```

was better than

`if (ha.CurrentBar.Open > ha.CurrentBar.Close)`

I'll add the CurrentBar property anyway and leave the Symbol empty for now.

mchandschuh: I don't disagree with your use case, but what if I want to maintain a `RollingWindow<TradeBar>` that is the history of the indicator, or pass the trade bar around to other parts of the system which don't know about the `HeikenAshi` indicator.

Sounds good, thanks!

StefanoRaggi: As always, you're right. I would have realized the need for it the first time I used it :smile:

mchandschuh:
Looks good, thanks.

PR #300, opened 2016-02-22
Synchronous results time set to UTC

AlexCatarino:
In the Backtesting results handler class, the synchronous result events, sampling and message reading uses the time informed by the algorithm in UTC instead of local. The former way was messing up statistics when Lean was executed in other countries.


mchandschuh:
@AlexCatarino - very nice!

PR #302, opened 2016-02-23
ToolBox changes

naicigam:
- Moved ToolBox helper classes to the Engine DLL, so they can be used by external tools.
- Fix issue in the DataWriter that generated the wrong zip file path.
- Remove unnecesary MySQL reference in Engine.


jaredbroad:
Thanks for the PR @naicigam! Please keep submissions to one topic so make them as focused as possible, and avoid changes which affect lots of files but don't have much impact (code churn - e.g. renaming the Data writer to LeanDataWriter). 

I'll close this for now but please submit a new PR for the MySQL reference. 

Possibly we could move the LeanDataWriter class to Common, but it doesn't really belong in the engine.

naicigam:
Sorry about that. Just submitted both the MySQL reference and the LeanDataWriter fix.

I would be happy to move LeanDataWriter, since I am currently copy/pasting it into my VS project so I can use it. Should I move it to Common or should we consider another option?

Thanks!

PR #294, opened 2016-02-16
Adds QCUSellinMay.py

AlexCatarino:
Adds Python version of QuantConnect University Sell in May

PR #303, opened 2016-02-23
Scripts includes all python script files in the DLL

AlexCatarino:
We may need use classes and methods that are stored in other files in the directory (like we do in the C# Algorithms project)
Also, we need all Python Algorithms in the correspondent DLL to run unit tests of all the algorithms in a row
The dll is now copied to the Tests project

Ref #294


jaredbroad:
Good work @AlexCatarino, great to see Python edging closer to production.

PR #305, opened 2016-02-24
TA-lib Candlestick Patterns (part 1)

StefanoRaggi:


kaffeebrauer:
Love the changes! Thanks for TALib support :+1:

Algorithm/CandlestickPatterns.cs
```diff
@@ -0,0 +1,508 @@
...
+        /// Initializes a new instance of the <see cref="CandlestickPatterns"/> class
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance</param>
+        public CandlestickPatterns(QCAlgorithm algorithm)
+        {
+            _algorithm = algorithm;
+        }
+
+        /// <summary>
+        /// Creates a new Two Crows pattern indicator.
```
mchandschuh: We may want to make these `<see cref="TwoCrows" />` so that they can navigate to a description of the `TwoCrows` indicator, thoughts?

Indicators/CandlestickPatterns/CandleSettings.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <summary>
+        /// Returns the candle setting for the requested type
+        /// </summary>
+        /// <param name="type">The candle setting type</param>
+        public static CandleSetting Get(CandleSettingType type)
+        {
+            CandleSetting setting;
+            DefaultSettings.TryGetValue(type, out setting);
+            return setting;
+        }
```
mchandschuh: Do we need to provide a Set here? I saw somewhere reference that these are settable settings.

mchandschuh:
@StefanoRaggi - couple minor comments, absolutely epic PR!

Algorithm/CandlestickPatterns.cs
StefanoRaggi: @mchandschuh Good idea, I'll add the links.

Indicators/CandlestickPatterns/CandleSettings.cs
StefanoRaggi: @mchandschuh The TA-lib library has a SetCandleSettings function to change the default values (which is referenced in the comments of each pattern), for maximum flexibility.
I'll add the Set method.

PR #306, opened 2016-02-25
Constants.DataFolder returns the correct value after calling Config.Set("data-folder", ...)

naicigam:


mchandschuh:
@naicigam - I'm curious what the use case is here? My concern is that this is a very frequently hit property, so ideally we wouldn't need to peruse a `JObject` each time we need it.

naicigam:
@mchandschuh I am trying to test the import of data by running a local backtests and making sure the algorithms can access the imported data.

Since I have multiple test cases that import data for the same instruments (same file names), I am switching the data folder between tests. and this was the less intrusive way I found to do this.

The Config and the Constant classes are tightly coupled inside Lean. Do you have any plans or intentions to reduce this coupling? I would be happy to start working on this if you can point me out in the right direction, since you know Lean much better than I do.

mchandschuh:
Maybe the easiest implementation here would be to implement get/set -- it seems that the `Constants` class is no longer really `Constants` -- any ideas on a better name?

naicigam:
We could also implement a Reset method, like the one in the Composer class.

mchandschuh:
I'll close this PR. Please submit another with the get/set change, and another if you'd like a Reset method. Thanks. Was thinking another good change may be to renames `Constants` to something else, maybe `Globals`

PR #310, opened 2016-03-03
Bitfinex Brokerage Integration

jameschch:
This is an integration with the REST and Websockets api of popular Bitcoin exchange Bitfinex, for the purposes of exchange and margin trading of BTCUSD only. The wss integration is original however, the REST and json classes are adapted from [https://github.com/jameschch/TradingApi.Bitfinex](https://github.com/jameschch/TradingApi.Bitfinex).

A reference to Moq has been added to allow unit testing of brokerage classes without an exchange account (or network connection). The brokerage classes have been extensively tested and have been live trading for several weeks, so are considered stable but with undiscovered defects.

There are also dependent fixes to several core classes. It might be preferred that these be submitted as separate PR's.

There are currently several known omissions that are considered for possible later implementation. Many of these are marked with todo: comments. They include:

1. Bitfinex margin call automated resolution.
2. Support for trading BTCLTC, LTCUSD, ETHBTC, ETHUSD.
3. Filtering of trades that breach Bitfinex's position reversal transaction limits.
4. Abandon the separate fork of TradingApi.Bitfinex and bring all classes into Lean repo.


jaredbroad:
Thanks for the PR @jameschch! Please split into smaller pull requests if possible. E.g. Just start with the data downloader for now and we'll merge one at a time.

By making PR's small the community is less likely to have merge conflicts and we're better able to review each PR. 

Please also remove the config changes you've made, and if possible the raw binary files in the tests. Can those binaries be added via Nuget?

There's also merge conflicts you might want to solve locally, and then push the smaller PR's to master one by one :)

jameschch:
No problem. I've broken down what I can into single commits. The Bitfinex integration could be broken down if you prefer but it makes sense to me to do a single PR for the entire feature.

The binaries have moved out of the Tests assembly. I'm considering bringing the source files into the main repo but there are a lot of json classes that might bloat the repo.

The only issue is related to the config changes: there is the environment structure and required brokerage parameters needed in the config.json so this has to change somehow. Will it make it easier with a separate PR with just the config changes?

mchandschuh:
I agree we can keep the entire `IBrokerage` implementation (and any tests) in one PR. This would be easiest for me. It should be as correct and complete as possible, including any required configuration changes for it to work in live mode.

jameschch:
Great, I'm also hoping to get this as complete and correct as possible. I believe this build should be functional and there is a sample algorithm, json.config and unit tests to get things started. In the meantime I've been discovering bugs in this build which we might also want to include.

Algorithm.CSharp/BaseBitcoin.cs
```diff
@@ -0,0 +1,87 @@
...
+                + " Total:" + Portfolio.TotalPortfolioValue);
+        }
+
+        /// <summary>
+        /// Must liquidate before reversing position
+        /// </summary>
+        //todo: implement transaction limits in brokerage model
+        protected virtual void Long()
+        {
+            if (Portfolio[BitcoinSymbol].IsShort)
```
mchandschuh: The brokerage implementation should handle this concern. Have a look at the `TradierBrokerage` which also has to look out for [zero crossings](https://github.com/QuantConnect/Lean/blob/master/Brokerages/Tradier/TradierBrokerage.cs#L999).

Common/Securities/BitfinexSecurityPortfolioManager.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (averageHoldingsLeverage <= 1.0m)
+            {
+                return new List<SubmitOrderRequest>();
+            }
+
+            //Will liquidate if position to equity ratio > 15%. We begin to issue warnings when Ticker to Position price ratio > 1.13            
+            foreach (var security in Securities.Values.Where(x => x.Holdings.Quantity != 0 && x.Price != 0))
+            {
+                decimal ratio = security.Holdings.Price / security.Holdings.AveragePrice;
+                if ((security.Holdings.IsShort && ratio > 1.13m) || (security.Holdings.IsLong && ratio < 0.87m))
```
mchandschuh: I wonder if there's a cleaner way to get the desired margin call behavior. If necessary, we can add another plugin point, but algorithms should not be expected to initialize a new `SecurityPortoflioManager` instance just to use bitfinex. Let's remove this from the brokerage implementation PR.

Algorithm.CSharp/BaseBitcoin.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Threading.Tasks;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+
+    /// <summary>
+    /// Base class for Bitcoin algorithms. This base class provides standard behaviour for trading BTCUSD on Bitfinex
+    /// trading on Bitfinex
+    /// </summary>
+    public abstract class BaseBitcoin : QCAlgorithm
```
mchandschuh: Any new algorithms should not be included in the brokerage implementation pull request, maybe we can split this algos into a separate PR. Also, ideally we wouldn't need a base bitcoin algorithm class.

Algorithm.CSharp/BaseBitcoin.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+
+        const string _bitcoinSymbol = "BTCUSD";
+
+        protected string BitcoinSymbol { get { return _bitcoinSymbol; } }
+
+        public BaseBitcoin()
+        {
+            Portfolio = new BitfinexSecurityPortfolioManager(Securities, Transactions);
+            SetBrokerageModel(BrokerageName.BitfinexBrokerage, AccountType.Margin);
```
mchandschuh: maybe we add the margin call behavior to the brokerage model (in a separate PR)

Brokerages/Bitfinex/BaseMessage.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <summary>
+    /// Base class for Bitfinex wss messages
+    /// </summary>
+    public abstract class BaseMessage
+    {
+
+        /// <summary>
+        /// Stored Keys for message
+        /// </summary>
+        protected string[] allKeys { get; set; }
```
mchandschuh: protected members should have pascal casing

Brokerages/Bitfinex/BaseMessage.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        /// <summary>
+        /// Returns typed value from untyped json array
+        /// </summary>
+        /// <param name="key"></param>
+        /// <returns></returns>
+        public decimal GetDecimal(string key)
+        {
+            return decimal.Parse(allValues[Array.IndexOf(allKeys, key)]);
```
mchandschuh: index of operations are notoriously slow since they are an O(n) operation. The comments say it comes from JSON, maybe it would be cleaner to save the data internally as a `JObject` which could give us dictionary-esque lookups

mchandschuh: Also, what if the key doesn't exist :)

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,469 @@
...
+        /// Divisor for prices. Scales prices/volumes to allow trades on 0.01 of unit
+        /// </summary>
+        protected decimal divisor = 100;
+        readonly object _fillLock = new object();
+        const string buy = "buy";
+        const string sell = "sell";
+        /// <summary>
+        /// Currently limited to BTCUSD
+        /// </summary>
+        protected Symbol symbol = Symbol.Create("BTCUSD", SecurityType.Forex, Market.Bitfinex);
```
mchandschuh: Ideally we would trust the symbol from incoming requests, such as the order objects or the subscription requests

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,469 @@
...
+        {
+            this.Initialize();
+        }
+
+        private void Initialize()
+        {
+
+            //todo: also stored in BrokerageData
+            apiSecret = Config.Get("bitfinex-api-secret");
+            apiKey = Config.Get("bitfinex-api-key");
```
mchandschuh: These initialization parameters should be accepted as constructor arguments. We could even include the `RestClient` as a constructor argument. This will provide consumers of this class more flexibility and remove the need for public get/set

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,469 @@
...
+        public override bool IsConnected
+        {
+            get { return this._tickerToken != null && !this._tickerToken.IsCancellationRequested; }
+        }
+
+        private decimal GetPrice(Order order)
+        {
+            if (order is StopMarketOrder)
+            {
+                return ((StopMarketOrder)order).StopPrice * divisor;
```
mchandschuh: The divisor in the brokerage implementation is a little intimidating. I think it all stems from the Order type using a integer for quantity. Maybe it's best we fix the problem there. I'm forseeing some ugly bugs if the algorithm and brokerage implementation's concept of 'divisor' are out of sync.

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,469 @@
...
+            var response = RestClient.GetActivePositions();
+            foreach (var item in response)
+            {
+                var ticker = RestClient.GetPublicTicker(TradingApi.ModelObjects.BtcInfo.PairTypeEnum.btcusd, TradingApi.ModelObjects.BtcInfo.BitfinexUnauthenicatedCallsEnum.pubticker);
+                list.Add(new Holding
+                {
+                    Symbol = Symbol.Create(item.Symbol, SecurityType.Forex, Market.Bitfinex),
+                    Quantity = decimal.Parse(item.Amount) * divisor,
+                    Type = SecurityType.Forex,
+                    CurrencySymbol = "B",
```
mchandschuh: should this be `BTC`?

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,469 @@
...
+                }
+            }
+            return list;
+        }
+
+        /// <summary>
+        /// Get queued tick data
+        /// </summary>
+        /// <returns></returns>
+        public IEnumerable<Data.BaseData> GetNextTicks()
```
mchandschuh: for cleanliness, we may want to split the `IDataQueueHandler` implementation into a separate partial class file, like was done with the`FxcmBrokerage`

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,469 @@
...
+                    Value = decimal.Parse(response.LastPrice) / divisor,
+                    TickType = TickType.Quote,
+                    Symbol = symbol,
+                    DataType = MarketDataType.Tick,
+                    Quantity = (int)(Math.Round(decimal.Parse(response.Volume), 2) * divisor)
+                });
+            }
+            if (!_tickerToken.IsCancellationRequested)
+            {
+                Thread.Sleep(8000);
```
mchandschuh: Sleeping threads on the .NET thread pool is not a [good idea](http://stackoverflow.com/questions/6305814/sleeping-in-a-pooled-c-sharp-thread)! You can introduce 'async-ness' in other ways if necessary. I think the ideal case would be to use a continuation or `Task.Delay` to run the `RequestTicker` function after the desired 8 seconds

Brokerages/Bitfinex/BitfinexBrokerage.cs
jaredbroad: Agree. Created issue https://github.com/QuantConnect/Lean/issues/330

Brokerages/Bitfinex/BitfinexBrokerageFactory.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <summary>
+        /// Create the Brokerage instance
+        /// </summary>
+        /// <param name="job"></param>
+        /// <param name="algorithm"></param>
+        /// <returns></returns>
+        public override Interfaces.IBrokerage CreateBrokerage(Packets.LiveNodePacket job, Interfaces.IAlgorithm algorithm)
+        {
+            var brokerage = new BitfinexWebsocketsBrokerage();
```
mchandschuh: we need to initialize the brokerage with the api-secret and api-key from the job packet's brokerage data dictionary. When running locally the values will be populated using the brokerage factory's brokerage data property

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
```diff
@@ -0,0 +1,214 @@
...
+                        this._channelId[(int)raw.chanId] = "ticker";
+                    }
+                }
+                else if (raw.chanId == 0)
+                {
+                    if (raw.status == "FAIL")
+                    {
+                        throw new Exception("Failed to authenticate with ws gateway");
+                    }
+                    Log.Trace("Successful wss auth");
```
mchandschuh: we like to prefix `Log.Trace` statements with the class and method, so this would become:
```C#
Log.Trace("BitfinexWebsocketsBrokerage.OnMessage(): Successful wss auth");
```
This makes it easy to know where log statements came from in the code

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
```diff
@@ -0,0 +1,214 @@
...
+                    AskPrice = msg.ASK / divisor,
+                    BidPrice = msg.BID / divisor,
+                    AskSize = (long)Math.Round(msg.ASK_SIZE * divisor, 0),
+                    BidSize = (long)Math.Round(msg.BID_SIZE * divisor, 0),
+                    Time = DateTime.UtcNow,
+                    Value = msg.LAST_PRICE / divisor,
+                    TickType = TickType.Quote,
+                    Symbol = symbol,
+                    DataType = MarketDataType.Tick,
+                    Quantity = (int)(Math.Round(msg.VOLUME, 2) * divisor)
```
mchandschuh: I think we should round after applying the divisor.

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
```diff
@@ -0,0 +1,214 @@
...
+                });
+            }
+        }
+
+        //todo: Currently populated but not used
+        private void PopulateWallet(string[][] data)
+        {
+            if (data.Length > 0)
+            {
+                lock (_cash)
```
mchandschuh: It's typically better to have a dedicated object for locking, so maybe we could have a `_cashLock` object instead of locking directly on the list

Common/Securities/BitfinexSecurityPortfolioManager.cs
jameschch: Yes, this is the simplest method for plugging in the margin warning behavior. This was a compromise that would avoid refactoring the base algorithm initialization. The alternative was to handle margin call events in the algo class and just remove them from the collection - which was much worse. If you don't override the vanilla margin call behavior you get common and unnecessary liquidations, which was even worse.

Brokerages/Bitfinex/BaseMessage.cs
jameschch: The problem here is the data is a raw array rather than key/value pairs, so this approach avoids accessing the json array directly by index number. So, this is an attempt to get dictionary-esque lookups. The alternative would be something like:

`decimal.Parse(allValues[3])`

..which makes me shudder.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: I have hard coded the symbol to limit the scope of the PR. I have a build with multi-currency support in test but was trying to make the review process simpler.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: Yes, the lack of DI was a pain point for the rest and wss clients. A constructor parameter is more common for injection.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: Great news! Bugs related to the divisor are basically a nightmare scenario, so a decimal quantity would be very welcome.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: There was recently a change to unicode to add the official bitcoin symbol: Ƀ (U+0243). I imagine this is currently poorly supported so "B" was a compromise. Using BTC is far less ambiguous.

Algorithm.CSharp/BaseBitcoin.cs
jameschch: I've provided the sample algorithm to show the required initialization for the brokerage. I'd appreciate advice on a cleaner method to do this. Perhaps a custom BrokerageSetupHandler?

Brokerages/Bitfinex/BaseMessage.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/e0ef2a953dfe6978a7bbb4fdf0fabcf8944f29ad)

Common/Securities/BitfinexSecurityPortfolioManager.cs
mchandschuh: After looking over the code again it appears you can implement your own `IMarginModel` and override the `GenerateMarginCallOrder` method. In this method, you could perform your special, security specific margin call behavior. Also, it looks like the way this method was implemented that it will only ever return a warning and not actually perform a real margin call. The margin call warning is a feature of the infrastructure to indicate to the algorithm that the portfolio is running out of remaining margin and is not directly overrideable. The pieces that are overrideable are the generation of margin call orders and the actual calculation of maintenance and initial margin requirements. If we can't get bitfinex to fit into the existing model then we'll need to modify the design to support the bitfinex case, but it won't be through subclassing of the `SecurityPortolioManager` as this type is not intended or designed to be subclassed and/or extended.

Brokerages/Bitfinex/BaseMessage.cs
mchandschuh: It looks like the keys are always constant indexes, so we could just define constants that represent each of the keys. For example:
```C#
public class TickerMessage
{
    private static class Key
    {
        public const int ChannelID = 0;
    }
}
```
Or, I would even prefer that each message parses the keys internally so that the knowledge of keys is unknown to the outside world. It kind of appears like this is what's happening, but it's still going through this indexing system. So it looks like the indexing system is only used internally by sub-classes in order to parse the input value array. I would prefer us to document the structure of the message and then use direct array look-ups since in the scope of the `TickerMessage` constructor it's expected to know use nitty gritty message contract details while parsing the message. But if you prefer to not use the indexes directly, we can instead define the constants.

My main issue with the indexing system is that is seems like an unnecessary layer of complexity and indirection in what is otherwise a very simple type hierarchy.

Brokerages/Bitfinex/BitfinexBrokerage.cs
mchandschuh: We can also add an entry to `Currencies.CurrencySymbols` dictionary, a mapping for BTC to Ƀ

Algorithm.CSharp/BaseBitcoin.cs
jameschch: I think people may run into problems without overriding the vanilla margin calls. Margin funding is cheap so lot of people are fully leveraged for extended periods of time. Having the engine automatically reduce a position due to maxed-out leverage would not be popular.

Brokerages/Bitfinex/BitfinexBrokerageFactory.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/c1f338168665f57a89072a2906c6e37f4b3802c0)

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/4f52cf5631f85cdf20cc5a4915b5a87c59cdab38)

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/ca4e9bfca39ab54614cff2dd439c77914c67d792)

Algorithm.CSharp/BaseBitcoin.cs
mchandschuh: The margin call behavior is already plugable/extensible via the `IMarginModel.GenerateMarginCallOrder` - so users can define how margin calls happen. By default, we use a system modeled off of interactive broker's margin call behavior. If BTC has different margin rules then we should add a dedicated `IMarginModel` implementation, but defining a new portfolio type is not the place to do this as the type wasn't designed to be extended.

Common/Securities/BitfinexSecurityPortfolioManager.cs
jameschch: I did try an alternative implementation with an override of the margincallmodel, and this provided a way to prevent the automated margin call liquidations generated by the portfolio. Without going into the details here, the margin call warnings coming from the ScanForMarginCall I don't think are relevant for Bitfinex and I'm also not sure it will be possible to automate margin call liquidations. I will commit the IMarginModel derived class and see if you think it's preferable.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/68c41cb3d90462d014791de7f4f34b5e7b5325df)

I agree with you on this. I notice that there are 66 calls to Thread.Sleep currently in the solution. I wouldn't mind helping in a migration to task based synchronization on a wider scale if you think it worthwhile.

mchandschuh: Many (hopefully all!) of the `Thread.Sleep` calls in the solution are on dedicated threads.

Common/Securities/BitfinexSecurityPortfolioManager.cs
jameschch: [proposed fix](https://github.com/jameschch/Lean/commit/574fa1825c86abee5fd8f57275066b94eb0d3719)

I'm not finding an IMarginModel for some reason, but have derived MargInCallModel. It would be advantageous to be able to override the ScanForMarginCall behavior in the MarginCallModel so that custom margin warnings could be implemented also. If you're happy with this approach I will also add unit tests.

My doubts about the margin call behavior were my biggest concern when submitting the PR so thanks for all you help,

mchandschuh: My apologies, the interface is called `ISecurityMarginModel` and can be found [here](https://github.com/QuantConnect/Lean/blob/master/Common/Securities/ISecurityMarginModel.cs).

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/14172b60a1670413553f00e5350f17defa7d252e)

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: That's interesting. Will have to look at how this has been done.

Brokerages/Bitfinex/BaseMessage.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/7ea7dce558f1a4410ecb884d690ba5ec8ded9ef6)

It does make more sense to declare int's for the indexes. I think the only advantage of key arrays is readability.

Common/Securities/BitfinexSecurityPortfolioManager.cs
jameschch: [proposed fix](https://github.com/jameschch/Lean/commit/430f0396c571e0a4d318da10801ffa116c5023b3)

Thanks. I have moved the functionality into a derivation of SecurityMarginModel. I could not find a simple way to inject this without changes to the Forex class. Is there a better approach?

Unit tests are included but I will be alpha testing this shortly.

Algorithm.CSharp/BaseBitcoin.cs
jameschch: I've now modified the BaseBitcoin class [here](https://github.com/jameschch/Lean/commit/430f0396c571e0a4d318da10801ffa116c5023b3).

There are still some required initialization steps that might be better as changes to core classes. I'm sure your views on these kind of changes will be more informed than mine.

Algorithm.CSharp/BaseBitcoin.cs
jameschch: I have this mostly complete at this point. It does occur to me that it might be simpler if the order provider could detect when a cross order was necessary and simply split one order into two. This would unify the cross order logic across brokerages and remove the need to cache and wait for multiple response from the broker to fill an order. To achieve something like this, I was hoping I could modify the orders collection via IOrderProvider but it appears this is intended as read only.

jameschch: This is now addressed:

[here](https://github.com/jameschch/Lean/commit/b7ecc7394a0e0677cbf446e546f7e6c3d5d554bc) and [here](https://github.com/jameschch/Lean/commit/84c742b914ccc5732181fe1de522cecda241e4da)

These are sensitive changes so will likely require some further testing.

jameschch:
I've checked in the last of the currently requested changes (I think). Thanks for all your help.

Due to constructor changes, you may notice that paper trading config is broken. Will get this back up and running fairly soon.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: [fixed](https://github.com/QuantConnect/Lean/pull/310/commits/0cf0f33a7c79289f6fc7321958c2fc321ad7817d)

mchandschuh:
This PR has gotten massive. Can we break out a separate one that has only the `IBrokerage` implementation when you feel it is complete and working as expected. It's much easier for me to review, test and validate smaller pieces in isolation.

jameschch:
No problem. I found much the same thing developing: it's a little too big for me to catch everything myself. The paper brokerage is there now and I have merged fixes for all known defects. I'll create a PR for just the changes to Brokerages assembly and dependencies as a start.

jameschch:
This PR has been split/superceded

cemdev:
What has happened to this? I note the bitfinex brokerage is not in the main lean github repo?

jameschch:
Currently testing a mostly complete PR that integrates gdax. 

I have prodution ready branches that support bitfinex and okcoin if interested.

cemdev:
*super interested* - but all this was from over a year ago, is it never getting merged into the main branch?

cemdev:
bitfinex is the one i'm primarily interested in, would love to see what you have. the only branch i found (i admit i did not look super hard) was a year old.

espirulina:
Its in James fork
https://github.com/jameschch/Lean/tree/master/Brokerages/Bitfinex

cemdev:
well, this is excellent, i was expecting to have to implement the api for the data to just run backtesting - a full on brokerage is really excellent. going to get started with this this weekend, many thanks

jaredbroad:
@cemdev we'll have the top most popular integrations in master over the next few months. @jameschch is kind enough to be leading the GDAX support.

PR #311, opened 2016-03-03
Bitcoin tick data downloader from CryptoIQ

jameschch:
Simple tool to download Bitcoin tick data.


mchandschuh:
Thanks for splitting this up! Can we add the license banner to the tops of these files?

jameschch:
Sure. Will go back and recheck all my other files too.

mchandschuh:
Another thing, I just pushed some changes in the toolbox, so these will probably have conflicts. I can bring down your changes locally and merge them for you if you prefer.

jameschch:
I'm just shutting down so be my guest otherwise will look at it tomorrow.

jameschch:
merged

ToolBox/CryptoiqDownloader/CryptoiqBitcoin.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace QuantConnect.ToolBox.CryptoiqDownloader
+{
+
+    public class CryptoiqBitcoin
+    {
+
+        public DateTime time;
```
jaredbroad: Where possible please try follow MS convention with code submissions. Public variables/properties have capital letters. https://msdn.microsoft.com/en-us/library/ms229002(v=vs.110).aspx

mchandschuh: I think this is all lower-case for simplicity with JSON parsing. That being said, I think by default `JsonConvert` will first make a case-sensitive member comparison and then a case-insensitive member comparison, so it should work with the proper casing as well.

PR #313, opened 2016-03-03
Fix for History<Tick>

jameschch:
Revisited change that fixes calls to History<Tick> that were previously returning nothing.


mchandschuh:
Thanks, this is certainly an improvement over what's currently there.

PR #314, opened 2016-03-03
Quandl Bitfinex downloader

jameschch:
Dedicated tool for downloading quandl bitfinex daily data I found useful. It may be this duplicates existing quandl integration.

### DIFF
--- a/Common/Market.cs
+++ b/Common/Market.cs
@@ -35,7 +35,8 @@ namespace QuantConnect
             Tuple.Create(USA, 1),
             Tuple.Create(FXCM, 2),
             Tuple.Create(Oanda, 3),
-            Tuple.Create(Dukascopy, 4)
+            Tuple.Create(Dukascopy, 4),
+            Tuple.Create(Bitcoin, 5)
         };
 
         static Market()
@@ -67,6 +68,8 @@ namespace QuantConnect
         /// Dukascopy Market
         /// </summary>
         public const string Dukascopy = "dukascopy";
+
+        public const string Bitcoin = "bitcoin";
 
         /// <summary>
         /// Adds the specified market to the map of available markets with the specified identifier.
--- a/Common/Market.cs
+++ b/Common/Market.cs
@@ -36,7 +36,7 @@ namespace QuantConnect
             Tuple.Create(FXCM, 2),
             Tuple.Create(Oanda, 3),
             Tuple.Create(Dukascopy, 4),
-            Tuple.Create(Bitcoin, 5)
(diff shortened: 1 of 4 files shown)


Common/Market.cs
```diff
@@ -68,6 +69,8 @@ static Market()
         /// </summary>
         public const string Dukascopy = "dukascopy";
 
+        public const string Bitcoin = "bitcoin";
```
mchandschuh: I think the market he should be bitfinex, thoughts? This would make it analogous to forex where fxcm/oanda/dukascopy are all foreign exchange markets, much like bitfinex is a bitcoin market.

ToolBox/QuandlBitfinexDownloader/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+                Console.WriteLine("Usage: Downloader FROMDATE APIKEY");
+                Console.WriteLine("FROMDATE = yyyymmdd");
+                Environment.Exit(1);
+            }
+
+            try
+            {
+                // Load settings from config.json
+                var dataDirectory = Config.Get("data-directory", "../../../Data");
+                bool useDivisor = bool.Parse(Config.Get("bitfinex-use-divisor", "false"));
```
jaredbroad: I'd recommend putting divisor of X, where X =1 or 100 depending on what the user sets it to.

jameschch: Great idea. Will make same change to cryptoiq

jaredbroad: We have similar concepts in equity data -- we call it the "scale factor". Please use the same terminology to keep it consistent.

ToolBox/QuandlBitfinexDownloader/Program.cs
```diff
@@ -38,11 +38,11 @@ static void Main(string[] args)
             {
                 // Load settings from config.json
                 var dataDirectory = Config.Get("data-directory", "../../../Data");
-                bool useDivisor = bool.Parse(Config.Get("bitfinex-use-divisor", "false"));
+                int divisor = int.Parse(Config.Get("bitfinex-divisor", "100"));
```
mchandschuh: in the future, use the `Config.GetInt` [method](https://github.com/QuantConnect/Lean/blob/master/Configuration/Config.cs#L153) as it is a little cleaner and also accepts the default value. It also will log when default values are being used as an indication to the user that it is a configurable setting

PR #317, opened 2016-03-10
Fixes bug in SecurityMarginModel.GetMarginRemaining

StefanoRaggi:
when switching positions from long to short and viceversa

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -592,10 +592,13 @@ namespace QuantConnect.Algorithm
             // can't order it if we don't have data
             if (price == 0) return 0;
 
+            // if targeting zero, simply return the negative of the quantity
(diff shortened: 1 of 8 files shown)


mchandschuh:
Just for those following, @StefanoRaggi is going to write more tests to try and identify the exact issue here. We have hunches that the `CalculateOrderQuantity` may need to reference the security's margin remaining and use that as an upper bound. That being said, the insufficient funds errors in the current implementation occur on a flip from `SetHoldings(sym, 1)` to `SetHoldings(sym, -1)` reliably using the hour SPY data over 10/07/2013 to 10/11/2013. It also appears to be about ~150 dollars off from having enough margin to make the purchase. This may be related to fees, as setting the brokerage model to Tradier reduced the overall number of errors.

jaredbroad:
Nice work! Great to have this stuff properly under tests!

StefanoRaggi:
Closing this PR because it is replaced by #327

PR #318, opened 2016-03-11
Remove "algorithm-location" from config.json

AlexCatarino:
The algorithm location can be determined by "algorithm-language" key from config.json, therefore "algorithm-location" did not provide original information.
Also, this change prevents conflits between "algorithm-language" and "algorithm-location". For exemple, the user could set  "algorithm-language": "FSharp" and "algorithm-location": "QuantConnect.Algorithm.CSharp.dll"


mchandschuh:
For others who might stumble here, the `algorithm-location` configuration value is actively used by the community and is especially useful for keeping one's algorithm code separated from the LEAN engine code via different projects and/or solutions. This setting allows one to specify the exact dll to be ingested regardless of conventions.

PR #319, opened 2016-03-11
Adds LimitFillRegressionAlgorithm.py

AlexCatarino:
Add Python version of LimitFillRegressionAlgorithm


mchandschuh:
Are we going to add the regression test for this one as well?

AlexCatarino:
Yes. I am working on the algorithms that belong to the regression tests first.

mchandschuh:
Comment was in regards to us not adding an automated regression test for this new algorithm, but from your comment [here](https://github.com/QuantConnect/Lean/pull/320#issuecomment-195396287) it appears there's an issue with adding the automated test, so this is OK.

PR #320, opened 2016-03-11
Adds RegressionAlgorithm.py

AlexCatarino:
Adds Python version of RegressionAlgorithm


mchandschuh:
Why didn't we include the regression test for this one?

AlexCatarino:
We are not including regression tests for Python (and Java) yet because we need to compile the dll manually.

PR #322, opened 2016-03-11
Adds ParameterizedAlgorithm.py

AlexCatarino:
Adds Python version of ParameterizedAlgorithm
Creates a method in the python algorithm class to deal with paramenters.

### DIFF
--- a/Algorithm.Python/ParameterizedAlgorithm.py
+++ b/Algorithm.Python/ParameterizedAlgorithm.py
@@ -1,0 +1,75 @@
+﻿# QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+# Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+# 
+# Licensed under the Apache License, Version 2.0 (the "License"); 
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import clr
+clr.AddReference("System")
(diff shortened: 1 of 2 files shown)


Algorithm.Python/ParameterizedAlgorithm.py
```diff
@@ -0,0 +1,75 @@
...
+        self.Fast = None
+        self.Slow = None
+        self.FastPeriod = 100
+        self.SlowPeriod = 200
+        self.GetParameters()
+
+
+    def GetParameters(self):
+        '''Get parameters from config.json'''
+        import json
```
mchandschuh: There's a function called `IAlgorithm.SetParameters` (implemented in `QCAlgorithm` [here](https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.cs#L498) that saves off all of the parameters. Maybe what we need is to make the `QCAlgorithm` method `virtual` and override it in each `Python` algorithm to fetch the parameter values from the dictionary. We will probably also need a method to `GetParameters()` which returns an `IEnumerable<ParameterAttribute>` - currently we use reflection to get parameter attributes that were declaratively declared on the members. It seems it is possible to do .NET style attributes in Python, but may require a [little work](http://stackoverflow.com/a/33151912). We would also might need to have a common `QCPythonAlgorithm` base class to handle the population of the proper fields.

AlexCatarino: We found a [solution]( https://github.com/QuantConnect/Lean/blob/5b91a14685bb3ae092084a66049f848161de1d5f/Algorithm.Python/ParameterizedAlgorithm.py#L46) that calls `IAlgorithm.GetParameters` at `Initialize`.
We could put the .NET style attributes in Python in the Python TODO list together with the `QCPythonAlgorithm` base class.

mchandschuh:
Yes, the only issue there is the algorithm doesn't expose what the parameters are. That is a good first step though. Next step we need to allow another component to inspect a python algorithm in order to detect what parameters it has. This will allow us to display the parameters on the website. We can tackle this later.

AlexCatarino:
Ok! I saved the stackoverflow discussion you sent to later reference.

PR #323, opened 2016-03-12
Constants renamed to Globals.

naicigam:
Globals values can be reloaded from configuration.


jaredbroad:
Thanks for the PR @naicigam! For some reason its failing TravisCI because of a couple of the tests. If we can get it passing these we can merge: https://travis-ci.org/QuantConnect/Lean/builds/115477241#L3705

naicigam:
@jaredbroad Sorry about that. Just fixed it.

PR #324, opened 2016-03-13
Fixed bug in MarketHoursDatabaseJsonConverter

StefanoRaggi:
In the constructor, holiday dates were formatted using the default date format set in the system preferences, instead of always using the fixed US format (as in the Convert method).


jaredbroad:
Nice catch!

PR #327, opened 2016-03-15
SetHoldings/CalculateOrderQuantity bug fixes + new tests

StefanoRaggi:
Replaces #317

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -614,6 +614,8 @@ namespace QuantConnect.Algorithm
             // continue iterating while we do not have enough margin for the order
             decimal marginRequired;
             var orderQuantity = 0;
+            decimal orderValue;
+            decimal orderFees;
 
             do
             {
@@ -628,11 +630,15 @@ namespace QuantConnect.Algorithm
                     orderQuantity--;
                 }
 
-                // calculate the margin required for the order
+                // generate the order
                 var order = new MarketOrder(security.Symbol, orderQuantity, UtcTime);
+                orderValue = order.GetValue(security);
+                orderFees = security.FeeModel.GetOrderFee(security, order);
+
+                // calculate the margin required for the order
                 marginRequired = security.MarginModel.GetInitialMarginRequiredForOrder(security, order);
 
-            } while (marginRequired > marginRemaining || marginRequired > targetOrderValue);
+            } while (orderQuantity > 0 && (marginRequired > marginRemaining || orderValue + orderFees > targetOrderValue));
 
             // add directionality back in
             return (direction == OrderDirection.Sell ? -1 : 1) * orderQuantity;
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -592,6 +592,9 @@ namespace QuantConnect.Algorithm
             // can't order it if we don't have data
             if (price == 0) return 0;
 
+            // if targeting zero, simply return the negative of the quantity
+            if (target == 0) return -security.Holdings.Quantity;
+
             // this is the value in dollars that we want our holdings to have
             var targetPortfolioValue = target*Portfolio.TotalPortfolioValue;
             var quantity = security.Holdings.Quantity;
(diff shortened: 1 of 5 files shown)


Tests/Algorithm/AlgorithmSetHoldingsTests.cs
```diff
@@ -0,0 +1,220 @@
...
+                orderQuantity = algorithm.CalculateOrderQuantity(_symbol, targetPercentage);
+                order = new MarketOrder(_symbol, orderQuantity, DateTime.UtcNow);
+                freeMargin = algorithm.Portfolio.GetMarginRemaining(_symbol, orderDirection);
+                requiredMargin = security.MarginModel.GetInitialMarginRequiredForOrder(security, order);
+
+                Console.WriteLine("Current price: " + security.Price);
+                Console.WriteLine("Target percentage: " + targetPercentage);
+                Console.WriteLine("Order direction: " + orderDirection);
+                Console.WriteLine("Order quantity: " + orderQuantity);
+                Console.WriteLine("Free margin: " + freeMargin);
```
mchandschuh: we may want to comment out this output, when running the automated tests it will spam the logs on travis

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -604,36 +607,32 @@ public int CalculateOrderQuantity(Symbol symbol, decimal target)
...
+                marginRequired = security.MarginModel.GetInitialMarginRequiredForOrder(security, order);
 
-                // we need to add the fee in as well, even though it's not value, it's still a cost for the transaction
-                // and we need to account for it to be sure we can make the trade produced by this method, imagine
-                // set holdings 100% with 1x leverage, but a high fee structure, it quickly becomes necessary to include
-                // otherwise the result of this function will be inactionable.
-
-                iterations++;
-            }
+            } while (marginRequired > marginRemaining || marginRequired > targetOrderValue);
```
mchandschuh: I'm curious as to how many times this iterates for common cases, since it looks like it's decrementing one at a time instead of estimating how many to decrement, so for large order fees (fees that are large compared to price of single share) I could see how this could loop hundreds of extra times.

Also, is there a case (specifically an exotic fee structure) where the initial estimate of `orderQuantity` will be less than the desired result? I'd like to see how this does with a progressive fee structure, maybe something like, less than 1k shares is $1 per share, but more than 1k shares is 1c per share. I know this is highly unlikely, but the iteration logic should be able to handle it. Before it was based on newton rhapson root finding, but this only decrements, may need to increment as well to be robust. Thoughts?

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -604,36 +607,38 @@ public int CalculateOrderQuantity(Symbol symbol, decimal target)
...
+            decimal orderValue;
+            decimal orderFees;
 
-                orderQuantity += deltaQuantity;
+            do
+            {
+                if (orderQuantity == 0)
+                {
+                    // compute the initial order quantity
+                    orderQuantity = (int)(targetOrderValue / unitPrice);
```
mchandschuh: it seems as though this is an initialization step and could be performed before the do loop is started.

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -604,36 +607,38 @@ public int CalculateOrderQuantity(Symbol symbol, decimal target)
...
-                // recompute order fees
+                // generate the order
                 var order = new MarketOrder(security.Symbol, orderQuantity, UtcTime);
-                var fee = security.FeeModel.GetOrderFee(security, order);
+                orderValue = order.GetValue(security);
+                orderFees = security.FeeModel.GetOrderFee(security, order);
 
-                orderValue = Math.Abs(order.GetValue(security)) + fee;
+                // calculate the margin required for the order
+                marginRequired = security.MarginModel.GetInitialMarginRequiredForOrder(security, order);
```
mchandschuh: I certainly like how we've brought this back into the 'margin' domain, I think it makes it easier to grock

Tests/Algorithm/AlgorithmTradingTests.cs
```diff
@@ -91,7 +102,7 @@ public void SetHoldings_ZeroToShort(decimal leverage)
         public void SetHoldings_ZeroToShort_SmallConstantFeeStructure(decimal leverage)
         {
             Security msft;
-            var algo = GetAlgorithm(out msft, 1, 1);
+            var algo = GetAlgorithm(out msft, leverage, 1);
```
mchandschuh: it's nice that these tests are set up using `TestCaseSource` but the way they're set up implies I can add random test cases and they will work. I don't think this is the case however. If we're going to use `TestCaseSource` then the test case needs to include the answer or the test method should manually compute it using the inputs

mchandschuh:
I'm tempted to just pull in the negative order fee fix directly for now, while we work through some of these other comments: f22224db5fc020bd25ab678c77c5bb67d11ab9dd

Tests/Algorithm/AlgorithmSetHoldingsTests.cs
StefanoRaggi: I will comment these in the next commit (they were so useful for debugging that I forgot them in)

Algorithm/QCAlgorithm.Trading.cs
StefanoRaggi: I'll move this out of the loop, although it would execute the division only the first time

Tests/Algorithm/AlgorithmTradingTests.cs
StefanoRaggi: I added the multiple leverage test cases, because in previous versions of the code I got different results with different leverages. In these test methods, answers are manually computed and checked, they just happen to (correctly) give the same answer for different values of the leverage input value. I didn't see an easier way to avoid proliferation of test methods. :)

Algorithm/QCAlgorithm.Trading.cs
StefanoRaggi: You are right, the number of iterations for the MSFT tests with 10K constant fee is 401 :(
One possible solution could be decrementing by the orderFee/price ratio instead of by one.
With large constant fees this should give the correct solution at the 2nd iteration, still not sure about the non-linear case you mentioned (more tests needed).

One question: why is the initial value of `orderQuantity` an "estimate" ? Isn't it always equal to the upper-bound, which is also the correct value in the absence of fees ?

mchandschuh: Thinking about it further, I think you're correct about the initial value of `orderQuantity`. It should be the greatest quantity possible (absent of fees) under the upper bound of the iteration. If we can improve the iteration step then I think we'll be solid on this.

Tests/Algorithm/AlgorithmTradingTests.cs
mchandschuh: Ah, you're completely correct again! The SetHoldings function is unlevered, so 1x is 1 times total portfolio value, so the leverage should only impact the required margin for the order, but the order should be the same. Perfect!

StefanoRaggi:
@mchandschuh About pulling in the order fee fix alone, I'm afraid the frequency of "Insufficient margin" errors would be increased. I think it should be merged along with the new `CalculateOrderQuantity`.

PR #333, opened 2016-03-16
Adds PriceCurrency in Order and OrderEvent classes

StefanoRaggi:


Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -820,11 +818,16 @@ private void HandleOrderEvent(OrderEvent fill)
             {
                 Log.Debug("BrokerageTransactionHandler.HandleOrderEvent(): " + fill);
                 Interlocked.Exchange(ref _lastFillTimeTicks, DateTime.Now.Ticks);
+
+                // set the price currency on the fill and on the order
+                var security = _algorithm.Securities[fill.Symbol];
+                fill.FillPriceCurrency = security.SymbolProperties.QuoteCurrency;
+                order.PriceCurrency = fill.FillPriceCurrency;
```
mchandschuh: This makes sense here since it's in a single place but, a piece of me thinks that the brokerages should be populating this field when they produce the fill.

StefanoRaggi: I thought about this as well, but looking at the brokerage code, I thought making changes to all brokerages at once was a bit too risky.
Perhaps for now we could set the fill currency (in the above code) only if empty (i.e. not set by the brokerage). This would allow us to update and test the brokerages independently.
Thoughts ?

mchandschuh: I agree with there being lots of changes if we were to update the brokerages to do this. Maybe the best way to enforce this is by making the currency a ctor parameter on the OrderEvent class. This will minimally force the concern everywhere it is constructed.

I am very hesitant to check if it's not set and then set it. I'd prefer us go one way or the other, either we always set it here and it is assumed that the brokerage populates the fills using the correct currency as specified on the security object, or the brokerage populates the order event and the transaction handler merely passes the data through the system. My preference here would be the latter since that's how things are currently handled. In most cases the currency should be easily available. For example, tradier will always produce USD denominated fills. I believe IB transmits the currency information via the contract details. Oanda/Fxcm I'm not as familiar with, but could use the `ISecurityProvider` to get access to the correct currency for the security.  I think the bigger question is what do we do if we encounter a mismatch, for example, a brokerage impl sends a fill that's quoted in currency A, but the security is configured to use currency B... that I do not currency have an answer for :)

StefanoRaggi: We also need to populate the PriceCurrency field in the Order class (needed for the serialization of trades). So should both fields be populated inside the brokerage fill events ?
Also, the OrderEvent constructor has an Order parameter, so maybe it's best to set the currency in the Order before passing it to the OrderEvent constructor ?
About the question of different currencies in brokerage vs. config, maybe log a warning and continue ?

mchandschuh: I think that's a good plan. We'll ensure that all orders are tagged with a currency when we create the order and if we find a mismatch we can `Log.Error` it and continue. For now we can assume the brokerage populates the order event with the price in proper currency units and use the order event ctor to set the currency from the order object

PR #334, opened 2016-03-18
Changed time buffer for MarketOnClose order submit from 11 to 16 minutes

StefanoRaggi:


jaredbroad:
Thanks Stefano. For others reading this -- the NYSE changed the market on close order time to 15 minutes prior to close.

Issue #331, opened 2016-03-16
Make indicators work with Renko bars


lithast:
A large number of the indicators derive from TradeBarIndicator (IndicatorBase<TradeBar>) even though they can just as easily be applied to RenkoBar. The common properties required are Open, High, Low, Close and in some cases Volume. Time is provided on BaseData.

I have explored 3 options so far to achieve this:

1. Create a common Interface across both of them with the required properties. This is complicated and I think would break backwards compatibility because IndicatorBase<T> expects T : BaseData. It would be possible to change this to expect IBaseData but the scope of that change is pretty large and has implications for compatibility (i.e methods on QCAlgorithim).

2. Put an implicit conversion on RenkoBar to convert it to trade bar. Really not desirable because of the  extra object allocations, plus the Period property would be an issue. I consider this a fall-back option in my local working copy if all else fails.

3. Derive TradeBar and RenkoBar from a common base class that has the required properties as virtual and inherits from BaseData. TradeBar and RenkoBar override the properties with their own implementation where needed. I was not really happy with this implementation but this looked to be the most workable. There is a mostly implemented fork of this at https://github.com/lithast/renko-indicator-support. There is one unit test that I still need to look into, a bit of duplication cleanup, some XML doc missing and the base class could really use a better name :)

Before I go any further with it wanted to get some opinions and see if you guys see any better options and what you thought on this one.

StefanoRaggi:
@lithast I think there may be a fourth option:

4. Derive `RenkoBar` from `TradeBar` (removing `Open`, `High`, `Low`, `Close`, `Volume` and `End `properties from `RenkoBar`). This way all indicators deriving from `TradeBarIndicator `can be used with `RenkoBar`.

lithast:
Ah yep. I looked into that approach and discounted it because I thought you guys would not go for it. The potential issues I ran into are:

1. TradeBar has a public Period property that does not apply to RenkoBar and would always have the default value of 1minute, which doesn't do anything. I could make that virtual and override it in RenkoBar with a NotSupportedException but that seemed less than ideal.

2. The public static ParseXyz methods on TradeBar all take T where T : TradeBar and the internals would not apply to the derived RenkoBar. I.e

` var tradeBar = new T
            {
                Period = config.Increment,
                Symbol = config.Symbol
            };`

Would throw if the Period above threw a NSE.

[...]

PR #343, opened 2016-03-21
Indicator support for renko bars

lithast:
Addresses #331 

Added a common base class called VolumeBar for both TradeBar and RenkoBar so that all indicators previously designed for TradeBar are now available for RenkoBar. 

All tests apart from the IchimokuKinkoHyoTests are passing for me and those also fail on a clean checkout of the QuantConnect master source. I think these failures are related to my development environment not the changes.


mchandschuh:
Looking at the size of this change and it's potential to break user code scares me :/  I think a safer implementation would be to add a `long Volume { get; set; }` to the `IBar` interface and have `RenkoBar` implement `IBar`.  We may need to make considerations for `QuoteBar` and the `Bar` implementation and what volume means in those cases.  Maybe a more generic term than volume can be used.

Also, it looks like the travis-ci build is [failing](https://travis-ci.org/QuantConnect/Lean/builds/117435516#L1752) for this PR.

lithast:
I looked into the build breakage at the time and it's very strange. It's failing to compile with some of the overloaded RegsiterIndicator methods on QCAlgorithim. My local environment is running VS2013 on .NET 4.5 and it compiles with all tests passing. From the build logs it looks like the CI is using mono to compile and my best guess is that it has some issues with that overload case. I don't have access to a mono environment to test it out I'm afraid.

Not sure if you saw my last comment on #331, but the TL;DR version is that an interface won't work here because `IndicatorBase<T>` expects a concrete `BaseData` instance as it's type parameter and changing that to `IBaseData` so `IBar : IBaseData` (required as an interface can't inherit from a class) is an ever bigger change. 

The other approach of:
```
public class BarIndicator<T> : IndicatorBase<T>
   where T : BaseData, IBar, new()
```
Would need two definitions of every indicator one `ExampleIndicatorRenko : BarIndicator<RenkoBar>` and one `ExampleIndicatorTradeBar : BarIndicator<TradeBar>`. I guess it would be possible to to put the indicator logic into an abstract base class and have the two above inherit from it while keeping the derived TradeBar version with the same signature to maintain compatibility but having to define everything twice and reference the Renko indicator by another class seems less than ideal. If that is acceptable though I can look into doing it. 

Otherwise I'm out of ideas for this one, can't see another way with the current architecture for doing this. Might have to revert back to my original hack of putting an implicit conversion on RenkoBar -> TradeBar in my local codebase and accept the slight performance and memory hit.

PR #348, opened 2016-03-22
Adds QuandlImporterAlgorithm.py

AlexCatarino:
Adds Python version of QuandlImporterAlgorithm

### DIFF
--- a/Algorithm.Python/QuandlImporterAlgorithm.py
+++ b/Algorithm.Python/QuandlImporterAlgorithm.py
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from datetime import date
+from datetime import date, timedelta
 
 from clr import AddReference
 AddReference("System")
@@ -40,11 +40,11 @@ class QuandlImporterAlgorithm(QCAlgorithm):
     def Initialize(self):
         '''Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.'''
 
-        today = date.today()
+        yesterday = date.today() - timedelta(1)
 
-        self.SetStartDate(2013,1,1)                          #Set Start Date
-        self.SetEndDate(today.year, today.month, today.day)  #Set End Date
-        self.SetCash(25000)                                  #Set Strategy Cash
+        self.SetStartDate(2013,1,1)                                      #Set Start Date
+        self.SetEndDate(yesterday.year, yesterday.month, yesterday.day)  #Set End Date
+        self.SetCash(25000)                                              #Set Strategy Cash
         self.AddData[Quandl](self.__quandlCode, Resolution.Daily)
         self.__sma = self.SMA(self.__quandlCode, 14)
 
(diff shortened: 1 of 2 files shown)


Algorithm.Python/QuandlImporterAlgorithm.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def Initialize(self):
+        '''Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.'''
+
+        today = date.today()
+
+        self.SetStartDate(2013,1,1)                          #Set Start Date
+        self.SetEndDate(today.year, today.month, today.day)  #Set End Date
+        self.SetCash(25000)                                  #Set Strategy Cash
+        self.AddData[Quandl](self.__quandlCode, Resolution.Daily)
```
mchandschuh: great to see iron python generics working!

Algorithm.Python/QuandlImporterAlgorithm.py
```diff
@@ -0,0 +1,61 @@
...
+    __sma = None
+    __quandlCode = "YAHOO/INDEX_SPY"
+
+    def Initialize(self):
+        '''Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.'''
+
+        today = date.today()
+
+        self.SetStartDate(2013,1,1)                          #Set Start Date
+        self.SetEndDate(today.year, today.month, today.day)  #Set End Date
```
mchandschuh: the original algorithm subtracts a day from the current day. Ideally we would be comparing the outputs of these python algorithms with the c# ones, even if we don't have the regression system set up to do this for us, it is expected that as you develop these you're verifying that the outputs match the c# example algorithms. If we find deviations, then it's something we must fix in the engine and/or python implementations.

Algorithm.Python/QuandlImporterAlgorithm.py
```diff
@@ -0,0 +1,61 @@
...
+class QuandlImporterAlgorithm(QCAlgorithm):
+    '''QuantConnect University: Generic Quandl Data Importer
+    Using the underlying dynamic data class "Quandl" we take care of the data 
+    importing and definition for you. Simply point QuantConnect to the Quandl Short Code.
+     
+    The Quandl object has properties which match the spreadsheet headers.
+    If you have multiple quandl streams look at data.Symbol to distinguish them.'''
+    
+    __sma = None
+    __quandlCode = "YAHOO/INDEX_SPY"
```
mchandschuh: are these akin to static variables in c#? Ideally we wouldn't make examples using static values in algorithms as it may cause issues for people trying to run algorithms in parallel.

Algorithm.Python/QuandlImporterAlgorithm.py
AlexCatarino: So far I could only make it work when the type is defined in C#. When I wrote a class in Python (see `QuandlFuturesAlgorithm` for example), it doesn't work.

Algorithm.Python/QuandlImporterAlgorithm.py
AlexCatarino: Yep, these variables are static.
I didn't know it matters. I will continue to define them inside a `__init__(self)` method to keep them non-static.

PR #351, opened 2016-03-23
Adds AbsolutePriceOscillator indicator (APO)

StefanoRaggi:


mchandschuh:
This looks exactly like the `MovingAverageConvergenceDivergence` indicator just without a signal line.  Also, I tend to prefer to expose the internal indicators as public properties.

StefanoRaggi:
Correct, APO is equal to our MACD (without the signal line). Probably TA-lib made it a different indicator because the original MACD did not have the MovingAverageType parameter and always used EMA. They do also have a MACDEXT though, that has two MovingAverageType parameters (one for the fast MA and one for the slow MA).
I'll derive APO from MACD.

PR #352, opened 2016-03-23
Adds PercentagePriceOscillator indicator (PPO)

StefanoRaggi:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -960,6 +960,24 @@ public MidPrice MIDPRICE(Symbol symbol, int period, Resolution? resolution = nul
...
+        /// Creates a new PercentagePriceOscillator indicator.
+        /// </summary>
+        /// <param name="symbol">The symbol whose PPO we want</param>
+        /// <param name="fastPeriod">The fast moving average period</param>
+        /// <param name="slowPeriod">The slow moving average period</param>
+        /// <param name="movingAverageType">The type of moving average to use</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value)</param>
+        /// <returns>The PercentagePriceOscillator indicator for the requested symbol over the specified period</returns>
+        public PercentagePriceOscillator APO(Symbol symbol, int fastPeriod, int slowPeriod, MovingAverageType movingAverageType, Resolution? resolution = null, Func<BaseData, decimal> selector = null)
```
mchandschuh: I think this should be `PPO`, correct?

mchandschuh:
This also looks just like the `MovingAverageConvergenceDivergence` indicator, just scaled to a percentage scale, maybe it could derive from the MACD... or if we decide to leave the `APO` then maybe it derives from the `APO` and then performs the scaling

Algorithm/QCAlgorithm.Indicators.cs
StefanoRaggi: Correct, it's a typo

StefanoRaggi:
Ok, I'll make it derive from APO.

Indicators/PercentagePriceOscillator.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IndicatorDataPoint input)
+        {
+            var value = base.ComputeNextValue(input);
+
+            return (IsReady && Slow != 0) ? 100 * value / Slow : 0m;
```
jaredbroad: This could be "return (Slow != 0) ? 100 * value / Slow : 0m;" to be in inline with other indicators. This way it always returns a value even if not ready and as it approaches "ready" the value will approach the real value.

StefanoRaggi: @jaredbroad Thanks, corrected in PR https://github.com/QuantConnect/Lean/pull/355

PR #355, opened 2016-03-24
PercentagePriceOscillator now returns a value during warmup

StefanoRaggi:


Indicators/PercentagePriceOscillator.cs
```diff
@@ -54,7 +54,7 @@ protected override decimal ComputeNextValue(IndicatorDataPoint input)
         {
             var value = base.ComputeNextValue(input);
 
-            return (IsReady && Slow != 0) ? 100 * value / Slow : 0m;
+            return Slow != 0 ? 100 * value / Slow : 0m;
```
jaredbroad: Do you think 100 is the infinity case with the PPO? Maybe when slow is zero it should return 100?

StefanoRaggi: The numerator is just the difference of two moving averages (can be positive or negative, small or large). In theory, this oscillator is unbounded, but it oscillates around the value of zero. I think the case of a zero denominator should be zero (as done in the TA-lib source code).

mchandschuh: Good point with the directionality

PR #357, opened 2016-03-25
Adds bitfinex brokerage

jameschch:
Contains core bitfinex brokerage integration, required dependencies and unit tests.


Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
```diff
@@ -0,0 +1,164 @@
...
+            decimal remainingAmount;
+            decimal executedAmount;
+            if (response.IsCancelled)
+            {
+                return OrderStatus.Canceled;
+            }
+            else if (decimal.TryParse(response.RemainingAmount, out remainingAmount) && remainingAmount > 0
+                && decimal.TryParse(response.ExecutedAmount, out executedAmount) && executedAmount > 0)
+            {
+                return OrderStatus.PartiallyFilled;
```
mchandschuh: should we check if `remainingAmount == 0` here and return `OrderStatus.Filled`?

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// </summary>
+        /// <param name="orderType"></param>
+        /// <returns></returns>
+        public string MapOrderType(OrderType orderType)
+        {
+            var result = _orderTypeMap.Where(o => o.Wallet == Wallet && o.OrderType == orderType);
+
+            if (result != null & result.Count() == 1)
+            {
+                return result.Single().BitfinexOrderType;
```
mchandschuh: This is just for extra information, no change required as the impact here is extremely minimal., but typically you don't want to enumerate an `IEnumerable` more than once. The call to `result.Count()` will enumerate the entire enumerable and then `result.Single()` will re-enumerate it, albeit at most the first two entries. As I said, in this case it's not going to impact much , if anything at all. Here's an [example](http://twistedoakstudios.com/blog/Post7694_achieving-exponential-slowdown-by-enumerating-twice) of where it could really matter!

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <param name="msg"></param>
+        /// <returns></returns>
+        public OrderStatus MapOrderStatus(TradeMessage msg)
+        {
+            if (Math.Abs(msg.FEE) != 0)
+            {
+                var cached = this.CachedOrderIDs.Where(c => c.Value.BrokerId.Contains(msg.TRD_ORD_ID.ToString())).FirstOrDefault();
+                if (cached.Value != null)
+                {
+                    if (msg.TRD_ORD_ID == cached.Value.BrokerId.Select(b => int.Parse(b)).Max(b => b))
```
mchandschuh: Calls to `Enumerable.Max(...)` can throw exceptions if there is nothing in the enumerable, so in this case, if `cached.Value.BrokerId` is empty it could throw, also, if you're using `Max(b => b)` you should be able to just write it as `Max()`

In this case, I think `cached.Value.BrokerId` is guaranteed to be populated before it makes its way into the cache. If it is not guaranteed, you can get around the exception by using the `DefaultIfEmpty` method like the following:
```c#
enumerable.DefaultIfEmpty(42).Max();
```
This will find the max value in the enumerable, but in the event that the enumerable is empty then it will return the default value of 42.

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+        public override bool IsConnected
+        {
+            get { return this._tickerToken != null && !this._tickerToken.IsCancellationRequested; }
+        }
+
+        private decimal GetPrice(Order order)
+        {
+            if (order is StopMarketOrder)
+            {
+                return ((StopMarketOrder)order).StopPrice * ScaleFactor;
```
mchandschuh: this scale factor still scares me. 

@jaredbroad: should we refactor `Order.Quantity` and `OrderEvent.Quantity` to support decimal before bringing this in, or are we good to bring this in and should just remember to remove the `ScaleFactor` later?

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+
+        /// <summary>
+        /// Place a new order
+        /// </summary>
+        /// <param name="order"></param>
+        /// <returns></returns>
+        public override bool PlaceOrder(Orders.Order order)
+        {
+            //todo: wait for callback from auth before posting
+            Authenticate();
```
mchandschuh: is this todo required for proper operation? do we need to authenticate before every call or just before the first call? It looks like the implementation in this file is blank.

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+
+            int quantity = (int)Math.Floor(SecurityProvider.GetHoldingsQuantity(order.Symbol));
+            Orders.Order crossOrder = null;
+            if (OrderCrossesZero(order, quantity))
+            {
+                crossOrder = order.Clone();
+                var firstOrderQuantity = -quantity;
```
mchandschuh: I'm a little confused as to why we're doing the zero crossing logic here, but then in the `private bool PlaceOrder` below we immediately recombine them into a single order.  Some brokerages require orders to be split up and some even require the order that brings the holdings to zero to fill before submitting the next order, but it doesn't look like that's required here.. or is it?

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+
+            bool hasFaulted = false;
+            foreach (string id in order.BrokerId)
+            {
+                var post = new BitfinexCancelReplacePost
+                {
+                    Amount = (order.Quantity / ScaleFactor).ToString(),
+                    CancelOrderId = int.Parse(id),
+                    Symbol = order.Symbol.Value,
+                    Price = order.Price <= 0 ? order.Id.ToString() : (order.Price * ScaleFactor).ToString(),
```
mchandschuh: I think we may need to consider the cross orders here. Assuming I submitted an order that crosses zero and then try to update that order, it looks like it will update each broker id order to be the same (the new order)... I think we'll have to re-do the order crossing detection... thoughts?

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+                            OriginalAmount = decimal.Parse(item.OriginalAmount) * ScaleFactor,
+                            RemainingAmount = decimal.Parse(item.RemainingAmount) * ScaleFactor,
+                            ExecutedAmount = decimal.Parse(item.ExecutedAmount) * ScaleFactor
+                        });
+                    }
+                }
+            }
+            catch (Exception)
+            {
+                throw;
```
mchandschuh: why have a try/catch here if we're just going to immediately throw?

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+            {
+                var response = _restClient.GetActiveOrders();
+                if (response != null)
+                {
+                    foreach (var item in response)
+                    {
+                        list.Add(new BitfinexOrder
+                        {
+                            Quantity = Convert.ToInt32(decimal.Parse(item.OriginalAmount) * ScaleFactor),
+                            BrokerId = new List<string> { item.Id.ToString() },
```
mchandschuh: Are all bitfinex orders market orders? I would have expected them to minimally also support limit orders.

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+        /// <returns></returns>
+        public override List<Holding> GetAccountHoldings()
+        {
+
+            var list = new List<Holding>();
+
+            var response = _restClient.GetActivePositions();
+            foreach (var item in response)
+            {
+                var ticker = _restClient.GetPublicTicker(TradingApi.ModelObjects.BtcInfo.PairTypeEnum.btcusd, TradingApi.ModelObjects.BtcInfo.BitfinexUnauthenicatedCallsEnum.pubticker);
```
mchandschuh: is BTCUSD the only pair offered by bitfinex? if so it looks like this `ticker` value can be moved outside of the loop. If not, then we may need to fetch different tickers for each position

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+                if (item.Type == Wallet)
+                {
+                    if (item.Currency == "usd")
+                    {
+                        list.Add(new Securities.Cash(item.Currency, item.Amount, 1));
+                    }
+                    else
+                    {
+                        var ticker = _restClient.GetPublicTicker(TradingApi.ModelObjects.BtcInfo.PairTypeEnum.btcusd, TradingApi.ModelObjects.BtcInfo.BitfinexUnauthenicatedCallsEnum.pubticker);
+                        list.Add(new Securities.Cash("BTC", item.Amount * ScaleFactor, decimal.Parse(ticker.Mid) / ScaleFactor));
```
mchandschuh: Instead of explicitly stating 'BTC', maybe using `item.Currency` is more appropriate. Also, similar comment as above, can we resolve the `ticker` using `item.Currency` instead of hard coding it for 'BTCUSD'?

Brokerages/Bitfinex/BitfinexBrokerageFactory.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="algorithm"></param>
+        /// <returns></returns>
+        public override Interfaces.IBrokerage CreateBrokerage(Packets.LiveNodePacket job, Interfaces.IAlgorithm algorithm)
+        {
+
+            string apiSecret = Config.Get("bitfinex-api-secret");
+            string apiKey = Config.Get("bitfinex-api-key");
+            string wallet = Config.Get("bitfinex-wallet");
+            //it's desirable to throw an exception here when failing parse
+            decimal scaleFactor = decimal.Parse(Config.Get("bitfinex-scale-factor", "1"));
```
mchandschuh: All of these config values need to come from the `job.BrokerageData` dictionary. You can move these config gets into the `BrokerageData` property above. Have a look at the [InteractiveBrokersBrokerageFactory](https://github.com/QuantConnect/Lean/blob/master/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerageFactory.cs#L55) for an example.

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,449 @@
...
+        /// </summary>
+        /// <returns></returns>
+        public override List<Order> GetOpenOrders()
+        {
+
+            var list = this.GetOpenBitfinexOrders().Select(o => (Order)o).ToList();
+
+            foreach (var item in list)
+            {
+                if (item.Status != OrderStatus.Canceled && item.Status != OrderStatus.Filled && item.Status != OrderStatus.Invalid)
```
mchandschuh: There's a handy [extension method](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/OrderExtensions.cs#L40) to test if an order status is open.

Brokerages/Bitfinex/BitfinexBrokerageFactory.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <summary>
+        /// provides brokerage connection data
+        /// </summary>
+        public override Dictionary<string, string> BrokerageData
+        {
+            get
+            {
+                return new Dictionary<string, string>
+                {
+                };
```
mchandschuh: You can initialize required parameters from configuration here. They'll end up in the job packet's brokerage data dictionary

Brokerages/Bitfinex/BitfinexOrder.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace QuantConnect.Brokerages.Bitfinex
+{
+
+    /// <summary>
+    /// Bitfinex order object with additional fields
+    /// </summary>
+    public class BitfinexOrder : MarketOrder
```
mchandschuh: Only support for market orders? Also, there may be unexpected side effects of subclassing like this... we may want to make sure we're not doing explicit type checking (hope not!) but also that the brokerage implementation doesn't mutate the values of these orders if the reference is being shared with other parts of the engine... this could lead to 'spooky action at a distance' kind of bugs, very hard to track down.

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
```diff
@@ -0,0 +1,215 @@
...
+            lock (Ticks)
+            {
+                Ticks.Add(new Tick
+                {
+                    AskPrice = msg.ASK / ScaleFactor,
+                    BidPrice = msg.BID / ScaleFactor,
+                    AskSize = (long)Math.Round(msg.ASK_SIZE * ScaleFactor, 0),
+                    BidSize = (long)Math.Round(msg.BID_SIZE * ScaleFactor, 0),
+                    Time = DateTime.UtcNow,
+                    Value = msg.LAST_PRICE / ScaleFactor,
```
mchandschuh: If we label the `Tick` instance using `TickType.Quote` then I think the value should be specified as the midpoint of the bid/ask. I think what this showcases is the need for a `TickType.Both` or something similar, and if we did go that route, we would want a dedicated `LastPrice` property (which there is but it's an alias for `Value`).

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <summary>
+        /// Map exchange order type
+        /// </summary>
+        /// <param name="orderType"></param>
+        /// <returns></returns>
+        public OrderType MapOrderType(string orderType)
+        {
+            var result = _orderTypeMap.Where(o => o.Wallet == Wallet && o.BitfinexOrderType == orderType);
+
+            if (result != null & result.Count() == 1)
```
StefanoRaggi: The single `&` is a non short circuiting operator and as such the null check would not avoid the `.Count()` call.
It should be replaced with the normal `&&` operator.

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
jameschch: Possibly. The Implemenatation will currently return Filled status when the order fee is returned, and this has proven to be reliable. I see no reason why an update with the fee could not be subsequent to zeroing the RemainingAmount where there are multiple trade executed messages, but I think it unlikely or impossible that the reverse may happen.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: No, the work indicated by the todo: is not required, but is desirable. There is a possible race condition where the auth may succeed after the order attempt. I have not encountered this yet, but is worthwhile addressing sometime.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: I had not considered the implications of cross zero here. This is a cancel and replace (rather than an update) so the replacing order needs cross zero detection also.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: Was helpful in debugging to have a break inside the catch rather than higher up the stack.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: There should be full support for Limit, Market and Stop orders. Unit Tests pass for all these order types.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: I have a branch (with all planned work completed) ready for another PR with support for all Bitfinex currency pairs.

It makes sense to pull the ticker price once for all positions, but it would mean iterating the holdings twice, once to find out which tickers are needed and then again to deserialize. Would you prefer to leave as if for the time being?

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: Yes, the currency is hard coded in this PR. There will be another PR supporting all currencies to follow on afterwards. Limiting the scope to the main currency pair is intended to reduces the time required for review and testing.

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
jameschch: Absolutely right. That's a typo.

[fixed](https://github.com/jameschch/Lean/commit/fd3d798956e6e6c0398476fb6f504203db2e4e10)

Brokerages/Bitfinex/BitfinexOrder.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/54b284c19145d05d50bd2b548f4b1421972b4e34)
You're right this class is no longer necessary. It has been unit tested but deserves some exchange testing and maybe more coverage

Brokerages/Bitfinex/BitfinexBrokerageFactory.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/4888ec82312f1b800c2baaa3641b6faec8dd0cc0)

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/947b92b18c45599b4b4a81fe6206b8cfeafe75fb)

Brokerages/Bitfinex/BitfinexBrokerageFactory.cs
jameschch: This was addressed elsewhere

jameschch:
All that currently remains is the cross zero on Update (and some live testing).

Once this PR is merged, I will add further changes with Ether and Litecoin support. We might then begin thinking about the switch which would give the option of routing trades to BitStamp.

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
jameschch: We are looking for the Lean order Id that corresponds to a broker id that accompanies a trade fill. If these don't states don't align, I think the Max should throw an exception.

I've removed the redundancy in the Max call [here](https://github.com/jameschch/Lean/commit/ab66b36e0077eb6e5aad6feb06eb2c03c4c554e2)

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: [fixed](https://github.com/jameschch/Lean/commit/79cbff62abc72d072eeb33e0bea5799ee993653f)

Brokerages/Bitfinex/BitfinexBrokerage.cs
jameschch: I notice the cfd support relies on something like a scale factor at security level so realize the question is larger than just this brokerage. The level of unit test coverage I've provided should be sufficient to instill confidence about making future changes.

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
jameschch: Not sure the action to take here. I guess it would be possible to emit two ticks for each type, but this might lead to unexpected algorithm behavior. The core changes required to do the refactor would be very sensitive.

Brokerages/Bitfinex/BitfinexBrokerage.cs
jaredbroad: Agree ideally need decimals to properly support bitcoin assets. Scale factors for CFD's are a different concept (1 oil contract is for 100 barrels of oil)

jameschch: Yes, the scaleFactor also fills me with dread. I can imagine insulating against regression problems at several levels will be a challenge with the quantity change. I may be able to help with adding some unit test coverage.

Brokerages/Bitfinex/BitfinexWebsocketsBrokerage.Messaging.cs
jameschch: I've looked into this further and really do agree that retaining last trade and volume data is basically a must-have. I can add clauses that will map fields based on the Market.Bitfinex which I hope will be a lower impact change than an additional TickType. This will likely be big enough to justify it's own PR.

jameschch: [fixed](https://github.com/jameschch/Lean/commit/fe2ba2e89efff95209a4a2bd56b7363aba00536f)

jameschch: The tick in question is a quote with a convenient last price but a cumulative volume, so it's not really a TickType.Both. There is a separate subscription for trade price and volume which is emitted as a separate message. Support for TickType.Trade can be provided in the future.

Brokerages/Bitfinex/BitfinexBrokerage.Util.cs
jameschch: The fill logic has been changed substantially to correctly handle multiple split fills.

jaredbroad:
Closing for now until we merge in order decimalization; then can pull in a BTC brokerage with fractional order support.

jameschch:
Yes the build has changed significantly since this was submitted. Quite apart from the fact that the future of bitfinex remains unclear.

PR #360, opened 2016-03-27
Null pointer check added

mblouin02:
Added a forgotten null pointer check after bumping into an exception.


jaredbroad:
Thanks for the PR @mblouin02!

PR #374, opened 2016-04-08
Upgrade Baseclass Contrib dependency for compatibility with VS2015

kaffeebrauer:


UserInterface/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Baseclass.Contrib.Nuget.Output" version="2.0.0" targetFramework="net45" />
+  <package id="Baseclass.Contrib.Nuget.Output" version="2.1.0" targetFramework="net45" />
```
mchandschuh: does this fix the build issue we were seeing in vs2015? I believe we saw errors from an MSBuild.Tasks dll.

kaffeebrauer: Hi Michael, according to the version upgrade from 2.0 to 2.1 the original version of 2.0 did not handle the situation where the MS Build version was 14 (used by VS2015), now in version 2.1 it does.

https://www.nuget.org/packages/Baseclass.Contrib.Nuget.Output/

Release Notes
- Added support for VS2015 RC

jaredbroad:
Not required anymore as used the Mono.WebBrowser classes

PR #378, opened 2016-04-12
Desktop charting for LEAN

jaredbroad:
Added form for local charting via packet streaming.


Launcher/Program.cs
```diff
@@ -125,5 +135,24 @@ static void Main(string[] args)
...
+        /// Form launcher method for thread.
+        /// </summary>
+        static void LaunchUX(LeanEngineSystemHandlers system, LeanEngineAlgorithmHandlers algorithm, AlgorithmNodePacket job)
+        {
+            var form = new Views.WinForms.LeanWinForm(system, algorithm, job);
+
+            // Pipe console log through to the UX
+            Log.LogHandler = new CompositeLogHandler(new ILogHandler[] { new ConsoleLogHandler(), 
+                new FunctionalLogHandler(
+                    x => system.Notify.Send(new DebugPacket(job.ProjectId, job.AlgorithmId, job.CompileId, x)),
```
mchandschuh: Doesn't this end up with duplicate messages being handled?

Launcher/QuantConnect.Lean.Launcher.csproj
```diff
@@ -34,6 +36,14 @@
     <StartupObject>QuantConnect.Lean.Launcher.Program</StartupObject>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Geckofx-Core, Version=45.0.6.0, Culture=neutral, PublicKeyToken=3209ac31600d1857, processorArchitecture=x86">
+      <HintPath>..\packages\GeckoFX.1.0.5\lib\Geckofx-Core.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Geckofx-Winforms, Version=45.0.6.0, Culture=neutral, PublicKeyToken=3209ac31600d1857, processorArchitecture=MSIL">
```
mchandschuh: it's a shame these dependencies ended up being added to the launcher project

Logging/QuantConnect.Logging.csproj
```diff
@@ -70,6 +70,7 @@
     <Reference Include="System.ComponentModel.Composition" />
     <Reference Include="System.Core" />
     <Reference Include="System.Numerics" />
+    <Reference Include="System.Windows.Forms" />
```
mchandschuh: Why does the logging project needs Windows.Forms?

Messaging/EventMessagingHandler.cs
```diff
@@ -0,0 +1,229 @@
...
+        public event RuntimeErrorEventRaised RuntimeErrorEvent;
+
+        public delegate void HandledErrorEventRaised(HandledErrorPacket packet);
+        public event HandledErrorEventRaised HandledErrorEvent;
+
+        public delegate void BacktestResultEventRaised(BacktestResultPacket packet);
+        public event BacktestResultEventRaised BacktestResultEvent;
+
+        public delegate void ConsumerReadyEventRaised();
+        public event ConsumerReadyEventRaised ConsumerReadyEvent;
```
mchandschuh: I envisioned this all being one event handler of type packet... I think that type of pattern will lead to less maintenance

Messaging/EventMessagingHandler.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        /// <summary>
+        /// Raise a backtest result event safely.
+        /// </summary>
+        protected virtual void OnBacktestResultEvent(BacktestResultPacket packet)
+        {
+            if (BacktestResultEvent != null)
+            {
+                BacktestResultEvent(packet);
```
mchandschuh: It's a threading concern to not assign the event handler to a local variable, the event handler may become null between the null check and the invocation

http://stackoverflow.com/questions/5147397/why-assign-a-handler-to-an-event-before-calling-it

UserInterface/WinForms/LeanWinForm.Extensions.cs
```diff
@@ -23,16 +25,35 @@ namespace QuantConnect.Views.WinForms
...
-            box.SelectionLength = 0;
+            box.SafeInvoke(b =>
+            {
+                box.SelectionStart = box.TextLength;
+                b.SelectionLength = 0;
+                box.SelectionColor = color;
+                box.AppendText(text + Environment.NewLine);
+                box.SelectionColor = box.ForeColor;
+                box.ScrollToCaret();
+                return 0;
```
mchandschuh: If this code needs to be thread safe (in order to prevent color bleed) then we'll need to lock it

UserInterface/WinForms/LeanWinForm.cs
```diff
@@ -0,0 +1,156 @@
...
+            _systemHandlers = systemHandlers;
+            _algorithmHandlers = algorithmHandlers;
+            _messaging = (EventMessagingHandler)systemHandlers.Notify;
+
+            //Create the browser control
+            _browser = new GeckoWebBrowser { Dock = DockStyle.Fill, Name = "browser" };
+            _browser.DOMContentLoaded += BrowserOnDomContentLoaded;
+            splitPanel.Panel1.Controls.Add(_browser);
+
+            var url = string.Format("https://beta.quantconnect.com/terminal/embedded?user={0}&token={1}&bid={2}&pid={3}&version={4}",
```
mchandschuh: should this be beta?

Issue #75, opened 2015-04-30
Add a WeightedAverage Indicator


mchandschuh:
Check out the discussion on the [LWMA issue](https://github.com/QuantConnect/Lean/issues/74) about weighted indicators and how it could be composed. This would allow things like VWAP to be easily defined.

jaredbroad:
To clarify: this means an indicator building block which takes multiple components and finds the weighted average of them using supplied weights.

```C#
var closeSPY = Identity("SPY", Field.Close);
var volumeSPY = Identity("SPY", Field.Volume);
// e.g. 14 Period VWAP
var vwap = closeSPY.WeightedBy(volumeSPY, 14); 
```

This should be an extension method which returns another indicator.

jaredbroad:
Completed by commit 32770af

PR #384, opened 2016-04-21
Adds WeightedBy indicator extension

AlexCatarino:
First, adds WindowIdentity indicator we need to implement the indicator extension
Second, adds the WeightedBy indicator extension itself and its test
Reference issue: #75

### DIFF
--- a/Indicators/IndicatorExtensions.cs
+++ b/Indicators/IndicatorExtensions.cs
@@ -16,6 +16,8 @@
 using System;
 using System.Globalization;
 using QuantConnect.Data;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace QuantConnect.Indicators
 {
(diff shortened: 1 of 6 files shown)


Indicators/IndicatorExtensions.cs
```diff
@@ -62,6 +64,44 @@ public static bool Update(this IndicatorBase<IndicatorDataPoint> indicator, Date
...
+        {
+            var x = new WindowIdentity(period);
+            var y = new WindowIdentity(period);
+            var numerator = new Sum("Sum_xy", period);
+            var denominator = new Sum("Sum_y", period);
+
+            value.Updated += (sender, consolidated) =>
+            {
+                x.Update(consolidated);
+                if (x.Samples == y.Samples)
```
mchandschuh: Comparing sample counts may cause issues if the `value` and `weight` indicators have different resolutions

AlexCatarino: If we passing `IndicatorDataPoint` via `Update(new IndicatorDataPoint(time, value)` how could we tell `WeightedBy` method the `IndicatorDataPoint` resolution?
I think that the issues will probably be solved with a input resolution check before `WeightedBy`is called.

AlexCatarino: We could, alternatively, only update the numerator when `x.Current.EndTime` equals `y.Current.EndTime`. Would that be better? It won't solve the resolution, but it would accept data with the same resolution but that are ready with different samples.

PR #388, opened 2016-04-22
Fixes Beta (PortfolioStatistics) calculation

AlexCatarino:
In CreateBenchmarkDifference method, a list of values of banchmark daily performance were not created properly. It added an extra zero entry in the beggining of the list and did not include the last day. That caused an one-day lag between the algorithm and benchmark performances.
Reference: [Strange backtester beta results](https://www.quantconnect.com/forum/discussion/1118)

Changes RegressionTests to reflect this correction.


mchandschuh:
Great stuff!

Issue #373, opened 2016-04-08
Custom Data Sourcing Emitting Erradically


jaredbroad:
Quandl sources are emitting unpredictably when there are multiple streams together.

Potentially FastForwardEnumerator's check against 'past data' doesn't include the equal case? So that 1 1/2 hours might be a function of the 30min max interval he refreshes at, and then its just emitting the same point over and over again. The enumerator stack is here (https://github.com/QuantConnect/Lean/blob/options-new/Engine/DataFeeds/LiveTradingDataFeed.cs#L401) and then the FastForwardEnumerator code here (https://github.com/QuantConnect/Lean/blob/options-new/Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs#L71)

PR #389, opened 2016-04-22
Prevents reread/repeated data from Quandl to be enqueued

AlexCatarino:
#373

### DIFF
--- a/Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs
@@ -70,6 +70,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
                 // make sure we never emit past data
                 if (_current != null && _current.EndTime > _enumerator.Current.EndTime)
                 {
+                    Logging.Log.Trace("FastForwardEnumerator.MoveNext(): Continuing because  of current end time");
                     continue;
                 }
 
--- a/Engine/DataFeeds/LiveTradingDataFeed.cs
+++ b/Engine/DataFeeds/LiveTradingDataFeed.cs
@@ -406,8 +406,12 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                     var enqueable = new EnqueueableEnumerator<BaseData>();
                     _customExchange.SetDataHandler(config.Symbol, data =>
                     {
-                        enqueable.Enqueue(data);
-                        if (subscription != null) subscription.RealtimePrice = data.Value;
+                        // Only enqueue new data if there is none (LastEnqueued == null) or if newer 
+                        if (enqueable.LastEnqueued == null || data.EndTime > enqueable.LastEnqueued.EndTime)
+                        {
+                            enqueable.Enqueue(data);
+                            if (subscription != null) subscription.RealtimePrice = data.Value;
+                        }
                     });
                     enumerator = enqueable;
                 }


Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs
```diff
@@ -70,6 +70,7 @@ public bool MoveNext()
                 // make sure we never emit past data
                 if (_current != null && _current.EndTime > _enumerator.Current.EndTime)
                 {
+                    Logging.Log.Trace("FastForwardEnumerator.MoveNext(): Continuing because  of current end time");
```
mchandschuh: This may end up flooding the live logs

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -409,8 +409,12 @@ protected Subscription CreateSubscription(Universe universe, Security security,
                     var enqueable = new EnqueueableEnumerator<BaseData>();
                     _customExchange.SetDataHandler(config.Symbol, data =>
                     {
-                        enqueable.Enqueue(data);
-                        if (subscription != null) subscription.RealtimePrice = data.Value;
+                        // Only enqueue new data if there is none (LastEnqueued == null) or if newer 
+                        if (enqueable.LastEnqueued == null || data.EndTime > enqueable.LastEnqueued.EndTime)
+                        {
```
mchandschuh: This may not be the best place for filtering logic, this handler is intended to pipe data out of a data exchange. I also worry that needing to add this here just hides another bug somewhere else.

I think  we could get similar behavior if we modified to this [line](https://github.com/AlexCatarino/Lean/blob/quandl-issue373/Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs#L71) in `FastForwardEnumerator` to filter out the equals case as well (or maybe allow it as a bool option, tick could allow the equals case, all others should not).

jaredbroad: If custom data emits multiple points for the same moment in time will this hide those other points? Agree we need to trace higher to see why its emitting more often than it should

AlexCatarino: We put that `Log.Trace` after that [line](https://github.com/AlexCatarino/Lean/blob/quandl-issue373/Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs#L71), because that if-condition was never met since `_current` had been always `null`. If `_current` on that line is always `null`, there is no other place other then this point we can check for new data.
I can run this branch in my machine for one or two days straight to confirm what I've observed so far.

AlexCatarino:
I traced higher to found who and why `_current` is set to `null`.

From `FastFowardEnumerator` back to [`enqueable.Enqueue(data);`](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/LiveTradingDataFeed.cs#L412), `_current` and `_enumerator` are modified by `MoveNext()` methods from three classes: `FrontierAwareEnumerator`, `RefreshEnumerator` and `RateLimitEnumerator`.

`RateLimitEnumerator` only receives `_enumerator` from `RefreshEnumerator` and [pass it forward](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/Enumerators/RateLimitEnumerator.cs#L77).

When the enumerator has passed the end of the collection at `FastFowardEnumerator.MoveNext()`, `_current` becomes `null` on `FrontierAwareEnumerator.MoveNext()` and sends a [`false`](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/Enumerators/FrontierAwareEnumerator.cs#L87) signal to `RefreshEnumerator.MoveNext()` that, [consistently sets `_current` and `_enumerator` to `null`](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/Enumerators/RefreshEnumerator.cs#L64). This step is important, because `RefreshEnumerator.MoveNext()` only polls new data when [`_enumerator` is set to `null` previously](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/Enumerators/RefreshEnumerator.cs#L54). This the 1-hour (two times [`minimumTimeBetweenCalls`](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/LiveTradingDataFeed.cs#L405) pattern.

AlexCatarino:
Does this [commit]( https://github.com/QuantConnect/Lean/commit/377bea18e45716d0fe89a7f157df8edf2ce8201) fix #373? 
If so we can close this PR.

jaredbroad:
@StefanoRaggi's commit almost fixes it; I'm hoping we can find a fix which qualifies uniqueness based on something other than time or doesn't require this uniqueness comparison. This would allow us to have custom data which has multiple points at the same instant of time (like how Ticks are rounded to the second)

Issue #387, opened 2016-04-22
PopulateDataDictionaries in TimeSlice is duplicating ticks


lithast:
I put some trip wires into my application to to catch issues I was having with out of order data merging live and historical ticks (unrelated). I noticed something else quite strange when I fixed the original issue. The trip wire was firing while loading historical data from the History Provider. I'm getting each tick in my data file repeated twice. Tracked it down to `PopulateDataDictionaries` in `TimeSlice`

![image](https://cloud.githubusercontent.com/assets/17696925/14741426/b4d1f124-08c8-11e6-82d0-0730d158c115.png)

`ticksList = new List<Tick> {(Tick) baseData};` 
Should this just create the list?

jaredbroad:
Great catch @lithast! Please submit a PR and we'll merge it into master.

jaredbroad:
+ An extra 10 bonus points if you submit a test with the PR :) /Tests/Engine/DataFeeds/TimeSliceTests.cs

lithast:
Well now, can't be leaving points on the table. I'll take a look at it tomorrow :)

jaredbroad:
Fixed by 36f3ef8

PR #390, opened 2016-04-23
Fixed an issue where TimeSlice was duplicating ticks

lithast:
Addresses #387 

Disclaimer: Last time I sent a pull request the build failed with what seemed to be some compiler differences between mono and native .Net. I'm not expecting issues here because the changes are simple, but in case those become famous last words here is a standard developer excuse:

image

### DIFF
--- a/Tests/Engine/DataFeeds/TimeSliceTests.cs
+++ b/Tests/Engine/DataFeeds/TimeSliceTests.cs
@@ -1,8 +1,22 @@
-﻿using System;
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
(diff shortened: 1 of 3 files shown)


Tests/Engine/DataFeeds/TimeSliceTests.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
```
mchandschuh: Can we add the license banner here?

jaredbroad:
Thanks for the great PR +10 bonus points @lithast :) Bonus points are tradable for free live trading subscriptions on QC. Send me your email (jared@quantconnect.com) and I'll send you a license.

lithast:
No problems Micheal, I have checked in those changes. Thanks Jared, I've flicked you a quick email.

PR #395, opened 2016-05-04
Bug Fix: Tick level data from FXCM needs to be converted to EST always

rbohac:


mchandschuh:
Does EST include DST?  If I recall correctly the market hours are fixed offset. https://github.com/QuantConnect/Lean/blob/master/Data/market-hours/market-hours-database.json

PR #396, opened 2016-05-04
FXCM Tick level timezone read from market-hours config file

rbohac:
FXCM tick level events time value is converted to EST via the setting in market-hours-database.json (currently UTC-05) rather than hard coding EST


rbohac:
I tested by setting my PC to several different timezones

UTC
EDT
Central
Hawaii

PR #401, opened 2016-05-11
FXCM Brokerage Dictionary index check fixed

rbohac:


Brokerages/Fxcm/FxcmBrokerage.Messaging.cs
```diff
@@ -387,9 +387,11 @@ private void OnExecutionReport(ExecutionReport message)
 
                         _isOrderSubmitRejected = true;
                     }
-
-                    _mapRequestsToAutoResetEvents[_currentRequest].Set();
-                    _mapRequestsToAutoResetEvents.Remove(_currentRequest);
+                    if (_mapRequestsToAutoResetEvents.ContainsKey(_currentRequest))
+                    {
```
StefanoRaggi: Maybe we could use `_mapRequestsToAutoResetEvents.TryGetValue` here, to have one less dictionary access ?

Brokerages/Fxcm/FxcmBrokerage.Messaging.cs
```diff
@@ -428,7 +430,7 @@ private void OnPositionReport(PositionReport message)
 
             if (message.getRequestID() == _currentRequest)
             {
-                if (message.isLastRptRequested())
+                if (message.isLastRptRequested() && _mapRequestsToAutoResetEvents.ContainsKey(_currentRequest))
```
StefanoRaggi: Same as above here

PR #403, opened 2016-05-11
FileSystemDataFeed.ScheduleEnumerator performance fix

StefanoRaggi:
On Windows, backtesting speed improvements ranged from 3x to 8x (especially with single symbol algorithms)

### DIFF
--- a/Engine/DataFeeds/FileSystemDataFeed.cs
+++ b/Engine/DataFeeds/FileSystemDataFeed.cs
@@ -148,7 +148,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
             // schedule the work on the controller
             var firstLoop = true;
             FuncParallelRunnerWorkItem workItem = null;
-            workItem = new FuncParallelRunnerWorkItem(() => enqueueable.Count < lowerThreshold, () =>
+            workItem = new FuncParallelRunnerWorkItem(() => true, () =>
             {
                 var count = 0;
                 while (enumerator.MoveNext())


mchandschuh:
You may want to run the 400 symbol benchmark as well. I believe  the count check was added to prevent memory overflow.

mchandschuh:
Also, consider what may happen if the algorithm thread takes a longer on each time step compared to the data feed.

StefanoRaggi:
Today I ran a test algorithm with 400 symbols from Jan 1998 to Mar 2016 (only AddEquity x 400, with an empty OnData method), the fixed version was marginally faster and memory usage was never more than double (max 300MB, it looks like the bounded BlockingCollection is working fine)

AddEquity 400 symbols (after perf fix):
20160512 13:55:54 Trace:: Debug: Algorithm Id:(BenchmarkTest) completed in 8336.65 seconds at 74k data points per second. Processing total of 616,083,755 data points.

AddEquity 400 symbols (before perf fix):
20160512 16:30:25 Trace:: Debug: Algorithm Id:(BenchmarkTest) completed in 8627.52 seconds at 71k data points per second. Processing total of 616,083,755 data points.

Tonight I'll repeat the test adding a Thread.Sleep(1) in OnData to slow down the algo thread and see how it behaves.

StefanoRaggi:
After some more tests I realized that `ScheduleEnumerator` is ok and this fix is not a good solution.

Issue #330, opened 2016-03-15
Make Order Quantity Fractional


jaredbroad:
Order quantity is currently an int. Change it to be a decimal. 

The change will be fairly extensive covering API methods and brokerage implementations.

jameschch:
I've looked into this and just changing the order quantity is not necessarily very problematic. There are a few explicit variable declarations but apart from this, everything ties together with just a few field types. There should not be a great deal of breaking changes to algorithms, although I think some will be unavoidable.

If this is followed through, there will also need to be change to Tick and Tradebar classes to support decimalised volumes.

StefanoRaggi:
Fixed in https://github.com/QuantConnect/Lean/pull/982

PR #404, opened 2016-05-11
Changes related to #330 order quantity decimalized

jameschch:
These are changes to all order classes that alter the type of Quantity from int to decimal. These are base class changes that would benefit from extensive testing. For such I would appreciate assistance from anyone who has relevant brokerage accounts to perform online testing.

~~I've gone for a brutalist approach that simply changes argument types from int to decimal and does not provide new overloads whilst marking existing methods obsolete. This means overloads that accepted a double quantity are removed and any existing algorithms that supply an int will have an implicit conversion to decimal. If it's preferred, I can soften the change by retaining obsolete overloads.~~

All unit tests currently pass however, some changes to results were required to account for rediscovered order quantity fragments.


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -46,19 +46,7 @@ public OrderTicket Buy(Symbol symbol, int quantity)
         /// Buy Stock (Alias of Order)
         /// </summary>
         /// <param name="symbol">string Symbol of the asset to trade</param>
-        /// <param name="quantity">double Quantity of the asset to trade</param>
-        /// <seealso cref="Buy(Symbol, int)"/>
-        public OrderTicket Buy(Symbol symbol, double quantity)
```
jaredbroad: Does this need to be removed?

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -117,26 +92,16 @@ public OrderTicket Sell(Symbol symbol, float quantity)
...
         public OrderTicket Sell(Symbol symbol, decimal quantity)
         {
-            return Order(symbol, Math.Abs(quantity) * -1);
-        }
-
-        /// <summary>
-        /// Issue an order/trade for asset: Alias wrapper for Order(string, int);
-        /// </summary>
-        /// <seealso cref="Order(Symbol, decimal)"/>
-        public OrderTicket Order(Symbol symbol, double quantity)
```
jaredbroad: Changes to API can break thousands of algorithms, ideally we don't remove these unless absolutely neccessary

Algorithm.CSharp/OpeningBreakoutAlgorithm.cs
```diff
@@ -288,7 +288,7 @@ private void ScanForEntrance()
 
             var allowedDollarLoss = MaximumPorfolioRiskPercentPerPosition * Portfolio.TotalPortfolioValue;
 
-            var shares = (int) (allowedDollarLoss/expectedCaptureRange);
+            var shares = allowedDollarLoss/expectedCaptureRange;
```
jaredbroad: Please leave this as int cast; equities (via the brokerage model "can place order") should throw/reject order on fractional trade requests.

jaredbroad:
Thank you for the PR @jameschch! Please try retain the older methods just to avoid breaking the QC algorithms. 

I think it would be best to include updates to the brokerage models in this PR; all the current supported brokerages cannot support fractional trading so orders which are fractional might break their API's e.g. [here](https://github.com/QuantConnect/Lean/blob/master/Common/Brokerages/TradierBrokerageModel.cs#L56) we could check order.Quantity != (int)order.Quantity return false. The paper trading brokerage should probably also do this, so its a default behavior.

Algorithm/QCAlgorithm.Trading.cs
jameschch: The problem with having both double and decimal overloads is you get:

> Error	231	The call is ambiguous between the following methods or properties

Removing the double overload results in an implicit conversion to decimal and no error.

jameschch:
I'm glad you've also considered the problem of algorithms having unwittingly used a fractional order method expecting the quantity to be floored. I think you're right that IBrokerage will need to change so that PlaceOrder and UpdateOrder have an explicit Floor and cast. I did want to make sure it's a safe assumption that none of the brokerages (apart from Bitfinex) need to support fractional order quantities at this time. I already have a bitfinex customized paper brokerage, so I don't see a problem with floored quantities being the default.

Algorithm/QCAlgorithm.Trading.cs
jameschch: Continuing from above, I could, add methods something like "OrderByDecimal(symbol, decimal)" and the overload problem goes away. However, I dislike the incongruity this introduces to the api.

jameschch: I've included overloads for double decimal and int in the latest commit. I've maintained float support but have not added additional methods. Have also added a dedicated regression test.

Algorithm.CSharp/OpeningBreakoutAlgorithm.cs
jameschch: No problem. I think we can see in this case the potential for problems with changing the return type of CalculateOrderQuantity. It might work to add another method something like 

` CalculateOrderQuantity<decimal>(Symbol, MaximumLeverage)`

jameschch: Think this is going to need to happen. Will also need:

`SetHoldings<decimal>(Symbol, percentage)`

Too keep consistency the non generic will return a floored quantity but the new generics will return a fractional quantity. Makes sense?

jameschch: Have balanced consistency and fractional support using generics in 7915b57. [This may not be the best approach](http://stackoverflow.com/a/8171547). I could instead add methods CalculateFractionalOrderQuantity() and SetFractionalHolding(). Any thoughts?

jameschch:
I think one other problem is demonstrated in the changes to CustomDataRegressionAlgorithm. With uses of the Order methods directly, backtest results will vary from paper/live due to lost decimal fragments. I think this may only be significant for large and possible very small denominations.

Common/Brokerages/DefaultBrokerageModel.cs
```diff
@@ -236,5 +242,22 @@ public virtual ISettlementModel GetSettlementModel(Security security, AccountTyp
             return new ImmediateSettlementModel();
         }
 
+        protected bool ValidateQuantityFloored(decimal quantity, out BrokerageMessageEvent message)
+        {
+            message = null;
+
+            // validate quantity floored
+            if (quantity != Math.Floor(quantity))
```
jaredbroad: Cool stuff! We might need ceiling for shorts (negative quantities), but since we're not checking buying power here maybe its if (quantity != floor && quantity != ceiling) to confirm that its a whole number. 

I think we can express that as "Math.Abs(quantity % 1) > 0"

Tests/RegressionTests.cs
```diff
@@ -150,21 +150,21 @@ private static TestCaseData[] GetRegressionTestParameters()
                 {"Total Trades", "1"},
                 {"Average Win", "0%"},
                 {"Average Loss", "0%"},
-                {"Compounding Annual Return", "155.210%"},
+                {"Compounding Annual Return", "191.086%"},
```
jaredbroad: Lets aim for getting this working without regression changes first to be safe. Its a fairly extensive change :) Test code might need updating to ensure they give rounded quantities.

Common/Brokerages/DefaultBrokerageModel.cs
jameschch: Yes, that's a better approach. I had also meant to add unit tests for the new method babbe81.

Tests/RegressionTests.cs
jameschch: Have reverted this change for now. I see what you are getting at: existing new order calls that supply a decimal will expect the implicit conversion but currently there is a behaviour change. I was hoping to avoid this, but I think it's necessary to add methods such as:

LimitOrderFractional(symbol, quantity, limitPrice)

...or to add overloads that cause retention of the decimal fragments:

LimitOrder(symbol, quantity, limitPrice, string tag = null, bool isQuantityFractional = false)

jameschch: Have opted for the overload option in 6a50816 and tried to harmonise with future symbolproperties changes. To minimize bloat of the API surface I have omitted some obvious method overloads but these can be added if preferred.

jaredbroad:
We recently added lot sizes to symbols and the SetHoldings now rounds to the nearest lot size for the security. Does this change this PR at all? This feels close to complete, -- if you can please squash some of the changes on the branch for a cleaner diff and we can do some "_final pre-flight checks_"

jameschch:
Yes I was following the other pr and was pleased to see it will simplify the scope of this.

jameschch:
I have made all the intended changes. Am getting odd results on some regression algorithms. Looking into this.

jameschch:
Can see tests are failing in upstream, so not due to this PR. If it'll help I can look at the failing tests anyway?

PR #406, opened 2016-05-15
Maintain original inner exception and log message

robcodes:
Simple change to keep the inner exception message and inner exception stack trace when logging errors in lean engine.

### DIFF
--- a/Engine/Engine.cs
+++ b/Engine/Engine.cs
@@ -245,11 +245,15 @@ namespace QuantConnect.Lean.Engine
                         // Algorithm runtime error:
                         if (algorithm.RunTimeError != null)
                         {
-                            throw algorithm.RunTimeError;
+                            throw new Exception("Algorithm.RunTimeError", algorithm.RunTimeError); //Keep original exception as inner exception to preserve stack trace
                         }
                     }
                     catch (Exception err)
                     {
+                        while (err != null && err.InnerException != null)
+                        {
+                            err = err.InnerException; //Find inner most exception as this will be most useful to algorithm designer
+                        }
                         //Error running the user algorithm: purge datafeed, send error messages, set algorithm status to failed.
                         Log.Error(err, "Breaking out of parent try catch:");
                         if (_algorithmHandlers.DataFeed != null) _algorithmHandlers.DataFeed.Exit();


Engine/Engine.cs
```diff
@@ -245,11 +245,15 @@ public void Run(AlgorithmNodePacket job, string assemblyPath)
...
                         {
-                            throw algorithm.RunTimeError;
+                            throw new Exception("Algorithm.RunTimeError", algorithm.RunTimeError); //Keep original exception as inner exception to preserve stack trace
                         }
                     }
                     catch (Exception err)
                     {
+                        while (err != null && err.InnerException != null)
+                        {
+                            err = err.InnerException; //Find inner most exception as this will be most useful to algorithm designer
```
mchandschuh: The logging should print out the entire stack, including inner exceptions, do it may be best to pass the full exception instance and let the logging component decide how to handle it

robcodes: Understand your point.
Could we make the below change though to prevent us from losing stack trace from the inner exception?
While re-throwing the exception it seems to lose the original stack trace unless u throw a new exception with the original exception as an inner exception.
`throw algorithm.RunTimeError;"`
to
`throw new Exception("Algorithm.RunTimeError", algorithm.RunTimeError)`

StefanoRaggi: I agree the exception should be thrown as @robcodes did, to avoid losing the inner exception (in the stack trace it will not be included though, but at least the information is in `err.InnerException`).
To recursively include **all** the inner exceptions in the logging, `err.ToString()` should be used instead of `err.StackTrace`, replacing:
`_algorithmHandlers.Results.RuntimeError(message, err.StackTrace);`
with
`_algorithmHandlers.Results.RuntimeError(message, err.ToString());`

The line below should probably be changed as well:
`_systemHandlers.Api.SetAlgorithmStatus(job.AlgorithmId, AlgorithmStatus.RuntimeError, message + " Stack Trace: " + err.StackTrace);`

jaredbroad: Thank you for the PR @robcodes. I think I agree with Stefano the [first part of PR](https://github.com/QuantConnect/Lean/pull/406/commits/ef6801e53afc5c0ff4416229a64b6c9ca789d890#diff-adae0a18f9996ae90d595c4ca363491aR248) is great but if possible can you remove the iterative loop and throw the ToString in the _algorithmHandlers.Results.RuntimeError()/_systemHandlers methods. This should give the user the whole InnerException chain.

jaredbroad:
Closing as stale; can resubmit with requested updates

mchandschuh:
For posterity, `Log.Error(Exception)` should handle this properly.

PR #412, opened 2016-05-24
Adds FRAMA indicator

jameschch:
I decided to try my hand at an indicator, so, here is the Fractal Adaptive Moving Average (FRAMA) created by John Ehlers. This is not to be confused with the FAMA or MAMA also created by Ehlers.

I'm not very familiar with the underlying theory so have not tested the results against a baseline. It might be a good idea to instead compare the results against an implementation known to be correct.


jaredbroad:
Thanks for the PR James! I reached out to John Ehlers and he'll hopefully send through a reference test dataset soon which we can use to validate the indicator properly.

jameschch:
You don't mean **THE** John Ehlers?

What I had in mind was running the code from another open source project and comparing results, but your approach is slightly more definitive.

jaredbroad:
*The* John Ehlers was kind enough to respond with the following data for the test framework :)
[FRAMA.txt](https://github.com/QuantConnect/Lean/files/282698/FRAMA.txt)
[FramaData.xlsx](https://github.com/QuantConnect/Lean/files/282701/FramaData.xlsx)

The date format is a little strange but roughly looks like xYYMMDD.00

jameschch:
On it.

jameschch:
I'm able to recreate the results with a variance of under 0.006 (due to rounding and recovered precision). However, the test data set contains some trailing results starting at date marked "1160525" in which one data point is repeated several times.

> 1160525	208.67	209.71	207.87	209.31	206.97	206.69
> 1160525	208.67	209.71	207.87	209.31	206.97	206.69
> 1160525	208.67	209.71	207.87	209.31	206.97	206.69

I'm fairly sure the computation stopped at this point, and these can be discarded.

The test data set assumes a hard coded long period target that I have parametrized, which we won't be able to test in the same way.

StefanoRaggi:
Nice job with this indicator, we should add it to the `MovingAverageType` enum:

  - Add a new enum value in MovingAverageType.cs
  - Add a case in the two `AsIndicator` methods in MovingAverageTypeExtensions.cs
  - Add two test lines in the `CreatesCorrectAveragingIndicator` method in MovingAverageTypeExtensionsTests.cs

jameschch:
@StefanoRaggi I was hoping to do this from the outset, however the test data made it clear that the indicator is intended to be based on the mid of a TradeBar rather than the close of an IndicatorDataPoint. I'm looking into what can be done here. What do you think?

StefanoRaggi:
@jameschch You're right: this moving average must be a `TradeBarIndicator` because it uses highs and lows in its calculations, while the `AsIndicator` extension method returns an `IndicatorBase<IndicatorDataPoint>` so we cannot add it to the `MovingAverageType` enum.

If in the future we will have more `TradeBarIndicator` moving averages, we could add an `AsTradeBarIndicator` extension method which would take a new `TradeBarMovingAverageType` enum and return a `TradeBarIndicator`.

jameschch:
That's a great idea. I imagine adding another entry point to the calculation of this average would still produce worthwhile results.

Tests/Indicators/FractalAdaptiveMovingAverageTests.cs
```diff
@@ -0,0 +1,78 @@
...
+    [TestFixture]
+    public class FractalAdaptiveMovingAverageTests
+    {
+        [Test]
+        public void ComputesCorrectly()
+        {
+            FractalAdaptiveMovingAverage frama = new FractalAdaptiveMovingAverage("", 6, 100);
+
+            decimal[] values = { 441m, 442m, 446m, 438m, 400m, 442m, 448m, 437m, 435m, 431m, 450m, 451m };
+            decimal[] expected = { 441m, 442m, 446m, 438m, 400m, 442m, 444.51m, 441.67m, 438.58m, 434.69m, 435.84m, 436.73m, 438m };
```
jaredbroad: Do you think this is needed with the new test data file? Does this test something different?

jameschch: It's not essential, but it does test different parameters so provides additional regression confidence. If you'd like to minimize the codebase it's a candidate for removal.

jaredbroad: Ok yes please remove then. If there's value in a second parameter set (i.e. tests a different code path) please submit another test-data-file in the same format as the other test sets (like frama.txt).

jameschch: This has been removed.

PR #413, opened 2016-05-27
Fixes universe subscription creation bug in LiveTradingDataFeed

StefanoRaggi:
*Occasionally when running EURUSD + Custom Data together the custom data would not get subscribed.*

1. LTDF.CreateUniverseSubscription is being called three times instead of twice, although only the first two of them are being added: https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/LiveTradingDataFeed.cs#L129
2. LTDF.CreateUniverseSubscription always adds the enumerator to the custom exchange even if it is a duplicate: https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/LiveTradingDataFeed.cs#L501
3. We always end up with two universe subscriptions, but the second one could have its enumerator replaced by the third "ghost" one.

My solution: check if existing first, then create and add (now it creates, then adds if not existing)
                               if (!_subscriptions.Contains(universe.Configuration))
                               {
                                   _subscriptions.TryAdd(CreateUniverseSubscription(universe, start, Time.EndOfTime));
                                   UpdateFillForwardResolution();
                               }

I tried many runs with this fix and was not able to repeat the problem.

### DIFF
--- a/Engine/DataFeeds/LiveTradingDataFeed.cs
+++ b/Engine/DataFeeds/LiveTradingDataFeed.cs
@@ -16,7 +16,6 @@
 
 using System;
 using System.Collections;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
@@ -24,7 +23,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using QuantConnect.Configuration;
 using QuantConnect.Data;
-using QuantConnect.Data.Auxiliary;
 using QuantConnect.Data.Custom;
 using QuantConnect.Data.Market;
 using QuantConnect.Data.UniverseSelection;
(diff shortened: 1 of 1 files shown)


mchandschuh:
Can something similar happen in backtesting?  I think this code is similar in both data feeds.

StefanoRaggi:
@mchandschuh I just ran a few backtests of the same algorithm and all symbols are correctly subscribed and firing events. 
With this code:

```
AddSecurity(SecurityType.Forex, "EURUSD", Resolution.Second);
AddData<Quandl>(_idx, Resolution.Daily);
AddData<Quandl>(_spy, Resolution.Daily);
```
one "unused" universe subscription is still being created as in the LiveTradingDataFeed before the fix. To prevent this, we could insert the following lines here:
https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/FileSystemDataFeed.cs#L304

```
if (_subscriptions.Contains(universe.Configuration))
{
      UpdateFillForwardResolution();
      return;
}
```

Maybe in the future this should be refactored (code is very similar).

PR #414, opened 2016-05-28
DailyFX+Json Reader+TimeSlice Create Update for Custom Data

jaredbroad:
Added a new custom data type DailyFX + JSON data format and SubscriptionDataSourceCollection adapter which breaks the Json array into individual objects. This is extendable so additional JSON structures can be decoded.

Daily FX data often had multiple events on the same moment in time so chance also tweaks Feed/Slice construction to fully emit all custom points even if they occur on the same moment in time.

OnData events get all the points but Slice API will only return the first for a given time. The alternative was (uglier) - adding 1 tick to each custom event which landed on the same timestep... or worse making the DailyFx object itself a collection... Given time we should find a way to get Slice.Get<T>() to return a list of events, maybe IEnumerable<T>?


Common/Data/Custom/DailyFx.cs
```diff
@@ -0,0 +1,323 @@
...
+    /// <remarks>
+    /// Data sourced by Thomson Reuters
+    /// DailyFX provides traders with an easy to use and customizable real-time calendar that updates automatically during 
+    /// announcements.Keep track of significant events that traders care about.As soon as event data is released, the DailyFX 
+    /// calendar automatically updates to provide traders with instantaneous information that they can use to formulate their trading decisions.
+    /// </remarks>
+    public class DailyFx : BaseData
+    {
+        //Private factory variables.
+        private Dictionary<string, DailyFx> _cache;
```
mchandschuh: In my car, so I don't have ReSharper, but looks like this guy is unused.

Common/Data/Custom/DailyFx.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "FALSE")]
+        Worse
+    }
+
+    /// <summary>
+    /// Helper to parse the Daily Fx API.
+    /// </summary>
+    public class DailyFxMeaningEnumConverter : JsonConverter
+    {
```
mchandschuh: Does the default StringEnumConverter not work?

Common/Data/SubscriptionDataSourceCollection.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    /// <remarks>
+    ///     Often data sources might return an array per request for efficiency e.g. a JSON request returning last 100 data points.
+    /// </remarks>
+    public class SubscriptionDataSourceCollection : SubscriptionDataSource
+    {
+        /// <summary>
+        /// Identifies where to get the subscription's data from
+        /// </summary>
+        public virtual IEnumerable<string> Explode(string source, SubscriptionDataConfig config, DateTime date, bool isLive)
```
mchandschuh: Sweet name!

mchandschuh:
Did we consider just returning a BaseDataCollection from the factory. Having a collection type derive from a single instance type seems misleading. Also, it looks like the collection tow only does json?  Ideally the source would be independent of the file format. 

Maybe we can have a quick call later today, I'm sure I'm missing something obvious, just want to understand the design here and help if  possible :)

Common/Data/Custom/DailyFx.cs
jaredbroad: Yeah strangely? Kept throwing errors and silently eating packets from the JSON. With the converter worked fine though. The test date 26th May had 40 events / GMT timing with one not specifying a time (https://www.dailyfx.com/calendar)

Engine/DataFeeds/JsonSubscriptionFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (source.TransportMedium != SubscriptionTransportMedium.Rest)
+            {
+                throw new NotImplementedException("Reading JSON from " + source.TransportMedium + " is not implemented.");
+            }   
+
+            // Convert to a collection string reader.
+            var collection = source as SubscriptionDataSourceCollection;
+            if (collection == null)
+            {
+                OnInvalidSource(source, new Exception("Json file format must be used with SubscriptionDataSourceCollection source."));
```
StefanoRaggi: Should we add `yield break;` here ?

Common/Data/Custom/DailyFx.cs
```diff
@@ -0,0 +1,319 @@
...
+        {
+            var start = date.ToString("yyyy", CultureInfo.InvariantCulture);
+            var end = start;
+
+            if (date.Month < 4)
+            {
+                start += "0101";
+                end += "03312359"; 
+            }
+            else if (date.Month < 6)
```
StefanoRaggi: Shouldn't be `date.Month < 7` ?

Common/Data/Custom/DailyFx.cs
```diff
@@ -0,0 +1,319 @@
...
+            {
+                start += "0401";
+                end += "06302359";
+            }
+            else if (date.Month < 9)
```
StefanoRaggi: Shouldn't be `date.Month < 10` ?

Engine/DataFeeds/SubscriptionDataReader.cs
```diff
@@ -266,7 +266,7 @@ public bool MoveNext()
                     // prevent emitting past data, this can happen when switching symbols on daily data
                     if (_previous != null && _config.Resolution != Resolution.Tick)
                     {
-                        if (_config.Resolution == Resolution.Tick)
+                        if (_config.Resolution == Resolution.Tick || _config.IsCustomData)
                         {
                             // allow duplicate times for tick data
                             if (instance.EndTime < _previous.EndTime) continue;
```
StefanoRaggi: This line was never executed before this change ?

jaredbroad:
I rationalize it by thinking its a way of reading the source, but if there's a way to use the BDC and keep it in the BaseData that's probably better too -- 

To brain storm - the alternative would be GetSource returning the URL for a collection like a normal SubscriptionDataSource(Transport.Rest, FileFormat.**Collection**?) -> Then have a CollectionReader which gets a BaseDataCollection back from the baseData.Reader() method, and yields each point individually? That nicely leaves all the array building to the BaseData object and keeps it in line with the other data types.

mchandschuh:
I like the idea of `FileFormat.Json` and a `JsonSubscriptionFactory`. I do not, however, like the idea of the `BaseData` instance defining how to be read in json. I've always viewed the `BaseData.Reader` method as a line by line reader, specifically only used by `BaseDataSubscriptionFactory`. I view this as the 'user-defined' file format where they can do as they please.

The `JsonSubscriptionFactory` (in my opinion) should not call into `BaseData.Reader`-esque methods, but rather should use the regular json attributes and infrastructure for handling the deserialization. This would greatly simplify this issue and immediately give json support to all data types, so long as the class is able to be deserialized. This also prevents adding more specialized methods to `BaseData`

If we do these things, then the `JsonSubscriptionFactory` can read its input stream as an `IEnumerable<T>` where `T` is the `BaseData` type.

In this configuration, I don't think we need a 'collection` source, but rather the individual factories define how a source is processed.

jaredbroad:
Elegant idea to use the JSON infrastructure. Not sure it covers a few cases; say when there's a root object? {"root":[array]} etc. We also need a little more control in this case (which is a good test case for other sources) the entries in the JSON might not actually be valid yet - e.g. In live trading they display the events for the whole day, including some which haven't happened yet, once the "actual" property has a value it yields the BaseData. Or scaling the properties like how we scale tradebars from millicents to dollars.

mchandschuh:
Great points, I hadn't considered scaling/projection of the data. These things could be handled via attributes, but I wouldn't recommend it, as it would all be custom. One thing I liked about my suggestion was being able to reuse the existing JSON attributes, without that the value degrades quickly.

Keeping the parsing/scaling/projection/ect in the `BaseData`, could we pass the entire JSON string to the `Reader` method, and the `Reader` would return a `BaseDataCollection`. In order to handle various object graphs, it may be required that we pass the entire object graph and not try to 'itemize' the JSON. So depending on the `BaseData.Reader` implementation and the string passed into it, it could return a collection or a single base data. This would also allow filtering, such as data elements that aren't valid yet.

Engine/DataFeeds/SubscriptionDataReader.cs
jaredbroad: True! Adjusted it to reflect real data flows

PR #416, opened 2016-05-31
Adds Swiss Army Knife Indicator

jameschch:
Continuing with Ehler's research papers, here is the Swiss Army Knife indicator. This is an amalgam of several algorithms Ehlers uses in various studies, for instance, the BandPass filter is the basis of the Empirical Mode decomposition.

His paper hints that tools may be used as either transformations or in a convergence/divergence pattern. Some of the tools lend well to chaining with the "Of" Method, others may be treated like moving averages.


jameschch:
At this stage I'd consider the implementation and tests to be complete, but perhaps not correct. The variance on the band pass results (0.043) is higher than expected, and since this is the main area of interest, this is a concern.

PR #418, opened 2016-06-03
Fixes on Lean OSX installation instructions

flatM:
Hi, 

Trying to install Lean on my macbook, I encountered one minor issue regarding VB project version (4.5 not supported on Mac OSX). The same sed command that would fix the issue under Linux did not work on OSX, due to the incompatibility between BSD sed (OSX) and GNU sed (Linux). I've revised the README to introduce the working version for OSX.

One more minor fix was also piggybacked into this PR: the file name of the launcher executable was also corrected.

FYI:
OS X: 10.11.5
Mono: MonoFramework-MDK-4.4.0.142.macos10.xamarin.universal.pkg
Xamarin Studio: 5.10.3

Thanks,
flatM

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -43,12 +43,15 @@ Install [Mono for Mac](http://www.mono-project.com/docs/getting-started/install/
 
 Install [MonoDevelop](http://www.monodevelop.com/download/) or [Xamarin Studio](http://xamarin.com/studio) for your IDE. If you use MonoDevelop also install its [FSharp Plugin](http://addins.monodevelop.com/Project/Index/48).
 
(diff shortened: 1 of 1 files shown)


jaredbroad:
What do you think about removing these from the build entirely @flatM? Users even report VB/F# issues in MS/Visual Studio sometimes.

flatM:
Yeah, I also had Lean installed on my Windows 7 desktop, and encountered an issue with F#. Couldn't recall what it was now. Could be the F# piece was not originally installed when I was installing VS2015 community.

I think it's your call if you'd like to keep VB/F# in Lean or not. IMHO, these issues with installations may have a negative impact over the potential increase of user base.

PR #421, opened 2016-06-09
Rename subscription factory

mchandschuh:
The name here was always a shame. What do you guys think about the rename? @jaredbroad @StefanoRaggi 

`ISubscriptionFactory` -> `ISubscriptionDataSourceReader`


jaredbroad:
Looks good! Much more appropriate

PR #425, opened 2016-06-11
Fixes TimeSlice bug preventing multiple custom symbols of same type

StefanoRaggi:
With this configuration in `Initialize`:

```
AddData<QuandlFuture>("SCF/CBOE_VX1_EW", Resolution.Daily);
AddData<QuandlFuture>("SCF/CBOE_VX2_EW", Resolution.Daily);
```
`OnData(QuandlFuture)` was called only for one of the symbols in each time step.

### DIFF
--- a/Engine/DataFeeds/TimeSlice.cs
+++ b/Engine/DataFeeds/TimeSlice.cs
@@ -16,15 +16,12 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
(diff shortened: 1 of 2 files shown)


jaredbroad:
Please test this with DailyFX data -- sometimes there are multiple points at the same time (like tick data). Ideally we need to support sending multiple -- making custom data a first class citizen.

PR #427, opened 2016-06-12
Fixes some failing regression tests

StefanoRaggi:


Common/Data/Auxiliary/LocalDiskFactorFileProvider.cs
```diff
@@ -71,7 +72,7 @@ public FactorFile Get(Symbol symbol)
                 return GetFactorFile(symbol, symbol.Value, market);
             }
 
-            var mapFile = mapFileResolver.ResolveMapFile(symbol.Value, symbol.ID.Date);
+            var mapFile = mapFileResolver.ResolveMapFile(symbol.Value, DateTime.Today);
```
mchandschuh: This seems suspect to me.

jaredbroad: If the map files are always using today's ticker string isn't this right? (symbol.Value is today's + the date of today?)

mchandschuh: Consider goocv before it was goog in a backtest. We would ask the map for resolver to find a goocv (symbol.Value) active today, but they're isn't one. Maybe what it should be doing is symbol.ID.Symbol and symbol.ID.Date. This would say give me the mail file based on the symbols first ticker/date combination.

jaredbroad: "Maybe what it should be doing is symbol.ID.Symbol and symbol.ID.Date" +1, this sounds right. This way the returned map file wouldn't change depending on the date today.

PR #429, opened 2016-06-15
Fixes bug: RealTime price not set every second

AlexCatarino:
On LiveTradingResultHandler.ProcessSynchronousEvents, realtime price was assigned to Security.Cache._lastData.Value and that assignment did not change Security.Price used to update Portfolio statistics (eg. TotalPortfolioValue).
For Forex securities, adds Portfolio.CashBook updates with realtime price.

### DIFF
--- a/Engine/Results/LiveTradingResultHandler.cs
+++ b/Engine/Results/LiveTradingResultHandler.cs
@@ -33,6 +33,7 @@ using QuantConnect.Packets;
 using QuantConnect.Securities;
 using QuantConnect.Statistics;
 using QuantConnect.Util;
+using QuantConnect.Securities.Forex;
 
 namespace QuantConnect.Lean.Engine.Results
 {
@@ -1022,6 +1023,15 @@ namespace QuantConnect.Lean.Engine.Results
                             if (last != null)
                             {
                                 last.Value = price;
+                                security.SetRealTimePrice(last);
+
+                                // Update CashBook for Forex securities
+                                Cash cash;
+                                var forex = security as Forex;
+                                if (forex != null && _algorithm.Portfolio.CashBook.TryGetValue(forex.BaseCurrencySymbol, out cash))
+                                {
+                                    cash.Update(last);
+                                }
                             }
                             else
                             {
--- a/Common/Securities/Security.cs
+++ b/Common/Securities/Security.cs
@@ -576,7 +576,7 @@ namespace QuantConnect.Securities
         }
 
         /// <summary>
-        /// Update any security properties based on the lastest market data and time
+        /// Update any security properties based on the latest market data and time
         /// </summary>
         /// <param name="data">New data packet from LEAN</param>
         public void SetMarketPrice(BaseData data)
@@ -589,7 +589,7 @@ namespace QuantConnect.Securities
         }
 
         /// <summary>
-        /// Update any security properties based on the lastest realtime data and time
+        /// Update any security properties based on the latest realtime data and time
         /// </summary>
         /// <param name="data">New data packet from LEAN</param>
         public void SetRealTimePrice(BaseData data)
--- a/Common/Securities/Security.cs
+++ b/Common/Securities/Security.cs
@@ -574,18 +574,30 @@ namespace QuantConnect.Securities
                 Exchange.SetLocalDateTimeFrontier(args.Time);
             };
         }
-        
+
         /// <summary>
         /// Update any security properties based on the lastest market data and time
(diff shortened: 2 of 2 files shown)


Common/Securities/SecurityCache.cs
```diff
@@ -37,7 +37,7 @@ public class SecurityCache
         /// <summary>
         /// Gets the most recent price submitted to this cache
         /// </summary>
-        public decimal Price { get; private set; }
+        public decimal Price { get; set; }
```
mchandschuh: I think there is a set market price function on security that should be called to do this. If a property is private set, care should be taken when changing to public.

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -840,29 +841,30 @@ public void StoreResult(Packet packet, bool async = true)
 
                         var highResolutionCharts = new Dictionary<string, Chart>(live.Results.Charts);
 
-                        // 10 minute resolution data, save today
-                        var tenminuteSampler = new SeriesSampler(TimeSpan.FromMinutes(10));
-                        var tenminuteCharts = tenminuteSampler.SampleCharts(live.Results.Charts, start, stop);
+                        // 1 minute resoluton data, save today
+                        var minuteSampler = new SeriesSampler(TimeSpan.FromMinutes(1));
+                        var minuteCharts = minuteSampler.SampleCharts(live.Results.Charts, start, stop);
```
mchandschuh: Are these changes intentional? They seem unrelated to the PR title.

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -840,29 +841,30 @@ public void StoreResult(Packet packet, bool async = true)
...
-                        live.Results.Charts = minuteCharts;
+                        live.Results.Charts = tenminuteCharts;
                         data_keys.Add(new
                         {
-                            Key = CreateKey("minute"),
+                            Key = CreateKey("10minute"),
                             Serialized = JsonConvert.SerializeObject(live.Results)
                         });
-
+
```
mchandschuh: Excess white space

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1021,17 +1022,26 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
...
+                                    cash.Update(last);
+                                }
                             }
                             else
                             {
                                 // we haven't gotten data yet so just spoof a tick to push through the system to start with
                                 security.SetMarketPrice(new Tick(DateTime.Now, subscription.Configuration.Symbol, price, price));
                             }
-
+
```
mchandschuh: Excess white space

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1021,17 +1022,26 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
...
                             {
                                 // we haven't gotten data yet so just spoof a tick to push through the system to start with
                                 security.SetMarketPrice(new Tick(DateTime.Now, subscription.Configuration.Symbol, price, price));
                             }
-
+                            
                             //Sample Asset Pricing:
                             SampleAssetPrices(subscription.Configuration.Symbol, time, price);
-                        }
+                        }
```
mchandschuh: Excess white space

Engine/Results/LiveTradingResultHandler.cs
AlexCatarino: They are. I am going to undo them and delete white spaces.

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1021,17 +1022,26 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
                             var last = security.GetLastData();
                             if (last != null)
                             {
-                                last.Value = price;
+                                security.Cache.Price = price;
```
mchandschuh: We can pump this new last value back into set market price. My fear is having two methods of doing this, but one updates other things, like models.

Common/Securities/SecurityCache.cs
jaredbroad: SetMarketPrice is updating volatility models so would cause different data in backtest/live; but might be good to have a SetRealTimePrice() setter method rather than setting the property directly

AlexCatarino: Ok. Let's do the SetRealTimePrice() option.

PR #430, opened 2016-06-15
Fixes enumerator stack for custom data in live trading

StefanoRaggi:
Issue #373


Engine/DataFeeds/Enumerators/FastForwardEnumerator.cs
```diff
@@ -83,7 +83,8 @@ public bool MoveNext()
             }
 
             // we've exhausted the underlying enumerator, iterator completed
-            return false;
+            _current = _enumerator.Current;
+            return true;
```
mchandschuh: Do we use this enumerator in backtest? In backtest we would want it to end (I think).

StefanoRaggi: `FastForwardEnumerator` is only used in live, backtesting uses `FillForwardEnumerator`

PR #431, opened 2016-06-15
Fixes minor bug on LiveTradingResultHandler.StoreResult

AlexCatarino:
10-minute resampling was computed before 1-minute resampling and its result was used to resample to 1-minute losing information in the process. Now, 1-minute resampling is performed prior to 10-minute.


mchandschuh:
I've always  wondered if there was something wrong in the sampling code there, nice!

Issue #428, opened 2016-06-15
Pushing cash balance updates from IBrokerage


jameschch:
It may be this is already possible, in which case, would appreciate someone pointing this out. It would be really useful to be able to push cash balance updates from the brokerage to the algorithm. Currently cash balances are only updated on initialization and on a daily schedule as far as I can tell.

My use case on Bitfinex: 

1. I have USD from a BTCUSD leveraged trade in trading wallet.
2. I transfer the USD from the trading wallet to the exchange wallet.
3. I make discretionary trades to convert USD to BTC.
4. I transfer btc to the trading wallet.
5. The algorithm makes leveraged trades with the correct cash balance.

The wss subscription will push wallet updates for each of these wallet transfers, and I am consuming them, I just need to push the updates. I'm considering exposing the method BrokerageTransactionHandler.PerformCashSync() to the brokerage. Is there already a way to do this? Any better ideas?

edit: Another option would be calling SecurityPortfolioManager.SetCash() directly. Is this advisable?

jaredbroad:
Correct we currently re-sync each day - which was intended to catch small differences like fees / interest (not large ones like withdrawals). To handle your use case we need "journal entry" events exposed by the brokerage via an event handler -- and having the BTH bind to the brokerage journal event - and trigger a cash sync when there's a cash movement.

The issue we found was syncing too often caused weird jumps / loss in state if we synced before or after a trade was withdrawn from the available cash. IB for example delays withdrawing the cash for a few minutes meaning we can't rely on them for "TotalPortfolioValue" state. 

A complete PR here would need through testing for all the brokerages to make sure they don't loose state   (e.g. portfolio value jumps 100% for 1-2 minutes after placing a trade)

jameschch:
Thanks. This makes sense. I noticed the defensiveness around the calls to PerformCashSync with relation to fills. Could this be implemented generally and then enabled on a brokerage by brokerage basis, perhaps as part for the BrokerageModel? I get the feeling you don't think this would be of great utility to any other brokerages.

jaredbroad:
It would be of great utility! We just need to make sure it doesn't break existing behavior :)

jameschch:
Right. I will look into this. For an event drive approach I would follow the pattern of Brokerage.OnOrderEvent.

jameschch:
The code to do this is just commented out in BrokerageTransactionHandler.HandleAccountChanged. Will just add a clause with IBrokerkageModel.AllowAccountUpdates

PR #434, opened 2016-06-16
Changes related to #428 allowing conditional push of cash balances

jameschch:
Very minor changes that re-enable existing feature for brokerage to push cash balance updates. This is disabled by default.

This was much too easy: it must be wrong. There is no direct test of the ITransactionHandler that might cover these changes. It's no problem to prove it's disabled by default, but where to add the test method?

### DIFF
--- a/Brokerages/Brokerage.cs
+++ b/Brokerages/Brokerage.cs
@@ -174,5 +174,13 @@ namespace QuantConnect.Brokerages
         /// </summary>
         /// <returns>The current cash balance for each currency available for trading</returns>
         public abstract List<Cash> GetCashBalance();
+
+        /// <summary>
+        /// Specifies whether the brokerage will instantly update account balances
+        /// </summary>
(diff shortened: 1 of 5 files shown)


jaredbroad:
Thank you for the PR @jameschch, as its disabled by default this seems safe enough. Then brokerages which have real-time cash balances can be enabled one by one.

jaredbroad:
@jameschch -- just brain storming with @StefanoRaggi -- what do you think about moving the boolean to the brokerage and making it a readonly property of the brokerages (e.g. "brokerage.AccountInstantlyUpdated")? I'm not sure it should be on the model since its a pure live trading brokerage concern. E.g. We wouldn't want a user overriding it with their own model. The user should also never interact or initiate this event -- its always coming from the live brokerages object.

jameschch:
Yes, that makes sense, this is an implementation detail rather than a logical capability of the brokerage.

PR #436, opened 2016-06-17
Adds ISubscriptionEnumeratorFactory

mchandschuh:
This is a series of mechanical refactorings to extract the enumerator creation gook from `FileSystemDataFeed` into a new abstraction: `ISubscriptionEnumeratorFactory`. This aims to further unify parts of the universe infrastructure. This change builds off of the [`data-feed-subscription-request`](https://github.com/QuantConnect/Lean/tree/data-feed-subscription-request) branch.

Where applicable, the live feed was also updated in separate commits so we can easily rip them out of need be.

Thoughts?


Engine/DataFeeds/Enumerators/Factories/OptionChainUniverseSubscriptionEnumeratorFactory.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators.Factories
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="ISubscriptionEnumeratorFactory"/> for the <see cref="OptionChainUniverse"/>
+    /// </summary>
+    public class OptionChainUniverseSubscriptionEnumeratorFactory : ISubscriptionEnumeratorFactory
+    {
+        private readonly BaseDataSubscriptionEnumeratorFactory _factory;
+        private readonly Func<SubscriptionRequest, IEnumerator<BaseData>, IEnumerator<BaseData>> _enumeratorConfigurator;
```
mchandschuh: It's a shame this is needed here. We need to apply fill forward/filtering to the individual subscriptions before they're sent into the `SynchronizingEnumerator` - Any ideas on a better way? For now I think this is better than the way it was (jumbled into `FileSystemDataFeed`)

PR #437, opened 2016-06-17
Adds Portfolio Optimization algorithm using Math.NET Numerics

AlexCatarino:
Adds PortfolioOptimizationAlgortithm.cs
This algorithm uses Math.NET Numerics library, specifically Linear Algebra object (Vector and Matrix) and operations, in order to solve a portfolio optimization problem.

### DIFF
--- a/Algorithm.CSharp/PortfolioOptimizationNumericsAlgorithm.cs
+++ b/Algorithm.CSharp/PortfolioOptimizationNumericsAlgorithm.cs
@@ -1,0 +1,186 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ * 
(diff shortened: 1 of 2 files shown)


Algorithm.CSharp/PortfolioOptimizationNumericsAlgorithm.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// </summary>
+    public class PortfolioOptimizationNumericsAlgorithm : QCAlgorithm
+    {
+        private const double _targetReturn = 0.1;
+        private const double _riskFreeRate = 0.01;
+        private double _lagrangeMultiplier;
+        private double _portfolioRisk;
+        private Dictionary<string, double> _mean;
+        private Dictionary<string, double> _stddev;
+        private Dictionary<string, double> _weights;
```
mchandschuh: These three values may make a nice 'SymbolData' class, instead of accessing parallel dictionaries. When possible, I try to avoid multiple parallel collections as they tend to hide the problem's natural objects.

Issue #439, opened 2016-06-20
Dockerize it


museyoucoulduse:
If Lean could be dockerized, it would be great. I can provide you with basic `Docker` file, but it is not the way it should be done.

What I think is that there should be some interaction with the Lean platform from outside the container, making it possible to use your own strategies.

Maybe using `config.json` and by mounting in a local directory with strategy? That way the end user will have no interaction with the internals of Lean and will be able to run it providing only calls to the API in his `Strategy.cs` file. Or better, giving the user something like admin panel in `Wordpress` where she can edit `BasicTemplateStrategy.cs` and customize it to her needs.

jaredbroad:
Welcome @museyoucoulduse, not sure I understand your use case -- can you describe it more?

We're tidying up the API classes now and will provide a fuller implementations for running algorithms via API (backtest and live) on quantconnect.com.  A docker file for users to run LEAN locally for their personal use or a hedge-fund use would be helpful to many people.

jaredbroad:
Closed by commit a7f85b9

PR #440, opened 2016-06-20
Dockerize it. Dockerfile.

museyoucoulduse:
#439 Add simple `Dockerfile`.

Use `mono:latest`, and follows `README` guide to spin up in `Linux`.


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM mono:latest
```
mchandschuh: @jaredbroad if I recall correctly we pinned to a certain mono version for performance reasons.

museyoucoulduse: @mchandschuh can you tell me the version number?

mchandschuh: I wish I could remember, I think 4.1, but am not positive.

jaredbroad: 4.051,

PR #446, opened 2016-06-28
Universe chaining

mchandschuh:
These changes provide functionality to allow chaining of universes. We chain universes by altering one to emit subscriptions for the other. For example, if we have coarse -> fine universe, then we'll alter the coarse universe such that it emits subscriptions for the fine universe. We'll also alter the fine universe such that it will prefilter the universe data to only include symbols selected by coarse.

The 'alteration' of a universe to done via the decorator pattern (same as used with data feed enumerator stacks). 

I've added extension methods to make these two use cases easier to follow in the code, see `UniverseExtensions`.


StefanoRaggi:
The commits in this PR are included in https://github.com/QuantConnect/Lean/pull/524 and have been merged.

Issue #383, opened 2016-04-21
Ensure Indicators Always Receive Adjusted Data


jaredbroad:
Indicators work best with smooth, fully adjusted data. When working with Raw or SplitOnly, DividendOnly adjusted data there are discontinuous jumps which break indicators.

This can be easily fixed by converting the data which goes into the indicator update manager and ensuring the data is always the fully adjusted version of the data (and if its already adjusted, as by default, then do nothing).

AlexCatarino:
Where should the data be adjusted (if it is not already)?
At [Algorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.Indicators.cs#L1151) or [Indicator](https://github.com/QuantConnect/Lean/blob/master/Indicators/IndicatorBase.cs#L75) level?

jaredbroad:
![](https://github.trello.services/images/mini-trello-icon.png) [Ensure Indicators Always Get Adjusted Data](https://trello.com/c/8kmvOASJ/986-ensure-indicators-always-get-adjusted-data)

mchandschuh:
Welp, I had a massive list of running comments here, that's what I get for not posting and updating.

**#lessons-learned**

1. Need to reset 'down stream' indicators (think `Of` extension)

2. Reset should happen depth first _before_ state is reset, allowing reset handlers to cache state if necessary
a. handled via `IIndicator.Resetting` event fires _before_ the reset is applied, which invokes sub-indicators into resetting (depth-first)

3. **:warning:[REQUIRES DISCUSSION]:warning:** -- sourcing data, applicable limits? Indicators can chain infinitely deep meaning it's impossible to hold all data in RAM required to get them back to speed. Perhaps initial impl focuses on  ~500 point data cache (configurable, perhaps computed via various indicators). 
a. Further, it's insufficient to maintain recent price data, a prime example is SPY second data consolidated into daily bars to create an SPY SMA_100_daily -- value is caching at consolidator level as it is a downsampling step that ultimately reduces the total recompute time (exponentially? only if nested consolidators, but even then I think it's quadratic, but I digress)...

[...]

PR #448, opened 2016-06-29
Ensure Indicators Always Receive Adjusted Data #383

AlexCatarino:
Adds AdjustSelectorOutput method that take a selector in and returns a selector that outputs adjusted data.


Common/Data/SubscriptionDataConfig.cs
```diff
@@ -101,6 +101,11 @@ public Symbol Symbol
         public decimal PriceScaleFactor;
```
jaredbroad: Why not use this?

AlexCatarino: I did try. 
`PriceScaleFactor` depends on `DataNormalizationMode`. When `DataNormalizationMode` is `Raw`, `PriceScaleFactor = 1` and we don't get the "real" price scale factor to adjust the price that goes into the indicator.
So I created a `UnconditionalPriceScaleFactor` on SubscriptionDataConfig and updated it along with `PriceScaleFactor` in a way that it would get the real price scale factor even if `DataNormalizationMode` is `Raw`. 
It wasn't working. Thus I printed out `PriceScaleFactor` and `UnconditionalPriceScaleFactor` and compare them with the value in the csv factor file. It was off. In the end, I decided to get the value from the `FactorFile` object and it is working.

mchandschuh:
Wow. Something is amiss with the commit history on this PR.

Issue #32, opened 2015-02-21
[Feature] Model Interest Paid on Leveraged Assets


jaredbroad:
Leverage is currently free in LEAN. Implement a system for deducting interest charges from cash when holding leveraged stock overnight. Apply a default interest rate of 1.5% on the leveraged fraction of holdings. Potentially best place is to apply cash charge on day change in AlgorithmManager data loop.

QANTau:
When looking at this feature, would it be possible to consider the calculation of swap for FX transactions also.  It would be nice to be able to specify a 'Calculation Time' that would fire an event where the calculation could be performed.

The default event code could calculate a 1.5% default rate, but could be overridden to perform custom calculation (in the case of FX swap, we could probably load external data to calc the swap rate) and return the +/- interest back into the account.

Applying on the 'day change' may not work in all cases as brokers may not charge swap at the end of the day (I have a local broker that charges the swap at 1930hrs local time), so the ability to set this to a custom time of day may be more appropriate.

I definately don't have the C# skills to build an event model like this, but will volunteer to put together an example of swap calculation code using external interest rate data if that helps.

jaredbroad:
Great feedback thanks @QANTau. We will factor this into the design of the leverage fee calculations. Its probably best done in the transaction models class however it has some similarities to the Benchmark system. 

One way to help would be starting with the source of the swap rate data, and how to import it into LEAN. It might be nice if this was standardized/similar to the way benchmark loads in data (https://github.com/QuantConnect/Lean/issues/30). If possible it would be sourced from a web-source like Yahoo to be self-updating.

jaredbroad:
OANDA - Historical Interest Data:
https://www.oanda.com/forex-trading/analysis/historical-rates

OANDA - How they calculate interest:
https://oanda.secure.force.com/AnswersSupport?urlName=Interest-Rate-Calculation-1436196464449&language=en_US

ebengtso:
I provided a sample code to calculate interests..

https://www.quantconnect.com/forum/discussion/5460/calculate-rollover-interest-for-oanda/p1

It would be interesting if the model could attach the interests to the trade, as costs or contribution.

lucasbrynte:
Bitfinex interest rates:
https://www.bitfinex.com/stats/#rates-on-margin-funding

Further info:
https://support.bitfinex.com/hc/en-us/articles/214424645-What-interest-rate-is-charged-on-margin-positions-on-Bitfinex
[...]

PR #449, opened 2016-06-30
Model Interest Paid on Leveraged Assets

AlexCatarino:
Implements IMarginInterestModel interface to model interest paid on leveraged assets.
Implements a ConstantMarginInterestModel to model a constant interest paid.
Modifies SecurityPortfolioManager and AlgorithmManager to enable the margin interest modelling.
Adds MarginInterestTests to test the implementation.
Ref. #32

### DIFF
--- a/Common/Orders/MarginInterest/ConstantMarginInterestModel.cs
+++ b/Common/Orders/MarginInterest/ConstantMarginInterestModel.cs
@@ -1,0 +1,102 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


jaredbroad:
Just a note from our discussion:

 - Perhaps the holiday math could be simplified by recording last time: (now  - lastRequest).TotalDays / 365. (Avoid needing to use exchange hours)

 - Total Margin Used isn't right number to use here. Need a "loan used" figure which factors in the amount of TPV/cash we have in the account. 
If you have $50k and spend $10k, it will report $5k margin used even though you have used none.
If you have $50k and spend $50, it will report $25k used. - even though $0 used (edited).
If you have $50k and spend $60k, it will report $30k, even though you've only used $10k loan.

 - Rename PayMarginInterest from the model to "GetMarginInterest".

 - Rename "PayMarginInterest" from the portfolio to "ApplyMarginInterest" following the patterns of the other model applications.

 - Insert live mode flag in algo manager to stop applying the interest in live mode. Brokerages will do that for you.

 - Put all the files in the securities folder with the other security models

jaredbroad:
Closing as buggy implementation. Can apply fixes and resubmit.

PR #470, opened 2016-07-08
Updated Quandl API to version 3

Phoenix1271:
- Quandl is currently supporting REST API for data in version 3. This change reflects new syntax of the API.
- IsAuthCodeSet property was incorrectly set to true if API token in config file is empty.


jaredbroad:
Thank you @Phoenix1271, good catch

PR #471, opened 2016-07-08
Updated QuandlFuturesDataAlgorithm sample

Phoenix1271:
- QuandlFutures data type was moved to Custom/Data folder because it's reused on multiple places outside the sample
- Name of class is matching the filename and is now consistent with rest of samples
- Added Symbol() call to remove compilation warning


jaredbroad:
Thank you for the PR @Phoenix1271 - sadly if the QuandlFuture is in Common it will be compiled into the quantconnect.com website algorithms too. This means anyone with that code in their files will start getting compile errors suddenly. For such a small wrapper it's probably not worth moving.

Issue #457, opened 2016-07-07
Lot size rounding orders to 0, then submitting 0-quantity orders.


jaredbroad:
Issue when submitting an order with quantity less than the lot size; then its rounded to 0 after the sanity checks. Then the 0-quantity order is submitted.

E.g. FXCM lot size 1k, submit order for 200 EUR, rounded to 0 but submitted anyway (and changed fees)

Should probably perform a non-zero check after rounding the order; &/or; round the lot size earlier in the process and check it still passes the non-zero check

![2016-07-06 18_09_58-algorithmic trading platform - quantconnect com](https://cloud.githubusercontent.com/assets/2243225/16639676/d3c934d4-43bf-11e6-8644-012dfd5591aa.png)

mchandschuh:
Some comments/thoughts/recommendations:

The method `BrokerageTransactionHandler.RoundOffOrder` seems misplaced and hard coded. Recommend moving to concern to `IBrokerageModel` via a new method -- `IBrokerageModel.PrepareSubmitOrderRequest(Security, SubmitOrderRequest)` (naming suggestions welcomed) -- which would give the model a chance to modify the request as soon as its created in the algorithm. This will allow the behavior to be pluggable and then also be subject to the various sanity checks within the algorithm.

We don't call `algorithm.BrokerageModel.CanSubmitOrder` in live mode. The documentation for the interface method seems to indicate its there to support things such as connectivity and size restrictions. Recommend allowing this logic to execute even in live mode, which will provide better consistency between backtest/live modes. This does not directly impact this bug, but is worth noting nevertheless.

@jaredbroad Let me know what you think and I can get started on this.

jaredbroad:
Agree RoundOffOrder was kind of randomly placed - but don't all orders pass through there? It was there since all brokerages needed rounding and rather than replicating it on each model it was only coded once (if done via interface-method on that IBrokerageModel that is fine too).

Agree re: CanSubmitOrder; certainly helps understand the error messages. Some of the brokerage errors are misleading so best if we filter them first. 

How would the "order rounded" messaging get back to the user?

mchandschuh:
###### RE: Duplicate rounding logic
I believe all of our models derive from a `DefaultBrokerageModel` which would likewise define the default rounding behavior. So no duplication really needed.
###### RE: Order rounded messasge
[...]

PR #472, opened 2016-07-08
Order is not placed if order is lower than minimum lot size

Phoenix1271:
- Added missing check after rounding of quantity
- Added Moq package to unit testing so it's possible to mock dependencies
- Fixed tests for BrokerageTransactionHandler that now tests what component should do instead of how it operates inside (larger test coverage and less brittle tests).

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -42,7 +42,7 @@ namespace QuantConnect.Algorithm
         public bool IsWarmingUp
         {
             get;
-            private set;
+            internal set;
         }
 
         /// <summary>
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -42,7 +42,7 @@ namespace QuantConnect.Algorithm
         public bool IsWarmingUp
         {
             get;
-            internal set;
+            private set;
         }
 
         /// <summary>
--- a/Common/Properties/SharedAssemblyInfo.cs
+++ b/Common/Properties/SharedAssemblyInfo.cs
@@ -1,5 +1,4 @@
 ﻿using System.Reflection;
-using System.Runtime.CompilerServices;
 
 // common assembly attributes
 [assembly: AssemblyDescription("Lean Algorithmic Trading Engine - QuantConnect.com")]
--- a/Common/Properties/SharedAssemblyInfo.cs
+++ b/Common/Properties/SharedAssemblyInfo.cs
@@ -1,4 +1,5 @@
 ﻿using System.Reflection;
+using System.Runtime.CompilerServices;
 
 // common assembly attributes
 [assembly: AssemblyDescription("Lean Algorithmic Trading Engine - QuantConnect.com")]
@@ -6,6 +7,9 @@
 [assembly: AssemblyCompany("QuantConnect")]
 [assembly: AssemblyTrademark("QuantConnect")]
 [assembly: AssemblyVersion("2.2.0.2")]
+
+[assembly: InternalsVisibleTo("QuantConnect.Tests")]
+
 #if DEBUG
     [assembly: AssemblyConfiguration("Debug")]
 #else
--- a/Common/Properties/SharedAssemblyInfo.cs
+++ b/Common/Properties/SharedAssemblyInfo.cs
@@ -7,9 +7,6 @@ using System.Runtime.CompilerServices;
 [assembly: AssemblyCompany("QuantConnect")]
 [assembly: AssemblyTrademark("QuantConnect")]
 [assembly: AssemblyVersion("2.2.0.2")]
-
-[assembly: InternalsVisibleTo("QuantConnect.Tests")]
-
 #if DEBUG
     [assembly: AssemblyConfiguration("Debug")]
 #else
--- a/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
+++ b/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
@@ -401,21 +401,7 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
             {
                 foreach(var request in _orderRequestQueue.GetConsumingEnumerable(_cancellationTokenSource.Token))
                 {
-                    OrderResponse response;
-                    switch (request.OrderRequestType)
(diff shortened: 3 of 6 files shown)


jaredbroad:
@mchandschuh - welcome feedback as lightweight submission for the issue #457

Algorithm/QCAlgorithm.History.cs
```diff
@@ -42,7 +42,7 @@ public IHistoryProvider HistoryProvider
         public bool IsWarmingUp
         {
             get;
-            private set;
+            internal set;
```
mchandschuh: Is this necessary? If you need to transition the algorithm from `IsWarmingUp==true` to `IsWarmingUp==false` you can use `SetFinishedWarmingUp()`. Ideally we don't need to use internal modifiers to enable testing.

Common/Properties/SharedAssemblyInfo.cs
```diff
@@ -1,11 +1,15 @@
...
+using System.Runtime.CompilerServices;
 
 // common assembly attributes
 [assembly: AssemblyDescription("Lean Algorithmic Trading Engine - QuantConnect.com")]
 [assembly: AssemblyCopyright("Copyright ©  2015")]
 [assembly: AssemblyCompany("QuantConnect")]
 [assembly: AssemblyTrademark("QuantConnect")]
 [assembly: AssemblyVersion("2.2.0.2")]
+
+[assembly: InternalsVisibleTo("QuantConnect.Tests")]
```
mchandschuh: If we need access to internal members in order to perform testing, I think that means we need to rethink the design and break things into smaller, testable pieces as needed.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -595,6 +581,26 @@ public void Exit()
         }
 
         /// <summary>
+        /// Handles a generic order request
+        /// </summary>
+        /// <param name="request"><see cref="OrderRequest"/> to be handled</param>
+        /// <returns><see cref="OrderResponse"/> for request</returns>
+        internal OrderResponse HandleOrderRequest(OrderRequest request)
```
mchandschuh: I think this is a great idea, but we should make it `public` and have it do the complete handling of the request. This will effectively change the `Run` method such that it's main responsibility is just consuming the queue and invoking this method. Then we can call this method directly from tests knowing that it handles all processing of an `OrderRequest`

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -609,6 +615,11 @@ private OrderResponse HandleSubmitOrderRequest(SubmitOrderRequest request)
             // rounds off the order towards 0 to the nearest multiple of lot size
             order.Quantity = RoundOffOrder(order, security);
 
+            if (order.Quantity == 0)
+            {
+                return OrderResponse.ZeroQuantity(request);
```
mchandschuh: Regardless of where the `RoundOffOrder` behavior ends up, I think this check should be here. It seems very reasonable that a transaction handler would prevent zero quantity orders.

Algorithm/QCAlgorithm.History.cs
Phoenix1271: I didn't notice that you have another method to set property. Can be reverted and `SetFinishedWarmingUp()` method called in the test instead.

Tests/Engine/BrokerageTransactionHandlerTests/BrokerageTransactionHandlerTests.cs
```diff
@@ -14,89 +14,118 @@
 */
 
 using System;
+using Moq;
```
mchandschuh: We haven't really thought out which library we would use for mocking, but I've heard/read good things about `Moq` and think its a great addition to the test project.

Common/Properties/SharedAssemblyInfo.cs
Phoenix1271: This has been added for BrokerageTransactionHandler message pump to be visible outside but if you want to make `HandleOrderRequest()` public it can be removed.

Tests/packages.config
```diff
@@ -1,8 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="IKVM-WithExes" version="7.3.4830.1" />
+  <package id="Castle.Core" version="3.3.3" targetFramework="net45" />
```
mchandschuh: What is `Castle.Core` required for?

Phoenix1271: Dependency of the Moq library to create dynamic proxies for mocked objects.

mchandschuh: Awesome! There's a few places I'll have to go clean up to use `Moq`

mchandschuh:
I've made a few comments, but overall the approach and design are good. I enjoy seeing things refactored to enable better testing. Great contribution @Phoenix1271!

Phoenix1271:
Removed mentioned internal modifiers and used `SetFinishedWarmingUp()` method.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -401,21 +401,7 @@ public void Run()
...
-                        case OrderRequestType.Cancel:
-                            response = HandleCancelOrderRequest((CancelOrderRequest) request);
-                            break;
-                        default:
-                            throw new ArgumentOutOfRangeException();
-                    }
+                    var response = HandleOrderRequest(request);
 
                     // we've finally finished processing the request, mark as processed
                     request.SetResponse(response, OrderRequestStatus.Processed);
```
mchandschuh: I think this should be moved to within the `HandleOrderRequest` method

Phoenix1271: You mean response processing?

mchandschuh: Yes, old line 421, new line 407. Without this line, the `HandleOrderRequest` method is incomplete.

Phoenix1271: Moved response processing inside `HandleOrderRequest()` method.

PR #473, opened 2016-07-08
Updated the GetCashBalance to reflect the correct cash in the cashbook and Updated the PerformCashSync

devalkeralia:
1. Updated the response timeout for the FXCM brokerage from 2.5 seconds to 5 seconds.

2. Changed the GetCashBalane() for FXCM brokerage to reflect the correct value of cash for the cashbook.

3. Updated the PerformCashSync() to sync the correct values for the portfolio.

### DIFF
--- a/Brokerages/Fxcm/FxcmBrokerage.Messaging.cs
+++ b/Brokerages/Fxcm/FxcmBrokerage.Messaging.cs
@@ -39,7 +39,7 @@ namespace QuantConnect.Brokerages.Fxcm
 
         private readonly object _locker = new object();
         private string _currentRequest;
-        private const int ResponseTimeout = 2500;
+        private const int ResponseTimeout = 5000;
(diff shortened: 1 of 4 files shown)


jaredbroad:
@StefanoRaggi @mchandschuh 
Proposing tweaking way cash sync works a little - when no matching cash found at the brokerage its removed from the model cash book. 

There were issues with left over small currency balances after a FX carry trade completed (profit/loss in the quote currency). Some brokerages leave these balances in the trade currency, other brokerages convert it immediately to account currency. When converted immediately to account currency the QCAlgorithm portfolio would still show the small difference in the trade-quote currency.

To make the behavior and TotalPortfolioValue consistent the cashbook was sync'd zero'ing out anything that wasn't found.

Previously the TPV was making large jumps on brokerages with singular account values.

PR #478, opened 2016-07-12
Remove buying power check when reducing long position size

AlexCatarino:
Checks whether a new order reduces a long position size, if so do not consider the mechanics to verify sufficient capital for order.

### DIFF
--- a/Common/Securities/SecurityTransactionManager.cs
+++ b/Common/Securities/SecurityTransactionManager.cs
@@ -313,6 +313,9 @@ namespace QuantConnect.Securities
                 return false;
             }
 
+            int remainingQuantity;
+            if (TryGetRemainingQuantity(order, security, out remainingQuantity)) return true;
+
             var freeMargin = security.MarginModel.GetMarginRemaining(portfolio, security, order.Direction);
             var initialMarginRequiredForOrder = security.MarginModel.GetInitialMarginRequiredForOrder(security, order);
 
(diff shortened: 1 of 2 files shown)


mchandschuh:
I'm curious on the background here.  Is this bug indicative of a problem with his we calculate sufficient capital for order? I think the margin calculations are supposed to account for the direction of the trade. @jaredbroad - maybe I'm not understanding something here?

AlexCatarino:
When we calculate sufficient capital for order, we account for [direction of the trade](https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityMarginModel.cs#L149). However it does not account for closing long positions that is independent of margin.
I thought on modifying how [`initialMarginRequiredForOrder`](https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityTransactionManager.cs#L317) is calculated, because it takes into account the whole order value as a cost but part of the cost is covered by selling current positions. However, it could be higher than the free margin (if it is negative) and prevent selling long positions.

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -353,5 +356,32 @@ private static bool Completed(Order order)
...
+        /// Returns true when order quantity close or reduce long position
+        /// </summary>
+        /// <param name="order">Order we're checking</param>
+        /// <param name="security">Security that the order tries to close or reduce</param>
+        /// <param name="value">Remaining quantity to close long position</param>
+        /// <returns>True if order close or reduce long position</returns>
+        private bool TryGetRemainingQuantity(Order order, Security security, out int value)
+        {
+            value = 0;
+            if (order.Direction != OrderDirection.Sell) return false;
```
jaredbroad: Buying can be used to close a short position. I don't think this can be based on the order direction.

jaredbroad:
@mchandschuh there were some edge cases for margin calls where the SufficientCapitalRemaining method was returning false and preventing liquidating the margin.

@AlexCatarino Please submit supporting tests which fails before and pass after your fix.

mchandschuh:
@jaredbroad thanks for the background, and I dig the red, green refractor.

Issue #465, opened 2016-07-07
Local Charting "Vector smash protection is enabled"


MoreDough:
There seems to be a race condition between the backtest and the charting implementation in LeanWinForm.

When the problem occurs, "Vector smash protection is enabled" is shown at the bottom of console output window.

If you put Thread.Sleep(15) in the OnData(Slice data) method in the algo, the backtest is slowed down and charting works.

jaredbroad:
It is a protection mechanism of Internet Explorer. Ideally we can use another cross platform web browser plugin but I haven't found one yet that works well. 

Maybe just the first few browser injections can be slowed down? The result handler only does it every second so its probably just the initial injections.

MoreDough:
I don't know what the problem is and will not look at this any more.

jameschch:
There is a slow answer and a quick answer to this. The quick answer is that the code attempts to use the most recent IE that's installed and I suspect the issue only occurs on older versions (I can't recreate on IE11).  It may be worth establishing the browser version(s) experiencing the problem, and displaying a warning that the user should upgrade, I'm not inclined to create a compatibility fix for old versions of IE.

The slow answer is that the IE WebBrowser control has always been and is still notoriously problematic and we are better off without it. I'm looking into alternatives.

jaredbroad:
Agree completely. It was the only thing I could get to compile/work on windows and linux (in mono the WebBrowser control gets converted to MonoBrowser under the hood). Open to alternatives though!

MoreDough:
Just tested again with no slowdown sleep in algo. Chart window is shown but contains no chart data, only the static stuff like scroller, 2 tabs and such.

From console window:
20160713 22:27:34 Trace:: Engine.Run(): Disconnecting from brokerage...
20160713 22:27:34 Trace:: Engine.Run(): Disposing of setup handler...
20160713 22:27:34 Trace:: Engine.Main(): Analysis Completed and Results Posted
20160713 22:27:34 Trace:: BacktestingResultHandler.ProcessLogMessages(): Ready
20160713 22:27:34 Trace:: FileSystemDataFeed.Exit(): Exit triggered.
Engine.Main(): Analysis Complete. Press any key to continue.
Vector smash protection is enabled.

My IE is ver 11, about box says:
Version: 11.0.9600.18349
Update versions: 11.0.32 (KB3160005)

I have Win 7 prof. (don't know if this could make a difference)

[...]

PR #489, opened 2016-07-17
Reinstate GeckoFX for non-mono

jameschch:
The mono library basically has an internal switch 
`if windows then IE else mono browser
`

I have changed this to a compiler directive of the form:
`#if !mono then geckoFX #else mono browser
`

I shan't go into the reasoning behind this decision here.

I'm not aware of the problems that led to GeckoFX being removed: presumably the main (only?) one was getting it to work on Linux. I have not made code changes to the mono case and am in the process of setting up my environment for testing, If this is a priority fix an existing mono environment would be a quicker option.


devalkeralia:
On Windows: Builds and runs correctly.

On Linux: Builds correctly but does not run. Always throws a NullReferenceException for Navigate. Wasn't working before and still isn't working. The ReadyState of the browser always remains "Uninitialised".

jaredbroad:
Thanks @jameschch - 

Before Gecko wasn't used as it never worked in linux and barely worked on Windows (because its BaseClass reference dependency wasn't completed yet). 

I propose we merge this in even though not perfect its slightly better than current and fixes issue #465

jaredbroad:
Nice PR @jameschch, all merged in!

PR #492, opened 2016-07-19
Add fine fundamental data classes

StefanoRaggi:


Common/Data/Fundamental/Generated/CompanyReference.cs
```diff
@@ -0,0 +1,277 @@
...
+		/// </summary>
+		/// <remarks>
+		/// Morningstar DataId: 6
+		/// </remarks>
+		[JsonProperty("6")]
+		public string CIK { get; set; }
+
+		/// <summary>
+		/// At the Company level; each company is assigned to 1 of 3 possible status classifications; (U) Public, (V) Private, or (O) Obsolete:
+		///  Public-Firm is operating and currently has at least one common share class that is currently trading on a public exchange.
```
jaredbroad: I wonder if this character will cause issues with readers / XML parsers? Can you please strip it out and replace all non A-Z,0-9 characters with safe versions? E.g. there are a few " ’ " instead of " ' " as well as " “1”  "  instead of "  "1" "

PR #494, opened 2016-07-21
Change platform target for Views and Launcher to x86

StefanoRaggi:
Unfortunately GeckoFx is 32bit only causing the following build warning:
warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "Geckofx-Core, Version=45.0.6.0, Culture=neutral, PublicKeyToken=3209ac31600d1857, processorArchitecture=x86", "x86". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.

### DIFF
--- a/Launcher/QuantConnect.Lean.Launcher.csproj
+++ b/Launcher/QuantConnect.Lean.Launcher.csproj
@@ -23,7 +23,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <PlatformTarget>x86</PlatformTarget>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -32,6 +32,7 @@
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <PlatformTarget>x86</PlatformTarget>
   </PropertyGroup>
   <PropertyGroup>
     <StartupObject>QuantConnect.Lean.Launcher.Program</StartupObject>
--- a/UserInterface/QuantConnect.Views.csproj
+++ b/UserInterface/QuantConnect.Views.csproj
@@ -23,7 +23,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <PlatformTarget>AnyCPU</PlatformTarget>
+    <PlatformTarget>x86</PlatformTarget>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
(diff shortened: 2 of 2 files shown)


mchandschuh:
This is a shame. Because of our user interface layer, our  entire application is forced into 32 bit. Maybe we can compile as  any cpu and have the UI project do 32 but, but launcher remains any cpu, which should default to 64 but on the appropriate machines.

mchandschuh:
All that being said, I should have asked if you tried just seeing views to 32 bit

StefanoRaggi:
The warning currently is on both projects, the UI appears to be working with or without this PR.
Test runners are also required to be running 32 bit to avoid Composer errors.
That said, I agree it's a shame (the alternative is to roll back to pre-GeckoFx, which is Windows only by the way)

jameschch:
It doesn't show it in the code, but before reinstating geckoFX I tried a few different web browser control libraries. This meant trying to change the build profiles in the way that's being described. I've come to two conclusions through doing this:

1. The flexibility is hamstrung by the fact that the browser control runs in a thread spawned by the launcher. This is to enable a communication channel. What might be more flexible is to spawn a process and change from inter-thread to inter-process communication (via wcf or whatever). This means the launcher and views can be separate build profiles and possibly separate solutions and repos.
2. It's a convenience to have one control that works in both Linux and Windows and will not significantly add to the code base to choose the right library for the right platform. There are CEF (chrome) based libraries that offer cross platform, but maybe something like an IBrowserProvider is worthwhile so these controls can be integrated in a uniform way.

The developers seem to intend GeckoFX to also run in linux. I'm just getting my mono environment setup to test this. The other library that I found to be fairly painless is:  http://cefsharp.github.io/

StefanoRaggi:
@mchandschuh The only effect of this PR is to remove the build warning mentioned above.

Launcher is already running as a 32bit process - can be verified with `Environment.Is64BitProcess`.
It is compiled using AnyCPU platform target with "Prefer 32 bit" option. If you uncheck it, on a 64bit machine Launcher will start as a 64bit process but `Composer.GetExportedValueByTypeName` will throw `BadImageFormatException` as expected (trying to load a 32bit library into a 64bit process).

@jameschch I agree that making Views an executable and using IPC would solve the 32/64 issue. 

I also looked at the other libraries available (CefSharp, CefGlue, ChromiumFx) and they all have missing or experimental Linux support:
CefSharp - https://github.com/cefsharp/CefSharp/issues/1450
CefGlue - https://bitbucket.org/xilium/xilium.cefglue/issues/108/linux-support
ChromiumFx - https://bitbucket.org/chromiumfx/chromiumfx/issues/9/mono-support

so it looks like GeckoFx is the right choice after all: https://bitbucket.org/geckofx/geckofx/wiki/Home

jaredbroad:
Sadly forcing 32bit isn't really a viable option; making it a separate executable isn't a bad idea -- then it could be long running and open a new tab-view for each backtest which is run. On load the UX exe could also check if another instance of itself is running and if so, switch focus to that window.

Ideally this would be in the same repo for now though since we could use the toolbox projects easily (imagine UX on a IDataDownloader[] for LEAN to allow people to download data they needed from a list of providers).

We also brainstormed allowing saving the local .json backtest result files and allow opening them later, and saving the .json quantconnect.com results and opening them locally.

@jameschch, What do you think about spearheading this project? Making UX a standalone executable; switch focus to self on load; make LEAN IPC messaging handler; and make a UX side IPC receiver.

jameschch:
It sounds like a nice idea to extend the capabilities of the desktop client. At the moment I have several things not-quite-done that I'd really like to get squared away first.

I'd want to make sure the client was consistent with the overall architectural vision of the project so there may need to be some dialogue before it gets moving.

jaredbroad:
Closing as addressed by PR: https://github.com/QuantConnect/Lean/pull/550

PR #495, opened 2016-07-22
Brokerage Specific Default BrokerageMessagingHandler + FXCM BrokerageMessagingHandler Bug Fix

devalkeralia:
Added the CreateBrokerageMessageHandler() which initializes the MessageHandler for the brokerages. Fixes the bug which caused the FXCM live algorithms to disconnect and throw runtime errors when the market opened on sunday evening.

For FXCM, I have chosen a timeout of negative 30 minutes since the market hours for FXCM are a bit flexible and they do not always open on time.

I also moved the DefaultBrokerageMessageHandler.cs into Common/Brokerages so that I can use it in the BrokerageFactory to get around the circular dependency.


Common/Interfaces/IBrokerageFactory.cs
```diff
@@ -54,5 +54,11 @@ public interface IBrokerageFactory : IDisposable
         /// <param name="algorithm">The algorithm instance</param>
         /// <returns>A new brokerage instance</returns>
         IBrokerage CreateBrokerage(LiveNodePacket job, IAlgorithm algorithm);
+
+        /// <summary>
+        /// Gets a brokerage message handler
+        /// </summary>
+        IBrokerageMessageHandler CreateBrokerageMessageHandler(IAlgorithm algorithm, AlgorithmNodePacket job,
+             IApi api, TimeSpan? initialDelay = null, TimeSpan? openThreshold = null);
```
mchandschuh: Do we need all of the parameters on the interface or can some of them be resolved from within the individual implementations

devalkeralia: @mchandschuh fair point. We don't. We would just need the algorithm, job and the api to initialize it. I could remove the rest.

PR #497, opened 2016-07-23
Fix invalid margin calls with equities when using leverage

StefanoRaggi:


jaredbroad:
Closing as discussed with @StefanoRaggi its not the right behavior we're seeking to model. IB typically liquidates partially before the cash limit is reached. (e.g. For a 2x levered holdings; once the holdings have reached 50% of value the algorithm should hold no shares - they'd all have been liquidated).

Issue #498, opened 2016-07-25
Large Equity Jumps with Live Paper Forex


jaredbroad:
Large account equity changes when trading FX paper trading. Jumps timed precisely with trades so likely is an issue with the cashbook model. 

![2016-07-25 13_07_54-algorithmic trading platform - quantconnect com](https://cloud.githubusercontent.com/assets/2243225/17112434/085c086c-5274-11e6-8679-1afe3ac2405d.png)

jaredbroad:
Issue appears using the symbols:
public string[] Symbols = { "EURUSD" , "GBPUSD", "NZDUSD", "AUDUSD", "USDCAD", "EURGBP", "GBPAUD", "USDCHF"};

Multiple jumps with these trades
[tradelog.xlsx](https://github.com/QuantConnect/Lean/files/382549/tradelog.xlsx)

![2016-07-25 17_53_36-algorithmic trading platform - quantconnect com](https://cloud.githubusercontent.com/assets/2243225/17118860/c3bb8e40-5290-11e6-92b9-e5ce0d58bcac.png)

PR #501, opened 2016-07-27
Added a check for non-USD currencies while updating cashbook

devalkeralia:
This is a proposed fix for the strange jumps in the equity when the trade gets made. #498  
 
The problem is that when a 'non-USD' currency comes in (for eg. EURGBP), the Cash.Update() replaces the rate for EUR with the rate for the 'non-USD' currency(EURGBP).

The proposed fix here is to have a check for the non-USD currency which comes in. The rate for the other currencies would already have been updated and this would prevent the over-writing of the conversion rates.

After testing this, the equity jumping has stopped by using this fix.

I would like to know what the others think about the fix. 

P.S. I have used the term 'non-USD' because the Account Currency is USD. So it's actually 'non-AccountCurrency'

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -84,6 +84,7 @@ namespace QuantConnect.Securities
         public void Update(BaseData data)
         {
             if (_isBaseCurrency) return;
+            if (!data.Symbol.Value.Contains(CashBook.AccountCurrency)) return;
             
             var rate = data.Value;
             if (_invertRealTimePrice)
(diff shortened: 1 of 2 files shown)


jaredbroad:
This might not be the complete fix here; how does the EURGBP currency get sent to the cashbook.update? Perhaps this PR will hide another bug we should also fix: e.g. https://github.com/QuantConnect/Lean/commit/14e7beb4157cf03dad3252c7397de3de6eec7af9

mchandschuh:
I think the issue here isn't with the cash object, but whomever is calling
update with the incorrect information. For the example provided, we need to
create a conversion by combining two currency conversions, EURUSD and
GBPUSD. I thought this was already handled, but forget where that code
lives.

PR #502, opened 2016-07-27
Fixes indicators lagging when combined with consolidators.

AlexCatarino:


Tests/Common/Data/BaseDataConsolidatorTests.cs
```diff
@@ -184,5 +184,45 @@ public void AggregatesPeriodInPeriodModeWithDailyDataAndRoundedTime()
...
         }
+
+        [Test]
+        public void ConsolidatesWithRegisterIndicator()
+        {
+            var consolidator = new BaseDataConsolidator(TimeSpan.FromMinutes(5));
+            consolidator.DataConsolidated += OnFiveMinutes;
+
+            indicator = new SimpleMovingAverage(2);
+            RegisterIndicator(indicator, consolidator);
```
AlexCatarino: At first I tried to subscribe to events by using an anonymous methods. That way, the order I subscribed them matter. 
Then, I emulated the bug reported by this [post](https://www.quantconnect.com/forum/discussion/1289/the-intrinsic-lag-created-by-the-consolidator/p1) on QC forum. In this case, **before** the fix, `RegisterIndicator` was always called after the `OnFiveMinutes`.

PR #505, opened 2016-07-29
Declare variable 'Log' with 'new' keyword

bryanwang107:
Log variable was already declared in base class, QCAlgorithm.cs, so in order to declare again in UpdateOrderLiveTestAlgorithm.cs, must use "new" keyword

Warning Code: CS0108
UpdateOrderLiveTestAlgorithm.cs Line 179


jaredbroad:
Thanks for the PR @bryanwang107, overall its good - just in the future please use a more descriptive commit message: "Change to UpdateOrderLiveTestAlgorithm.cs" isn't quite enough ;)

Issue #487, opened 2016-07-14
Live realtime prices update after hours even if extended market hours are false


jaredbroad:
In live trading -- the charting/data sampling continues to get updated with updated TotalPortfolioValue values even after the market has closed - indicating the SetRealTimePrice() is being set after market close, and then the result handler is sampling that updated price.

If the subscription sets extended: false we probably shouldn't set the real-time prices after-hours either.

mchandschuh:
I wonder how IB evaluates portfolio value for margin calculations in these
events and whether or not we should follow IB here.

jaredbroad:
Not sure, but it would seem unethical/impossible to trigger margin calls on pre/post market prices (they can vary wildly with no liquidity). Its probably market hours only?

Its not the data loop getting called, just the charting sampling. 95% of people don't care about pre-post market data so no reason to show it in results and potentially scare people (with large drops in portfolio value because of a single post market trade etc).

mchandschuh:
Don't disagree with the margin calls. I was thinking about trading multiple
markets, say forex and us equities. If the after market price plummets,
will it affect how much currency you can trade in forex market.

jaredbroad:
Fixed by pr #809

PR #507, opened 2016-07-29
Prevents realtime prices update after hours

AlexCatarino:
If market is closed for any asset we are trading, do not set security real time price nor update the cashbook.
If market is closed for all assets, do not update charting/data sampling with updated TotalPortfolioValue.
Ref. #487

### DIFF
--- a/Engine/Results/LiveTradingResultHandler.cs
+++ b/Engine/Results/LiveTradingResultHandler.cs
@@ -1007,6 +1007,8 @@ namespace QuantConnect.Lean.Engine.Results
                 //Set next sample time: 4000 samples per backtest
                 _nextSample = time.Add(ResamplePeriod);
 
+                var canSampleEquity = false;
+
                 //Update the asset prices to take a real time sample of the market price even though we're using minute bars
                 if (_dataFeed != null)
                 {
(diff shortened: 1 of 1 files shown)


Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1019,6 +1021,12 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
                             //Sample Portfolio Value:
                             var price = subscription.RealtimePrice;
 
+                            var now = DateTime.UtcNow.ConvertFromUtc(security.Exchange.TimeZone);
+                            if (!security.Exchange.Hours.IsOpen(now, security.IsExtendedMarketHours)) continue;
```
jaredbroad: Does security.Exchange.ExchangeOpen not work since no internal local frontier updates?

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1050,11 +1058,14 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
                     }
                 }
 
-                //Sample the portfolio value over time for chart.
-                SampleEquity(time, Math.Round(_algorithm.Portfolio.TotalPortfolioValue, 4));
+                if (canSampleEquity)
```
jaredbroad: The aim was to not update real time price; not stop sampling. Stopping sampling will stop chart movement as well and make the chart not reflect market closed periods (pros and cons of that; so for now lets keep current behavior of showing market closed periods).

Issue #520, opened 2016-08-09
Tradier failed to get cash balance


jaredbroad:
"Object reference not set to an instance of an object" exceptions when fetching cash from a live trading Tradier brokerage algorithm with their new API.

jaredbroad:
Fixed by commit 97ce555

PR #523, opened 2016-08-09
changed tradier accountID type from int to string

devalkeralia:
Changed the dataType of the Account ID for Tradier from an 'int' to a 'string' due to the changes made in the API by tradier.


devalkeralia:
Fixes #520

Issue #542, opened 2016-08-24
Algorithm keeps receiving quotes after the stock is delisted


quant1729:
Hello, 

I have faced this interesting issue. I run DelistingEventsAlgorithm locally with provided input files (I am using AAPL, SPY) to see delistings happenning (I modified map files for this). 
In DelistingEventsAlgorithm.cs I changed just initialization method: 
        public override void Initialize()
        {
            SetStartDate(2015, 07, 3);  //Set Start Date
            SetEndDate(2015, 07, 20);    //Set End Date
            SetCash(100000);             //Set Strategy Cash
            // Find more symbols here: http://quantconnect.com/data
            AddSecurity(SecurityType.Equity, "AAPL", Resolution.Daily);
            AddSecurity(SecurityType.Equity, "SPY", Resolution.Daily);
        }

What I noticed is that when the file is over and there is no more data in it, the algo keeps receiving the forward filled prices up to the EndDate. So e.g. if I say in algo "SetEndDate(2015, 07, 20)" and there is no data for SPY after 2015-07-15 my algo will still be getting last price 210.63 on 16th, 17th.. 20th of July ..even after receiving the delisting event. 

Modified map file for SPY to emulate delisting:
[spy.csv.txt](https://github.com/QuantConnect/Lean/files/435769/spy.csv.txt)

Screenshot:
![image](https://cloud.githubusercontent.com/assets/11757550/17949998/a16a2f24-6a58-11e6-8789-eae45560ab40.png)

jaredbroad:
Fixed in 5b3227ef501bc687e28231f26e9007f080d943f0

PR #545, opened 2016-08-26
Delistings bugfix

AlexCatarino:
On SubscriptionDataReader.TryGetNextDate, the day after the delisting was not been considered an event date, thus the algorithm kept receiving quotes after the security was delisted.
Ref #542

### DIFF
--- a/Common/Interfaces/IAlgorithm.cs
+++ b/Common/Interfaces/IAlgorithm.cs
@@ -450,6 +450,13 @@ namespace QuantConnect.Interfaces
         Security AddSecurity(SecurityType securityType, string symbol, Resolution resolution, string market, bool fillDataForward, decimal leverage, bool extendedMarketHours);
 
         /// <summary>
+        /// Removes the security with the specified symbol. This will cancel all
+        /// open orders and then liquidate any existing holdings
+        /// </summary>
+        /// <param name="symbol">The symbol of the security to be removed</param>
+        bool RemoveSecurity(Symbol symbol);
+
+        /// <summary>
         /// Set the starting capital for the strategy
         /// </summary>
         /// <param name="startingCash">decimal starting capital, default $100,000</param>
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -141,7 +141,7 @@ namespace QuantConnect.Lean.Engine
             var settlementScanFrequency = TimeSpan.FromMinutes(30);
             var nextSettlementScanTime = DateTime.MinValue;
 
-            var delistingTickets = new List<OrderTicket>();
+            var delistings = new List<Delisting>();
 
             //Initialize Properties:
             _algorithmId = job.AlgorithmId;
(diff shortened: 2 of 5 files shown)


Common/Interfaces/IAlgorithm.cs
```diff
@@ -450,6 +450,13 @@ TradeBuilder TradeBuilder
         Security AddSecurity(SecurityType securityType, string symbol, Resolution resolution, string market, bool fillDataForward, decimal leverage, bool extendedMarketHours);
 
         /// <summary>
+        /// Removes the security with the specified symbol. This will cancel all
+        /// open orders and then liquidate any existing holdings
+        /// </summary>
+        /// <param name="symbol">The symbol of the security to be removed</param>
+        bool RemoveSecurity(Symbol symbol);
```
StefanoRaggi: This should have been here since the `public QCAlgorithm.RemoveSecurity(Symbol)` method was added.
RemoveSecurity isn't used internally up to now, but is currently used in `AddRemoveSecurityRegressionAlgorithm`.

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -345,13 +345,17 @@ private ISubscriptionEnumeratorFactory GetEnumeratorFactory(SubscriptionRequest
...
                 {
-                    // Trigger universe selection when security added manually after Initialize
+                    // Trigger universe selection when security added/removed after Initialize
                     var universe = (UserDefinedUniverse) request.Universe;
                     universe.CollectionChanged += (sender, args) =>
                     {
-                        if (args.NewItems == null || _frontierUtc == DateTime.MinValue) return;
+                        IList items =
+                            args.Action == NotifyCollectionChangedAction.Add ? items = args.NewItems :
+                            args.Action == NotifyCollectionChangedAction.Remove ? items = args.OldItems : null;
```
StefanoRaggi: This assignment can be simplified a bit:
```
 var items =
     args.Action == NotifyCollectionChangedAction.Add ? args.NewItems :
     args.Action == NotifyCollectionChangedAction.Remove ? args.OldItems : null;

```

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -501,12 +501,16 @@ private Subscription CreateUniverseSubscription(SubscriptionRequest request)
...
                 enumerator = enqueueable;
 
-                // Trigger universe selection when security added manually after Initialize
+                // Trigger universe selection when security added/removed after Initialize
                 userDefined.CollectionChanged += (sender, args) =>
                 {
-                    if (args.NewItems == null || _frontierUtc == DateTime.MinValue) return;
+                    IList items =
+                           args.Action == NotifyCollectionChangedAction.Add ? items = args.NewItems :
+                           args.Action == NotifyCollectionChangedAction.Remove ? items = args.OldItems : null;
```
StefanoRaggi: Same as above comment

Engine/AlgorithmManager.cs
```diff
@@ -846,7 +860,11 @@ private static void HandleDelistedSymbols(IAlgorithm algorithm, Delistings newDe
                 {
                     Log.Trace("AlgorithmManager.Run(): Security delisting warning: " + delisting.Symbol.ToString());
                     var security = algorithm.Securities[delisting.Symbol];
-                    if (security.Holdings.Quantity == 0) continue;
+                    if (security.Holdings.Quantity == 0)
+                    {
+                        algorithm.RemoveSecurity(delisting.Symbol);
```
StefanoRaggi: If we call RemoveSecurity here (on a delisting warning event), will we receive the delisting event later ?

AlexCatarino: This is wrong. Doing that, we don't allow trades on the last day the security is listed.

Engine/DataFeeds/FileSystemDataFeed.cs
mchandschuh: Be careful with nested ternary operators... They can be easily misread. Code readability should always trump code brevity.

Issue #143, opened 2015-07-19
How do I create an ATR indicator from a RenkoConsolidator source


sblanchard:
In the RegisterIndicator method it sees the outputType as Tick and throws an exception.
Is this possible to?

mchandschuh:
In order to compute the AverageTrueRange, a TradeBar input value is
required. If you'd like to use a RenkoConsolidator's output as the input
for the ATR, you'll need to do some projection from the RenkBar ->
TradeBar, or, you could write a new version of the ATR that takes RenkoBar
as an input value. Here's some code that will take the output of a
RenkoConsolidator and pas it through to the AverageTrueRange indicator. The
first two lines create and plot the usual ATR14. The next several lines
create and plot the ATR14 of the consolidated renko output. An important
note here: The AverageTrueRange of a Renko bar should always be in the
neighborhood of the RenkoBarSize. By definition, the RenkoBar is a range
bound bar, not a time bound bar, so the true range of bars for which range
is already defined, will be about the size of the bar!

I hope these code snippets help!

-Michael

            // compute the 14 period average true range of daily prices
            var atr14 = ATR(symbol, 14);
            PlotIndicator(symbol, atr14);



            // comute the 14 period average true range of $5 renko bars

            // define a $5 renko bar consolidator
            var renko = new RenkoConsolidator(5m);
            SubscriptionManager.AddConsolidator(symbol, renko);

            // define our ATR indicator
            var atr = new AverageTrueRange(14);

            // each time we get a new renko bar, this event will be fired,
            // we'd like to update our ATR indicator with this data
            renko.DataConsolidated += (sender, args) =>
            {
                // project the output renko bar into a trade bar and pass
that
                // into our atr indicator
                atr.Update(new TradeBar
                {
                    Symbol = args.Symbol,
                    Time = args.Start,
                    Period = (args.End - args.Start),
                    Close = args.Close,
                    High = args.High,
                    Low = args.Low,
                    Open = args.Open,
                    Volume = args.Volume,
                    Value = args.Close
                });
            };

            PlotIndicator(symbol, atr);

PR #546, opened 2016-08-27
Renko Bar w/Gap Support and Consolidator Updates + Tests

squideyes:
The WickoConsolidator is a "Renko with Wicks" consolidator, with 13 unit tests and a sample algorithm.  

For more info about Wickos, please check out https://squideyes.com/2016/08/18/dont-renko-wicko/.  Wicko's are Renko's with wicks (I know, my cleverness in naming knows no bounds!) The main reason I made an entirely new set of classes and tests is that I didn't want to jigger around with the somewhat-broken RenkoConsolidator (which doesn't handle multi-brick moves and reversals correctly!) since people may have taken a hard dependency on it.  The RenkoConsolidator is indeed broken, but depending on trading style brick size a trader may not notice or care!  Anyway, feel free to accept my PR, fix the existing consolidator or maybe merge them together.


Common/Data/Consolidators/WickoConsolidator .cs
```diff
@@ -0,0 +1,203 @@
+﻿/*
```
StefanoRaggi: This source file appears to be mistakenly included in the commit

Common/Data/Consolidators/WickoConsolidator.cs
```diff
@@ -0,0 +1,301 @@
...
+        private bool firstTick = true;
+        private WickoBar lastWicko = null;
+
+        private DateTime openOn;
+        private DateTime closeOn;
+        private decimal openRate;
+        private decimal highRate;
+        private decimal lowRate;
+        private decimal closeRate;
+
```
StefanoRaggi: Please add leading underscores to private field names

Tests/Common/Data/WickoConsolidatorTests.cs
```diff
@@ -0,0 +1,586 @@
...
+using NUnit.Framework;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
+using QuantConnect.Indicators;
+using System.Collections.Generic;
+
+namespace QuantConnect.Tests.Common.Data
+{
+    [TestFixture]
+    public class WickoConsolidatorTests
```
StefanoRaggi: Nice set of tests!

Algorithm.CSharp/RenkoConsolidatorAlgorithm.cs
```diff
@@ -29,38 +29,38 @@ public class RenkoConsolidatorAlgorithm : QCAlgorithm
         /// </summary>
         public override void Initialize()
         {
-            SetStartDate(2012, 01, 01);
-            SetEndDate(2013, 01, 01);
+            SetStartDate(2014, 05, 01);
+            SetEndDate(2014, 05, 15);
```
StefanoRaggi: Is there a particular reason to update RenkoConsolidatorAlgorithm? If not, we can undo the change to this file

Tests/Common/Data/WickoConsolidatorTests.cs
```diff
@@ -28,559 +28,559 @@ public class WickoConsolidatorTests
         [Test]
         public void OutputTypeIsRenkoBar()
         {
-            var consolidator = new WickoConsolidator(10.0m, x => x.Value);
+            var _consolidator = new WickoConsolidator(10.0m, x => x.Value);
```
StefanoRaggi: We do not use the underscore prefix for local variables.
Please look at our contributing guidelines here:
https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md#code-style-and-testing

jaredbroad:
Thank you, its a really high quality PR @squideyes. I just have a couple of questions before merging it in:

The industry seems to have a few variants of Renko which are quite similar; but universally they're all called _Renko_.  If the existing RenkoBar class can't directly support the wicked case; please extend/ override the Renko class to make your variant. The current Renko bar also has wicks, why doesn't this support your needs?

I appreciate you don't want to step on existing user Renko usages but if Renko can't support it directly; could you please call your variant a "WickedRenko"; this has a google search presence and will appear when people type in "Renko".

Please replace the WickoTrend with a more generic BarDirection. Trend is a much wider topic; BarDirection is explicitly just the bar.

The obsolete End property can be removed. Its been there a long while. If we need a renko variant, please also remove the Obsolete message on the new class.

The "OnDataConsolidated" should be fired 0-1 times per input data point; the way your loop is setup it looks like it might be fired multiple times? (If so; it won't behave like people expect). Can you please also demonstrate an example why you need a loop and a simple online implementation won't work?

WickoConsolidator(decimal barSize, bool evenBars = true)
bool evenBars = true -- this appears unused and can be removed.

squideyes:
So there are two kinds of Renko; with and without wicks.  “With” is really fake Renko, but we like it nonetheless; without is the standard.  The whole “evenBars” thing, however is screwy, and in my opion should be removed.

 

In terms of the DataConsolidator firing multiple times, that is both essential and by design.  If it breaks the underlying code then you’ll either need to fix it or remove the RenkoConsolidator entirely, since the current implementation is clearly broken.

 

In Renko, a single tick can yield zero or more bricks.  For instance, in the ThreeRisingGapWickos() unit test, the first tick is 10.0 and the second is 14.0; a very plausible scenario.  Given a 1.0 brick size, that means that three OnDataConsolidated events will need to be fired on that one tick.

 

In terms updating the existing RenkoConsolidator to work as it should and getting rid of WickoConsolidator, that’s no problem.  

 

As to the rest, I’ll be happy to make the changes.  But before I go any further I need to hear back from you about OnDataConsolidated…

jaredbroad:
> "For instance, in the ThreeRisingGapWickos() unit test, the first tick is 10.0 and the second is 14.0; a very plausible scenario.  Given a 1.0 brick size, that means that three OnDataConsolidated events will need to be fired on that one tick."
Great explanation thank you, that makes perfect sense. Please be sure to put a detailed comment to explain this. I might have missed that comment :)

> In terms updating the existing RenkoConsolidator to work as it should and getting rid of WickoConsolidator, that’s no problem.
Thank you this would be preferred thank you. 

Since the charting isn't block/candle specific we can include all the data in the "model" and then worry about charting it after. Later we could feasibly we could make a QCAlgorithm PlotRenko/PlotRenkoWicks helper methods to trim the candles.

squideyes:
Cool.  So do you need me to do anything or is it in your hands?

jaredbroad:
Please do update it @squideyes! You're the expert now :)

StefanoRaggi:
I agree on updating the existing RenkoConsolidator instead of adding a new one, here are a few more thoughts:

**Renko variants**
A `enum RenkoType { Classic, Wicked }` could be used as a constructor parameter, so in the future if needed we could add other variants.

**Gap handling**
I think the filling of gaps with many bricks/bars at the same time should be optional/user choice. The reason I would prefer not to have consecutive bricks (so extra OnDataConsolidated events) is that (in backtesting) market orders filled on the closes of the earlier (fake) ones are unrealistic and an algorithm using this setting would likely get very different results in live trading. A new bool constructor parameter could allow to enable/disable this feature.
 
**Indicator support** (issues #331 and #143)
Currently this consolidator is generating objects of class RenkoBar which are not compatible with many of our indicators that require TradeBar inputs.
My question is: do we really need RenkoBars ? Couldn't we just generate TradeBars instead ? I know it is a minor breaking change, but I think the benefit in doing so is important and now could be the right time to do it. I guess @jaredbroad will have to decide on this :)

jaredbroad:
We shouldn't remove Renko entirely but agree RenkoBar.cs should be rewritten to extend TradeBar, overriding the GetSource/Reader methods to throw; and adding the BrickSize, RekoType property. That should maintain backwards compatibility with existing implementations.

I think we should still iterate though/throw multiple events since this will be used for charting in the majority of cases. People using it will just need to be aware of the gap case since that is fundamental to Renko. For now lets implement it as the hard/gap case then worry about disabling that feature (as an ease of use feature) once there's demand.

squideyes:
Thanks for the clarifications.  I started to coalesce things but I ran out of time.  I won’t be able to finalize the code before the weekend.  Sorry for the delay…

jaredbroad:
Thank you @squideyes for the rework & tests. It fits nicely with the other LEAN code now.

PR #550, opened 2016-09-01
Decouple QuantConnect.Views build configuration

andrewhart098:
The purpose of this pull request is to decouple the QuantConnect.Views build configuration from the rest of the solution. This is achieved by changing QuantConnect.Views 'Output Type' to 'Console Application' and removing the QuantConnect.Views reference in the QuantConnect.Launcher project. 

This will enable QuantConnect.Views to run as a separate 32 bit process (which it must because of its dependency on GeckoFX) and allow the rest of the solution to run as a 64 bit process. 

The interprocess communication between Lean and the UI is achieved over tcp sockets and leverages the NetMQ library - a 100% native C# implementation of ZeroMQ for .NET. NetMQ was chosen because of it’s ability to asynchronously send messages over tcp, while preserving the message order. 

A new message handler, StreamingMessageHandler, is added which processes the messages in-line using a NetMQ Push Socket. This is now the default message handler for the desktop environments.  The port on which the NetMQ communicates, as well as the location of the UI executable, is now configurable in config.json.

The UI uses a NetMQ Pull Socket in a new class called DesktopClient.cs.  The desktop client uses an instance of a new interface, IDesktopMessageHandler, to push messages to the UI.  LeanWinForms.cs now implements this interface.

### DIFF
--- a/Launcher/Program.cs
+++ b/Launcher/Program.cs
@@ -87,11 +87,7 @@ namespace QuantConnect.Lean.Launcher
 
             if (environment.EndsWith("-desktop"))
             {
-                Application.EnableVisualStyles();
-                var messagingHandler = leanEngineSystemHandlers.Notify;
-                var thread = new Thread(() => LaunchUX(messagingHandler, job));
-                thread.SetApartmentState(ApartmentState.STA);
-                thread.Start();
+
             }
 
             // log the job endpoints
@@ -135,17 +131,6 @@ namespace QuantConnect.Lean.Launcher
                 leanEngineAlgorithmHandlers.Dispose();
                 Log.LogHandler.Dispose();
             }
-        }
-
-        /// <summary>
-        /// Form launcher method for thread.
-        /// </summary>
-        static void LaunchUX(IMessagingHandler messaging, AlgorithmNodePacket job)
-        {
-            //Launch the UX
-            //var form = Composer.Instance.GetExportedValueByTypeName<Form>("desktop-ux-classname");
-            var form = new Views.WinForms.LeanWinForm(messaging, job);
-            Application.Run(form);
         }
     }
 }
--- a/Launcher/Program.cs
+++ b/Launcher/Program.cs
@@ -16,6 +16,8 @@
 
 using System;
 using System.ComponentModel.Composition;
+using System.Diagnostics;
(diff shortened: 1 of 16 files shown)


jaredbroad:
Replacement for PR: https://github.com/QuantConnect/Lean/pull/494

jameschch:
So glad this was completed. I had made some progress with a version running over http but this is a much better approach. One of the things that will be possible now the UI is a separate process is switching to a chromium based browser engine that may play nicely in linux, or perform better in backtest.

@andrewhart098 I'm interested in NetMQ. How does this compare to RabbitMQ and MSMQ? How do you think it will scale in an enterprise?

andrewhart098:
@jameschch - One interesting option for building a cross-platform, desktop client is [electron](http://electron.atom.io/).  There are already zeromq bindings for  [node](https://github.com/JustinTulloss/zeromq.node).  It would be great if the client could be connected to multiple instances of Lean and have different tabs for each backtest.  

Compared to other messaging queues, ZeroMQ is unique because it is a broker-less model which means that unlike most other products, you don’t need to install and run a message queuing server/broker. Simply reference the NetMQ library on NuGet.

The performance numbers for zeromq can be found [here](http://zeromq.org/area:results). I don't know what your standards for perf are, but I think a throughput of over 9000 Mb/s for 1Mb messages with just over 1 millisecond in latency is really good.  From what I've read, brokerless is generally way faster than brokered (e.g. RabbitMQ, MSMQ). Brokerless is really the only messaging queue option in this situation anyways.

That being said, without a broker – it’s up to the application to handle things when they go wrong.  I’m not entirely clear what the requirements for Lean are in an enterprise setting, but I think that with the flexibility a separate client process provides, there is a simple solution to make a really robust desktop client using zeromq for messaging.

jameschch:
I love the fact that zeromq is brokerless. This aligns well with the deployability of a microservice architecture, which is the main reason for reignited interest in messaging.

In terms of the desktop client I was trying to think of a way to provide a ui service to local or remote lean instances for monitoring and alerting. As you say it's conceptually tied to something in particular. I'd like it if it listened to whoever was talking. Standing it up as a service can be straightforward, tabbing multiple input streams sounds more laborious.

Issue #547, opened 2016-08-30
Market Hours Database Holidays


jaredbroad:
FXCM and Oanda aren't open on January 1st, 

market-hours-database.json should be updated to include fxcm and oanda holidays.

richardcurteis:
Fix submitted.
https://github.com/QuantConnect/Lean/pull/551

jaredbroad:
Fixed by 727c3c1

PR #551, opened 2016-09-01
Added FXCM holidays

richardcurteis:
Fix for [https://github.com/QuantConnect/Lean/issues/547](url)
Added FXCM specific holidays.
They are closed 25th/26th of December and 1st/2nd January. Including when those dates fall on normal trading hours.
Screenshot of this confirmation from FXCM support is below.
Oanda holidays was omitted as they are open on all days that fall within normal trading hours, screenshot below.

**FXCM**
fxcm conversation

**Oanda**
oanda conversation


jaredbroad:
Thank you for the PR @richardcurteis

PR #558, opened 2016-09-15
Implements IPriceVariationModel

AlexCatarino:
- Makes a IPriceVariationModel which takes the security object and returns a decimal variation.
- Implements the RoundedPrice using the IPriceVariationModel to control the rounding value.

### DIFF
--- a/Common/Securities/Forex/Forex.cs
+++ b/Common/Securities/Forex/Forex.cs
@@ -47,7 +47,8 @@ namespace QuantConnect.Securities.Forex
                 new ImmediateSettlementModel(),
                 Securities.VolatilityModel.Null,
                 new SecurityMarginModel(50m),
-                new ForexDataFilter()
+                new ForexDataFilter(),
+                new SecurityPriceVariationModel()
                 )
         {
             Holdings = new ForexHolding(this);
@@ -78,7 +79,8 @@ namespace QuantConnect.Securities.Forex
                 new ImmediateSettlementModel(),
                 Securities.VolatilityModel.Null,
                 new SecurityMarginModel(50m),
-                new ForexDataFilter()
+                new ForexDataFilter(),
+                new SecurityPriceVariationModel()
                 )
         {
             Holdings = new ForexHolding(this);
--- a/Common/Securities/Security.cs
+++ b/Common/Securities/Security.cs
@@ -297,6 +297,19 @@ namespace QuantConnect.Securities
         }
 
         /// <summary>
+        /// Customizable price variation model used to define the minimum price variation of this security.
+        /// By default minimum price variation is a constant find in the symbol-properties-database.
+        /// </summary>
+        /// <seealso cref="AdjustedPriceVariationModel"/>
+        /// <seealso cref="SecurityPriceVariationModel"/>
+        /// <seealso cref="EquityPriceVariationModel"/>
+        public IPriceVariationModel PriceVariationModel
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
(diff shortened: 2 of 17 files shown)


jaredbroad:
Nice work Alex, please add "NoRounding Model" for the backtesting adjusted prices case when the price factor is less than 1. Configure the security to use the no-rounding model by default in backtesting.

Common/Securities/SecurityPriceVariationModel.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// for use in defining the minimum price variation.
+    /// </summary>
+    public class SecurityPriceVariationModel : IPriceVariationModel
+    {
+        /// <summary>
+        /// Get the minimum price variation from a security
+        /// </summary>
+        /// <param name="security">Security which we want the minimum price variation from</param>
+        /// <returns>Decimal minimum price variation of a given security</returns>
+        public virtual decimal GetMinimumPriceVariation(Security security)
```
jaredbroad: This seems unnecessary. The way to get minimum variation is always just "security.SymbolProperties.MinimumPriceVariation" -- in what cases does it change?

jaredbroad: Or even better the rounding interface is unnecessary as its a by product of the minimum variation. Rounding isn't something which should ever change security to security? Its a convenience method for when the user's prices aren't set correctly. Minimum variation 0 would be a pass-through for adjusted pricing,

Common/Securities/Security.cs
```diff
@@ -618,6 +632,32 @@ public void SetDataNormalizationMode(DataNormalizationMode mode)
...
+                SetLiveModePriceVariationModel(true);
+            }
+        }
+
+        /// <summary>
+        /// In LiveMode, we set up the security price variation model to assure that 
+        /// order prices are rounded to meet brokerages precision requirements
+        /// </summary>
+        /// <param name="liveMode">True if live mode</param>
+        public void SetLiveModePriceVariationModel(bool liveMode)
```
jaredbroad: Please remove this and when the security is constructed setup the PriceVariationModel. Follow the patterns established by all the other model properties.

Common/Securities/Forex/Forex.cs
```diff
@@ -47,7 +47,8 @@ public Forex(SecurityExchangeHours exchangeHours, Cash quoteCurrency, Subscripti
                 new ImmediateSettlementModel(),
                 Securities.VolatilityModel.Null,
                 new SecurityMarginModel(50m),
-                new ForexDataFilter()
+                new ForexDataFilter(),
+                new AdjustedPriceVariationModel()
```
jaredbroad: Do you think Forex/CFD types should use the SecurityPriceVariationModel()?

jaredbroad: (In backtesting FX/CFD is not using adjusted pricing)

AlexCatarino: If we use SecurityPriceVariationModel() as default, we will change results from previous backtests, since it will start rounding limit/stop orders. Is that what we want?

StefanoRaggi: I think this should be done because backtests will now be more accurate and behavior same as in live trading, although there will be some small difference in overall results (may have to check and update some regression stats as well)

jaredbroad: Yes movement towards more accuracy is positive even if it changes previous backtests.

jaredbroad:
Nice work Alex! Ready to merge in!

PR #560, opened 2016-09-19
API Updates

andrewhart098:
This PR adds the ability to reach two new endpoints introduced in QC API v2.

The first of the endpoints allows users to download data they have purchased from QuantConnect.  The data is stored to disk in the same folder-hierarchy as Lean uses to store data.

The second endpoint allows users to read logs of live running algorithms.

### DIFF
--- a/Api/Api.cs
+++ b/Api/Api.cs
@@ -314,12 +314,7 @@ namespace QuantConnect.Api
                 return false;
 
             // Save csv in same folder heirarchy as Lean
-            string path;
-            
-            if (resolution == Resolution.Daily || resolution == Resolution.Hour)
-                path = Config.Get("data-folder") + "/" + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() + "/" + resolution.ToLower() + "/" + symbol.Value.ToLower() + ".zip";
-            else
-                path = Config.Get("data-folder") + "/" + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() + "/" + resolution.ToLower() + "/" + symbol.Value.ToLower() + "/" + date.ToString("yyyyMMdd") + "_quote.zip";
+            var path = CreatePathForDownloadedData(symbol, resolution, date);
 
             // Make sure the directory exist before writing
             (new FileInfo(path)).Directory.Create();
@@ -415,6 +410,28 @@ namespace QuantConnect.Api
             {
                 yield return segment;
             }
+        }
+
+        /// <summary>
+        /// Creates the path for the downloaded data to be saved in the same folder heirarchy as Lean
+        /// </summary>
+        /// <param name="symbol">Symbol of security of which data will be requested.</param>
+        /// <param name="resolution">Resolution of data requested.</param>
+        /// <param name="date">Date of the data requested.</param>
+        /// <returns>Path that specifies where data will be downloaded</returns>
+        private string CreatePathForDownloadedData(Symbol symbol, Resolution resolution, DateTime date)
+        {
+            string path;
+
+            if (resolution == Resolution.Daily || resolution == Resolution.Hour)
+                path = Config.Get("data-folder") + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
+                       "/" + resolution.ToLower() + "/" + symbol.Value.ToLower() + ".zip";
+            else
+                path = Config.Get("data-folder") + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
+                       "/" + resolution.ToLower() + "/" + symbol.Value.ToLower() + "/" + date.ToString("yyyyMMdd") +
+                       "_quote.zip";
+
+            return path;
         }
 
         /// <summary>
--- a/Api/Api.cs
+++ b/Api/Api.cs
@@ -275,6 +275,25 @@ namespace QuantConnect.Api
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


Api/Api.cs
```diff
@@ -358,6 +432,28 @@ public virtual IEnumerable<MarketHoursSegment> MarketToday(DateTime time, Symbol
...
+        /// <returns>Path that specifies where data will be downloaded</returns>
+        private string CreatePathForDownloadedData(Symbol symbol, Resolution resolution, DateTime date)
+        {
+            string path;
+
+            if (resolution == Resolution.Daily || resolution == Resolution.Hour)
+                path = Config.Get("data-folder") + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
+                       "/" + resolution.ToLower() + "/" + symbol.Value.ToLower() + ".zip";
+            else
+                path = Config.Get("data-folder") + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
```
jaredbroad: Please pass in a "data-folder" variable in the constructor, and save it to the class, so its not loading from config each time.

Common/Interfaces/IApi.cs
```diff
@@ -132,7 +132,32 @@ public interface IApi : IDisposable
...
         LiveList LiveList();
-        
+
+        /// <summary>
+        /// Gets the logs of a specific live algorithm 
+        /// </summary>
+        /// <param name="projectId">Project Id of the live running algorithm</param>
+        /// <param name="algorithmId">Algorithm Id of the live running algorithm</param>
+        /// <returns>List of strings that represent the logs of the algorithm</returns>
+        LiveLog ReadLiveLogs(int projectId, string algorithmId);
```
jaredbroad: Please pass in a start, end and format specifier:

format - raw or json
start (api takes timestamp start)
end (api takes timestamp end)

The C# method should have a datetime input which is converted to -> timestamp in unix sec. Defaulting the start to 0 and end to unix-now.

Api/Api.cs
andrewhart098: There is no constructor for the Api class.  I will instead pass it in to the Initialize() method and save it off there. I think this is what you meant.

Api/Api.cs
```diff
@@ -443,17 +453,28 @@ private string CreatePathForDownloadedData(Symbol symbol, Resolution resolution,
             string path;
 
             if (resolution == Resolution.Daily || resolution == Resolution.Hour)
-                path = Config.Get("data-folder") + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
+                path = _dataFolder + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
                        "/" + resolution.ToLower() + "/" + symbol.Value.ToLower() + ".zip";
             else
-                path = Config.Get("data-folder") + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
+                path = _dataFolder + symbol.ID.SecurityType.ToLower() + "/" + symbol.ID.Market.ToLower() +
```
jaredbroad: Path.Combine() is a safe cross platform alternative here; or even better use the:
"LeanData.GenerateRelativeZipFilePath" method in the toolbox LEAN Data Writer,

Api/Api.cs
```diff
@@ -443,17 +453,28 @@ private string CreatePathForDownloadedData(Symbol symbol, Resolution resolution,
...
         }
 
         /// <summary>
+        /// Convert C# dateTime to Unix Epoch DateTime
+        /// </summary>
+        /// <param name="dateTime">DateTime to be converted</param>
+        /// <returns>The total number of seconds since Jan 1, 1970</returns>
+        private static long DateTimeToUnixTimestamp(DateTime dateTime)
+        {
+            return (long) (dateTime - new DateTime(1970, 1, 1, 0, 0, 0, 0, System.DateTimeKind.Utc)).TotalSeconds;
```
jaredbroad: We have a static method Time.DateTimeToUnixTimeStamp()

Tests/API/Api.cs
```diff
@@ -256,8 +254,8 @@ public void Download_AndSave_DataCorrectly()
             Assert.IsTrue(downloadedMinuteData);
             Assert.IsTrue(downloadedDailyData);
             
-            Assert.IsTrue(File.Exists(Config.Get("data-folder") + "forex/oanda/daily/eurusd.zip"));
-            Assert.IsTrue(File.Exists(Config.Get("data-folder") + "forex/oanda/minute/eurusd/20131011_quote.zip"));
+            Assert.IsTrue(File.Exists(_dataFolder + "forex/oanda/daily/eurusd.zip"));
```
jaredbroad: Perhaps its worth deleting the file if it exists already? Otherwise its might read a false positive

Api/Api.cs
```diff
@@ -271,6 +276,85 @@ public LiveList LiveList()
...
+            var path = Path.Combine(_dataFolder, LeanData.GenerateRelativeZipFilePath(symbol.Value, symbol.ID.SecurityType, symbol.ID.Market, date, resolution));
+
+            // Make sure the directory exist before writing
+            (new FileInfo(path)).Directory.Create();
+
+            // Download and save the data
+            var uri     = new Uri(link.DataLink);
+            var client  = new RestClient(uri.Scheme + "://" + uri.Host);
+            var request = new RestRequest(uri.PathAndQuery, Method.GET);
+            client.DownloadData(request).SaveAs(path);
```
jaredbroad: Nice

PR #571, opened 2016-09-27
Makes IAlgorithm Debug/Error/LogMessages a ConcurrentQueue

AlexCatarino:
IAlgorithm DebugMessage, ErrorMessage and LogMessage are now ConcurrentQueue<string> instead of List<string> because List.Add is not thread safe.
Copying a list (using ToList()) at LiveTradingResultHandler while adding an element to it at QCAlgorithm caused a runtime error.

### DIFF
--- a/Engine/Results/BacktestingResultHandler.cs
+++ b/Engine/Results/BacktestingResultHandler.cs
@@ -865,18 +865,36 @@ namespace QuantConnect.Lean.Engine.Results
                 }
             }
 
-            //Send out the debug messages:
-            _algorithm.DebugMessages.ForEach(x => DebugMessage(x));
-            _algorithm.DebugMessages.Clear();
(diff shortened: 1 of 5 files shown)


Engine/Results/BacktestingResultHandler.cs
```diff
@@ -865,18 +865,36 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
                 }
             }
 
-            //Send out the debug messages:
-            _algorithm.DebugMessages.ForEach(x => DebugMessage(x));
-            _algorithm.DebugMessages.Clear();
+            //Send out the debug messages:            
+            while (_algorithm.DebugMessages.Count > 0)
```
jaredbroad: This could potentially hang forever. We should make a timeout for the loops or restructure it to copy to a list first before sending.

PR #572, opened 2016-09-27
Ib api migration

devalkeralia:
Migrated the IB Api used in the platform from the old 3rd party Krs.Ats.IBNet implementation to the newer official TWS IB Api.

I have migrated the code from the event handlers used in the older implementation to the callbacks used in the newer implementation. 

It uses the latest version of the TWS Api downloaded from [http://interactivebrokers.github.io/](url)

The documentation for the TWS api can be found here [https://www.interactivebrokers.com/en/software/api/api.htm](url)

Do let me know if any bugs are spotted or any implementation needs to be changed.


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.Callbacks.cs
```diff
@@ -0,0 +1,741 @@
...
+            Log.Trace("InteractiveBrokersBrokerage.Callbacks.nextValidId(): " + orderId);
+        }
+
+        /// <summary>
+        /// Returns the active account list
+        /// </summary>
+        /// <param name="accountsList">List of accounts</param>
+        public virtual void managedAccounts(string accountsList)
+        {
+            Console.WriteLine("Account list: " + accountsList + "\n");
```
jaredbroad: Log.Trace please

Brokerages/InteractiveBrokers/HistoricalDataDetails.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+
+        }
+
+        /// <summary>
+        /// The ticker Id of the request to which this bar is responding.
+        /// </summary>
+        public int RequestId
+        {
+            get { return requestId; }
```
jaredbroad: All these can be autoproperties

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.Callbacks.cs
```diff
@@ -0,0 +1,741 @@
...
+
+        /// <summary>
+        /// Acknowledges the closing of the connection
+        /// </summary>
+        public virtual void connectionClosed()
+        {
+            Log.Debug("TWS Connection Closed.");
+        }
+
+        public virtual void accountSummary(int reqId, string account, string tag, string value, string currency)
```
jaredbroad: Can you please group all the empty methods at the bottom of the file?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.Callbacks.cs
```diff
@@ -0,0 +1,741 @@
...
+        /// <param name="order">Order</param>
+        /// <param name="orderState">The state of the order</param>
+        public void openOrder(int orderId, Contract contract, Order order, OrderState orderState)
+        {
+//            var orders = new List<Orders.Order>();
+            // convert IB order objects returned from RequestOpenOrders\
+            var convertedOrder = ConvertOrder(order, contract);
+            if (convertedOrder == null)
+            {
+                Console.WriteLine("Conversion failed. NULL!!");
```
jaredbroad: Log.Error + proper error message

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -145,36 +167,17 @@ public InteractiveBrokersBrokerage(IOrderProvider orderProvider, ISecurityProvid
...
-            _client.UpdatePortfolio += HandlePortfolioUpdates;
-            _client.OrderStatus += HandleOrderStatusUpdates;
-            _client.UpdateAccountValue += HandleUpdateAccountValue;
-            _client.Error += HandleError;
-            _client.TickPrice += HandleTickPrice;
-            _client.TickSize += HandleTickSize;
-            _client.CurrentTime += HandleBrokerTime;
-
+            _ibClient = new EClientSocket(this);
+            Thread.Sleep(10000);
```
jaredbroad: Can't we do this with a callback? Blind sleeping 10sec isn't ideal

Launcher/config.json
```diff
@@ -42,15 +42,15 @@
...
-  "ib-port": "4001",
-  "ib-agent-description": "Individual",
-  "ib-use-tws": false,
-  "ib-tws-dir": "C:\\Jts",
-  "ib-controller-dir": "C:\\IBController",
+ "ib-account": "",
+ "ib-user-name": "",
+ "ib-password": "",
+ "ib-host": "127.0.0.1",
+ "ib-port": "7496",
```
jaredbroad: Shouldn't need to change this

Launcher/config.json
```diff
@@ -6,7 +6,7 @@
   // two predefined environments, 'backtesting' and 'live', feel free
   // to add more!
 
-  "environment": "backtesting",
+  "environment": "live-interactive",
```
jaredbroad: Please leave this line as it was.

Brokerages/packages.config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="IKVM-WithExes" version="7.3.4830.1" />
-  <package id="Krs.Ats.IBNet" version="9.66.0.31" targetFramework="net45" />
+  <package id="Krs.Ats.IBNet" version="9.66.0.30" targetFramework="net45" />
```
jaredbroad: Krs.Ats.IBNet should go away and your TWSLib will take its place, please rename the TWS lib to a more appropriate name reflecting the name of the ib github project. (TWS is just Trader Work Station -- we should make sure this works with the Gateway as well)

devalkeralia: is the name IBApi.dll fine?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -334,55 +315,31 @@ public override List<Cash> GetCashBalance()
...
-                }
-
-                // remove our event handlers
-                client.ExecDetails -= clientOnExecDetails;
-                client.ExecutionDataEnd -= clientOnExecutionDataEnd;
+                throw new TimeoutException("InteractiveBrokersBrokerage.GetExecutions(): Operation took longer than 1 second.");
             }
-
-            return details;
+            return _exectionDetailsList;
```
mchandschuh: I don't think this pattern is thread safe. Imagine multiple threads calling into this method, and then all the callbacks trying to set the same `_executionDetailsList`. My suggestion would be to use something similar to what was previously in place.

Brokerages/InteractiveBrokers/HistoricalDataDetails.cs
```diff
@@ -0,0 +1,169 @@
...
+*/
+
+using System;
+
+namespace QuantConnect.Brokerages.InteractiveBrokers
+{
+    /// <summary>
+    /// Historical Data Details
+    /// </summary>
+    [Serializable()]
```
mchandschuh: curiosity: Why is `[Serializable]` required?

Tests/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerageDataQueueHandlerTest.cs
```diff
@@ -23,7 +23,7 @@
 namespace QuantConnect.Tests.Brokerages.InteractiveBrokers
 {
     [TestFixture]
-    [Ignore("These tests require the IBController and IB TraderWorkstation to be installed.")]
+   // [Ignore("These tests require the IBController and IB TraderWorkstation to be installed.")]
```
mchandschuh: This is the likely cause of the travis build failure as well

devalkeralia: I had commented it for running the tests locally and forgot to put it back in. My bad.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
devalkeralia: would you suggest putting it inside a "using" block?

mchandschuh: Another solution is to implement the `EWrapper` in its own class and expose the callbacks as events. This would make the port look much closer to the existing implementation as well, by adding event handlers dedicated to a specific request while at the same time removing all the member variables which can cause concurrent heartburn :)

devalkeralia: I went the `ConcurrentDictionary` route.

mchandschuh: In addition, you'll need a reset event per request id. So maybe the dictionary value is a tuple or small private class containing the reset event and the result. The callback would get the reset event for the specific request id and flip the switch, allowing the calling code to proceed execution and know that the result is now populated.

devalkeralia: I have created one more property for the `ExecutionDetails` class for the reset event. Since the key is the request id, a reset event will be created per request id. The event will be "Set()" once it hits the `execDetailsEnd` gets a response indicating that the Executions have arrived for a particular request id.
Take a look at my latest commit.
Let me know if the implementation seems correct.

the other way could be to create the `ConcurrentDictionary` something like:
`ConcurrentDictionary<int, Tuple<ExecutionDetails, ManualResetEvent>>`

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -334,55 +315,29 @@ public override List<Cash> GetCashBalance()
...
-                Time = timeSince ?? DateTime.MinValue,
-                Side = side ?? IB.ActionSide.Undefined
+                Time = timeSince.ToString("yyyymmdd hh:mm:ss") ?? DateTime.MinValue.ToString("yyyymmdd hh:mm:ss"),
+                Side = side ?? ""
             };
-
-            var details = new List<IB.ExecDetailsEventArgs>();
-            using (var client = new IB.IBClient())
+            var client = new EClientSocket(this);
+            client.eConnect(_host, _port, IncrementClientID());
```
mchandschuh: does this new `client` ever get used?

devalkeralia:
its used to make the request for the executions.
used nowhere else.
it was in a "using" block earlier in the older implementation where it was
created and used only inside "using" to make request for executions.

i could just remove it and use the client already defined.
what would you suggest?

mchandschuh:
Take a look at this [diff](https://github.com/QuantConnect/Lean/compare/example/ib-api-migration-begin-diff...example/ib-api-migration).

This is a diff between what was your current implementation at the time I started (branch `example/ib-api-migration-begin-diff`) and a few examples of what I'm talking about (branch `example/ib-api-migration`).

If you look at yet another [diff](https://github.com/QuantConnect/Lean/compare/master...example/ib-api-migration), this one against `origin/master`, you'll see how similar the old and new implementation become using this pattern (There's lots of changes, so just find InteractiveBrokersBrokerage.cs). The key is that I've created something similar to the old client. I've called this class the `InteractiveBrokersClient` and his entire purpose is to take `EWrapper` methods and invoke events. These events can be bound to from the brokerage implementation. See `GetExecutions` for the direct method in question, but this issue applies to all usages of `_ibClient` and `Client` property. We have to be *very* careful with this code... it was written over countless hours/days/weeks and should be altered with *extreme* care. By allowing the code to look almost identical to the previous code, it will give us more confidence in the correctness of the migration.

I've also provided an example of binding callbacks to things that the brokerage needs at the class level (vs per request level such as `GetExecutions`). Such examples include the `CurrentTimeUtc` callback and the `NextValidId` callback.

Please reach out if you have any questions regarding the sample code provided. I did most of the heavy typing by stubbing out most of the `EventArgs` types and the architecture of the `EWrapper` implementation (`InteractiveBrokersClient`).

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
devalkeralia: Makes sense. This one is almost exactly like the old implementation.
I will change the code according to this new pattern.
it looks the same as the old one but with newer callbacks. Events remain the same. 
thanx for writing up the event args classes. 
I'll get to it.

jaredbroad:
Closing this for now as it need a large refactor; but lets keep at it and we'll merge it in soon.

PR #577, opened 2016-09-29
Live Algorithm Management with QC Api v2

andrewhart098:
This PR expands the C# wrapper for the QuantConnect Api v2 to enable the creation and management of algorithms deployed live on QuantConnect.  Classes are added that allows the configuration of live algorithms to trade using specific brokerages: Interactive Brokers, Tradier, Oanda and FXCM.

In addition to the tests added with this pull request, I successfully tested deploying a live algorithms that paper traded forex on FXCM and live algorithms that paper traded equities on Interactive Brokers. 

As for Oanda, QC only supports v1 accounts.  I was only able to sign up for a v2 account.  I have contacted customer support to see if I can get a v1 account.  I do not have a brokerage account with Tradier.  

Until I successfully paper trade forex on Oanda, I will leave this as WIP.


Common/API/ApiAlgorithmStatus.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace QuantConnect.API
+{
+    /// <summary>
+    /// An enum representing the possible status of a live running algorithm on QC
+    /// </summary>
+    public enum ApiAlgorithmStatus
```
jaredbroad: Perhaps AlgorithmStatus in common would work here,

andrewhart098: There are a couple values in AlgorithmStatus that have no equivalent in the Api.  In other words, the Api will reject certain values in AlgorithmStatus.  I thought it was better to create a new enum that only held values the Api will accept.

Common/API/LiveAlgorithmResults.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// <summary>
+        /// Represents data about the live running algorithm returned from the server
+        /// </summary>
+        public LiveResults LiveResults { get; set; }
+    }
+
+    /// <summary>
+    /// Holds information about the state and operation of the live running algorithm
+    /// </summary>
+    public class LiveResults
```
jaredbroad: Can we potentially reuse the "Common\Packets\LiveResult" class here? BacktestResult and LiveResult are almost identical in theory on the view side. Reusing the same classes would help future view implementations.

Potentially we could even do a slightly reshuffle to make the classes extended from a "AlgorithmResult" base common implementation.

Common/API/LiveAlgorithmSettings.cs
```diff
@@ -0,0 +1,299 @@
...
+        /// Type of server being used to run live algorithm
+        /// </summary>
+        [JsonProperty(PropertyName = "serverType")]
+        public string ServerType { get; private set; }
+
+        /// <summary>
+        /// The API expects the settings as part of a brokerage object
+        /// </summary>
+        [JsonProperty(PropertyName = "brokerage")]
+        public BaseLiveAlgorithmSettings Brokerage { get; private set; }
```
jaredbroad: Nice

Common/API/LiveAlgorithmSettings.cs
```diff
@@ -0,0 +1,299 @@
...
+        /// Contructor for live trading with FXCM
+        /// </summary>
+        /// <param name="user">Username associated with brokerage</param>
+        /// <param name="password">Password associated with brokerage</param>
+        /// <param name="brokerage">Name of the brokerage</param>
+        /// <param name="environment">'live'/'paper'</param>
+        /// <param name="account">Account id for brokerage</param>
+        public FXCMLiveAlogrithmSettings(string user,
+                                         string password,
+                                         BrokerageName brokerage,
```
jaredbroad: the "FXCMLiveAlogrithmSettings" can't accept any other brokerage names; this shouldn't be a constructor arg. Same for all the other *LiveAlogrithmSettings parameters
```base(user, password, BrokerageName.FxcmBrokerage, environment, account)"```

Tradier cannot do "paper" that environment can be removed there, Tradier "string lifetime," is always 86399 seconds

File: Common/API/LiveAlgorithmSettings.cs [hunk 5]
```diff
@@ -0,0 +1,299 @@
...
+        /// Account of the associated brokerage
+        /// </summary>
+        [JsonProperty(PropertyName = "account")]
+        public string Account { get; private set; }
+    }
+
+    /// <summary>
+    /// Default live algorithm settings
+    /// </summary>
+    public class DefaultLiveAlogrithmSettings : BaseLiveAlgorithmSettings
```
StefanoRaggi: Small typo in class name

andrewhart098:
Thank you everyone who reviewed my code so quickly.  I have successfully launched an algorithm that paper trades using Oanda as a brokerage using the api.  I will make the suggestions other contributors have mentioned above and resubmit my PR.

Common/API/LiveAlgorithmResults.cs
andrewhart098: @jaredbroad - I did look for an existing class in the solution before creating this one to deserialize the results.  It's such a minor difference, but LiveResult uses a struct called ChartPoint (Common\Charting.cs) that is incompatible with the response from the Api.  

ChartPoint is what holds the x and y values.  For some reason, when you fetch the results from the Api, the first y value in the series of x and y values is null.  This minor difference throws an error when you try to deserilize.  I tried making the "y" property nullable, but this would mean fixing code everywhere that uses this struct.  

I didn't like it either, but the easiest path to getting the Api response deserialized was to add this class.  I'm all ears for a better way though.

nooooooooo

jaredbroad: The nulls are to demark the start and endpoint of various deployments of the same algorithm. The null breaks the line so its not a constant and looks more like a dash "----   --  ---     -" etc.

Probably best we'll need a nullable int (wide spread impact) or custom deserializer (localized impact) for this like how we do custom Order type deserialization. Please see how wide the impact would be setting it to a nullable int & we can decide the right path. Setting to -1 or 0 would throw chart scales off; so the only other option would be to skip the point. 

Wishing we'd made ChartPoint a class now :) Then we could set the whole point to null.

Tests/API/ApiTests.cs
```diff
@@ -0,0 +1,725 @@
...
+using NUnit.Framework;
+using QuantConnect.Api;
+using QuantConnect.API;
+using QuantConnect.Brokerages;
+using QuantConnect.Configuration;
+using RestSharp.Extensions.MonoHttp;
+
+namespace QuantConnect.Tests.API
+{
+    [TestFixture, Category("TravisExclude")]
```
andrewhart098: Should I also add the 'Ignore' attribute because these tests require configured and active Brokerage accounts?

jaredbroad: Yes good idea

PR #580, opened 2016-09-30
Feature/daily quotes

fbertram:
these changes implement download of daily quotes for the Google downloader. The code includes 2 workarounds worth mentioning:

(1) Google seems to taint their data by replacing some quotes with '-'. The code tries its best to put reasonable values in, but of course cannot recover what was really lost here.

(2) Google does not recognize SPY, but requires NYSEARCA:SPY instead. There is a function for translating this, I am sure there will be a dict requires at some point in the future.

Cheers, Felix

### DIFF
--- a/ToolBox/GoogleDownloader/GoogleDataDownloader.cs
+++ b/ToolBox/GoogleDownloader/GoogleDataDownloader.cs
@@ -50,11 +50,9 @@ namespace QuantConnect.ToolBox.GoogleDownloader
                 case Resolution.Minute:
                 case Resolution.Hour:
                     return GetMinuteOrHour(symbol, resolution, startUtc, endUtc);
-                    break;
 
                 case Resolution.Daily:
                     return GetDaily(symbol, resolution, startUtc, endUtc);
-                    break;
 
                 default:
                     throw new NotSupportedException("Resolution not available: " + resolution);
@@ -240,7 +238,7 @@ namespace QuantConnect.ToolBox.GoogleDownloader
                 return "NYSEARCA:SPY";
             }
 
-            return symbol;
+            return symbol.Value;
         }
     }
 }
--- a/ToolBox/GoogleDownloader/GoogleDataDownloader.cs
+++ b/ToolBox/GoogleDownloader/GoogleDataDownloader.cs
@@ -207,17 +207,6 @@ namespace QuantConnect.ToolBox.GoogleDownloader
             return epoch.AddSeconds(unixTime);
         }
 
-        private static string ConvertTicker(Symbol symbol)
-        {
-            // FIXME: this is a hack to make sure google finds our index:
-            if (symbol.Value == "SPY")
-            {
-                return "NYSEARCA:SPY";
-            }
-
-            return symbol.Value;
-        }
-
         /// <summary>
         /// Get historical data enumerable for a single symbol, type and resolution given this start and end time (in UTC).
         /// </summary>
--- a/ToolBox/GoogleDownloader/GoogleDataDownloader.cs
+++ b/ToolBox/GoogleDownloader/GoogleDataDownloader.cs
@@ -29,36 +29,6 @@ namespace QuantConnect.ToolBox.GoogleDownloader
     /// </summary>
     public class GoogleDataDownloader : IDataDownloader
     {
-        /// <summary>
-        /// Get historical data enumerable for a single symbol, type and resolution given this start and end time (in UTC).
-        /// </summary>
-        /// <param name="symbol">Symbol for the data we're looking for.</param>
-        /// <param name="resolution">Resolution of the data request</param>
-        /// <param name="startUtc">Start time of the data in UTC</param>
-        /// <param name="endUtc">End time of the data in UTC</param>
-        /// <returns>Enumerable of base data for this symbol</returns>
-        public IEnumerable<BaseData> Get(Symbol symbol, Resolution resolution, DateTime startUtc, DateTime endUtc)
-        {
-            if (symbol.ID.SecurityType != SecurityType.Equity)
-                throw new NotSupportedException("SecurityType not available: " + symbol.ID.SecurityType);
-
-            if (endUtc < startUtc)
-                throw new ArgumentException("The end date must be greater or equal than the start date.");
-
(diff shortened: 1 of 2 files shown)


ToolBox/GoogleDownloader/GoogleDataDownloader.cs
```diff
@@ -44,21 +39,122 @@ public class GoogleDataDownloader : IDataDownloader
...
+                    columns[3] != "-" ? (Decimal?)columns[3].ToDecimal() : null,
+                    columns[4] != "-" ? (Decimal?)columns[4].ToDecimal() : null
+                };
+
+                // let's try hard to fix any issues as good as we can
+                // this code assumes that there is at least 1 good value
+                if (ohlc[1] == null) ohlc[1] = ohlc.Where(val => val != null).Max();
+                if (ohlc[2] == null) ohlc[2] = ohlc.Where(val => val != null).Min();
+                if (ohlc[0] == null) ohlc[0] = ohlc.Where(val => val != null).Average();
+                if (ohlc[3] == null) ohlc[3] = ohlc.Where(val => val != null).Average();
```
mchandschuh: I would be reluctant to improvise data. How often does this happen?

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
```diff
@@ -134,5 +230,15 @@ private static DateTime FromUnixTime(long unixTime)
             return epoch.AddSeconds(unixTime);
         }
 
+        private static string ConvertTicker(Symbol symbol)
+        {
+            // FIXME: this is a hack to make sure google finds our index:
+            if (symbol.Value == "SPY")
+            {
+                return "NYSEARCA:SPY";
```
mchandschuh: Is it only SPY that behaves this way?  How did it work before without this?

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
```diff
@@ -27,13 +29,6 @@ namespace QuantConnect.ToolBox.GoogleDownloader
     /// </summary>
     public class GoogleDataDownloader : IDataDownloader
     {
-        // q = SYMBOL
-        // i = resolution in seconds
-        // p = period in days
-        // ts = start time
-        // Strangely Google forces CHLO format instead of normal OHLC.
-        private const string UrlPrototype = @"http://www.google.com/finance/getprices?q={0}&i={1}&p={2}d&f=d,c,h,l,o,v&ts={3}";
-
```
mchandschuh: Looks like we just moved this somewhere else in the file, obfuscating the diff a little bit.

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
```diff
@@ -44,21 +39,122 @@ public class GoogleDataDownloader : IDataDownloader
...
+
+            // Create the Google formatted URL.
+            var url = string.Format(UrlPrototypeDaily, ConvertTicker(symbol), startdate, enddate);
+
+            // Download the data from Google.
+            string[] lines;
+            using (var client = new WebClient())
+            {
+                var data = client.DownloadString(url);
+                lines = data.Split('\n');
```
mchandschuh: I don't think this is an issue, but as a rule of thumb it is typically better to not directly split by lines, but instead use a stream reader and iterate by lines. In this case I suspect the 'file' is small enough that it doesn't matter, so this is just food for thought :)

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
fbertram: moved privates to top of file.

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
fbertram: thanks for the feedback, I actually copied this from the original routine. A typical download might be about 10 years, resulting in 2,500 lines here. Left code as is.

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
fbertram: I am not sure what Google does here, I see this about 20 times in 20 years. I agree with your concerns, and it makes me cringe too.
I see the following options:
(1) drop the bar, which will have some side effects.
(2) fill missing values with zero. this might trigger quite erratic reactions in an algorithm.
(3) try to repair. while this is not ideal, I feel that this creates the least amount of side effects

Left code as is, in the spirit of providing the best simulation we can. Basically, we have to admit that downloading data from Google is not ideal. The data from Yahoo are better, but they are not adjusted...

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
fbertram: I was seeing client.DownloadString failing and not returning any content, but can no longer reproduce this. I have removed this routine for now. Not sure why the patient has healed itself, this may or may not be related to the '-' issue discussed above.

ToolBox/GoogleDownloader/GoogleDataDownloader.cs
```diff
@@ -28,37 +30,101 @@ namespace QuantConnect.ToolBox.GoogleDownloader
...
+                // date = 20160927
+                var day = DMY[0].ToInt32();
+                var month = (int)Enum.Parse(typeof(ConvertMonths), DMY[1]);
+                var year = (DMY[2].ToInt32() > 70) ? 1900 + DMY[2].ToInt32() : 2000 + DMY[2].ToInt32();
+                var time = new DateTime(year, month, day, 0, 0, 0);
+
+                // occasionally, the columns will have a '-' instead of a proper value
+                List<Decimal?> ohlc = new List<Decimal?>()
+                {
+                    columns[1] != "-" ? (Decimal?)columns[1].ToDecimal() : null,
```
jaredbroad: Could you please make this log an error message when a null appears? Log.Error() with the date and price that google returned so consumer knows its an invalid point.

fbertram: done

Issue #573, opened 2016-09-28
Create "No Margin Calls" Model to Rely on Brokerage Margin Calls


jaredbroad:
For live trading its sometimes best to just not do any internal LEAN margin call / modelling and rely on the  brokerage margin calls. We should create a class which users can use to disable LEAN margin modelling.

jaredbroad:
![](https://github.trello.services/images/mini-trello-icon.png) [NoMarginCallMarginModel](https://trello.com/c/OgYEMlUZ/982-nomargincallmarginmodel)

PR #582, opened 2016-10-01
Implements NoMarginCallMarginModel

AlexCatarino:
NoMarginCallMarginModel is a version of SecurityMarginModel which does not issue margin call orders.
Ref.: #573


jaredbroad:
Thank you Alex,

PR #583, opened 2016-10-03
Add Tradier Brokerage IHistoryProvider implementation

StefanoRaggi:


jaredbroad:
Thank you, that is a nice efficient implementation and tests.

Brokerages/Tradier/TradierBrokerage.HistoryProvider.cs
```diff
@@ -0,0 +1,218 @@
...
+                .Select(g => new TradeBar(
+                    g.Key,
+                    symbol,
+                    g.First().Open,
+                    g.Max(t => t.High),
+                    g.Min(t => t.Low),
+                    g.Last().Close,
+                    g.Sum(t => t.Volume),
+                    Time.OneHour))
+                .Select(tradeBar => new Slice(tradeBar.Time, new[] { tradeBar }))
```
mchandschuh: should we be passing `tradeBar.EndTime` into slice here?

StefanoRaggi: Yes, good catch, fixed here: https://github.com/QuantConnect/Lean/pull/588

Issue #474, opened 2016-07-10
Add Regression Channel as an Indicator


abhishekc05:
Is it possible to include linear regression channels as a built in Indicator ?

jameschch:
I've considered putting together linear regression channels as I believe it's just offset periods of the existing LSMA indicator. Was there a reference implementation you have in mind?

abhishekc05:
Hi,

Thanks for that. The regression channel needs just a few extra outputs
other than the last predicted value. I was using the following crude
version - am not really confident of being able to add an indicator
properly to the Engine as am relatively new to C#.

The input is the period.

The important outputs are: slope, standard deviation of errors and last
predicted value. The bands can be calculated using the current price and
standard deviation of errors or can be part of the output (in which case
the number of standard deviations to use would also be an input); slope is
useful to identify the trend.


--------------------------------------------------------------
public void AddSample(decimal price)
{
if (_samples<_period)
{
_slope = 1m;
_upBand = price;
_lowBand =price;
_isReady = false;
_samples++;
_yBuffer.Add((double)price);
_xBuffer.Add(_samples);
return;
}
else
{
_slopeCoefficients = Fit.Polynomial(_xBuffer.ToArray(),
_yBuffer.ToArray());
_isReady =  true;
double[] _errors = new double[_xBuffer.Count()];
            double[] _modelledValues = new double[_yBuffer.Count()];
            for (int _curCount = 0;_curCount<_xBuffer.Count();_curCount++)
        {
              //_modelledValues[_curCount] =
Evaluate.Polynomial(_xBuffer[_curCount], _slopeCoefficients);
             _modelledValues [_curCount] = _slopeCoefficients[0] +
_slopeCoefficients[1]*_xBuffer[_curCount];
             _errors[_curCount] = _modelledValues[_curCount] -
_yBuffer[_curCount];
                }

            _oneStdDev =
(decimal)ArrayStatistics.PopulationStandardDeviation(_errors);
 //Standard
Devation of errors  - Population uses N
            _lastPredictedValue = (decimal)(_slopeCoefficients[0] +
_slopeCoefficients[1]*_xBuffer.Last());        //Getting the last
calculated value
            _slope = (decimal)_slopeCoefficients[1];
            _upBand = _lastPredictedValue + _oneStdDev*(decimal)_deviation;
            _lowBand = _lastPredictedValue - _oneStdDev*(decimal)_deviation;

            _yBuffer.Add((double)price); //Preparing for next band value
[...]

PR #584, opened 2016-10-03
Implements the RegressionChannel indicator

AlexCatarino:
The Regression Channel indicator extends the LeastSquaresMovingAverage with the inclusion of two (upper and lower) channel lines that are distanced from the linear regression line by a user defined number of standard deviations.
Ref.: #474

### DIFF
--- a/Indicators/LeastSquaresMovingAverage.cs
+++ b/Indicators/LeastSquaresMovingAverage.cs
@@ -36,12 +36,12 @@ namespace QuantConnect.Indicators
         /// <summary>
         /// The point where the regression line crosses the y-axis (price-axis)
         /// </summary>
-        public WindowIndicator<IndicatorDataPoint> Intercept { get; private set; }
+        public IndicatorBase<IndicatorDataPoint> Intercept { get; private set; }
         
         /// <summary>
         /// The regression line slope
         /// </summary>
-        public WindowIndicator<IndicatorDataPoint> Slope { get; private set; }
+        public IndicatorBase<IndicatorDataPoint> Slope { get; private set; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LeastSquaresMovingAverage"/> class.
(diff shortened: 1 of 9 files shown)


Indicators/LeastSquaresMovingAverage.cs
```diff
@@ -63,23 +75,31 @@ public LeastSquaresMovingAverage(int period)
...
+                .OrderBy(i => i.Time)
+                .Select(i => Convert.ToDouble(i.Value))
+                .ToArray();
+            // Fit OLS
+            var ols = Fit.Line(x: t, y: series);
+            Intercept.Update(input.Time, (decimal)ols.Item1);
+            Slope.Update(input.Time, (decimal)ols.Item2);
+
+            // Calculate the fitted value corresponding to the input
+            return Intercept + Slope * Period;
```
mchandschuh: Looks like you already fixed it :) 

You could squash the commit that fixed this into the commit that produced it. Not required, but nice easy way of dipping your toes into the `git rebase` world of goodness :)

AlexCatarino: In my first try, Intercept and Slope were not window indicator, but then I used Intercept.Plus(Slope.Times(Period)), it was returning zeroes. Then I moved to window and use that expression. Finally, when you suggested reviewing the usage of window indicator, I re-access everything and come up with this solution.

I will wait for final reviews and squash some commits (if everyone agrees) that were mostly typos and variable renaming.

mchandschuh: Variable renames are nice in their own commits. Ideally,  any change that has no effect on behavior should be in its own commit. This includes doc changes,  white space changes, renames and mechanical refactorings. This helps the diff engine during merges and helps the poor sap that needs to look through the history to find out how something for some way... It's much better to be able to completely ignore the renames in the diff and instead look at a change that is primarily changing behavior, vs a commit that changes behaviors and has renames and refactorings, ect...

AlexCatarino: Thanks for the pro-tip.

jaredbroad:
Thanks Alex

PR #592, opened 2016-10-05
FactorFileGenerator

andrewhart098:
+ Expanded YahooDataDownloader to download Split and Dividend data from the Yahoo Api
+ Add utility to create factor files from Yahoo Split/Dividend data and daily equity data


mchandschuh:
Unable to do a full review at this time, but the prospect of a utility that can generate these files sounds awesome. Perhaps there is a way to abstract the yahoo-ness from the factor for generation code so it could be used with other data sources? Once again, didn't dig into the implementation, so I understand if this is not feasible.

andrewhart098:
@mchandschuh - I think you are absolutely right, I abstracted the Yahoo-ness from the factor file generator by adding a class called EquityEvent.  EquityEvent is supposed to represent a Split or dividend.  If this isn't the best name, I would be happy to change it.  As long as you can put the split dividend data from a provider into an EquityEvent, then you should be able to generator factor files.

mchandschuh:
Would the built in `Split` and  `Dividend` classes work?  Your queue would need to be changed to accept `BaseData`.

andrewhart098:
@mchandschuh - Yes, that was a better idea.

andrewhart098:
Moving the discussion from quantconnect.com to here.
https://www.quantconnect.com/forum/discussion/1411/splitdividend-factor-files/p1

Jeff Miller wrote: 

> The only thing I didn’t notice in your pull request is how you are handling a split and dividend events that occur on the same day.  Dividends should be applied pre-split in most cases.

Under what conditions would a split be applied before a dividend?  Should I trust the order of the split dividend data from the provider or should I manually check the split dividend data for same day events and order them appropriately?

je-miller:
> Should I trust the order of the split dividend data from the provider or should I manually check the split dividend data for same day events and order them appropriately?
I would explicitly check for same day events and apply the dividend event first.   I thought I read somewhere that on rare occasions dividends could be paid post split.  I'm not sure about this though.

andrewhart098:
I rebased my local git branch to clean up the git commit history and in the process made this branch and my local branch incompatible.  Please see my new PR #605.

@jaredbroad - please discard this PR.

PR #593, opened 2016-10-07
Add DisposeSafely extension methods

mchandschuh:
I had originally written these elsewhere, but feel that they're general enough
to be moved into Util. Let me know if you think they don't belong here.

Add `DisposableExtensions` to hold extension methods for `IDisposable`.
Add two overloads of `DisposeSafely`. This will allow you to call
`myDisposable.DisposeSafely()` and it will wrap the `Dispose` call in a
try/catch and log the error. The other overload allows you to
handler the error using an `Action<Exception> errorHandler` parameter

### DIFF
--- a/Common/QuantConnect.csproj
+++ b/Common/QuantConnect.csproj
@@ -436,6 +436,7 @@
     <Compile Include="Util\BusyBlockingCollection.cs" />
     <Compile Include="Util\CircularQueue.cs" />
     <Compile Include="Util\Composer.cs" />
+    <Compile Include="Util\DisposableExtensions.cs" />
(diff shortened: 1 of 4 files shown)


jaredbroad:
Looks good. Curious - who was throwing during a dispose call?

mchandschuh:
I've just written it so many times and found myself writing it again last night and decided enough was enough! haha, also it sometimes causes a blow-out of the indentation, figure you're using an `IDisposable` within a try catch/finally or something, and in the finally you want to try/dispose since throwing in a finally is foolhardy :)

PR #596, opened 2016-10-10
Prevents InsufficientBuyingPower message flooding

AlexCatarino:
When there is no sufficiente capital for issueing new orders, warning messages are sent for every try. These do not contains relevant new information, but their creation and disposal can make memory run out.


jaredbroad:
Hiding this is dangerous as it can hide potential errors in the algorithm.

PR #600, opened 2016-10-17
Increases default symbol limits & adds config option to change these

rbohac:


Queues/JobQueue.cs
```diff
@@ -93,6 +93,12 @@ public AlgorithmNodePacket NextJob(out string location)
                     RamAllocation = int.MaxValue,
                     Parameters = parameters,
                     Language = Language,
+                    Controls = new Controls()
```
rbohac: ok i'll have to do this later

rbohac:
@jaredbroad config.json has merge conflicts. Looks like some other changes related to IB & Tradier & Local Desktop were made. Do you want to handle merging those?

PR #601, opened 2016-10-17
IFileProvider

andrewhart098:
The pull request introduces the IFileProvider interface whose purpose is to retrieve data if it is not already on the local file system.  The IFileProvider is configured in config.json.  Out of the box, the DefaultFileProvider is used which does nothing.  I also added an ApiFileProvider which attempts to download data from a users data library on QuantConnect.com if it is not already on the local file system.  IFileProvider has one method, Fetch(), that takes a symbol, resolution and date. It returns a bool if it was successfully able to retrieve the data.

The implementation got a bit messy....

The actual checking for the data on disk is done in ZipEntryNameSubscriptionDataSourceReader and TextSubscriptionDataSourceReader.  So this is where I also call the IFileProvider's Fetch() method.  Getting the IFileProvider from Engine.cs to SubscriptionDataSourceReader.cs (where ZipEntryNameSubscriptionDataSourceReader  and TextSubscriptionDataSourceReader are called) is where things got weird.

There are essentially four places where SubscriptionDataSourceReader is used:

- 1 implementation of IHistoryProvider (SubscriptionDataReaderHistoryProvider.cs)
-  ISubscriptionEnumeratorFactory's CreateEnumerator() method
- FileSystemDataFeed
- LiveTradingDataFeed

I tried to make as few changes as possible.  I am open to suggestions on how to make this implementation cleaner.  

Any thoughts? @mchandschuh @StefanoRaggi

### DIFF
--- a/Common/Interfaces/IDataFileProvider.cs
+++ b/Common/Interfaces/IDataFileProvider.cs
@@ -14,25 +14,24 @@
 */
 
 using System;
-using QuantConnect.Interfaces;
+using System.ComponentModel.Composition;
 
-namespace QuantConnect.Lean.Engine.DataFeeds
+namespace QuantConnect.Interfaces
 {
     /// <summary>
-    /// Default file provider functionality that does not attempt to retrieve any data
+    /// Fetches a remote file for a security using a Symbol, Resolution and DateTime
+    /// Used to get remote files when they are not found locally
     /// </summary>
-    public class DefaultFileProvider : IFileProvider
+    [InheritedExport(typeof(IDataFileProvider))]
+    public interface IDataFileProvider
     {
         /// <summary>
-        /// Does not attempt to retrieve any data
+        /// Gets and downloads the remote file
         /// </summary>
         /// <param name="symbol">Symbol of the security</param>
         /// <param name="resolution">Resolution of the data requested</param>
         /// <param name="date">DateTime of the data requested</param>
-        /// <returns>False</returns>
-        public bool Fetch(Symbol symbol, Resolution resolution, DateTime date)
-        {
-            return false;
-        }
+        /// <returns>Bool indicating whether the remote file was fetched correctly</returns>
+        bool Fetch(Symbol symbol, Resolution resolution, DateTime date);
     }
 }
--- a/Common/Interfaces/IDataFileProvider.cs
+++ b/Common/Interfaces/IDataFileProvider.cs
@@ -19,8 +19,8 @@ using System.ComponentModel.Composition;
 namespace QuantConnect.Interfaces
 {
     /// <summary>
-    /// Fetches a remote file for a security using a Symbol, Resolution and DateTime
-    /// Used to get remote files when they are not found locally
+    /// Fetches a remote file for a security.
+    /// Must save the file to Globals.DataFolder.
     /// </summary>
     [InheritedExport(typeof(IDataFileProvider))]
     public interface IDataFileProvider
@@ -28,10 +28,11 @@ namespace QuantConnect.Interfaces
         /// <summary>
         /// Gets and downloads the remote file
         /// </summary>
-        /// <param name="symbol">Symbol of the security</param>
-        /// <param name="resolution">Resolution of the data requested</param>
+        /// <param name="symbol"><see cref="Symbol"/> of the security</param>
+        /// <param name="resolution"><see cref="Resolution"/> of the data requested</param>
         /// <param name="date">DateTime of the data requested</param>
+        /// <param name="tickType"><see cref="TickType"/> of the security</param>
         /// <returns>Bool indicating whether the remote file was fetched correctly</returns>
-        bool Fetch(Symbol symbol, Resolution resolution, DateTime date);
+        bool Fetch(Symbol symbol, DateTime date, Resolution resolution, TickType tickType);
     }
 }
--- a/Engine/DataFeeds/ApiDataFileProvider.cs
+++ b/Engine/DataFeeds/ApiDataFileProvider.cs
@@ -21,14 +21,15 @@ using QuantConnect.Logging;
 namespace QuantConnect.Lean.Engine.DataFeeds
 {
     /// <summary>
-    /// Attempts to download data from the api
+    /// Attempts to download data from the api and save it in the data folder specified in config.json.
+    /// This implementation will overwrite data if it already exists.
     /// </summary>
     public class ApiDataFileProvider : IDataFileProvider
     {
         private readonly int _uid = Config.GetInt("job-user-id", 0);
         private readonly string _token = Config.Get("api-access-token", "1");
         private readonly string _dataPath = Config.Get("data-folder", "../../../Data/");
-        public bool Fetch(Symbol symbol, Resolution resolution, DateTime date)
+        public bool Fetch(Symbol symbol, DateTime date, Resolution resolution, TickType tickType)
         {
             Log.Trace(
                 string.Format(
--- a/Engine/DataFeeds/ApiDataFileProvider.cs
+++ b/Engine/DataFeeds/ApiDataFileProvider.cs
@@ -23,7 +23,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
     /// <summary>
     /// Attempts to download data from the api
     /// </summary>
-    public class ApiFileProvider : IFileProvider
+    public class ApiDataFileProvider : IDataFileProvider
     {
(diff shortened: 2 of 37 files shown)


jaredbroad:
@mchandschuh - thoughts for lower touch method welcome.

Engine/DataFeeds/ApiFileProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+using QuantConnect.Interfaces;
+using QuantConnect.Logging;
+
+namespace QuantConnect.Lean.Engine.DataFeeds
+{
+    /// <summary>
+    /// Attempts to download data from the api
+    /// </summary>
+    public class ApiFileProvider : IFileProvider
+    {
```
StefanoRaggi: We already have `MapFileProvider` and `FactorFileProvider` class names, perhaps we could rename `XxxFileProvider` to `XxxDataFileProvider` for consistency ?

Launcher/config.json
```diff
@@ -1,226 +1,221 @@
...
+            "real-time-handler": "QuantConnect.Lean.Engine.RealTime.LiveTradingRealTimeHandler",
+            "transaction-handler": "QuantConnect.Lean.Engine.TransactionHandlers.BrokerageTransactionHandler",
+            "messaging-handler": "QuantConnect.Messaging.StreamingMessageHandler",
+            "log-handler": "QuantConnect.Logging.QueueLogHandler",
+            "desktop-http-port": "1234",
+            "desktop-exe": "../../../UserInterface/bin/Release/QuantConnect.Views.exe"
+        }
     }
-  }
 }
```
StefanoRaggi: With VS 2015 the default indent/tab size for JSON files is now 2 spaces instead of 4, I think we should keep it like this for easier updates and diffs.

StefanoRaggi:
@andrewhart098 Great job, just added a couple minor comments inline.
The PR looks good, I can't see an easier way to implement this.

Engine/DataFeeds/ApiFileProvider.cs
andrewhart098: I like XxxDataFileProvider.  Any thoughts on that @jaredbroad?

Common/Interfaces/IDataFileProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.ComponentModel.Composition;
+
+namespace QuantConnect.Interfaces
+{
+    /// <summary>
+    /// Fetches a remote file for a security using a Symbol, Resolution and DateTime
+    /// Used to get remote files when they are not found locally
+    /// </summary>
+    [InheritedExport(typeof(IDataFileProvider))]
+    public interface IDataFileProvider
```
mchandschuh: I think it's important to document here that in order for this to work within Lean, the `IDataFileProvider` implementations *must* save the file to `Globals.DataFolder` in it's current form. Something better would be if the `Fetch` method accepts the root data directory. Accepting it as an argument is nice because it moves the dependency outside of this implementation and forces the consumer to decide where to save it.

Engine/DataFeeds/ApiDataFileProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            Log.Trace(
+                string.Format(
+                    "Attempting to get data from QuantConnect.com's data library for symbol({0}), resolution({1}) and date({2}).",
+                    symbol.ID, resolution, date.Date.ToShortDateString()));
+
+            var api = new Api.Api();
+            api.Initialize(_uid, _token, _dataPath);
+
+            var download = api.DownloadData(symbol, resolution, date);
```
mchandschuh: Since the purpose of `Fetch` is to save a certain file to a certain place on disk, does it make sense to check for the file first, or is it more correct for the caller to understand that calling `Fetch` may overwrite existing data at the generated file path.

Engine/DataFeeds/ZipEntryNameSubscriptionDataSourceReader.cs
```diff
@@ -58,9 +62,9 @@ public ZipEntryNameSubscriptionDataSourceReader(SubscriptionDataConfig config, D
         /// <returns>An <see cref="IEnumerable{BaseData}"/> that contains the data in the source</returns>
         public IEnumerable<BaseData> Read(SubscriptionDataSource source)
         {
-            if (!File.Exists(source.Source))
+            if (!File.Exists(source.Source) && !_dataFileProvider.Fetch(_config.Symbol, _config.Resolution, _date))
             {
-                OnInvalidSource(source, new FileNotFoundException("The specified file was not found", source.Source));
+                OnInvalidSource(source, new FileNotFoundException("The specified file was not found", source.Source));
```
mchandschuh: Please remove the trailing white space at the end of this line

#### Git Tip!
>You can run `git diff --check` to look for white space errors before you commit changes.

Common/Interfaces/IDataFileProvider.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Gets and downloads the remote file
+        /// </summary>
+        /// <param name="symbol">Symbol of the security</param>
+        /// <param name="resolution">Resolution of the data requested</param>
+        /// <param name="date">DateTime of the data requested</param>
+        /// <returns>Bool indicating whether the remote file was fetched correctly</returns>
+        bool Fetch(Symbol symbol, Resolution resolution, DateTime date);
```
mchandschuh: I think this signature would be best. I chose this order of parameters to match that of `LeanData.GenerateZipFilePath`.

```csharp
public interface IDataFileProvider
{
    /// <summary>
    /// Fetches the zipped file bytes for the data associated with the specified parameters
    /// </summary>
    byte[] Fetch(Symbol symbol, DateTime date, Resolution resolution, TickType tickType);
}
```

@jaredbroad I could actually reuse these abstractions elsewhere in current development, which would be schweet!

mchandschuh: Editted above comment to return a `Stream`. This will be more reusable. If we already have a `byte[]` then it's an almost free operation to put it into a `MemoryStream`, whereas if you're truly streaming from a remote source, this could save processing time and memory resources.

Engine/DataFeeds/ApiDataFileProvider.cs
andrewhart098: I think it makes more sense to have the caller understand that calling Fetch may overwrite the data.  This is the current case in both TextSubscriptionDataSourceReader and ZipEntryNameSubscriptionDataSourceReader.  I will add documentation to that effect.

Common/Interfaces/IDataFileProvider.cs
andrewhart098: @mchandschuh - Can you be more specific about a situation where you would want to stream the result of the fetch method?

mchandschuh: This is more of a long term vision thing about unifying how data feeds source data. No need to worry about it for this PR. Saving directly to disk is the best option for now.

Engine/DataFeeds/ApiDataFileProvider.cs
BradGessner: Hi guys, I just found this cool work!  I was in the middle of writing one on my own but am happy to see an interface.  I would suggest that if the file already exists, the backtester would just read that during backtesting.  Otherwise, it would fetch the IFileDataProvider and write that to disk in Lean format after feeding to the algorithm.  Perhaps it will be extensible by the consumer to do this too.
Either way you decide, I'm excited to see this new feature so I can pipe other data sources quickly.

BradGessner: Hi guys, I just found this cool work!  I was in the middle of writing one on my own but am happy to see an interface.  I would suggest that if the file already exists, the backtester would just read that during backtesting.  Otherwise, it would fetch the IFileDataProvider and write that to disk in Lean format after feeding to the algorithm.  Perhaps it will be extensible by the consumer to do this too.
Either way you decide, I'm excited to see this new feature so I can pipe other data sources quickly.

andrewhart098:
@BradGessner, that is exactly how the IDataFileProvider works - it checks
if the file is on disk and if not attempts to use the IDataFileProvider. If
the file is already on disk, Lean makes no attempt to use the
IDataFileProvider. See TextSubscriptionDataSourceReader.cs and
ZipEntryNameSubscriptionDataSourceReader.cs

PR #602, opened 2016-10-17
Ability to switch between internal feeds and normal feeds

StefanoRaggi:
This PR solves two problems:

1. In live trading, if in the existing holdings at the brokerage there is also the benchmark symbol (SPY), the data feed is still marked as internal (so OnData does not fire for this symbol)

2. The same problem happens if the benchmark symbol is selected by universe selection or added manually with AddSecurity in OnData


Engine/Setup/BrokerageSetupHandler.cs
```diff
@@ -331,6 +330,8 @@ public bool Setup(IAlgorithm algorithm, IBrokerage brokerage, AlgorithmNodePacke
                     return false;
                 }
 
+                algorithm.PostInitialize();
```
jaredbroad: Great solution!

PR #606, opened 2016-10-23
Add config setting to save list of trades in csv local file

StefanoRaggi:


Engine/Engine.cs
```diff
@@ -384,5 +392,26 @@ public void Run(AlgorithmNodePacket job, string assemblyPath)
                 _algorithmHandlers.RealTime.Exit();
             }
         }
+
+        private static void SaveListOfTrades(IOrderProvider transactions, string csvFileName)
+        {
+            var orders = transactions.GetOrders(x => x.Status.IsFill());
+
+            using (var writer = new StreamWriter(csvFileName))
```
mchandschuh: May want to ensure the directory exists

PR #608, opened 2016-10-31
Use .NET zip implementation for ZipBytes

mchandschuh:
The replaced zip implementation had issues in high performance,
multi-threaded scenarios. The exact cause of the issue was not
determined, but invalid zip files were being generated when using this
method directly.

The replacement uses the built in .NET zip archive implementation
which should (in theory) have better cross-platform characteristics
than 3rd party implementations.

The unit test was also update as it made incorrect usage of the
Compression.Unzip routine. It has been replaced with a more reliable
method that unzips the first entry and returns a stream reader.


mchandschuh:
The first version of the unit test didn't seem to play well on travis. Updated the unit test to use a more mono friendly version. Let's see how travis likes this one.

PR #620, opened 2016-11-08
ApiDataQueueHandler

andrewhart098:
- Added new IDataQueueHandler, ApiDataQueueHandler that can retrieve live data from a websocket connection
- Added ApiWebSocketConnection.cs, which manages the connection to the websocket providing data using [websocket-sharp](https://github.com/sta/websocket-sharp)
- Added new methods to IApi.cs and Api.cs that adds a layer of abstraction to ApiWebSocketConnection.cs
- Added ApiDataQueueHandlerTests.cs
- Added new config options to specify websocket connection


Api/ApiConnection.cs
```diff
@@ -84,7 +85,7 @@ public bool TryRequest<T>(RestRequest request, out T result)
                 // Add the UTC timestamp to the request header.
                 // Timestamps older than 1800 seconds will not work.
                 var timestamp = (int)Time.TimeStamp();
-                var hash = CreateSecureHash(timestamp);
+                var hash = ApiExtensions.CreateSecureHash(timestamp, _token);
```
jaredbroad: Not an extension; but it works better as a static method on the API -- i.e. Api.CreateSecureHash(timestamp, _token)

Launcher/config.json
```diff
@@ -49,6 +49,10 @@
   "job-user-id": "0",
   "api-access-token": "",
 
+  // live data configuration
+  "live-data-url": "ws://127.0.0.1",
+  "live-data-port": 8080,
```
jaredbroad: Please default to API URL and 443 port

Api/Api.cs
```diff
@@ -613,5 +615,35 @@ public virtual void Dispose()
         {
             // NOP
         }
+
+
+        /// <summary>
+        /// Generate a secure hash for the authorization headers.
+        /// </summary>
+        /// <returns>Time based hash of user token and timestamp.</returns>
+        public static string CreateSecureHash(int timestamp, string token)
```
mchandschuh: Arguably, this need not be public. But if it is public, perhaps a better name is `CreateTimestampedToken` or even `CreatePassword`

Common/Data/BaseData.cs
```diff
@@ -301,5 +305,41 @@ public virtual string GetSource(SubscriptionDataConfig config, DateTime date, Da
         {
             throw new InvalidOperationException("Please implement GetSource(SubscriptionDataConfig, DateTime, bool) on your custom data type: " + GetType().Name);
         }
+
+        /// <summary>
+        /// Deserialize the message from the data server
+        /// </summary>
+        /// <param name="serialized">The data server's message</param>
+        /// <returns>An enumerable of base data, if unsuccessful, returns an empty enumerable</returns>
+        public static IEnumerable<BaseData> DeserializeMessage(string serialized)
```
mchandschuh: I see this copied everywhere. I'm equally guilty here.

PR #622, opened 2016-11-09
Use IBaseData in type constraints

mchandschuh:
Refactors existing consolidators, indicators, and helper methods to depend on
IBaseData instead of BaseData. These updates also defines an IBaseDataBar to
act as an abstraction point between TradeBar and QuoteBar.

Tried to think of a way of making this in smaller commits but it really is a
systemic change. Most of the changes involve changing `TradeBar` references
to the new `IBaseDataBar` (open to suggestions on naming). Some other
changes were to the indicator selectors, now `Func<IBaseData, decimal>` and
`Func<IBaseData, IBaseDataBar>`

### DIFF
--- a/Algorithm.CSharp/IndicatorSuiteAlgorithm.cs
+++ b/Algorithm.CSharp/IndicatorSuiteAlgorithm.cs
@@ -198,7 +198,7 @@ namespace QuantConnect
         /// <summary>
         /// Function used to select a trade bar that has double the values of the input trade bar
         /// </summary>
-        private static TradeBar SelectorDoubleTradeBar(BaseData baseData)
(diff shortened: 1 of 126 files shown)


mchandschuh:
With the code changes done, we'll want to re-assess the potential impact on user code. I was able to make these changes with only one update required in an algorithm, so if that counts as a litmus than we should be pretty safe with these changes.

andrewhart098:
Should QuoteBar implement IBaseDataBar rather than IBar? That way, QuoteBar can be used in all the indicators.

So this: `public class QuoteBar : BaseData, IBaseDataBar`

Rather than: `public class QuoteBar : BaseData, IBar`

mchandschuh:
@andrewhart098: yes, absolutely.

PR #625, opened 2016-11-09
Removed Disconnect call from IB thread catch block

StefanoRaggi:


mchandschuh:
I wasn't 100% about this, just commented on it to see if it was something we considered. As always, I'm open to rebuttal.

PR #629, opened 2016-11-15
Add trading mode in IB BrokerageData + fix typo in run-ib-controller.bat

StefanoRaggi:


jaredbroad:
Thank you :)

mchandschuh:
Just curious, what does this mean/possible values?

StefanoRaggi:
From IBController.ini:
# Trading Mode
# ------------
#
# TWS 955 introduced a new Trading Mode combo box on its login 
# dialog. This indicates whether the live account or the paper
# trading account corresponding to the supplied credentials is
# to be used. The allowed values are 'live' (the default) and 
# 'paper'. Note that if the supplied credentials are for a 
# paper trading account, the setting is ignored. For earlier 
# versions of TWS this setting has no effect.

jaredbroad:
IB stopped automatically picking live/paper based on the credentials

PR #637, opened 2016-11-22
Fix InvalidOperationException in OS.CpuUsage on Windows

StefanoRaggi:


Common/OS.cs
```diff
@@ -83,7 +83,8 @@ public static PerformanceCounter CpuUsage
             {
                 if (_cpuUsageCounter == null)
                 {
-                    _cpuUsageCounter = new PerformanceCounter("Process", "% Processor Time", Process.GetCurrentProcess().Id.ToString());
+                    _cpuUsageCounter = new PerformanceCounter("Process", "% Processor Time", 
+                        IsWindows ? Process.GetCurrentProcess().ProcessName : Process.GetCurrentProcess().Id.ToString());
```
mchandschuh: Nice. I noticed this recently too when running all of the tests. Thanks for fixing!

PR #638, opened 2016-11-23
FX subscriptions compatible with Quote data

andrewhart098:
- Securities can now be configured to use one or more BaseData types (QuoteBar, TradeBar etc) in a subscription
- Forex subscriptions now uses QuoteBar by default and are backwards compatible with TradeBars
- SpreadSlippageModel removed and replaced with ConstantSlippageModel
- ImmediateFillModel now uses spread to calculate the fill price for Tick data


Algorithm/QCAlgorithm.cs
```diff
@@ -539,6 +548,35 @@ public void SetParameters(Dictionary<string, string> parameters)
...
+            foreach (var dataFeed in availableDataFeeds)
+            {
+                AvailableDataFeeds[dataFeed.Key] = dataFeed.Value;
+            }
+        }
+
+        /// <summary>
+        /// Hard code the set of default available data feeds
+        /// </summary>
+        public Dictionary<SecurityType, List<TickType>> DefaultDataFeeds()
```
quant1729: Hey Andrew, 

Thanks for the PR. 

Wondering what is the idea behind these default data feeds? Currently we have tradebars, quotebars, oi supported in futures, options. Should we reflect it in the default list or use the file with overrides (via "security-data-feeds")?

I think it makes sense to fully test this in local live/backtest configurations for futures/options.

quant1729: Lets chat when you are online

Common/Data/UniverseSelection/Universe.cs
```diff
@@ -174,7 +174,7 @@ public IEnumerable<Symbol> PerformSelection(DateTime utcTime, BaseDataCollection
         public virtual Security CreateSecurity(Symbol symbol, IAlgorithm algorithm, MarketHoursDatabase marketHoursDatabase, SymbolPropertiesDatabase symbolPropertiesDatabase)
         {
             // by default invoke the create security method to handle security initialization
-            return SecurityManager.CreateSecurity(algorithm.Portfolio, algorithm.SubscriptionManager, marketHoursDatabase, symbolPropertiesDatabase,
+            return SecurityManager.CreateSecurity(algorithm.GetAvailableDataFeeds(), algorithm.Portfolio, algorithm.SubscriptionManager, marketHoursDatabase, symbolPropertiesDatabase,
```
mchandschuh: It seems strange that invoking `CreateSecurity` will automatically add subscriptions for all available quote types... wouldn't it make more sense to push this responsibility to the caller of this function.

This function would accept `QuoteType` and pass it along to whomever needs it, but it would be a shame if I only want `QuoteType.Trade` data and all of a sudden now I'm being bombarded with `QuoteType.Quote` data as well.

Algorithm/QCAlgorithm.cs
andrewhart098: Ultimately, with this PR, I'm trying to make it so that the 'Reader' method used by the ISubscriptionDataSourceReader can be configured by the user. The problem I found with moving the idea of the 'default data feed' outside the scope of the Algorithm is that everything hinges on the SubscriptionDataConfig, which is created within the scope of the algorithm.  It's the Type of the SubscriptionDataConfig that determines which Reader method to use.

While technically still in the scope of the algorithm, perhaps SubscriptionManager is a better place to hang the 'default data feeds'?

Common/Data/SubscriptionDataConfigList.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /// <summary>
+        /// <see cref="Symbol"/> for which this class holds <see cref="SubscriptionDataConfig"/>
+        /// </summary>
+        public Symbol Symbol { get; private set; }
+
+        /// <summary>
+        /// Assume that the InternalDataFeed is the same for both <see cref="SubscriptionDataConfig"/>
+        /// </summary>
+        public bool IsInternalFeed
```
mchandschuh: Internal feed is a per subscription property, not per symbol.

Common/Interfaces/IAlgorithm.cs
```diff
@@ -133,6 +133,11 @@ AlgorithmStatus Status
             set;
         }
 
+        Dictionary<SecurityType, List<TickType>> AvailableDataFeeds
```
mchandschuh: Seems strange that the algorithm exposes available dates  feeds, but rather should probably be told what feeds are available.

Common/Securities/SecurityManager.cs
```diff
@@ -327,9 +328,18 @@ protected virtual void OnCollectionChanged(NotifyCollectionChangedEventArgs chan
...
-            var config = subscriptionManager.Add(factoryType, symbol, resolution, dataTimeZone, exchangeHours.TimeZone, isCustomData, fillDataForward,
-                extendedMarketHours, isInternalFeed, isFilteredSubscription);
+            // Add the symbol to Data Manager -- generate unified data streams for algorithm events
+            SubscriptionDataConfigList configList = new SubscriptionDataConfigList(symbol);
+
+            // Get the type that will be used in the data feed
+            // Could be more than one for a given security - i.e. More than one subscription needed
+            foreach (var dataFeed in availableDataFeeds[symbol.ID.SecurityType])
+            {
+                var dataFeedType = GetDataFeedType(factoryType, dataFeed);
```
mchandschuh: Reusing the term data feed even makes me uneasy. Read up on the concept of ubiquitous language. 

There can only be on definition of data feed on the system and it's already been taken.

PR #639, opened 2016-11-23
Oanda v2 API

QANTau:
This request updates the Oanda Brokerage and Downloader code to support string type when passing in AccountID (as per documentation on http://developer.oanda.com/rest-live-v20/account-df/).

This request provides the ability to connect using the new v2 AccountID (“-“-delimited string with format “{siteID}-{divisionID}-{userID}-{accountNumber}”) and doesn't impact on passing of legacy "AccountNumber" only.


jaredbroad:
Awesome thank you @QANTau. I can see it won't make a difference to V1 accounts - will it work with V2 accounts? (as all the rest endpoints are /v1/)

QANTau:
As far as I can tell from my testing, this change works fine with all the /v1 endpoints (ie. can pass the old or new AccountID as strings and all is OK) but not the other way around (ie. we can't move to using /v2 endpoints until the AccountID is passed in the new format.

PR #647, opened 2016-12-02
Accern data

devalkeralia:
Added Accern.cs file to read the accern news data from the data folder into a Accern object

Added the AccernDataSplitter project to split the huge accern data gzip files into small chucks of indicidual months

Would be happy to make any suggested changes.

### DIFF
--- a/Common/Data/Custom/Accern.cs
+++ b/Common/Data/Custom/Accern.cs
@@ -1,0 +1,359 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Globalization;
+using System.IO;
+
+namespace QuantConnect.Data.Custom
+{
+    /// <summary>
+    /// Data Source used to get Accern news data
+    /// </summary>
+    public class Accern : BaseData
+    {
+        /// <summary>
+        /// Unique Identifier for the articles
+        /// </summary>
+        public string ArticleId;
+
+        /// <summary>
+        /// A story is a unique company and event pair. Each story that is detected gets a unique ID and several articles can talk about the same story, hence, sharing the same story ID
+        /// </summary>
+        public string StoryId;
+
+        /// <summary>
+        /// Name of the company mentioned In the article
+        /// </summary>
+        public string Name;
+
+        /// <summary>
+        /// Unique global ID of the company, derived from Bloomberg Open Symbology
+        /// </summary>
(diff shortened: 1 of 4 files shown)


Common/Data/Custom/Accern.cs
```diff
@@ -0,0 +1,359 @@
...
+            if (data[36] != "") accernData.ArticleSentiment = Convert.ToDecimal(data[36]);
+            if (data[37] != "") accernData.OverallSourceRank = Convert.ToInt32(data[37]);
+            if (data[38] != "") accernData.EventSourceRank = Convert.ToInt32(data[38]);
+            if (data[39] != "") accernData.EventSourceRank2 = Convert.ToInt32(data[39]);
+            if (data[40] != "") accernData.OverallAuthorRank = Convert.ToInt32(data[40]);
+            if (data[41] != "") accernData.EventAuthorRank = Convert.ToInt32(data[41]);
+            if (data[42] != "") accernData.EventAuthorRank2 = Convert.ToInt32(data[42]);
+            if (data[43] != "") accernData.OverallEventImpactScore = Convert.ToDecimal(data[43]);
+            if (data[44] != "") accernData.EventEntityImpactScore = Convert.ToInt32(data[44]);
+            if (data[45] != "") accernData.EventEntityImpactScore2 = Convert.ToInt32(data[45]);
```
mchandschuh: Nullable values can be a pain, but I think these should probably be nullables, otherwise how will the algorithm be able to discern between a real value of 0 vs it not being specified?

ToolBox/AccernDataSplitter/Program.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        public static string GenerateDirectoryPath(string symbol)
+        {
+            return Path.Combine(Globals.DataFolder, "news", "accern", symbol.ToLower());
+        }
+
+        public static string GenerateFilePath(string symbol, DateTime date)
+        {
+            return Path.Combine(GenerateDirectoryPath(symbol), date.ToString("yyyyMM") + ".csv");
```
mchandschuh: Should the file path generation be added to `LeanData` with all the other file generation helper methods? Perhaps the argument could be made that this new data is 'outside' the scope of `LeanData` -- but if it's in LEAN, it's nice to have all the file pathing logic in one place coupled with the unit tests in `LeanDataTests`

ToolBox/AccernDataSplitter/config.json
```diff
@@ -0,0 +1,11 @@
+﻿{
+    "environment": "desktop",
+
+	"environments": {
+		"desktop": {}
+	},
+
+	// Delete the configuration after loading
+	"accern-data-folder": "",
```
mchandschuh: I know I'm being nit-picky, but this is the input folder, correct? I had to trace the value to prove that to myself, perhaps we rename this guy to `accern-input-data-folder` or even `accern-input-data-directory` -- also, do we need a specific location for the cache directory? Since it's a cache can't we just throw it into `accern-input-data-directory`/`cache` or something similar -- less required configuration is always a good thing :) If we do this, the code would need to create it at start (it's OK to create a dir that already exists, no error would be thrown).

Also, the comment on line 8 seems out of place.

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -191,6 +192,9 @@
     <Compile Include="ZipStreamProvider.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="AccernDataSplitter\config.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
mchandschuh: Typically we set this to copy if newer so if you modify the `config.json` in the bin directory it doesn't get overwritten with each build.

jaredbroad:
Closing for now lets review when higher priority or PR submitted by Accern.

PR #651, opened 2016-12-09
Use Minute as the minimum resolution in GetLastKnownPrice

StefanoRaggi:
This PR fixes a memory issue in backtesting when adding securities at tick resolution, due to the History call in security seeding.

An interesting question is if a security should be seeded at all (in backtesting).
Thoughts ?


jaredbroad:
The idea for backtesting was to seed so the very first bar of a fill
forward algorithm would have data everytime.

PR #657, opened 2016-12-15
Fix holiday bug in SecurityExchangeHours

StefanoRaggi:


Common/Securities/SecurityExchangeHours.cs
```diff
@@ -196,7 +196,7 @@ public DateTime GetNextMarketOpen(DateTime localDateTime, bool extendedMarket)
             do
             {
                 var marketHours = GetMarketHours(time.DayOfWeek);
-                if (!marketHours.IsClosedAllDay && !_holidays.Contains(time.Ticks))
+                if (!marketHours.IsClosedAllDay && !_holidays.Contains(time.Date.Ticks))
```
mchandschuh: Interesting. This is a great change, just curious how it just became an issue? Presumably all consumers of this method were at one point only sending in the `DateTime.Date` value, but this is definitely more correct. Perhaps now that this change is here, we should check consumers to see if they're doing the `.Date` and remove it (why do it twice)?

StefanoRaggi: It wasn't an issue, I just noticed it while working on the other task (early market closes).
I saw that other locations calling `_holidays.Contains(x)` were always passing in `time.Date`, so I just guessed it was a bug, although in this case probably hidden by the `IsClosedAllDay` on the same line :)

I just checked and `GetNextMarketOpen` callers are not doing the` .Date`, so we should be ok.

PR #665, opened 2016-12-22
Log location on disc displayed for local backtests

andrewhart098:
+ Removes broken "Your logs can be downloaded from ..." message at the end of local backtests
+ Api implements a new method, StoreLogs, that returns the location of logs on disc.  
+ Removed unnecessary log processing from BacktestingResultsHandler ProcessLogMessages method
+ Removed unnecessary log processing methods from Api.cs - ReadLogAllowance and UpdateDailyLogUsed

### DIFF
--- a/Engine/Results/BacktestingResultHandler.cs
+++ b/Engine/Results/BacktestingResultHandler.cs
@@ -691,13 +691,12 @@ namespace QuantConnect.Lean.Engine.Results
         }
 
         /// <summary>
-        /// Terminate the result thread and apply any required exit proceedures.
+        /// Terminate the result thread and apply any required exit procedures.
         /// </summary>
         public void Exit() 
         {
-            //Process all the log messages and send them to the S3:
-            var logURL = ProcessLogMessages(_job);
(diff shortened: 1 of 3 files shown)


Engine/Results/BacktestingResultHandler.cs
```diff
@@ -767,7 +767,11 @@ private string ProcessLogMessages(AlgorithmNodePacket job)
         {
             if (_exitTriggered)
             {
-                _api.StoreSystemLogs(job);    
+                // Process the system logs once exit has been triggered. This should occur the second time this method is called
+                if (!_api.StoreSystemLogs(job))
+                {
+                    Log.Error("BacktestingResultsHandler.ProcessLogMessages(): Could not upload the system logs to S3.");
```
mchandschuh: Wouldn't this logging be more appropriate inside the api implementation? This way, the error message would be specific to the implementation vs the logging assuming an S3 implementation

jaredbroad:
Nice work @andrewhart098, this has bugged people for ever.

PR #671, opened 2017-01-02
Add new CancelPending order status

StefanoRaggi:
This new status is being added to solve the following problem with order cancellations:

Working orders (limit or stop) are usually cancelled by calling the `OrderTicket.Cancel()` method. This method is asynchronous in both backtesting and live, so reading the `OrderStatus` immediately after the Cancel call, can result in different values seen over different runs.

The solution introduces a new `CancelPending` value to the `OrderStatus` enum and guarantees that the order will have this status (or the `Canceled` status) when the Cancel method returns (unless there was an error earlier, such as invalid order id). This status is meant to be temporary and will be overwritten/replaced by the `Canceled` value after the brokerage has completed the cancel operation.

### DIFF
--- a/Common/Orders/OrderTypes.cs
+++ b/Common/Orders/OrderTypes.cs
@@ -129,6 +129,11 @@ namespace QuantConnect.Orders
         Filled = 3,
 
         /// <summary>
+        /// Order waiting for confirmation of cancellation
+        /// </summary>
+        CancelPending = 4,
+
+        /// <summary>
         /// Order cancelled before it was filled
         /// </summary>
         Canceled = 5,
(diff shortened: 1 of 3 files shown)


Common/Orders/OrderTypes.cs
```diff
@@ -129,6 +129,11 @@ public enum OrderStatus {
         Filled = 3,
 
         /// <summary>
+        /// Order waiting for confirmation of cancellation
+        /// </summary>
+        CancelPending = 4,
```
jaredbroad: Best to add to the bottom of the enum incase anything depends on the number

StefanoRaggi: Ok, will move it to the bottom with a new number

PR #674, opened 2017-01-05
Disable margin calls by default in live mode

StefanoRaggi:
Margin calls in live mode are now automatically disabled in the `BrokerageSetupHandler.Setup` method, using the `NullMarginCallModel` and `NoMarginCallMarginModel` models.


Common/Securities/NullMarginCallModel.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using QuantConnect.Orders;
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Represents a margin call model which will always return no margin call orders
+    /// </summary>
+    public class NullMarginCallModel : MarginCallModel
+    {
```
mchandschuh: It's more common for the null implementation to directly derive  from the interface vs a base class. It makes behavior more explicit

StefanoRaggi: Currently there is no `IMarginCallModel` interface, I'll create it with the `ExecuteMarginCall` method and implement it in both `MarginCallModel` and `NullMarginCallModel` classes.

mchandschuh: Interesting!  Not sure how we missed adding that when we added the base model. Thanks for adding the interface.

Issue #678, opened 2017-01-06
Leverage not working anymore


stefan-j:
I am unsure how regularly changes are pushed to QuantConnect. However, recently leverage isn't working as usual anymore. 

Buying on margin with both SetHoldings (with arg > 1) and MarketOrder worked on the 5th Jan. Since today SetHoldings seems to have no effect with (arg > 1) and MarketOrder results in a invalid order. I have set the leverage via SetLeverage as well without success.

[Example backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_cc5cf84484811315b311fb5cef0a64c7.html) where SetHoldings(5) results in the same behavior as SetHoldings(1). Changing it to MarketOrder(16000) would result in the order getting filled the same as MarketOrder(8000). On the Oanda brokerage it would result in an invalid order.

stefan-j:
I'm not exactly sure what the changes where, but the leverage defaults to 1 in the initialization (or after).  From a glance this seems to have been introduced in the merging of PR #675 in commit fe58550f447b7d5a6e8645e40f4a026e6690b5f3 .

The workaround I'm currently using is to set the leverage manually _after_ the initialization.

jaredbroad:
Thanks @stefan-j -- fixed in #679

PR #679, opened 2017-01-06
Disable margin calls in live mode and refactor margin call models

StefanoRaggi:
This PR reverts and replaces PR #674 because it caused issue #678.
The margin call model has been refactored and now contains both methods for margin call order handling (generation and execution).

Disabling margin calls is now much simpler:
`Portfolio.MarginCallModel = MarginCallModel.Null;`

List of changes:
- Added a new IMarginCallModel interface
- Renamed MarginCallModel to DefaultMarginCallModel
- Refactored margin call models by moving the GenerateMarginCallOrder method from ISecurityMarginModel to IMarginCallModel
- Disabled margin calls by default for live trading in BrokerageSetupHandler initialization


Common/Securities/IMarginCallModel.cs
```diff
@@ -46,6 +57,12 @@ public static class MarginCallModel
 
         private sealed class NullMarginCallModel : IMarginCallModel
         {
+            public SubmitOrderRequest GenerateMarginCallOrder(Security security, decimal netLiquidationValue, decimal totalMargin,
```
mchandschuh: Will we ever need multiple orders here? Just thinking about generalizing the solution.

StefanoRaggi: If we leave the call site as a loop on holdings, probably we don't need multiple orders:
https://github.com/QuantConnect/Lean/pull/679/files#diff-ea81a360172dd59a9b5b98d829e4f26fR575

Maybe a better generalization (for another PR) could be replacing

`SubmitOrderRequest GenerateMarginCallOrder(Security security, decimal netLiquidationValue, decimal totalMargin, decimal maintenanceMarginRequirement);`

with

`List<SubmitOrderRequest> GenerateMarginCallOrders(SecurityPortfolioManager portfolio);`

?

mchandschuh: I agree this work is for another PR, and that seems like the next reasonable improvement here. We'll want to do that before anyone starts writing their own margin models (probably rare and advanced users only, so they could probably handle the breaking change easily)

PR #687, opened 2017-01-11
Filter consolidator inputs on data resolution

StefanoRaggi:
This update reverts PR #664 and includes an improved filtering solution.

It prevents consolidators and indicators from receiving inputs at a resolution higher than the data subscription resolution, typically fill-forward bars.
Volume-based indicators and consolidators will now be calculated correctly with fill-forward data.

It also fixes the RegressionAlgorithm failing test.

### DIFF
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -482,12 +482,20 @@ namespace QuantConnect.Lean.Engine
                 {
                     foreach (var update in timeSlice.ConsolidatorUpdateData)
                     {
+                        var resolutionTimeSpan = update.Target.Resolution.ToTimeSpan();
                         var consolidators = update.Target.Consolidators;
                         foreach (var consolidator in consolidators)
                         {
                             foreach (var dataPoint in update.Data)
                             {
-                                consolidator.Update(dataPoint);
+                                // Filter out data with resolution higher than the data subscription resolution.
+                                // This is needed to avoid feeding in higher resolution data, typically fill-forward bars.
+                                // It also prevents volume-based indicators or consolidators summing up volume to generate
+                                // invalid values.
+                                if (algorithm.Time == dataPoint.EndTime.RoundUp(resolutionTimeSpan))
+                                {
+                                    consolidator.Update(dataPoint);
+                                }
                             }
 
                             // scan for time after we've pumped all the data through for this consolidator
--- a/Engine/DataFeeds/SubscriptionSynchronizer.cs
+++ b/Engine/DataFeeds/SubscriptionSynchronizer.cs
@@ -99,11 +99,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                         var clone = subscription.Current.Clone(subscription.Current.IsFillForward);
                         clone.Time = clone.Time.ExchangeRoundDown(configuration.Increment, subscription.Security.Exchange.Hours, configuration.ExtendedMarketHours);
 
-                        // do not add fill-forward data if rounded down to the previous day
-                        if (!clone.IsFillForward || clone.Time.Date == subscription.Current.Time.Date)
-                        {
-                            packet.Add(clone);
-                        }
+                        packet.Add(clone);
 
                         if (!subscription.MoveNext())
                         {


Engine/AlgorithmManager.cs
```diff
@@ -482,12 +482,20 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, IDataFeed feed, I
...
                             foreach (var dataPoint in update.Data)
                             {
-                                consolidator.Update(dataPoint);
+                                // Filter out data with resolution higher than the data subscription resolution.
+                                // This is needed to avoid feeding in higher resolution data, typically fill-forward bars.
+                                // It also prevents volume-based indicators or consolidators summing up volume to generate
+                                // invalid values.
+                                if (algorithm.Time == dataPoint.EndTime.RoundUp(resolutionTimeSpan))
+                                {
+                                    consolidator.Update(dataPoint);
```
mchandschuh: Would this be better placed on the time slice create method? Also,  for volume consolidators,  fill forward would also copy the volume property. Perhaps we can change the clone method to not populate volume when filling forward.

jaredbroad:
We considered it but can have multiple consolidators so seemed like it
wouldnt fit?

Fillforward approach is specifically filtering out when no trades vs fixing
the root of it which was piping data beyond the subscription into the
consolidators.

With that OBV & other volume indicators are then fixed by changing the
subscription of data going into them vs editing the data itself or making
assumptions on how the subscriptions would be used + we were hoping to
preserve the separation of consolidatora and indicators. Downside of that
approach is it might be difficult to get fill forward data, and use a
volume indicator

On Jan 11, 2017 7:15 PM, "Michael" <notifications@github.com> wrote:

*@mchandschuh* commented on this pull request.
------------------------------

In Engine/AlgorithmManager.cs
<https://github.com/QuantConnect/Lean/pull/687#pullrequestreview-16266763>:

>                          var consolidators = update.Target.Consolidators;
                         foreach (var consolidator in consolidators)
                         {
                             foreach (var dataPoint in update.Data)
                             {
-                                consolidator.Update(dataPoint);
+                                // Filter out data with resolution
higher than the data subscription resolution.
+                                // This is needed to avoid feeding in
higher resolution data, typically fill-forward bars.
+                                // It also prevents volume-based
indicators or consolidators summing up volume to generate
+                                // invalid values.
+                                if (algorithm.Time ==
dataPoint.EndTime.RoundUp(resolutionTimeSpan))
+                                    consolidator.Update(dataPoint);

Would this be better placed on the time slice create method? Also, for
volume consolidators, fill forward would also copy the volume property.
Perhaps we can change the clone method to not populate volume when filling
forward.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/QuantConnect/Lean/pull/687#pullrequestreview-16266763>,
or mute the thread

mchandschuh:
Maybe get the best of both worlds by updating OBV to check for fill forward and ignore volume from fill forward bars. The consumer has enough data  to do the right thing now that I think about it.

StefanoRaggi:
While I agree that indicators using volume such as `OBV` can be updated to ignore volume from fill-forward bars, the issues this PR is trying to address are also others (as you explained in great detail in #636):

1. Fill-forward data sent back in time: #664 was an attempt to solve this one by filtering data in `SubscriptionSynchronizer`, but caused `RegressionAlgorithm` test to fail

2. Algorithms with multiple symbols at different resolutions: fill-forward bars (at higher resolution) are sent into consolidators used internally by all indicators, causing indicators to output incorrect values

2. `TradeBarConsolidator` summing volume

I believe this PR should fix all these issues, I looked at other possible places for the filtering (like `TimeSlice.Create`) and this one seems the best to me, for these two main reasons:
- _no changes to data in generated slices_
- _no duplicate data as input to consolidators and indicators_

mchandschuh:
My reasoning for suggesting time slice create is that other components use that function, such as the  history provider. It looks like given this condition, we don't want to update the consolidator, so why add it to the time slice consolidator update data if it isn't going to be used to update consolidators?

StefanoRaggi:
Maybe its possible that one piece of data in the time slice consolidator update data is needed by one consolidator but not another one ?

mchandschuh:
I thought each consolidator got its own update list. If that's incorrect then you're right, it may not work in time slice create.

StefanoRaggi:
Yes, AlgorithmManager loops over the same update data multiple times, once for each consolidator

PR #688, opened 2017-01-12
Added LookupSubscriptionConfigDataTypes to lookup Subscription Type

andrewhart098:
+ Fixes a bug in the SecurityManager.CreateSecurity that would not generate the appropriate SubscriptionDataConfigs for Canonical Options and Futures
+ New method added, LookupSubscriptionConfigDataTypes, to SubscriptionManager that uses the SecurityType, Resolution and whether the symbol is canonical, to determine the subscription Types.

### DIFF
--- a/Common/Data/SubscriptionManager.cs
+++ b/Common/Data/SubscriptionManager.cs
@@ -15,9 +15,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using NodaTime;
+using QuantConnect.Data.Auxiliary;
 using QuantConnect.Data.Consolidators;
 using QuantConnect.Data.Market;
+using QuantConnect.Util;
 
 namespace QuantConnect.Data
 {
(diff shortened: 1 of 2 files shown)


Common/Data/SubscriptionManager.cs
```diff
@@ -180,7 +183,27 @@ public IReadOnlyList<TickType> GetDataTypesForSecurity(SecurityType securityType
...
+        public IEnumerable<Type> LookupSubscriptionConfigDataTypes(SecurityType symbolSecurityType, Resolution resolution, bool isCanonical)
+        {
+            if (isCanonical)
+            {
+                return new List<Type>() {typeof(ZipEntryName)};
+            }
 
+            if (resolution == Resolution.Tick)
+            {
+                return new List<Type>() { typeof(Tick) };
```
mchandschuh: Is this valid for `SecurityType.Equity.Base`? 
If this method is not intended to be called with that argument then maybe a guard clause to check and throw would be helpful.

quant1729:
I think the bug is quite subtle and it is in the core, so it makes sense to add unit test(s) that fail on old code, but pass on new one.

andrewhart098:
See #689

PR #694, opened 2017-01-16
Moved futures margin folder into 'future' folder. Tests green.

quant1729:
Moved futures margin folder into 'future' folder. Tests green.


Common/Securities/Futures/FutureMarginModel.cs
```diff
@@ -220,7 +220,7 @@ private MarginRequirementsEntry[] LoadMarginRequirementsHistory(string symbol)
         {
             lock (DataFolderSymbolLock)
             {
-                var directory = Path.Combine(Globals.DataFolder, "margins");
+                var directory = Path.Combine(Globals.DataFolder, Path.Combine(SecurityType.Future.ToLower(), "margins"));
```
jaredbroad: Following the pattern for other asset types it should be: "future/{market}/margin" since the symbols in the margin folder are specific to the USA market. When we support international futures markets we'll have other margins as well.

quant1729:
Updated PR. My bad. Now it should be correct place to store those files.

PR #697, opened 2017-01-16
Add null check in SymbolJsonConverter.ReadJson()

andrewhart098:
SymbolJsonConverter.ReadJson() method throws an uncaught NullReference exception when it can't find the "ID" attribute in the json.

### DIFF
--- a/Common/SymbolJsonConverter.cs
+++ b/Common/SymbolJsonConverter.cs
@@ -72,7 +72,14 @@ namespace QuantConnect
                 underlyingSymbol = new Symbol(SecurityIdentifier.Parse(underlying["ID"].ToString()), underlying["Value"].ToString());
             }
 
-            return new Symbol(SecurityIdentifier.Parse(jobject["ID"].ToString()), jobject["Value"].ToString(), underlyingSymbol);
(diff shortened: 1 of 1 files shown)


Common/SymbolJsonConverter.cs
```diff
@@ -72,7 +72,14 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
...
-            return new Symbol(SecurityIdentifier.Parse(jobject["ID"].ToString()), jobject["Value"].ToString(), underlyingSymbol);
+            var symbolId = jobject["ID"];
+
+            if (symbolId == null)
+            {
+                return null;
+            }
+
+            return new Symbol(SecurityIdentifier.Parse(symbolId.ToString()), jobject["Value"].ToString(), underlyingSymbol);
         }
```
quant1729: Not sure if it is a real use-case, but if we check jobject["ID"] == null, it's also worth checking if other fields may be missing as well (e.g. "Value", etc).

mchandschuh: Just  some  context: the reason we check it for null is backwards compatibility with the previous symbol definition, which didn't have the security identifier property

PR #707, opened 2017-01-18
Fix makes sure that NaNs, Infs from QLNet don't disrupt our calculati…

quant1729:
…ons.

Cosmetic fix in IQFeed test.

### DIFF
--- a/Common/Securities/Option/QLOptionPriceModel.cs
+++ b/Common/Securities/Option/QLOptionPriceModel.cs
@@ -232,7 +232,13 @@ namespace QuantConnect.Securities.Option
         {
             try
             {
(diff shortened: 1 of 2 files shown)


quant1729:
I think ideally this should be solved using a sum type one level higher in Greeks/Results classes. So I thought about adding a flag "IsReady" confirming that computations were successful, and if yes, then deliver all numbers.  As for QLNet, they port/copy from QuantLib trying to minimize new bugs from porting. As a result of this, NPV is double(not decimal) and there are no safe functions in entire library to check if values were produced. We catch exceptions. In C++ it was ok, in C# this is bad. Will suggest QLNet authors to change this pattern.

PR #710, opened 2017-01-24
Custom Data for Python Support

AlexCatarino:
Tidy up Python Management
Adds Import Custom Data support


Engine/DataFeeds/ZipEntryNameSubscriptionDataSourceReader.cs
```diff
@@ -52,7 +53,7 @@ public ZipEntryNameSubscriptionDataSourceReader(IDataFileProvider dataFileProvid
             _config = config;
             _date = date;
             _isLiveMode = isLiveMode;
-            _factory = (BaseData) Activator.CreateInstance(config.Type);
+            _factory = _factory = (BaseData)ObjectActivator.GetActivator(config.Type).Invoke(new object[] { config.Type });
```
jaredbroad: @andrewhart098 will this be OK?

PR #712, opened 2017-01-25
Added basic file locking mechanism. First cut. Tested - Windows/Linux

quant1729:
Added basic file locking mechanism. First cut. Tested - Windows/Linux. Locking is not referred by other LEAN code yet.


mchandschuh:
Might want to look into named mutexes

PR #713, opened 2017-01-25
RollingWindow(N) now becomes ready after N additions

StefanoRaggi:
Previously it became ready only after N+1 additions.
Also updated all indicators dependent on this behavior of RollingWindow.

Fixes this issue:
https://www.quantconnect.com/forum/discussion/1680/why-rolling-window-did-not-work-as-expected

### DIFF
--- a/Common/Indicators/RollingWindow.cs
+++ b/Common/Indicators/RollingWindow.cs
@@ -150,9 +150,9 @@ namespace QuantConnect.Indicators
                 {
                     _listLock.EnterReadLock();
 
-                    if (i >= Count)
+                    if (i < 0 || i > Count - 1)
                     {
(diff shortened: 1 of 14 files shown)


mchandschuh:
May want to add documentation that even though `rollingWindow.IsReady` returns true that it does not guarantee that `rollinWindow.MostRecentlyRemoved` will have a value. Perhaps we can add an additional to to the property to check for `_mostRecentlyRemoved == null`

StefanoRaggi:
Good catch, thank you. 

I forgot to update the check in `MostRecentlyRemoved` property from:

` if (!IsReady)`

to

` if (Samples <= Size)`

Will fix now

PR #715, opened 2017-01-30
Fixing fill forward resolution race bug in backtesting and live. Test.

quant1729:
Fill forward resolution was initialized too late from time to time. HistoryAlgo test failed due to this under CPU stress. In this test fillforward enumerator produced bars with 1 day data resolution, but 1 minute fill forward resolution. As a result of this, around 200 unexpected bars arrived to the algo while only 5 were expected. Algo statistics produced incorrect numbers. Now test is updated to scream if unexpected bars have arrived.

### DIFF
--- a/Engine/DataFeeds/FileSystemDataFeed.cs
+++ b/Engine/DataFeeds/FileSystemDataFeed.cs
@@ -405,9 +405,29 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// </summary>
         private void UpdateFillForwardResolution()
         {
-            _fillForwardResolution.Value = _subscriptions
-                .Where(x => !x.Configuration.IsInternalFeed)
-                .Select(x => x.Configuration.Resolution)
+            UpdateFillForwardResolution(_subscriptions.Select( x => x.Configuration ));
+        }
+
+        /// <summary>
+        /// Updates the fill forward resolution by checking specified subscription configurations and
+        /// selecting the smallest resoluton not equal to tick
+        /// </summary>
+        /// <param name="subscriptionConfigs">Subscription configurations list</param>
+        private void UpdateFillForwardResolution(IEnumerable<SubscriptionDataConfig> subscriptionConfigs)
+        {
+            _fillForwardResolution.Value = GetFillForwardResolution(subscriptionConfigs);
+        }
+
+        /// <summary>
+        /// Returns the fill forward resolution by checking specified subscription configurations and
+        /// selecting the smallest resoluton not equal to tick
+        /// </summary>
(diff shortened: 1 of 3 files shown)


jaredbroad:
Did you find a way to semi-reliably repeat the issue? It would be nice to replicate it locally first before merging.

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -405,9 +405,29 @@ public void Exit()
...
+        {
+            _fillForwardResolution.Value = GetFillForwardResolution(subscriptionConfigs);
+        }
+
+        /// <summary>
+        /// Returns the fill forward resolution by checking specified subscription configurations and
+        /// selecting the smallest resoluton not equal to tick
+        /// </summary>
+        /// <param name="subscriptionConfigs">Subscription configurations list</param>
+        private TimeSpan GetFillForwardResolution(IEnumerable<SubscriptionDataConfig> subscriptionConfigs)
```
mchandschuh: What value does the separation of methods here provide? It looks the same as what was there previously.

mchandschuh: Ah, I now see... it allows you to call update *before* adding the subscription in the `ConfigureEnumerator` method.

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -502,6 +522,10 @@ private IEnumerator<BaseData> ConfigureEnumerator(SubscriptionRequest request, b
             // optionally apply fill forward logic, but never for tick data
             if (request.Configuration.FillDataForward && request.Configuration.Resolution != Resolution.Tick)
             {
+                var subscriptionConfigs = _subscriptions.Select(x => x.Configuration).Concat(new[] { request.Configuration });
+
+                UpdateFillForwardResolution(subscriptionConfigs);
```
mchandschuh: This could potentially be done less expensively by checking the current fill forward resolution, and if this subscription is lower and non-tick, replace the value.

quant1729:
To replicate the issue locally just run HIstoryAlgorithm (or any other algorithm with resolution of 1 hour, 1 day), and do it with CPUSTRES.EXE. Regression test fails with around 200 unexpected bars in the algo.

Engine/DataFeeds/FileSystemDataFeed.cs
quant1729: In this case we become dependent on the logic of the UpdateFillForwardResolution() itself. I would like fillforward resolution update logic to be encapsulated in the method. If one wants to change it later, this change should not affect the lines above.

PR #719, opened 2017-01-31
Add missing DataNormalizationMode in history requests

StefanoRaggi:
Previously, `DataNormalizationMode` in history requests was always being set to the `Adjusted` default value, so `SetDataNormalizationMode` calls were not being respected in `SubscriptionDataReaderHistoryProvider.GetHistory` and warmup.


jaredbroad:
Awesome thank you

PR #725, opened 2017-02-02
Remove call to Environment.Exit in Program.cs

andrewhart098:
+ Remove call to `Environment.Exit(0)` in `Program.cs`.  This was a hacky fix to a simple threading problem.  The threading problem has fixed.


andrewhart098:
Closing, see #776

PR #730, opened 2017-02-07
Fixing trello #1874 bug: Algo dies with System.InvalidOperationExcept…

quant1729:
…ion: Collection was modified; enumeration operation may not execute.

### DIFF
--- a/Common/Securities/CashBook.cs
+++ b/Common/Securities/CashBook.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using QuantConnect.Data;
+using System.Collections.Concurrent;
 
 namespace QuantConnect.Securities
 {
@@ -34,7 +35,7 @@ namespace QuantConnect.Securities
         /// </summary>
         public const string AccountCurrency = "USD";
 
-        private readonly Dictionary<string, Cash> _currencies;
+        private readonly ConcurrentDictionary<string, Cash> _currencies;
 
         /// <summary>
         /// Gets the total value of the cash book in units of the base currency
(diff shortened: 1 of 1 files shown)


jaredbroad:
Thanks @quant1729, did you identify who the two threads were? Could there be any state issues by the two threads accessing it at once?

mchandschuh:
@jaredbroad - my thoughts exactly. This 'fix' may just be hiding a bug, ideally we would synchronize the two threads accessing this data, otherwise it will be non deterministic

quant1729:
in both modes 3 - 4 threads touch cashbook throughout its lifetime. Measured with my counter in class methods. Example threads in backtest: brokerage transaction handler, algorithm manager, main engine. Example in live mode: IB brokerage, algorithm manager, main engine. 

@mchandschuh Can you please elaborate on this with code details? As the author of the code I believe you know more on the ideas behind cashbook in MT environment.

mchandschuh:
The concern is not specific to the cash book, but rather any data being accessed by multiple threads. The exception is caused because one thread is reading the data while another is modifying it, so there exists the potential for a bug. Is it safe for a thread to 'miss'  a value? The only way to make that determination is to figure out who is reading and writing the data and do they need to be synchronized to ensure data is not dropped from the read operation

jaredbroad:
TODO on a future date: review all the live trading cross thread calls and use locks to sync. This PR points to larger work needing done in the asynchronous calls in live (brokerage events, result handler etc)

For now PR approved as most common cause is the result handler displaying the portfolio values; which isn't critical and is recalculated a second later; so safe enough to merge in. 

@quant1729 please log when these events occur so the information/request isn't lost and if it causes follow on errors we'll know where they come from.

quant1729:
Added todo item to trello (on me). Adding logging now.

PR #732, opened 2017-02-07
Fixed bug in FxcmBrokerage.GetAccountHoldings

StefanoRaggi:
Previously, holdings generated with multiple orders were reported incorrectly, considering only the quantity of the last order.
Now, multiple positions for the same symbol are correctly aggregated.

### DIFF
--- a/Brokerages/Fxcm/FxcmBrokerage.cs
+++ b/Brokerages/Fxcm/FxcmBrokerage.cs
@@ -333,7 +333,22 @@ namespace QuantConnect.Brokerages.Fxcm
         {
             Log.Trace("FxcmBrokerage.GetAccountHoldings()");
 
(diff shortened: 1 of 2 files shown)


Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -333,7 +333,22 @@ public override List<Holding> GetAccountHoldings()
...
+                .Select(group => new Holding
+                {
+                    Symbol = group.Key,
+                    Type = group.First().Type,
+                    AveragePrice = group.Sum(x => x.AveragePrice * x.Quantity) / group.Sum(x => x.Quantity),
+                    ConversionRate = group.First().ConversionRate,
+                    CurrencySymbol = group.First().CurrencySymbol,
+                    Quantity = group.Sum(x => x.Quantity)
+                })
+                .ToList();
```
jaredbroad: Nice use of LINQ

PR #733, opened 2017-02-08
Changed time rules to explicitly require a user to specify only posit…

quant1729:
…ive numbers in offsets for both live and backtest modes. Tests.

This commit refers to trello ticket #1770. In backtests this negative offset is ignored and event is fired at the moment when markets are open. In live mode, we don't ignore negative offset and attempt to fire event N minutes before markets are open. I propose to fix this issue by explicitly requiring user to specify only positive numbers in offsets for both live and backtest modes.


quant1729:
[Schedule Event - Streamline backtesting and live trading. Ideally making backtesting more accurate.](https://trello.com/c/jHGQmjA7/1770-schedule-event-streamline-backtesting-and-live-trading-ideally-making-backtesting-more-accurate)

PR #734, opened 2017-02-08
Trello #1862. Updated BasicTemplateOptionsAlgorithm to be less strict…

quant1729:
…. New version of the algo should work with any symbol. Updated corresponding regression test.

Namely, new algo: 
doesn't expect the symbol to expire today
doesn't rely on weeklys (most symbols don't have them)
has enough capital to write naked options (just is case some users may want it)
doesn't expect larger chain to exist
trades ATM contract


quant1729:
[Update BasicTemplateOptionAlgorithm to be less strict and trigger on more contracts](https://trello.com/c/F2XQok0e/1862-update-basictemplateoptionalgorithm-to-be-less-strict-and-trigger-on-more-contracts)

PR #737, opened 2017-02-08
Add IDataCacheFileProvider

andrewhart098:
+ Added new interface, `IDataCacheFileProvider`, which defines the ability to cache data files and retrieve the data from cache.
+ Changed the existing `DataFileCacheProvider` class to inherit from the new `IDataCacheFileProvider` interface
+ Added the `IDataCacheFileProvider` to the `IHistoryProvider.Initialize()` method as a parameter with a default value of null
+ Moved the `IStreamReader` from the `QuantConnect.Lean.Engine` project to the `QuantConnect` project so that it can be returned in `IDataCacheFileProvider.Fetch()` method.

These changes, while fairly extensive, should not affect any existing code.


quant1729:
Changes look good to me!

jaredbroad:
General style guide for contributors; method arguments should go from "large" to "small" in general size and significance, they should be grouped in general function, and optional parameters are avoided wherever possible.

andrewhart098:
Closing in favor of https://github.com/QuantConnect/Lean/pull/746

PR #745, opened 2017-02-14
Tick consolidator minor fix

AlexCatarino:
WorkingBar resolution was not set explicitly and that made it incoherent with the Consolidator, since the default is minute. It causes IDataConsolidator.Scan to emit a bar that was already emited


AlexCatarino:
@mchandschuh, I am tagging you for review: 
Do you think this is the right place to include this?

PR #751, opened 2017-02-16
Custom plot utc bug

jaredbroad:
Fixes backtesting and live trading charting bug where data was saved in algorithm time instead of UTC.

### DIFF
--- a/Engine/Results/LiveTradingResultHandler.cs
+++ b/Engine/Results/LiveTradingResultHandler.cs
@@ -564,8 +564,8 @@ namespace QuantConnect.Lean.Engine.Results
         /// Add a sample to the chart specified by the chartName, and seriesName.
         /// </summary>
         /// <param name="chartName">String chart name to place the sample.</param>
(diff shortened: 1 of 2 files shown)


Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1045,7 +1045,7 @@ public void ProcessSynchronousEvents(bool forceProcess = false)
                             else
                             {
                                 // we haven't gotten data yet so just spoof a tick to push through the system to start with
-                                security.SetMarketPrice(new Tick(DateTime.Now, subscription.Configuration.Symbol, price, price));
+                                security.SetMarketPrice(new Tick(time, subscription.Configuration.Symbol, price, price));
```
mchandschuh: May want to convert this time value to the subscription time zone

jaredbroad: Thanks, I traced it through & its not used anywhere (likely why it was DateTime.Now before).

PR #758, opened 2017-02-20
History request by security type

andrewhart098:
Currently, Lean makes all history requests to seed securities as if they are `TradeBars`.  This will not work for securities with data formatted as `QuoteBars`, e.g. Forex. These changes update the history requests that seed securities to pull the security type from the first type in `Security.Subscriptions`.  

While `Security.Subscriptions` is an `IEnumerable`, Equity and Forex data by default should only have one subscription. Options and Futures do not have their prices seeded. If no subscriptions are available, a security type of `TradeBar` is assumed.

The `BrokerageModelSecurityInitializerTests` have been updated to test seeding a security with data formatted as `QuoteBars`.

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -414,6 +414,7 @@ namespace QuantConnect.Algorithm
         /// <returns>A single <see cref="BaseData"/> object with the last known price</returns>
         public BaseData GetLastKnownPrice(Security security)
         {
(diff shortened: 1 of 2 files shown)


StefanoRaggi:
I looked into this as well, 
the solution I found was to get the config with `GetMatchingSubscription` which is already used in other history requests, then passing in `config.Type` instead of `typeof(TradeBar)`

PR #762, opened 2017-02-21
fixes caching issue causing IB error No security definition has been …

rbohac:
…found for the request

Cause: Caching is set up in the InteractiveBrokersBroker.cs for symbon contract details. When the caching was queried it was only using the IB Symbol String eg "AAPL" rather than an identifier that considers if the contract that is being looked up is an equity, option etc. PR uses contract.ToString() rather than contract.Symbol. contract.ToString() has additional data such as "OPT AAPL USD Smart"

Fixes issue in https://trello.com/c/P8ef0zoD/2016-live-ib-paper-no-security-definition-has-been-found-for-the-request

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -613,7 +613,7 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
         private string GetPrimaryExchange(Contract contract)
         {
             ContractDetails details;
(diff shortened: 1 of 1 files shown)


jaredbroad:
Thank you for the PR;

Potentially the ToString() its generating still isn't sufficiently unique and it could cause other latent bugs; could you update it to generate a unique string/hash from the contract data? Maybe something like:

```
private string GetUniqueKey(Contract contract) {
   return string.Format("{0} {1} {2}", contract.ToString(), contract.Strike, contract.Expiration);
}
```
Welcome other suggestions

rbohac:
Added Expiration, Right and Strike to unique key

PR #763, opened 2017-02-21
QuoteBar.Reader will process Cfd data

andrewhart098:
Currently, Cfd's are configured to use `QuoteBars`. The `QuoteBar.Reader` method will not work with Cfd data because of a trailing 0 in the Cfd data.  This change should be temporary until the Cfd data can be reprocessed into full QuoteBars or the trailing zero can be removed.


jaredbroad:
Closing as bug in CFD data & will reprocess there to address issue

PR #768, opened 2017-02-23
Log on live failure

andrewhart098:
+ Changed the Api StoreLog definition  differentiate between backtest and live logs.
+ Added the UserToken to the AlgorithmNodePacket.
+ Use the Api to upload logs when an algorithm fails during initialization when run in live mode.


andrewhart098:
Closing and PRing from another branch

PR #771, opened 2017-02-24
Fixes ib pacing flag

mchandschuh:
We should definitely get a couple more eyes on this, but
I'm fairly certain this is fixing a subtle bug :)

be true for every subsequent iteration of the loop. IOW, once
you get pacing violation then it will always interpret timeouts
as a pacing violation via line 2079. The fix here moves the
declaration of the flag to inside the loop, so that on each
iteration we're guaranteed that it's reset to the false state.


jaredbroad:
Nice catch!

PR #772, opened 2017-02-24
Synchronize results from IB history provider

mchandschuh:
Posting for some early feedback regarding the approach.

The current IB history implementation sychronously executes each `HistoryRequest`
and forms slices and emits them. This causes issues with multiple requests since
the individual requests never get synced. This PR aims to synchronize the results
from each history request.

This PR *also* fixes a would-be latent bug in the time zoning of the resultant trade
bars. It was returning the trade bars time stamped in the slice's time zone instead of
the security's configured time zone.

### This code has not been tested yet!

IEnumerator<BaseData> instances and synchronizing them into a single
IEnumerator<BaseData> instance. This can be handy for many things. This move is
to enable usage in the QuantConnect.Brokerages projects where the IB history
provider impl needs to synchronize multiple trader bar streams into Slices.

This enables easy usages with the enumerator stack, specifically the
SynchronizingEnumerator.

>In addition, the method was updated to return an IList<T> to represent the inner
grouped collection. The current implementation of this method is such that it
emits only full groupings, so this return type change codifies that implementation
detail into the type signature, allowing consumers to depend on it.

here. Each request results in a BaseDataCollection time stamped in the slice time
zone. The separate collections are then synchronized by the SynchronizingEnumerator
and formed into Slice instances.

>I've modified the time zone handling of the raw data. It appears as though we're
modifying the time zone of the data to match the requested sliceTimeZone, but
really the data should get fed to the algorithm in the configured exchange time
zone.

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -35,6 +35,7 @@ using IB = QuantConnect.Brokerages.InteractiveBrokers.Client;
 using IBApi;
 using NodaTime;
 using System.Text;
+using QuantConnect.Data.UniverseSelection;
 
 namespace QuantConnect.Brokerages.InteractiveBrokers
 {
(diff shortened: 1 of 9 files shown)


jaredbroad:
Closing as discussed -- seeking a cross brokerage implementation with full history request support (fillforward etc).

PR #774, opened 2017-02-24
Add start-date parameter to CoarseUniverseGenerator

mushketyk:
When `CoarseUniverseGenerator` generates fundamental data it starts from the date of the latest file in the fundamental data folder.
This works well if one need to generate newer data but did not work for me when I had to generate data for local backtesting from 1st January 2000.

The PR adds a new parameter `start-date` to the `config.json` in `CoarseUniverseGenerator`. This parameter is optional. If it is omitted the code would work as before.

### DIFF
--- a/ToolBox/CoarseUniverseGenerator/Program.cs
+++ b/ToolBox/CoarseUniverseGenerator/Program.cs
@@ -50,6 +50,7 @@ namespace QuantConnect.ToolBox.CoarseUniverseGenerator
             var ignoreMaplessSymbols = false;
             var updateMode = false;
             var updateTime = TimeSpan.Zero;
+            DateTime? startDate = null;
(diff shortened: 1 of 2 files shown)


jaredbroad:
Thank you for the PR @mushketyk! 

 - If possible could you please rename "start-date" to something more specific (so people don't get confused with the algorithm start date) -- e.g. "coarse-universe-generator-start-date"

 - Please remove the data-folder changes from the config; they look like accidental anyway.

Overall looks good and we can merge in once those adjustments are fixed.

mushketyk:
Hi @jaredbroad,

Thank you for your review!

I've renamed the parameter to "coarse-universe-generator-start-date" and changed data directory to "../../../Data" (same is used by Google and Yahoo downloaders), this should change these kind of issues in the future.

PR #778, opened 2017-02-28
Only process the logs in the BacktesstingResultsHandler once

andrewhart098:
IResultHandler.Exit() is called twice in Engine.cs which causes the logs to be processed twice in BacktestingResultsHandler Exit method. These changes will ensure that the logs are processed only on the first call to BacktestingResultsHandler.Exit()


jaredbroad:
Merging with a mental note for future todo; "refactor the IHandler plugins to use a cancellation token instead of the Exit() method".

PR #788, opened 2017-03-09
ApiDataProvider

andrewhart098:
+ Added the ApiDataProvider.cs. The ApiDataProvider is a new instance of the IDataProvider that will check for the data on disc.  If it the data is not present on disc, it will reach out to QuantConnect.com to see if the data is available.
+ Added LeanData.TryParsePath(). This method takes a path and will return a Symbol, DateTime and resolution.  This is useful in the ApiDataProvider that uses the api to retrieve files that are not present on disc.  The IDataProvider only has a path to the file.  Using this method, the ApiDataProvider can extract the useful information from the path to communicate with the api.

### DIFF
--- a/Engine/DataFeeds/ApiDataProvider.cs
+++ b/Engine/DataFeeds/ApiDataProvider.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.IO;
+using QuantConnect.Configuration;
+using QuantConnect.Interfaces;
+using QuantConnect.Logging;
+using QuantConnect.Util;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Lean.Engine.DataFeeds
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


Engine/DataFeeds/ApiDataProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        /// <summary>
+        /// Retrieves data to be used in an algorithm.
+        /// If file does not exist, an attempt is made to download them from the api
+        /// </summary>
+        /// <param name="key">A string representing where the data is stored</param>
+        /// <returns>A <see cref="Stream"/> of the data requested</returns>
+        public Stream Fetch(string key)
+        {
+            if (File.Exists(key))
```
mchandschuh: Does this work with the configured value for days directory?

mchandschuh: Whoops, typo,  meant data directory

PR #789, opened 2017-03-09
Fix IB data time zones for Futures

StefanoRaggi:
Ticks emitted were not respecting the dataTimeZone specified in market hours database, so ticks for some Future symbols were not passed to the algorithm.

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -90,6 +90,8 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
             { Market.CBOE, "CFE" }
         };
 
+        // data time zones by symbol (required and only used for futures)
+        private readonly Dictionary<string, DateTimeZone> _dataTimeZones = new Dictionary<string, DateTimeZone>(); 
 
         /// <summary>
         /// Returns true if we're currently connected to the broker
@@ -1647,6 +1649,13 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
 
             foreach (var tick in ticks)
             {
+                // if we have a cached data time zone for this symbol, perform the conversion 
+                DateTimeZone dataTimeZone;
+                if (_dataTimeZones.TryGetValue(tick.Symbol.Value, out dataTimeZone))
+                {
+                    tick.Time = tick.Time.ConvertTo(TimeZones.NewYork, dataTimeZone);
+                }
+
                 yield return tick;
 
                 if (_underlyings.ContainsKey(tick.Symbol))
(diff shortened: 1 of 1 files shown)


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1647,6 +1649,13 @@ public IEnumerable<BaseData> GetNextTicks()
 
             foreach (var tick in ticks)
             {
+                // if we have a cached data time zone for this symbol, perform the conversion 
+                DateTimeZone dataTimeZone;
+                if (_dataTimeZones.TryGetValue(tick.Symbol.Value, out dataTimeZone))
+                {
+                    tick.Time = tick.Time.ConvertTo(TimeZones.NewYork, dataTimeZone);
```
mchandschuh: Should we use the configured exchange time zone from the market hours database here?

StefanoRaggi: I thought the data time zone is used for the conversion, isn't the exchange time zone only used for reading/converting market hours ?

mchandschuh: Data time zone is the raw tone zone from the data source and exchange time zone is the time zone of data sent into the algorithm. So I was expecting us to convert from data time zone to the exchange home zone instead of new York tz

StefanoRaggi: The change above is in `IDataQueueHandler.GetNextTicks`, the ticks returned by IB have been previously assigned NY time zone here (with the exception of Forex which is set to EST):
https://github.com/QuantConnect/Lean/blob/master/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs#L1787
https://github.com/QuantConnect/Lean/blob/master/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs#L1622

The problem I am trying to solve is that with `AddFuture(symbol, Resolution.Tick)` the algorithm only receives data for futures with NY data time zone:

CL works:
    "Future-usa-CL": {
     "dataTimeZone": "America/New_York",
     "exchangeTimeZone": "America/New_York",

ES doesn't work:
    "Future-usa-ES": {
     "dataTimeZone": "America/Chicago",
     "exchangeTimeZone": "America/Chicago",

Another solution to this problem (without changing code) was to convert all futures market hours to NY time zone, but it didn't seem ideal because usually exchanges publish market hours on their websites in local exchange time, users adding new symbols to the market hours database would need to manually convert the time values to NY time zone (actually not in the case of CME, which reports both time zones on their web site).

jaredbroad: Perhaps instead of double converting we should make "GetBrokerageTime()" return UTC instead of NY time?  Then this method would just convert from UTC -> Data Timezone we specify which makes more sense than hard coding NY-time anywhere.

It seems like we've gotten away with the "if-forex-then-est" statement because we only had 2 assets. We could make it all UTC then remove this part?

mchandschuh:
@jaredbroad -  did you mean convert  from UTC  to the exchange time zone?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
jaredbroad: Meant data timezone; I'm assuming we convert it from data timezone to UTC later in the stack and that is why the data isn't arriving as expected (maybe the conversion takes it out of exchange hours & it gets ignored etc).

Regardless -- If the raw live ticks arrived in the same timezone as the raw backtesting files it would keep things consistent?

mchandschuh: My only concern here was to make sure we are emitting data on the exchange time zone. If that's already covered elsewhere, then no worries, but it looks like we're emitting  data in the data time zone which is incorrect.  Data time zone is for reading raw data and then it should be immediately converted into the exchange time zone for consumption by the algorithm

Issue #773, opened 2017-02-24
Create IEX IDataQueueHandler for Free Live Market Data


jaredbroad:
IEX has rolled out a pretty awesome API for free live streaming market data. This would mean equities data would be free for LEAN users!

https://iextrading.com/developer/#tops-tops-endpoint

It looks like a polling API - so we'd need to collate the subscription requests and poll & parse each second.

PR #795, opened 2017-03-13
Implement IEX IDataQueueHandler

buybackoff:
This PR implements IEX IDataQueueHandler from #773.

The first commit has all code related to the handler. 

In the second commit I had to update Json.Net, because Socket.IO client has the 8.0.1 version as a dependency. Also the current dependency 7.0.1 is almost two years old, while 9.0.1 works on any platform (even .NET 2.0), has no dependencies and has improved performance since 7.0. However, if for some reason that package update is undesirable, I could revert the last commit and will use assembly redirect to an older version in the Toolbox project.

I was able to test basic functionality today before market close.

### DIFF
--- a/Common/Global.cs
+++ b/Common/Global.cs
@@ -627,7 +627,8 @@ namespace QuantConnect
             {"W", "Chicago Board Options Exchange"},
             {"X", "Philadelphia Stock Exchange"},
             {"Y", "BATS Y-Exchange, Inc"},
-            {"Z", "BATS Exchange, Inc"}
+            {"Z", "BATS Exchange, Inc"},
+            {"IEX", "Investors Exchange"},
         };
 
         /// Canada Market Short Codes:
--- a/Tests/Engine/DataFeeds/IEXDataQueueHandlerTests.cs
+++ b/Tests/Engine/DataFeeds/IEXDataQueueHandlerTests.cs
@@ -1,0 +1,215 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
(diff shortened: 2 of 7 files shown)


jaredbroad:
Nice work! Pretty solid first attempt @buybackoff. 

One conceptual mistake we'll have to fix is the use of the market IEX. In LEAN market broadly means "where one ticker represents the same physical thing". So AAPL.NYSE is the same as AAPL.NASDAQ; so they're both classified as Apple.USA. In the same manner IEX is still trading the same physical apple company so its USA market. This is undocumented so its an easy misunderstanding.


Updating Newtonsoft is a wide spread change and unrelated to this PR specifically so it might be best merged in a separate PR. This way the community can merge it in separately and test it all still works as expected. We've had issues upgrading Newtonsoft in the past.

Lets do some more testing tomorrow and if all goes well we can merge into the toolbox. Can you please confirm with the firehose everything works as expected.

jaredbroad:
Please PR the Newtonsoft v9 update only to this branch (https://github.com/QuantConnect/Lean/tree/newtonsoft-v9)

buybackoff:
I will change IEX to USA and remove it from the markets list.

The problem with this PR and JSON.NET is that it won't work as is, so I have to redirect assembly to an older version and then the update won't be required (but I still think it is right thing to do).

I tested firehose after 4:30 and there were very few messages, but it worked.

jaredbroad:
Understood; I'll leave this PR here until we've updated the version then can merge it in without issue.
Please make sure to run all the NUnit tests for the Newtonsoft change (v8 had weird issues serializing the Symbol class).

buybackoff:
In this PR I am redirecting to 7.0, there is no v8. I guess I also need to add the redirect to the Launcher, will test paper trading tomorrow when the market is open.

jaredbroad:
Thank you for the PR @buybackoff; I've dropped an email to IEX reporting the single subscription issue and will ping this thread when they fix it.

Issue #799, opened 2017-03-14
Live oanda environment + warmup + indicator issue


Jay-Jay-D:
This is a slightly modified `BasicTemplateForexAlgorithm`.
```cs
using QuantConnect.Data;
using QuantConnect.Indicators;
using System.Linq;

namespace QuantConnect.Algorithm.CSharp
{
    public class BasicTemplateForexAlgorithm : QCAlgorithm
    {
        IIndicator<IndicatorDataPoint> _indicator;

        public override void Initialize()
        {
            SetStartDate(2013, 10, 07);  //Set Start Date
            SetEndDate(2013, 10, 11);    //Set End Date
            SetCash(100000);             //Set Strategy Cash
            var eurusd = AddForex("EURUSD", Resolution.Minute, market:"oanda");
            AddForex("GBPUSD", Resolution.Minute, market: "oanda");
            AddForex("EURGBP", Resolution.Minute, market: "oanda");
            History(5, Resolution.Daily);
            History(5, Resolution.Hour);
            History(5, Resolution.Minute);            
            var history = History(System.TimeSpan.FromSeconds(5), Resolution.Second);
            foreach (var data in history.OrderBy(x => x.Time))
            {
                foreach (var key in data.Keys)
                {
                    Log(key.Value + ": " + data.Time + " > " + data[key].Value);
                }
            }

            SetWarmUp(120);
            _indicator = SMA(eurusd.Symbol, 120);

        }
        public override void OnData(Slice data)
        {
            // Print to console to verify that data is coming in
            foreach (var key in data.Keys)
            {
                Log(key.Value + ": " + data[key].Time + " > " + data[key].Value);
            }
        }
    }
}
```
Note that the only changes respect to the original are the indicator and the `SetWarmup` method.

When I try to run this algorithm in the **live-oanda environment**, the engine throws this error:
```
2017-03-14T20:04:35.9482036Z ERROR:: AlgorithmManager.Run(): RuntimeError: Consolidators update: System.ArgumentNullException: Received type of TradeBar but expected QuoteBar
Parameter name: data
   at QuantConnect.Data.Consolidators.DataConsolidator`1.Update(IBaseData data) in C:\Users\jjd\Downloads\Lean-master\Common\Data\Consolidators\DataConsolidator.cs:line 46
[...]

PR #801, opened 2017-03-15
Oanda HistoryProvider returns QuoteBars instead of TradeBars

StefanoRaggi:
Solves issue #799


benjaminthedonkey:
The introduction of QuoteBars will help model trade costs  more realistically. It's a big step in the right direction. However, I see that Volume information was not included and I feel this is a great opportunity to do it.   I wonder what  other peoples  thoughts are on this regard.

jaredbroad:
There is no volume available from the API for OANDA or FXCM.

benjaminthedonkey:
@jaredbroad  sorry, I'm talking about tick volume.   I didn't check FXCM but OANDA has it.    [live rates](http://developer.oanda.com/rest-live/rates/)

PR #806, opened 2017-03-19
Backtest plugin

mushketyk:
This PR adds a project for a VisualStudio plugin. The plugin allows to send files for backtesting or simply save them to QuantConnect.

The plugin has few limitations:

* Plugin is not signed, since other QuantConnect projects are not signed. This probably will not allow to publish the plugin to the VisualStudio plugins gallery, but it still can be built and installed manually.

* It would be convenient to open a backtesting page when a user sends a file to backtesting in VisualStudio, but it seems there is no way to form a URL to point to a particular backtest

* The plugin does not have good menu icons and generate boring backtest names


rbohac:
Hey I'm trying to test this out. I have the plugin build and I've added it by double clicking on the .vsix file. It said it added successfully. After reloading visual studio, I'm not sure what to next. I don't see a menu option. Can you give me some guidance?

mushketyk:
Hi @rbohac,

Thank you for your interest in the plugin.
The plugin should add two buttons in the section Tools -> QuantConnect: "Log in" and "Log out". You can use them at any time to log in into your QuantConnect account using your [credentials](https://www.quantconnect.com/account).
Also when you open a solution the plugin adds a context menu item to the Solution Explorer window. It adds an option to save a file to QuantConnect and an option to send a file to a backtest.

PR #808, opened 2017-03-20
Fixes IEX missing Time information

AlexCatarino:
Tick.Time was not set with the information from IEX.

Tick.TickTime was defined as TickType.Quote even when it was a TickType.Trade tick. This distinction is helpful when an algorithm filters out quote ticks.


jaredbroad:
Thanks Alex

PR #809, opened 2017-03-20
RealTimePrice is not set when the exchange is closed

andrewhart098:
+ The `LiveTradingDataFeed` will only update the RealTimePrice of a security if the exchange is open.
+ Added tests to demonstrate that if the subscription is null or the market is closed, the function `LiveTradingDataFeed.SubscriptionShouldUpdateRealTimePrice()` will return false.

### DIFF
--- a/Tests/Engine/RealTimePriceUpdateTests.cs
+++ b/Tests/Engine/RealTimePriceUpdateTests.cs
@@ -1,0 +1,117 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+using QuantConnect.Algorithm.Examples;
+using QuantConnect.Brokerages;
+using QuantConnect.Data.Auxiliary;
+using QuantConnect.Lean.Engine.DataFeeds;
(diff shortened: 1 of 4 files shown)


Tests/Engine/RealTimePriceUpdateTests.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            public TimeZoneOffsetProviderNeverOpen() 
+                : base(TimeZones.NewYork, DateTime.Parse("1/1/2016"), DateTime.Parse("1/1/2018"))
+            {
+            }
+
+            public override DateTime ConvertFromUtc(DateTime utcTime)
+            {
+                // return a date that's always closed for equities
+                return DateTime.Parse("3/18/2017 11:00:00 PM");
```
StefanoRaggi: To avoid an exception with international users, we should specify `CultureInfo.InvariantCulture` here, or use this DateTime constructor: `return new DateTime(2017, 3, 18, 23, 0, 0);`

PR #827, opened 2017-04-04
Prevent crash TWS 964

edrouwendaal:
If TWS not started prevent signal block in IB.Connect

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -457,6 +457,7 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
 
                         // no response, disconnect and retry
                         _client.ClientSocket.eDisconnect();
+                        signal.issueSignal();
                         messageProcessingThread.Join();
 
                         // if existing session detected from IBController log file, log error and throw exception


jaredbroad:
Welcome to LEAN and thank you for the submission @edrouwendaal!

StefanoRaggi:
@edrouwendaal Thank you for the PR.

I just installed TWS 964 locally, but wasn't able to repeat the crash (with both 960 and 964).
Perhaps you could post a portion of the log file showing the problem.

Could you please give us some more detail on the type of error/crash you are experiencing ?
Are you using the default IB configuration settings in config.json ?

AnshulYADAV007:
Hi @jaredbroad, 

I am new to this forum(Lean in particular and Github in general) but genuinely wants to contribute to the software in whatever way I can. I have a strong computer science background and I know C# language well enough. Can you please mentor me during my initial days? or recommend someone whom I can ask for help. Maybe do some skype calls. 

Thanks
Anshul

edrouwendaal:
The crash occurred at InteractiveBrokerage.cs line 461, it would hang there forever (.Join).  It threw an exception that the reader could not read beyond the stream. Signaling line 439, caused successful continuation. Unfortunately I am not able to reproduce the error. I will report back when I encounter this again but for now it seems fixed.

edrouwendaal:
screenshot

jaredbroad:
Thank you @edrouwendaal

PR #830, opened 2017-04-05
Trim all trailing 0s when generating Lean formatted lines

andrewhart098:
Remove all trailing 0s in LeanData.GenerateLine(). This creates smaller data files and prevents large numbers of trailing zeros on data points.  This change will apply to all data written with LeanData.GenerateLine.

### DIFF
--- a/Tests/Common/Util/ExtensionsTests.cs
+++ b/Tests/Common/Util/ExtensionsTests.cs
@@ -15,6 +15,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using Newtonsoft.Json;
 using NUnit.Framework;
 using QuantConnect.Securities;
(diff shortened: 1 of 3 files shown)


andrewhart098:
@mchandschuh - I see the extension method.  Updating PR to use it.

Tests/Common/Util/ExtensionsTests.cs
```diff
@@ -216,6 +217,14 @@ public void SafeDecimalCastRespectsLowerBound()
             Assert.AreEqual(decimal.MinValue, output);
         }
 
+        [Test]
+        public void NormalizeDecimalReturnsNoTrailingZeros()
+        {
+            var input = 1.72340000m;
+            var output = input.Normalize();
+            Assert.IsFalse(output.ToString(CultureInfo.InvariantCulture).EndsWith("0"));
```
mchandschuh: I made this comment from my phone but wasn't seeing it pop-up here -- apologies if you get it twice:

I think we need to test three cases here, one of which I think would have failed the initial implementation.

  * Trailing zeros after decimal `1.200`
  * Trailing zeros before decimal `1200` -- this one may have failed the initial impl
  * No trailing zeros `1234.5678`

Can use `[TestCase]` like [this](https://github.com/nunit/docs/wiki/TestCase-Attribute):
```csharp
[Test]
[TestCase(1.200, "1.2")]
[TestCase(1200, "1200")]
[TestCase(123.456, "123.456")]
public void NormalizeDecimalReturnsNoTrailingZeros(decimal input, string expectedOutput)
{
    var output = input.Normalize();
    Assert.AreEqual(expectedOutput, output);
}
```

andrewhart098: Cool, thanks mike.  I've never really used the TestCase attribute in Nunit.  Very clean!

PR #846, opened 2017-04-11
Minor improvement loading assembly

edrouwendaal:


edrouwendaal:
Prevent wrong load, example: ".TestAlgo" vs ".TestAlgo2" vs ".TestAlgoA" the wrong one could be loaded when trying to load TestAlgo

Brokerages/InteractiveBrokers/run-ib-controller.bat
```diff
@@ -2,7 +2,7 @@
 
 :: IB Controller Sourced-Modified from https://github.com/ib-controller/ib-controller
 @echo off
-set TWS_MAJOR_VRSN=960
+set TWS_MAJOR_VRSN=964
```
jaredbroad: Thank you @edrouwendaal; we can't merge this part. The docker container is pinned at 960 for now. Do you have the install.sh script for 964?

PR #850, opened 2017-04-17
Remove BaseData underlying symbol from Future symbol

StefanoRaggi:


mchandschuh:
Just a thought I had and not sure if it matters, but there may be places in
the code where we check the IsDerivative property on Symbol, as this will
now return false for futures.

StefanoRaggi:
@mchandschuh There is no `IsDerivative` property on Symbol, maybe you meant `IsCanonical` ?
This one (and a few similar checks somewhere else) has been updated to reflect the change.
https://github.com/QuantConnect/Lean/blob/master/Common/Symbol.cs#L165

mchandschuh:
As always, sounds like you've got it covered :)

For some reason I thought there was an IsDerivative property some where,
must have been thinking of the canonical check. Sounds like you did the
necessary analysis here. Thanks for easing my concern :)

PR #865, opened 2017-04-24
Work completed for issue #824 "Liquidate to take order tag"

### LINKED ISSUE #824: Liquidate to take order tag
When liquidating it can cause orders to occur but it's not 100% clear from the output in the trade list that this was due to a "Liquidate()" call. It would be nice if Liquidate had an optional variable to set an order tag.

oswaldozapata:
Now when Liquidating, the tag field on the trade list will show as "Liquidated" to indicate that this was due to a Liquidate() call. This is to resolve #824

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -632,7 +632,7 @@ namespace QuantConnect.Algorithm
                     }
                     else
                     {
-                        Transactions.CancelOrder(order.Id);
+                        Transactions.CancelOrder(order.Id,"Liquidated");
                     }
                 }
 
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -632,7 +632,7 @@ public List<int> Liquidate(Symbol symbolToLiquidate = null)
                     }
                     else
                     {
-                        Transactions.CancelOrder(order.Id);
+                        Transactions.CancelOrder(order.Id,"Liquidated");
```
mchandschuh: Perhaps we could accept the string as an argument with a default value of liquidated. This would be consistent with the other methods that create orders and accept a tag.

oswaldozapata: Hey thank you for your response, I will update this.

oswaldozapata:
ok I updated the code

PR #868, opened 2017-04-25
Read/write new futures data format

andrewhart098:
The new futures data format uses decimals rather than long integers.  This PR both writes and reads futures data using decimals rather than longs and does not apply any scaling factor when reading and writing futures data.


andrewhart098:
Added to this PR a general framework for calculating expiry dates of future contracts programmatically.

PR #869, opened 2017-04-25
Handle corrupt zip files in data cache providers

StefanoRaggi:
Instead of leaving zip exceptions to the data feed enumerator stack, we now log file names of corrupt zip files, return null and continue.

The file system data feed should now handle corrupt files like missing files.

### DIFF
--- a/Engine/DataFeeds/ZipDataCacheProvider.cs
+++ b/Engine/DataFeeds/ZipDataCacheProvider.cs
@@ -74,21 +74,25 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
                     if (dataStream != null)
                     {
-                        _zipFileCache.AddOrUpdate(filename,
-                            x =>
+                        Lazy<CachedZipFile> existingEntry;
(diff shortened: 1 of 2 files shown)


Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -74,22 +74,26 @@ public Stream Fetch(string key)
...
+                        if (!_zipFileCache.TryGetValue(filename, out existingEntry))
+                        {
+                            try
                             {
                                 var newItem = new CachedZipFile(ZipFile.Read(dataStream), filename);
                                 stream = CreateStream(newItem.ZipFile, entryName);
-                                return newItem;
-                            },
-                            (x, existingEntry) =>
+                                _zipFileCache.TryAdd(filename, new Lazy<CachedZipFile>(() => newItem));
```
jaredbroad: @andrewhart098 do you know if we needed/relied on the previous behavior where it recreates the stream part each time?

PR #870, opened 2017-04-27
Fix for issue #854

### LINKED ISSUE #854: When PortfolioValue is less than Zero, Halt Backtest
Sometimes with high leverage & volatile stocks its possible to go negative in portfolio value. There's nothing stopping LEAN at the moment from continuing. We should halt the backtest.

oswaldozapata:
It was added a new control to avoid going negative in portfolio value, Each second Lean engine will verify the TotalPortfolioValue, if this is less or equal to zero then Lean will stop the execution .
This will resolve #854

### DIFF
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -115,7 +115,7 @@ namespace QuantConnect.Lean.Engine
 
                 if (_algorithm.Portfolio.TotalPortfolioValue <= 0)
                 {
-                    return "Portfolio value is less or equal to zero";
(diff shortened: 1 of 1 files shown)


jaredbroad:
The proposed solution won't be deterministic. The check should be in the primary algorithm loop; and only applies to backtesting. Please update to fix this and can merge in.

e.g. Potentially here: 
https://github.com/QuantConnect/Lean/pull/870/files#diff-2e5bf591f5af42f335c0daaea8b78b7eR286

oswaldozapata:
I will update this, thanks Jared :)

PR #872, opened 2017-04-27
Fix SetBrokerageModel resetting leverage when called after AddSecurity

StefanoRaggi:


Algorithm/QCAlgorithm.cs
```diff
@@ -877,8 +877,14 @@ public void SetBrokerageModel(IBrokerageModel model)
                 // update models on securities added earlier (before SetBrokerageModel is called)
                 foreach (var security in Securities.Values)
                 {
+                    // save the leverage specified in AddSecurity, just in case the SecurityInitializer overwrites it
+                    var leverage = security.Leverage;
```
jaredbroad: A kitten dies every time we merge a hack.. ideally we should detect that AddSecurity leverage has been set and only save the leverage off if its changed. How feasible is that without breaking everyone's algorithms :)?

StefanoRaggi:
Closing this PR, as fixing this issue requires more discussion/brainstorming on this point:

- In algorithm.Initialize, do we want to support calls to `SetSecurityInitializer`, `SetBrokerageModel` and `AddSecurity `in any random order ?
- Or, do we want to require these calls to be in a specific order such as:
1. SetSecurityInitializer
2. SetBrokerageModel
3. AddSecurity

In the current master code, calling AddSecurity before SetSecurityInitializer or SetBrokerageModel has some issues.

PR #873, opened 2017-04-27
Hull moving average implementation

Jay-Jay-D:


StefanoRaggi:
@Jay-Jay-D Thank you for the PR, I would add a few minor changes:

- Rename the helper method from `HULL` to `HMA`, to be consistent with the other ones (SMA, EMA, etc.)

- Add a new `Hull` value in the `MovingAverageType` enumeration, so it can be used as input to other indicators: https://github.com/QuantConnect/Lean/blob/master/Indicators/MovingAverageType.cs#L21

- Add a new case in to the two `AsIndicator` methods in `MovingAverageTypeExtensions` and `MovingAverageTypeExtensionsTests`:
https://github.com/QuantConnect/Lean/blob/master/Indicators/MovingAverageTypeExtensions.cs
https://github.com/QuantConnect/Lean/blob/master/Tests/Indicators/MovingAverageTypeExtensionsTests.cs

Jay-Jay-D:
Done!

PR #875, opened 2017-04-28
Check for TotalPortfolioValue now is into the primary algorithm loop.

### LINKED ISSUE #854: When PortfolioValue is less than Zero, Halt Backtest

oswaldozapata:
Now the check for TotalPortfolioValue will be performed in the primary algorithm loop;
Resolve #854

### DIFF
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -115,7 +115,7 @@ namespace QuantConnect.Lean.Engine
 
                 if (_algorithm.Portfolio.TotalPortfolioValue <= 0)
                 {
-                    return "Portfolio value is less or equal to zero";
+                    return "Portfolio value is less than or equal to zero";
                 }
 
                 return null;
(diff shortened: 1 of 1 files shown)


Engine/AlgorithmManager.cs
```diff
@@ -262,6 +263,12 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, IDataFeed feed, I
                             results.SamplePerformance(_previousTime.Date, Math.Round((algorithm.Portfolio.TotalPortfolioValue - portfolioValue) * 100 / portfolioValue, 10));
                         }
                         portfolioValue = algorithm.Portfolio.TotalPortfolioValue;
+
+                        if (portfolioValue <= 0)
+                        {
+                            Log.Trace("AlgorithmManager.Run(): Portfolio value is less than or equal to zero");
```
jaredbroad: This won't pass the error message outside of the console (it's a silent death). We should pass it out via a algorithm log or message as well.  e.g. result.SystemDebugMessage();

PR #878, opened 2017-04-28
Calculate futures expiry dates programmatically

andrewhart098:
This PR introduces a method for calculating the futures expiry dates programmatically according to the rules of each futures contract. This should be extended to all futures contracts.

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -580,6 +580,16 @@ namespace QuantConnect
         }
 
         /// <summary>
+        /// Business day here is defined as any day of the week that is not saturday or sunday
+        /// </summary>
+        /// <param name="date">The date to be examined</param>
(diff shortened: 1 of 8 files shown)


Common/Extensions.cs
```diff
@@ -580,6 +580,16 @@ public static DateTime ConvertToUtc(this DateTime time, DateTimeZone from, bool
         }
 
         /// <summary>
+        /// Business day here is defined as any day of the week that is not saturday or sunday
+        /// </summary>
+        /// <param name="date">The date to be examined</param>
+        /// <returns>A bool indicating wether the datetime is a weekday or not</returns>
+        public static bool IsCommonBusinessDay(this DateTime date)
+        {
+            return (date.DayOfWeek != DayOfWeek.Saturday && date.DayOfWeek != DayOfWeek.Sunday);
```
mchandschuh: Maybe `IsWeekday` is a more apt name

PR #880, opened 2017-04-30
Full python support with python libraries (numpy etc)

jaredbroad:
This PR moves LEAN from using IronPython to PythonNet. This change allows importing of Numpy/Scipy etc. The design requires a wrapper layer in C# which passes the events through to the Python object with a threading lock.

It has poor performance now but in the coming days will be improved to run faster.


tomhunter-gh:
Hi @jaredbroad, @AlexCatarino, great effort with this.  Would it be possible to have [the readme in the `Algorithm.Python`](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Python/readme.md) project updated?  Also, I'd be very interested to hear your thoughts on Python 2 vs Python 3.  I was hoping to work with Python 3.. But if we're sticking to Python 2 for the foreseeable future then I'll switch my environment.  Would be really nice to have some way to switch between the two but that would mean referencing a different PythonNet etc.. Thanks

tomhunter-gh:
Also quick question - what's the difference between UCS2 and UCS4?  I tried googling it but I didn't really understand.  Is it related to 32 bit vs 64 bit?  Or Windows vs Linux/Mac?  Thanks

AlexCatarino:
Hi @tomhunter-gh. 
I believe that it won't be difficult to support both Python 2 and 3, but we want to focus on Python 2 at the moment, since there are still some improvements to make, namely on documentation and examples.
On the examples front, we are starting to write some strategies in python using the third party packages we couldn't with IronPython framework.
The difference between UCS2 and UCS4 is related to Windows vs Mac. In both cases, we are working on 64 bit.

tomhunter-gh:
Thanks @AlexCatarino.  In this case I will switch my default environment back to Python 2.7, however I would like to use Python 3 in future.  Please let me know if I can help with enabling Python 3.  

With regards to the UCS2/4 question, I noticed that the HintPaths in the project files reference the UCS4 Python.Runtime.dll.

This StackOverflow answer suggests how to find out if your Python uses UCS2 or UCS4: http://stackoverflow.com/a/1446456/62072

Here's the output from my Python 2 64-bit environment:

> C:\Users\tomhu_000>C:\Users\tomhu_000\Anaconda2\python.exe
> Python 2.7.12 |Anaconda 4.2.0 (64-bit)| (default, Jun 29 2016, 11:07:13) [MSC v.1500 64 bit (AMD64)] on win32
> Type "help", "copyright", "credits" or "license" for more information.
> Anaconda is brought to you by Continuum Analytics.
> Please check out: http://continuum.io/thanks and https://anaconda.org
> \>>> import sys
> \>>> print sys.maxunicode
> 65535

Which according the the SO answer means my Python uses UCS2.  Does this mean that the current Python support in Lean will only work on a Mac?  Unfortunately I'm not hugely familiar with Nuget - is there an easy way to switch all the references to the UCS2 dll?

Many thanks

jaredbroad:
From my (limited) understanding its about the character encoding of the input python documents. With UC2 it assumes only 2 bytes per character; with UC4 it uses 4 bytes per character for the file.py. When the encoder loads the files into Python they cannot be interpreted if the byte encoding is wrong.

If you run this from the LEAN directory via command line using MingW it should convert all the project files across.
```
sed -i -e 's/UCS4/UCS2/' Common/QuantConnect.csproj
sed -i -e 's/UCS4/UCS2/' Algorithm/QuantConnect.Algorithm.csproj
sed -i -e 's/UCS4/UCS2/' Algorithm.CSharp/QuantConnect.Algorithm.CSharp.csproj
sed -i -e 's/UCS4/UCS2/' Algorithm.Python/QuantConnect.Algorithm.Python.csproj
sed -i -e 's/UCS4/UCS2/' AlgorithmFactory/QuantConnect.AlgorithmFactory.csproj
```

tomhunter-gh:
Thanks @jaredbroad 

Sorry to bang on about it - any chance we could have different build configurations like they do in the pythonnet solution to switch between versions and platforms?

image

Understand you might not want to 'pollute' the project files.. Let me know what your thoughts are.  Thanks

PR #881, opened 2017-05-01
Python: Increase execution speed when not using custom data

AlexCatarino:
When not using custom data, do not wrap slice object.


mchandschuh:
Or maybe we can just add a check when removing a security to properly set the flag there as well.

PR #882, opened 2017-05-01
Write Expiry function for Futures with short cuts

AnshulYADAV007:


Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,114 @@
...
+                this.symbol = symbol;
+                this.dateList = list;
+            }
+        }
+
+        [Test]
+        public void TestAllExpiryDateFunctions()
+        {
+            var _path = Directory.GetCurrentDirectory();
+            _path = _path.Substring(0,_path.Length-10) + "\\Common\\Securities\\Futures\\FuturesExpiryFunctionsTestData.xml";
```
StefanoRaggi: @AnshulYADAV007 There is a `TestData` folder in the Tests project where we keep all data files for tests, 
if you move the file there you can get the path using: 
`_path = Path.Combine("TestData", "FuturesExpiryFunctionsTestData.xml")`

AnshulYADAV007: Thanks Stefano, will do that :)

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,114 @@
...
+                    var symbol = _symdata.symbol;
+                    foreach (var dates in _symdata.dateList)
+                    {
+                        var _security = Symbol.CreateFuture(symbol, Market.USA, dates.contractMonth);
+                        var func = FuturesExpiryFunctions.FuturesExpiryFunction(_security.ID.Symbol);
+                        var calculated = func(_security.ID.Date);
+                        var expected = dates.lastTrade;
+                        if (symbolsForNineThirtyEasternTime.Contains(_security.ID.Symbol))
+                        {
+                            expected = expected + new TimeSpan(13,30,0);
```
andrewhart098: You may consider using the Nunit `[TestCase]` attribute (http://nunit.org/?p=testCase&r=2.5) instead of calculating what the expected value of the test should be. It may make it easier to understand what the expected value of a test is if a test is failing.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -51,6 +153,27 @@ public void TestGoldExpiryDateFunction()
...
+            Assert.AreEqual(june2017Func(june2017.ID.Date), new DateTime(2017, 6, 28));
+
+            var july2017 = Symbol.CreateFuture(QuantConnect.Securities.Futures.Metals.Silver, Market.USA, new DateTime(2017, 7, 15));
+            var july2017Func = FuturesExpiryFunctions.FuturesExpiryFunction(july2017.ID.Symbol);
+            Assert.AreEqual(july2017Func(july2017.ID.Date), new DateTime(2017, 7, 27));
+
+        }
+
+        [Test]
         public void TestSP500EMiniExpiryDateFunction()
```
andrewhart098: This test is now failing in TravisCI as well as my machine.

> Test Failure : QuantConnect.Tests.Common.Securities.Futures.FuturesExpiryFunctionsTests.TestSP500EMiniExpiryDateFunction
> Expected: 2017-06-16 13:30:00.000
> But was:  2017-06-16 09:30:00.000

Common/Securities/Futures/FuturesExpiryFunctions.cs
```diff
@@ -40,44 +39,487 @@ public class FuturesExpiryFunctions
...
+                                   select _day).Reverse().ElementAt(0);
+            return lastBusinessDay;
+        }
+
+        /// <summary>
+        /// Method to check whether a given time is holiday or not
+        /// </summary>
+        /// <param name="time"></param>
+        /// <returns></returns>
+        private static bool notHoliday(DateTime time)
```
andrewhart098: Please capitalize the first letter in the naming of methods. `NotHoliday()`

Common/Securities/Futures/FuturesExpiryFunctions.cs
```diff
@@ -40,44 +39,487 @@ public class FuturesExpiryFunctions
...
+        /// Method to retrieve last business day of the month preceding the delivery month. 
+        /// </summary>
+        private static DateTime LastBusinessDayPrecedingMonth(DateTime time)
+        {
+            var precedingMonth = time.AddMonths(-1);
+            var daysInMonth = DateTime.DaysInMonth(precedingMonth.Year, precedingMonth.Month);
+            var lastBusinessDay = (from day in Enumerable.Range(1, daysInMonth)
+                                   let _day = new DateTime(precedingMonth.Year, precedingMonth.Month, day)
+                                   where notHoliday(_day)
+                                   select _day).Reverse().ElementAt(0);
```
andrewhart098: Local variables, even within lambdas, typically do not use the underscore in their name

Common/Securities/Futures/FuturesExpiryFunctions.cs
```diff
@@ -40,44 +39,487 @@ public class FuturesExpiryFunctions
...
+        }
+
+        /// <summary>
+        /// Method to retrieve 9:16 a.m. Central Time (CT) on the business day immediately preceding the third Wednesday of the contract month (usually Tuesday).
+        /// </summary>
+        private static DateTime BusinessDayPrecedingThirdWednesdayAtNineSixteen(DateTime time)
+        {
+            // Get Third Wednesday
+            var daysInMonth = DateTime.DaysInMonth(time.Year, time.Month);
+            var ThirdWednesday = (from day in Enumerable.Range(1, daysInMonth)
```
andrewhart098: Local variables typically do not capitalize the first letter in the name in C#. So, `thirdWednesday`

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
AnshulYADAV007: Hey Stefano,

I have moved my xlm file to TestData folder. But the function
XmlReader.Create() was taking this path (
'C:\Users\Anshul\Documents\work\Lean\Tests\bin\Debug\TestData\FuturesExpiryFunctionsTestData.xml'.
) as input when I passed "TestData\FuturesExpiryFunctionsTestData.xml"
to it so I have to use GetParent twice to fix the issue. I am unable to find
a better solution for this problem.
Thanks

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,114 @@
...
+                QuantConnect.Securities.Futures.Financials.Y2TreasuryNote
+            };
+            IList<String> symbolsForTwelveOclock = new List<String>
+            {
+                QuantConnect.Securities.Futures.Meats.LiveCattle,
+                QuantConnect.Securities.Futures.Meats.LeanHogs
+            };
+            using (XmlReader reader = XmlReader.Create(_path))
+            {
+                List<SymbolData> data = new List<SymbolData>();
```
AnshulYADAV007: I have updated it everywhere. And I will keep it in mind from next time onwards.

AnshulYADAV007: Great technique. I have implemented it.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
AnshulYADAV007: I am really sorry, you told me about this and I forgot to use them.
Thanks Again. I have used them now.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
AnshulYADAV007: That was a silly mistake due to the confusion in timezones, I have updated it.

Common/Securities/Futures/FuturesExpiryFunctions.cs
AnshulYADAV007: Again a Silly mistake. Sorry, Updated it will keep a note of it.

Common/Securities/Futures/FuturesExpiryFunctions.cs
AnshulYADAV007: Will follow this guideline from now onwards, Thanks

Common/Securities/Futures/FuturesExpiryFunctions.cs
AnshulYADAV007: Done, Thanks

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
StefanoRaggi: Did you set the `"Copy if newer"` flag in the new data file ?

In VS, File Properties -> Copy to output directory -> Copy if newer

AnshulYADAV007: Thanks now it is working 👍

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,189 @@
...
+        [TestCase(QuantConnect.Securities.Futures.Indices.NASDAQ100EMini, nineThirtyEasternTime)]
+        [TestCase(QuantConnect.Securities.Futures.Indices.Dow30EMini, nineThirtyEasternTime)]
+        [TestCase(QuantConnect.Securities.Futures.Meats.LiveCattle, twelveOclock)]
+        [TestCase(QuantConnect.Securities.Futures.Meats.LeanHogs, twelveOclock)]
+        [TestCase(QuantConnect.Securities.Futures.Meats.FeederCattle, zero)]
+        [TestCase(QuantConnect.Securities.Futures.Metals.Gold, zero)]
+        [TestCase(QuantConnect.Securities.Futures.Metals.Silver, zero)]
+        [TestCase(QuantConnect.Securities.Futures.Metals.Platinum, zero)]
+        [TestCase(QuantConnect.Securities.Futures.Metals.Palladium, zero)]
+        public void TestExpiryDateFunction(string symbol, string dayTime)
```
andrewhart098: It could be a good idea to split the test cases up in an effort to easily diagnose what is going wrong if one of the test cases fails.  So maybe, a `TestMetalsExpiryDates`, `TestMeatsExpiryDates`, etc.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,189 @@
...
+            {
+                var serializer = new XmlSerializer(typeof(Item[]));
+                data = ((Item[])serializer.Deserialize(reader)).ToDictionary(i=>i.symbol,i=>i.symbolDates);
+            }
+        }
+
+        /// <summary>
+        /// Utility method to do the common work for all the symbols i.e. asserting the 
+        /// </summary>
+        /// <param name="symbol"></param>
```
andrewhart098: Please leave comments for parameters.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,114 @@
...
+            public SymbolData() { }
+            public SymbolData(String symbol, List<Dates> list)
+            {
+                this.symbol = symbol;
+                this.dateList = list;
+            }
+        }
+
+        [Test]
+        public void TestAllExpiryDateFunctions()
```
andrewhart098: I think it would be best to have a separate test for each ExpiryDateFunction. That way, in two years from now when someone makes a minor tweak and these tests start to fail, it's easier to identify which expiry date function is the problem.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,189 @@
...
 
 namespace QuantConnect.Tests.Common.Securities.Futures
 {
     [TestFixture]
     public class FuturesExpiryFunctionsTests
     {
+        /// <summary>
+        /// Dates for Termination Conditions of futures
+        /// </summary>
+        public class Dates
```
andrewhart098: I always like to put any classes needed for a test at the end of the test file rather than the beginning.  Also, does this need to be public?

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -51,27 +228,136 @@ public void TestGoldExpiryDateFunction()
...
+        [Test]
+        public void TestLeanHogsExpiryDateFunction()
+        {
+            var jun2017 = Symbol.CreateFuture(QuantConnect.Securities.Futures.Meats.LeanHogs, Market.USA, new DateTime(2017, 6, 1));
+            var jun2017Func = FuturesExpiryFunctions.FuturesExpiryFunction(jun2017.ID.Symbol);
+            Assert.AreEqual(jun2017Func(jun2017.ID.Date), new DateTime(2017, 6, 14, 12, 0, 0));
+        }
+
+        [Test]
+        public void TestFeederCattleExpiryDateFunction()
```
andrewhart098: There are several tests like this one that I believe are testing the futures contracts with more complex and custom expiry rules. It might be a good idea to add more than one date to test. You may be able to use Test Cases here too

AnshulYADAV007: I have removed these functions

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
AnshulYADAV007: Done that, Thanks

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
AnshulYADAV007: Sorry, I overlooked it. Will keep a note of that from now onwards.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -16,12 +16,189 @@
...
+            public Dates() { }
+            public Dates(DateTime c, DateTime l)
+            {
+                contractMonth = c;
+                lastTrade = l;
+            }
+        }
+
+        /*
+        /// <summary>
```
AnshulYADAV007: Sorry, Removed

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
AnshulYADAV007: Okay, I will also follow that convention. Unfortunately yes it needs to
be public to use XmlSerializer. Otherwise, it gives the following error:

SetUp : System.InvalidOperationException :
QuantConnect.Tests.Common.Securities.Futures.FuturesExpiryFunctionsTests+Item
is inaccessible due to its protection level. Only public types can be
processed.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -27,12 +27,12 @@ namespace QuantConnect.Tests.Common.Securities.Futures
     [TestFixture]
     public class FuturesExpiryFunctionsTests
     {
-        private IDictionary<String, List<Dates>> data = new Dictionary<String, List<Dates>>();
-        private const string zero = "00:00:00";
-        private const string nineSixteenCentralTime = "14:16:00";
-        private const string nineThirtyEasternTime = "13:30:00";
-        private const string twelveOclock = "12:00:00";
-        private const string twelveOne = "12:01:00";
+        private static IDictionary<String, List<Dates>> _data = new Dictionary<String, List<Dates>>();
```
andrewhart098: Why static?

jaredbroad:
Merged in through another PR

Issue #883, opened 2017-05-01
Feature:  ALMA moving average


michaelsrew:
If you find some time one day there is a moving average which trader can use in IB TWS: 
arnaud legoux moving average
[https://www.interactivebrokers.com/en/software/tws/usersguidebook/technicalanalytics/arnaudlegoux.htm](https://www.interactivebrokers.com/en/software/tws/usersguidebook/technicalanalytics/arnaudlegoux.htm)
from:
[http://www.arnaudlegoux.com/index.html](http://www.arnaudlegoux.com/index.html)
There is also an C# implementation on his website which could be adapted.
It would be interesting to play around with that new stuff with your data.
thanks :)

jaredbroad:
Can you please attach the example data of the daily SPY + ALMA generated in TWS/TS? This will be used for the test set.

Jay-Jay-D:
I take it

Jay-Jay-D:
This is my [Excel implementation](https://docs.google.com/spreadsheets/d/1PoFqUP4w9o6JZMuH1Cx8y3Lvi0bPLDpsbAUBs13_1vE/edit?usp=sharing) of the ALMA, I'll use this data in the tests unless @michaelsrew gives us data from TWS.

michaelsrew:
wow i didn't expect it so fast. thanks again. I only wanted to play around with some new stuff and also they are writing that this moving average is the new thing working better that others(ema,..)
back to reality: i am not good with that trading tool(tws) from ib. If i would now how to provide you that data  you will get it by tomorrow. I've seen that Jay-Jay-D started testing his implementation from the 22.2.2013 SPY daily values and they are quite ok compared with the line drawed by the chart-trader in tws. I could write down some values what i see from the charting tool. i am a beginner sorry :(

michaelsrew:
Starting jay-jay-d's data on 22 February 2013 i took around 41 spy day close values starting at this specific date to have something to compare. The alma moving average parameters used in TWS are the same values as he used for his test [ALMA(9,0.85,6.0)]. This
[alma.xlsx](https://github.com/QuantConnect/Lean/files/974340/alma.xlsx)

 file contains only IB SPY daily Close values and the corresponding alma values to compare/test.
If you need more data i will get you some more of tws.
thanks!

Jay-Jay-D:
[...]

PR #885, opened 2017-05-02
ALMA implementation 

Jay-Jay-D:
As this MA needs two arguments (at least), I'm not sure if it should be added to the `MovingAverageTypeExtensions` #883


StefanoRaggi:
As this is another type of moving average, I would add it to the `MovingAverageType` enum.
About the number of arguments, for the other ones it was solved by adding default values for all parameters except for period (for example in `T3MovingAverage`).

Tests/Indicators/ArnaudLegouxMovingAverageTest.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
```
StefanoRaggi: Please add the QC license banner and XML comments, otherwise the build warning count keeps growing :)

StefanoRaggi: Sorry, no need for the XML comments in the test project (the XML documentation flag is disabled in the Tests project)

Jay-Jay-D:
This indicator is new for me, but seems that what matters between the parameters is the ratio. So I'll make some overloads with just the period as argument and eht second argument will be estimated to keep a default ratio, e.g period / sigma = 1.5

Jay-Jay-D:
Done! ALMA was added to the `MovingAverageType` enum.
I also took the liberty of sorting the helpers methods in the `QCAlgorithm.Indicators.cs` file alphabetically.

StefanoRaggi:
Awesome! Thank you, @Jay-Jay-D

PR #887, opened 2017-05-03
Fetch IB option chains from the OCC website

StefanoRaggi:
IB requests for full option chains are rate limited and responses can be delayed up to a minute for each underlying, so we fetch them from the OCC website instead of using the IB API.


jaredbroad:
OCC Permission Received :)

PR #888, opened 2017-05-03
Adds python sklearn package

AlexCatarino:
Adds python [sklearn package](http://scikit-learn.org/) to foundation docker image.


DockerfileLeanFoundation
```diff
@@ -1,5 +1,5 @@
 #
-#	LEAN Foundation Docker Container March-2017
+#	LEAN Foundation Docker Container May-2017
```
StefanoRaggi: Maybe we should also add the day of month here, just in case we need to release more than one update in the same month ?

PR #893, opened 2017-05-06
Add parameter validation in FutureMarginModel.GetInitialMarginRequirement

StefanoRaggi:


jaredbroad:
How was the order getting to this point? Normally pre-order sanity checks prevent placing the order when there's no market price.

PR #898, opened 2017-05-08
Compress futures and options files using System.IO.Compression

andrewhart098:
+ Instead of using 7z to zip potentially thousands of small files in parallel, use System.IO.Compression to zip the processed Lean futures and options files.  This results in better performance and improved overall reliability
+ Fixed bug where all empty csv files were created in the futures and options data.  This bug was present in the data because the `StreamWriter`, when instantiated, creates an empty file.  Many times, the `StreamWriter` would not be used - resulting in empty files.  Created class called `LazyStreamWriter`, which only creates a file if the `WriteLine()` method is called.

### DIFF
--- a/ToolBox/AlgoSeekFuturesConverter/AlgoSeekFuturesConverter.cs
+++ b/ToolBox/AlgoSeekFuturesConverter/AlgoSeekFuturesConverter.cs
@@ -148,11 +148,11 @@ namespace QuantConnect.ToolBox.AlgoSeekFuturesConverter
                             if (!processors.TryGetValue(tick.Symbol, out symbolProcessors))
                             {
                                 symbolProcessors = new List<List<AlgoSeekFuturesProcessor>>(3)
-                                {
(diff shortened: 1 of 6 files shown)


mchandschuh:
FYI, I recall issues with the .Net compression libraries running in mono,
especially in a multithreaded environment.

andrewhart098:
@mchandschuh - Any chance you remember what those issues were? I'm running the data conversion in a docker environment and things may have changed.

The issue I observed using 7z was that the AlgoSeekFuturesConverter would hang while zipping the converted Lean future files.  Using `htop`, I observed several (hundreds?) of running mono processes just sitting there.  I determined that it was hanging on the final conversion step by looking at the logs of both 7z and the c# process running it in parallel.  When I replaced 7z for the final compression step with the native System.IO.Compression, I no longer had this error. I have yet to find inconsistencies in the converted futures data which would indicate that the final compression step is creating invalid or incorrect futures data.

jaredbroad:
@mchandschuh the mono-team improved the System.IO stability in Mono 4.8+

PR #908, opened 2017-05-11
Fixes 18 warnings with code CS1572

AnshulYADAV007:
In this PRm I fixed the Warnings with code CS1572, there were 5 warnings in
this category. They are of the form:
XML comment has a param tag for ___, but there is no parameter by that name

Solution to these warnings is simple to Remove the tag from the XML
comments. Mostly they are the codes which are not refactored
properly.

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -369,7 +369,6 @@ namespace QuantConnect.Algorithm
         /// <param name="period">The period of the FRAMA</param>
         /// <param name="longPeriod">The long period of the FRAMA</param>
         /// <param name="resolution">The resolution</param>
-        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value)</param>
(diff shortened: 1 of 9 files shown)


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -369,7 +369,6 @@ public FilteredIdentity FilteredIdentity(Symbol symbol, TimeSpan resolution, Fun
         /// <param name="period">The period of the FRAMA</param>
         /// <param name="longPeriod">The long period of the FRAMA</param>
         /// <param name="resolution">The resolution</param>
-        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value)</param>
         /// <returns>The FRAMA for the given parameters</returns>
         public FractalAdaptiveMovingAverage FRAMA(Symbol symbol, int period, int longPeriod = 198, Resolution? resolution = null)
```
StefanoRaggi: Actually this warning is pointing to a bug here :)
All indicator helper methods have an optional selector parameter, we should add the missing parameter and leave the XML comment.

PR #911, opened 2017-05-12
fixes 4 warnings with code CS0162

AnshulYADAV007:
In this PR, the warnings I am fixing are corresponding to the code CS0162
"Unreachable code detected"
There are four instances of this warning.
1) One in LeanData.cs, which can be fixed by removing the break
   statement after the if/else code
2) and rest in OpeningBreakoutAlgorithm, These can be fixed by using
   C# preprocessor directives instead of const for plotting and logging controls

### DIFF
--- a/Algorithm.CSharp/OpeningBreakoutAlgorithm.cs
+++ b/Algorithm.CSharp/OpeningBreakoutAlgorithm.cs
@@ -13,6 +13,10 @@
  * limitations under the License.
 */
 
+// plotting and logging controls
+#define ENABLE_PLOTTING
+#undef ENABLE_ORDER_UPDATE_LOGGING
+
 using System;
 using System.Linq;
 using QuantConnect.Brokerages;
@@ -57,7 +61,6 @@ namespace QuantConnect.Algorithm.CSharp
         private const string Symbol = "SPY";
 
         // plotting and logging control
-        private const bool EnablePlotting = true;
         private const bool EnableOrderUpdateLogging = false;
         private const int PricePlotFrequencyInSeconds = 15;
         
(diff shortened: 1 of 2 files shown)


Algorithm.CSharp/OpeningBreakoutAlgorithm.cs
```diff
@@ -13,6 +13,10 @@
  * limitations under the License.
 */
 
+// plotting and logging controls
+#define ENABLE_PLOTTING
+#undef ENABLE_ORDER_UPDATE_LOGGING
```
StefanoRaggi: Although it's exactly the same, I personally find it more readable to have a commented #define instead:
// #define ENABLE_ORDER_UPDATE_LOGGING

Algorithm.CSharp/OpeningBreakoutAlgorithm.cs
```diff
@@ -57,7 +61,6 @@ public class OpeningBreakoutAlgorithm : QCAlgorithm
         private const string Symbol = "SPY";
 
         // plotting and logging control
-        private const bool EnablePlotting = true;
         private const bool EnableOrderUpdateLogging = false;
```
StefanoRaggi: Shouldn't this line also be removed ?

AnshulYADAV007: it is used at line 380 ( if (LiveMode || orderEvent.Status.IsFill() || EnableOrderUpdateLogging)), I didn't want to touch that if you want I can update

StefanoRaggi: You need to touch that, otherwise you have two flags (one const + one preprocessor directive) :)

As @jaredbroad doesn't like preprocessor directives, in order to remove the build warnings we'll have to leave the bool constants and add `#pragma warning` directives instead (removing the `#define` statements you added)

AnshulYADAV007: Okay I will do that, Thanks

PR #921, opened 2017-05-17
Fixes for regression tests

AnshulYADAV007:
[TestFixtureSetUp] of FactorFileTests.cs was setting "data-folder" in Config.cs to "./TestData". This was causing test failures for the tests corresponding to RegressionTests.cs. This commit removes the line editing "data-folder" and moves the files from TestData to Data.


Tests/Engine/FactorFileTests.cs
```diff
@@ -28,7 +28,6 @@ public class FactorFileTests
         [TestFixtureSetUp]
         public void Setup()
         {
-            Config.Set("data-folder", "./TestData");
             Globals.Reset();
         }
```
StefanoRaggi: We can safely remove the entire method now, the `Globals.Reset` call was only needed because of the previous line.

PR #944, opened 2017-05-25
Add example cfd data

andrewhart098:
The data is for XAUUSD. Tick, minute and second files match the dates of the example forex data (20140501 - 20140515). Total, the data is about 20MB.


jaredbroad:
Next time we update the samples I think we should trim all these by half, e.g. maybe 7 days instead of 14. Repo is a 106Mb zip now

PR #955, opened 2017-06-01
Data utils

andrewhart098:
+ Added ability to return a list of symbols from the OandaSymbolMapper, rather than a list of strings. 
+ Added the LeanDataReader that can read a specified lean data file directly without using the rest of the Lean data enumeration stack.


ToolBox/LeanDataReader.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Enumerate over the tick zip file and return a list of BaseData.
+        /// </summary>
+        /// <returns>IEnumerable of ticks</returns>
+        public IEnumerable<BaseData> Parse()
+        {
+            var factory = (BaseData) ObjectActivator.GetActivator(_config.Type).Invoke(new object[0]);
+            ZipFile zipFile;
+            using (var unzipped = Compression.Unzip(_zipPath, out zipFile))
```
jaredbroad: This zipfile should be tidied up after or ideally streamed in memory

PR #958, opened 2017-06-03
Python loader

jaredbroad:
Addresses loader issue and speeds up python


AlgorithmFactory/Python/Wrappers/AlgorithmPythonWrapper.cs
```diff
@@ -870,5 +868,56 @@ public void SetParameters(Dictionary<string, string> parameters)
...
             _baseAlgorithm.SetParameters(parameters);
         }
+
+        /// <summary>
+        /// Creates Util module
+        /// </summary>
+        /// <returns>PyObject with utils</returns>
+        private PyObject ImportUtil()
+        {
+            var code =
```
jaredbroad: This should be put into a separate file; copied to the output directory and loaded as a string.

PR #970, opened 2017-06-13
Fixes yahoo downloader split/dividend data and factor files

jameschch:
Fixes download of split and dividend data from Yahoo. Re-adds factor file generator tests.


jaredbroad:
Nice PR thank you @jameschch

PR #973, opened 2017-06-15
Enables implicit conversion for python algorithm

AlexCatarino:
- Updates pythonnet package
This update implemented implicit conversion.
From now on, python algorithms will be able to use string object instead of Symbol object when a method overload requires Symbol.

- Modifies python example algorithms to show implicit conversion benefits

### DIFF
--- a/Algorithm.Python/CustomChartingAlgorithm.py
+++ b/Algorithm.Python/CustomChartingAlgorithm.py
@@ -11,12 +11,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import clr
-clr.AddReference("System")
-clr.AddReference("System.Collections")
-clr.AddReference("QuantConnect.Algorithm")
-clr.AddReference("QuantConnect.Indicators")
-clr.AddReference("QuantConnect.Common")
+from clr import AddReference
+AddReference("System")
+AddReference("System.Collections")
+AddReference("QuantConnect.Algorithm")
+AddReference("QuantConnect.Indicators")
+AddReference("QuantConnect.Common")
 
 from System import *
 from System.Collections.Generic import List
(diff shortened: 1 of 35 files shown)


Algorithm.Python/CustomChartingAlgorithm.py
```diff
@@ -56,7 +56,11 @@ def Initialize(self):
         self.resamplePeriod = (self.EndDate - self.StartDate) / 2000
         
     def OnData(self, slice):
-        self.lastPrice = slice[self.spy].Close
+        if slice["SPY"] is None:
```
jaredbroad: 👍

Algorithm.Python/CustomDataBitcoinAlgorithm.py
```diff
@@ -41,18 +41,17 @@ def Initialize(self):
 
         # Define the symbol and "type" of our generic data:
         self.AddData(Bitcoin, "BTC")
-        self.btc = self.Securities["BTC"].Symbol
-
+        
 
     def OnData(self, data):
-        if self.btc not in data: return
+        if "BTC" not in data: return
```
jaredbroad: Does the symbol method still work as well?

Algorithm.Python/DailyAlgorithm.py
```diff
@@ -11,11 +11,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import clr
-clr.AddReference("System")
-clr.AddReference("QuantConnect.Algorithm")
-clr.AddReference("QuantConnect.Indicators")
-clr.AddReference("QuantConnect.Common")
+from clr import AddReference
```
jaredbroad: What about the rest of the CLR, you don't need that?

AlexCatarino: No. In fact, so far we haven't used any methods other than AddReference.

Algorithm.Python/CustomDataBitcoinAlgorithm.py
AlexCatarino: Yes. I made it work for both cases.

PR #980, opened 2017-06-19
Fix for using daily or hourly resolution market data for futures.

smartquant:
- Open interest should use yyyyMMdd HH:mm in csv file when frequency is daily or hourly
- Filename should be [baseticker]_[quote|openinterest|trade]_yyyyMM.csv


jaredbroad:
Thank you for the PR @smartquant; can you please add a few days of daily futures data in this format for testing of this PR? (ideally in ES or GC).

smartquant:
Hi Jared, I have uploaded an example data set for hourly and daily futures (GC and ES) as suggested converting the example minute data provided in the \Data folder. It took me a while to get my head around it, but I was able to use the consolidators for the job.
I am happy to write a test for loading the daily / hourly data, but I have not found an easy way to do this outside an algorithm, particularly when we want to load a whole future chain. Would need some guidance here.

andrewhart098:
@smartquant - When running the `BasicTemplateFuturesAlgorithm` with your changes (and the resolutions set to Daily)- I see that `Slice.FuturesChain` in `OnData()` has it's `TradeBars` and `QuoteBars` hydrated with the daily data.  I am also able to modify the filter and change the contracts selected. I think one way to test would be to add another futures example algorithm that uses the daily data and add a regression test for it.  

There is also the `LeanDataReader` which could be helpful for writing unit tests.

smartquant:
Thanks @andrewhart098. Indeed the daily and hourly futures data was not correctly generated. I had to adapt the LeadDataReader to read the correct future within the zip file and re-generate the files. Code to generate the data is added as a unit test (but I have commented out the writing to the zip file). 
As suggested I have added a regression test for daily futures data. Note the small bug fix in the Engine to make the history provider work correctly.

PR #982, opened 2017-06-20
Order and Holding Decimalization Support

StefanoRaggi:
This PR updates #404 to latest master


jameschch:
These changes were made prior to the options support, amongst other things, and probably need close examination before progressing further. I am aware of at least one potential issue with the PR as it stands.

jameschch:
I don't mean to be cryptic: the specific issue I'm aware of is that DefaultMarginCallModel was omitted and still has a cast to int on quantity. Unfortunately, this was not covered by unit tests that validated the outputs. My concern is that there might be several other areas in which these subtle issues may be lurking.

It's pleasing that you're pursing this PR but I consider the code to be fairly stale given that there have been such significant changes since then, specifically: validation of minimum lot size and options support. I had tested fairly thoroughly in live trading, but I would now consider this testing to be partial at best given that my current production build that trades with decimal quantities has drifted a fair bit from this PR.

StefanoRaggi:
@jameschch Thank you for your comments on this PR.
First of all, sorry for delaying this follow through, should have been done months ago, about 1000+ commits earlier :)

Our intention is to get this foundation code working, so we can fully support cryptocurrencies, not sure which will be the first brokerage implementation though. I agree there still might be some bugs in this PR, which I'm sure will be fixed when testing live trading with the new brokerages. 

Our first priority right now is to ensure compatibility with existing algorithms and brokerages (at the moment I'm working on this).

Data/market-hours/market-hours-database.json
```diff
@@ -10225,7 +10225,81 @@
...
-    }
+      ]    
+    },
+        "Forex-bitfinex-[*]": {
+            "dataTimeZone": "UTC",
+            "exchangeTimeZone": "UTC",
+            "sunday": [
+                {
+                    "start": "00:00:00",
+                    "end": "1.00:00:00",
```
andrewhart098: What happens when either a `MarketOnOpenOrder` or `MarketOnCloseOrder` is submitted if the exchange is always open?

StefanoRaggi: If the brokerage accepts orders 24/7, these order types should be rejected in `BrokerageModel.CanSubmitOrder`

jameschch: Since submitting this I also realized that the exchange in question does not operate during Chinese new year, which might need addressing at some point. For this PR, I have just supplied what is necessary to define a new market for the purposes of backtesting using decimalized order quantities.  You're right that any non supported order types will be rejected with an exception, however, this is implemented in the brokerage integration.

jaredbroad:
Personally I welcome our new crypto overlords.

PR #1002, opened 2017-06-30
Fix open interest bug setting value to zero for contracts added intraday

StefanoRaggi:
When using a filter function creating a dynamic chain universe, contracts added during the day would have the value of open interest set to zero.

### DIFF
--- a/Common/Data/UniverseSelection/SubscriptionRequest.cs
+++ b/Common/Data/UniverseSelection/SubscriptionRequest.cs
@@ -103,7 +103,13 @@ namespace QuantConnect.Data.UniverseSelection
             Universe = universe;
             Security = security;
             Configuration = configuration;
-            StartTimeUtc = startTimeUtc;
+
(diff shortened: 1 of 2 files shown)


Common/Data/UniverseSelection/SubscriptionRequest.cs
```diff
@@ -103,7 +103,13 @@ public IEnumerable<DateTime> TradableDays
             Universe = universe;
             Security = security;
             Configuration = configuration;
-            StartTimeUtc = startTimeUtc;
+
+            // open interest data comes in once a day before market open,
+            // make the subscription start from midnight
+            StartTimeUtc = configuration.TickType == TickType.OpenInterest ?
```
jaredbroad: Suggest we merge this as scaffolding to patch the issue but would be nice to keep brain storming. Perhaps there's a more generic solution? Can we set the start date earlier in the stack?  Could certain subscriptions be full day?

PR #1011, opened 2017-07-10
Implements Jupyter Project

AlexCatarino:
With this new project, we are able to use Jupyter with QuantConnect data for quantitative analysis

### DIFF
--- a/DockerfileJupyter
+++ b/DockerfileJupyter
@@ -1,6 +1,6 @@
 #
-#	LEAN Jupyter Docker Container 20170802
-#	
+#	LEAN Jupyter Docker Container 20170803
+#
 
 # Use base system for cleaning up wayward processes
 FROM quantconnect/lean:foundation
@@ -13,21 +13,17 @@ RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.10.0/tini 
     mv tini /usr/local/bin/tini && \
     chmod +x /usr/local/bin/tini
 
-#Install Jupyter
-RUN pip install jupyter
-	
-# Install Python.NET
-RUN apt-get update && apt-get install -y libglib2.0-dev \
-  && wget --quiet https://github.com/QuantConnect/pythonnet/archive/master.zip \
-  && unzip master.zip && cd pythonnet-master && python setup.py install
+#Install Jupyter and Python.NET
+RUN apt-get update && apt-get install -y git libglib2.0-dev && \
+    pip install jupyter git+https://github.com/QuantConnect/pythonnet
 
 # Setting some environment variables
-ENV WORK /root/Lean/Jupyter/bin/Release
+ENV WORK /root/Lean/Jupyter/bin/Debug
 ENV LEAN /usr/lib/python2.7/Lean
 ENV PYTHONPATH=${LEAN}:${PYTHONPATH}
 
 # Copy Lean files to convinient locations
-COPY ./Jupyter/bin/Release ${LEAN}
+COPY ./Jupyter/bin/Debug ${LEAN}
 COPY ./Jupyter/*.ipynb ${WORK}/Examples/
 RUN rm -irf ${LEAN}/*Python*
 
@@ -38,5 +34,5 @@ CMD jupyter notebook --ip='*' --port=8888 --no-browser --allow-root
 
 # Usage: 
 # docker build -t quantconnect/lean:foundation -f DockerfileLeanFoundation .
-# docker build -t quantconnect/lean:jupyter -f DockerfileJupyter .
-# docker run -v "(absolute to your data folder):/root/Lean/Data" quantconnect/lean:jupyter 
+# docker build -t quantconnect/jupyter -f DockerfileJupyter .
+# docker run -v (absolute to your data folder):/root/Lean/Data:ro quantconnect/jupyter 
--- a/DockerfileJupyter
+++ b/DockerfileJupyter
@@ -1,0 +1,42 @@
+#
+#	LEAN Jupyter Docker Container 20170802
+#	
+
+# Use base system for cleaning up wayward processes
+FROM quantconnect/lean:foundation
+
+MAINTAINER QuantConnect <contact@quantconnect.com>
+
+# Install Tini
+RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.10.0/tini && \
+    echo "1361527f39190a7338a0b434bd8c88ff7233ce7b9a4876f3315c22fce7eca1b0 *tini" | sha256sum -c - && \
+    mv tini /usr/local/bin/tini && \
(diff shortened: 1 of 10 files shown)


DockerfileJupyter
```diff
@@ -0,0 +1,42 @@
...
+#	LEAN Jupyter Docker Container 20170802
+#	
+
+# Use base system for cleaning up wayward processes
+FROM quantconnect/lean:foundation
+
+MAINTAINER QuantConnect <contact@quantconnect.com>
+
+# Install Tini
+RUN wget --quiet https://github.com/krallin/tini/releases/download/v0.10.0/tini && \
```
andrewhart098: Possibly we don't need tini? From the tini project page 

> NOTE:  If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just pass the --init flag to docker run.

DockerfileJupyter
```diff
@@ -0,0 +1,42 @@
...
+    echo "1361527f39190a7338a0b434bd8c88ff7233ce7b9a4876f3315c22fce7eca1b0 *tini" | sha256sum -c - && \
+    mv tini /usr/local/bin/tini && \
+    chmod +x /usr/local/bin/tini
+
+#Install Jupyter
+RUN pip install jupyter
+	
+# Install Python.NET
+RUN apt-get update && apt-get install -y libglib2.0-dev \
```
andrewhart098: Isn't python.Net included in the quantconnect/lean:foundation docker file?

DockerfileJupyter
```diff
@@ -0,0 +1,42 @@
...
+RUN rm -irf ${LEAN}/*Python*
+
+WORKDIR $WORK
+
+ENTRYPOINT [ "/usr/local/bin/tini", "--" ]
+CMD jupyter notebook --ip='*' --port=8888 --no-browser --allow-root
+
+# Usage: 
+# docker build -t quantconnect/lean:foundation -f DockerfileLeanFoundation .
+# docker build -t quantconnect/lean:jupyter -f DockerfileJupyter .
```
andrewhart098: Why not quantconnect/jupyter:latest?

DockerfileJupyter
```diff
@@ -0,0 +1,42 @@
...
+# docker run -v "(absolute to your data folder):/root/Lean/Data" quantconnect/lean:jupyter
```
andrewhart098: May want to recommend that docker mounts the data folder in readonly mode so,
`docker run -v (absolute to your data folder):/root/Lean/Data:ro quantconnect/lean:jupyter `

DockerfileJupyter
AlexCatarino: In the foundation, we have a pythonnet installation that only works in Lean. For instance, if you open python and type "import clr", it says the package is not installed.
We could change the foundation to include this installation method that is more general.

DockerfileJupyter
andrewhart098: The projects in Examples won't work with this run command.  The problem is that the example projects live in the path `~/Lean/Jupyter/bin/Release/Examples` and the data folder is specified in the config as `../../../Data`. Therefore, when in one of the Example projects, the absolute path where Lean is looking for data is `~/Lean/Jupyter/Data` when this run command mounts it at `~/Lean/Data`. The data path should probably be configured to an absolute path in Lean

DockerfileJupyter
AlexCatarino: I thought that since it is part of Lean, it made sense.

DockerfileJupyter
AlexCatarino: Maybe we just move them to the parent directory.

Jupyter/QuantBook.cs
```diff
@@ -0,0 +1,452 @@
...
+        }
+
+        /// <summary>
+        /// Get the Lean <see cref = "Symbol"/> object given its ticker, security type and market.
+        /// </summary>
+        /// <param name="ticker">The asset symbol name</param>
+        /// <param name="type">The asset security type</param>
+        /// <param name="market">The asset market</param>
+        /// <returns><see cref = "Symbol"/> object wrapped in a <see cref = "PyObject"/></returns>
+        public PyObject get_symbol(string ticker, string type = "Equity", string market = null)
```
StefanoRaggi: This method should be renamed to `GetSymbol`

Jupyter/QuantBook.cs
```diff
@@ -0,0 +1,452 @@
...
+
+            return SymbolCache.GetSymbol(ticker).ToPython();
+        }
+
+        /// <summary>
+        /// Gets symbol information from a list of tickers
+        /// </summary>
+        /// <param name="pyObject">The tickers to retrieve information for</param>
+        /// <returns>A pandas.DataFrame containing the symbol information</returns>
+        public PyObject print_symbols(PyObject pyObject)
```
StefanoRaggi: This method should be renamed to `PrintSymbols`

Jupyter/QuantBook.cs
```diff
@@ -0,0 +1,452 @@
...
+                    DataNormalizationMode = dataNormalizationMode,
+                    IncludeExtendedMarketHours = extendedMarket,
+                    DataType = symbol.ID.SecurityType == SecurityType.Equity ? typeof(TradeBar) : typeof(QuoteBar)
+                };
+            });
+
+            return CreateDataFrame(requests, _historyProvider.GetHistory(requests, TimeZones.NewYork), selector);
+        }
+
+        public PyObject GetFundamental(PyObject pyObject, string selector, DateTime? start = null, DateTime? end = null)
```
StefanoRaggi: XML comment is missing for this public method

Jupyter/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="NodaTime" version="1.3.1" targetFramework="net452" />
```
StefanoRaggi: This reference should be updated to version `1.3.4`

idcrisis:
Thank you.

Wonder if exchange data integrity can be signed. 

Also multiplicative UDP broadcast of quotes free. Hourly especially.

Jupyter/AlgorithmAnalysis.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            try
+            {
+                using (Py.GIL())
+                {
+                    _pandas = Py.Import("pandas");
+                }
+
+                // Set the data-folder
+                Config.Set("data-folder", "/root/Lean/Data");
```
jaredbroad: Don't hard code data folder, should be configured.

2aspace (commented):
Je ill

PR #1025, opened 2017-07-20
FXCM Forex Real volume and Transactions ToolBox Project

Jay-Jay-D:


Common/Global.cs
```diff
@@ -335,7 +335,7 @@ public enum MarketDataType
         /// Option chain data
         OptionChain,
         /// Futures chain data
-        FuturesChain
+        FuturesChain,
```
jaredbroad: Should remove this comma

Common/Data/Slice.cs
```diff
@@ -320,9 +320,6 @@ private static DataDictionary<SymbolData> CreateDynamicDataDictionary(IEnumerabl
                 switch (datum.DataType)
                 {
                     case MarketDataType.Base:
-                        symbolData.Type = SubscriptionType.Custom;
-                        symbolData.Custom = datum;
```
jaredbroad: Looks like this was removed by accident

Common/Util/LeanData.cs
```diff
@@ -59,17 +60,17 @@ public static string GenerateLine(IBaseData data, SecurityType securityType, Res
                     switch (resolution)
                     {
                         case Resolution.Tick:
-                            var tick = (Tick) data;
+                            var tick = (Tick)data;
```
jaredbroad: Lots of white space changes, please revert.

Common/Util/LeanData.cs
```diff
@@ -238,8 +239,25 @@ public static string GenerateLine(IBaseData data, SecurityType securityType, Res
                             throw new ArgumentOutOfRangeException("resolution", resolution, null);
                     }
                     break;
-            }
 
+                case SecurityType.Base:
+
+                    var vol = data as ForexVolume;
```
jaredbroad: As the ForexVolume is a custom data type quite specific to FXCM this would be classified as a custom data type. To avoid polluting the LEAN code base we'd prefer these are kept isolated. Imagine we applied the code pattern below to the 1 Million custom data types in quandl -- it would be a mess! So best to do any CSV writing inside the project and not make any changes to *LEAN* Data writer. If required you can tweak the ToCSV to be public and access it from your ForexVolume Toolbox project.

Common/Util/LeanData.cs
```diff
@@ -375,15 +393,15 @@ public static string GenerateZipEntryName(Symbol symbol, DateTime date, Resoluti
                 case SecurityType.Cfd:
                     if (isHourOrDaily)
                     {
-                        return string.Format("{0}.csv", 
+                        return string.Format("{0}.csv",
```
jaredbroad: Lots more white space changes. Please remove these -- they just make the diff and merge harder for others later on.

Tests/ToolBox/ForexVolume/ForexVolumeDownloaderTest.cs
```diff
@@ -0,0 +1,342 @@
...
+        public void SetUpTemporatyFolder()
+        {
+            var randomFolder = "ForexVolumeTesting_" + Guid.NewGuid().ToString("N").Substring(startIndex: 0, length: 6);
+            _dataDirectory = Path.Combine(Path.GetTempPath(), randomFolder);
+            _downloader = new ForexVolumeDownloader(_dataDirectory);
+        }
+
+        private string _dataDirectory;
+        private ForexVolumeDownloader _downloader;
+        private readonly Symbol _symbol = Symbol.Create("EURUSD", SecurityType.Base, Market.Decode(code: 20));
```
jaredbroad: This is a FXCM volume indicator. It should be market FXCM?

Tests/ToolBox/ForexVolume/ForexVolumeDownloaderTest.cs
```diff
@@ -0,0 +1,342 @@
...
+
+    [TestFixture]
+    public class ForexVolumeDownloaderTest
+    {
+        [SetUp]
+        public void SetUpTemporatyFolder()
+        {
+            var randomFolder = "ForexVolumeTesting_" + Guid.NewGuid().ToString("N").Substring(startIndex: 0, length: 6);
+            _dataDirectory = Path.Combine(Path.GetTempPath(), randomFolder);
+            _downloader = new ForexVolumeDownloader(_dataDirectory);
```
jaredbroad: Is there anyway to rethink these tests so we don't touch things we dont own? Maybe rethink your generator class methods so it outputs a string which is separately written to disk by your program. This way you can string compare the output of the generator & avoid testing writing to disk,

ToolBox/LeanDataWriter.cs
```diff
@@ -59,9 +59,10 @@ public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory
             {
                 _dataType = TickType.Quote;
             }
-
+            
             // Can only process Fx and equity for now
-            if (_securityType != SecurityType.Equity && _securityType != SecurityType.Forex && _securityType != SecurityType.Cfd)
+            if (_securityType != SecurityType.Equity && _securityType != SecurityType.Forex && _securityType != SecurityType.Cfd
```
jaredbroad: There should ideally be no changes to LDW.

Jay-Jay-D: I'll implement a writer for this custom data. 

But maybe worth consider making an abstraction from the LDW in order to leverage the actual system of files for minute data.

Tests/ToolBox/ForexVolume/ForexVolumeDownloaderTest.cs
Jay-Jay-D: The idea here is to test if the data is written *as* I expect and *where* I expect. The data correctness is tested as you suggest in other tests.

I see as alternative, use the same testing assembly folder and make a temporary folder there, run the assert and delete the temporary folder after that. Is that any better? 

I'll use the Nunit tests cases features to shrink the testing file.

Jay-Jay-D:
@jaredbroad, the implementation was refactored in order to follow the suggestions:

- The core Lean files were reverted to its original state.
- The custom data has its own writer, I don’t use the `LeanDataWriter`.
- The tests were refactored. It stills write to disk but in the assembly folder and all the written data is deleted with and `TestFixtureTearDown` method.

Launcher/config.json
```diff
@@ -155,17 +155,17 @@
...
-        // real brokerage implementations require the BrokerageTransactionHandler
-        "live-mode-brokerage": "InteractiveBrokersBrokerage",
-        "setup-handler": "QuantConnect.Lean.Engine.Setup.BrokerageSetupHandler",
-        "result-handler": "QuantConnect.Lean.Engine.Results.LiveTradingResultHandler",
-        "data-feed-handler": "QuantConnect.Lean.Engine.DataFeeds.LiveTradingDataFeed",
-        "data-queue-handler": "QuantConnect.Brokerages.InteractiveBrokers.InteractiveBrokersBrokerage",
-        "real-time-handler": "QuantConnect.Lean.Engine.RealTime.LiveTradingRealTimeHandler",
-        "transaction-handler": "QuantConnect.Lean.Engine.TransactionHandlers.BrokerageTransactionHandler",
-        "history-provider": "BrokerageHistoryProvider"
+      "live-mode": true,
```
andrewhart098: @Jay-Jay-D - Can you remove these spacing changes in the config?

Tests/TestData/fxVolumeHourly.csv
```diff
@@ -0,0 +1,218 @@
...
+2015/01/01 12:00,0,0
+2015/01/01 13:00,0,0
+2015/01/01 14:00,0,0
+2015/01/01 15:00,0,0
+2015/01/01 16:00,0,0
+2015/01/01 17:00,0,0
+2015/01/01 18:00,0,0
+2015/01/01 19:00,0,0
+2015/01/01 20:00,0,0
+2015/01/01 21:00,0,0
```
andrewhart098: Shouldn't the data go all the way to hour 23:00?

Common/Data/Custom/ForexVolume.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// <param name="date">Date of the requested data</param>
+        /// <param name="isLiveMode">true if we're in live mode, false for backtesting mode</param>
+        /// <returns>
+        ///     Instance of the T:BaseData object generated by this line of the CSV
+        /// </returns>
+        public override BaseData Reader(SubscriptionDataConfig config, string line, DateTime date, bool isLiveMode)
+        {
+            DateTime time;
+            var obs = line.Split(',');
+            if (config.Resolution == Resolution.Minute)
```
andrewhart098: Are you sure that it is only minute data that has this special time formatting? That seems strange. Is minute data the highest resolution for forex volume data in this case? It would make more sense  to me if it applied to all resolutions under hourly.

Tests/ToolBox/ForexVolume/ForexVolumeDownloaderTest.cs
```diff
@@ -0,0 +1,226 @@
...
+using QuantConnect.ToolBox;
+using QuantConnect.ToolBox.FxVolumeDownloader;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+
+namespace QuantConnect.Tests.ToolBox.FxVolume
```
andrewhart098: Try to avoid abbreviations in namespaces, `FxVolume`

Jay-Jay-D: In this case I think it worth because there is a class called `ForexVolume`

Jay-Jay-D: Another alternative is rename the namespace as `FxcmForexVolme`.

Common/Data/Custom/ForexVolume.cs
Jay-Jay-D: I tried to emulate the behavior of the LeanDataWriter, so minute data is saved using the milliseconds format.

Yes, minute is the highest resolution for this data set.

Tests/TestData/fxVolumeHourly.csv
Jay-Jay-D: Yes, that makes sense. Working on it.

Common/Data/Custom/ForexVolume.cs
andrewhart098: Maybe there should be an argument check for resolutions higher than minute?

ToolBox/ForexVolumeDownloader/ForexVolumeDownloader.cs
```diff
@@ -0,0 +1,278 @@
...
+                data.AddRange(Get(symbol, resolution, intermediateStartDate, intermediateEndDate));
+                // If we have 300k observations in memory, write it.
+                if (resolution == Resolution.Minute && counter % 30 == 0)
+                {
+                    writer.Write(data);
+                    data.Clear();
+                }
+            } while (intermediateEndDate != endUtc);
+            // Seems the data has som duplicate values! this makes the writer throw an error.
+            var duplicates = data.GroupBy(x => x.Time)
```
andrewhart098: If you have greater than 300K observations, the code Clears the `data` variable on line 137. If that's the case, won't this code only catch duplicates for those observations that occurred after 300K?

ToolBox/ForexVolumeDownloader/ForexVolumeDownloader.cs
```diff
@@ -0,0 +1,278 @@
...
+        /// <param name="symbol">Symbol for the data we're looking for.</param>
+        /// <param name="resolution">Resolution of the data request</param>
+        /// <param name="startUtc">Start time of the data in UTC</param>
+        /// <param name="endUtc">End time of the data in UTC</param>
+        /// <returns>
+        ///     Enumerable of base data for this symbol
+        /// </returns>
+        public IEnumerable<BaseData> Get(Symbol symbol, Resolution resolution, DateTime startUtc, DateTime endUtc)
+        {
+            startUtc = startUtc.AddDays(value: -1);
```
andrewhart098: Why subtract a day here from the `startUtc` variable?

ToolBox/ForexVolumeDownloader/ForexVolumeWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+        private readonly string _folderPath;
+
+        public ForexVolumeWriter(Resolution resolution, Symbol symbol, string dataDirectory)
+        {
+            _symbol = symbol;
+            _resolution = resolution;
+            _dataDirectory = dataDirectory;
+            _market = _symbol.ID.Market;
+            _folderPath = Path.Combine(new[] {_dataDirectory, "base", _market.ToLower(), _resolution.ToString().ToLower()});
+            if (resolution == Resolution.Minute)
```
andrewhart098: If the resolution is unsupported (minute or second) you might considering throwing and `ArgumentException`

ToolBox/ForexVolumeDownloader/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+                Console.WriteLine("Usage: ForexVolumeDownloader SYMBOLS RESOLUTION FROMDATE TODATE");
+                Console.WriteLine("SYMBOLS = eg EURUSD,USDJPY\n" +
+                                  "\tAvailable pairs:\n" +
+                                  "\tEURUSD, USDJPY, GBPUSD, USDCHF, EURCHF, AUDUSD, USDCAD,\n" +
+                                  "\tNZDUSD, EURGBP, EURJPY, GBPJPY, EURAUD, EURCAD, AUDJPY");
+                Console.WriteLine("RESOLUTION = Minute/Hour/Daily/All");
+                Console.WriteLine("FROMDATE = yyyymmdd");
+                Console.WriteLine("TODATE = yyyymmdd");
+
+#if DEBUG
```
andrewhart098: Is there another way to achieve this without using the #if directive? I would try to avoid it unless absolutely necessary

ToolBox/ForexVolumeDownloader/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+                    resolutions[0] = (Resolution)Enum.Parse(typeof(Resolution), args[1]);
+                }
+
+                var startDate = DateTime.ParseExact(args[2], "yyyyMMdd", CultureInfo.InvariantCulture);
+                var endDate = DateTime.ParseExact(args[3], "yyyyMMdd", CultureInfo.InvariantCulture);
+
+                // Load settings from config.json
+
+                var dataDirectory = Config.Get("data-directory", "../../../Data");
+                //var dataDirectory = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.DesktopDirectory), "TestingFXVolumeData");
```
andrewhart098: Comment cleanup

ToolBox/ForexVolumeDownloader/Program.cs
jaredbroad: Should be removed from production code.

ToolBox/ForexVolumeDownloader/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+                var downloader = new ForexVolumeDownloader(dataDirectory);
+                foreach (var ticker in tickers)
+                {
+                    var symbol = Symbol.Create(ticker, SecurityType.Base, Market.Decode(code: 20));
+                    foreach (var resolution in resolutions)
+                    {
+                        downloader.Run(symbol, resolution, startDate, endDate);
+                    }
+                }
+                Console.WriteLine("\n => Timer: {0} milliseconds.", (DateTime.Now - timer).TotalMilliseconds);
```
andrewhart098: You should probably choose Log.Trace over Console.WriteLine and set the Log Handler

Tests/ToolBox/ForexVolume/ForexVolumeDownloaderTest.cs
andrewhart098: In that case I would remove the `FxVolume` from the namespace entirely and just use `QuantConnect.Tests.ToolBox` as the namespace

andrewhart098:
Looks great @Jay-Jay-D - I left a few comments for your consideration.  There are also currently some merge conflicts that need to be resolved before this PR is merged.

ToolBox/ForexVolumeDownloader/ForexVolumeDownloader.cs
Jay-Jay-D: I don't understand what do you mean.

I changed the comment maybe is a little more clear now. BTW, Have you seen [the tests](https://github.com/Jay-Jay-D/LeanSTP/blob/patch-1/Tests/ToolBox/ForexVolume/ForexVolumeDownloaderTest.cs#L164) for this functionality?

ToolBox/ForexVolumeDownloader/ForexVolumeWriter.cs
Jay-Jay-D: The only calls who calls the `ForexVolumeWriter` is the `ForexVolumeDownloader` and there is already a check for that. In this case the Exception will be redundant.

Common/Data/Custom/ForexVolume.cs
Jay-Jay-D: Done. Also added an argument check for the Symbol

ToolBox/ForexVolumeDownloader/ForexVolumeDownloader.cs
Jay-Jay-D: My bad, I  misinterpretation of the API. Fixing it.

Tests/TestData/fxVolumeHourly.csv
Jay-Jay-D: Done!

Jay-Jay-D:
@andrewhart098 and @jaredbroad , all the suggestions were incorporated and the errors fixed.

ToolBox/ForexVolumeDownloader/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+                Console.WriteLine("FROMDATE = yyyymmdd");
+                Console.WriteLine("TODATE = yyyymmdd");
+                Environment.Exit(exitCode: 1);
+            }
+
+            try
+            {
+                var timer = DateTime.Now;
+                Log.DebuggingEnabled = true;
+                var logHandlers = new ILogHandler[]
```
andrewhart098: `Log.LogHandler = new CompositeLogHandler(logHandlers );`

jaredbroad:
Great work @Jay-Jay-D -- long way but its done!

PR #1030, opened 2017-07-24
Log stack trace when exceptions are thrown Initialize()

andrewhart098:
Log the exception message and stack trace for errors thrown in Initialize.

### DIFF
--- a/Engine/Setup/BrokerageSetupHandler.cs
+++ b/Engine/Setup/BrokerageSetupHandler.cs
@@ -218,7 +218,7 @@ namespace QuantConnect.Lean.Engine.Setup
                     }
                     catch (Exception err)
                     {
-                        AddInitializationError(err.Message);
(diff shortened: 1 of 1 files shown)


Engine/Setup/BacktestingSetupHandler.cs
```diff
@@ -193,7 +193,7 @@ public bool Setup(IAlgorithm algorithm, IBrokerage brokerage, AlgorithmNodePacke
                 catch (Exception err)
                 {
                     Log.Error(err);
-                    Errors.Add("Failed to initialize algorithm: Initialize(): " + err);
+                    Errors.Add("Failed to initialize algorithm: Initialize(): " + err.Message + " StackTrace: " + err.StackTrace);
                 }
```
StefanoRaggi: I believe err.ToString() is the best way to get detailed stack traces, including inner exceptions recursively

PR #1032, opened 2017-07-24
Li xiang618 patch 1

LiXiang618:
These are the commands for installing R base. 
Users could add libraries based on their demands.


idcrisis:
Thanks. Some feature requests if I may:

- Integration of:
 - Numerical Recipes code
 - Knime / Torch
 - Distributed GPU computation over BOINC/Gridcoin 

QuantConnect indeed!

PR #1036, opened 2017-07-26
Support R.Net and R-base in C#

LiXiang618:
Install R-base.
Write CallingRFromCSharp.cs file telling the usage of some APIs.


Algorithm.CSharp/CallingRFromCSharp.cs
```diff
@@ -0,0 +1,68 @@
...
+                engine.SetSymbol("group1", group1);
+                // Direct parsing from R script.
+                NumericVector group2 = engine.Evaluate("group2 <- c(29.89, 29.93, 29.72, 29.98, 30.02, 29.98)").AsNumeric();
+                // Test difference of mean and get the P-value.
+                GenericVector testResult = engine.Evaluate("t.test(group1, group2)").AsList();
+
+                double p = testResult["p.value"].AsNumeric().First();
+                // you should always dispose of the REngine properly.
+                // After disposing of the engine, you cannot reinitialize nor reuse it
+                engine.Dispose();
```
StefanoRaggi: @LiXiang618 Nice job with the R integration !

Only one minor comment:
if `engine.Dispose()` should only be called once, I think the correct place for this call would be the `OnEndOfAlgorithm` method (although this sample algorithm is fine because of the `SetHoldings` call).

LiXiang618: Yeah, you are right. I will change that. Thanks very much!

PR #1062, opened 2017-08-13
Fill forward bid/ask in QuoteBars with Options and Futures

StefanoRaggi:
When Bid or Ask bars are null in the current `QuoteBar`, we now copy forward the previous values (if fill-forward is enabled on the subscription).

Regression stats for `OptionRenameRegressionAlgorithm` have also been updated to reflect this change.


Engine/DataFeeds/TimeSlice.cs
```diff
@@ -380,6 +380,20 @@ private static bool HandleOptionData(DateTime algorithmTime, BaseData baseData,
 
                 case MarketDataType.QuoteBar:
                     var quote = (QuoteBar)baseData;
+
+                    // if fill-forward enabled, copy previous bid/ask when current values are null
+                    if (security.IsFillDataForward)
+                    {
+                        if (quote.Bid == null)
+                        {
+                            quote.Bid = new Bar(contract.BidPrice, contract.BidPrice, contract.BidPrice, contract.BidPrice);
```
jaredbroad: This fill forward should probably be done in the subscription enumerator stack somewhere; so we can pull in the full OHLC bar of the bid/ask objects and not just the current price. 

It would also stop the FF from polluting the time series create pathway.

PR #1064, opened 2017-08-15
Use Tick.Price for fill price when bid/ask spread is not available

andrewhart098:
When an equity at tick resolution is used in Lean, it is seeded with a price using minute data.  For equities, this adds a TradeBar to the security.SecurityCache. When the `ImmediateFillModel` is then used with this security (and tick Bid/Ask spreads are not available), the ImmediateFillModel uses the tradebar that was used to seed the security to generate the fill price.

This PR uses the tick.Price in the `ImmediateFillModel` when the tick bid/ask spreads are not available.


jaredbroad:
Also potentially impacts some algorithms as when the security is removed the price will be reset to $0; where previously it would maintain its last price.

PR #1065, opened 2017-08-16
Fix Options and Futures subscriptions at Tick resolution

StefanoRaggi:
Previously `Tick` resolution subscriptions only received ticks with `TickType.Trade`, now `TickType.Quote` and `TickType.OpenInterest` are received as well.


jaredbroad:
Brain storming how we can solve this without weirdness like: 

`new SubscriptionDataType(typeof(CoarseFundamental), TickType.Trade);`

PR #1070, opened 2017-08-17
Updates Jupyter project

AlexCatarino:
- Modifies the pandas dataframe created from a history request:
History requests should not return a dictionary with a dataframe, but a multi-index dataframe. It is more common to work with multi-index dataframes rather than multi-column.

- Refactors Jupyter Project:
a. QuantBook wraps an instance of QCAlgorithm, enabling the use some its methods 
b. Creates an OptionHistory class to help users to manage historical ootions data requests
c. Fix typo in BasicTemplateAnalysisTemplate notebook

- Implements Indicator History in Jupyter project:
With simple commands, we can use Lean indicators in QuantBook. It fetchs the historical data from the symbol, calculates the indicator and saves the output in a pandas.DataFrame.


Common/Python/PandasConverter.cs
```diff
@@ -0,0 +1,196 @@
...
+            {
+                var index = CreateIndex(data.First().Symbol, data.Select(x => x.Time));
+
+                var pyDict = new PyDict();
+                var columns = "open,high,low,close";
+
+                pyDict.SetItem("low", _pandas.Series(data.Select(x => (double)x.Low).ToList(), index));
+                pyDict.SetItem("open", _pandas.Series(data.Select(x => (double)x.Open).ToList(), index));
+                pyDict.SetItem("high", _pandas.Series(data.Select(x => (double)x.High).ToList(), index));
+                pyDict.SetItem("close", _pandas.Series(data.Select(x => (double)x.Close).ToList(), index));
```
jaredbroad: Please add QuoteBar handling.

jaredbroad:
Good update. Thanks @AlexCatarino

PR #1075, opened 2017-08-17
Fix path comparison in Loader

StefanoRaggi:
These two path expressions evaluate to the same physical directory, differing only by the trailing separator char:
- `new FileInfo(assemblyPath).DirectoryName`
- `AppDomain.CurrentDomain.BaseDirectory`


jaredbroad:
Hillarious line of code. Everyone's debugging broke :D

PR #1081, opened 2017-08-21
Require specification of DataTimeZone on history requests

andrewhart098:
- Renamed `HistoryRequest.TimeZone` to `HistoryRequest.DataTimeZone` using resharper in order to be clearer on what this timezone is used for.
- Removed the parameterless `HistoryRequest` constructor. Defaults are not assumed by the `HistoryRequest` and must be specified.
- Added new `AddData` method that allows users to specify separate exchange time zones and the data time zones.
- Added new method, `Security.IsCustomData` that determines if the security contains at least one subscription that represents custom data 
- Added `HistoryRequest.TickType` to avoid ambiguity between Options/Futures history request with Tick resolution

### DIFF
--- a/Common/Data/HistoryRequest.cs
+++ b/Common/Data/HistoryRequest.cs
@@ -69,7 +69,7 @@ namespace QuantConnect.Data
         /// <summary>
         /// Gets the time zone of the time stamps on the raw input data
         /// </summary>
-        public DateTimeZone TimeZone { get; set; }
+        public DateTimeZone DataTimeZone { get; set; }
 
         /// <summary>
         /// Gets true if this is a custom data request, false for normal QC data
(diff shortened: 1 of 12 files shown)


Jupyter/QuantBook.cs
```diff
@@ -343,7 +343,10 @@ public OptionHistory GetOptionHistory(Symbol symbol, DateTime date, Resolution?
                                         Resolution.Minute, 
                                         underlying.IsExtendedMarketHours, 
                                         underlying.IsCustomData(), 
-                                        DataNormalizationMode.Raw)
+                                        DataNormalizationMode.Raw,
+                                        symbol.ID.SecurityType == SecurityType.Equity ?
```
jaredbroad: This should always evaluate to false as its GetOptionHistory.

Common/Data/HistoryRequest.cs
```diff
@@ -120,6 +126,7 @@ DataNormalizationMode dataNormalizationMode
             DataType = dataType;
             IsCustomData = isCustomData;
             DataNormalizationMode = dataNormalizationMode;
+            TickType = LeanData.GetCommonTickTypeForCommonDataTypes(DataType, symbol.SecurityType);
```
jaredbroad: Passed in then not used

PR #1087, opened 2017-08-24
Add Dispose() method to IBrokerage

andrewhart098:
Currently only `InteractiveBrokersBrokerage` has a dispose method with custom logic for cleaning up resources. This PR adds the call to `IBrokerage.Dispose()` at the end of `Engine.Run()`.

### DIFF
--- a/Brokerages/Brokerage.cs
+++ b/Brokerages/Brokerage.cs
@@ -100,6 +100,14 @@ namespace QuantConnect.Brokerages
         public abstract void Disconnect();
 
         /// <summary>
+        /// Dispose of the brokerage instance
+        /// </summary>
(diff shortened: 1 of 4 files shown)


Brokerages/Brokerage.cs
```diff
@@ -100,6 +100,11 @@ protected Brokerage(string name)
         public abstract void Disconnect();
 
         /// <summary>
+        /// Dispose of the brokerage instance
+        /// </summary>
+        public abstract void Dispose();
```
StefanoRaggi: Perhaps we could make this `virtual` instead (with the empty implementation), so we only need the override in IB brokerage ?

PR #1088, opened 2017-08-24
Fix InteractiveBrokersBrokerageModel fee model with Forex

StefanoRaggi:
When calling `SetBrokerageModel(BrokerageName.InteractiveBrokersBrokerage)` fees were reported as zero for `Forex` orders.


jaredbroad:
Nice PR thank you @StefanoRaggi

PR #1090, opened 2017-08-25
Gdax Brokerage

jameschch:
Current state of changes to introduce support for gdax brokerage.

Planned features are complete. Testing is on-going as are quality improvements. Currently, unit tests pass and provide good coverage of new features. Gdax production trading tests are still being debugged.

Once the support for a crypto security type is ready, master can be merged into this branch and minor changes made to output Crypto rather than Forex as is currently.


Tests/Brokerages/GDAX/GDAXBrokerageTests.cs
```diff
@@ -0,0 +1,313 @@
...
+        string _holdingData;
+        Symbol _symbol;
+        const string _brokerId = "d0c5340b-6d6c-49d9-b567-48c4bfca13d2";
+        const string _matchBrokerId = "132fb6ae-456b-4654-b4e0-d681ac05cea1";
+
+        [SetUp()]
+        public void Setup()
+        {
+            _unit = new GDAXBrokerage("wss://localhost", _wss.Object, _rest.Object, "abc", "MTIz", "pass");
+            _orderData = File.ReadAllText("TestData\\gdax_order.txt");
```
StefanoRaggi: Could be that Travis build is failing because Linux requires `"/"` path separator instead of `"\"` ?

Brokerages/BaseWebsocketsBrokerage.cs
```diff
@@ -0,0 +1,281 @@
...
+                }
+                catch (Exception exception)
+                {
+                    Log.Error(exception);
+                }
+            });
+            _connectionMonitorThread.Start();
+            while (!_connectionMonitorThread.IsAlive)
+            {
+                Thread.Sleep(1);
```
Martin-Molinero: Sleep 1 this seems kind of small

Brokerages/GDAX/GDAXBrokerage.Utility.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            if (order.FilledSize != 0 && order.FilledSize != order.Size)
+            {
+                return Orders.OrderStatus.PartiallyFilled;
+            }       
+            else if (order.Status == "open" || order.Status == "pending" || order.Status == "active")
+            {
+                return Orders.OrderStatus.Submitted;
+            }
+            else if (order.Status == "done" || order.Status == "settled")
```
Martin-Molinero: using string for the order status seems kind of error prone

Martin-Molinero: *undefined string

jameschch: Well, we could null check these strings, but what should we do if we encounter a null? I would say, given the context, we should throw an exception anyway.

I'm not aware that the project adheres strictly to the principle that these kind of strings should be constants. It might help to have constants here.

Brokerages/BaseWebsocketsBrokerage.cs
jameschch: I'm glad you highlighted this section. The reconnection code is from the existing implementation in the oanda brokerage. It's really an organizational call whether the risk of changing oanda is worth reducing duplicated code.

Tests/Brokerages/GDAX/GDAXBrokerageTests.cs
jameschch: Thanks, this fixed the last of the travis problems.

jameschch:
Status update: I've now completed almost all planned production debugging and testing. It is now possible to execute trades and manage orders and receive fills and ticks through a QCAlgorithm with GDAX.

Proving all of this using the production test structure in GDAXBrokerageIntegrationTests is the new focus. 

There are several areas that could be implemented later or not at all but I haven't planned for any other features in this PR. If something is missing: please flag this up. Anyone that wishes to assist in identifying bugs would be very welcome and I will do what I can to get you up and running.

jameschch:
The latest GDAX wss API has been introduced and now provides a best bid and ask ticker channel. This removes the need to reconstruct the order book in the brokerage, so is well worth using. This will mean some significant additional changes but should result in a lot less code.

StefanoRaggi:
Doesn't the ticker channel only receive a message when a trade/match happens ?
In this case we would only receive bid/ask updates when there is a trade: this can be a problem on less traded instruments like ETH/EUR where there can be no trades for minutes, but bid/ask prices are moving.

jameschch:
The work to use the new wss api is complete and orders, tickers and fills should work again.

jameschch:
One point that is worth raising: the exchange does not support update of orders. This method currently throws a not supported exception if called.
It might be preferable to instead cancel and place a new order to simulate an update. However, this might be misleading for algorithms that expect an update and rely on there not being an exposed window between the cancel and place (like with a stop loss).
In terms of the update tests in QuantConnect.Tests.Brokerages.BrokerageTests I'm probably going to add logic to exclude these for gdax.
We should also catch the update order higher in the stack. I'm thinking CanSubmitOrder in BrokerageModel but there might be somewhere more appropriate?

jaredbroad:
@jameschch correct CanSubmitOrder is a good place for that. We used to simulate some orders for brokerages but it is too risky so cancelling is fine until the broker has that feature.

jameschch:
The brokerage integration tests have now passed when live trading a cash account. There are some limitations on the scope of testing which will need to be covered off after further discussion.

I am still running reliability and stability tests and will likely continue this until release. Unit tests will fail currently as the needs of production testing conflicts until we merge in the dedicated Crypto security type.

I have added a service to supply missing conversion rates for non USD denominations to enable more production testing scenarios. I do think the missing conversion rates should be a concern of the crypto security type, or the portfolio itself, rather then the brokerage, but we've not yet discussed how to tackle this.

jaredbroad:
Merged with a fresh PR skipping all the other LEAN changes for cleaner diff #1211

PR #1091, opened 2017-08-25
Fix quantity bug in option exercise/assignment

StefanoRaggi:
When assigning shares of the underlying equity for an option at expiration, the equity quantity was incorrectly set equal to the option quantity. Now this quantity is multiplied by `option.ContractUnitOfTrade` (usually 100 for US options).

This bug was hidden by the option assignment tests in `SecurityPortfolioManagerTests`, which have also been corrected.
The regression stats for `OptionExerciseAssignRegressionAlgorithm` have also been updated.

In some cases this bug was also causing division by zero exceptions in `OnOrderEvent`, which seem to be fixed with this PR.

### DIFF
--- a/Common/Orders/OptionExercise/DefaultExerciseModel.cs
+++ b/Common/Orders/OptionExercise/DefaultExerciseModel.cs
@@ -14,12 +14,7 @@
  *
 */
 
-using System;
 using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


jaredbroad:
As discussed; looks like the contract multiplier might need more thought / review.

PR #1098, opened 2017-08-29
Adds Crypto security type

jameschch:
This supplies the minimal changes required to create a Crypto security type with the behavior parallel to that of Forex. Very little refactoring of existing Forex code was required, and most of the changes are of the form:

```
if (securityType == Forex || securityType == Crypto)
{
   ExistingForexBehaviour();
}
```

This diff can be used as a basis to decide at which points the behavior of Forex and Crypto should diverge. Given that this is at a pre-specified stage, only a small amount of additional test coverage has been added.

Please feel free to share any feature ideas and to clarify with additional commits as necessary.


Common/Orders/OrderJsonConverter.cs
```diff
@@ -106,8 +106,12 @@ public static Order CreateOrderFromJObject(JObject jObject)
             order.ContingentId = jObject["ContingentId"].Value<int>();
 
             var market = Market.USA;
+
+            //todo: can't safely derive market from security type
             if (securityType == SecurityType.Forex) market = Market.FXCM;
 
+            if (securityType == SecurityType.Crypto) market = Market.Bitfinex;
```
Martin-Molinero: 'else-if'. Maybe :108 could be the 'else'

Common/Data/Custom/FxcmVolume.cs
```diff
@@ -0,0 +1,130 @@
...
+    ///     - EURUSD, USDJPY, GBPUSD, USDCHF, EURCHF, AUDUSD, USDCAD,
+    ///       NZDUSD, EURGBP, EURJPY, GBPJPY, EURAUD, EURCAD, AUDJPY
+    /// FXCM only provides support for FX symbols which produced over 110 million average daily volume (ADV) during 2013.
+    /// This limit is imposed to ensure we do not highlight low volume/low ticket symbols in addition to other financial reporting concerns.
+    ///
+    /// </summary>
+    /// <seealso cref="QuantConnect.Data.BaseData" />
+    public class FxcmVolume : BaseData
+    {
+        private string[] _availableSymbols =
```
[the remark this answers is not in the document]
jameschch: Sorry this was merge of new features from master. I should have merged master at the end of the review to make things clearer.

Common/Orders/OrderJsonConverter.cs
jameschch: I'm glad you highlighted this. The class looks "legacy" to me and I hope another dev can confirm if this is obsolete. The practice of deriving a market from a security type doesn't make sense to me.

jaredbroad: The orders are pretty tricky to decode from JSON so we keep this class around to help people opening JSON results from the GUI. I agree deriving market from sectype is incorrect; we should be able to get that from the symbol object. 

If possible can we remove references to Bitfinex as it not integrated yet and can add it when the brokerage implementation is ready? Anywhere you're forced into having a market like that perhaps it can start with GDAX

jameschch: Yes, I was thinking the market should come from symbol, however, you're not guaranteed to have market, which leaves us back to guessing based on a default, throwing an exception, or leaving it null and hoping nothing blows up down the line. I'm not very familiar of the context in which this is used. Is it just a serialization utility?

jameschch: I think I've removed most (all?) Bitfinex references in the gdax branch, but there's no reason it can't be done here.

Algorithm/QCAlgorithm.cs
```diff
@@ -944,6 +945,7 @@ public void SetBrokerageMessageHandler(IBrokerageMessageHandler handler)
         public void SetBenchmark(SecurityType securityType, string symbol)
         {
             var market = securityType == SecurityType.Forex ? Market.FXCM : Market.USA;
+            market = securityType == SecurityType.Crypto ? Market.Bitfinex : market;
```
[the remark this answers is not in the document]
jameschch: I feel like manually setting the benchmark is sufficiently advanced that anyone doing it should be supplying all the relevant information, including market. I believe in the short term this should be marked deprecated and the code changed to use:
`
BrokerageModel.DefaultMarkets.TryGetValue(securityType, out market) `

Going further from this, the logic should derive the suggested benchmark given the available history providers to ensure the Alpha statistic is consistent in backtesting and live.

Common/Data/SubscriptionManager.cs
```diff
@@ -206,7 +208,8 @@ public void AddConsolidator(Symbol symbol, IDataConsolidator consolidator)
                 {SecurityType.Option, new List<TickType>() { TickType.Quote, TickType.Trade, TickType.OpenInterest } },
                 {SecurityType.Cfd, new List<TickType>() { TickType.Quote } },
                 {SecurityType.Future, new List<TickType>() { TickType.Quote, TickType.Trade, TickType.OpenInterest } },
-                {SecurityType.Commodity, new List<TickType>() { TickType.Trade } }
+                {SecurityType.Commodity, new List<TickType>() { TickType.Trade } },
+                {SecurityType.Crypto, new List<TickType>() { TickType.Quote } }
```
StefanoRaggi: I think we should also add `TickType.Trade` here so we can have both quotes and trades. 
The Gdax PR #1090 will also require changes to emit both types of ticks.

jameschch: I see what you're getting at here. In the gdax brokerage I'm using the order stream to construct best bid/ask and emit a Quote tick, but you're saying Lean will do this for me if i emit TickType.Trade?

StefanoRaggi: The brokerage implementation should emit best bid/ask ticks as TickType.Quote and all trades ("match" message in GDAX) as TickType.Trade. 
Lean will receive these ticks and pass them to the algorithm as Tick objects (if Resolution.Tick) or consolidate quotes in QuoteBars and trades in TradeBars (for other resolutions).

Common/Data/SubscriptionManager.cs
```diff
@@ -209,7 +209,8 @@ public void AddConsolidator(Symbol symbol, IDataConsolidator consolidator)
                 {SecurityType.Cfd, new List<TickType>() { TickType.Quote } },
                 {SecurityType.Future, new List<TickType>() { TickType.Quote, TickType.Trade, TickType.OpenInterest } },
                 {SecurityType.Commodity, new List<TickType>() { TickType.Trade } },
-                {SecurityType.Crypto, new List<TickType>() { TickType.Quote } }
+                {SecurityType.Crypto, new List<TickType>() { TickType.Quote } },
+                {SecurityType.Crypto, new List<TickType>() { TickType.Trade } }
```
StefanoRaggi: Should be: 
`{SecurityType.Crypto, new List<TickType>() { TickType.Quote, TickType.Trade } }`

jameschch: Yes, so this is the first step of a fairly significant global change and goes beyond anything I had planned.  How are you imagining the rest of this playing out?

StefanoRaggi: I'm assuming the global change you mention is inside the GDAX brokerage implementation, not in LEAN engine ?
Multiple tick types for a security type (and their subscriptions) are already supported inside LEAN (currently enabled only for Futures and Options, only brokerage implementation that has them is IB). 
Since crypto brokerages like GDAX make both quotes and trades available, I think it would be nice to have them both included in the initial version, the full OrderBook should be considered a new feature to be added later on (requiring new data types in LEAN).

jameschch: That's great. You were talking about introducing multiple tick types when we floated the bitfinex brokerage. I didn't realize it had been done since then. You guys have been busy.

PR #1100, opened 2017-08-30
Fixes default Equity leverage for Cash accounts

dannypurcell:
When trading in a cash only account, the maximum leverage for the account should be 1.0. This change is a step in the right direction but we still need a good way to cap leverage for the whole account such that multiple positions can not total up to be more than 1.0.
I'm new to this code base so if anyone knows a good place where such a rule may be set, please let me know.
I also do not know anything about Forex and Cfd trading, if those need to consider a "cash only" type account please let me know and I'll add that in.

### DIFF
--- a/Common/Brokerages/DefaultBrokerageModel.cs
+++ b/Common/Brokerages/DefaultBrokerageModel.cs
@@ -146,13 +146,14 @@ namespace QuantConnect.Brokerages
         /// <returns>The leverage for the specified security</returns>
         public decimal GetLeverage(Security security)
         {
+            if (AccountType == AccountType.Cash)
+            {
+                return 1m;
+            }
+                    
             switch (security.Type)
             {
                 case SecurityType.Equity:
-                    if (AccountType == AccountType.Cash)
-                    {
-                        return 1m;
-                    }
                     return 2m;
 
                 case SecurityType.Forex:
--- a/Common/Brokerages/DefaultBrokerageModel.cs
+++ b/Common/Brokerages/DefaultBrokerageModel.cs
@@ -149,6 +149,10 @@ namespace QuantConnect.Brokerages
             switch (security.Type)
             {
                 case SecurityType.Equity:
+                    if (AccountType == AccountType.Cash)
+                    {
+                        return 1m;
+                    }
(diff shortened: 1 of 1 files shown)


Common/Brokerages/DefaultBrokerageModel.cs
```diff
@@ -149,6 +149,10 @@ public decimal GetLeverage(Security security)
             switch (security.Type)
             {
                 case SecurityType.Equity:
+                    if (AccountType == AccountType.Cash)
```
mchandschuh: Would the leverage==1 apply to more than just equity?

StefanoRaggi: Probably `SecurityType.Crypto`, when we merge this one: https://github.com/QuantConnect/Lean/pull/1098 ?

jameschch: Yes, most crypto exchanges offer cash and margin accounts. In some cases cash is the only option due to regulatory constraints.

mchandschuh: I guess my question really is... if the user selects cash account, then shouldn't it apply that to all security types instead of just equity?

jaredbroad: @mchandschuh agree. It is more of a brokerage level property. 

@dannypurcell thank you for the PR; & welcome to LEAN! -- Please move this to the top of the method & we can merge it in.

dannypurcell: Thanks! 
Done
Forex trading requires margin right? Do we already block Forex accounts from setting account type to Cash?

jaredbroad: Its conceivable people would trade forex with a cash account as well. Is there something else I'm missing specific to fx?

dannypurcell: More likely there is something I am missing. 
I only looked at Forex trading once, a long time ago. The brokerage I was using at the time only offered Forex to margin accounts. That does not seem to be the case for Oanda or IB.

Issue #1069, opened 2017-08-16
Add Indicator Mass Index


LiXiang618:
Mass Index (Normally N1 = 25, N2 = 9)

Developed by Donald Dorsey, the Mass Index uses the high-low range to identify trend reversals based on range expansions. In this sense, the Mass Index is a volatility indicator that does not have a directional bias. Instead, the Mass Index identifies range bulges that can foreshadow a reversal of the current trend.

Formula:
Single EMA = N2-period exponential moving average (EMA) of the high-low differential  
Double EMA = N2-period EMA of the N2-period EMA of the high-low differential (i.e. N2-period EMA of Single EMA)
EMA Ratio = Single EMA divided by Double EMA 
Mass Index = N1-period sum of the EMA Ratio 

Note:
N1 in the sample data is 25.
N2 in the sample data is supposed to be 9, but needs verification.

Sample Data:
[spy_mass_index_25.txt](https://github.com/QuantConnect/Lean/files/1229400/spy_mass_index_25.txt)

Reference:
http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:mass_index

Jay-Jay-D:
I take it

Jay-Jay-D:
This issue should be closed.

PR #1109, opened 2017-09-01
Indicator mass index

Jay-Jay-D:
Mass indicator index implementation #1069


Indicators/MassIndex.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>
+        /// A new value for this indicator
+        /// </returns>
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            _highLow.Update(new IndicatorDataPoint
```
mchandschuh: Does this need to be an indicator?

Jay-Jay-D: No, is a remnant of a previous version. I'll fix it now.

The first attempt I made was [this](https://github.com/Jay-Jay-D/LeanSTP/blob/df74ac295c2f0400bdb8d6293307970675ba7874/Tests/Indicators/MassIndexIndicatorTests.cs), but fails the `ComparesAgainstExternalDataAfterReset` test.

Indicators/MassIndex.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <summary>
+    /// The Mass Index uses the high-low range to identify trend reversals based on range expansions.
+    /// In this sense, the Mass Index is a volatility indicator that does not have a directional
+    /// bias. Instead, the Mass Index identifies range bulges that can foreshadow a reversal of the
+    /// current trend. Developed by Donald Dorsey.
+    /// </summary>
+    /// <seealso cref="QuantConnect.Indicators.IndicatorBase{QuantConnect.Data.Market.TradeBar}"/>
+    public class MassIndex : IndicatorBase<TradeBar>
+    {
+        private readonly int _sumPeriod;
```
Martin-Molinero: doesn't Sum make the period available?

Jay-Jay-D: Actually yes, thanks!

PR #1111, opened 2017-09-02
Fixes FF daily data with different data/exchange tz

### LINKED ISSUE #1073: FillForwardEnumerator enumerating Oanda daily data incorrectly
For the algorithm below, the enumerator used for the Oanda EURUSD subscription is the `FillForwardEnumerator`.
```
    {
        private Symbol _symbol;
        private DateTime _previous;
        private SimpleMovingAverage _sma;

        {
            HistoryProvider = null;
            SetStartDate(2012, 7, 20);
            SetEndDate(2012, 7, 26);
            _symbol = AddForex("EURUSD", Resolution.Daily, Market.Oanda).Symbol;
            _sma = SMA(_symbol, 5, Resolution.Daily);
            SetBenchmark(date => 0);
        }

        {
            if (_previous.Date == Time.Date) return;

            Log(Time.Date.ToString("yyyyMMdd") + ": " + Securities[_symbol].Close + " " + _sma);
            _previous = Time;
        }
    }
```

In the algorithm above, the SMA indicator is never updated.The following table lists the first four bars emitted by the FillForwardEnumerator for the above algorithm:

| BaseData Emitted By FFE  | BaseData.Time | BaseData.EndTime |
| ------------- | ------------- | ------------- |
| 1 | Thursday 7/19/2012 8:00:00 PM  | Friday 7/20/2012 8:00:00 PM |
| 2  | Friday 7/20/2012 12:00:00 AM  | Saturday 7/21/2012 12:00:00 AM  |
| 3  | Saturday 7/21/2012 8:00:00 PM  | Sunday 7/22/2012 8:00:00 PM  |
| 4  | Sunday 7/22/2012 8:00:00 PM  | Monday 7/23/2012 8:00:00 PM |

The result of this `FillForwardEnumerator` is is that 8 hours of precision are lost when the second bar is emitted.

The following test can reproduce the issue above:

```
        [Test]
        public void OandaFillsForwardDailyForexOnWeekends()
        {
            var dailyBarsEmitted = 0;
            var fillForwardBars = new List<BaseData>();

            // 3 QuoteBars as they would be read from the EURUSD oanda daily file by QuoteBar.Reader()
            // The conversion from dataTimeZone to exchangeTimeZone has been done by hand
            var data = new BaseData[]
            {
                // fri 7/20
                new QuoteBar{Value = 0, Time =  new DateTime(2012, 7, 19, 20, 0, 0), Period = Time.OneDay},
                // sunday 7/22
                new QuoteBar{Value = 1, Time = new DateTime(2012, 7, 21, 20, 0, 0), Period = Time.OneDay},
                // monday 7/23
                new QuoteBar{Value = 2, Time = new DateTime(2012, 7, 22, 20, 0, 0), Period = Time.OneDay},
            }.ToList();
            var enumerator = data.GetEnumerator();

            var market = Market.Oanda;
            var symbol = Symbol.Create("EURUSD", SecurityType.Forex, market);

            var marketHours = MarketHoursDatabase.FromDataFolder();
            var dataTimeZone = marketHours.GetDataTimeZone(market, symbol, SecurityType.Forex);
            var exchange = new ForexExchange(MarketHoursDatabase.FromDataFolder().GetExchangeHours(market, symbol, SecurityType.Forex, dataTimeZone));

            var fillForwardEnumerator = new FillForwardEnumerator(enumerator, exchange, Ref.Create(TimeSpan.FromDays(1)), false, data.Last().EndTime, Time.OneDay);

            while (fillForwardEnumerator.MoveNext())
            {
                fillForwardBars.Add(fillForwardEnumerator.Current);
                Console.WriteLine(fillForwardEnumerator.Current.Time.DayOfWeek + " " + fillForwardEnumerator.Current.Time + " - " + fillForwardEnumerator.Current.EndTime.DayOfWeek + " " + fillForwardEnumerator.Current.EndTime);
                dailyBarsEmitted++;
            }

            Assert.AreEqual(dailyBarsEmitted, 3);
            Assert.AreEqual(fillForwardBars[0].Time, new DateTime(2012, 7, 19, 20, 0, 0));
            Assert.AreEqual(fillForwardBars[1].Time, new DateTime(2012, 7, 22, 20, 0 , 0));
            Assert.AreEqual(fillForwardBars[2].Time, new DateTime(2012, 7, 23, 20, 0, 0));
        }
```

The Assert statements in the above test list what I *think* the `BaseData` objects emitted from the `FillForwardEnumerator` should look like. I'm still a slightly hazy on what happens to the first hour the Oanda exchange is open on Sunday (7:00PM to 8:00PM EST).

mchandschuh:
The issue with having a different data and exchange time zone is that
daily data no longer follows a common assumption of beginning and ending
at midnight each day. This assumption was baked into the existing impl
to its own detriment. Specifically, the get next market open date calc
was being improperly executed. In most case (data resolution < one day)
we would use the start time of the bar, but since daily data spans a day,
we need to use the end time of the bar to properly calculate this value.

Fixes #1073

### DIFF
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -461,21 +461,13 @@ namespace QuantConnect.Lean.Engine
                 {
                     foreach (var update in timeSlice.ConsolidatorUpdateData)
                     {
-                        var resolutionTimeSpan = update.Target.Resolution.ToTimeSpan();
                         var consolidators = update.Target.Consolidators;
                         foreach (var consolidator in consolidators)
                         {
                             foreach (var dataPoint in update.Data)
                             {
-                                // Filter out data with resolution higher than the data subscription resolution.
-                                // This is needed to avoid feeding in higher resolution data, typically fill-forward bars.
-                                // It also prevents volume-based indicators or consolidators summing up volume to generate
-                                // invalid values.
-                                var algorithmTimeSpan = resolutionTimeSpan == TimeSpan.FromTicks(0)
-                                    ? TimeSpan.FromTicks(0)
-                                    : TimeSpan.FromSeconds(1);
-                                if (update.Target.Resolution == Resolution.Tick ||
-                                    algorithm.UtcTime.RoundDown(algorithmTimeSpan) == dataPoint.EndTime.RoundUp(resolutionTimeSpan).ConvertToUtc(update.Target.ExchangeTimeZone))
+                                // only push data into consolidators on the native, subscribed to resolution
+                                if (EndTimeIsInNativeResolution(update.Target, dataPoint.EndTime))
                                 {
                                     consolidator.Update(dataPoint);
                                 }
@@ -996,6 +988,20 @@ namespace QuantConnect.Lean.Engine
                 _algorithm.Status = AlgorithmStatus.RuntimeError;
                 Log.Error(err);
             }
+        }
+
+        /// <summary>
+        /// Determines if a data point is in it's native, configured resolution
+        /// </summary>
+        private static bool EndTimeIsInNativeResolution(SubscriptionDataConfig config, DateTime dataPointEndTime)
+        {
+            if (config.Increment == TimeSpan.Zero)
+            {
+                return true;
+            }
+
+            var roundedDataPointEndTime = dataPointEndTime.RoundDownInTimeZone(config.Increment, config.ExchangeTimeZone, config.DataTimeZone);
+            return dataPointEndTime == roundedDataPointEndTime;
         }
     }
 }
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -113,7 +113,7 @@ namespace QuantConnect.Lean.Engine
                 {
                     return "Algorithm took longer than 10 minutes on a single time loop.";
                 }
-                
+
                 return null;
             };
             _liveMode = liveMode;
@@ -133,7 +133,7 @@ namespace QuantConnect.Lean.Engine
         /// <param name="leanManager">ILeanManager implementation that is updated periodically with the IAlgorithm instance</param>
         /// <param name="token">Cancellation token</param>
         /// <remarks>Modify with caution</remarks>
-        public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, IDataFeed feed, ITransactionHandler transactions, IResultHandler results, IRealTimeHandler realtime, ILeanManager leanManager, CancellationToken token) 
+        public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, IDataFeed feed, ITransactionHandler transactions, IResultHandler results, IRealTimeHandler realtime, ILeanManager leanManager, CancellationToken token)
         {
             //Initialize:
             _dataPointCount = 0;
@@ -173,10 +173,10 @@ namespace QuantConnect.Lean.Engine
                 .FirstOrDefault(x => x.DeclaringType == algorithm.GetType()) != null;
 
             //Go through the subscription types and create invokers to trigger the event handlers for each custom type:
-            foreach (var config in algorithm.SubscriptionManager.Subscriptions) 
+            foreach (var config in algorithm.SubscriptionManager.Subscriptions)
             {
                 //If type is a custom feed, check for a dedicated event handler
-                if (config.IsCustomData) 
+                if (config.IsCustomData)
                 {
                     //Get the matching method for this event handler - e.g. public void OnData(Quandl data) { .. }
                     var genericMethod = (algorithm.GetType()).GetMethod("OnData", new[] { config.Type });
@@ -219,7 +219,7 @@ namespace QuantConnect.Lean.Engine
                     return;
                 }
 
-                // Update the ILeanManager 
+                // Update the ILeanManager
                 leanManager.Update();
 
                 var time = timeSlice.Time;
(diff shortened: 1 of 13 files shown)


mchandschuh:
@jaredbroad @andrewhart098 @StefanoRaggi 

Comments very welcome. This is super sensitive code. The commit message describes my
line of thinking as to why this happened and also the reason for the fix. In short, with daily
data being filled forward on daily increments, we need to choose a different reference
start time in order to properly calculate next market open.

Also, please check the test added as it is a modified form of #1073 because I think the
assertions outlined there were incorrect as they were missing the Sat-Sun bar, since the
market opens at 5PM `America/New_York` and the daily bar ends at 8PM same tz.

andrewhart098:
@mchandschuh - I think this PR fixes at least one bug in the `FillForwardEnumerator`. I have a hunch that there's still another bug somewhere in the FFE.  When running an algorithm using Oanda daily data, the consolidators are sporadically updated (if at all).

I'm struggling to place my finger on it exactly, but basically in the main AlgorithmManager loop, at the point where the AlgorithmManager updates the consolidators, [there's a conditional](https://github.com/QuantConnect/Lean/blob/eeb3e55696885e0db36cbfba45710182fb96b42b/Engine/AlgorithmManager.cs#L478) that ensures that data is only passed into the consolidator if it has the same Date as `Algorithm.UtcTime`. Even with this PR, virtually all of the data in `ConsolidatorUpdateData` is one day behind  `Algorithm.UtcTime`.

My hunch is that it's the FFE, but I'm still digging.  This PR does affect how the consolidators are updating. For now, I'm going to leave this PR open until I understand why the Oanda daily consolidators are not updating. It could be something else too! There's a lot of code between the AlgorithmManager and the FFE.

Thanks for looking into this. If you have any idea why the Oanda daily data consolidators are not updating I'd love to hear it.

mchandschuh:
yes, this is a similar bug as the other one addressed. The real issue is anytime we do 'daily' rounding on data with different data/exchange time zones. Daily data with different data/exchange time zones will be timestamped at, for example, 1600, so the daily rounding drastically changes the value.

Really what we're trying to do here is determine if it's a fill forwarded piece of data from a lower resolution, such as daily data fill forwarding every minute. In this case, we don't want to push the duplicate data through. I suppose the correct solution here would be to round in the data time zone.

I have some code working locally that performs the rounding in the data time zone. I'm trying to find other cases where this might be an issue as well, should have this PR updated this afternoon.

mchandschuh:
fyi, first commit includes whitespace changes, easier to view diff commit by commit

andrewhart098:
Hey @mchandschuh - the `OandaFillsForwardDailyForexOnWeekends` test is failing with the latest commit. There are now 5 bars emitted from the FFE. The 5 bars emitted are:

| Bar.Time | Bar.EndTime |
| --- | --- |
{7/19/2012 8:00:00 PM} | {7/20/2012 8:00:00 PM}
{7/19/2012 4:00:00 AM} | {7/20/2012 4:00:00 AM} 
{7/20/2012 12:00:00 AM} | {7/21/2012 12:00:00 AM}
{7/21/2012 8:00:00 PM} | {7/22/2012 8:00:00 PM}
{7/22/2012 8:00:00 PM} | {7/23/2012 8:00:00 PM}

The second bar looks freaky - it's before the first. Stepping through the test, I notice that when the test first gets to `RequiresFillForwardData()`, the `nextFillForwardTime` is calculated correctly. However, when the `fillForward.Time` is calculated you use `RoundDownInTimeZone`. Correct me if I'm wrong, but I think it would be incorrect to calculate the `fillForward.Time` in the data tz because the fill forward frontier time is based on the exchange Tz not the data Tz, i.e.

```
// find the next fill forward time after the next market open
 var nextFillForwardTime = Exchange.Hours.GetNextMarketOpen(previous.EndTime, _isExtendedMarketHours) + fillForwardResolution;
```

In other words, I think we should swap `RoundDownInTimeZone` for `RoundDown` one line 277 of the FFE.

mchandschuh:
The rounding must be performed in the data time zone. That extension method
converts to data tz, performs rounding, and then converts it back.

There is definitely something wrong with the existing impl, which is what I
get for playing around with it too much. I've been beyond swamped this past
week, but hoping to take another look at it soon.

If you have time feel free to modify this PR. My thought process here is
that rounding must be performed in the data time zone. Consider daily
rounding, which will always end up at midnight. But then consider data time
zone != exchange time zone. Daily rounding should be rounded to midnight in
the data time zone, not the exchange time zone.

I did a little analysis and also identify consolidators as needing some
attention as they perform a similar check. Basically the logical check that
is usually being performed is to check if a piece of data is on the
requested resolution of not. For example, daily data will fill forward on
minutes, but you don't want daily bars pumped into consolidators every
minute.

Feel free to call and chat and I how to get back to this issue soon.

Engine/DataFeeds/Enumerators/LiveFillForwardEnumerator.cs
```diff
@@ -43,8 +44,8 @@ public class LiveFillForwardEnumerator : FillForwardEnumerator
         /// <param name="isExtendedMarketHours">True to use the exchange's extended market hours, false to use the regular market hours</param>
         /// <param name="subscriptionEndTime">The end time of the subscrition, once passing this date the enumerator will stop</param>
         /// <param name="dataResolution">The source enumerator's data resolution</param>
```
Martin-Molinero: missing the new "dataTimeZone" param

mchandschuh:
@andrewhart098 - noticed `CustomDataRegressionAlgorithm` is failing when I run on master, is this normal?

mchandschuh:
Newest implementation lazily evaluates the bar end times and pre sorts based on fill forward/data resolution values

Engine/AlgorithmManager.cs
```diff
@@ -997,5 +989,19 @@ private void SampleBenchmark(IAlgorithm algorithm, IResultHandler results, DateT
...
+        /// Determines if a data point is in it's native, configured resolution
+        /// </summary>
+        private static bool EndTimeIsInNativeResolution(SubscriptionDataConfig config, DateTime dataPointEndTime)
+        {
+            if (config.Increment == TimeSpan.Zero)
+            {
+                return true;
+            }
+
+            var roundedDataPointEndTime = dataPointEndTime.RoundDownInTimeZone(config.Increment, config.ExchangeTimeZone, config.DataTimeZone);
```
Martin-Molinero: I'm not 100% sure how this works but: During an execution, if this is true once, wouldn't it be true always for that data source? or can it change?
If you can trust it will return always the same, maybe there could be some type of cache/memory of previous checks, to avoid calling RoundDownInTimeZone for each data.

mchandschuh:
@jaredbroad - still need to get some benchmark numbers on old vs new -- was able to run through the daily oanda case and the data is getting pumped through as expected -- not a huge fan of the rounding we do in `SubscriptionSynchronizer` -- basically the FFE can emit something on say an hour 0900-100 and if it's daily data, it would round that down to the previous day before being pumped into the algo

mchandschuh:
@jaredbroad - this is about as good as it gets without unrolling the enumerable into a giant if/else block.

This implementation is ~40% slower than the existing, but is still only about a second slower over 2.7 million fill forward bars.

#### Existing
```
Elapsed (s): 2.48
DataPoints : 2765521
```

#### New
```
Elapsed (s): 3.43
DataPoints : 2765521
```


I think this is ready for a final review.

Engine/AlgorithmManager.cs
mchandschuh: the date time value here will be different for every data point so we can't cache the value

PR #1125, opened 2017-09-11
Enable security seeding for Options and Futures

StefanoRaggi:


Algorithm/QCAlgorithm.History.cs
```diff
@@ -444,13 +444,20 @@ public BaseData GetLastKnownPrice(Security security)
             var startTime = GetStartTimeAlgoTzForSecurity(security, 1, resolution);
             var endTime   = Time.RoundDown(resolution.ToTimeSpan());
 
+            // request QuoteBar for Options and Futures
+            var dataType = typeof(BaseData);
+            if (security.Type == SecurityType.Option || security.Type == SecurityType.Option)
```
jaredbroad: :+1:

StefanoRaggi: Oops, copy/paste bug on that line :)

Algorithm/QCAlgorithm.History.cs
```diff
@@ -444,13 +444,20 @@ public BaseData GetLastKnownPrice(Security security)
             var startTime = GetStartTimeAlgoTzForSecurity(security, 1, resolution);
             var endTime   = Time.RoundDown(resolution.ToTimeSpan());
```
jaredbroad: On Sunday night; the Start was 3.59 Friday -- and Sunday 8pm is more than +2 days from that start time.
This meant the history request is more than 2 days -- the maximum period for IB history API.
We hacked around it with StartTime.AddHours(-1), EndTime = StartTime.AddHours(1) -- and then taking the "last()" item on the history request result.

Issue #1079, opened 2017-08-21
System.AccessViolationException referencing self.UtcTime in .py


se-l:
When referencing **self.UtcTime** in .OnData, I always receive a

_**System.AccessViolationException**: 'Attempted to read or write protected memory. This is often an indication that other memory is corrupt.'_

like in this issue @AlexCatarino  : #950 
I cloned the master branch repos last week. Is there a chance I missed the fix?

Below a forex variation of BasicTemplateAlgorithm.py, that reproduces the error for me. Tried reproducing it in the browser IDE, but would get stuck 'waiting for results' / 'analyzing' not returning results. Works fine with self.Time

Happy to provide more details. First time use of .NET / VS for quantconnect.

`from clr import AddReference
AddReference("System")
AddReference("QuantConnect.Algorithm")
AddReference("QuantConnect.Common")

from System import *
from QuantConnect import *
from QuantConnect.Algorithm import *
import decimal as d

class BasicTemplateAlgorithm(QCAlgorithm):
    '''Basic template algorithm simply initializes the date range and cash'''

    def Initialize(self):
        '''Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.'''
        
        self.SetStartDate(2017,1,10)  #Set Start Date
        self.SetEndDate(2017,1,17)    #Set End Date
        self.SetCash(100000)           #Set Strategy Cash
        # Find more symbols here: http://quantconnect.com/data
        self.AddForex("EURUSD", Resolution.Minute)

    def OnData(self, data):
        '''OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
        
        Arguments:
            data: Slice object keyed by symbol containing the stock data
        '''
        newTicket = self.MarketOrder("EURUSD", 1000 )
        #print newTicket.Time#self.Transactions.GetOrderTicket(self.Transactions.LastOrderId).OrderEvents[0].UtcTime
        print self.UtcTime
        #print self.Time`

Goal is to subtract current time and order Time.
For now, I would continue using self.Time with a workaround. Since the order tz and self.Time tz have an offset of 4h, I thought it's best to use Utc. Will adjust time (+4) or adjust time zones until I figure out what's wrong. Appreciating any hints....

Full error:
[...]

Issue #1031, opened 2017-07-24
System.AccessViolationException from Python Runtime


tomhunter-gh:
Hi @AlexCatarino, thanks for all the hard work with the Python integration.

Unfortunately, I'm getting a System.AccessViolationException when running a Python algorithm.  I have created the following minimal test case:

```python
from clr import AddReference
AddReference("System")
AddReference("QuantConnect.Algorithm")
AddReference("QuantConnect.Common")

from System import *
from QuantConnect import *
from QuantConnect.Algorithm import *


class BasicTemplateAlgorithm_AccessViolationExceptionTest(QCAlgorithm):

    def Initialize(self):
        
        self.SetStartDate(2013,10,07)
        self.SetEndDate(2013,10,11)
        self.SetCash(100000)
        
        self.AddEquity("SPY", Resolution.Second)

        self.orderTicket = None

    def OnData(self, data):
        
        if not self.Portfolio.Invested:            
            self.orderTicket = self.Order("SPY", 1)

        openTime = self.orderTicket.SubmitRequest.Time
        print openTime
```

This prints the `openTime` datetime a few hundred times and then crashes with:

```
Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Python.Runtime.Runtime.Py_CompileString(String code, String file, IntPtr tok)
   at Python.Runtime.PythonEngine.ModuleFromString(String name, String code)
   at Python.Runtime.Converter.TzInfo(Int32 hours, Int32 minutes)
   at Python.Runtime.Converter.ToPython(Object value, Type type)
   at Python.Runtime.PropertyObject.tp_descr_get(IntPtr ds, IntPtr ob, IntPtr tp)
   at Python.Runtime.Runtime.PyObject_Call(IntPtr pointer, IntPtr args, IntPtr kw)
   at Python.Runtime.PyObject.Invoke(PyTuple args, PyDict kw)
   at Python.Runtime.PyObject.InvokeMethod(String name, PyTuple args, PyDict kw)
   at Python.Runtime.PyObject.TryInvokeMember(InvokeMemberBinder binder, Object[] args, Object& result)
   at CallSite.Target(Closure , CallSite , Object , Slice )
[...]

PR #1127, opened 2017-09-12
Upgrades pythonnet version

AlexCatarino:
In this version, we have fixed a bug that led to System.AccessViolationException when we add timezone information in the datetime.datetime object.


andrewhart098:
Resolves #1031 
Resolves #1079

PR #1131, opened 2017-09-14
Crypto security

jaredbroad:
Implementation of a crypto security type by @jameschch


Common/Securities/Crypto/Crypto.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <param name="symbolProperties">The symbol properties for this security</param>
+        public Crypto(SecurityExchangeHours exchangeHours, Cash quoteCurrency, SubscriptionDataConfig config, SymbolProperties symbolProperties)
+            : base(config,
+                quoteCurrency,
+                symbolProperties,
+                new CryptoExchange(exchangeHours),
+                new ForexCache(),
+                new SecurityPortfolioModel(),
+                new ImmediateFillModel(),
+                new InteractiveBrokersFeeModel(),
```
jaredbroad: Perhaps constant fee model (0)? I thought most exchanges were spread based?

Common/Brokerages/DefaultBrokerageModel.cs
```diff
@@ -155,6 +156,9 @@ public decimal GetLeverage(Security security)
                 case SecurityType.Cfd:
                     return 50m;
 
+                case SecurityType.Crypto:
+                    return 3m;
```
jaredbroad: Perhaps default leverage should be 1 as most exchanges/brokerages don't support leverage yet? Later we can add custom leverage by brokerage model.

Common/Orders/OrderJsonConverter.cs
```diff
@@ -105,8 +106,19 @@ public static Order CreateOrderFromJObject(JObject jObject)
...
+            //does data have market?
+            var suppliedMarket = jObject.SelectTokens("Symbol.ID.Market");
+            if (suppliedMarket.Any())
+            {
+                market = suppliedMarket.Single().Value<string>();
+            }
+            else
+            {
+                //no data, use default                
+                new DefaultBrokerageModel().DefaultMarkets.TryGetValue(securityType, out market);
```
jaredbroad: This is nice.

Common/Securities/Cash.cs
```diff
@@ -179,7 +180,7 @@ public Security EnsureCurrencyDataFeed(SecurityManager securities, SubscriptionM
             var currencyPairs = Currencies.CurrencyPairs.Select(x =>
             {
                 // allow XAU or XAG to be used as quote currencies, but pairs including them are CFDs
-                var securityType = Symbol.StartsWith("X") ? SecurityType.Cfd : SecurityType.Forex;
+                var securityType = Symbol.StartsWith("XAU") ? SecurityType.Cfd : SecurityType.Forex;
```
jaredbroad: I think this drops XAG as quote currency, might need second If statement.

Common/Brokerages/DefaultBrokerageModel.cs
jameschch: It's only really the US regulated exchanges that cannot offer leverage. But I see your point, we should not treat gdax behaviour as the default for crypto.

Common/Securities/Crypto/Crypto.cs
jameschch: The gdax model is fee-free for makers after rebates. The other exchanges I have in mind, OKCoin, Bitfinex and Kraken have a maker/taker split at around 0.01%/0.02%. The Chinese exchanges used to offer fee free until the PBOC "visits" earlier this year. It makes sense to provide a constant fee model with the expectation that it needs overriding as anyone writing an algorithm for crypto that does not specify a valid fee model will get uselessly biased results.

Common/Securities/Cash.cs
jameschch: Fixed

jameschch:
Changes are here:

[https://github.com/QuantConnect/Lean/pull/1133](https://github.com/QuantConnect/Lean/pull/1133)

PR #1137, opened 2017-09-14
Added 6 python algorithms

jingwu74:


Algorithm.Python/EmaCrossUniverseSelectionAlgorithm.py
```diff
@@ -0,0 +1,80 @@
+﻿from System.Collections.Generic import List
```
jaredbroad: Please add the apache license header

Algorithm.Python/BasicTemplateForexAlgorithm.py
```diff
@@ -0,0 +1,57 @@
...
+        history = self.History(TimeSpan.FromSeconds(5), Resolution.Second)
+         
+        for data in sorted(history, key=lambda x: x.Time):
+            for key in data.Keys:
+                self.Log(str(key.Value) + ": " + str(data.Time) + " > " + str(data[key].Value))
+        
+    def OnData(self, data):
+        # Print to console to verify that data is coming in
+        for key in data.Keys:
+            self.Log(str(key.Value) + ": " + str(data.Time) + " > " + str(i[key].Value))
```
Martin-Molinero: is there a typo in the last str?

Algorithm.Python/BasicTemplateForexAlgorithm.py
```diff
@@ -0,0 +1,57 @@
...
+        self.AddForex("GBPUSD", Resolution.Minute)
+        self.AddForex("EURGBP", Resolution.Minute)
+        
+        self.History(5, Resolution.Daily)
+        self.History(5, Resolution.Hour)
+        self.History(5, Resolution.Minute)
+
+        history = self.History(TimeSpan.FromSeconds(5), Resolution.Second)
+         
+        for data in sorted(history, key=lambda x: x.Time):
```
Martin-Molinero: data is not sorted in history?

Algorithm.Python/EmaCrossUniverseSelectionAlgorithm.py
```diff
@@ -0,0 +1,104 @@
...
+        # Filter the values of the dict: we only want up-trending securities 
+        values = filter(lambda x: x.is_uptrend, self.averages.values())
+        
+        # Sorts the values of the dict: we want those with greater difference between the moving averages
+        values.sort(key=lambda x: x.scale, reverse=True)
+
+        # we need to return only the symbol objects
+        list = List[Symbol]()
+        for x in values[:self.coarse_count]:
+            print x.symbol, x.scale
```
Martin-Molinero: Hi ! I see you generally use self.Log() & here you used print, should it be self.Log() too?

Algorithm.Python/BasicTemplateForexAlgorithm.py
jingwu74: It is sorted in history

Algorithm.Python/EmaCrossUniverseSelectionAlgorithm.py
jingwu74: Yes, it should be Log

PR #1151, opened 2017-09-21
ReadZipEntry fix

Jay-Jay-D:
The issue is that after a file using the `Quantconnect.Compression.ReadLines` static method, the file can be modified or deleted.

More in detail, just by using the Quantconnect.Compression.ReadLines, the file can be accessed, but the error arises then one tries to _interact_ with the results. E.g.:
```cs
var lines = QuantConnect.Compression.ReadLines(zipFilePath);
var linesN =lines.Count();
```
Even when the underlying methods disposing the zip handlers correctly, the problem is the use of the `yield` in the `Quantconnect.Compression.ReadZipEntry`.

This PR contains the tests to prove the issue and a fix to the problem.

### DIFF
--- a/Compression/Compression.cs
+++ b/Compression/Compression.cs
@@ -611,13 +611,15 @@ namespace QuantConnect
 
         private static IEnumerable<string> ReadZipEntry(Ionic.Zip.ZipEntry entry)
         {
+            var output = new List<string>();
             using (var entryReader = new StreamReader(entry.OpenReader()))
             {
                 while (!entryReader.EndOfStream)
                 {
-                    yield return entryReader.ReadLine();
+                    output.Add(entryReader.ReadLine());
                 }
             }
+            return output;
         }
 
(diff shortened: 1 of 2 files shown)


jaredbroad:
Hi JJ -- what is the *bug* you see from trying to interact with the data? Why would someone delete data library files?

We can't do a "copy-to-cache" style system at this level as it would explode the RAM usage in a universe example where you're opening 1,000 files. LEAN is built to stream all the way up to keep RAM low e even in tough universe cases.

Jay-Jay-D:
> ... modified or deleted.
Ok, forget about delete. You cannot update a zip file after reading it e.g. daily and hourly data.

But I see your point, and now understand the implications of this change.

jaredbroad:
OK understood; for backtesting we assume it'll take less than 24hrs per backtest (or at least the data is steady state)

PR #1153, opened 2017-09-21
Optimize conversion of Python lists to CLR generic Lists

young-zhang:
Since we're copying from a Python list to a CLR List, and we already know the size of the Python list, we should pre-allocate the CLR list with the size.


jaredbroad:
Nice! Thank you @young-zhang 

The IDE docs are now driven by this repo -- changes will be quickly reflected in the IDE :)

PR #1161, opened 2017-09-25
Fixes to the VS2017 Plugin

Sarel-Esterhuizen:
- Few visual tweaks to the login form, the user and access token fields were too low making the form look weird
- There is a bug with VS plugins where it does not include downstream dependencies in the VSIX, this means that the QuantConnect.Configuration dll for example will not make it into the package since it is an indirect reference via QuantConnect dll. Explicitly specifying on the references to BuiltProjectOutputGroupDependencies solves this bug.
- Fixed funny casing for the Solution Explorer menu so that it looks neater and more consistent with the VS UI.
- The plugin language inference does not work if you select a folder since it relies on extensions. Added code to determine if the user's context action is on a folder and changed the method to get all child code files from that point. Change also respects subfolders so that it maps neatly to the web ui.


jaredbroad:
Awesome! Thank you for the PR and welcome to LEAN @Sarel-Esterhuizen!

PR #1163, opened 2017-09-26
Improves Universe Selection for python

AlexCatarino:
In python algorithm using Universe Selection, the selector method should return a List<Symbol>. To make it more pythonic, we allow returning python list. The conversion is, then, performed in C# side.


jaredbroad:
💯 :boom:

PR #1176, opened 2017-10-02
Set backtesting env history provider

mchandschuh:
Fresh clone of repo gives this exception when running backtest

image


mchandschuh:
This same default value is enforced via code:

https://github.com/QuantConnect/Lean/blob/master/Engine/Engine.cs#L439

PR #1179, opened 2017-10-03
Bugfix/live btc

mchandschuh:
This example custom data implementation was setting both the `Time` and `EndTime`
properties of `BaseData`. The issue with this is that `BaseData` uses the same storage
for each property. So the `coin.Time = coin.EndTime - config.Increment` call was always
pushing the end time far enough in the past that we would try to fast forward.

A thought here is that perhaps we should never try to fast forward rest subscription
sources. The fast forward was implemented to support remote files which may have
data into the future that we don't want to yield until the correct time.


mchandschuh:
An even bigger issue was uncovered here in the `RefreshEnumerator`. The existing implementation
was constantly regenerating the enumerator stack on every `MoveNext` invocation. The intent
behind this enumerator is that it would only refresh after the enumerator had exhausted itself.

PR #1182, opened 2017-10-05
Minor string trimming optimization

atifaziz:
Using `String.Trim` is more efficient than trimming the start and then the end as you avoid a string allocation between trimming of the head and the tail.

This is something I came across while causally browsing the source code and thought I'd turn it into a PR.


jaredbroad:
Thanks @atifaziz ! :)

PR #1188, opened 2017-10-06
Feature/set algorithm name

mchandschuh:
Set the algorithm's `Name` property to something reasonable. When running backtest/live we'll use information from the job packet. When running using the console handler we'll just use the algorithm's type name.


Engine/Setup/BacktestingSetupHandler.cs
```diff
@@ -164,6 +164,7 @@ public bool Setup(IAlgorithm algorithm, IBrokerage brokerage, AlgorithmNodePacke
             if (algorithm == null)
             {
                 Errors.Add("Could not create instance of algorithm");
+                Errors.Add("Could not create instance of algorithm");
```
StefanoRaggi: Why is this line duplicated ? Hope it's not a bug of the clean whitespace plugin :)

mchandschuh: great catch!, missed that when diffing -- not sure, it might have been a fat finger ctrl+d (duplicate line)

mchandschuh: Actually, must have been, ctrl+s and ctrl+d are easy to fat finger :/

PR #1189, opened 2017-10-06
Add python algorithms

jingwu74:


jaredbroad:
Hurrah for regression tests!

PR #1193, opened 2017-10-09
FxcmVolumeDownloader new options

Jay-Jay-D:
This PR implements two new options to the `FxcmVolumeDownloader`:

- The option `all` will download all the available data
- The option `update` will search for the pairs and resolution already downloaded and will update only those pairs.

Given that the FXCM Real Volume API slices days of data at 21:00 UTC, the implementation has some redundancies in case the latest day of data is incomplete. E.g. the `update` option will try to update also the last day of data already downloaded.


Tests/ToolBox/ForexVolume/FxcmVolumeDownloaderTest.cs
```diff
@@ -137,7 +136,7 @@ public void RequestWithMoreThan10KMinuteObservationIsCorrectlySaved()
             Assert.AreEqual(expected: 27, actual: files.Length);
         }
 
-        [Ignore("Long test")]
+//        [Ignore("Long test")]
```
mchandschuh: are these intended to remain commented out? I suspect we want these to **not** run on the ci server

ToolBox/FxcmVolumeDownloader/FxcmVolumeAuxiliaryMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+                case Resolution.Daily:
+                case Resolution.Hour:
+                    var expectedFilePath = Path.Combine(folderPath,
+                                                        String.Format("{0}_volume.zip", symbol.Value.ToLower()));
+                    if (File.Exists(expectedFilePath))
+                    {
+                        var lastStrDate = ReadZipFileData(expectedFilePath).Last()     // last observation                       
+                                                                           .First()    // first string (date)
+                                                                           .Substring(startIndex: 0, length: 8);
+                        lastAvailableDate = DateTime.ParseExact(lastStrDate, "yyyyMMdd", CultureInfo.InvariantCulture);
```
mchandschuh: Please use `DateFormat.EightCharacter` instead of the literal `"yyyyMMdd"` -- also occurs below on line 83.

ToolBox/FxcmVolumeDownloader/Program.cs
```diff
@@ -5,12 +5,35 @@
...
+                if (args[0] == "all" || args[0] == "update")
+                {
+                    if (args[0] == "update")
+                    {
+                        isUpdate = true;
+                    }
+
+                    args = new[]
+                    {
+                        "EURUSD,USDJPY,GBPUSD,USDCHF,EURCHF,AUDUSD,USDCAD,NZDUSD,EURGBP,EURJPY,GBPJPY,EURAUD,EURCAD,AUDJPY",
```
mchandschuh: Is this all the symbols that fxcm provides? Just curious why this list is the way it is.

Tests/ToolBox/ForexVolume/FxcmVolumeDownloaderTest.cs
Jay-Jay-D: I forgot to uncomment it.

I forgot to uncomment it, the good news are they passed 👍

ToolBox/FxcmVolumeDownloader/FxcmVolumeAuxiliaryMethods.cs
Jay-Jay-D: TIL... `DateFormat.EightCharacter`

ToolBox/FxcmVolumeDownloader/Program.cs
Jay-Jay-D: Those are the pairs that FXCM provides Volume data. The list was given by mail with the API document.

PR #1194, opened 2017-10-09
add HourReverseSplitRegressionAlgorithm

jingwu74:


Algorithm.Python/HourReverseSplitRegressionAlgorithm.py
```diff
@@ -0,0 +1,42 @@
...
+AddReference("System")
+AddReference("QuantConnect.Algorithm")
+AddReference("QuantConnect.Indicators")
+AddReference("QuantConnect.Common")
+
+from System import *
+from QuantConnect import *
+from QuantConnect.Algorithm import *
+from QuantConnect.Indicators import *
+from datetime import datetime, timedelta
```
AlexCatarino: I think we could avoid adding unnecessary imports. In this case, we would remove:
`AddReference("QuantConnect.Indicators")`
`from QuantConnect.Indicators import *`
`from datetime import datetime, timedelta`

Algorithm.Python/HourReverseSplitRegressionAlgorithm.py
```diff
@@ -0,0 +1,42 @@
...
+class HourReverseSplitRegressionAlgorithm(QCAlgorithm):
+
+    def Initialize(self):
+        self.SetStartDate(2013, 11, 7)
+        self. SetEndDate(2013, 11, 8)
+        self.SetCash(100000)
+        self.symbol = self.AddEquity("VXX", Resolution.Hour).Symbol
+    
+    def OnData(self, tradeBars):
+        if tradeBars.HasData:
```
AlexCatarino: Since the tradeBars object is actuall a Slice object, we could use slice or data. Otherwise it can confuse users. Also, instead of HasData, we better use 
`if slice.Bars.Count == 0: return` 
Because we actually have data and the Buy order is being placed earlier than the regression test requires to.

PR #1195, opened 2017-10-10
Add support for Financial Advisors in IB Brokerage

StefanoRaggi:
This PR also adds a new `DefaultOrderProperties` member to `QCAlgorithm`, including a nested `OrderFinancialAdvisorProperties` class.

These properties can be modified by users during algorithm `Initialize` and are always copied into all new orders submitted with any method available (`Order`, `SetHoldings`, etc.).

The `Order` base class has also been expanded to include the `OrderProperties` nested class.

Usage examples can be found in the newly added C#/Python algorithm: `FinancialAdvisorDemoAlgorithm`.


jaredbroad:
Good first pass. Please include a python example algorithm for the feature as well

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -551,19 +555,31 @@ public override void Connect()
...
-                        Disconnect();
-
-                        if (attempt++ < maxAttempts)
+                        if (!DownloadFinancialAdvisorAccount(_account))
                         {
-                            Thread.Sleep(1000);
-                            continue;
+                            Disconnect();
+
+                            throw new TimeoutException("InteractiveBrokersBrokerage.Connect(): DownloadFinancialAdvisorConfiguration failed.");
```
mchandschuh: Do we want similar re-try logic here as below?

```c#
if (attempt++ < maxAttempts)
{
    Thread.Sleep(1000);
    continue;
 }
```

Common/Orders/OrderFinancialAdvisorProperties.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string Account { get; set; }
+
+        /// <summary>
+        /// The account group for the order
+        /// </summary>
+        public string Group { get; set; }
+
+        /// <summary>
+        /// The allocation method for the account group order
+        /// Supported allocation methods are: EqualQuantity, NetLiq, AvailableEquity, PctChange
```
mchandschuh: For now this is fine since we're only doing IB, but if/when we add support for other FA accounts we'll want to revisit this design and see how it can be generalized. I suspect different brokerages support different options which leads me to think that FA properties would be brokerage specific.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Fixed

PR #1199, opened 2017-10-11
Update handling of IB fills to have actual IB fees in order events

StefanoRaggi:
Previously, the IB brokerage was calling the `InteractiveBrokerageFeeModel` to calculate the fee for the full order quantity, ignoring the actual executions. This solution had a couple of issues:
- it didn't include fees for orders partially filled, then cancelled
- it couldn't calculate fees for orders sent to financial advisor groups/profiles

Now every `OrderEvent` fired (for total or partial fills) will only include the actual IB fee paid for each single execution.

Order fill handling logic has also been simplified by only processing fills in the IB `ExecutionDetails` and `CommissionReport` events, instead of `OrderStatusChanged` (which can fire multiple times or not at all in some cases, such as market orders executed immediately).


Brokerages/InteractiveBrokers/Client/CommissionReportEventArgs.cs
```diff
@@ -35,5 +35,13 @@ public CommissionReportEventArgs(CommissionReport commissionReport)
         {
             CommissionReport = commissionReport;
         }
+
+        /// <summary>
+        /// Returns a string that represents the current object.
+        /// </summary>
+        public override string ToString()
+        {
+            return $"ExecId: {CommissionReport.ExecId}, Commission: {CommissionReport.Commission}, Currency: {CommissionReport.Currency}";
```
mchandschuh: <3 -- these make debugging/logging so much easier.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1404,18 +1396,10 @@ private void HandleExecutionDetails(object sender, IB.ExecutionDetailsEventArgs
...
+                // so we can fire the order event for the fill (or partial fill) with the actual IB fees
 
-                // set order status based on remaining quantity
-                var status = remainingQuantity > 0 ? OrderStatus.PartiallyFilled : OrderStatus.Filled;
-
-                if (!IsFillAlreadyHandled(order, executionDetails.Execution.OrderId, remainingQuantity))
-                {
-                    HandleFill(order, currentQuantityFilled, totalQuantityFilled, remainingQuantity, price, status);
-                }
+                _orderExecutions[executionDetails.Execution.ExecId] = executionDetails.Execution;
```
mchandschuh: Are we guaranteed to receive the fill *before* the commission event? Any documentation that provides such guarantee? I'm thinking of receiving the commission event first and then this one and we'll never emit the fill (unless I've ready this incorrectly)

StefanoRaggi: This is the only documentation page regarding these two events (it doesn't mention specific order of events):
https://interactivebrokers.github.io/tws-api/executions_commissions.html#gsc.tab=0
While testing I have never seen a commission report received before the execution it refers to (by ExecId).

I agree it is safer not to assume the current ordering, I'll change the logic to support receiving the events in any order.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1422,7 +1442,8 @@ private void HandleCommissionReport(object sender, IB.CommissionReportEventArgs
                 Execution execution;
                 if (!_orderExecutions.TryGetValue(e.CommissionReport.ExecId, out execution))
                 {
-                    Log.Error("InteractiveBrokersBrokerage.HandleCommissionReport(): execution not found: " + e.CommissionReport.ExecId);
+                    // save execution in dictionary and wait for execution event
```
mchandschuh: nit - comment save 'commission' in dictionary

Issue #1068, opened 2017-08-16
Add Indicator Detrended Price Oscillator


LiXiang618:
Detrended Price Oscillator (In the sample N = 21)

The Detrended Price Oscillator (DPO) is an indicator designed to remove trend from price and make it easier to identify cycles.

Formula:
Price {N/2 + 1} periods ago less the N-period simple moving average.

Note:
N refers to the number of periods used to calculate the Detrended Price Oscillator. A 20-day DPO would use a 20-day SMA that is displaced by 11 periods {20/2 + 1 = 11}. This displacement shifts the 20-day SMA 11 days to the left, which actually puts it in the middle of the look-back period. The value of the 20-day SMA is then subtracted from the price in the middle of this look-back period. In short, DPO(20) equals price 11 days ago less the 20-day SMA. 

Sample Data:
[spy_detrended_price_oscillator_21.txt](https://github.com/QuantConnect/Lean/files/1229245/spy_detrended_price_oscillator_21.txt)

Reference:
http://www.stockcharts.com/school/doku.php?id=chart_school:technical_indicators:detrended_price_osci

Jay-Jay-D:
This issue should be closed.

PR #1200, opened 2017-10-11
DetrendedPriceOscillator indicator.

Jay-Jay-D:
+ small redundant code fix at CompositeIndicator.cs
#1068


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -313,6 +313,24 @@ public DoubleExponentialMovingAverage DEMA(Symbol symbol, int period, Resolution
         }
 
         /// <summary>
+        /// Creates a new <see cref="DetrendedPriceOscillator"/> indicator.
+        /// </summary>
+        /// <param name="symbol">The symbol whose DPO we want</param>
+        /// <param name="period">The period over which to compute the DPO</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to the Value property of BaseData (x => x.Value)</param>
+        /// <returns>A new registerd DetrendedPriceOscillator indicator for the requested symbol over the specified period</returns>
```
mchandschuh: typo, registerd -> registered

Honestly, this typo probably occurs many times in this file and has probably just been copied around since the inception of this file

Indicators/CompositeIndicator.cs
```diff
@@ -98,13 +98,8 @@ public CompositeIndicator(string name, IndicatorBase<T> left, IndicatorBase<T> r
...
         public CompositeIndicator(IndicatorBase<T> left, IndicatorBase<T> right, IndicatorComposer composer)
-            : base(string.Format("COMPOSE({0},{1})", left.Name, right.Name))
-        {
-            _composer = composer;
-            Left = left;
-            Right = right;
-            ConfigureEventHandlers();
-        }
+            : this(string.Format("COMPOSE({0},{1})", left.Name, right.Name), left, right, composer)
+        { }
```
mchandschuh: no action required -- I was always more partial to these being on separate lines.

mchandschuh: also, I like the removal of the duplication here, good stuff.

Indicators/DetrendedPriceOscillator.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>
+        ///     A new value for this indicator
+        /// </returns>
+        protected override decimal ComputeNextValue(IndicatorDataPoint input)
+        {
+            _priceLag.Update(input);
+            _sma.Update(input);
+            return IsReady ? _priceLag - _sma : decimal.Zero;
```
mchandschuh: I think we typically try to return estimated values before `IsReady == true` -- any reason why we can't just *always* return `_priceLag - _sma` -- no worry of div 0 here or anything similar. I'm picturing plotting this and I'll get zeros until it's ready and then a straight line jump to the value.

Algorithm/QCAlgorithm.Indicators.cs
Jay-Jay-D: No, just once, my typo.
Fixed.

Indicators/DetrendedPriceOscillator.cs
Jay-Jay-D: Suggestion implemented.

In any case, please note this is an oscillator that swings around zero, so the difference will not be significant.

PR #1201, opened 2017-10-11
Support for GoodTillDate with Interactive Brokers

patrickstar1:


Common/Orders/Order.cs
```diff
@@ -81,6 +81,12 @@ public abstract class Order
         public OrderDuration Duration { get; internal set; }
 
         /// <summary>
+        /// Good till date(time) - The date and time until the order will be active. The order
+		/// will be automatically canceled by the exchange after this date/time.
+        /// </summary>
+        public DateTime GoodTillDate { get; set; }
```
mchandschuh: Given the logic on line 1493 of `InteractiveBrokersBrokerage.cs` I tihnk this will need to be nullable -- `DateTime?` -- otherwise the null check there will *always* return true

mchandschuh: Upon further inspection, perhaps we should be using this field on `Order.cs`:

```csharp
        /// <summary>
        /// Order Expiry on a specific UTC time.
        /// </summary>
        public DateTime DurationValue;
```

patrickstar1: @mchandschuh 

Agreed. Sorry I am a Java developer and am still learning the differences between Java and C#. In Java there is no such thing as not-nullable objects. Only primitive values are not nullable.

We could use DurationValue. Only thing is, although DurationValue in here is a DateTime and not really a duration - conceptually duration and end time are different (e.g. a duration of 5 days vs. 2017-10-16 12:54:53.123)

Another thing is, the document says "UTC time". This is good but only if QC's servers' time is set to UTC. Can you confirm that?

Thanks for your review.

patrickstar1: Also, please consider that GoodTillDate is a known terminology in trading world. Duration is not.

StefanoRaggi: I believe the most used term for order duration/validity is **Time In Force** and common values for this attribute are: **GTC** (Good Till Cancelled), **GTD** (Good Till Date), **DAY** (valid to end of day).

Naming issues aside, I think for a **GTD** order the enum value (currently named **OrderDuration**) should also be set by the caller in addition to the expiry/validity date.

patrickstar1: @StefanoRaggi  - currently OrderDuration only has GTC (and Custom that is not used anywhere). Do you want to add GTD as well? I'm thinking this will mean that other broker classes should implement it once you add it to the enum but I don't have enough knowledge about how to handle it with those brokers.

Another option is to reject it in runtime if a broker doesn't support the function.

BTW, here is IB's documentation (sorry, I should have attached the document link at the beginning). They use GoodTillDate as the name of the attribute:

https://interactivebrokers.github.io/tws-api/classIBApi_1_1Order.html#ade4fdefe2079ea5826872b9f52e3444e

So here is what so far I understand from your comments:

1- Do not use null check as the condition but the value of OrderDuration
2- Add GTD to OrderDuration
3- Use DurationValue as the end time (can we rename it? the attribute was not used anyway so it shouldn't break anyone's code).
4- If a broker doesn't have implementation for GTD, throw an exception with a message saying that "GoodTillDate has not been implemented for {broker_name}".

Please feel free to modify the requirements.

StefanoRaggi: @patrickstar1  - I think the real question is if: 
(a) we only want to add GTD support for IB or 
(b) it should be added as standard feature in LEAN, including other brokerages and backtesting (which may be the hardest part). 
@mchandschuh, @jaredbroad, thoughts ?

1. Ok
2. OrderDuration.Custom is currently only used internally in the Oanda brokerage implementation and used as GTD, I think it should be renamed from Custom to GTD
3. Ok for using DurationValue as the end time (default value: DateTime.MaxValue), not sure if renaming is safe because order objects are also serialized and deserialized in the QC cloud
4. Ok, we'll have to update Oanda and FXCM brokerages to support GTD

patrickstar1: Very nice. I will wait for the final decisions then.

mchandschuh: RE: other brokerages -- we can update the brokerage models for the other brokerages to return false for `CanSubmitOrder` if we don't support certain order features.

mchandschuh: RE: utc time -- as much as possible we try to keep things in utc for simplicity. All orders are timestamped explicitly in UTC, independent of the server's configuration. I think the name duration was chosen for alignment with `OrderDuration` -- I'm historically not the best namer of things, so I'll let @jaredbroad chime in on that one -- if we want to rename it, we can come up with a shim to prevent old algorithms from breaking if they used it.

patrickstar1: @mchandschuh thanks for the comments. We can always mark it as deprecated and add a new attribute that mirrors the old/deprecated attribute. This way we will have backward compatibility and some few months later get rid of it.

jaredbroad: :D Its relatively easy as an individual @patrickstar1; but it is harder to migrate a community of 40k. We prefer to put in the design work up front. If you can wait 3-7 days we'll do that design work and get back to you.

patrickstar1: That'd be awesome. Thanks Jared. I totally agree with a well designed system rather than a patch even if it delays the change.

mchandschuh:
I think a few things need to happen to implement the desired behavior (assuming starting from `QuantConnect/Lean/master`)
 - [ ] Work on new branch from fresh `upstream/master`
 - [ ] Add `OrderDuration.GTD` Good til date
 - [ ] In IB Brokerage, check for `order.OrderDuration == OrderDuration.GTD` and use `order.DurationValue` as the expiry time

jaredbroad:
Closing for now @patrickstar1 but this is close! Please submit a new PR with Mike's suggestions above leveraging Stefano's work.

PR #1202, opened 2017-10-12
Refactor custom live data enumerator stack

mchandschuh:
>NOTE: Cloud integration testing completed.
- [x] Test bitcoin rest 1 second, 1 minute
- [x] Test bitcoin rest daily
- [x] Test dailyfx in cloud 1 second
- [x] Test quandl daily
- [x] Test TopGainers - refactored selection logic into proper universe
- [x] Test custom data universe, nyse top gainers

This change aims to provide some more guarantees around our live custom data stack.
The big improvement here (besides test and code organization) is that we're now being
smart about applying rate limit and fast forward behaviors. Enumerator stack refreshes
are limited manually while rest rate limits are enforced via `RefreshEnumerator`. We make
this choice based on the `SubscriptionTransportMedium` -- basically, in live it doesn't make
sense to fast forward a rest endpoint. Likewise it doesn't make sense to rate limit access to
a file that's already been downloaded, instead we want to rate limit actual communication
with remote servers which happens every call for rest subscriptions and on each enumerator
stack refresh from file subscriptions.

I added some fairly simple test cases to cover the big 3 cases being addressed here:
1. Normal rest service, data point per invocation
2. Collection rest service, multiple data points per invocation
3. Remote files with historical and/or future data

### DIFF
--- a/Engine/DataFeeds/Enumerators/FrontierAwareEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/FrontierAwareEnumerator.cs
@@ -1,11 +1,11 @@
 /*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -94,20 +94,9 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
 
             if (underlyingCurrent != null && underlyingCurrent.EndTime <= localFrontier)
             {
-                // only emit if new data available (skip custom data if same time/value)
-                if (_lastEmittedValue != null && 
-                    _lastEmittedValue.DataType == MarketDataType.Base && 
-                    _lastEmittedValue.EndTime == underlyingCurrent.EndTime &&
-                    _lastEmittedValue.Value == underlyingCurrent.Value)
-                {
-                    _current = null;
-                }
-                else
-                {
-                    _needsMoveNext = true;
-                    _current = underlyingCurrent;
-                    _lastEmittedValue = _current;
-                }
+                _needsMoveNext = true;
+                _current = underlyingCurrent;
+                _lastEmittedValue = _current;
             }
             else
             {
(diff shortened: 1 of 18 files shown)


jaredbroad:
hey @akoeplinger! We're seeing TravisCI build errors on previously successful builds. Looks like an expired key in the base image?

akoeplinger:
@jaredbroad please see https://github.com/travis-ci/travis-ci/issues/8594 :)

Engine/DataFeeds/Enumerators/FrontierAwareEnumerator.cs
```diff
@@ -94,20 +94,9 @@ public bool MoveNext()
 
             if (underlyingCurrent != null && underlyingCurrent.EndTime <= localFrontier)
             {
-                // only emit if new data available (skip custom data if same time/value)
-                if (_lastEmittedValue != null && 
-                    _lastEmittedValue.DataType == MarketDataType.Base && 
-                    _lastEmittedValue.EndTime == underlyingCurrent.EndTime &&
-                    _lastEmittedValue.Value == underlyingCurrent.Value)
-                {
```
StefanoRaggi: Awesome! Glad you managed to remove this hackish check :)

Tests/Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactoryTests.cs
```diff
@@ -0,0 +1,462 @@
...
+using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Interfaces;
+using QuantConnect.Lean.Engine.DataFeeds;
+using QuantConnect.Lean.Engine.DataFeeds.Enumerators.Factories;
+using QuantConnect.Securities;
+using QuantConnect.Securities.Equity;
+
+namespace QuantConnect.Tests.Engine.DataFeeds.Enumerators.Factories
+{
+    public class LiveCustomDataSubscriptionEnumeratorFactoryTests
```
StefanoRaggi: Nice tests!

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -272,12 +272,15 @@ public void Run()
...
+                        var packet = new DataFeedPacket(subscription.Security, config);
 
                         // dequeue data that is time stamped at or before this frontier
                         while (subscription.MoveNext() && subscription.Current != null)
                         {
-                            packet.Add(subscription.Current);
+                            var clone = subscription.Current.Clone(subscription.Current.IsFillForward);
+                            clone.Time = clone.Time.RoundDownInTimeZone(config.Increment, config.ExchangeTimeZone, config.DataTimeZone);
+                            packet.Add(clone);
                         }
```
StefanoRaggi: Just curious, is this time rounding an additional bug fix (impact on not only custom data) ?

mchandschuh: The time rounding here is unfortunate, I do agree, but mirrors the backtesting behavior. We depend on rounded times later in the pipe line, specifically when invoking consolidators. The end time is used to determine if a piece of data is in it's subscribed to, native resolution. A later iteration can communicate this information in another way, for now, this is in line with how backtesting does it.

PR #1207, opened 2017-10-12
Replaced BTC with Nifty in IndicatorSuiteAlgorithm

andrewhart098:


Algorithm.CSharp/IndicatorSuiteAlgorithm.cs
```diff
@@ -121,8 +121,8 @@ public override void Initialize()
         /// <summary>
         /// Custom data event handler:
         /// </summary>
-        /// <param name="data">Bitcoin - dictionary of TradeBarlike Bars of Bitcoin Data</param>
-        public void OnData(Bitcoin data)
+        /// <param name="data">Nifty - dictionary Bars of Nifty Data</param>
+        public void OnData(Nifty data)
```
jaredbroad: Not sure this solves the issue; in the IDE the other types aren't accessible; you'll have to add a simple type (maybe quandl instead?) 
Build Error: File: main.cs Line:60 Column:20 - The type or namespace name 'Nifty' could not be found (are you missing a using directive or an assembly reference?)

PR #1209, opened 2017-10-13
Refactors indicator Update(DateTime, decimal) method

AlexCatarino:
- Moves indicator Update(DateTime, decimal) method overload from the extension class to the base class.
With this change, python algorithms can access this overload since pythonnet doesn't support extension methods.
- Changes python algorithms to use Update(DateTime, decimal) method

### DIFF
--- a/Indicators/IndicatorBase.cs
+++ b/Indicators/IndicatorBase.cs
@@ -98,6 +98,25 @@ namespace QuantConnect.Indicators
         }
 
         /// <summary>
+        /// Updates the state of this indicator with the given value and returns true
+        /// if this indicator is ready, false otherwise
+        /// </summary>
+        /// <param name="time">The time associated with the value</param>
+        /// <param name="value">The value to use to update this indicator</param>
+        /// <returns>True if this indicator is ready, false otherwise</returns>
+        public bool Update(DateTime time, decimal value)
+        {
+            if (typeof(T) == typeof(IndicatorDataPoint))
+            {
+                return Update((T)Activator.CreateInstance(typeof(T), new object[] { time, value }));
+            }
+            else
+            {
+                throw new NotSupportedException(string.Format("{0} does not support Update(DateTime, decimal) method overload. Use Update({1}) instead.", GetType().Name, typeof(T).Name));
+            }
+        }
+
+        /// <summary>
         /// Resets this indicator to its initial state
         /// </summary>
         public virtual void Reset()
--- a/Indicators/IndicatorBase.cs
+++ b/Indicators/IndicatorBase.cs
@@ -108,7 +108,7 @@ namespace QuantConnect.Indicators
         {
             if (typeof(T) == typeof(IndicatorDataPoint))
             {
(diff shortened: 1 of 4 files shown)


Indicators/IndicatorBase.cs
```diff
@@ -98,6 +98,25 @@ public bool Update(T input)
...
+        /// if this indicator is ready, false otherwise
+        /// </summary>
+        /// <param name="time">The time associated with the value</param>
+        /// <param name="value">The value to use to update this indicator</param>
+        /// <returns>True if this indicator is ready, false otherwise</returns>
+        public bool Update(DateTime time, decimal value)
+        {
+            if (typeof(T) == typeof(IndicatorDataPoint))
+            {
+                return Update((T)Activator.CreateInstance(typeof(T), new object[] { time, value }));
```
mchandschuh: if `T` is `IndicatorDataPoint`, why not just use `new IndicatorDataPoint(time, value)` ?

Indicators/IndicatorExtensions.cs
```diff
@@ -27,19 +27,6 @@ namespace QuantConnect.Indicators
...
     {
         /// <summary>
-        /// Updates the state of this indicator with the given value and returns true
-        /// if this indicator is ready, false otherwise
-        /// </summary>
-        /// <param name="indicator">The indicator to be updated</param>
-        /// <param name="time">The time associated with the value</param>
-        /// <param name="value">The value to use to update this indicator</param>
-        /// <returns>True if this indicator is ready, false otherwise</returns>
-        public static bool Update(this IndicatorBase<IndicatorDataPoint> indicator, DateTime time, decimal value)
```
mchandschuh: This could break any algorithm that explicitly calls this via `IndicatorExtensons.Update(time,value)` -- the odds are low, but leaving it in shouldn't cause issues as the C# compiler will prefer instance methods over extension methods.

AlexCatarino: Ok. So I will undo this part and leave it as it was.

Indicators/IndicatorBase.cs
AlexCatarino: I tried that, but `Update` didn't accept, since it is looking for an instance of `T`.

Indicators/IndicatorBase.cs
```diff
@@ -98,6 +98,25 @@ public bool Update(T input)
...
+        /// if this indicator is ready, false otherwise
+        /// </summary>
+        /// <param name="time">The time associated with the value</param>
+        /// <param name="value">The value to use to update this indicator</param>
+        /// <returns>True if this indicator is ready, false otherwise</returns>
+        public bool Update(DateTime time, decimal value)
+        {
+            if (typeof(T) == typeof(IndicatorDataPoint))
+            {
+                return Update((T)(object)new IndicatorDataPoint(time, value));
```
mchandschuh: cool, the run time will thank you  for removing a reflection call :)

PR #1216, opened 2017-10-17
Crypto securities only have one default SecurityType - Quote

andrewhart098:
Gdax api returns bid/ask spreads and its most accurately modeled with quotes. In addition, the historical data contains bid/ask spreads.  Using Quotes instead of Trades is the more appropriate for modeling crypto in Lean. Updated the gdax BTCUSD daily data to reflect this change


Common/Data/SubscriptionManager.cs
```diff
@@ -209,7 +209,7 @@ public void AddConsolidator(Symbol symbol, IDataConsolidator consolidator)
                 {SecurityType.Cfd, new List<TickType>() { TickType.Quote } },
                 {SecurityType.Future, new List<TickType>() { TickType.Quote, TickType.Trade, TickType.OpenInterest } },
                 {SecurityType.Commodity, new List<TickType>() { TickType.Trade } },
-                {SecurityType.Crypto, new List<TickType>() { TickType.Quote, TickType.Trade } },
+                {SecurityType.Crypto, new List<TickType>() { TickType.Quote } },
```
StefanoRaggi: I believe we need both

mchandschuh: I think if we have access to the trades we should pipe them in by default. I'm thinking of two particular use cases.
1. Ability to use indicators that require volume
2. Ability to better model liquidity/partial fills

jaredbroad: Agree ideal to have both but backtesting doesn't have the trade data; so it means a stream of error messages for anyone doing backtesting....

PR #1217, opened 2017-10-17
 Adds support for Historical Requests of Custom Data for python

AlexCatarino:
- Adds support for DynamicData in PandasConverter
- Adds python version of HistoryAlgorithm
This algorithm has examples of how to make historical data requests of custom data.
Adds this algorithm to the regression tests.

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -543,6 +543,136 @@ namespace QuantConnect.Algorithm
         }
 
         /// <summary>
+        /// Gets the historical data for the specified symbols between the specified dates. The symbols must exist in the Securities collection.
+        /// </summary>
+        /// <param name="type">The data type of the symbols</param>
+        /// <param name="tickers">The symbols to retrieve historical data for</param>
+        /// <param name="start">The start time in the algorithm's time zone</param>
+        /// <param name="end">The end time in the algorithm's time zone</param>
+        /// <param name="resolution">The resolution to request</param>
+        /// <returns>pandas.DataFrame containing the requested historical data</returns>
+        public PyObject History(PyObject type, PyObject tickers, DateTime start, DateTime end, Resolution? resolution = null)
+        {
+            var symbols = GetSymbolsFromPyObject(tickers);
+            if (symbols == null) return null;
+
+            var requests = symbols.Select(x =>
+            {
+                var security = Securities[x];
+                var config = security.Subscriptions.OrderByDescending(s => s.Resolution)
+                        .FirstOrDefault(s => s.Type.BaseType == CreateType(type).BaseType);
+                if (config == null) return null;
+
+                return CreateHistoryRequest(security, config, start, end, resolution);
+            });
+
+            return _converter.GetDataFrame(History(requests.Where(x => x != null)).Memoize());
+        }
+
+        /// <summary>
+        /// Gets the historical data for the specified symbols. The exact number of bars will be returned for
+        /// each symbol. This may result in some data start earlier/later than others due to when various
+        /// exchanges are open. The symbols must exist in the Securities collection.
+        /// </summary>
+        /// <param name="type">The data type of the symbols</param>
+        /// <param name="tickers">The symbols to retrieve historical data for</param>
+        /// <param name="periods">The number of bars to request</param>
+        /// <param name="resolution">The resolution to request</param>
+        /// <returns>pandas.DataFrame containing the requested historical data</returns>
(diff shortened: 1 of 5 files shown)


Common/Python/PandasConverter.cs
```diff
@@ -115,31 +116,46 @@ public PyObject GetDataFrame<T>(IEnumerable<T> data)
...
                 if (typeof(T) == typeof(TradeBar))
                 {
-                    Func<IBaseDataBar, double> getVolume = x => { var bar = x as TradeBar; return (double)bar.Volume; };
-                    pyDict.SetItem("volume", _pandas.Series(data.Select(x => getVolume(x)).ToList(), index));
+                    var bars = data.Select(x => x as TradeBar);
+                    pyDict.SetItem("low", _pandas.Series(bars.Select(x => (double)x.Low).ToList(), index));
+                    pyDict.SetItem("open", _pandas.Series(bars.Select(x => (double)x.Open).ToList(), index));
+                    pyDict.SetItem("high", _pandas.Series(bars.Select(x => (double)x.High).ToList(), index));
+                    pyDict.SetItem("close", _pandas.Series(bars.Select(x => (double)x.Close).ToList(), index));
+                    pyDict.SetItem("volume", _pandas.Series(bars.Select(x => (double)x.Volume).ToList(), index));
```
mchandschuh: >NOTE: This is an existing issue with the `PandasConverter` that perhaps should be looked at separately.

Each one of these `bars.Select` statement causes an extra enumeration of the `bars` enumerable. To make matters worse, this is all invoked 4*N times, where N is the number of data points.

I'm wondering if we need to invoke `GetDataFrame<T>` multiple times, or perhaps we just invoke it once and check the type of each item.

Ideally this would be implemented in such a way that we only cause one iteration of the enumerable. During this single iteration we would be building up all the various lists required to instantiate the pandas data frame.

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -566,6 +566,136 @@ public PyObject History(PyObject tickers, DateTime start, DateTime end, Resoluti
...
+            var requests = symbols.Select(x =>
+            {
+                var security = Securities[x];
+                var config = security.Subscriptions.OrderByDescending(s => s.Resolution)
+                        .FirstOrDefault(s => s.Type.BaseType == CreateType(type).BaseType);
+                if (config == null) return null;
+                
+                Resolution? res = resolution ?? security.Resolution;
+                var start = GetStartTimeAlgoTz(x, periods, resolution).ConvertToUtc(TimeZone);
+                return CreateHistoryRequest(security, config, start, UtcTime.RoundDown(res.Value.ToTimeSpan()), resolution);
```
mchandschuh: It looks like many of these methods are copy/paste from the `QCAlgorithm.History.cs` file -- perhaps instead of copy/paste we can just invoke the appropriate function?

AlexCatarino: I couldn't find a better way to do it. In these methods, the type is a variable, so how could we invoke the appropriate method that requires the type?

Common/Python/PandasConverter.cs
AlexCatarino: Agree. We can talk how would be the best way to do it.
I have made an experiment, where I would rewrite the `IEnumerable<Slice>` as a `Dictionary<string,List<double>>` that could be used to set the items in the PyDict object, but it got slower (I didn't make ver extensive performance tests).

mchandschuh: awesome that you benchmarked it! What kind of throughput is `GetDataFrame(IEnumerable<Slice> slice)` method getting?  I'd be interested in timings containing 100 data points for 1, 10 and 100 symbols. I suspect we have an exponential slowdown here (w.r.t num data points and/or num symbols)

I'll take a peek at this later -- I'm fairly convinced we can get it to a single iteration. One idea is this method could accept `IEnumerable` instead of `IEnumerable<T>` and while it enumerates (`while(enumerator.MoveNext())`) it can check each type `var type = enumerator.Current.GetType()` -- there's lots of options here.

AlexCatarino: That method is returning a `pandas.DataFrame` (it is represented as a `PyObject` object). That dataframe is an enhanced dictionary that is created based on a python dictionary (`PyDict` object).

Algorithm/QCAlgorithm.Python.cs
mchandschuh: Hmm, the only real other way is with reflection which is pretty messy... I was just hopeful for a little more code reuse here.

PR #1225, opened 2017-10-19
Allow trading of benchmark security when selected by universe selection

mchandschuh:
This change forces subscriptions added via universe selection to not be internal feeds. The internal feed flag, when set, prevents data from being piped into the algorithm which prevents a user from trading that security.


StefanoRaggi:
@mchandschuh A few regression tests are failing now: 
`CoarseFineFundamentalRegressionAlgorithm`, `CoarseFundamentalTop5Algorithm`, `CustomDataRegressionAlgorithm`, `DropboxBaseDataUniverseSelectionAlgorithm`, `DropboxUniverseSelectionAlgorithm`, `WeeklyUniverseSelectionRegressionAlgorithm`

Not sure if regression stats need to be updated as well.

PR #1228, opened 2017-10-19
Update IBController for Financial Advisor support

StefanoRaggi:
Updates to IBController: https://github.com/QuantConnect/ib-controller/pull/1


andrewhart098:
May want to update header comment in DockerfileLeanFoundation to reflect the date this file was last updated. Very helpful when testing

PR #1229, opened 2017-10-19
Begun process of adding Aussie equites and market to be available to algorithms.

thonmakerformvp:
I have updated the market.cs file to have an aussie market variable.

I have also updated the market-hours-database.json to have the correct market hours for aussie markets. within the market-hours-database.json the holidays and early closes need to be refined.

i have added one .zip file of an aussie equity - BHP daily data.

I have edited BasicTemplateAlgorithm.cs to try and run on the aussie BHP stock instead of the SPY.

When I run the algo, I get the following errors which stop the algo from working:

20171019 04:08:58 ERROR:: LocalDiskMapFileProvider.GetMapFileResolver(au): The specified directory does not exist: ../../../Data/equity/au/map_files

20171019 04:08:59 ERROR:: DefaultDataProvider.Fetch(): The specified file was not found: ../../../Data/equity/usa/daily/bhp.zip

Obviously, I need to add a map_files file - but am unsure what that is or how to do it. the algo is also obviously getting confused about where to look for data as it needs to be told to look in /Data/equity/au... not /Data/equity/usa... - I am unsure how to specify this though.


Algorithm.CSharp/BasicTemplateAlgorithm.cs
```diff
@@ -26,22 +26,22 @@ namespace QuantConnect.Algorithm.CSharp
...
+            SetStartDate(2011, 10, 07);  //Set Start Date
             SetEndDate(2013, 10, 11);    //Set End Date
             SetCash(100000);             //Set Strategy Cash
 
             // Find more symbols here: http://quantconnect.com/data
             // Forex, CFD, Equities Resolutions: Tick, Second, Minute, Hour, Daily.
             // Futures Resolution: Tick, Second, Minute
             // Options Resolution: Minute Only.
-            AddEquity("SPY", Resolution.Minute);
+            AddEquity("BHP", Resolution.Daily);
```
mchandschuh: These changes appear to be for local testing. Further, these will break the regression tests. Please remove these types of changes from the PR.

Common/Market.cs
```diff
@@ -118,6 +119,11 @@ static Market()
         /// GDAX
         /// </summary>
         public const string GDAX = "gdax";
+        
+        /// <summary>
+        /// AU
+        /// </summary>
```
mchandschuh: nit - perhaps spell out Australia (maybe not everyone knows AU == Australia?)

Algorithm.CSharp/BasicTemplateAlgorithm.cs
mchandschuh: If keeping this is critical, we could create a different `BasicAustralianMarketTemplateAlgorithm` or something similar.

Common/Market.cs
thonmakerformvp: I was instructed by @jaredbroad to use AU over Aus or Australia after some discussion in our slack channel. That said, this is commented out so perhaps we could change it to Australia for clarity?

Algorithm.CSharp/BasicTemplateAlgorithm.cs
thonmakerformvp: I do think keeping this is important as we are attempting to get Lean working on foreign equities and as such we need an algo attempting to run on these foreign equities as we try and configure everything. That said - it was silly of me to overwrite BasicTemplateAlgorithm like this instead of creating a new algo like you suggested. I will add a BasicAustralianMarketTemplateAlgorithm and revert BasicTemplateAlgorithm.

In regards to the meow.txt - that was just me being silly haha :)

thonmakerformvp:
When debugging please set your config.json to use AustralianEquityBasicTemplateAlgorithm.cs instead of BasicTemplateAlgorithm.cs

Common/Market.cs
mchandschuh: Agreed with using `Market.AU` -- matches `Market.USA` -- comment was only in regards to the xml comment on line 124.

Looks good as updated, thanks.

thonmakerformvp:
By adding a map_files directory I was able to get rid of the following error when running the AustralianEquity algo:

The specified directory does not exist: ../../../Data/equity/au/map_files

The following error persists:

ERROR:: DefaultDataProvider.Fetch(): The specified file was not found: ../../../Data/equity/usa/daily/bhp.zip

Upon inspecting the DefaultDataProvider class it seems the Fetch function is being passed in a bad string pointing to the usa directory. I have tried to figure out where in Lean calls this Fetch function so that I can change it to pass in the correct au directory but I have so far been unable to locate this. Any pointers on where to change the parameter to the DefaultDataProvider.Fetch() function are appreciated.

thonmakerformvp:
The AustralianEquityBasicTemplateAlgorithm can now successfully be backtested on the BHP stock within an Aussie market! 

Now moving on to handle currency and live trading.

jaredbroad:
Lots of work required here before a PR. Try keeping it generic and thinking of other/multi currency applications

PR #1233, opened 2017-10-20
Implements Portfolio Statistics in QuantBook

AlexCatarino:
Implements the GetPortfolioStatistics method that computes the portfolio statistics from a pandas.DataFrame using the existing Portfolio Statitics class from Lean.


AlexCatarino:
I will work on this PR.
Maybe it would be better to implement SetBenchmark and the GetPortfolioStatistics would make a Historical data request for the benckmark, instead of asking the user to include the benchmark in the input.
What do you think, @jingwu74 ?

jingwu74:
Yes great idea

PR #1237, opened 2017-10-20
Swap WebSocket4Net for WebSocketSharpFork

mchandschuh:
>NOTE: This still needs to be properly tested in the cloud

WebSocket4Net was throwing exceptions with newer mono versions.


Brokerages/GDAX/GDAXBrokerage.cs
```diff
@@ -99,7 +93,13 @@ public override bool PlaceOrder(Orders.Order order)
 
                 if (order.Type == OrderType.Market)
                 {
-                    OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, (decimal)raw.fill_fees, "GDAX Order Event") { Status = OrderStatus.Filled });
+                    Logging.Log.Trace("GDAXBrokerage.PlaceOrder(Market): Response: " + response.Content);
+                    OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, (decimal) raw.FillFees, "GDAX Order Event")
+                    {
+                        Status = OrderStatus.Filled,
```
jameschch: think this should be:
                        FillPrice = raw.Price,
                        FillQuantity = raw.FilledSize
This change should get all unit tests passing.

jaredbroad: Mono 5.+ web sockets worked well but caused a memory leak. We've revered to Mono 4.6.1.5 until can patch the memory leak or write a custom websocket implementation from scratch.

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -66,7 +66,6 @@ public partial class GDAXBrokerage : BaseWebsocketsBrokerage, IDataQueueHandler
         public GDAXBrokerage(string wssUrl, IWebSocket websocket, IRestClient restClient, string apiKey, string apiSecret, string passPhrase, IAlgorithm algorithm)
             : base(wssUrl, websocket, restClient, apiKey, apiSecret, Market.GDAX, "GDAX")
         {
-            throw new Exception("GDAX currently under maintenance and will be back online Monday 23rd October 2017");
```
jaredbroad: I removed this in https://github.com/QuantConnect/Lean/pull/1237/commits/410348b9a89a849144e1b6c98ebe6e2c07532281
strange it got back in.

PR #1251, opened 2017-10-24
Fix fee percentage in GDAXFeeModel

StefanoRaggi:


andrewhart098:
I believe the FractionalQuantityRegressionAlgorithm regression test statistics will need to change as well as it uses the GDAX brokerage

PR #1252, opened 2017-10-25
Add TradeDirection to Tick class and in GDAX ticks

StefanoRaggi:
Some brokerage tick/trade messages (including GDAX) include a direction field (aka side), indicating whether the tick was a Buy or a Sell.


StefanoRaggi:
Closing as new feature to be implemented later on, see also: #1277

PR #1256, opened 2017-10-26
Kraken Downloader

matthewsedam:
This is an implementation of a data downloader for the Kraken crypto exchange.


matthewsedam:
This is ready for review @jaredbroad.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Convert a DateTime object into a Unix time long value
+        /// </summary>
+        /// <param name="utcDateTime">The DateTime object (UTC)</param>
+        /// <returns>A Unix long time value.</returns>
+        /// <remarks>When we move to NET 4.6, we can replace this with DateTimeOffset.ToUnixTimeSeconds()</remarks>
+        private static long ToUnixTime(DateTime utcDateTime)
```
jaredbroad: Please use similar methods in the Time class

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,141 @@
...
+                var value = Decimal.Parse(i[0]);
+
+                yield return new Tick
+                {
+                    Time = time,
+                    Symbol = symbol,
+                    Value = value,
+                    AskPrice = value,
+                    BidPrice = value,
+                    TickType = TickType.Trade
```
jaredbroad: Is there a volume property in the data?

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,141 @@
...
+                    {
+                        errorCount++;
+                        System.Threading.Thread.Sleep(6000);
+                        response = client.DownloadString(url);
+                        result = JsonConvert.DeserializeObject<dynamic>(response);
+                    }
+
+                    if (result.error.Count != 0 && errorCount >= 10)
+                    {
+                        throw new Exception("Error in Kraken API: " + result.error[0]);
```
jaredbroad: Should probably check if the error array; chance of out of bounds exception here

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,141 @@
...
+                }
+
+                var value = Decimal.Parse(i[0]);
+
+                yield return new Tick
+                {
+                    Time = time,
+                    Symbol = symbol,
+                    Value = value,
+                    AskPrice = value,
```
jaredbroad: Are these quote ticks? If not; these should be unset.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,141 @@
...
+                    }
+
+                    if (result.error.Count != 0 && errorCount >= 10)
+                    {
+                        throw new Exception("Error in Kraken API: " + result.error[0]);
+                    }
+
+                    List<List<string>> newData = result.result[symbol.Value].ToObject<List<List<string>>>();
+                    data.AddRange(newData);
+                    last = Convert.ToInt64(result.result.last);
```
jaredbroad: Its OK for this PR but generally more efficient to parse and yield inline vs building up a memory array and iterating twice.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
jaredbroad: DataType = MarketDataType.Tick,

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
matthewsedam: What do you mean? result.error is an empty JSON array if the API call returned correctly. If we enter the if-statement, we have guaranteed that the size of the array >=1 (an error happened and Kraken is telling us why), so result.error[0] should never have an out-of-bounds exception. Am I reading this correctly?

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
matthewsedam: These are actual trades that happened on the exchange. How should I format that with the Tick class?

jaredbroad: DataType = MarketDataType.Tick,
TickType = TickType.Trade

And dont set the spread properties

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
matthewsedam: I think DateTimeOffset.ToUnixTimeSeconds() is only available in .NET 4.6. It is not allowing me to use in in the VS project.  It seems we are using a version of .NET 4.5. Is this correct?

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
matthewsedam: Fixed.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
matthewsedam: Fixed in latest commit.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
matthewsedam: Added volume data.

matthewsedam:
Ready for a second review @jaredbroad.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
mchandschuh: `QuantConnect.Time.DateTimeToUnixTimeStamp` and `QuantConnect.Time.UnixMillisecondTimeStampToDateTime`

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,166 @@
...
+                {
+                    url = string.Format(UrlPrototype, symbol.Value, last);
+                    response = client.DownloadString(url);
+                    result = JsonConvert.DeserializeObject<dynamic>(response);
+
+                    var errorCount = 0;
+                    while (result.error.Count != 0 && errorCount < 10)
+                    {
+                        errorCount++;
+                        System.Threading.Thread.Sleep(6000);
```
mchandschuh: I'm guessing this is self-imposed rate limiting?

If so, there's a pretty easy to use type available, `QuantConnect.Util.RateGate`. You can use it like this:
```csharp
// define gate allowing, for example, 10 calls per 10 minutes -- which is different than 1 call per minute
var rateGate = new RateGate(10, TimeSpan.FromMinutes(10));
// blocking call until the gate allows you through
rateGate.WaitToProceed();
```

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -13,6 +13,21 @@
     <FileAlignment>512</FileAlignment>
     <LangVersion>6</LangVersion>
     <TargetFrameworkProfile />
+    <PublishUrl>publish\</PublishUrl>
+    <Install>true</Install>
+    <InstallFrom>Disk</InstallFrom>
+    <UpdateEnabled>false</UpdateEnabled>
```
mchandschuh: Are any of these changes required?

If not, please revert them.

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -278,7 +295,18 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <BootstrapperPackage Include=".NETFramework,Version=v4.5.2">
+      <Visible>False</Visible>
+      <ProductName>Microsoft .NET Framework 4.5.2 %28x86 and x64%29</ProductName>
+      <Install>true</Install>
+    </BootstrapperPackage>
```
mchandschuh: I think these changes are related to clicking some of the install features.

Please revert these changes.

matthewsedam:
I have added in all suggestions @jaredbroad, @mchandschuh.

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,148 @@
...
+using Newtonsoft.Json;
+using QuantConnect.Data.Market;
+using QuantConnect.Util;
+
+namespace QuantConnect.ToolBox.KrakenDownloader
+{
+    /// <summary>
+    /// Kraken Data Downloader class
+    /// </summary>
+    class KrakenDataDownloader : IDataDownloader
```
mchandschuh: Can we make this class public?

matthewsedam:
It's now public @mchandschuh. Thanks!

ToolBox/KrakenDataDownloader/KrakenDataDownloader.cs
```diff
@@ -0,0 +1,148 @@
...
+                        DataType = MarketDataType.Tick,
+                        Symbol = symbol,
+                        TickType = TickType.Trade,
+                        Quantity = volume,
+                        Exchange = "kraken"
+                    };
+                }
+
+                var last = Convert.ToInt64(result.result.last);
+                while (last < endUnixTime)
```
jaredbroad: Can you explain why there are 2 yielding loops?

matthewsedam: The Kraken API only returns a subset of the set of all trades that has happened since the time the user specifies. As a result, we have to repeatedly call the API until it tells us that we have no more trades to pull from it. Having two yielding loops allows for the data to be processed and written to a file while we are waiting for the RateGate to allow us to request from Kraken more trades.

PR #1262, opened 2017-10-27
Add QCAlgorithm.AddSeries method

mchandschuh:
Adding a scatter plot is an involved process because there's not direct and easy way to just define a series object. This change allows an algorithm to simply define the series without pumping any data into it. This allows the user to set the series type and the unit for the series.

### DIFF
--- a/Algorithm/QCAlgorithm.Plotting.cs
+++ b/Algorithm/QCAlgorithm.Plotting.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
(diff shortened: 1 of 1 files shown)


Algorithm/QCAlgorithm.Plotting.cs
```diff
@@ -184,13 +184,33 @@ public void Plot(string chart, string series, decimal value)
...
+        /// <param name="unit">The unit of the y axis, usually $</param>
+        public void AddSeries(string chart, string series, SeriesType seriesType, string unit = "$")
+        {
+            Chart c;
+            if (!_charts.TryGetValue(chart, out c))
+            {
+                _charts[chart] = c = new Chart(chart);
+            }
+
+            c.Series[series] = new Series(series, seriesType, unit);
```
mchandschuh: Perhaps we should check if the series already exists and throw a error?

PR #1264, opened 2017-10-27
Add Financial Advisor support and Real Order Fees to IB Brokerage

StefanoRaggi:
This PR is a resubmission of https://github.com/QuantConnect/Lean/pull/1195 and https://github.com/QuantConnect/Lean/pull/1199

### Financial Advisor support

The `Order` base class has been expanded to include the `OrderProperties` nested class.
A new `DefaultOrderProperties` member was added to `QCAlgorithm` which can be set to an implementation of the `IOrderProperties` interface.
The `InteractiveBrokersOrderProperties` class was added to allow users to set Financial Advisor properties and will be expanded in the future to add support for additional IB features.

Order properties can be modified by users during algorithm `Initialize` and are always copied into all new orders submitted with any method available (`Order`, `SetHoldings`, etc.).

Usage examples can be found in the newly added C#/Python algorithm: `FinancialAdvisorDemoAlgorithm`.

### Real Order Fees

Previously, the IB brokerage was calling the `InteractiveBrokerageFeeModel` to calculate the fee for the full order quantity, ignoring the actual executions. This solution had a couple of issues:
- it didn't include fees for orders partially filled, then cancelled
- it couldn't calculate fees for orders sent to financial advisor groups/profiles

Now every `OrderEvent` fired (for total or partial fills) will only include the actual IB fee paid for each single execution.

Order fill handling logic has also been simplified by only processing fills in the IB `ExecutionDetails` and `CommissionReport` events, instead of `OrderStatusChanged` (which can fire multiple times or not at all in some cases, such as market orders executed immediately).


Algorithm/QCAlgorithm.cs
```diff
@@ -297,6 +297,11 @@ public AlgorithmSettings Settings
         public IOptionChainProvider OptionChainProvider { get; private set; }
 
         /// <summary>
+        /// Gets the default order properties
+        /// </summary>
+        public IOrderProperties DefaultOrderProperties { get; set; }
```
mchandschuh: Out of curiosity, what would I have to go through in order to *not* use these default properties? Say for example, I'd like to issue this trade with different algo parameters or a different fa method?

Brokerages/InteractiveBrokers/FinancialAdvisor/FinancialAdvisorConfiguration.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+    /// <summary>
+    /// Contains configuration data for a Financial Advisor
+    /// </summary>
+    public class FinancialAdvisorConfiguration
+    {
+        // Financial Advisor configuration data
+        private List<AccountAlias> _accountAliases = new List<AccountAlias>();
+        private List<Group> _accountGroups = new List<Group>();
+        private List<AllocationProfile> _allocationProfiles = new List<AllocationProfile>();
```
mchandschuh: nit - these properties should probably be made `readonly` - although, if not `readonly`, could instantiate a new instance instead of clearing them -- would remove the possibility of collection modified during enumeration error RE: line 48 and line 58. (Note sure if these are ever invoked from separate threads).

Common/Interfaces/IOrderProperties.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Contains additional properties and settings for an order
+    /// </summary>
+    public interface IOrderProperties
+    {
+        /// <summary>
+        /// Returns a new instance clone of this object
+        /// </summary>
+        IOrderProperties Clone();
```
mchandschuh: my mind rambling - it does seem slightly peculiar that the only method on this interface is a `Clone` method.

Common/QuantConnect.csproj
```diff
@@ -150,6 +150,8 @@
     <Compile Include="Brokerages\DefaultBrokerageModel.cs" />
     <Compile Include="Brokerages\FxcmBrokerageModel.cs" />
     <Compile Include="Brokerages\GDAXBrokerageModel.cs" />
+	<Compile Include="Orders\InteractiveBrokersOrderProperties.cs" />
```
mchandschuh: not action required -- interesting that a hard tab made its way in here though

Launcher/config.json
```diff
@@ -156,17 +156,17 @@
...
-        "real-time-handler": "QuantConnect.Lean.Engine.RealTime.LiveTradingRealTimeHandler",
-        "transaction-handler": "QuantConnect.Lean.Engine.TransactionHandlers.BrokerageTransactionHandler",
-        "history-provider": "BrokerageHistoryProvider"
+      "live-mode": true,
+
+      // real brokerage implementations require the BrokerageTransactionHandler
+      "live-mode-brokerage": "InteractiveBrokersBrokerage",
+      "setup-handler": "QuantConnect.Lean.Engine.Setup.BrokerageSetupHandler",
+      "result-handler": "QuantConnect.Lean.Engine.Results.LiveTradingResultHandler",
+      "data-feed-handler": "QuantConnect.Lean.Engine.DataFeeds.LiveTradingDataFeed",
```
mchandschuh: I"m guessing all the changes in this file are just white space?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -125,6 +126,11 @@ public override bool IsConnected
         }
 
         /// <summary>
+        /// Returns true if the connected user is a financial advisor
+        /// </summary>
+        public bool IsFinancialAdvisor => _account.Contains("F");
```
mchandschuh: Contains may be sufficient, but I thought it was explicitly a starts with relationship

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1293,39 +1334,30 @@ private void HandleOrderStatusUpdates(object sender, IB.OrderStatusEventArgs upd
...
-                    var totalQuantityFilled = Convert.ToInt32(order.AbsoluteQuantity - update.Remaining);
-                    var remainingQuantity = update.Remaining;
-                    var price = Convert.ToDecimal(update.LastFillPrice);
+                    // fill events will be only processed in HandleExecutionDetails and HandleCommissionReports
+                    return;
+                }
 
-                    if (!IsFillAlreadyHandled(order, update.OrderId, remainingQuantity))
+                // IB likes to duplicate/triplicate some events, so we fire non-fill events only if status changed
+                if (status != order.Status)
```
mchandschuh: Just realized that we could receive multiple back to back partial fills in which case the status wouldn't change... would we then not process those subsequent partial fills?
 - [ ] Investigate whether or not this if statement breaks partial fill support

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1388,18 +1430,84 @@ private void HandleExecutionDetails(object sender, IB.ExecutionDetailsEventArgs
...
+        /// <summary>
+        /// Emits an order fill (or partial fill) including the actual IB commission paid
+        /// </summary>
+        private void EmitOrderFill(Order order, Execution execution, CommissionReport commissionReport)
+        {
+            var currentQuantityFilled = Convert.ToInt32(execution.Shares);
+            var totalQuantityFilled = Convert.ToInt32(execution.CumQty);
+            var remainingQuantity = Convert.ToInt32(order.AbsoluteQuantity - totalQuantityFilled);
+            var price = Convert.ToDecimal(execution.Price);
+            var orderFee = Convert.ToDecimal(commissionReport.Commission);
```
mchandschuh: Wasn't there something about partial fills leading to duplicate fees or was that for a different brokerage impl?

Algorithm/QCAlgorithm.cs
StefanoRaggi: If you mean sending an order with different properties from those in `DefaultOrderProperties`, I believe we would need to have new overloads for Order, Buy, Sell and SetHoldings methods.
Maybe better address this in a separate PR ?

jaredbroad:
```
DefaultOrderProperties = new RiskyGroupOrderProperties();
MarketOrder("SPY", 100);

DefaultOrderProperties = new SafeGroupOrderProperties();
MarketOrder("SPY", 100);
```

mchandschuh: Definitely agreed not in the PR, but suspect it will be something that comes up

Common/Interfaces/IOrderProperties.cs
StefanoRaggi: Yes, seems strange, but for now the only `IOrderProperties` implementation is `InteractiveBrokersOrderProperties`, not sure when we'll add other implementations and if there may be common methods to be added to the interface.
The `Clone` method is the only one currently needed, as it is called internally by order methods.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: We have `Fxxxxx` accounts for real FA accounts and `DFxxxxx` for paper FA accounts

Launcher/config.json
StefanoRaggi: Yes, no changes, I'll revert, it's the new white space removal plugin we're all using :slightly_smiling_face:

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Order fills (partial or not) are no longer emitted in `HandleOrderStatusUpdates`:

```
                if (status == OrderStatus.Filled || status == OrderStatus.PartiallyFilled)
                {
                    // fill events will be only processed in HandleExecutionDetails and HandleCommissionReports
                    return;
                }
```

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Yes, the issue is in the GDAX brokerage (work in progress)

Brokerages/InteractiveBrokers/FinancialAdvisor/FinancialAdvisorConfiguration.cs
StefanoRaggi: The `Clear` and `Load` are always called sequentially by `Connect`, the `SubAccounts` property is not currently used.
I could have used `readonly`, but it would have required XML deserialization to temp objects followed by deep copy. Now the new instances returned from Deserialize are directly assigned to the members (line 131)

PR #1265, opened 2017-10-27
Kaiko data converter

### LINKED ISSUE #1260: Sample BTC data with no volume
The code expects trading volume information but the sample BTC data at Data\crypto\gdax\daily\btcusd.zip lacks it.

andrewhart098:
This project can take trade as well as orderbook information from Kaiko
and convert the data into Lean compatible data at all resolutions in
modeled as both trades and quotes.

Fixes #1260


Common/Time.cs
```diff
@@ -142,6 +142,27 @@ public static DateTime UnixTimeStampToDateTime(double unixTimeStamp)
...
+            DateTime time;
+            try
+            {
+                // Unix timestamp is seconds past epoch
+                time = EpochTime.AddMilliseconds(unixTimeStamp);
+            }
+            catch (Exception err)
+            {
+                Log.Error(err, "UnixTimeStamp: " + unixTimeStamp);
+                time = DateTime.Now;
```
mchandschuh: I was never a fan of catch clauses like these... they swallow errors unless you're looking at the logs

PR #1266, opened 2017-10-27
Improves PandasConverter

AlexCatarino:
In the previous implementation, each one of bars.Select statement causes an extra enumeration of the bars enumerable. To make matters worse, this is all invoked 4*N times, where N is the number of data points.

Another bottleneck was related to the way we concatenate pandas.DataFrames: we would add a new data frame to an existing data frame individually. Now, we collect all data frames and concatenate them in one operation.

With this new implementation, we have accomplshed a reduction of 90% of memory usage in a history request of 1000 symbols.

### DIFF
--- a/Common/Python/PandasConverter.cs
+++ b/Common/Python/PandasConverter.cs
@@ -61,116 +61,43 @@ namespace QuantConnect.Python
         /// <returns><see cref="PyObject"/> containing a pandas.DataFrame</returns>
         public PyObject GetDataFrame(IEnumerable<Slice> data)
         {
-            var symbols = data.SelectMany(x => x.Keys).Distinct().OrderBy(x => x.Value);
+            var sliceDataDict = new Dictionary<Symbol, PandasData>();
 
-            // If data contains derivatives and its underlying, 
-            // we get the underlying to exclude it from the dataframe 
-            Symbol underlying = null;
-            var derivatives = symbols.Where(x => x.HasUnderlying);
-            if (derivatives.Count() > 0)
+            foreach (var slice in data)
             {
-                underlying = derivatives.First().Underlying;
+                foreach (var baseData in slice.Values)
+                {
+                    if (!sliceDataDict.ContainsKey(baseData.Symbol))
+                    {
+                        sliceDataDict.Add(baseData.Symbol, new PandasData(baseData));
+                    }
+                    sliceDataDict[baseData.Symbol].Add(baseData);
+                }
             }
-            
+
+            var maxLevels = sliceDataDict.Max(x => x.Value.Levels);
+
             using (Py.GIL())
             {
-                var dataFrame = _pandas.DataFrame();
-
-                foreach (var symbol in symbols)
-                {
-                    if (symbol == underlying)
-                    {
-                        continue;
-                    }
-
-                    var items = new PyObject[]
-                    {
-                        dataFrame,
-                        GetDataFrame(data.Get<QuoteBar>(symbol)),
-                        GetDataFrame(data.Get<TradeBar>(symbol))
-                    };
-
-                    dataFrame = _pandas.concat(new PyList(items));
-                }
-
-                return dataFrame;
+                var dataFrames = sliceDataDict.Select(x => x.Value.ToPandasDataFrame(_pandas, maxLevels));
+                return _pandas.concat(dataFrames.ToArray());
             }
         }
 
         /// <summary>
-        /// Converts an enumerable of <see cref="IBaseDataBar"/> in a pandas.DataFrame
+        /// Converts an enumerable of <see cref="IBaseData"/> in a pandas.DataFrame
         /// </summary>
         /// <param name="data">Enumerable of <see cref="Slice"/></param>
         /// <returns><see cref="PyObject"/> containing a pandas.DataFrame</returns>
         public PyObject GetDataFrame<T>(IEnumerable<T> data)
-            where T : IBaseDataBar
+            where T : IBaseData
         {
-            if (data.Count() == 0)
+            var sliceData = new PandasData(data.FirstOrDefault());
+            foreach (var baseData in data)
             {
-                return _pandas.DataFrame();
+                sliceData.Add(baseData);
             }
-
-            using (Py.GIL())
-            {
-                var index = CreateIndex(data.First().Symbol, data.Select(x => x.Time));
-
-                var pyDict = new PyDict();
-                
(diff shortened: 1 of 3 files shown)


Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
-                        GetDataFrame(data.Get<TradeBar>(symbol))
-                    };
-
-                    dataFrame = _pandas.concat(new PyList(items));
+                    sliceDataDict[baseData.Symbol].Add(baseData);
                 }
+            }
 
-                return dataFrame;
+            var maxLevels = sliceDataDict.Max(x => x.Value.Levels);
```
mchandschuh: Can we compute this during the iteration above?

Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
-            
-            using (Py.GIL())
-            {
-                var dataFrame = _pandas.DataFrame();
-
-                foreach (var symbol in symbols)
+                foreach (var baseData in slice.Values)
                 {
-                    if (symbol == underlying)
+                    if (!sliceDataDict.ContainsKey(baseData.Symbol))
```
mchandschuh: If we use TryGetValue then we won't need the additional dictionary access below

Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
-                pyDict.SetItem("high", _pandas.Series(data.Select(x => (double)x.High).ToList(), index));
-                pyDict.SetItem("close", _pandas.Series(data.Select(x => (double)x.Close).ToList(), index));
 
-                if (typeof(T) == typeof(TradeBar))
+                foreach (var kvp in data)
                 {
-                    Func<IBaseDataBar, double> getVolume = x => { var bar = x as TradeBar; return (double)bar.Volume; };
-                    pyDict.SetItem("volume", _pandas.Series(data.Select(x => getVolume(x)).ToList(), index));
+                    var index = kvp.Value.Select(x => x.EndTime).ToList();
+                    var values = kvp.Value.Select(x => (double)x.Value).ToList();
```
mchandschuh: We could build both of these lists in a single iteration of kvp.Value

Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
+    }
+
+    /// <summary>
+    /// Organizes a list of data to create pandas.DataFrames 
+    /// </summary>
+    public class PandasData
+    {
+        private Symbol _symbol;
+        private List<DateTime> _timeIndex;
+        private Dictionary<string, List<double>> _series;
```
mchandschuh: Should the be read only?

Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
+        /// Implied levels of a multi index pandas.Series (depends on the security type)
+        /// </summary>
+        public int Levels
+        {
+            get
+            {
+                var levels = 2;
+                if (_symbol.SecurityType == SecurityType.Future) levels += 1;
+                if (_symbol.SecurityType == SecurityType.Option) levels += 3;
+                return levels;
```
mchandschuh: Why not calculate this once in constructor?

Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
+            _timeIndex = new List<DateTime>();
+        }
+
+        /// <summary>
+        /// Adds an object to the end of the lists
+        /// </summary>
+        /// <param name="baseData"><see cref="IBaseData"/> object that contains information to be saved in an instance of <see cref="PandasData"/></param>
+        public void Add(IBaseData baseData)
+        {
+            _timeIndex.Add(baseData.Time);
```
mchandschuh: Should we be using time or end time? Philosophical question really. For daily data people would expect time, the date of the data, vs other data end time makes sense. I want the data emitted at 9:42:00 vs saying I want the data that started at 9:39:59

mchandschuh: Perhaps we need a test of this pandas converter that puts data on and reads it back out and confirms the data is all the same and accessible in the same way

Common/Python/PandasConverter.cs
```diff
@@ -61,149 +61,271 @@ public PandasConverter(PyObject pandas = null)
...
+                }
             }
 
-            return _pandas.MultiIndex.from_tuples(tuples.ToArray(), names: names.Split(','));
+            var data = baseData as DynamicData;
+            if (data != null)
+            {
+                foreach (var kvp in data.GetStorageDictionary())
+                {
+                    _series[kvp.Key].Add(Convert.ToDouble(kvp.Value));
```
mchandschuh: I think we want the value property also for custom data types, such as Bitcoin

AlexCatarino:
The proposed changes were made. Thank you for the input.
Concerning the performance, it got about 40-50% faster. The 1000-equity history request before the change took 45-50 seconds and it now takes 24-26 seconds.

Common/Python/PandasConverter.cs
AlexCatarino: That was included in the line below. :)

Common/Python/PandasConverter.cs
```diff
@@ -189,7 +193,7 @@ public PandasData(IBaseData baseData)
         /// <param name="baseData"><see cref="IBaseData"/> object that contains information to be saved in an instance of <see cref="PandasData"/></param>
         public void Add(IBaseData baseData)
         {
-            _timeIndex.Add(baseData.Time);
+            _timeIndex.Add(baseData.EndTime);
```
mchandschuh: I was hoping for a little discussion around this change, @jaredbroad -- any comments?

One issue I see is with daily data. To access data *for* to 30th of October I would need to enter a date of the 31st

AlexCatarino: I think we should go for EndTime. If the users makes a History request to warm up and indicator, he should use EndTime in Indicator.Update method. If we give him Time. he would need to discount the Period.

mchandschuh: That's a pretty good point, also, I think it's rare people are accessing the data frame by date/time

PR #1267, opened 2017-10-29
GDAXDownloader added to Toolbox

espirulina:
GDAXDownloader Folder and files


ToolBox/GDAXDownloader/GDAXCandle.cs
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+*/
+using System;
+
+namespace QuantConnect.ToolBox.GDAXDownloader
+{
+    /// <summary>
+    /// Defines the JSON response structure from the GDAX API
+    /// </summary>
+    public class GDAXCandle
```
jaredbroad: I don't think this is used?

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,114 @@
...
+{
+    /// <summary>
+    /// Cryptoiq Data Downloader class 
+    /// </summary>
+    public class GDAXDownloader : IDataDownloader
+    {
+        private readonly string _exchange;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CryptoiqDownloader"/> class
```
jaredbroad: Old comments

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// </summary>
+        /// <param name="exchange">The bitcoin exchange</param>
+        /// <param name="scaleFactor">Scale factor used to scale the data, useful for changing the BTC units</param>
+        public GDAXDownloader(string exchange = Market.GDAX)
+        {
+            _exchange = exchange;
```
jaredbroad: This its safe to remove this constructor, seems like a duplication

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            var counter = startUtc;
+            const string url = "http://api.gdax.com/products/{0}/candles?start={1}&end={2}&granularity={3}";
+
+            while (counter <= endUtc)
+            {
+                System.Threading.Thread.Sleep(1000);
+                Console.WriteLine("Getting " + counter.ToShortDateString() + " data..");
+                DateTime endDate = counter.AddDays(1);
+                int granularity = 3600;
```
jaredbroad: Try using var syntax everywhere; eg 
var granularity = 3600;

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,114 @@
...
+                string data;
+                using (var reader = new System.IO.StreamReader(response.GetResponseStream(), encoding))
+                {
+                    data = reader.ReadToEnd();
+                }
+
+                data = data.Remove(0, 2);
+                data = data.Substring(0,data.Length - 2);
+
+                string[] bars = data.Split(new string[] { "],[" },StringSplitOptions.RemoveEmptyEntries);
```
jaredbroad: Can you use JSON.NET? Parsing the JSON manually seems too fragile.

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,114 @@
...
+                        Symbol = symbol,
+                        Low = decimal.Parse(items[1]),
+                        High = decimal.Parse(items[2]),
+                        Open = decimal.Parse(items[3]),
+                        Close = decimal.Parse(items[4]),
+                        Volume = decimal.Parse(items[5]),
+                        Value = decimal.Parse(items[4]),
+                        DataType = MarketDataType.TradeBar,
+                        Period = Time.OneHour,
+                        EndTime = new DateTime(1970, 1, 1, 0, 0, 0, 0).AddSeconds(long.Parse(items[0])).ToLocalTime().AddHours(1)
```
jaredbroad: I think there's helpers in Time class for creating a DateTime from Unix

ToolBox/GDAXDownloader/Program.cs
```diff
@@ -0,0 +1,63 @@
+﻿using System;
```
jaredbroad: Please add license headers to all the files

ToolBox/GDAXDownloader/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+                writer.Write(data);
+                
+                Console.ForegroundColor = ConsoleColor.Cyan;
+                Console.WriteLine("Finish data download");
+                
+            }
+            catch (Exception err)
+            {
+                Log.Error(err);
+                Console.WriteLine(err.Message);
```
jaredbroad: Generally we use Log.Trace() for general messaging so people can redirect to another logging service (Log4Net etc).

ToolBox/GDAXDownloader/GDAXDownloader.cs
espirulina: Thanks. have to check how to use JSON.NET to parse this kind of return data:

  | [[1498953600,222.72,242.13,230,235,1512.2072792299994],[1498867200,220.21,249,243.86,230.07,4865.78202293998],[1498780800,238.09,272,258.31,243.86,6538.636067629987]]]
-- | --

ToolBox/GDAXDownloader/GDAXDownloader.cs
espirulina: I only find such helper in .Net 4.6 
http://gigi.nullneuron.net/gigilabs/converting-tofrom-unix-timestamp-in-c/

StefanoRaggi: @espirulina I think Jared meant `QuantConnect.Time` static class:
https://github.com/QuantConnect/Lean/blob/master/Common/Time.cs#L128

espirulina:
Thank you @jaredbroad !
I updated according your recomendations.

ToolBox/GDAXDownloader/Program.cs
```diff
@@ -0,0 +1,75 @@
...
+                
+                var writer = new LeanDataWriter(Resolution.Hour, symbolObject, dataDirectory, TickType.Quote);
+                writer.Write(data);
+                
+                Console.WriteLine("Finish data download");
+                
+            }
+            catch (Exception err)
+            {
+                Log.Error(err);
```
jaredbroad: Generally just this line is enough; it automatically logs the stack trace as part of the Log.Error(err)

ToolBox/GDAXDownloader/Program.cs
```diff
@@ -0,0 +1,75 @@
...
+                const string market = Market.GDAX;
+                var downloader = new GDAXDownloader();
+
+                // Download the data
+                var symbolObject = Symbol.Create(args[0], SecurityType.Crypto, market);
+                var data = downloader.Get(symbolObject, Resolution.Hour, startDate, endDate);
+
+                // Save the data
+                
+                var writer = new LeanDataWriter(Resolution.Hour, symbolObject, dataDirectory, TickType.Quote);
```
jaredbroad: I think this is meant to be TickType.Trade as its returning TradeBars from the API?

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>QuantConnect.ToolBox</RootNamespace>
     <AssemblyName>QuantConnect.ToolBox</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
jaredbroad: The rest of solution is 4.5.2 lets not change this yet until we do so more systematically.

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,107 @@
...
+                {
+                    var a = JArray.Parse(data);
+
+                    foreach (var token in a.Children())
+                    {
+                        var barValues = JArray.Parse(token.ToString());
+
+                        for (int i = 0; i < barValues.Count; i++)
+                        {
+                            var dt = DateTimeOffset.FromUnixTimeSeconds(long.Parse(barValues[0].ToString())).DateTime;
```
jaredbroad: If we keep .NET 4.5.2 we'll have to use the QuantConnect Time class for this.

jameschch:
I think it does makes sense to pull backtest data from the exchange. I've found with small modifications to this code it's also possible to pull minute data in 4 hour chunks. At some point I was hoping to generalize one of the downloader tools to create consolidated bars in the same manner as the oanda downloader.

espirulina:
Hi @jaredbroad ! ready for review. Changed acording your suggestions and also increased freqency of request to 0.5 seconds, as GDAX api limit is 3 per second.
Take care,
Marco

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,115 @@
...
+                                };
+                                returnData.Add(tradeBar);
+                            }
+                        }
+                    }
+                    counter = counter.AddDays(1);
+                }
+                catch(System.Net.WebException ex)
+                {
+                    Console.WriteLine("BAD REQUEST: " + requestURL);
```
jaredbroad: Log.Error(ex);

jaredbroad: Generally anything in a re-usable class like this must use Log.X as the users can override the log handlers. Forcing console output in a reusable class means entries won't make it to the system log.

Console.Write etc is only passable in the Program.cs files where its theoretically an "application" of the reusable class object. Even there its better to use Log.Trace to keep it all standardized.

espirulina: Thanks

ToolBox/GDAXDownloader/GDAXDownloader.cs
```diff
@@ -0,0 +1,107 @@
...
+                var encoding = ASCIIEncoding.ASCII;
+
+                string data;
+                using (var reader = new System.IO.StreamReader(response.GetResponseStream(), encoding))
+                {
+                    data = reader.ReadToEnd();
+                }
+                if (data.Length > 0)
+                {
+                    var a = JArray.Parse(data);
```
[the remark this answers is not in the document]
espirulina: Thanks @mchandschuh

ToolBox/GDAXDownloader/Program.cs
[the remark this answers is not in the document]
espirulina: You are correct. Nothing is preventing us from using options for different Resolutions. This was like version 1. Thanks

espirulina:
@jaredbroad @mchandschuh  updates done. Please review. After its ok, ill move into to MultipleResolution options - only 1 hour woring right now

ToolBox/GDAXDownloader/Program.cs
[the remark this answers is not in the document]
mchandschuh: While I agree this may be version 1, the ability to down _only_ hourly data is not very useful to the rest of the community. Hourly data is probably the _least_ often resolution used. The other downloader programs accept a resolution as a command line argument, I'd recommend scooping the code from there.

jaredbroad:
thank you @espirulina !

PR #1269, opened 2017-10-30
When string cannot be converted into a symbol, log an error message

andrewhart098:
Often users will try to use a string in place of a symbol when the Symbol has not been registered using a method like AddSecurity, AddCrypto, AddEquity, etc. When this happens Lean silently fails to implicitly convert the string into a symbol object. This PR adds an error message when Lean is unable to convert the string into a Symbol object. 

This should add more context when users try to, for example, add an EMA using a string instead of `Symbol` which has not been registered in the Algorithm. In such a case, Lean produces a cryptic error message such as:

 ```System.Exception: Please register to receive data for symbol ' ' using the AddSecurity() function```. 

The error message is cryptic precisely because Lean cannot determine what symbol the user meant to use.

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -98,6 +98,13 @@ namespace QuantConnect.Algorithm
         }
 
         /// <summary>
+        /// Message for exception that is thrown when the implicit conversion between symbol and string fails
+        /// </summary>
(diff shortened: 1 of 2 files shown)


andrewhart098:
Reverted the log message in the implicit conversion between string and Symbol. Trailing whitespace still removed from Symbol class

PR #1283, opened 2017-11-03
Fix GDAX order fees with partial fills

StefanoRaggi:
Previously each partial fill was being attributed the total order fee, now each partial fill event includes the pro rated partial fee.

Also, the REST request to fetch the total order fee is sent only once, preventing the 5-second timeout in `BrokerageTransactionHandler` when receiving lots of partial fills with market orders.

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
@@ -178,9 +178,9 @@ namespace QuantConnect.Brokerages.GDAX
 
                 OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Information, -1, ("GDAXWebsocketsBrokerage.OnMessage: Unexpected message format: " + e.Message)));
             }
-            catch (Exception ex)
+            catch (Exception exception)
             {
-                OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Error, -1, $"Parsing wss message failed. Data: {e.Message} Exception: {ex.ToString()}"));
+                OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Error, -1, $"Parsing wss message failed. Data: {e.Message} Exception: {exception}"));
                 throw;
             }
         }
@@ -188,7 +188,7 @@ namespace QuantConnect.Brokerages.GDAX
         private void OrderMatch(string data)
         {
             var message = JsonConvert.DeserializeObject<Messages.Matched>(data, JsonSettings);
-            
+
             EmitTradeTick(message);
 
             var cached = CachedOrderIDs
--- a/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
@@ -47,6 +47,7 @@ namespace QuantConnect.Brokerages.GDAX
         private CancellationTokenSource _canceller = new CancellationTokenSource();
         private ConcurrentQueue<WebSocketMessage> _messageBuffer = new ConcurrentQueue<WebSocketMessage>();
         private volatile bool _streamLocked;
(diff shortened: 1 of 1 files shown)


Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -187,37 +188,39 @@ private void OnMessageImpl(object sender, WebSocketMessage e)
...
+            if (cached.Count == 0)
             {
                 return;
             }
 
             Log.Trace($"GDAXBrokerage.OrderMatch(): Match: {message.ProductId} {data}");
-            var orderId = cached.First().Key;
-            var orderObj = cached.First().Value;
+            var orderId = cached[0].Key;
+            var order = cached[0].Value;
```
mchandschuh: Surprised we didn't go with:
```c#
var order = cached[0].Value;
var orderId = order.Id;
```

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -226,19 +229,32 @@ private void OrderMatch(string data)
...
 
+            decimal totalOrderFee;
+            if (!_orderFees.TryGetValue(orderId, out totalOrderFee))
+            {
+                totalOrderFee = GetFee(order);
+                _orderFees[orderId] = totalOrderFee;
+            }
+
+            // apply order fee on a pro rata basis
+            var orderFee = totalOrderFee * Math.Abs(message.Size) / Math.Abs(order.Quantity);
```
mchandschuh: I think this is great for now, but just considered something -- someone edits the quantity of an order _after_ we've received a partial fill. This would cause the _total_ order feed to change as well as the pro rata calculations.

I'm not considering we attempt to address this now, just thinking out loud and thinking that there are even more complicated cases that we may have not yet anticipated.

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -226,19 +229,32 @@ private void OrderMatch(string data)
...
-                GetFee(cached.First().Value), $"GDAX Match Event {direction}"
+                orderFee, $"GDAX Match Event {direction}"
             );
 
             //if we're filled we won't wait for done event
             if (orderEvent.Status == OrderStatus.Filled)
             {
-                Orders.Order outOrder = null;
-                CachedOrderIDs.TryRemove(cached.First().Key, out outOrder);
+                Order outOrder;
```
mchandschuh: Perhaps this conditional should read `if (orderEvent.Status.IsClosed())`

I'm not sure if canceled events get pumped through here as well, but we would certainly want to clean `CachedOrderIDs` and `_orderFees` each time an order transitions into the closed stated, not just the filled state.

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -257,30 +273,33 @@ private void OrderDone(string data)
             }
 
             //is this our order?
-            var cached = CachedOrderIDs.Where(o => o.Value.BrokerId.Contains(message.OrderId));
+            var cached = CachedOrderIDs.Where(o => o.Value.BrokerId.Contains(message.OrderId)).ToList();
```
mchandschuh: This whole `CachedOrderIDs` collection is somewhat bothersome to me. LEAN provides a type, `IOrderProvider` that will fetch orders. My concern with this approach is we now have multiple stores of active orders (transaction handler and brokerage). I realize `CachedOrderIDs` is existing code, but once again, just thinking out loud.

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
jaredbroad: I don't think you can update orders in GDAX; agree with your point that Caching the order id's in the brokerage is a bad pattern -- unfortunately its a pattern we've replicated on all the brokerages now.

PR #1292, opened 2017-11-07
Emit GDAX quote ticks when best bid/ask prices or sizes changed

StefanoRaggi:
Previously quote ticks were only emitted when a trade was received.

A new preliminary implementation of the `OrderBook` class was added to the GDAX brokerage, so we can track the best bid/ask prices and sizes and emit a quote `Tick` when one of them has changed.

Also, quote ticks now include values for the `BidSize` and `AskSize` properties.

### DIFF
--- a/Brokerages/GDAX/BestBidAskUpdatedEventArgs.cs
+++ b/Brokerages/GDAX/BestBidAskUpdatedEventArgs.cs
@@ -1,0 +1,53 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+
+namespace QuantConnect.Brokerages.GDAX
+{
+    /// <summary>
+    /// Event arguments class for the <see cref="OrderBook.BestBidAskUpdated"/> event
+    /// </summary>
+    public sealed class BestBidAskUpdatedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the new best bid price
+        /// </summary>
+        public readonly Symbol Symbol;
+
+        /// <summary>
+        /// Gets the new best bid price
+        /// </summary>
+        public readonly decimal BestBidPrice;
+
+        /// <summary>
+        /// Gets the new best ask price
+        /// </summary>
+        public readonly decimal BestAskPrice;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BestBidAskUpdatedEventArgs"/> class
+        /// </summary>
+        /// <param name="symbol">The symbol</param>
+        /// <param name="bestBidPrice">The newly updated best bid price</param>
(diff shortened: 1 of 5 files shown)


Brokerages/GDAX/BestBidAskUpdatedEventArgs.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    /// <summary>
+    /// Event arguments class for the <see cref="OrderBook.BestBidAskUpdated"/> event
+    /// </summary>
+    public sealed class BestBidAskUpdatedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the new best bid price
+        /// </summary>
+        public readonly Symbol Symbol;
```
mchandschuh: No action required, but curious as to why the preference of field over read-only property here? Typical C# conventions say we should prefer properties. In this specific case I don't see arguments one way or another, but just curious as to what inspired the decision

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -185,6 +191,102 @@ private void OnMessageImpl(object sender, WebSocketMessage e)
...
+
+        private void OnL2Update(string data)
+        {
+            try
+            {
+                var message = JsonConvert.DeserializeObject<Messages.L2Update>(data);
+
+                var symbol = ConvertProductId(message.ProductId);
+
+                var orderBook = _orderBooks[symbol];
```
mchandschuh: Is it possible for the order book to not be created yet? Does GDAX documentation define a certain order of messages, regarding snapshot vs L2 update?

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -441,7 +525,7 @@ public override void Subscribe(LiveNodePacket job, IEnumerable<Symbol> symbols)
             WebSocket.Send(json);
 
 
-            OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Information, -1, "GDAXBrokerage.Subscribe: Sent subcribe."));
+            OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Information, -1, "GDAXBrokerage.Subscribe: Sent subscribe."));
```
mchandschuh: Instead of -1, perhaps we could use `"Subscribe"` or `"SentSubscribe"` as the code

Brokerages/GDAX/OrderBook.cs
```diff
@@ -0,0 +1,173 @@
...
+    {
+        private readonly object _locker = new object();
+        private readonly Symbol _symbol;
+        private readonly SortedDictionary<decimal, decimal> _bids = new SortedDictionary<decimal, decimal>();
+        private readonly SortedDictionary<decimal, decimal> _asks = new SortedDictionary<decimal, decimal>();
+
+        /// <summary>
+        /// Event fired each time <see cref="BestBidPrice"/> or <see cref="BestAskPrice"/> are changed
+        /// </summary>
+        public event EventHandler<BestBidAskUpdatedEventArgs> BestBidAskUpdated;
```
mchandschuh: This is cool. Could see this type of event being super useful in algorithms if/when we support L2 data

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
StefanoRaggi: According to [GDAX docs for level2](https://docs.gdax.com/#the-code-classprettyprintlevel2code-channel), a _snapshot_ message will be sent after subscription, followed by _l2update_ messages.

Brokerages/GDAX/BestBidAskUpdatedEventArgs.cs
StefanoRaggi: No reason at all to use fields, should be get properties, I'll fix it now

Brokerages/GDAX/GDAXBrokerage.Messaging.cs
StefanoRaggi: Only fixed the typo on that line, changing codes should be also done in other places, next PR :)

PR #1298, opened 2017-11-09
TickAggregator

andrewhart098:
Refactored method of consolidating tick data into data at lower resolutions used in the KaikoDataConverter into a general class in the Toolbox project. These new classes should be of interest to anyone writing a data converter or downloader.

These classes use consolidators to aggregate trade or quote ticks into a specified resolution. They hold the consolidated data in memory. An additional method, Flush, will add the current working bar to the consolidated data and return the consolidated data. These classes are especially useful in data conversion from one data type to the Lean data format.


ToolBox/TickAggregatingConsolidator.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// The resolution that the data is being aggregated into
+        /// </summary>
+        public Resolution Resolution { get; protected set; }
+
+        /// <summary>
+        /// This method is intended to return all of the
+        /// aggregated data and the bar that the consolidator is working on
+        /// </summary>
+        public abstract List<BaseData> Flush();
```
mchandschuh: Why does this method need to be abstract? It appears all implementations do the same thing save the unnecessary cast to a specific type (`TradeBar` vs `QuoteBar`)

ToolBox/TickAggregatingConsolidator.cs
```diff
@@ -0,0 +1,99 @@
...
+using QuantConnect.Data;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.ToolBox
+{
+    /// <summary>
+    /// Class that uses consolidators to aggregate tick data data
+    /// </summary>
+    public abstract class TickAggregatingConsolidator
```
mchandschuh: *Aggregating* consolidator almost seems redundant. I'm not the best with names but to me this type is able to perform consolidation *and* keep track of the history of what's been consolidated.

ToolBox/TickAggregatingConsolidator.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// The consolidated data
+        /// </summary>
+        public List<BaseData> Consolidated { get; protected set; }
+
+        /// <summary>
+        /// The resolution that the data is being aggregated into
+        /// </summary>
+        public Resolution Resolution { get; protected set; }
```
mchandschuh: Consider making this `private set` and having it set via the constructor of this class. Same for `Consolidated`. Also, `Consolidated` might be better typed as `IReadOnlyList` or `IEnumerable` publicly to prevent the *possibility* of external code being able to modify our data storage.

ToolBox/TickAggregatingConsolidator.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <summary>
+    /// Class that uses consolidators to aggregate tick data data
+    /// </summary>
+    public abstract class TickAggregatingConsolidator
+    {
+        /// <summary>
+        /// The consolidator used to aggregate data from
+        /// higher resolutions to data in lower resolutions
+        /// </summary>
+        public abstract IDataConsolidator Consolidator { get; }
```
mchandschuh: Consider setting this via a constructor argument in this class instead of having it abstract

ToolBox/TickAggregatingConsolidator.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Return all the consolidated data as well as the
+        /// bar the consolidator is currently working on
+        /// </summary>
+        public override List<BaseData> Flush()
+        {
+            // Add the last bar
+            Consolidated.Add(Consolidator.WorkingData as QuoteBar);
+            return Consolidated;
```
mchandschuh: If I call flush multiple times won't I end up with multiple *partially* completed working bars? Perhaps a better implementation here is to return a *copy* of `Consolidated` and add the working bar to the copy. Also, these cases are unnecessary since it's a list of `BaseData`

PR #1301, opened 2017-11-10
Don't include tick type in the filename for crypto

andrewhart098:
The tick archive already includes the tickType


mchandschuh:
@andrewhart098 please squash these two commits. In an ideal world every commit is a valid state of the repository (build/test passes). Ping me if you want some help with the git commands necessary.

PR #1305, opened 2017-11-13
Prevent duplicate security creation

mchandschuh:
Since securities are created AND THEN added to the algorithm, it was possible for two different universes to create the SAME security at the same time step. This led to two different Security instances and only one of them being added. Ironically, it appears that it was a last in wins situation.

This change aims to track securities that are created by a universe in a given time step and re-use the already created security. A more holistic solution here may be to encapsulate the creation/fetching of securities into a new type and then ensure we use that abstraction everywhere, but this would have led to many more changes.

### DIFF
--- a/Engine/DataFeeds/UniverseSelection.cs
+++ b/Engine/DataFeeds/UniverseSelection.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -156,14 +157,30 @@ public SecurityChanges ApplyUniverseSelection(Universe universe, DateTime dateTi
...
 
+            var keys = _pendingSecurityAdditions.Keys;
+            if (keys.Any() && keys.Single() != dateTimeUtc)
+            {
+                // if the frontier moved forward then we've added these securities to the algorithm
+                _pendingSecurityAdditions.Clear();
+            }
+
+            Dictionary<Symbol, Security> pendingAdditions;
+            if (!_pendingSecurityAdditions.TryGetValue(dateTimeUtc, out pendingAdditions))
```
mchandschuh: So this whole thing is kind of hacky, *but works* and is contained within this single file which is kind of nice. A more holistic approach would be to refactor everywhere in the solution where we try to get a security to use some common abstraction which would also be able to track these 'pending' security changes

StefanoRaggi: I like the single file change too! Nice fix.

jaredbroad: Could there be any issues from clearing the pending array using the Utc - day change over? Perhaps this should be something like "Latest Universe TimeZone Day"?

PR #1306, opened 2017-11-13
Enable History request based subscription TickType

andrewhart098:
Currently, Crypto security types are restricted to history requests that return QuoteBars. This removes that restriction and allows Crypto securities to make TradeBar history requests.

In addition, HistoryRequest includes information regarding the tick type, but that information is not passed to the HistoryPacket. Therefore, only the common data types for a history request are returned for history requests. This is problematic for crypto history requests that specify Trade History.


mchandschuh:
Nice catch, surprised that wasn't being communicated. This is reminiscent of days of yore where we had a one to one mapping of security type to tick type (as far as our data library was concerned). This assumption is quickly becoming incorrect and I posit that an analysis of tick type usage and where it's value comes from is warranted.

These changes look good to me. We'll at least be transporting/respecting the value on the packet.

PR #1308, opened 2017-11-13
Support headers in SubscriptionDataSource

mchandschuh:
It is not uncommon that a web service would require some form of authentication headers to be provided. This change aims to enable this by adding the header information to the source. This design
allows a user's custom data type to return the source properly hydrated with the required header values before being sent back through the rest of the LEAN infrastructure.


jaredbroad:
@Jay-Jay-D

PR #1309, opened 2017-11-13
Added LatestPriceFillModel

andrewhart098:
This fill model is provided because currently the data sourced for Crypto is limited to one minute snapshots for Quote data. This fill model will ignore the trade/quote distinction and return the latest pricing information in order to determine the correct fill price. This fill model is used for crypto securities in backtesting


Common/Orders/Fills/ImmediateFillModel.cs
```diff
@@ -448,7 +448,7 @@ private static bool IsExchangeOpen(Security asset)
             return true;
         }
 
-        private class Prices
+        public class Prices
```
mchandschuh: could this be `protected` ?

Common/Orders/Fills/LatestPriceFillModel.cs
```diff
@@ -0,0 +1,94 @@
...
+            var subscriptionTypes = asset.Subscriptions.Select(x => x.Type).ToList();
+
+            // Tick
+            var tick = asset.Cache.GetData<Tick>();
+            if (subscriptionTypes.Contains(typeof(Tick)) && tick != null)
+            {
+                var price = direction == OrderDirection.Sell ? tick.BidPrice : tick.AskPrice;
+                if (price != 0m)
+                {
+                    return new Prices(price, 0, 0, 0, 0);
```
mchandschuh: I think these zeros are bugs, should probably be `return new Prices(price, price, price, price, price);` since some methods, such as `MarketOnOpenFill` use the opening price and this would return zero. Similar comment for line 66.

EDIT: Upon further inspection the constructor properly does the initialization. Perhaps a constructor that takes a single decimal, `public Prices(decimal price)` would make this look less like a bug upon first inspection.

Common/Orders/Fills/LatestPriceFillModel.cs
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+
+            // Get both the last trade and last quote
+            // Assume that the security has both a trade and quote subscription
+            // This should be true for crypto securities
+            var quoteBar = asset.Cache.GetData<QuoteBar>();
+            var tradeBar = asset.Cache.GetData<TradeBar>();
+
+            if (tradeBar.Time > quoteBar.Time)
```
mchandschuh: **ALWAYS** use `EndTime` to compare two pieces of data. This is especially relevant when considering data with different resolutions. The `EndTime` is the time at which that data was emitted whereas the `Time` is the beginning of the data interval.

Common/Orders/Fills/ImmediateFillModel.cs
andrewhart098: I did this in order to test the `GetPrices` method directly in `LatestPriceFillModelTests`. If the accessibility of `Prices` anything less than public, I cannot test the `GetPrices` method because this is the type returned by `GetPrices`. I agree it's not ideal to make it a public class, but I'm not sure what else to do.

Common/Orders/Fills/LatestPriceFillModel.cs
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+
+            // Get both the last trade and last quote
+            // Assume that the security has both a trade and quote subscription
+            // This should be true for crypto securities
+            var quoteBar = asset.Cache.GetData<QuoteBar>();
+            var tradeBar = asset.Cache.GetData<TradeBar>();
+
+            if (tradeBar.EndTime > quoteBar.EndTime)
```
mchandschuh: @andrewhart098 @jaredbroad -- should this be the default behavior for immediate fill model as well? This statement basically states that if a trade has happened since the last quote, use the trade price, otherwise, use the quoted price. I think this makes sense as part of default behavior of ImmediateFillModel

PR #1310, opened 2017-11-14
Add time range check in UniverseSelection

StefanoRaggi:
With universe selection at daily resolution, when removing a security at the end of a backtest, this method is called with a start time greater than the end time, causing the error "No data loaded for XYZ because there were no tradeable dates for this security".

### DIFF
--- a/Common/Data/UniverseSelection/Universe.cs
+++ b/Common/Data/UniverseSelection/Universe.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


Common/Data/UniverseSelection/Universe.cs
```diff
@@ -188,6 +188,11 @@ public virtual Security CreateSecurity(Symbol symbol, IAlgorithm algorithm, Mark
         /// <returns>All subscriptions required by this security</returns>
         public virtual IEnumerable<SubscriptionRequest> GetSubscriptionRequests(Security security, DateTime currentTimeUtc, DateTime maximumEndTimeUtc)
         {
+            if (currentTimeUtc > maximumEndTimeUtc)
```
mchandschuh: Perhaps more correctly would be to avoid running universe selection at all when this condition is true. One way to accomplish this would be at the top of `ApplyUniverseSelection` to check the current frontier time and the algorithm end time (which gets passed in here as `currentTimeUtc` and `maximumEndTimeUtc` respectively). In this case, we'd still fetch data for universe selection that we won't act on, so maybe it even belongs a level higher... so we *prevent* universe selection from even fetching data to run on when the frontier is past the algorithm end time, This would probably require inspecting the current start/end times of the universe subscriptions to ensure they don't go past the algorithm end date.

Either way we go, we want to make sure that this condition isn't required to be implemented by derived types since this is a `virtual` method.

StefanoRaggi: I agree on moving the time check up one level, to the top of `ApplyUniverseSelection`, not sure about moving it higher: in the daily resolution universe case, doesn't the last bar of the backtesting period have to be emitted anyway (`EndTime` is one second after algorithm end time) ?

This is a simple test case algorithm:
```
    {
        {
            SetStartDate(2017, 1, 4);
            SetEndDate(2017, 1, 10);
            AddUniverse(Universe.DollarVolume.Top(2));
        }
    }
```

mchandschuh:
@StefanoRaggi -- thanks for squashing!
@jaredbroad -- this is good to be merged.

PR #1311, opened 2017-11-14
Implements historical futures data requests

AlexCatarino:
- Creates a FutureHistory class to help users to manage historical future data requests
- Adds working example in the template notebook

### DIFF
--- a/Engine/DataFeeds/CachingFutureChainProvider.cs
+++ b/Engine/DataFeeds/CachingFutureChainProvider.cs
@@ -61,17 +61,17 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
             return symbols;
         }
-    }
 
-    internal class FutureChainCacheEntry
-    {
-        public DateTime Date { get; }
-        public List<Symbol> Symbols { get; }
-
-        public FutureChainCacheEntry(DateTime date, List<Symbol> symbols)
+        private class FutureChainCacheEntry
         {
-            Date = date;
-            Symbols = symbols;
+            public DateTime Date { get; }
+            public List<Symbol> Symbols { get; }
+
+            public FutureChainCacheEntry(DateTime date, List<Symbol> symbols)
+            {
+                Date = date;
+                Symbols = symbols;
+            }
         }
     }
 }
--- a/Engine/DataFeeds/CachingFutureChainProvider.cs
+++ b/Engine/DataFeeds/CachingFutureChainProvider.cs
@@ -1,0 +1,77 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); 
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Lean.Engine.DataFeeds
+{
+    /// <summary>
+    /// An implementation of <see cref="IFutureChainProvider"/> that will cache by date future contracts returned by another future chain provider.
+    /// </summary>
+    public class CachingFutureChainProvider : IFutureChainProvider
+    {
+        private readonly ConcurrentDictionary<Symbol, FutureChainCacheEntry> _cache = new ConcurrentDictionary<Symbol, FutureChainCacheEntry>();
+        private readonly IFutureChainProvider _futureChainProvider;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CachingFutureChainProvider"/> class
+        /// </summary>
+        /// <param name="futureChainProvider"></param>
+        public CachingFutureChainProvider(IFutureChainProvider futureChainProvider)
+        {
+            _futureChainProvider = futureChainProvider;
+        }
+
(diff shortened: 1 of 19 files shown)


Jupyter/BacktestingFutureChainProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace QuantConnect.Lean.Engine.DataFeeds
+{
+    /// <summary>
+    /// An implementation of <see cref="BacktestingFutureChainProvider"/> that reads the list of contracts from open interest zip data files
+    /// </summary>
+    public class BacktestingFutureChainProvider
```
mchandschuh: If we're following the pattern of `IOptionsChainProvider` then I think we'll want to add an `IFuturesChainProvider` so this type can derive from the interface.

Jupyter/FutureHistory.cs
```diff
@@ -0,0 +1,75 @@
...
+using Python.Runtime;
+using QuantConnect.Data;
+using QuantConnect.Python;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Jupyter
+{
+    /// <summary>
+    /// Class to manage information from History Request of Options
```
mchandschuh: copy pasta in comment, options -> futures

Jupyter/BacktestingFutureChainProvider.cs
AlexCatarino: At team stand up meeting, we have decided to keep `BacktestingFutureChainProvider` in Jupyter and to not create the interface and have it all in `Common` project, since it is not as necessary as `IOptionsChainProvider`.

jaredbroad: *Until we can find a source for it in live trading.

jaredbroad: @AlexCatarino we should still add the interface even if no live implementation yet.

Engine/DataFeeds/LiveFutureChainProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// Gets the list of future contracts for a given underlying symbol
+        /// </summary>
+        /// <param name="symbol">The underlying symbol</param>
+        /// <param name="date">The date for which to request the future chain (only used in backtesting)</param>
+        /// <returns>The list of future contracts</returns>
+        public IEnumerable<Symbol> GetFutureContractList(Symbol symbol, DateTime date)
+        {
+            // Returns an empty list of contracts until we can find a source for it in live trading.
+            return Enumerable.Empty<Symbol>();
```
jaredbroad: This should probably throw a NotImplementedException?

Engine/DataFeeds/CachingFutureChainProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Lean.Engine.DataFeeds
+{
+    /// <summary>
+    /// An implementation of <see cref="IFutureChainProvider"/> that will cache by date future contracts returned by another future chain provider.
+    /// </summary>
+    public class CachingFutureChainProvider : IFutureChainProvider
```
jaredbroad: 👍 cool!

Engine/DataFeeds/LiveFutureChainProvider.cs
AlexCatarino: When I first discussed this with @StefanoRaggi, we thought of that...
I agree. It would be better to throw and let people know they cannot use it yet.

Engine/DataFeeds/CachingFutureChainProvider.cs
```diff
@@ -0,0 +1,77 @@
...
+            else
+            {
+                symbols = entry.Symbols;
+            }
+
+            return symbols;
+        }
+    }
+
+    internal class FutureChainCacheEntry
```
mchandschuh: Perhaps this should be a private class nested inside the `CachinFuturesChainProvider` ?

Engine/Setup/BacktestingSetupHandler.cs
```diff
@@ -198,6 +198,9 @@ public bool Setup(IAlgorithm algorithm, IBrokerage brokerage, AlgorithmNodePacke
                     // set the option chain provider
                     algorithm.SetOptionChainProvider(new CachingOptionChainProvider(new BacktestingOptionChainProvider()));
 
+                    // set the future chain provider
+                    algorithm.SetFutureChainProvider(new CachingFutureChainProvider(new BacktestingFutureChainProvider()));
```
mchandschuh: I love seeing composability :D

Jupyter/QuantBook.cs
```diff
@@ -368,6 +368,41 @@ public OptionHistory GetOptionHistory(Symbol symbol, DateTime date, Resolution?
...
+                                        typeof(QuoteBar),
+                                        x,
+                                        resolution ?? future.Resolution,
+                                        future.Exchange.Hours,
+                                        MarketHoursDatabase.FromDataFolder().GetDataTimeZone(future.Symbol.ID.Market, future.Symbol, future.Type),
+                                        Resolution.Minute,
+                                        future.IsExtendedMarketHours,
+                                        future.IsCustomData(),
+                                        DataNormalizationMode.Raw,
+                                        LeanData.GetCommonTickTypeForCommonDataTypes(typeof(QuoteBar), future.Type))
```
mchandschuh: this method, `LeanData.GetCommonTickTypeForCommanDataTypes` will always return `TickType.Quote` when given `typeof(QuoteBar` -- any reason to call the method if we always want quotes?

Jupyter/QuantBook.cs
```diff
@@ -368,6 +368,41 @@ public OptionHistory GetOptionHistory(Symbol symbol, DateTime date, Resolution?
...
+                    );
+
+            return new FutureHistory(_algorithm.HistoryProvider.GetHistory(requests, _algorithm.TimeZone).Memoize());
```
mchandschuh: I would tend towards placing the `Memoize` call inside the constructor -- the makes it so consumers don't need to worry about the multiple iteration issue and instead it becomes a feature of the `FutureHistory` class`

mchandschuh:
@jaredbroad -- anything left here that needs to be addressed?

PR #1312, opened 2017-11-14
Add GDAX post-only option for Limit orders + bug fixes

StefanoRaggi:
The `PostOnly` flag for Limit orders has been added as a property in `GDAXOrderProperties` class.
It can be set as default for all limit orders in Initialize:
```
DefaultOrderProperties = new GDAXOrderProperties { PostOnly = true };
```

A couple of bugs have also been fixed:
- In GDAX subscribe, the SPY benchmark was being passed to `PollTick`
- Rejected orders were not being marked as Invalid

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
@@ -490,10 +490,13 @@ namespace QuantConnect.Brokerages.GDAX
         /// </summary>
         public override void Subscribe(IEnumerable<Symbol> symbols)
         {
-
             foreach (var item in symbols)
             {
-                if (item.Value.Contains("UNIVERSE")) continue;
+                if (item.Value.Contains("UNIVERSE") ||
+                    item.SecurityType != SecurityType.Forex && item.SecurityType != SecurityType.Crypto)
+                {
+                    continue;
+                }
 
                 if (!IsSubscribeAvailable(item))
                 {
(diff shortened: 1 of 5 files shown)


Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -490,10 +490,13 @@ private void EmitTradeTick(Messages.Matched message)
         /// </summary>
         public override void Subscribe(IEnumerable<Symbol> symbols)
         {
-
             foreach (var item in symbols)
             {
-                if (item.Value.Contains("UNIVERSE")) continue;
+                if (item.Value.Contains("UNIVERSE") ||
+                    item.SecurityType != SecurityType.Forex && item.SecurityType != SecurityType.Crypto)
```
mchandschuh: I'm beginning t think we need a more deterministic way of defining what is a universe symbol. Consider a user defined universe where they give it whatever symbol they desire.

I'm not recommending a change for this PR, but just planting the seed for future thought :)

Brokerages/GDAX/GDAXBrokerage.cs
```diff
@@ -57,6 +57,15 @@ public override bool PlaceOrder(Order order)
                 payload.overdraft_enabled = true;
             }
 
+            var orderProperties = order.Properties as GDAXOrderProperties;
+            if (orderProperties != null)
+            {
+                if (order.Type == OrderType.Limit)
+                {
+                    payload.post_only = orderProperties.PostOnly;
```
mchandschuh: Love seeing this abstraction being used... gives the control to the algorithm.

PR #1313, opened 2017-11-14
fix zero implied volatility and greeks by adding price model

jingwu74:
change the format of the output (the history request returns a multi-index pandas data frame)

### DIFF
--- a/Algorithm.Python/BasicTemplateFuturesHistoryAlgorithm.py
+++ b/Algorithm.Python/BasicTemplateFuturesHistoryAlgorithm.py
@@ -36,8 +36,8 @@ from datetime import timedelta
 class BasicTemplateFuturesHistoryAlgorithm(QCAlgorithm):
 
     def Initialize(self):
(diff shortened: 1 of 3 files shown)


Algorithm.Python/BasicTemplateFuturesHistoryAlgorithm.py
```diff
@@ -36,8 +36,8 @@
 class BasicTemplateFuturesHistoryAlgorithm(QCAlgorithm):
 
     def Initialize(self):
-        self.SetStartDate(2013, 10, 8)
-        self.SetEndDate(2013, 10, 9)
+        self.SetStartDate(2016, 8, 17)
+        self.SetEndDate(2016, 8, 20)
```
jaredbroad: https://github.com/QuantConnect/Lean/tree/master/Data/future/usa/minute/es
We don't have this date range in the open source so it'll break the regression tests.

PR #1317, opened 2017-11-15
Implements Tick data support in PandasConverter

AlexCatarino:


Common/Python/PandasConverter.cs
```diff
@@ -66,15 +66,32 @@ public PyObject GetDataFrame(IEnumerable<Slice> data)
 
             foreach (var slice in data)
             {
-                foreach (var baseData in slice.Values)
+                foreach (var key in slice.Keys)
```
mchandschuh: from a performance standpoint it's much better to iteration directly over the dictionary which will yield `KeyValuePair<TKey, Value>`  -- this prevents us from needing to re-index into the dictionary every time we want the value.

mchandschuh: from a performance standpoint it's much better to iteration directly over the dictionary which will yield `KeyValuePair<TKey, Value>`  -- this prevents us from needing to re-index into the dictionary every time we want the value.

Common/Python/PandasConverter.cs
```diff
@@ -66,15 +66,32 @@ public PyObject GetDataFrame(IEnumerable<Slice> data)
...
+                        value.Add(slice[key]);
+                    }
+                    else
+                    {
+                        if (!sliceDataDict.TryGetValue(key, out value))
+                        {
+                            sliceDataDict.Add(key, value = new PandasData(ticks.First()));
+                            maxLevels = Math.Max(maxLevels, value.Levels);
+                        }
+                        foreach (var tick in ticks)
```
mchandschuh: This plus line 87 -- do we end up with a duplicate first tick?

Common/Python/PandasConverter.cs
```diff
@@ -166,7 +183,8 @@ public class PandasData
     {
         private readonly Symbol _symbol;
         private readonly List<DateTime> _timeIndex;
-        private readonly Dictionary<string, List<double>> _series;
+        private readonly Dictionary<string, List<double>> _doubleSeries;
```
mchandschuh: Not for this PR, but perhaps we should move `PandasData` to its own file in a separate change.

Common/Python/PandasConverter.cs
```diff
@@ -346,7 +387,18 @@ public PyObject ToPandasDataFrame(dynamic pandas, int levels = 2)
                 var index = pandas.MultiIndex.from_tuples(tuples, names: names.Split(','));
 
                 // Returns a dictionary keyed by column name where values are pandas.Series objects
-                return _series.ToDictionary(k => k.Key, v => pandas.Series(v.Value, index));
+                var pandasSeries = new Dictionary<string, dynamic>();
+                foreach (var kvp in _doubleSeries)
+                {
+                    if (kvp.Value.Count != tuples.Length) continue;
```
mchandschuh: If this does happen what does that mean? In what case would this condition hold true and are we hiding a bug by continuing in this case?

Common/Python/PandasConverter.cs
```diff
@@ -346,7 +387,18 @@ public PyObject ToPandasDataFrame(dynamic pandas, int levels = 2)
...
+                    pandasSeries.Add(kvp.Key, pandas.Series(kvp.Value, index));
+                }
+                foreach (var kvp in _stringSeries)
+                {
```
mchandschuh: Instead of having a string series and a double series, we could define the series to have values of type object.  This maybe be especially relevant when decomposing dynamic data type's storage dictionary since we don't know what the property type would be on the custom data.

Common/Python/PandasConverter.cs
AlexCatarino: When we create a pandas Series in the following line, it is required that the length of those list are the same.
This check was created because ticks that are of type trade does not have bid and ask prices, so I am not adding zeroes to the dictionary and, consequently, I am not adding that information to the data frame

Common/Python/PandasConverter.cs
AlexCatarino: I went for that solution because slice[key] is dynamic. On the other hand, `KeyValuePair<TKey, Value>` is `KeyValuePair<Symbol, BaseData>`, so I wouldn't be able to cast to to `List<Tick>`.

Common/Python/PandasData.cs
```diff
@@ -0,0 +1,280 @@
...
+
+        /// <summary>
+        /// Initializes an instance of <see cref="PandasData"/> with a sample list of <see cref="Tick"/> object
+        /// </summary>
+        /// <param name="ticks">List of <see cref="Tick"/> objects that contains information to be saved in an instance of <see cref="PandasData"/></param>
+        public PandasData(IEnumerable<Tick> ticks)
+        {
+            _series = "askprice,asksize,bidprice,bidsize,lastprice,quantity,exchange,suspicious".Split(',')
+                .ToDictionary(k => k, v => new List<object>());
+            _symbol = ticks.FirstOrDefault().Symbol;
```
mchandschuh: awesome, no longer duplicate first tick!

Common/Python/PandasData.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            var tick = baseData as Tick;
+            if (tick != null)
+            {
+                if (tick.TickType == TickType.Quote)
+                {
+                    _series["askprice"].Add((double)tick.AskPrice);
+                    _series["asksize"].Add((double)tick.AskSize);
+                    _series["bidprice"].Add((double)tick.BidPrice);
+                    _series["bidsize"].Add((double)tick.BidSize);
```
mchandschuh: not sure if we can handle this, but sometimes ticks only have bid or ask values, so may end up with zeros here.

Common/Python/PandasData.cs
```diff
@@ -0,0 +1,280 @@
...
+                }
+                else
+                {
+                    _series["exchange"].Add(tick.Exchange);
```
mchandschuh: Are we sure none of our quote ticks have the exchange specified?

Common/Python/PandasData.cs
```diff
@@ -0,0 +1,280 @@
...
+                _series["lastprice"].Add((double)tick.LastPrice);
+                _series["quantity"].Add((double)tick.Quantity);
+            }
+
+            var data = baseData as DynamicData;
+            if (data != null)
+            {
+                foreach (var kvp in data.GetStorageDictionary())
+                {
+                    _series[kvp.Key].Add(Convert.ToDouble(kvp.Value));
```
mchandschuh: `_series[kvp.Key]` is a `List<object>` no need to convert to double... maybe only convert if:
```csharp
var value = kvp.Value;
if (value is decimal) value = (double) value;
_series[kvp.Key].Add(value);
```

Common/Python/PandasData.cs
AlexCatarino: I just confirm it and we do have exchange for quote ticks.

PR #1323, opened 2017-11-16
Changes AddData method return type

AlexCatarino:
Unlike other methods used to add securities to the algorithm that return a Security or a derived object, AddData was returning void. This commit change it to return Security for consistency.


jaredbroad:
Thank you @AlexCatarino! @StefanoRaggi !

PR #1325, opened 2017-11-17
Implements QCAlgorithm.Fetch method

AlexCatarino:
- Fetch wraps the WebClient.DownloadString method. It allows python users to fetch data from a uri without a third party python library.
- Changes DropboxUniverseSelectionAlgorithm to showcase Fetch method

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -1832,13 +1832,22 @@ namespace QuantConnect.Algorithm
         /// The resource to download is specified as a <see cref="string"/> containing the URI.
         /// </summary>
         /// <param name="address">A string containing the URI to download</param>
+        /// <param name="headers">Defines header values to add to the request</param>
         /// <param name="userName">The user name associated with the credentials</param>
         /// <param name="password">The password for the user name associated with the credentials</param>
         /// <returns>The requested resource as a <see cref="string"/></returns>
-        public string Fetch(string address, string userName = null, string password = null)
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.cs
```diff
@@ -1809,5 +1810,21 @@ public void SetStatus(AlgorithmStatus status)
...
+
+        /// <summary>
+        /// Downloads the requested resource as a <see cref="string"/>.
+        /// The resource to download is specified as a <see cref="string"/> containing the URI.
+        /// </summary>
+        /// <param name="address">A string containing the URI to download</param>
+        /// <param name="userName">The user name associated with the credentials</param>
+        /// <param name="password">The password for the user name associated with the credentials</param>
+        /// <returns>The requested resource as a <see cref="string"/></returns>
+        public string Fetch(string address, string userName = null, string password = null)
```
mchandschuh: Some web services don't use network credentials, but instead use authorization headers. Do we want/need to support such a case?

Algorithm/QCAlgorithm.cs
```diff
@@ -1809,5 +1810,21 @@ public void SetStatus(AlgorithmStatus status)
...
+        {
+            using (var client = new WebClient { Credentials = new NetworkCredential(userName, password) })
+            {
+                return client.DownloadString(address);
```
mchandschuh: Also, we might want to setup the proxy as we do similarly in the `RemoteFileSubscriptionStreamReader` so it will use the local machine's proxy settings. This was a PR from a few years ago and helps people when they run locally through a proxy.

mchandschuh:
@AlexCatarino - bump :)

PR #1328, opened 2017-11-17
Cashbook result

jaredbroad:


Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -488,7 +489,8 @@ public void Update()
                 {
                     Statistics = deltaStatistics,
                     RuntimeStatistics = runtimeStatistics,
-                    ServerStatistics = serverStatistics
+                    ServerStatistics = serverStatistics,
+                    Cash = cashbook
```
mchandschuh: Might the cashbook make more sense to be in the packet containing the holdings vs the statistics?

PR #1329, opened 2017-11-17
Prevent duplicate security creation

mchandschuh:
Since securities are created AND THEN added to the algorithm, it was possible for two different universes to create the SAME security at the same time step. This led to two different Security instances and only one of them being added. Ironically, it appears that it was a last in wins situation.

This change aims to track securities that are created by a universe in a given time step and re-use the already created security. A more holistic solution here may be to encapsulate the creation/fetching of securities into a new type and then ensure we use that abstraction everywhere, but this would have led to many more changes.

### DIFF
--- a/Engine/DataFeeds/UniverseSelection.cs
+++ b/Engine/DataFeeds/UniverseSelection.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -157,14 +158,30 @@ public SecurityChanges ApplyUniverseSelection(Universe universe, DateTime dateTi
...
+                _pendingSecurityAdditions[dateTimeUtc] = pendingAdditions;
+            }
+
             // find new selections and add them to the algorithm
             foreach (var symbol in selections)
             {
                 // create the new security, the algorithm thread will add this at the appropriate time
                 Security security;
-                if (!_algorithm.Securities.TryGetValue(symbol, out security))
+                if (!pendingAdditions.TryGetValue(symbol, out security) && !_algorithm.Securities.TryGetValue(symbol, out security))
```
mchandschuh: The bug was this line here which was incorrectly using `||` instead of `&&`

PR #1332, opened 2017-11-20
Fix race condition in BacktestingBrokerage order cancellations

StefanoRaggi:
This PR fixes occasional `NullReferenceException`s in `BacktestingBrokerage.Scan` and in `OrderTicket.Cancel`

### DIFF
--- a/Common/Orders/OrderTicket.cs
+++ b/Common/Orders/OrderTicket.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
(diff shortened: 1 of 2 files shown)


Common/Orders/OrderTicket.cs
```diff
@@ -162,7 +162,13 @@ public IReadOnlyList<UpdateOrderRequest> UpdateRequests
         /// </summary>
         public CancelOrderRequest CancelRequest
         {
-            get { return _cancelRequest; }
+            get
+            {
+                lock (_cancelRequestLock)
+                {
+                    return _cancelRequest;
```
mchandschuh: There's never a need to lock around a return statement like this since returning an object reference is an atomic operation, IOW, it's impossible for this line to be corrupted by a race condition. No absolute need to change this, but figured I'd add some technical detail.

PR #1333, opened 2017-11-21
Fixes CalculateOrderQuantity to handle decimal quantities

AlexCatarino:
In quantity calculation, we didn't reduce the order quantity by decimal numbers, leading to a big step when dealing with crypto-currencies. For example, from 2.3456 it would drop to 1.3456 where values in between should've been tested.
- Updates RegressionTests statitics to reflect changes in CalculateOrderQuantity

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -957,48 +957,40 @@ namespace QuantConnect.Algorithm
             decimal marginRequired;
             decimal orderValue;
             decimal orderFees;
-            var feeToPriceRatio = 0;
+            var feeToPriceRatio = 0m;
 
             // compute the initial order quantity
-            decimal orderQuantity = targetOrderValue / unitPrice;
+            var orderQuantity = targetOrderValue / unitPrice;
 
-            if (orderQuantity % security.SymbolProperties.LotSize != 0)
-            {
-                orderQuantity = orderQuantity - (orderQuantity % security.SymbolProperties.LotSize);
-            }
-
-            var iterations = 0;
+            // rounding off Order Quantity to the nearest multiple of Lot Size
+            orderQuantity -= orderQuantity % security.SymbolProperties.LotSize;
 
             do
             {
-                // decrease the order quantity
-                if (iterations > 0)
-                {
-                    // if fees are high relative to price, we reduce the order quantity faster
-                    if (feeToPriceRatio > 0)
-                        orderQuantity -= feeToPriceRatio;
-                    else
-                        orderQuantity--;
-                }
+                // reduce order quantity by feeToPriceRatio, since it is faster than by lot size
+                // if it becomes nonpositive, return zero
+                orderQuantity -= feeToPriceRatio;
+                if (orderQuantity <= 0) return 0;
 
                 // generate the order
                 var order = new MarketOrder(security.Symbol, orderQuantity, UtcTime);
                 orderValue = order.GetValue(security);
                 orderFees = security.FeeModel.GetOrderFee(security, order);
-                feeToPriceRatio = (int)(orderFees / unitPrice);
+
+                // calculate feeToPriceRatio, if lower than lot size, use lot size instead
+                feeToPriceRatio = orderFees / unitPrice;
+                if (feeToPriceRatio < security.SymbolProperties.LotSize)
+                {
+                    feeToPriceRatio = security.SymbolProperties.LotSize;
+                }
 
                 // calculate the margin required for the order
                 marginRequired = security.MarginModel.GetInitialMarginRequiredForOrder(security, order);
 
-                iterations++;
+            } while (marginRequired > marginRemaining || orderValue + orderFees > targetOrderValue);
 
-            } while (orderQuantity > 0 && (marginRequired > marginRemaining || orderValue + orderFees > targetOrderValue));
-
-            //Rounding off Order Quantity to the nearest multiple of Lot Size
-            if (orderQuantity % security.SymbolProperties.LotSize != 0)
-            {
-                orderQuantity = orderQuantity - (orderQuantity % security.SymbolProperties.LotSize);
-            }
+            var remainder = orderQuantity % security.SymbolProperties.LotSize;
+            if (remainder > 0) orderQuantity -= remainder;
 
             // add directionality back in
             return (direction == OrderDirection.Sell ? -1 : 1) * orderQuantity;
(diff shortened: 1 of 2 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -957,48 +957,40 @@ public decimal CalculateOrderQuantity(Symbol symbol, decimal target)
...
+            var orderQuantity = targetOrderValue / unitPrice;
 
-            if (orderQuantity % security.SymbolProperties.LotSize != 0)
-            {
-                orderQuantity = orderQuantity - (orderQuantity % security.SymbolProperties.LotSize);
-            }
-
-            var iterations = 0;
+            // rounding off Order Quantity to the nearest multiple of Lot Size
+            orderQuantity -= orderQuantity % security.SymbolProperties.LotSize;
```
StefanoRaggi: If `security.SymbolProperties.LotSize` is zero we would get a `DivideByZeroException` here,
maybe better add a check and throw an error message if zero ?

AlexCatarino: If `security.SymbolProperties.LotSize` is non-positive, we should also throw and error message, because negative number is also incorrect, right? It would be an error in the `symbol-properties-database`
```
// if security lot size is non-positive, throw exception
if (security.SymbolProperties.LotSize <= 0)
{
    throw new Exception($"{security.Symbol} lot size cannot be less than or equal to zero");
}
```

mchandschuh: Maybe we should enforce these constraints in the symbol properties class instead of everywhere we consume it. I think enforcing a positive valued lot size is reasonable and correct.

AlexCatarino: What value would be correct? It is normally 1, but FXCM Forex is 1000 and crypto is 0.00000001. We could throw in `SymbolPropertiesDatabase.FromCsvLine` where we read the data.

jaredbroad: All out of scope for this PR. Agree safe assumption the input data is correct at this level of the application. If it was incorrect it should through on load as lot size of 0 isn't realistic.

mchandschuh:
I see some regression changes here but no unit tests for these changes. Perhaps we can add some unit tests for decimal cases for this method? I'm assuming they don't exist since regression results changed but no unit tests were updated.

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -957,48 +957,40 @@ public decimal CalculateOrderQuantity(Symbol symbol, decimal target)
...
 
-            } while (orderQuantity > 0 && (marginRequired > marginRemaining || orderValue + orderFees > targetOrderValue));
-
-            //Rounding off Order Quantity to the nearest multiple of Lot Size
-            if (orderQuantity % security.SymbolProperties.LotSize != 0)
-            {
-                orderQuantity = orderQuantity - (orderQuantity % security.SymbolProperties.LotSize);
-            }
+            var remainder = orderQuantity % security.SymbolProperties.LotSize;
+            if (remainder > 0) orderQuantity -= remainder;
```
mchandschuh: Not important, but this could easily be written without an if statement:

```csharp
orderQuantity -= orderQuantity % security.SymbolProperties.LotSize;
```

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -957,48 +957,40 @@ public decimal CalculateOrderQuantity(Symbol symbol, decimal target)
...
+                orderQuantity -= feeToPriceRatio;
+                if (orderQuantity <= 0) return 0;
 
                 // generate the order
                 var order = new MarketOrder(security.Symbol, orderQuantity, UtcTime);
                 orderValue = order.GetValue(security);
                 orderFees = security.FeeModel.GetOrderFee(security, order);
-                feeToPriceRatio = (int)(orderFees / unitPrice);
+
+                // calculate feeToPriceRatio, if lower than lot size, use lot size instead
```
mchandschuh: This comment describes exactly the next few lines of code, however it doesn't tell me *why* the next few lines of code are there. I think what this is really trying to do is find an incremental delta value for the next iteration step. This algorithm is basically a newton rhapson method on an undefined function (due to fee model).

Also, the removal of the `iterations` variable from this loop is somewhat concerning. I'm curious how many iterations the algorithm needs now. Previously I think most requests only required ~2 or ~3 iterations to complete (max of 2 for linear fee model with #shares).

Tests/RegressionTests.cs
```diff
@@ -597,22 +597,22 @@ private static TestCaseData[] GetRegressionTestParameters()
...
                 {"Average Loss", "-2.03%"},
-                {"Compounding Annual Return", "2313.556%"},
+                {"Compounding Annual Return", "2305.677%"},
                 {"Drawdown", "4.500%"},
-                {"Expectancy", "0.473"},
-                {"Net Profit", "4.458%"},
-                {"Sharpe Ratio", "3.408"},
+                {"Expectancy", "0.470"},
+                {"Net Profit", "4.453%"},
+                {"Sharpe Ratio", "3.405"},
```
mchandschuh: Changes here are expected.

Tests/RegressionTests.cs
```diff
@@ -229,22 +229,22 @@ private static TestCaseData[] GetRegressionTestParameters()
...
+                {"Sharpe Ratio", "1.381"},
                 {"Loss Rate", "31%"},
                 {"Win Rate", "69%"},
                 {"Profit-Loss Ratio", "1.64"},
                 {"Alpha", "0.055"},
                 {"Beta", "0.379"},
                 {"Annual Standard Deviation", "0.099"},
                 {"Annual Variance", "0.01"},
-                {"Information Ratio", "-0.703"},
+                {"Information Ratio", "-0.702"},
```
mchandschuh: I'm unsure of why there are changes here since the algorithm doesn't use a security that has decimalized quantity. Did we inspect why these values change? Can we define unit tests that would have failed *before* this change and that now pass due to his change (and subsequently describes why these values have changed).

I'm always hesitant when a regression test changes and I can't definitely say why.

Algorithm/QCAlgorithm.Trading.cs
AlexCatarino: I removed the iterations, because it was not being really used.
You are totally right, I ran the tests and it never needed more than 2 iterations.

PR #1334, opened 2017-11-21
Adds python version of BasicTemplateFrameworkAlgorithm

AlexCatarino:


Algorithm.Python/BasicTemplateFrameworkAlgorithm.py
```diff
@@ -0,0 +1,66 @@
...
+
+        self.SetStartDate(2013,10,07)  #Set Start Date
+        self.SetEndDate(2013,10,11)    #Set End Date
+        self.SetCash(100000)           #Set Strategy Cash
+
+        # Find more symbols here: http://quantconnect.com/data
+        # Forex, CFD, Equities Resolutions: Tick, Second, Minute, Hour, Daily.
+        # Futures Resolution: Tick, Second, Minute
+        # Options Resolution: Minute Only.
+        symbols = [ Symbol.Create("SPY", SecurityType.Equity, Market.USA) ]
```
mchandschuh: turned out much simpler than I expected!

So you were able to run this algo locally?

AlexCatarino: Yes, I was.
It is indeed super simple. If we want more symbols we could do something like this:
```
tickers = [ "SPY", "IBM" ]
symbols = [ Symbol.Create(x, SecurityType.Equity, Market.USA) for x in tickers ]
```

PR #1347, opened 2017-11-22
Refactors QuantBook

AlexCatarino:
It inherits from QCAlgorithm, since it is more useful for them to be standardized under one API.

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -34,14 +34,14 @@ namespace QuantConnect.Algorithm
 {
     public partial class QCAlgorithm
     {
-        private PandasConverter _converter;
+        public PandasConverter PandasConverter { get; private set; }
 
         /// <summary>
         /// Sets pandas converter
         /// </summary>
-        public void SetPandas()
+        public void SetPandasConverter()
         {
-            _converter = new PandasConverter();
+            PandasConverter = new PandasConverter();
         }
 
         /// <summary>
@@ -509,7 +509,7 @@ namespace QuantConnect.Algorithm
             var symbols = GetSymbolsFromPyObject(tickers);
             if (symbols == null) return null;
 
-            return _converter.GetDataFrame(History(symbols, periods, resolution));
+            return PandasConverter.GetDataFrame(History(symbols, periods, resolution));
         }
 
         /// <summary>
@@ -525,7 +525,7 @@ namespace QuantConnect.Algorithm
             var symbols = GetSymbolsFromPyObject(tickers);
             if (symbols == null) return null;
 
-            return _converter.GetDataFrame(History(symbols, span, resolution));
+            return PandasConverter.GetDataFrame(History(symbols, span, resolution));
         }
 
         /// <summary>
@@ -541,7 +541,7 @@ namespace QuantConnect.Algorithm
             var symbols = GetSymbolsFromPyObject(tickers);
             if (symbols == null) return null;
 
-            return _converter.GetDataFrame(History(symbols, start, end, resolution));
+            return PandasConverter.GetDataFrame(History(symbols, start, end, resolution));
         }
 
         /// <summary>
@@ -568,7 +568,7 @@ namespace QuantConnect.Algorithm
                 return CreateHistoryRequest(security, config, start, end, resolution);
             });
 
-            return _converter.GetDataFrame(History(requests.Where(x => x != null)).Memoize());
+            return PandasConverter.GetDataFrame(History(requests.Where(x => x != null)).Memoize());
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Algorithm/QCAlgorithm.Python.cs
```diff
@@ -34,14 +34,14 @@ namespace QuantConnect.Algorithm
...
     public partial class QCAlgorithm
     {
-        private PandasConverter _converter;
+        public PandasConverter PandasConverter { get; private set; }
 
         /// <summary>
         /// Sets pandas converter
         /// </summary>
-        public void SetPandas()
+        public void SetPandasConverter()
```
mchandschuh: No change required, but just a design comment. Typically a `SetXXX` method would accept the value to be set. In this case, something like `SetPandasConverter(IPandasConverter)` or something similar. Once again, not recommending the change now, just providing some general design feedback here.

Common/Python/PandasData.cs
```diff
@@ -63,8 +65,15 @@ public PandasData(IBaseData baseData)
                 // and add the field named 'value' since it is the reference value
                 columns = "value," + string.Join(",", ((DynamicData)baseData).GetStorageDictionary().Keys);
             }
-            
-            _series = columns.Split(',').ToDictionary(k => k, v => new List<object>());
+            // C# custom data
+            else if (baseData is BaseData)
+            {
+                _type = baseData.GetType();
+                var properties = _type.GetProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly);
```
mchandschuh: Won't `DeclaredOnly` cause issues with subclasses? From [MSDN](https://msdn.microsoft.com/en-us/library/system.reflection.bindingflags(v=vs.110).aspx):

```
Specifies that only members declared at the level of the supplied type's hierarchy should be considered. Inherited members are not considered.
```

If we don't have an explicit reason to use `BindingFlags.DeclaredOnly` then I would recommend removing it from this call. Also, might want to test this out. Consider the following class structure:

```csharp
public class MyCustomDataBaseClass : BaseData
{
    public int MyCustomDataProperty { get; set; }
}

public class MyDerivedCustomDataClass : MyCustomDataBaseClass
{
    public int MyDerivedCustomDataProperty { get; set; }
}
```

I think with what you've provided we won't get access to the `MyCustomDataProperty`

AlexCatarino: That is the idea: we don't want the subclass properties. We just want `EndTime`, `Value` and those there were defined/declared in the custom class.

mchandschuh: Wait... why don't we want subclass properties? I think you're saying we only want `EndTime` and `Value` from `BaseData`... but we should get all other sub class properties for the exact case described above. It reasonable to think that the user would want the `MyCustomDataBaseClass.MyCustomDataProperty` to be in the data frame, right?

A more tangible example perhaps is when a user subclasses `TradeBar` to get the OHLC properties for free. With this example, I think we would miss out on those properties given a type:
```csharp
public class ExtendedTradeBar : TradeBar
{
    public decimal TheAnswerToLife { get; set; } = 42;
}
```

I think with the proposed solution, only `TheAnswerToLife` would end up in the frame (along with `EndTime`, and `Value`) but I would also expect `Open`, `High`, `Low,` and `Close` from `TradeBar` as well.

AlexCatarino: I am not accounting for subclassing `TradeBar`.
```csharp
else if (baseData is BaseData)
```
Maybe I should do the other way around: remove the undesired properties like `DataType`, `Symbol`, etc...

mchandschuh: maybe something like this would do the trick:
```csharp
var properties = type.GetProperties()
    .Where(p => p.DeclaringType != typeof (BaseData)
             && p.DeclaringType != typeof (IBaseData));
```

And then manually add `EndTime` and `Value` as before.

Common/Python/PandasData.cs
```diff
@@ -43,28 +44,41 @@ public class PandasData
...
-            else if (baseData is DynamicData)
+            // C# custom data
+            else if (type != typeof(TradeBar) && type != typeof(QuoteBar))
             {
-                // We get the fields of DynamicData from the storage dictionary
-                // and add the field named 'value' since it is the reference value
-                columns = "value," + string.Join(",", ((DynamicData)baseData).GetStorageDictionary().Keys);
+                columns = "Value";
+                var properties = type.GetProperties();
+                var baseDataProperties = typeof(BaseData).GetProperties().Select(p => p.Name).ToHashSet();
```
mchandschuh: This can be moved to a static member variable since it won't change

Common/Python/PandasData.cs
```diff
@@ -103,6 +117,17 @@ public void Add(IBaseData baseData)
         {
             _timeIndex.Add(baseData.EndTime);
 
+            if (_type != null)
```
mchandschuh: Is this really a `IsCustomData` check? Somewhat obscured

PR #1348, opened 2017-11-24
Fix custom data time zone issue

Jay-Jay-D:
Previously, the custom data time zone was override with UTC even when it was defined in the `AddData` method.

This PR removes the `GetDataTimeZone` method and replaces those calls by retrieving the data time zone directly from the subscription.


Algorithm/QCAlgorithm.History.cs
```diff
@@ -479,7 +479,7 @@ public BaseData GetLastKnownPrice(Security security)
                 security.Symbol,
                 resolution,
                 security.Exchange.Hours,
-                MarketHoursDatabase.FromDataFolder().GetDataTimeZone(security.Symbol.ID.Market, security.Symbol, security.Symbol.SecurityType),
```
mchandschuh: We're not guaranteed that the data time zone and exchange time zone are the same. We can do something like this:
```csharp
var dataTimeZone = security.Subscriptions.Select(config => config.DataTimeZone).First();
```

Another possibility here is when we do `MarketHoursDatabase.GetEntry` and don't find an entry **and** an `overrideTimeZone` was specified, add the entry to the dictionary. This would guarantee that we're always fetching the configured overriden time zone.

@jaredbroad @StefanoRaggi -- Thoughts on adding entries to market hours database for custom data (would happen as part of `AddData<T>`)

Common/Securities/MarketHoursDatabase.cs
```diff
@@ -98,20 +98,6 @@ public SecurityExchangeHours GetExchangeHours(string market, Symbol symbol, Secu
...
 
-        /// <summary>
-        /// Performs a lookup using the specified information and returns the data's time zone if found,
-        /// if an entry is not found, an exception is thrown
-        /// </summary>
-        /// <param name="market">The market the exchange resides in, i.e, 'usa', 'fxcm', ect...</param>
-        /// <param name="symbol">The particular symbol being traded</param>
-        /// <param name="securityType">The security type of the symbol</param>
-        /// <returns>The raw data time zone for the specified security</returns>
-        public DateTimeZone GetDataTimeZone(string market, Symbol symbol, SecurityType securityType)
```
mchandschuh: The reasoning behind removing this was to force the caller to think about the override time zone. I'm beginning to think that a better route is to add an entry to market hours database when we do `AddData<T>` -- this will ensure the entire system gets the right value without needing to concern itself about custom data and overriden time zones.

A side effect s that we would remove the concept of overriding a time zone; replaced by entering a symbol specific entry to define the override.

Algorithm/QCAlgorithm.cs
```diff
@@ -1620,20 +1620,25 @@ public Security AddData<T>(string symbol, Resolution resolution, bool fillDataFo
         public Security AddData<T>(string symbol, Resolution resolution, DateTimeZone timeZone, bool fillDataForward = false, decimal leverage = 1.0m)
             where T : IBaseData, new()
         {
-            var marketHoursDbEntry = _marketHoursDatabase.GetEntry(Market.USA, symbol, SecurityType.Base, timeZone);
+            var key = new SecurityDatabaseKey(Market.USA, symbol, SecurityType.Base);
+            var entry = new MarketHoursDatabase.Entry(timeZone, SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork));
```
mchandschuh: I think we want to use the `timeZone` parameter passed in here

Common/Securities/MarketHoursDatabase.cs
```diff
@@ -237,5 +237,10 @@ public override Entry GetEntry(string market, string symbol, SecurityType securi
                 return new Entry(tz, SecurityExchangeHours.AlwaysOpen(tz));
             }
         }
+
+        public void AddEntry(SecurityDatabaseKey key, Entry entry)
+        {
+            _entries[key] = entry;
```
mchandschuh: This should be `_entries.Add` so that we get an exception when adding an entry that already exists. This also means that consumers who call add should first try to get a matching, symbol specific entry -- perhaps a `Contains` method is needed to make that easier for consumers?

mchandschuh:
Closing. This is superseded by #1466

PR #1350, opened 2017-11-27
Fix GDAX fill events for limit order executed as taker

StefanoRaggi:
Limit orders executed as a taker (for example submitting a buy price above market) had fill order events reporting the opposite buy/sell direction.

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
@@ -336,8 +336,8 @@ namespace QuantConnect.Brokerages.GDAX
                 ? OrderStatus.Filled : OrderStatus.PartiallyFilled;
 
             OrderDirection direction;
-            // Messages are always from the perspective of the market maker. Flip it in cases of a market order.
-            if (order.Type == OrderType.Market)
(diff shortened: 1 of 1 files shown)


Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -336,8 +336,8 @@ private void OrderMatch(string data)
                 ? OrderStatus.Filled : OrderStatus.PartiallyFilled;
 
             OrderDirection direction;
-            // Messages are always from the perspective of the market maker. Flip it in cases of a market order.
-            if (order.Type == OrderType.Market)
+            // Messages are always from the perspective of the market maker. Flip direction if executed as a taker.
+            if (order.BrokerId[0] == message.TakerOrderId)
```
mchandschuh: Any concern with `BrokerId` being empty here? What about multiple broker ids? Do we need to check `order.BrokerId.Any(bid => bid == message.TakerOrderId` ?

StefanoRaggi: I believe with GDAX there will always only be one element in order.BrokerId, the empty case should already be handled in line 317.

PR #1355, opened 2017-11-29
Updates pythonnet

AlexCatarino:
In this update, we fixed type correspondence from C# Decimal to python decimal that caused wrong method overload binding.

- Fixes FractionalQuantityRegressionAlgorithm:
With the pythonnet update we cannot pass a python float where a decimal is required.

### DIFF
--- a/Algorithm.Python/FractionalQuantityRegressionAlgorithm.py
+++ b/Algorithm.Python/FractionalQuantityRegressionAlgorithm.py
@@ -47,7 +47,7 @@ class FractionalQuantityRegressionAlgorithm(QCAlgorithm):
 
         self.SetTimeZone(DateTimeZone.Utc)
 
-        security = self.AddSecurity(SecurityType.Crypto, "BTCUSD", Resolution.Daily, Market.GDAX, False, 3.3, True)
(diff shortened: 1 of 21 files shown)


Algorithm.Python/FractionalQuantityRegressionAlgorithm.py
```diff
@@ -47,7 +47,7 @@ def Initialize(self):
 
         self.SetTimeZone(DateTimeZone.Utc)
 
-        security = self.AddSecurity(SecurityType.Crypto, "BTCUSD", Resolution.Daily, Market.GDAX, False, 3.3, True)
+        security = self.AddSecurity(SecurityType.Crypto, "BTCUSD", Resolution.Daily, Market.GDAX, False, d.Decimal(3.3), True)
```
mchandschuh: I'm unfamiliar with `d.Decimal` -- am I correct to assume that it's a function we added to pythonnet to convert float->decimal? Do we need documentation updates so users know to use this?

AlexCatarino: Yes, you are correct. I will scan the docs and see if we have examples without it and update them accordingly.

mchandschuh:
@AlexCatarino -- anyway to avoid breaking existing algorithms -- or would they just not work without this change?

PR #1358, opened 2017-11-30
Added SetWarmUp resolution parameter.

Maly-Lemire:


jaredbroad:
Nice idea thank you @Maly-Lemire

PR #1359, opened 2017-11-30
Fixes BasicTemplateFxcmVolumeAlgorithm

AlexCatarino:
The previous version was using unexisting data. At the moment, we only have hour and daily-resolution data for forex volume.
Once the resolution is changed, we have to modify the resolution of the indicator since both indicators in WeightedBy must have the same resolution.


Jay-Jay-D:
There is Minute data available for the FXCM Real Volume data and it can be downloaded using the [ToolBox's downloader](https://github.com/QuantConnect/Lean/blob/master/ToolBox/FxcmVolumeDownloader/Program.cs)

PR #1360, opened 2017-11-30
Intrinio economic data indices 

Jay-Jay-D:
This PR implements the [Intrinio US & Global Economic DataFeed](https://intrinio.com/data/economic) as custom data.


Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,218 @@
...
+                                                     string.Format("Basic ({0})", Convert.ToBase64String(byteKey)))
+                };
+
+                subscription = new SubscriptionDataSource(url, SubscriptionTransportMedium.RemoteFile, FileFormat.Csv,
+                                                          authorizationHeaders);
+                Thread.Sleep(millisecondsTimeout: 1000);
+            }
+            else
+            {
+                subscription = new SubscriptionDataSource("", SubscriptionTransportMedium.RemoteFile);
```
mchandschuh: We might want a comment here about what we're doing... and if it *works* perhaps returning null is more descriptive of why we're doing this. IIRC,this is so that we do not hit the intrinio API again

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,218 @@
...
+                var byteKey = Encoding.ASCII.GetBytes(string.Format("{0}:{1}", _user, _password));
+                var authorizationHeaders = new List<KeyValuePair<string, string>>
+                {
+                    new KeyValuePair<string, string>("Authorization",
+                                                     string.Format("Basic ({0})", Convert.ToBase64String(byteKey)))
+                };
+
+                subscription = new SubscriptionDataSource(url, SubscriptionTransportMedium.RemoteFile, FileFormat.Csv,
+                                                          authorizationHeaders);
+                Thread.Sleep(millisecondsTimeout: 1000);
```
mchandschuh: This seems very suspicious to me... why is this here in the `GetSource` method?

Jay-Jay-D: I added new comments explaining why. Basically, in case we have many sources of data, is the way to avoid excess the Intrinio's API limit of 1 call per second.

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
Jay-Jay-D: Is it returns null throws error:
`ParallelRunnerWorker.ThreadEntry():  System.NullReferenceException: Object reference not set to an instance of an object.`

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,228 @@
...
+                var authorizationHeaders = new List<KeyValuePair<string, string>>
+                {
+                    new KeyValuePair<string, string>("Authorization",
+                                                     string.Format("Basic ({0})", Convert.ToBase64String(byteKey)))
+                };
+
+                subscription = new SubscriptionDataSource(url, SubscriptionTransportMedium.RemoteFile, FileFormat.Csv,
+                                                          authorizationHeaders);
+                // Finally, here we force the engine to wait one second in case the we have multiples data sources added.
+                Thread.Sleep(millisecondsTimeout: 1000);
```
jaredbroad: //You can use a private class static to save the last API poll time and delay n-ms until at least 1s reached. 
// Doesn't scale well for 100 data sources though.. doesn't make Intrino very quant friendly.
public class IntrinioEconomicData : BaseData
{
   private static DateTime LastPollOfIntrinoAPI = new DateTime()
}

Common/Data/Custom/Intrinio/IntrinioEconomicDAta.cs
```diff
@@ -151,7 +153,9 @@ public override SubscriptionDataSource GetSource(SubscriptionDataConfig config,
                 subscription = new SubscriptionDataSource(url, SubscriptionTransportMedium.RemoteFile, FileFormat.Csv,
                                                           authorizationHeaders);
                 // Finally, here we force the engine to wait one second in case the we have multiples data sources added.
-                Thread.Sleep(millisecondsTimeout: 1000);
+                var intrinioApiCallLimit = 1000 - (DateTime.Now - LastIntrinoAPICall).Milliseconds;
+                if (intrinioApiCallLimit > 0) Thread.Sleep(millisecondsTimeout: intrinioApiCallLimit);
+                LastIntrinoAPICall = DateTime.Now;
```
Jay-Jay-D: Maybe a better implementation is this:
```cs
var intrinioApiCallLimit = 1000 - (DateTime.Now - LastIntrinoAPICall).Milliseconds;
do
{
	if (intrinioApiCallLimit > 0) Thread.Sleep(intrinioApiCallLimit);
	intrinioApiCallLimit = 1000 - (DateTime.Now - LastIntrinoAPICall).Milliseconds;
} while (intrinioApiCallLimit > 0);
LastIntrinoAPICall = DateTime.Now;
```
To avoid race condition issues.

mchandschuh:
The handling rate limits from within the `BaseData` type seems wrong to me. I think it would be fairly easy to implement this in the actual enumerator stack via the `RateLimitEnumerator<T>`.
@jaredbroad -- thoughts on `SubscriptionDataSource.RateLimit` -- or something similar? Basically thinking of a way for `BaseData` to *tell* the data feed it has a rate limit, vs the data feed being blocked *because* of a rate limit.

mchandschuh:
@Jay-Jay-D  -- what's the status on this? Are we happy with it? If you're able to solve the rate limiting issue from within the `BaseData` implementation then let's do that. Baking it into LEAN is a bigger concern for later.

Jay-Jay-D:
I think you're right this is not the most elegant solution to limit the API calls rate, but it works.

The only concern left is the race condition issue in my last comment.

Common/Data/Custom/Intrinio/IntrinioEconomicDAta.cs
Jay-Jay-D: @mchandschuh here what I mention below.

Algorithm.CSharp/BasicTemplateIntrinioEconomicData.cs
```diff
@@ -0,0 +1,92 @@
...
+            SetStartDate(2010, 01, 01); //Set Start Date
+            SetEndDate(2013, 12, 31); //Set End Date
+            SetCash(100000); //Set Strategy Cash
+
+            // Find more symbols here: http://quantconnect.com/data
+            // Forex, CFD, Equities Resolutions: Tick, Second, Minute, Hour, Daily.
+            // Futures Resolution: Tick, Second, Minute
+            // Options Resolution: Minute Only.
+            _uso = AddEquity("USO", Resolution.Daily, leverage: 2m).Symbol;
+            _bno = AddEquity("BNO", Resolution.Daily, leverage: 2m).Symbol;
```
mchandschuh: Ideally the example algorithms use data that is available in the open source repo.

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,231 @@
...
+    /// <seealso cref="QuantConnect.Data.BaseData" />
+    public class IntrinioEconomicData : BaseData
+    {
+        private static DateTime LastIntrinoAPICall;
+
+        private readonly string _baseUrl = @"https://api.intrinio.com/historical_data.csv?";
+
+        private readonly IntrinioDataTransformation _dataTransformation;
+        private readonly string _password = Config.Get("intrinio-password");
+        private readonly string _user = Config.Get("intrinio-username");
```
mchandschuh: Do we need a static method to set these? I'm thinking about quantconnect backtesting users where we wouldn't be specifying this in the configuration files.

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,231 @@
...
+                                                     string.Format("Basic ({0})", Convert.ToBase64String(byteKey)))
+                };
+
+                subscription = new SubscriptionDataSource(url, SubscriptionTransportMedium.RemoteFile, FileFormat.Csv,
+                                                          authorizationHeaders);
+                LastIntrinoAPICall = DateTime.Now;
+            }
+            else
+            {
+                subscription = new SubscriptionDataSource("", SubscriptionTransportMedium.LocalFile);
```
mchandschuh: Awesome. Empty source means do nothing. This is valid so long as we retrieve everything for backtest on the first invocation. What about multiple symbols?

Algorithm.CSharp/BasicTemplateIntrinioEconomicData.cs
Jay-Jay-D: I added two small files (29 KB) with the daily data needed to run this sample.

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
Jay-Jay-D: I tested it in the `BasicTemplateIntrinioEconomicData` algorithm and it works fine with two symbols.

Algorithm.CSharp/BasicTemplateIntrinioEconomicData.cs
jaredbroad: @Jay-Jay-D please submit a python example version of the algorithm

Common/Data/Custom/Intrinio/EconomicDataSources.cs
```diff
@@ -0,0 +1,477 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+namespace QuantConnect.Data.Custom.Intrinio
+{
+    public static class EconomicDataSources
```
mchandschuh: Should this be `IntrinioEconomicDataSources` ? Are these constants specific to Intrinio or are they common _tickers_ ?

Common/Data/Custom/Intrinio/IntrinioConfig.cs
```diff
@@ -0,0 +1,28 @@
...
+        public static string User { get; set; }
+        /// <summary>
+        ///     Intrinio API password
+        /// </summary>
+        public static string Password { get; set; }
+
+        /// <summary>
+        ///     Check if Intrinio API user and password are not empty or null.
+        /// </summary>
+        public static bool AreUserAndPasswordSet => !(string.IsNullOrWhiteSpace(User) && string.IsNullOrWhiteSpace(Password));
```
mchandschuh: not required, just me being nit picky :)
Naming -> just thinking of the if statement:
```csharp
if (IntrinioConfig.AreUserAndPasswordSet) { }
```
Or even
```csharp
if (IntrinioConfig.IsInitialized) { }
```

Might read better without the `Are` prefix.


Also, should this be `!(string.IsNullOrWhiteSpace(User) || string.IsNullOrWhiteSpace(Password)` ?

Consider user is null and password is set, we would want this to return `false`, but I think it would return true:
`! ( true && false )` => `!(false)` -> `true`

VS
`! (true || false )` => !(true)` -> `false` as desired

Another way to write it would be to lose the parenthesis:
```csharp
public static bool IsInitialized => !string.IsNullOrWhiteSpace(User) && !string.IsNullOrWhiteSpace(Password)
```

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,256 @@
...
+        /// Set the Intrinio API user and password.
+        /// </summary>
+        public static void SetUserAndPassword(string user, string password)
+        {
+            IntrinioConfig.User = user;
+            IntrinioConfig.Password = password;
+
+            if (IntrinioConfig.AreUserAndPasswordSet)
+            {
+                throw new NotImplementedException("Please set a valid Intrinio user and password.");
```
mchandschuh: Is this condition backwards? If the user name and password _are_ set... throw exception?

Common/Data/Custom/Intrinio/IntrinioEconomicData.cs
```diff
@@ -0,0 +1,256 @@
...
+                Time = time,
+                EndTime = time + QuantConnect.Time.OneDay,
+                Value = value
+            };
+        }
+
+        /// <summary>
+        /// Set the Intrinio API user and password.
+        /// </summary>
+        public static void SetUserAndPassword(string user, string password)
```
mchandschuh: Out of curiosity, why place this set method here instead of on `IntrinioConfig` ?

If this were placed on `IntrinioConfig` then the setters could be made private which is kind of nice, forces people to invoke he set username/password method directly.

Common/Data/Custom/Intrinio/EconomicDataSources.cs
Jay-Jay-D: Done!

PR #1361, opened 2017-11-30
Verify data validity after base data.reader

mchandschuh:
### Filter null or uninitialized data immediately

The convention throughout the documentation and source code is to either
return null or a new, uninitialized object to indicate that the reader method
was unable to produce an instance from the previous content. It appears over
time that these filters were forgotten within the data enumerator stacks.

### Update the FractionalQuantityRegressionAlgorithm statistics
This regression test presents itself occassionaly as a flickering test.
It seems the statistics were changed during such a flicker incorrectly.
By looking at the design of the algorith, one would expect it to trade
on consecutive days, which these statistcs represent. With the recent
filtering changes the flicker rate seems to be waning.

---

#### Original Analysis (might have some truths, left for posterity):
The issue stems from the bitcoin reader returning default instances instead of
null. These default instances get piped all the way through to the algorithm.
While this is happening, the trusty ol' benchmark security, SPY, is keeping
real time. The race is whether or not the SPY subscription can pass the first
real date in the bitcoin data. If this happens, multiple bitcoin dates would
all be dequeued at once since their in the past.

This change solves the race condition by filtering out invalid data points as
early as possible in the pipeline.

### DIFF
--- a/Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
+++ b/Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
@@ -118,7 +118,10 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                     {
                         foreach (var instance in instances.Data)
                         {
-                            yield return instance;
+                            if (instance != null && instance.EndTime != default(DateTime))
+                            {
+                                yield return instance;
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


mchandschuh:
Here's where the flicker in the algorithm occurs:
image

Note how one of them doesn't place a trade on 11/13/2015. I posit that the version that _does_ place the trade on that date is the correct version.

PR #1370, opened 2017-12-05
Fix option chain underlying price in TimeSlice

StefanoRaggi:
When creating time slices, in the option chain object the price of the underlying was set to the value of the previous time step when using `AddOption` and was equal to zero when using `AddOptionContract`.

For each time step in `algorithm.OnData`, these expressions are now guaranteed to have the same value:
- algorithm.Securities[underlyingSymbol]
- optionChain.Underlying.Price
- optionContract.UnderlyingLastPrice

### DIFF
--- a/Engine/DataFeeds/TimeSlice.cs
+++ b/Engine/DataFeeds/TimeSlice.cs
@@ -1,11 +1,11 @@
 /*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


Engine/DataFeeds/TimeSlice.cs
```diff
@@ -145,13 +147,14 @@ public static TimeSlice Create(DateTime utcDateTime, DateTimeZone algorithmTimeZ
             var futuresChains = new FuturesChains(algorithmTime);
             var symbolChanges = new SymbolChangedEvents(algorithmTime);
 
-            foreach (var packet in data)
+            // ensure we read equity data before option data, so we can set the current underlying price
+            foreach (var packet in data.OrderBy(x => x.Configuration.Symbol.SecurityType))
```
mchandschuh: I wonder what impact this has on performance? Order by is historically slow since it has to enumerate the entire collection into a sorted collection and then re-enumerates the sorted collection

StefanoRaggi: I don't like it much either, but couldn't think of other simple ways to guarantee that equities will be listed before options. Perhaps forcing order of subscriptions for data feeds (not enough though, the method is also called for history requests) ?

mchandschuh: maybe we can benchmark it to see if it's worth addressing now or later?

mchandschuh: @StefanoRaggi -- were we able to get a benchmark on this? If it has little impact that I'd say we just merge it as is. Honestly, even if there is a noticeable impact we still may want to merge it and resolve the performance issue later -- I wonder what @jaredbroad thinks?

StefanoRaggi: @mchandschuh I ran backtests on a benchmark algorithm with 500 equities (3 runs on master + 3 runs on rebased branch) and didn't notice any backtesting speed reduction

PR #1372, opened 2017-12-05
Feature algorithm framework

mchandschuh:
This PR adds a new framework for creating modular algorithms by inheriting from the `QCAlgorithmFramework` class and setting the various model properties. These include defining an `ISignalModel`, and `IPortfolioConstructionModel`, and optionally an `IExecutionModel` and an `IRiskManagementModel`.

On each time step, the signals model is updated with the current `Slice` to generate an enumerable of `ISignal`. These signals are piped into the `IPortfolioConstructionModel` to generate an enumerable of `IPortfolioTarget` which is then piped into the `IExecutionModel` to be converted into trades. The `IRiskManagementModel` runs at the end of each time step to ensure the algorithm's doesn't exceed it's risk parameters.


Engine/Signals/DefaultSignalHandler.cs
```diff
@@ -0,0 +1,259 @@
...
+        private DateTime _nextPersistenceUpdate;
+        private DateTime _nextChartUpdateAlgorithmTimeUtc;
+
+        private IMessagingHandler _messagingHandler;
+        private CancellationTokenSource _cancellationTokenSource;
+        private readonly ConcurrentQueue<Packet> _messages = new ConcurrentQueue<Packet>();
+        private readonly Dictionary<SignalScoreType, Series> _seriesByScoreType = new Dictionary<SignalScoreType, Series>();
+
+        /// <inheritdoc />
+        public bool IsActive => !_cancellationTokenSource.IsCancellationRequested;
```
StefanoRaggi: Does this need to be checked for null?

Algorithm.Framework/Selection/CoarseFundamentalPortoflioSelectionModel.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections.Generic;
+using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Framework.Selection
+{
+    /// <summary>
+    /// Portoflio selection model that uses coarse selectors. For US equities only.
+    /// </summary>
+    public class CoarseFundamentalPortoflioSelectionModel : FundamentalPortfolioSelectionModel
```
StefanoRaggi: Small typo in the class name

Algorithm.Framework/Selection/FineFundamentalPortoflioSelectionModel.cs
```diff
@@ -0,0 +1,63 @@
...
+using QuantConnect.Data.Fundamental;
+using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Framework.Selection
+{
+    /// <summary>
+    /// Portoflio selection model that uses coarse/fine selectors. For US equities only.
+    /// </summary>
+    public class FineFundamentalPortoflioSelectionModel : FundamentalPortfolioSelectionModel
```
StefanoRaggi: Small typo in the class name

Algorithm.Framework/Selection/CoarseFundamentalPortoflioSelectionModel.cs
```diff
@@ -0,0 +1,51 @@
...
+            Func<IEnumerable<CoarseFundamental>, IEnumerable<Symbol>> coarseSelector,
+            UniverseSettings universeSettings = null,
+            ISecurityInitializer securityInitializer = null
+            )
+            : base(false, universeSettings, securityInitializer)
+        {
+            _coarseSelector = coarseSelector;
+        }
+
+        /// <inheritdoc />
```
StefanoRaggi: Nice XML doc tag :) Not sure if our API documentation generator supports it ?

Algorithm.Framework/packages.config
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="NodaTime" version="1.3.4" targetFramework="net461" />
+  <package id="QuantConnect.pythonnet" version="1.0.5.1" targetFramework="net461" />
```
StefanoRaggi: Curious why `net461` instead of `net452`, the target framework for the project is correct (v4.5.2)

Algorithm.Framework/QCAlgorithmFramework.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            base.PostInitialize();
+        }
+
+        /// <summary>
+        /// Used to send data updates to algorithm framework models
+        /// </summary>
+        /// <param name="slice">The current data slice</param>
+        public override void OnFrameworkData(Slice slice)
```
StefanoRaggi: `OnFrameworkData` and `OnFrameworkSecuritiesChanged` can be declared `sealed` to prevent overriding them in user algorithms. This could avoid using reflection in the constructor to check if the methods have been overridden.

Algorithm.Framework/Selection/CoarseFundamentalPortoflioSelectionModel.cs
mchandschuh: Good call, @andrewhart098 -- do you know if we support `<inheritdoc />` tags?

Algorithm.Framework/QCAlgorithmFramework.cs
mchandschuh: That's a great idea and much better!

Algorithm/QCAlgorithm.cs
```diff
@@ -1858,5 +1890,14 @@ public string Download(string address, IEnumerable<KeyValuePair<string, string>>
...
         }
+
+        /// <summary>
+        /// Event invocator for the <see cref="SignalsGenerated"/> event
+        /// </summary>
+        /// <param name="signals">The collection of signals generaed at the current time step</param>
+        protected virtual void OnSignalsGenerated(IEnumerable<Signal> signals)
+        {
+            SignalsGenerated?.Invoke(this, new SignalCollection(UtcTime, signals));
+        }
```
StefanoRaggi: Being declared `virtual`, if user algorithms `override` this method, they have to call `base.OnSignalGenerated()`.
Maybe better add a separate virtual method for users to override ?

Engine/Signals/DefaultSignalHandler.cs
mchandschuh: Absolutely. It is unlikely, however, that one would check `IsActive` before calling `Run`, but I completely agree with you. Change to be available in next push.

Algorithm.Framework/Selection/CoarseFundamentalPortoflioSelectionModel.cs
mchandschuh: hehe, at least I was consistent with fine as well. Change available with next push.

Algorithm.Framework/Selection/FineFundamentalPortoflioSelectionModel.cs
mchandschuh: Addressed.

Algorithm.Framework/packages.config
mchandschuh: another great catch -- I think this is because when the project was first created it was targeting that version and then I must have added the nuget packages before changing it to 452. Fixed in next push.

Algorithm/QCAlgorithm.cs
mchandschuh: Hmm, in fact, maybe we just *don't* have this as virtual. Users can tap into the event stream via `IAlgorithm.SignalsGenerated += (sender, collection) => MyHandler(collection);` -- I don't see a need for them to insert themselves *before* our implementation runs.

Good call.

Algorithm.Framework/Selection/CoarseFundamentalPortoflioSelectionModel.cs
mchandschuh: I checked with @andrewhart098 and the doc server doesn't currently support this. I'll copy the docs around for good measure.

Common/Algorithm/Framework/Alphas/Analysis/Functions/BinaryAlphaScoreFunction.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            var startingValue = context.InitialValues.Get(alpha.Type);
+            var currentValue = context.CurrentValues.Get(alpha.Type);
+
+            switch (alpha.Direction)
+            {
+                case AlphaDirection.Down:
+                    return currentValue < startingValue ? 1 : 0;
+
+                case AlphaDirection.Flat:
```
jaredbroad: We should make flat indicate "None" -- realistically can't do flat predictions; that should be a volatility prediction.

PR #1373, opened 2017-12-05
Fix potential division by zero in BacktestingResultHandler

StefanoRaggi:
This can happen with algorithms that do not call `AddSecurity` and have a non-security benchmark such as `SetBenchmark(x => 0)`


jaredbroad:
Bug squash master! 🐛 🥇

Issue #1364, opened 2017-12-02
unrestricted growth of Portfolio size


AnnaZuzex:
While working on the algorithm we've found a problem with the increasing size of Portfolio and Securities storage lists.
Even if the securities are not used in the code or the algorithm and if they don't hold any stock or affected by current or later
transactions they will not be removed. Especially it becomes reasonable when algorithm selects stocks to work with on each iteration by selector
within AddUniverse. It leads to 2 types of problems which we noticed in our algorithm which works with different stocks every day:
1) it leads to out of memory error if backtesting is done for a significant period. New securities are added every day to university,
size can only increase even if we trade only 1 or 2 symbols it doesn't help.
2) Processing time between calls of our functions increase significantly over time as all the Portfolio is processed outside.

When we tried to solve this problem by removing elements which were never used in the trades the error would appear because those elements 
are accessed outside without any verification.

Where the AlgorithmManager tries to access the securities:
https://github.com/QuantConnect/Lean/blob/c475bde19ed462879dd29e8eb35a9f9c1bbdc05c/Engine/AlgorithmManager.cs#L432
in the line 432 of AlgorithManager.cs method ApplyDividend is called without any verification if the symbol exists in Portfolio or if it holds any stock etc.
The method ApplyDividend itself in SecurityPortfolioManager.cs 
https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityPortfolioManager.cs#L599
access variable Securities[dividend.Symbol] without verification if the symbol exists within security.
Later in the code of the algorihm manager method ApplySplit called in the similar manner
https://github.com/QuantConnect/Lean/blob/c475bde19ed462879dd29e8eb35a9f9c1bbdc05c/Engine/AlgorithmManager.cs#L441
Later on in the AlgorithmManager.cs it access securities again without any verification: 
https://github.com/QuantConnect/Lean/blob/c475bde19ed462879dd29e8eb35a9f9c1bbdc05c/Engine/AlgorithmManager.cs#L443
The structure of the code doesn't allow to affect the parameters stored in Slice.Dividends or Slice.Splits as algorithm doesn't have access to them.
===
How we see it is possible to solve this problem:
Option 1. Update code of the algorithmManager.cs to check existance before accessing dictionary
Option 2. Update methods ApplyDividend and ApplySplit in QCAlgorithm.cs to safely access the method
[...]

PR #1375, opened 2017-12-06
Issue 1364/unrestricted growth of portfolio size

AnnaZuzex:
Added checks if the symbol exists in Securities before it is accessed, in order to prevent from failures after removal of items from Securities
Problem is described in https://github.com/QuantConnect/Lean/issues/1364

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -928,6 +928,7 @@ namespace QuantConnect.Algorithm
         /// <returns>Order quantity to achieve this percentage</returns>
         public decimal CalculateOrderQuantity(Symbol symbol, decimal target)
         {
+            if (!Securities.ContainsKey(symbol)) return 0; // check if symbol exists in Securities
             var security = Securities[symbol];
             var price = security.Price;
 
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -928,7 +928,6 @@ namespace QuantConnect.Algorithm
         /// <returns>Order quantity to achieve this percentage</returns>
         public decimal CalculateOrderQuantity(Symbol symbol, decimal target)
         {
-            if (!Securities.ContainsKey(symbol)) return 0; // check if symbol exists in Securities
             var security = Securities[symbol];
             var price = security.Price;
 
--- a/Common/Securities/SecurityPortfolioManager.cs
+++ b/Common/Securities/SecurityPortfolioManager.cs
@@ -596,6 +596,11 @@ namespace QuantConnect.Securities
         /// <param name="dividend">The dividend to be applied</param>
         public void ApplyDividend(Dividend dividend)
         {
+            // check if symbol exists in Securities
+            if (!Securities.ContainsKey(dividend.Symbol))
+            {
+                return;
+            }
             var security = Securities[dividend.Symbol];
 
             // only apply dividends when we're in raw mode or split adjusted mode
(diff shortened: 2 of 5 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -928,6 +928,7 @@ public decimal CalculateOrderQuantity(Symbol symbol, double target)
         /// <returns>Order quantity to achieve this percentage</returns>
         public decimal CalculateOrderQuantity(Symbol symbol, decimal target)
         {
+            if (!Securities.ContainsKey(symbol)) return 0; // check if symbol exists in Securities
```
AlexCatarino: If it returns zero when the symbol is not in Securities, we hide this user error:
```This asset symbol ( ) was not found in your security list. Please add this security or check it exists before using it with 'Securities.ContainsKey(" ")'```

File: Common/Securities/SecurityPortfolioManager.cs [hunk 2]
```diff
@@ -596,6 +596,11 @@ public List<SubmitOrderRequest> ScanForMarginCall(out bool issueMarginCallWarnin
         /// <param name="dividend">The dividend to be applied</param>
         public void ApplyDividend(Dividend dividend)
         {
+            // check if symbol exists in Securities
+            if (!Securities.ContainsKey(dividend.Symbol))
+            {
+                return;
+            }
```
AlexCatarino: We are double-checking, since the same if-statement is found in `AlgorithmManager`. 
Same for `ApplySplitTo Equities`.

AnnaZuzex: I have removed the double check both here and in ApplyToEqities, 
commit c4f54610232a296fc5f0985f7b23a75b443492a6

Algorithm/QCAlgorithm.Trading.cs
AnnaZuzex: I have removed this check,
commit c4f54610232a296fc5f0985f7b23a75b443492a6

jaredbroad:
Hi @AnnaZuzex - thanks for the PR; when you physically remove the securities from the portfolio do you see the RAM usage go down in python?

AnnaZuzex:
Hello, @jaredbroad. We have not been able to test it on the main system due to the problems described in this issue. Locally we don't have enough data to see the difference. And on the platform the algorithm will not pass further than 1 iteration due to the problem of accessing securities in the apply dividend. When we see the memory problem no elements within our algorithm class increase in size except for the Portfolio. That's why we would like to try to remove elements to check if that would help. 
For now based on the analysis of the code we don't see what else can affect it. 
We use C# version of the platform.

mchandschuh:
@AnnaZuzex - This PR doesn't seem to remove any securities, just adding checks for when securities are removed. I posit that these checks are more likely to hide bugs than solve them. How are you removing the security objects? Perhaps we need holistic approach here.

AnnaZuzex:
@mchandschuh We've tried different approaches. Last one was Portfolio.Securities.Remove() all of them failed due to usage of the security later in the AlgorithmManager.cs. When we remove them we check that algorithm don't have securities used in the transactions and that they don't hold any stock. What we are trying to achieve is to run our algorithm with dynamic security selection (similar to the one attached to the issue) without out of memory problems and increasing processing time between iterations of our algorithm. We are open to other more effective approaches if you can suggest them.

jaredbroad:
We appreciate the interest and help to make LEAN more efficient! But for now will close this PR as it seems exploratory and not a finished piece of work with proof of the solution.

We save the portfolio objects to keep track of the fees/trading volumes etc. Our theory was these objects are references (so small) and wouldn't affect the overall memory of the backtest. 

If we can remove these objects & prove the RAM is markedly different then we can work on a different solution to saving the trade fee / volume data  or "archiving" the objects. 

@AnnaZuzex I recommend downloading FX data; renaming it over and over into many symbol names so you can test a universe of a few hundred stocks and hack locally first; then update the Github _issue_ when you've found more information. Once we've got the fix we can reopen the PR  :+1:

PR #1376, opened 2017-12-06
Fix bugs in WeeklyUniverseSelectionRegressionAlgorithm

StefanoRaggi:
This algorithm had a couple of issues:
- it was using a coarse universe with no data files available, so it has been changed to use a custom universe
- it was using algorithm time instead of selection time in the selector function

The regression stats have also been updated to match the new algorithm code.


Algorithm.CSharp/WeeklyUniverseSelectionRegressionAlgorithm.cs
```diff
@@ -26,22 +26,22 @@ namespace QuantConnect.Algorithm.CSharp
...
         /// </summary>
         public override void Initialize()
         {
-            UniverseSettings.Resolution = Resolution.Hour;
-
             SetStartDate(2013, 10, 1);  //Set Start Date
             SetEndDate(2013, 10, 31);    //Set End Date
             SetCash(100000);             //Set Strategy Cash
 
+            SetBenchmark(dt => 1m);
```
mchandschuh: Ideally we could have this running without this `SetBenchmark` line -- I do realize that that with the 2% change the benchmark dependent stats will change, but the orders/fills/security changes should be identical.

StefanoRaggi: Ok, removed SetBenchmark call

PR #1382, opened 2017-12-08
set the data normalization mode to be raw for options algorithms

jingwu74:


mchandschuh:
We'll want to make similar changes to the csharp versions as well as update the regression statistics.

Perhaps this should be default behavior for options?

mchandschuh:
Superseded by https://github.com/QuantConnect/Lean/pull/1390 which handles this in `PostInitialize` to force compliance.

Issue #1023, opened 2017-07-20
Don't use Suspicious ticks in fill models


4real:
Okay, so I found one algo I was using was making its profit from Suspicious ticks. Apparently, these aren't marked in the real time stream (so filtering them out altogether is a bad idea), but fill models in backtests shouldn't be using them.

jaredbroad:
When we build the bars we filter many exchanges on top of the suspicious
ticks. This is a heuristic decision as we hope to make the data as close to
what you'd really get at the second and minute level.

Generally we expect people using ticks to understand them fully as they are
harder to use. Equally people could request fill models which only use the
suspicious or off exchange fills.

4real:
Pretty sure it's going to be a surprise for quite a few people writing their first tick algos. It's a trap for e.g. mean reversion algos. ML algos also love picking up signals like this.

jaredbroad:
Implemented in https://github.com/QuantConnect/Lean/pull/1388/commits/c5da472abef61f898d85a99a41ac6a0f704988af

PR #1388, opened 2017-12-12
Changes SecurityCache.AddData to not save suspicious tick

AlexCatarino:
When Tick data is suspicious, do not save it. It will prevent orders to be filled with suspicious data.
Ref. #1023


mchandschuh:
We spoke about this earlier, but for posterity -- I think it's best we remove suspicious ticks from the list of security update data in `TimeSlice.Create` -- this should then flow through the rest of the system.

https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/TimeSlice.cs#L225

PR #1389, opened 2017-12-12
Add new scheduler DateRules

StefanoRaggi:
- `MonthEnd()`: fires on the last day of each month
- `MonthEnd(Symbol)`: fires on the last tradeable date for the specified symbol of each month
- `WeekStart()`: fires on Monday each week
- `WeekStart(Symbol)`: fires on the first tradeable date for the specified symbol of each week
- `WeekEnd()`: fires on Friday each week
- `WeekEnd(Symbol)`: fires on the last tradeable date for the specified symbol of each week

### DIFF
--- a/Common/Scheduling/DateRules.cs
+++ b/Common/Scheduling/DateRules.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -92,7 +92,7 @@ namespace QuantConnect.Scheduling
         public IDateRule EveryDay(Symbol symbol)
         {
             var security = GetSecurity(symbol);
-            return new FuncDateRule(symbol.ToString() + ": EveryDay", (start, end) => Time.EachTradeableDay(security, start, end));
+            return new FuncDateRule(symbol.Value + ": EveryDay", (start, end) => Time.EachTradeableDay(security, start, end));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Common/Scheduling/DateRules.cs
```diff
@@ -92,7 +92,7 @@ public IDateRule EveryDay()
         public IDateRule EveryDay(Symbol symbol)
         {
             var security = GetSecurity(symbol);
-            return new FuncDateRule(symbol.ToString() + ": EveryDay", (start, end) => Time.EachTradeableDay(security, start, end));
+            return new FuncDateRule(symbol.Value + ": EveryDay", (start, end) => Time.EachTradeableDay(security, start, end));
```
mchandschuh: I know outputting SID's isn't the prettiest thing, but a nice property is that they don't change. Scheduled events can be added but they can also be removed, and in order to removed them you need to know the name of the event. If we create the event and then some time later the symbol's ticker changes and we try to remove it we won't be able to find it.

Was this change purely for aesthetics?

Common/Scheduling/DateRules.cs
```diff
@@ -108,12 +108,75 @@ public IDateRule MonthStart()
...
+        public IDateRule MonthEnd(Symbol symbol)
+        {
+            return new FuncDateRule(symbol.Value + ": MonthEnd", (start, end) => MonthEndIterator(GetSecurity(symbol), start, end));
+        }
+
+        /// <summary>
+        /// Specifies an event should fire on Monday each week
+        /// </summary>
+        /// <returns>A date rule that fires on Monday each week</returns>
+        public IDateRule WeekStart()
```
mchandschuh: Why wouldn't we just use `Every(DayOfWeek.Monday)` ?

Common/Scheduling/DateRules.cs
```diff
@@ -108,12 +108,75 @@ public IDateRule MonthStart()
...
+        /// <summary>
+        /// Specifies an event should fire on the first tradeable date for the specified
+        /// symbol of each week
+        /// </summary>
+        /// <param name="symbol">The symbol whose exchange is used to determine the first
+        /// tradeable date of the week</param>
+        /// <returns>A date rule that fires on the first tradeable date for the specified security each week</returns>
+        public IDateRule WeekStart(Symbol symbol)
+        {
+            return new FuncDateRule(symbol.Value + ": WeekStart", (start, end) => WeekStartIterator(GetSecurity(symbol), start, end));
```
mchandschuh: This is great to have!

Common/Scheduling/DateRules.cs
```diff
@@ -161,5 +224,88 @@ private static IEnumerable<DateTime> MonthStartIterator(Security security, DateT
...
+                    }
+                    else
+                    {
+                        // find previous date when market is open
+                        var currentDate = date;
+                        while (!security.Exchange.Hours.IsDateOpen(currentDate))
+                        {
+                            currentDate = currentDate.AddDays(-1);
+                        }
+                        yield return currentDate;
```
mchandschuh: Awesome stuff!!

Common/Scheduling/DateRules.cs
StefanoRaggi: When possible, especially in logging, I often replace `symbol.ToString()` with `symbol.Value`.

In this case, I thought it wouldn't be an issue: 
removing an event with `ScheduleManager.Remove` from an algorithm only requires passing `scheduledEvent.Name` (returned by `ScheduleManager.On`), instead of recreating the same name generated when the event was created.

Common/Scheduling/DateRules.cs
StefanoRaggi: Or maybe better leave it as is, to keep the new `WeekStart` event name ?
It's a one-liner in both cases :)

jaredbroad: Week Start could be tuesday if Monday is a holiday

mchandschuh: This method is security agnostic, so it will fire every monday. RE: new name -- I dig it :)

Common/Scheduling/DateRules.cs
mchandschuh: I suppose it's not a big issue since we do return the scheduled event instance, so if they wanted to remove it later they could save off the instance or the name when they create it.

mchandschuh: I would recommend against updating logs to use `symbol.Value` -- for a pretty good reason why, check out `UniverseSelectionRegressionAlgorithm` -- where we have multiple GOOGs back to back that are different securities. Without the full identifier, it can be hard to tell what is what.

PR #1390, opened 2017-12-12
Sets Equity data to raw in option algorithms

AlexCatarino:
In the options trading, the strike price, the options settlement and exercise are all based on the raw price of the underlying asset instead of the adjusted price.  In order to select the accurate contracts, we need to set the data normalization mode of the underlying asset to be raw.

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -530,6 +530,10 @@ namespace QuantConnect.Algorithm
                 // all based on the raw price of the underlying asset instead of the adjusted price. 
                 // In order to select the accurate contracts, we need to set
                 // the data normalization mode of the underlying asset to be raw
+                else if (equity.DataNormalizationMode != DataNormalizationMode.Raw)
+                {
+                    Debug($"Warning: The {underlying.ToString()} equity security was set the raw price normalization mode to work with options.");
+                }
                 equity.SetDataNormalizationMode(DataNormalizationMode.Raw);
 
                 // set the underlying property on the option chain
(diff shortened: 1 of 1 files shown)


Algorithm/QCAlgorithm.cs
```diff
@@ -526,6 +526,11 @@ public void PostInitialize()
                     // if it wasn't manually added, add a subscription for underlying updates
                     equity = AddEquity(underlying.Value, option.Resolution, underlying.ID.Market, false);
                 }
+                // In the options trading, the strike price, the options settlement and exercise are
+                // all based on the raw price of the underlying asset instead of the adjusted price. 
+                // In order to select the accurate contracts, we need to set
+                // the data normalization mode of the underlying asset to be raw
+                equity.SetDataNormalizationMode(DataNormalizationMode.Raw);
```
jaredbroad: If the equity isn't already raw mode -> please send a Debug/Warning message notifying the user of this behaviour. 
"The AAPL equity security was set the raw price normalization mode to work with options."

AlexCatarino:
The regression tests didn't change. I was surprise by it, since we are not setting manually to `Raw` in all algorithms.

StefanoRaggi:
@mchandschuh Maybe the regression tests didn't change because of this line ?
https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityManager.cs#L383

mchandschuh:
@StefanoRaggi  - I think that line would only affect the option security itself and not the underlying equity security?

AlexCatarino:
@StefanoRaggi After this PR, will we need [that line](https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityManager.cs#L383)?

StefanoRaggi:
@AlexCatarino I think it is needed, so algorithms don't need to set Raw mode on option securities ?

PR #1394, opened 2017-12-13
Fix option assignment simulation in backtesting

StefanoRaggi:
The option assignment simulator used in backtesting was firing assignment events multiple times before expiration, but the actual assignment was not being performed until the expiration date.

The order messages for option assignment and expiration were inverted and have also been corrected.

### DIFF
--- a/Brokerages/Backtesting/BacktestingBrokerage.cs
+++ b/Brokerages/Backtesting/BacktestingBrokerage.cs
@@ -37,6 +37,7 @@ namespace QuantConnect.Brokerages.Backtesting
         private bool _needsScan;
         private readonly ConcurrentDictionary<int, Order> _pending;
         private readonly object _needsScanLock = new object();
+        private readonly HashSet<Symbol> _pendingOptionAssignments = new HashSet<Symbol>();
 
         /// <summary>
         /// This is the algorithm under test
@@ -387,8 +388,7 @@ namespace QuantConnect.Brokerages.Backtesting
         public void SimulateMarket()
         {
             // if simulator is installed, we run it
-            if (MarketSimulation != null)
-                MarketSimulation.SimulateMarketConditions(this, Algorithm);
+            MarketSimulation?.SimulateMarketConditions(this, Algorithm);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Engine/AlgorithmManager.cs
```diff
@@ -955,7 +955,7 @@ private static void ProcessDelistedSymbols(IAlgorithm algorithm, List<Delisting>
                     }
                     else
                     {
-                        var message = option.GetPayOff(option.Underlying.Price) < 0
+                        var message = option.GetPayOff(option.Underlying.Price) > 0
```
mchandschuh: Whoa, was this the actual bug here?

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -398,19 +398,29 @@ public void SimulateMarket()
...
+            Log.Trace($"BacktestingBrokerage.ActivateOptionAssignment(): OrderId: {ticket.OrderId}");
+        }
+
+        /// <summary>
+        /// Event invocator for the OrderFilled event
+        /// </summary>
+        /// <param name="e">The OrderEvent</param>
+        protected override void OnOrderEvent(OrderEvent e)
+        {
+            if (e.Status == OrderStatus.Filled && _pendingOptionAssignments.Contains(e.Symbol))
```
mchandschuh: Should we also be checking for canceled and such? Perhaps using e.Status.IsClosed() is more inclusive?

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -398,19 +398,29 @@ public void SimulateMarket()
...
+            // do not process the same assignment more than once
+            if (_pendingOptionAssignments.Contains(option.Symbol)) return;
 
-            foreach (var fill in fills)
-            {
-                // processing assignment in the portfolio first
-                portfolioModel.ProcessFill(Algorithm.Portfolio, option, fill);
+            _pendingOptionAssignments.Add(option.Symbol);
+
+            var request = new SubmitOrderRequest(OrderType.OptionExercise, option.Type, option.Symbol, -quantity, 0m, 0m, Algorithm.UtcTime, "Simulated option assignment before expiration");
```
mchandschuh: Nice! I love the tags

Engine/AlgorithmManager.cs
StefanoRaggi: Not the real bug, only wrong message shown, I introduced it in my previous PR :)

Brokerages/Backtesting/BacktestingBrokerage.cs
StefanoRaggi: This check was only added to prevent `_pendingOptionAssignments` from growing indefinitely, but you're right, the assignment could fail for insufficient margin and the order status could be `Invalid`. Agree `e.Status.IsClosed()` is a better choice.

PR #1395, opened 2017-12-13
Refactors options algorithms

AlexCatarino:
AddEquity and SetDataNormalization calls were removed because they are not mandatory since PostInitialize calls them.

### DIFF
--- a/Algorithm.CSharp/BasicTemplateOptionStrategyAlgorithm.cs
+++ b/Algorithm.CSharp/BasicTemplateOptionStrategyAlgorithm.cs
@@ -34,9 +34,7 @@ namespace QuantConnect.Algorithm.CSharp
     /// <meta name="tag" content="filter selection" />
     public class BasicTemplateOptionStrategyAlgorithm : QCAlgorithm
     {
-        private const string UnderlyingTicker = "GOOG";
-        public readonly Symbol Underlying = QuantConnect.Symbol.Create(UnderlyingTicker, SecurityType.Equity, Market.USA);
-        public readonly Symbol OptionSymbol = QuantConnect.Symbol.Create(UnderlyingTicker, SecurityType.Option, Market.USA);
+        private Symbol _optionSymbol;
 
         public override void Initialize()
         {
@@ -44,16 +42,14 @@ namespace QuantConnect.Algorithm.CSharp
             SetEndDate(2015, 12, 24);
             SetCash(1000000);
 
-            var equity = AddEquity(UnderlyingTicker);
-            var option = AddOption(UnderlyingTicker);
-
-            equity.SetDataNormalizationMode(DataNormalizationMode.Raw);
+            var option = AddOption("GOOG");
+            _optionSymbol = option.Symbol;
 
             // set our strike/expiry filter for this option chain
             option.SetFilter(-2, +2, TimeSpan.Zero, TimeSpan.FromDays(180));
 
             // use the underlying equity as the benchmark
-            SetBenchmark(equity.Symbol);
+            SetBenchmark("GOOG");
         }
 
         public override void OnData(Slice slice)
@@ -61,7 +57,7 @@ namespace QuantConnect.Algorithm.CSharp
             if (!Portfolio.Invested)
             {
                 OptionChain chain;
-                if (slice.OptionChains.TryGetValue(OptionSymbol, out chain))
+                if (slice.OptionChains.TryGetValue(_optionSymbol, out chain))
                 {
                     var atmStraddle = chain
                         .OrderBy(x => Math.Abs(chain.Underlying.Price - x.Strike))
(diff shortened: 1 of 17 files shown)


Algorithm.Python/BasicTemplateOptionTradesAlgorithm.py
```diff
@@ -21,51 +21,47 @@
...
-        self.SetBenchmark(equity.Symbol)
+        self.SetBenchmark("GOOG")
 
 
     def OnData(self,slice):
         if not self.Portfolio.Invested:
             for kvp in slice.OptionChains:
                 chain = kvp.Value
                 # find the second call strike under market price expiring today
-                contracts = sorted(sorted(chain, key = lambda x: abs(chain.Underlying.Price - x.Strike)),
```
jingwu74: In order to find the most at-the-money contract, should we sort the chain by expiry first then by the absolute difference between underlying price and the strike?

Algorithm.CSharp/BasicTemplateOptionStrategyAlgorithm.cs
```diff
@@ -34,34 +34,30 @@ namespace QuantConnect.Algorithm.CSharp
...
-            equity.SetDataNormalizationMode(DataNormalizationMode.Raw);
+            var option = AddOption("GOOG");
+            _optionSymbol = option.Symbol;
 
             // set our strike/expiry filter for this option chain
             option.SetFilter(-2, +2, TimeSpan.Zero, TimeSpan.FromDays(180));
 
             // use the underlying equity as the benchmark
-            SetBenchmark(equity.Symbol);
+            SetBenchmark("GOOG");
```
mchandschuh: Does this work? I half expect the `SymbolCache` to resolve `"GOOG"` to the canonical option symbol and not the equity symbol. Please confirm by stepping through to see how it gets resolved and let me know :)

Algorithm.CSharp/BasicTemplateOptionsHistoryAlgorithm.cs
```diff
@@ -32,25 +31,19 @@ namespace QuantConnect.Algorithm.CSharp
...
             SetCash(1000000);
 
-            var equity = AddEquity(UnderlyingTicker);
-            var option = AddOption(UnderlyingTicker);
+            var option = AddOption("GOOG");
 
-            equity.SetDataNormalizationMode(DataNormalizationMode.Raw);
             option.PriceModel = OptionPriceModels.CrankNicolsonFD();
-            option.SetFilter(-2, +2, TimeSpan.FromDays(00), TimeSpan.FromDays(180));
+            option.SetFilter(-2, +2, TimeSpan.Zero, TimeSpan.FromDays(180));
```
mchandschuh: Cool, maybe we should have an overload which assumes `TimeSpan.Zero` for the earliest expiry? No need to change here, just thinking out loud.

Algorithm.CSharp/OptionRenameRegressionAlgorithm.cs
```diff
@@ -112,18 +107,5 @@ public override void OnData(Slice slice)
                 }
             }
         }
-
-        /// <summary>
-        /// Order fill event handler. On an order fill update the resulting information is passed to this method.
-        /// </summary>
-        /// <param name="orderEvent">Order event details containing details of the evemts</param>
-        /// <remarks>This method can be called asynchronously and so should only be used by seasoned C# experts. Ensure you use proper locks on thread-unsafe objects</remarks>
-        public override void OnOrderEvent(OrderEvent orderEvent)
```
mchandschuh: Any particular reason why we're removing this logging? I'm not strongly opinionated one way or another, just curious as to the reasoning.

Algorithm.CSharp/BasicTemplateOptionStrategyAlgorithm.cs
AlexCatarino: This works. Since the equity doesn't exist yet, it is created in SetBenchmark and refers to it.

Algorithm.CSharp/BasicTemplateOptionsHistoryAlgorithm.cs
AlexCatarino: Good point. I think that most of people only changes the latest expiry.

Algorithm.CSharp/OptionRenameRegressionAlgorithm.cs
AlexCatarino: I remove it, because there are plenty of those in other examples and this algorithm is a regression one.

Algorithm.Python/BasicTemplateOptionTradesAlgorithm.py
AlexCatarino: @mchandschuh I asked @jingwu74 to check whether the change would make a difference, she test it and didn't find it necessary.

Algorithm.CSharp/OptionRenameRegressionAlgorithm.cs
mchandschuh: I think it's nice to have the order event logging in regression algorithms. Consider the regression test fails, the first thing I would want to do is compare the trades and security additions/removals to get a sense of *why* it failed.

PR #1396, opened 2017-12-13
Add More Crypto Markets to Market Class

andrewhart098:
These changes add the 9 more crypto markets to the market enum. This will enable the KaikoDataConverter to process the data from these markets. These changes do not add Brokerage functionality for the additional markets.

In addition, the KaikoDataConverter now checks for scientific notation on all quantity and price fields. If the KaikoDataConverter encounters a poorly formatted row off raw data, it skips that single row and continues with the conversion process.

### DIFF
--- a/ToolBox/KaikoDataConverter/Program.cs
+++ b/ToolBox/KaikoDataConverter/Program.cs
@@ -1,11 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using NodaTime;
 using QuantConnect.Data;
-using QuantConnect.Data.Consolidators;
 using QuantConnect.Data.Market;
+using QuantConnect.Logging;
 using QuantConnect.Util;
 
 namespace QuantConnect.ToolBox.KaikoDataConverter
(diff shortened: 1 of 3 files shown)


ToolBox/KaikoDataConverter/Program.cs
```diff
@@ -306,17 +336,18 @@ private static IEnumerable<Tick> ParseKaikoTradeFile(Symbol symbol, string unzip
...
-        private static decimal ParseQuantity(string[] lineParts, int quantityColumn)
+        private static decimal ParseScientificNotationToDecimal(string[] lineParts, int column)
         {
-            var quantity = lineParts[quantityColumn];
-            if (quantity.Contains("e"))
+            var value = lineParts[column];
+            if (value.Contains("e"))
             {
-                return Decimal.Parse(quantity, System.Globalization.NumberStyles.Float);
+                return Decimal.Parse(value, System.Globalization.NumberStyles.Float);
```
StefanoRaggi: I would add the `CultureInfo.InvariantCulture` argument on this line for non-US users

ToolBox/KaikoDataConverter/Program.cs
```diff
@@ -306,17 +336,18 @@ private static IEnumerable<Tick> ParseKaikoTradeFile(Symbol symbol, string unzip
...
             }
 
-            return Convert.ToDecimal(lineParts[quantityColumn]);
+            return Convert.ToDecimal(lineParts[column]);
```
StefanoRaggi: Same here

PR #1397, opened 2017-12-13
Fix GDAX submit order errors stopping algorithms

StefanoRaggi:
The expected behavior for rejected orders is to log/display the error and allow the algorithm to continue execution. The GDAX brokerage implementation was incorrectly sending a runtime error, causing the algorithm to terminate.

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.cs
@@ -26,7 +26,6 @@ namespace QuantConnect.Brokerages.GDAX
 {
     public partial class GDAXBrokerage : BaseWebsocketsBrokerage
     {
-
         #region IBrokerage
         /// <summary>
         /// Checks if the websocket connection is connected or in the process of connecting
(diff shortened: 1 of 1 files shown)


Brokerages/GDAX/GDAXBrokerage.cs
```diff
@@ -77,16 +76,16 @@ public override bool PlaceOrder(Order order)
 
                 if (raw?.Id == null)
                 {
-                    OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Error, (int)response.StatusCode, "GDAXBrokerage.PlaceOrder: Error parsing response from place order: " + response.Content));
+                    OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Warning, (int)response.StatusCode, "GDAXBrokerage.PlaceOrder: Error parsing response from place order: " + response.Content));
                     UnlockStream();
-                    return false;
+                    return true;
```
mchandschuh: Should we be returning true here? I think we should still be returning false so the transaction handle invalidates the QC order.
image

mchandschuh: Maybe not this exact case since we couldn't get an ID, but the 'rejected' case below should certainly return `false`, yes/no?

StefanoRaggi: Hmm, if we return `false` in these cases, the order status will be correctly set to `Invalid`, but the algorithm will also be terminated, due to the `_algorithm.Error` call above -- which is the issue we are trying to solve in this PR :)
Looking better at the code, I should probably still return `true` but add a call this line in all error cases (like in the last error case at the end of the method) ?
```
 OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, 0, "GDAX Order Event") { Status = OrderStatus.Invalid });
```
This line also updates the order status

mchandschuh:
I didn't notice the sending of the invalid order event before the return true. These changes look good.

PR #1399, opened 2017-12-15
Disable order log tracing when backtesting

StefanoRaggi:
These logging instructions were only meant to be executed in live mode.


StefanoRaggi:
I previously added those log traces so that PaperBrokerage would be consistent with all other brokerages. I forgot that in with frequent trades this could have a small impact on backtesting speed.

PR #1401, opened 2017-12-16
Update python Framework example

DMTSource:
Updating to match the current c# example. I backtested in QC to ensure it ran:

https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_68eb80ea8a205ace27aaf13e8de1b6a2.html


mchandschuh:
Looks like I missed some cases during the massive rename :)

DMTSource:
Force of habit with the equals signs... and also sorry about the messy commits Alex!

PR #1403, opened 2017-12-17
Expose GDAX minimum order sizes

mchandschuh:
This data is extremely useful for algorithms. I think the long term solution is to actually expose this data through the symbol properties database, but for now this will suffice :)


mchandschuh:
Minimum order sizes should be moved to `symbol-properties-database.csv`

PR #1409, opened 2017-12-18
Fix race condition in SubscriptionCollection.GetEnumerator

StefanoRaggi:
An occasional `NullReferenceException` was thrown when enumerating and concurrently adding or removing items.

A new unit test that always throws this exception was also added.

### DIFF
--- a/Engine/DataFeeds/SubscriptionCollection.cs
+++ b/Engine/DataFeeds/SubscriptionCollection.cs
@@ -163,12 +163,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// </returns>
         public IEnumerator<Subscription> GetEnumerator()
         {
-            foreach (var subscriptionsBySymbol in _subscriptions)
+            foreach (var subscriptionsByConfig in _subscriptions
+                .Select(x => x.Value))
             {
-                var subscriptionsByConfig = subscriptionsBySymbol.Value;
-                foreach (var kvp in subscriptionsByConfig.OrderBy(x => x.Key.TickType))
+                foreach (var subscription in subscriptionsByConfig
+                    .Select(x => x.Value)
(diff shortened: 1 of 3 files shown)


Engine/DataFeeds/SubscriptionCollection.cs
```diff
@@ -163,12 +163,13 @@ public bool TryRemoveAll(Symbol symbol, out ICollection<Subscription> subscripti
         /// </returns>
         public IEnumerator<Subscription> GetEnumerator()
         {
-            foreach (var subscriptionsBySymbol in _subscriptions)
+            foreach (var subscriptionsByConfig in _subscriptions
+                .Select(x => x.Value))
```
mchandschuh: I wonder if using the `.Value` property is safe. The code before and this code look so similar to me that we may have just decreased the odds of the bad runner winning the race. By using `.Value` we'd remove the access of `kvp.Value` in the `Select` statements. Also, after looking through decompiled code, the `.Values` property on a `ConcurrencDictionary` implements a thread-safe enumerator as well, will volatile reads and such, and should guarantee all non-null values.

StefanoRaggi: After reading this article I was convinced that the `.Values` property should be avoided:
https://arbel.net/2013/02/03/best-practices-for-using-concurrentdictionary/

Also the source for `ConcurrentDictionary` shows that it calls `AcquireAllLocks` and this locking is also seen when inspecting the output of the Concurrency Visualizer extension for VS.

mchandschuh: Great article. Years ago I would always be a proponent of enumerating the kvp's VS the keys or the values and as of recently I forgot *why* -- article nicely describes why usage of `ConcurrentDictionary.Keys` and `ConcurrentDictionary.Values` should be avoided like the plague!

PR #1411, opened 2017-12-18
Emit order canceled events from GDAX brokerage

mchandschuh:
Since we can't correlate these events via async responses from the client, we'll just need to fire these off synchronously as the request comes in and is successful.

This is my attempt at resolving the GDAX cancel order event issue. I'm open to feedback and other suggestions to make this as robust as possible.

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.cs
@@ -27,7 +27,6 @@ namespace QuantConnect.Brokerages.GDAX
 {
     public partial class GDAXBrokerage : BaseWebsocketsBrokerage
     {
-
         #region IBrokerage
         /// <summary>
         /// Checks if the websocket connection is connected or in the process of connecting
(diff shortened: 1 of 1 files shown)


jaredbroad:
Does this suggest missing tests from the IBrokerage test suite?

Brokerages/GDAX/GDAXBrokerage.cs
```diff
@@ -143,7 +144,11 @@ public override bool CancelOrder(Order order)
                 var req = new RestRequest("/orders/" + id, Method.DELETE);
                 GetAuthenticationToken(req);
                 var response = RestClient.Execute(req);
-                success.Add(response.StatusCode == System.Net.HttpStatusCode.OK);
+                success.Add(response.StatusCode == HttpStatusCode.OK);
+                if (response.StatusCode == HttpStatusCode.OK)
+                {
+                    OnOrderEvent(new OrderEvent(order, LastHeartbeatUtcTime, 0, order.Tag));
```
StefanoRaggi: Doesn't this fire an order event with the current status  -- `OrderStatus.CancelPending` ?

This line should pass the expected status:
```
OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, 0, "GDAX Order Event") { Status = OrderStatus.Canceled });
```

mchandschuh:
RE: missing tests
Absolutely. Although this guy is a little different in that he doesn't allow for updating orders which is also a big part of the brokerage test suite. Agreed we should take another look at that test suite.

PR #1416, opened 2017-12-19
Speed up SecurityIdentifier parsing from string

StefanoRaggi:
With this update there is a **3x** speedup in `SecurityIdentifier.TryParse`.

This method is heavily used in coarse universe selection, when parsing coarse data files.


Common/SecurityIdentifier.cs
```diff
@@ -472,8 +472,10 @@ private static ulong DecodeBase36(string symbol)
                     ? c - '0'
                     : c - 'A' + 10);
 
-                result += value * Pow(36, pos++);
+                result += baseValue * value;
```
jaredbroad: Nice idea

mchandschuh: Awesome!

PR #1417, opened 2017-12-19
Alpha total asset breakdown pie chart

mchandschuh:
Add `SeriesType.Pie` -- result handlers only save last data point in each series
Add 'Alpha Assets' pie chart for total population asset breakdown over life of algorithm
Use `Formatting.Indented` when serializing results to disk in `BaseResultHandler` (makes life easier for me :) )
Fixes bug in reporting finalized alpha scores

### DIFF
--- a/Engine/Results/LiveTradingResultHandler.cs
+++ b/Engine/Results/LiveTradingResultHandler.cs
@@ -697,13 +697,31 @@ namespace QuantConnect.Lean.Engine.Results
                     foreach (var series in update.Series.Values)
                     {
                         //If we don't already have this record, its the first packet
-                        if (!Charts[update.Name].Series.ContainsKey(series.Name))
+                        var chart = Charts[update.Name];
+                        if (!chart.Series.ContainsKey(series.Name))
                         {
-                            Charts[update.Name].Series.Add(series.Name, new Series(series.Name, series.SeriesType, series.Index, series.Unit));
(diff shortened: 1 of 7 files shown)


mchandschuh:
This actually still has a bug in it... I have a solution, just confirming the fix now.

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -697,13 +697,31 @@ public void SampleRange(List<Chart> updates)
...
                     {
                         //If we don't already have this record, its the first packet
-                        if (!Charts[update.Name].Series.ContainsKey(series.Name))
+                        var chart = Charts[update.Name];
+                        if (!chart.Series.ContainsKey(series.Name))
                         {
-                            Charts[update.Name].Series.Add(series.Name, new Series(series.Name, series.SeriesType, series.Index, series.Unit));
+                            chart.Series.Add(series.Name, new Series(series.Name, series.SeriesType, series.Index, series.Unit)
+                            {
+                                Color = series.Color, ScatterMarkerSymbol = series.ScatterMarkerSymbol
```
mchandschuh: Added this here to match backtesting -- @jaredbroad -- please confirm that this wasn't intentionally left off of here.

jaredbroad: Nice catch ty

PR #1420, opened 2017-12-20
Provide better default margin model for GDAX cash accounts

mchandschuh:
> READY TO BE REVIEWED/MERGED
---
> NOTE: FractionalQuantityRegressionAlgorithm is failing due to recent changes in minimum order sizes. This PR has no ill effect on the regression suit.
---

#### GDAXBrokerageModel.GetMarginModel -> NonShortableCashAccountMarginModel

GDAX cash accounts aren't allowed to short and require sufficient quantity of quote currency
to purchase and sufficient quantity of base currency to sell. This model aims to enforce
these constraints

#### Add IBrokerageModel.GetMarginModel and invoke from BrokerageModelSecurityInitializer

It seems all the other models are being provided via the brokerage model but all the margin models
are assigned via the security constructors. This allows us to set margin models using the same
pattern that we used for overriding default margin models. This becomes particularly important
when considering how different exchanges/brokerages manage cash and margin in the account.

#### Add NonShortableCashAccountMarginModel

This margin model forces us to purchase everthing outright with cash.
This also means there's no maintenance margin. This also means that in order
to buy/sell currency pairs we require the quote currency to buy and the base
currency to sell. For example, you can't purchase LTCBTC with USD, and likewise
you can't sell LTCBTC without having LTC to sell.

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerageFactory.cs
+++ b/Brokerages/GDAX/GDAXBrokerageFactory.cs
@@ -27,6 +27,8 @@ namespace QuantConnect.Brokerages.GDAX
     /// </summary>
     public class GDAXBrokerageFactory : BrokerageFactory
     {
+        private IAlgorithm _algorithm;
+
         /// <summary>
         /// Factory constructor
         /// </summary>
@@ -53,9 +55,10 @@ namespace QuantConnect.Brokerages.GDAX
         };
 
         /// <summary>
-        /// The brokerage model
+        /// The brokerage model. This uses the alorithm instance passed in <see cref="CreateBrokerage"/>
+        /// in order to properly populate the model with a reference to the algorithm's cashbook
         /// </summary>
-        public override IBrokerageModel BrokerageModel => new GDAXBrokerageModel();
+        public override IBrokerageModel BrokerageModel => new GDAXBrokerageModel(_algorithm?.Portfolio?.CashBook);
 
         /// <summary>
         /// Create the Brokerage instance
--- a/Brokerages/GDAX/GDAXBrokerageFactory.cs
+++ b/Brokerages/GDAX/GDAXBrokerageFactory.cs
@@ -27,8 +27,6 @@ namespace QuantConnect.Brokerages.GDAX
     /// </summary>
     public class GDAXBrokerageFactory : BrokerageFactory
     {
-        private IAlgorithm _algorithm;
-
         /// <summary>
         /// Factory constructor
         /// </summary>
@@ -58,7 +56,7 @@ namespace QuantConnect.Brokerages.GDAX
         /// The brokerage model. This uses the alorithm instance passed in <see cref="CreateBrokerage"/>
         /// in order to properly populate the model with a reference to the algorithm's cashbook
         /// </summary>
-        public override IBrokerageModel BrokerageModel => new GDAXBrokerageModel(_algorithm?.Portfolio?.CashBook);
+        public override IBrokerageModel BrokerageModel => new GDAXBrokerageModel();
 
         /// <summary>
         /// Create the Brokerage instance
--- a/Common/Brokerages/GDAXBrokerageModel.cs
+++ b/Common/Brokerages/GDAXBrokerageModel.cs
@@ -166,7 +166,7 @@ namespace QuantConnect.Brokerages
         }
 
         /// <summary>
-        /// Gets a new margin model for the security. For cash accounts the <see cref="NonShortableCashAccountMarginModel"/>
+        /// Gets a new margin model for the security. For cash accounts the <see cref="CashAccountMarginModel"/>
         /// is returned, while margin accounts will return the default margin model
         /// </summary>
         /// <param name="security">The security to get a margin model for</param>
@@ -176,7 +176,7 @@ namespace QuantConnect.Brokerages
         {
             if (accountType == AccountType.Cash)
             {
-                return new NonShortableCashAccountMarginModel(_cashBook);
+                return new CashAccountMarginModel(_cashBook);
             }
 
             return base.GetMarginModel(security, accountType);
--- a/Common/Brokerages/GDAXBrokerageModel.cs
+++ b/Common/Brokerages/GDAXBrokerageModel.cs
@@ -28,7 +28,8 @@ namespace QuantConnect.Brokerages
     /// </summary>
     public class GDAXBrokerageModel : DefaultBrokerageModel
     {
-        private static BrokerageMessageEvent _message = new BrokerageMessageEvent(BrokerageMessageType.Warning, 0, "Brokerage does not support update. You must cancel and re-create instead.");
+        private readonly CashBook _cashBook;
+        private static readonly BrokerageMessageEvent UpdateNotSupportedErrorMessage = new BrokerageMessageEvent(BrokerageMessageType.Warning, 0, "Brokerage does not support update. You must cancel and re-create instead.");
 
         // https://support.gdax.com/customer/portal/articles/2725970-trading-rules
         private static readonly Dictionary<string, decimal> MinimumOrderSizes = new Dictionary<string, decimal>
@@ -57,9 +58,10 @@ namespace QuantConnect.Brokerages
         /// </summary>
         /// <param name="accountType">The type of account to be modelled, defaults to
         /// <see cref="AccountType.Margin"/></param>
-        public GDAXBrokerageModel(AccountType accountType = AccountType.Margin)
+        public GDAXBrokerageModel(CashBook cashBook, AccountType accountType = AccountType.Margin)
             : base(accountType)
         {
+            _cashBook = cashBook;
             if (accountType == AccountType.Margin)
             {
                 new BrokerageMessageEvent(BrokerageMessageType.Warning, 0,
@@ -102,7 +104,7 @@ namespace QuantConnect.Brokerages
         /// <returns></returns>
         public override bool CanUpdateOrder(Security security, Order order, UpdateOrderRequest request, out BrokerageMessageEvent message)
         {
-            message = _message;
+            message = UpdateNotSupportedErrorMessage;
             return false;
         }
 
@@ -117,7 +119,7 @@ namespace QuantConnect.Brokerages
         {
             if (order.BrokerId != null && order.BrokerId.Any())
             {
-                message = _message;
+                message = UpdateNotSupportedErrorMessage;
                 return false;
             }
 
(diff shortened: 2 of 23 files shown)


Common/Securities/NonShortableCashAccountMarginModel.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <param name="direction">The direction of the trade</param>
+        /// <returns>The margin available for the trade</returns>
+        public decimal GetMarginRemaining(SecurityPortfolioManager portfolio, Security security, OrderDirection direction)
+        {
+            switch (direction)
+            {
+                case OrderDirection.Hold:
+                case OrderDirection.Buy:
+                    // increasing position, purchasing in units of the quote currency
+                    return security.QuoteCurrency.Amount;
```
mchandschuh: I think the only thing missing from this `GetMarginRemaining` implementation is checking open orders. This is kind of tricky, since we need to check for orders not only for `security` but also for all other securities that have the currency while keeping in mind the side of the trade.

Common/Securities/NonShortableCashAccountMarginModel.cs
```diff
@@ -0,0 +1,158 @@
...
+using QuantConnect.Orders;
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="ISecurityMarginModel"/> that requires cash of the correct
+    /// currency to exist in the portfolio to make trades. This also only applies to securities where shorting
+    /// is not an option. IOW, holdings.Quantity >= 0 at all times.
+    /// </summary>
+    public class NonShortableCashAccountMarginModel : ISecurityMarginModel
```
mchandschuh: Perhaps this should just be `CashAccountMarginModel` -- Do any cash accounts allow you to naked short an asset?

Brokerages/GDAX/GDAXBrokerageFactory.cs
```diff
@@ -53,9 +55,10 @@ public override void Dispose()
         };
 
         /// <summary>
-        /// The brokerage model
+        /// The brokerage model. This uses the alorithm instance passed in <see cref="CreateBrokerage"/>
+        /// in order to properly populate the model with a reference to the algorithm's cashbook
         /// </summary>
-        public override IBrokerageModel BrokerageModel => new GDAXBrokerageModel();
+        public override IBrokerageModel BrokerageModel => new GDAXBrokerageModel(_algorithm?.Portfolio?.CashBook);
```
mchandschuh: This instance will be used in the event the user doesn't manually invoke `SetBrokerageModel` -- Since GDAX disabled margin, I think we should include `AccountType.Cash` here

Common/Brokerages/GDAXBrokerageModel.cs
```diff
@@ -162,5 +164,22 @@ public override IFillModel GetFillModel(Security security)
...
+        /// <param name="accountType">The account type</param>
+        /// <returns>The margin model for this brokerage/security</returns>
+        public override ISecurityMarginModel GetMarginModel(Security security, AccountType accountType)
+        {
+            if (accountType == AccountType.Cash)
+            {
+                return new CashAccountMarginModel(_cashBook);
+            }
+
+            return base.GetMarginModel(security, accountType);
```
mchandschuh: Maybe a message to the user saying margin accounts aren't supported by GDAX?

Common/Securities/CashAccountMarginModel.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// <summary>
+        /// Cash accounts require the full value of the order plus any fees
+        /// </summary>
+        /// <param name="security"></param>
+        /// <param name="order">The order to be executed</param>
+        /// <returns>The total margin in terms of the currency quoted in the order</returns>
+        public decimal GetInitialMarginRequiredForOrder(Security security, Order order)
+        {
+            var fees = security.FeeModel.GetOrderFee(security, order);
+            var marginRequiredInAccountCurrency = order.GetValue(security) + fees;
```
mchandschuh: nit -- terminology here is incorrect, perhaps we should be phrasing it as `cashRequired` or similar

Common/Securities/CashAccountMarginModel.cs
```diff
@@ -0,0 +1,242 @@
...
+            {
+                // find the target currency for the requested direction and the securities potentially involved
+                targetCurrency = direction == OrderDirection.Buy
+                    ? security.QuoteCurrency.Symbol
+                    : baseCurrency.BaseCurrencySymbol;
+
+                foreach (var portfolioSecurity in portfolio.Securities.Values)
+                {
+                    var basePortfolioSecurity = portfolioSecurity as IBaseCurrencySymbol;
+                    if (basePortfolioSecurity == null) continue;
```
mchandschuh: I think we need to remove this check. Specifically in the case of crypto this will always be true, but in the general case it doesn't need to be. Consider an equity portfolio with assets priced in USD and other assets priced in EUR. Since they're equities, they won't have a base currency symbol, but we'll still need to potentially check them.

Maybe this condition should instead be
```csharp
if (basePortfolioSecurity == null && direction == OrderDirection.Sell) { /*...*/ }
```

The thinking here is that when we're selling we care about the base currency, so if we're selling and there is no base currency, it should be safe to skip.

Common/Securities/CashAccountMarginModel.cs
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                }
+            }
+
+            // fetch open orders with matching symbol/side
+            var openOrders = portfolio.Transactions.GetOpenOrders(x =>
+                {
+                    OrderDirection dir;
+                    return symbolDirectionPairs.Count == 0 ||
+                           symbolDirectionPairs.TryGetValue(x.Symbol, out dir) && dir == x.Direction;
```
mchandschuh: Awesome! Out of curiosity, what is the case where `symbolDirectionPairs.Count == 0` ? I'm guessing this is the equity/stock case? I think even in that case we want to have symbol direction pairs for all the stocks' quote currency?

Common/Securities/CashAccountMarginModel.cs
```diff
@@ -0,0 +1,242 @@
...
+                        case OrderType.StopLimit:
+                            orderPrice = ((StopLimitOrder) order).LimitPrice;
+                            break;
+                    }
+
+                    // convert order value to target currency
+                    decimal orderValue;
+                    if (orderSecurity.QuoteCurrency.Symbol == targetCurrency)
+                    {
+                        orderValue = orderPrice * order.Quantity;
```
mchandschuh: Why aren't we using `order.GetValue(security)` here?  Where possible, I like to use polymorphism rather than conditional branching (ifs/switches) -- polymorphism encapsulates the logic where and conditional branching leaks to logic. An blurb from [refactoring guru](https://refactoring.guru/replace-conditional-with-polymorphism) on refactoring switches into polymorphic calls and benefits/ect.

Common/Securities/CashAccountMarginModel.cs
```diff
@@ -0,0 +1,242 @@
...
+                    if (baseCurrency != null)
+                    {
+                        // we have this much 'cash' that can be sold
+                        return _cashBook[baseCurrency.BaseCurrencySymbol].Amount - Math.Abs(openOrdersTotalValue);
+                    }
+
+                    // we have this much stock value that can be sold... this is in the account currency,
+                    // the assumption being that since the security doesn't implement IBaseCurrencySymbol
+                    // that the holdings are either 'units' of stock or similar and not currency swaps/virtual positions
+                    return (security.Holdings.AbsoluteQuantity - Math.Abs(openOrdersTotalQuantity)) * security.Holdings.AveragePrice;
```
mchandschuh: This math looks wrong. We have a quantity and subtract from it cost. This can return units of quantity or value. I think the more correct thing is to return quantity. You can sell 5 units of SPY, rather than you can sell 1,000 USD of SPY. The important part is that the other methods in this implementation are consistent.

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -257,6 +257,16 @@ public List<Order> GetOpenOrders(Symbol symbol)
...
         }
 
+        /// <summary>
+        /// Get a list of open orders satisfying the filter.
+        /// </summary>
+        /// <returns>List of open orders.</returns>
+        public List<Order> GetOpenOrders(Func<Order, bool> filter)
+        {
+            filter = filter ?? (x => true);
+            return _orderProcessor.GetOrders(x => x.Status.IsOpen() && filter(x)).ToList();
```
mchandschuh: I'm weary of bloating the API. This behavior is easily accomplished via `Transactions.GetOrders`, although I suppose this is the more common use case.. algos are almost always requesting open orders.

Tests/Common/Securities/CashAccountMarginModelTests.cs
```diff
@@ -0,0 +1,326 @@
...
+
+        [Test]
+        public void EquityBuyOrderRequiresFullMargin()
+        {
+            _spy.SetMarketPrice(new Tick { Value = 250m });
+
+            var order = new MarketOrder(_spy.Symbol, 2, DateTime.UtcNow);
+            var initialMargin = _marginModel.GetInitialMarginRequiredForOrder(_spy, order);
+
+            Assert.AreEqual(501m, initialMargin);
```
mchandschuh: Comments on how you got `501` are a must in these types of tests. The tests serve as documentation, which, by the way, I do like how readable these tests are. I'm normally against running the world but it does make these tests easy to follow. I think we should just add a comment:
```csharp
// (current price * quantity) + fees = 501
```

Tests/Common/Securities/CashAccountMarginModelTests.cs
```diff
@@ -0,0 +1,326 @@
...
+        {
+            _spy = _algorithm.AddEquity("SPY");
+            _portfolio[_spy.Symbol].SetHoldings(150, 2);
+            _spy.SetMarketPrice(new Tick { Value = 250m });
+            _algorithm.SetFinishedWarmingUp();
+
+            var order = new MarketOrder(_spy.Symbol, -2, DateTime.UtcNow);
+            var initialMargin = _marginModel.GetInitialMarginRequiredForOrder(_spy, order);
+
+            Assert.AreEqual(-499m, initialMargin);
```
mchandschuh: I'm thinking that in the sell case we should be tracking units of stock (or currency/whatever). The model is really saying you can sell 2 units, and that's true even if the price moves quickly.

Same here for comment, how did you get -499?

mchandschuh:
This is being improve by @StefanoRaggi -- closing as it's now obsolete.

PR #1431, opened 2017-12-21
Fix subscription removal in SubscriptionCollection.TryRemove

StefanoRaggi:
This method was not removing the symbol dictionary entry when removing the last subscription, allowing the collection to grow excessively over time (especially with universe selection algorithms).

This PR also includes a few minor performance fixes.

### DIFF
--- a/Engine/DataFeeds/SubscriptionCollection.cs
+++ b/Engine/DataFeeds/SubscriptionCollection.cs
@@ -139,7 +139,12 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                 return false;
             }
 
-            return dictionary.Any() || _subscriptions.TryRemove(configuration.Symbol, out dictionary);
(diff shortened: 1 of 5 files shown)


Engine/DataFeeds/SubscriptionCollection.cs
```diff
@@ -133,7 +133,13 @@ public bool TryRemove(SubscriptionDataConfig configuration, out Subscription sub
...
             }
 
-            return dictionary.TryRemove(configuration, out subscription);
+            if (!dictionary.TryRemove(configuration, out subscription))
+            {
+                subscription = null;
+                return false;
+            }
+
+            return dictionary.Any() || _subscriptions.TryRemove(configuration.Symbol, out dictionary);
```
mchandschuh: I think this should still return true since the subscription was removed.

Great catch on this dictionary!

PR #1433, opened 2017-12-22
Disable automatic security seeding

StefanoRaggi:
In this PR we are disabling the default security seeding (automatically getting the last price for a security when added to the algorithm) for a couple reasons, both when using large universes:
- In live trading, these history requests are sent to a history server, potentially causing timeouts
- In backtesting, depending on the algorithm this could also cause slowdowns up to 30%

### DIFF
--- a/Tests/Common/Securities/BrokerageModelSecurityInitializerTests.cs
+++ b/Tests/Common/Securities/BrokerageModelSecurityInitializerTests.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


Tests/Common/Securities/BrokerageModelSecurityInitializerTests.cs
```diff
@@ -119,10 +118,10 @@ public void BrokerageModelSecurityInitializer_CanSetPrice_ForQuoteBar()
             _algo.SetDateTime(dateForWhichDataExist);
 
             // Act
-            _brokerageInitializer.Initialize(_quoteBarSecurity, true);
+            _brokerageInitializer.Initialize(_quoteBarSecurity);
 
             // Assert
-            Assert.IsFalse(_quoteBarSecurity.Price == 0);
+            Assert.IsTrue(_quoteBarSecurity.Price == 0);
```
mchandschuh: I would either comment why this is expected to be true or remove the entire test method.

Arguably tests pertaining to setting a security's price don't belong in this test class, but rather would belong in tests or the seeder.

Not recommending, but if we wanted to confirm that the seeder either is or is not invoked, typically a fake or stub implementation would be used to achieve this. This way the test doesn't depend on an implementation of the seeder, but just the seeder contract (interface).

Issue #1435, opened 2017-12-24
CustomChartingAlgorithm backtest fails to run with exception: chart reserved names


gwahgdy:
Tested in Algorithm Lab on cloud.  **CustomChartingAlgorithm.py** backtest fails to run with the following exception.

`2016-01-01 00:00:00 Launching analysis for e2c4c89ebe9c273f679aaa3050855fa6 with LEAN Engine v2.4.0.1.2731
2016-01-01 00:00:00 Runtime Error: Python.Runtime.PythonException: Exception : Algorithm.Plot(): 'Strategy Equity','Equity','Daily Performance','Alpha','Alpha Count','Alpha Asset Breakdown' and 'Meta' are reserved chart names created for all charts.
  at QuantConnect.Algorithm.QCAlgorithm.Plot (System.String chart, System.String series, System.Decimal value) [0x00074] in <45ea18ccece04cf3a027049fe320d8f4>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00038] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at Python.Runtime.PyObject.Invoke (Python.Runtime.PyTuple args, Python.Runtime.PyDict kw) [0x00033] in <59c711440fed482b8e57b026917e4706>:0 
  at Python.Runtime.PyObject.InvokeMethod (System.String name, Python.Runtime.PyTuple args, Python.Runtime.PyDict kw) [0x00007] in <59c711440fed482b8e57b026917e4706>:0 
  at Python.Runtime.PyObject.TryInvokeMember (System.Dynamic.InvokeMemberBinder binder, System.Object[] args, System.Object& result) [0x0003e] in <59c711440fed482b8e57b026917e4706>:0 
  at (wrapper dynamic-method) System.Object:CallSite.Target (System.Runtime.CompilerServices.Closure,System.Runtime.CompilerServices.CallSite,object,QuantConnect.Data.Slice)
[...]

PR #1437, opened 2017-12-25
Fix use of protected name in custom charting algo

DMTSource:
Switch away from use of protected-chart name which was preventing the example from running. Sorry for wrong branch name...should be something like: issue-1435/fix-custom-charting-protected
https://github.com/QuantConnect/Lean/issues/1435


jaredbroad:
Thats strange it should mean it shows up on the Strategy Equity plot. Will try repeating and see if can fix in frontend.

DMTSource:
Oh my bad, just before it was reported by @gwahgdy in #1435 I had also seen the protected chart error in a personal algo and assumed it was related to recent changes in charting/framework and therefore systematic.

PR #1442, opened 2017-12-27
Minor performance fix in SubscriptionSynchronizer.Sync

StefanoRaggi:
This method was adding unnecessarily two types of `DataFeedPacket` objects.
The first type of packet contained the coarse data just used for universe selection, the other one contained only empty packets.
Since these packets would be filtered out later anyway, there is no point in passing them to `TimeSlice.Create` which needs to enumerate them.

### DIFF
--- a/Engine/DataFeeds/SubscriptionSynchronizer.cs
+++ b/Engine/DataFeeds/SubscriptionSynchronizer.cs
@@ -65,7 +65,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds
             SecurityChanges newChanges;
             do
             {
-
                 universeData.Clear();
                 newChanges = SecurityChanges.None;
                 foreach (var subscription in subscriptions)
@@ -87,7 +86,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                     }
 
                     var packet = new DataFeedPacket(subscription.Security, subscription.Configuration);
-                    data.Add(packet);
 
                     var configuration = subscription.Configuration;
                     var offsetProvider = subscription.OffsetProvider;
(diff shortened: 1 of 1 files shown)


Common/Symbol.cs
```diff
@@ -174,6 +174,15 @@ public bool IsCanonical()
                 ID.Date == SecurityIdentifier.DefaultDate;
         }
 
+        /// <summary>
+        /// Method returns true, if symbol is a universe symbol
+        /// </summary>
+        /// <returns>true, if symbol is a universe symbol</returns>
+        public bool IsUniverseSymbol()
+        {
+            return Value.Contains("-UNIVERSE");
```
mchandschuh: This can't be guaranteed to be correct in either case (universe or non-universe).

Engine/DataFeeds/SubscriptionSynchronizer.cs
```diff
@@ -145,6 +143,11 @@ public TimeSlice Sync(DateTime frontier, IEnumerable<Subscription> subscriptions
                         }
                     }
 
+                    if (packet.Count > 0 && !packet.Configuration.Symbol.IsUniverseSymbol())
```
mchandschuh: The subscription defines whether or not it's a universe subscription [here](https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/Subscription.cs#L79).

For posterity: This change will prevent universe subscription data from making its way into the `Slice` object.

Engine/DataFeeds/SubscriptionSynchronizer.cs
```diff
@@ -108,40 +106,47 @@ public TimeSlice Sync(DateTime frontier, IEnumerable<Subscription> subscriptions
...
-                                collection = new BaseDataCollection(frontier, subscription.Configuration.Symbol, packetData);
+                                collection.Data.AddRange(packetData);
                             }
-
-                            universeData[subscription.Universe] = collection;
                         }
                         else
                         {
-                            collection.Data.AddRange(packetData);
+                            data.Add(packet);
```
mchandschuh: no changed required, but personal style, if an else block has a single line I typically invert the if so that the single line is up top near the condition and the larger block is in the else. This helps keeps the condition closer to it's downstream impact

PR #1444, opened 2017-12-28
Exclude data feed packets for removed symbols

StefanoRaggi:
With universe selection, data packets for symbols just removed from the universe were still being added to the `TimeSlice` (in the same time step) and also included in the `Slice` passed to `OnData`.


jaredbroad:
Reluctant to merge this as points to wider structural issue. Lets let it simmer until the final solution is reached.

mchandschuh:
Opened issue https://github.com/QuantConnect/Lean/issues/1604

PR #1461, opened 2018-01-03
Fix scheduled event firing order in backtesting

StefanoRaggi:
When backtesting with multiple scheduled events, these were fired in unpredictable order, not respecting the expected sequence sorted by time.

### DIFF
--- a/Engine/RealTime/BacktestingRealTimeHandler.cs
+++ b/Engine/RealTime/BacktestingRealTimeHandler.cs
@@ -16,6 +16,8 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 using QuantConnect.Interfaces;
 using QuantConnect.Lean.Engine.Results;
 using QuantConnect.Logging;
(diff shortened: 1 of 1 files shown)


DMTSource:
The heart emoji isn't enough, this one needs a **THANK YOU!** as this required some workarounds using _ExtendedMarketHours_ for pre-open compute work.

Question, if a scheduled call waits for first data point with normal market data(at 9:31 the logs show), would this then be a sort of look-ahead compared to live trading which would work in real time/data? I use things like History and Price around 9:25-9:29 to process a large universe so that I can act at open. In backtesting, I began using ExtendedMarketHours=True to ensure things were firing more accurately in the backtester.

StefanoRaggi:
@DMTSource Glad you like the fix :)

To answer your question, there is no look-ahead in backtesting, because events scheduled before the `EndTime` of each bar are fired _before_ the current bar is processed and the security current price is updated.

For example, assuming `Resolution.Minute` and `ExtendedMarketHours=False`, when you receive the first minute bar of the day at 9:31, the events scheduled at 9:25 and 9:29 will fire and event handler code will not "see" the prices in the current bar, but only prices received earlier (in this case the close of the previous day).

PR #1465, opened 2018-01-04
Historical Data Requests fetchs data from non-subscribed securities

AlexCatarino:
It was not possible to add a security based on its historical data, since we needed to add the security before requesting its security data. Universe Selection algorithms are an example of such usage.

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -245,13 +245,12 @@ namespace QuantConnect.Algorithm
         {
             var requests = symbols.Select(x =>
             {
-                var security = Securities[x];
-                var config = GetMatchingSubscription(security, typeof(T));
+                var config = GetMatchingSubscription(x, typeof(T));
                 if (config == null) return null;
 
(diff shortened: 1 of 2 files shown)


Algorithm/QCAlgorithm.History.cs
```diff
@@ -647,18 +648,71 @@ private HistoryRequest CreateHistoryRequest(Security security, SubscriptionDataC
...
+                case SecurityType.Option:
+                    return new Securities.Option.OptionExchange(exchangeHours);
+                case SecurityType.Forex:
+                    return new Securities.Forex.ForexExchange(exchangeHours);
+                case SecurityType.Future:
+                    return new Securities.Future.FutureExchange(exchangeHours);
+                case SecurityType.Cfd:
+                    return new Securities.Cfd.CfdExchange(exchangeHours);
+                case SecurityType.Crypto:
+                    return new Securities.Crypto.CryptoExchange(exchangeHours);
```
mchandschuh: do we actually _need_ the exchange object? I think we just need the hours and the time zone which are all contained within the `SecurityExchangeHours`

PR #1468, opened 2018-01-04
Adds OrderResponse error message for quantity less than lot size

AlexCatarino:
Invalid orders due to quantity less than lot size were marked as zero quantity. In order to improve error message, independent order response errors were implemented and hold more information.

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -649,9 +649,14 @@ namespace QuantConnect.Algorithm
             }
 
             //Ordering 0 is useless.
-            if (request.Quantity == 0 || request.Symbol == null || request.Symbol == QuantConnect.Symbol.Empty || Math.Abs(request.Quantity) < security.SymbolProperties.LotSize)
+            if (request.Quantity == 0)
             {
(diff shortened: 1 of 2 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -649,11 +649,16 @@ private OrderResponse PreOrderChecksImpl(SubmitOrderRequest request)
             }
 
             //Ordering 0 is useless.
-            if (request.Quantity == 0 || request.Symbol == null || request.Symbol == QuantConnect.Symbol.Empty || Math.Abs(request.Quantity) < security.SymbolProperties.LotSize)
+            if (request.Quantity == 0)
```
jaredbroad: Removed two checks here; do you think they'll not be needed? 
"request.Symbol == null || request.Symbol == QuantConnect.Symbol.Empty"

AlexCatarino: Before that if-statement, there is this one: 
`if (!Securities.TryGetValue(request.Symbol, out security))`
If `request.Symbol == null`, it will throw an exception.
If `request.Symbol == QuantConnect.Symbol.Empty`, it will be true, since Securities keys should not be `Symbol.Empty`.

PR #1475, opened 2018-01-10
Fix bug in Symbol inequality operator

### LINKED ISSUE #1454: Option Symbols Incorrect Across Splits
Trading AAPL options: 

2014-06-09 00:00:00 : | Runtime Error: System.ArgumentNullException: Value cannot be null. Parameter name: key   at System.Collections.Concurrent.ConcurrentDictionary`2[TKey,TValue].ContainsKey (TKey key) [0x0000b] in <dca3b561b8ad4f9fb10141d81b39ff45>:0    at QuantConnect.Securities.SecurityManager.get_Item (QuantConnect.Symbol symbol) [0x00001] in <88b912b0fd84484faf8b4c982c08e712>:0    at QuantConnect.Securities.SecurityPortfolioManager.ApplySplitToOptions (QuantConnect.Data.Market.Split split) [0x000a3] in <88b912b0fd84484faf8b4c982c08e712>:0    at QuantConnect.Securities.SecurityPortfolioManager.ApplySplit (QuantConnect.Data.Market.Split split) [0x00009] in <88b912b0fd84484faf8b4c982c08e712>:0    at QuantConnect.Lean.Engine.AlgorithmManager.Run (QuantConnect.Packets.AlgorithmNodePacket job, QuantConnect.Interfaces.IAlgorithm algorithm, QuantConnect.Lean.Engine.DataFeeds.IDataFeed feed, QuantConnect.Lean.Engine.TransactionHandlers.ITransactionHandler transactions, QuantConnect.Lean.Engine.Results.IResultHandler results, QuantConnect.Lean.Engine.RealTime.IRealTimeHandler realtime, QuantConnect.Lean.Engine.Server.ILeanManager leanManager, QuantConnect.Lean.Engine.Alpha.IAlphaHandler alphas, System.Threading.CancellationToken token) [0x00bd9] in <2516aeb381a54c2a9787b534047df995>:0

StefanoRaggi:
The operator method was returning true when comparing a null symbol against null.

Fixes #1454

### DIFF
--- a/Tests/Common/SymbolTests.cs
+++ b/Tests/Common/SymbolTests.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -21,7 +21,6 @@ using Newtonsoft.Json;
 using NUnit.Framework;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
-using QuantConnect.Securities;
 using QuantConnect.Securities.Option;
 
 namespace QuantConnect.Tests.Common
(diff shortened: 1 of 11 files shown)


Tests/Common/SymbolTests.cs
```diff
@@ -270,6 +269,22 @@ public void ComparesAgainstStringIgnoringCase()
...
 
+        [Test]
+        public void ComparesAgainstNull()
+        {
+            var validSymbol = Symbols.SPY;
+            var emptySymbol = Symbol.Empty;
+            Symbol nullSymbol = null;
+
+            Assert.IsTrue(validSymbol != null);
+            Assert.IsTrue(emptySymbol != null);
```
mchandschuh: I think we want the empty symbol to behave just like null, it's a 'null impl' of the symbol type. Conceptually `Symbol.Empty` and `(Symbol)null` are exactly the same -> the symbol wasn't specified, so I think they should compare the same

StefanoRaggi: If we want `Symbol.Empty` to behave like `(Symbol)null`, don't we just need to change its definition ?
from:
        `public static readonly Symbol Empty = new Symbol(SecurityIdentifier.Empty, string.Empty);`
to:
        `public static readonly Symbol Empty = null;`

PR #1476, opened 2018-01-12
Option split bug fixes

StefanoRaggi:
### Fix split adjustment for option holding price
This bug was causing equity sampling (after the split, before market open) to use a non-split-adjusted price in the calculation of `TotalPortfolioValue`.

The `Drawdown` regression stat for `OptionSplitRegressionAlgorithm` has changed due to the removal of the price spike and has been corrected.

### Fix option split handling for symbols added with AddOptionContract
Option contracts added to the algorithm with `AddOptionContract` were not replaced with new symbols after underlying splits, causing the algorithm to stop receiving data for them and invalid holdings in the portfolio.


jaredbroad:
Need a more holistic solution which doesn't require changing the symbol object -- symbol objects are meant to represent the security and a split doesn't mean it different contract.

PR #1477, opened 2018-01-12
Move alpha scoring off algo thread

mchandschuh:
Analyzing the alphas on the algorithm thread was a major performance killer. This PR moves the analysis to the alpha thread and takes snapshots at each time step so the analysis can be consistently performed without the need for synchronization of the threads.


Common/Algorithm/Framework/Alphas/Analysis/AlphaManager.cs
```diff
@@ -70,51 +65,94 @@ public class AlphaManager
...
-            foreach (var alpha in collection.Alphas)
+            if (generatedAlphas != null && generatedAlphas.Alphas.Count > 0)
             {
-                // save initial security values and deterine analysis period
-                var initialValues = _securityValuesProvider.GetValues(alpha.Symbol);
-                var analysisPeriod = alpha.Period + TimeSpan.FromTicks((long) (_extraAnalysisPeriodRatio * alpha.Period.Ticks));
+                foreach (var alpha in generatedAlphas.Alphas)
+                {
+                    // save initial security values and deterine analysis period
+                    var initialValues = securityValuesCollection[alpha.Symbol];
```
jaredbroad: Possible key not found?

Common/Algorithm/Framework/Alphas/Analysis/AlphaAnalysisContext.cs
```diff
@@ -168,5 +173,25 @@ public override string ToString()
...
+        /// <summary>Determines whether the specified object is equal to the current object.</summary>
+        /// <returns>true if the specified object  is equal to the current object; otherwise, false.</returns>
+        /// <param name="obj">The object to compare with the current object. </param>
+        /// <filterpriority>2</filterpriority>
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != GetType()) return false;
+            return Id.Equals(((AlphaAnalysisContext)obj).Id);
```
jaredbroad: Curious what is the main reason for this?

Common/AlphaRuntimeStatistics.cs
```diff
@@ -17,34 +20,67 @@ public class AlphaRuntimeStatistics
...
+            var accountCurrencySymbol = Currencies.GetCurrencySymbol(CashBook.AccountCurrency);
+            return new Dictionary<string, string>
+            {
+                {"Total Alphas Generated", $"{TotalAlphasGenerated}"},
+                {"Total Alphas Closed", $"{TotalAlphasClosed}"},
+                {"Total Alphas Analysis Completed", $"{TotalAlphasAnalysisCompleted}"},
+                {"Long Alpha Count", $"{LongCount}"},
+                {"Short Alpha Count", $"{ShortCount}"},
+                {"Long/Short Ratio", $"{Math.Round(100*LongShortRatio, 2)}%"},
+                {"Total Estimated Alpha Value", $"{accountCurrencySymbol}{TotalEstimatedAlphaValue.SmartRounding()}"},
```
jaredbroad: Should we tweak this to monthly or this might vary depending on how long the backtest runs?

Engine/Alphas/DefaultAlphaHandler.cs
```diff
@@ -174,20 +144,9 @@ public void OnAfterAlgorithmInitialized(IAlgorithm algorithm)
...
-            }
-            else
-            {
-                // space out backtesting samples evenly
-                var backtestPeriod = algorithm.EndDate - algorithm.StartDate;
-                ChartUpdateInterval = TimeSpan.FromTicks(backtestPeriod.Ticks / BacktestChartSamples);
-            }
+            // send date ranges to extensions for initialization -- this data wasn't available when the handler was
+            // initialzied, so we need to invoke it here
+            AlphaManager.InitializeExtensionsForRange(algorithm.StartDate, algorithm.EndDate, algorithm.UtcTime);
```
jaredbroad: Shame about this; the order of init is becoming increasingly challenging,

Engine/Alphas/DefaultAlphaHandler.cs
```diff
@@ -275,9 +191,19 @@ public virtual void Run()
                     throw;
                 }
 
-                Thread.Sleep(50);
+                Thread.Sleep(1);
```
jaredbroad: Curious if this made a difference?

Common/Algorithm/Framework/Alphas/Analysis/AlphaAnalysisContext.cs
mchandschuh: `Equals` and `GetHashCode` are used in hashset and concurrent dictionary implementations for the key types. It's a performance boost to implement these members, otherwise they would all have the same hashcode and the hashset or concurrenct dictionary is then about as useful as a list

Common/Algorithm/Framework/Alphas/Analysis/AlphaManager.cs
mchandschuh: hmm, good point, I suppose it's *possible* if the algorithm creates a symbol it doesn't have data for

Common/AlphaRuntimeStatistics.cs
mchandschuh: There's another property that is the mean alpha value, basically it's the total/# alphas closed. The normalized value is probably what I would use as a metric over the total value since total value (as mentioned) is proportional to the backtest length

Engine/Alphas/DefaultAlphaHandler.cs
mchandschuh: This was 50 before because we were just doing persistence/messaging updates, but now since we're doing scoring as well, the 50ms sleep was way too long

Issue #1378, opened 2017-12-06
SecurityHolding's  UnrealizedProfit and TotalCloseProfit methods


benjaminthedonkey:
I think It would be beneficial to add the concept of "Trades", as a result of filled orders,  and maybe also "Positions" ( consolidated trades by security ).  Currently, the class SecurityHolding only have a total quantity and an average filled price. This works fine in most cases but also results in proximate values for some important methods on that class, like UnrealizedProfit() or TotalCloseProfit().  Now that we have bid-ask quotes, all this calculations could be made with more accuracy.

jaredbroad:
Thanks Benjamin; why do you think the UnrealizedProfit / TotalCloseProfit is approximate? Can you please give an example where SecurityHolding doesn't model this close enough? Thanks

AlexCatarino:
The concept of [`Trade`](https://github.com/QuantConnect/Lean/blob/master/Common/Statistics/Trade.cs) does exist. We can get the list of all closed trades in the `TradeBuilder`:
```csharp
public override void OnEndOfAlgorithm()
{
    foreach (var trade in TradeBuilder.ClosedTrades)
    {
        //
    }
}
```

benjaminthedonkey:
Thanks guys. Regarding UnrealizedProfit / TotalCloseProfit methods, I just think they should use bid - ask quotes and not mid point.  For example, when I buy, the moment I get filled, I owe the difference between the filled price and the the bid .

AlexCatarino:
By default in Lean, if there is quote data, when you buy you get filled with the ask price and when you sell you get filled with the bid price (see [ImmediateFillModel.GetPrices](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/Fills/ImmediateFillModel.cs#L379)).

benjaminthedonkey:
I also suggest to add a method in the ForexHolding class to return the TotalCloseProfit in Pips (or pipettes if the broker support it). This is useful for people that is used to work with this price representation. thanks!

``` C#
         // <summary>
        /// Profit if we closed the holdings right now including the approximate fees.
        /// </summary>
        /// <remarks>Does not use the transaction model for market fills but should.</remarks>
        public virtual decimal TotalCloseProfitPips()
        {
            var pipValue = (_security.SymbolProperties.MinimumPriceVariation * AbsoluteQuantity) / AveragePrice;
            return Math.Round((TotalCloseProfit() / pipValue), 2);
        }
```

[...]

PR #1480, opened 2018-01-15
TotalCloseProfitPips method added to ForexHolding

Jay-Jay-D:
This PR adds a new method to the ForexHolding class to get the total profit in pips. #1378 

[Here](https://gist.github.com/Jay-Jay-D/80b33ec27e1a115438a0a4d0541cfacd) is a simple QCAlgorithm to test it.


mchandschuh:
Great stuff, maybe we can also add a unit test?

PR #1484, opened 2018-01-16
Fix CashBook serialization error

StefanoRaggi:
The following error was being thrown in `BaseResultsHandler.SaveResults`:
_Token PropertyName in state Property would result in an invalid JSON object. Path 'Cash.USD'._

### DIFF
--- a/Common/SymbolJsonConverter.cs
+++ b/Common/SymbolJsonConverter.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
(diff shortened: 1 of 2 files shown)


Common/Securities/Cash.cs
```diff
@@ -138,7 +138,7 @@ public Security EnsureCurrencyDataFeed(SecurityManager securities, SubscriptionM
         {
             if (Symbol == CashBook.AccountCurrency)
             {
-                SecuritySymbol = QuantConnect.Symbol.Empty;
+                SecuritySymbol = null;
```
mchandschuh: Do we know why `Symbol.Empty` failed serialization to begin with?

StefanoRaggi: This bug has been introduced in the last Symbol equality operator change, which now evaluates  `Symbol.Empty == null`. The check at the top of `SymbolJsonConverter.WriteJson()` will now return for both `Symbol.Empty` and `null`.

mchandschuh: Maybe we can update `SymbolJsonConverter` to use:
`if (ReferenceEquals(symbol, null) { return; }`

PR #1487, opened 2018-01-17
Adds null-check in LatestPriceFillModel.GetPrices

AlexCatarino:
When there is not quote bar data, we can return the asset prices.

### DIFF
--- a/Common/Orders/Fills/LatestPriceFillModel.cs
+++ b/Common/Orders/Fills/LatestPriceFillModel.cs
@@ -71,20 +71,23 @@ namespace QuantConnect.Orders.Fills
             // Assume that the security has both a trade and quote subscription
             // This should be true for crypto securities
             var quoteBar = asset.Cache.GetData<QuoteBar>();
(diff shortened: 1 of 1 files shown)


Common/Orders/Fills/LatestPriceFillModel.cs
```diff
@@ -71,20 +71,23 @@ protected override Prices GetPrices(Security asset, OrderDirection direction)
...
-            }
-            else
+            if (quoteBar != null)
             {
-                // The latest pricing data came from a quote
-                var bar = direction == OrderDirection.Sell ? quoteBar.Bid : quoteBar.Ask;
-                if (bar != null)
+                var tradeBar = asset.Cache.GetData<TradeBar>();
+
+                if (tradeBar.EndTime > quoteBar.EndTime)
```
mchandschuh: should we also be checking if `tradeBar` is null?

AlexCatarino: I thought it would be covered but proved wrong by unit tests that I have just added. Thank you!

PR #1495, opened 2018-01-20
Make HeikinAshi work with Forex data

### LINKED ISSUE #1286: Make HeikinAshi indicator work with QuoteBars
## Problem

Unable to use the HeikinAshi with QuoteBar Securities (e.g. Forex)

## Reproduce

```
ha = self.HeikinAshi("EURUSD")

# Error: Type mismatch found between consolidator and indicator for symbol: 
#   EURUSD.Consolidator outputs type QuoteBar but indicator expects input type TradeBar

```

## Proposed Solution

As far as I can tell, there is no reason we cannot have the HeikinAshi indicator work with QuoteBar. Would it be as simple as changing the inherited class from `TradeBarIndicator` to `BarIndicator`?

```
# Lean/Indicators/HeikinAshi.cs

namespace QuantConnect.Indicators
{
    public class HeikinAshi : BarIndicator
```

Worthy7:
Changed HeikinAshi to inherit from BarIndicator instead of TradeBarIndicator to facilitate forex data. This is done by removing volume as it is unnecessary in HeikinAshi.

Fixes #1286

### DIFF
--- a/Indicators/HeikinAshi.cs
+++ b/Indicators/HeikinAshi.cs
@@ -26,7 +26,7 @@ namespace QuantConnect.Indicators
     /// HA_High[0] = MAX(High[0], HA_Open[0], HA_Close[0])
     /// HA_Low[0] = MIN(Low[0], HA_Open[0], HA_Close[0])
     /// </summary>
-    public class HeikinAshi : TradeBarIndicator
+    public class HeikinAshi : BarIndicator
     {
         /// <summary>
         /// Gets the Heikin-Ashi Open
@@ -96,7 +96,7 @@ namespace QuantConnect.Indicators
         /// </summary>
         /// <param name="input">The input given to the indicator</param>
         /// <returns> A new value for this indicator </returns>
-        protected override decimal ComputeNextValue(TradeBar input)
+        protected override decimal ComputeNextValue(IBaseDataBar input)
         {
             if (!IsReady)
             {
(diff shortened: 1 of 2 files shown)


mchandschuh:
@Worthy7 -- let's go with @AlexCatarino's suggestion of keeping the `Volume` property and using a type check to update it if it is trade bar data

Worthy7:
Sorry guys, but yes @StefanoRaggi you are correct, it was from a pull request you made lol https://github.com/QuantConnect/Lean/pull/662

For this request:
https://www.quantconnect.com/forum/discussion/1566/getting-volume-in-a-heikin-ashi-tradebar

So, I actually considered this before I PR'ed, I found the posts above etc, but the point is, it is simple not the responsibility of the HeikinAshi class to pass Volume around. That should be left to the parent class or some kind of extension. IMO.

I can obviously just add the code you asked for now, but I think it's not the correct way to to do this (Imagine if we had loads of Heikin Ashi type indicators....)

Worthy7:
Rebased on to today's master.
I don't really understand about how chaining indicators works for now but anyway, there it is.

AlexCatarino:
Thank you, @Worthy7!

mchandschuh:
@Worthy7 - apologies for the flip flop here. My initial concern with removing the `Volume` property was the potential for compile errors in user algorithms. We've determined this concern to be low risk.

I agree that we should aim to be as *correct* as possible. Looking at various definitions of the indicator online, not a single one mentions volume, so let's remove it.

Any users that are currently using the volume property will experience a compile error and will need to obtain that data from somewhere else.

mchandschuh:
Merged #1495 in 41b233fb4226f0db77d4846157754dd56dfccad8

Worthy7:
O, but I put the volume back already and you merged it lol

Issue #1483, opened 2018-01-16
Daily FX Locking Up After March 31st.


jaredbroad:
```
using QuantConnect.Data.Custom;
using NodaTime;

namespace QuantConnect.Algorithm.CSharp
{
    public class BasicTemplateAlgorithm : QCAlgorithm
    {
        public override void Initialize()
        {
            SetStartDate(2015, 1, 07);
            SetEndDate(2017, 12, 11);
            SetCash(100000);

            AddData<DailyFx>("DFX", Resolution.Minute);
            AddForex("EURUSD", Resolution.Minute);
        } 
        public override void OnData(Slice data)
        { 
        }
    }
}
```

jaredbroad:
Fixed in #1502

PR #1500, opened 2018-01-22
RestSubscriptionStreamReader issue.

Jay-Jay-D:
#1483 

The problem has no to do with the `DailyFx` class itself, instead the `CollectionSubscriptionDataSourceReader` never reaches the end of the stream when the `SubscriptionTransportMedium` is `Rest`. And in this case is because, as you can see in the `RestSubscriptionStreamReader` implementation, [it never _ends_ to read the stream](https://github.com/QuantConnect/Lean/blob/42c782e312efb8ba210285daa95136d9c39b526d/Engine/DataFeeds/Transport/RestSubscriptionStreamReader.cs#L65).

The fix was simply to incorporate a static bool variable called ´delivered´; that variable will flag once the whole content of the collection is pumped in via the `ReadLine` method and it will encapsulated by the `EndOfStream` method.

### DIFF
--- a/Common/Data/Custom/DailyFx.cs
+++ b/Common/Data/Custom/DailyFx.cs
@@ -172,50 +172,52 @@ namespace QuantConnect.Data.Custom
         /// <returns></returns>
         public override BaseData Reader(SubscriptionDataConfig config, string content, DateTime date, bool isLiveMode)
         {
-            if (_previousContent == content)
+            if (isLiveMode)
             {
-                return null;
+                if (_previousContent == content)
+                {
+                    return null;
+                }
+
+                _previousContent = content;
+
+                // clean old entries from memory
+                var clearingDate = date.Date.AddDays(-2);
+                var oldEntries = _previous.Where(kvp => kvp.Value.DisplayDate.UtcDateTime.Date < clearingDate).ToList();
+                oldEntries.ForEach(oe => _previous.Remove(oe.Key));
             }
-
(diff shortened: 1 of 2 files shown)


mchandschuh:
@Jay-Jay-D -- I'm not sure this will work. As I'm reading it, this stream would end after _every_ response, which is not ideal. The rest stream is typically used for rest endpoints in live mode.

Is this issue pertaining to backtest or live? If it's in backtest, then we may need to come up with another solution.

mchandschuh:
For posterity:

We spoke directly regarding the usage of the `RestSubscriptionStreamReader` in backtest mode. This reader was designed specifically for the live case, hence when `EndOfStream` always returns true. We could make an update to the `RestSubscriptionStreamReader` to accept a `liveMode` flag, and if `liveMode`, then keep existing behavior, and in backtest mode, end the stream after each invocation to the source. This will allow the stream to recycle itself in backtest. Now this may still have issues in this exact case because I think the factory needs to maintain state, and therefore, _not_ be recycled.

Another option is to improve the `RemoteFileSubscriptionStreamReader`  so that when invoked from the `CollectionSubscriptionDataSourceReader`, it reads to the end of the file and sends the entire file contents into the `BaseData.Reader` method. Arguably, this is what the collection format is intended to do, push the entire file/response/ect into the reader so it can create a collection of data in one invocation.

mchandschuh:
This is superceded by #1502

PR #1502, opened 2018-01-23
 Implementing live mode flag into RestSubscriptionStreamReader

Jay-Jay-D:
This PR implements a new behavior for the `RestSubscriptionStreamReader`, depending if the algorithm is running live or not. ~The argument is passed through the `SubscriptionDataSource` class.~

- In live mode it behaves as default, i.e. the `RestSubscriptionStreamReader.EndOfStream` is always false.
- In backtesting mode, the `RestSubscriptionStreamReader.EndOfStream` will be true once the file content is delivered.

This implementation is a better, correct implementation of the #1500 and fixes the issue #1483.

### DIFF
--- a/Common/Data/SubscriptionDataSource.cs
+++ b/Common/Data/SubscriptionDataSource.cs
@@ -58,8 +58,8 @@ namespace QuantConnect.Data
         /// Initializes a new instance of the <see cref="SubscriptionDataSource"/> class.
         /// </summary>
         /// <param name="source">The subscription's data source location</param>
-        /// <param name="format">The format of the data within the source</param>
         /// <param name="transportMedium">The transport medium to be used to retrieve the subscription's data from the source</param>
+        /// <param name="format">The format of the data within the source</param>
         public SubscriptionDataSource(string source, SubscriptionTransportMedium transportMedium, FileFormat format)
             : this(source, transportMedium, format, null)
         {
--- a/Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
+++ b/Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
@@ -79,7 +79,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                 {
                     default:
                     case SubscriptionTransportMedium.Rest:
-                        reader = new RestSubscriptionStreamReader(source.Source, source.Headers);
+                        reader = new RestSubscriptionStreamReader(source.Source, source.Headers, _isLiveMode);
                         break;
                     case SubscriptionTransportMedium.LocalFile:
                         reader = new LocalFileSubscriptionStreamReader(_dataCacheProvider, source.Source);
--- a/Engine/DataFeeds/Transport/RestSubscriptionStreamReader.cs
+++ b/Engine/DataFeeds/Transport/RestSubscriptionStreamReader.cs
@@ -29,16 +29,20 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Transport
     {
         private readonly RestClient _client;
         private readonly RestRequest _request;
+        private bool _isLiveMode;
+        private bool _delivered;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RestSubscriptionStreamReader"/> class.
         /// </summary>
         /// <param name="source">The source url to poll with a GET</param>
         /// <param name="headers">Defines header values to add to the request</param>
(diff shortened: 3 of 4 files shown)


Common/Data/SubscriptionDataSource.cs
```diff
@@ -44,24 +44,33 @@ public class SubscriptionDataSource : IEquatable<SubscriptionDataSource>
         /// </summary>
         public readonly IReadOnlyList<KeyValuePair<string, string>> Headers;
 
+        /// <summary>
+        /// Flags if algorithm is running on a live server.
+        /// </summary>
+        public readonly bool? IsLiveMode;
```
mchandschuh: @Jay-Jay-D -- I don't think we need this here. The individual subscription factories will know whether or not they're in live mode or not. Ideally the `SubscriptionDataSource` doesn't care what type of algorithm we're running. Let's remove this and source the `isLiveMode` value elsewhere. Ping me on slack if you get caught in any particular spot, but IIRC, the data should already be available where it's needed

Common/Data/SubscriptionDataSource.cs
```diff
@@ -44,24 +44,33 @@ public class SubscriptionDataSource : IEquatable<SubscriptionDataSource>
...
         /// <summary>
         /// Initializes a new instance of the <see cref="SubscriptionDataSource"/> class.
         /// </summary>
         /// <param name="source">The subscription's data source location</param>
         /// <param name="transportMedium">The transport medium to be used to retrieve the subscription's data from the source</param>
-        public SubscriptionDataSource(string source, SubscriptionTransportMedium transportMedium)
-            : this(source, transportMedium, FileFormat.Csv)
+        /// <param name="isLiveMode">Flags if algorithm is running on a live server.</param>
+        /// <remarks>In case of implementing REST as Transport Medium you must declare if the Algorithm is running in live mode through the <code>isLiveMode</code> argument.</remarks>
+        public SubscriptionDataSource(string source, SubscriptionTransportMedium transportMedium, bool isLiveMode)
```
mchandschuh: Sadly, changing these constructors would break many many algorithms and would also require users to properly set the value, which is odd considering the `GetSource` method has a `isLiveMode`parameter in it and would just pipe it through. This means that the calling context already has access to this data so it shouldn't be needed on this object.

Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
```diff
@@ -79,7 +80,8 @@ public IEnumerable<BaseData> Read(SubscriptionDataSource source)
                 {
                     default:
                     case SubscriptionTransportMedium.Rest:
-                        reader = new RestSubscriptionStreamReader(source.Source, source.Headers);
+                        var isliveMode = source.IsLiveMode != null && source.IsLiveMode.Value;
```
mchandschuh: For example, this class already has a `_isLiveMode` class member variable we can use here.

Common/Data/SubscriptionDataSource.cs
```diff
@@ -73,6 +75,7 @@ public SubscriptionDataSource(string source, SubscriptionTransportMedium transpo
         /// <param name="transportMedium">The transport medium to be used to retrieve the subscription's data from the source</param>
         /// <param name="format">The format of the data within the source</param>
         /// <param name="headers">The headers to be used for this source</param>
+        /// <remarks>In case of implementing REST as Transport Medium you must declare if the Algorithm is running in live mode through the <code>isLiveMode</code> argument.</remarks>
```
mchandschuh: Can we revert these left over comments from the previous commits?

Engine/DataFeeds/Transport/RestSubscriptionStreamReader.cs
```diff
@@ -75,6 +79,7 @@ public string ReadLine()
                 var response = _client.Execute(_request);
                 if (response != null)
                 {
+                    _delivered = !_isLiveMode;
```
mchandschuh: Should this care about live mode? `EndOfStream` properly checks both flags, not sure we need to differentiate between live/non-live on this line.

mchandschuh:
Before merging, would be nice to clean up the history on this branch using `git rebase` -- ping me if you'd like some help.

[Here's](https://sethrobertson.github.io/GitBestPractices/) a great resource for git best practices (of note here is the section on sausage making :) )

jaredbroad:
Nice work @Jay-Jay-D

PR #1503, opened 2018-01-23
Liquidate options on split warning

mchandschuh:
>NOTE: As this PR currently stands, it is *possible* for a SID collision with options. IOW, a pre-split option contract SID may compare equal to a post-split option contract SID. Given that we mark split contracts as non-tradable, this could cause a post-split contract to be not tradable.

>The SID collision is still possible, however unlikely. In addition, all holdings are liquidated and open orders cancelled, so there's no risk of cross contamination of holdings. The only risk is in post-backtest reporting where if a SID collision is present, then we may report two different contracts (pre/post split) as being the same contract. Resolving this turned into a much bigger and more involved task than previously thought and will require some more design. This PR is a middle ground to keep the backtest simulations accurate while also making users aware of the shortcoming.

>This PR is based on top of [PR #1501](https://github.com/QuantConnect/Lean/pull/1501): Add split warning event

#### Add MarketOnCloseOrder.DefaultSubmissionTimeBuffer

This is the default time span before market close that we'll allow an MOC order to be placed. Nominally set to 15 minutes + 30 second buffer for processing/transport.

#### Add tag to SecurityTransactionManager.CancelOpenOrders

When bulk canceling orders for a particular symbol it's nice to tag the order so we know why they were all canceled.

#### Liquidate option contract holdings before split events

Using the newly added split warnings, submit market on close order at the last possible moment to liquidate option contract holdings. After the liquidate order has been submitted, we now mark the option contract as non-tradable, thereby preventing any further trading in post-split option contracts.

This is added because we currently do not properly support option contract splits. To properly support, we'll need some extra data sources defining how the contracts are rewritten due to the split event. This data is freely available via the OCC in pdf format.

#### Remove unused option split code

This code was an attempt at modeling option splits. The reality of the situation
is far more complicated and requires an external data source to be properly modeled.
Some things to consider are handling a split event before all the contracts from the
previous split event have expired.

With respect to the holdings, I think a more generic method to rewrite the holdings
would be beneficial when we have the data, but it isn't as simple as applying a split
factor. During splits and other corporate events, the OCC will rewrite the options
conract such that you are 'made whole'. Sometimes this involves adjusting the contract
multiplier away from the default 100 (or 10 for minis). Another fairly common case
(especially with M&A's) is to have part of the contract paid out in cash. So it's
possible for pretty much every value in the contract/holdings to change:

        1. Multipler
        2. Strike Price
        3. Number of contracts held
        4. Cash payout in leiu of fractional stock
        5. Constant cash payout

Links to removed code for easy reference later:
https://github.com/QuantConnect/Lean/blob/5f64665598b5e6119b42d11d84eb1062f12542a6/Common/Securities/SecurityPortfolioManager.cs#L696
https://github.com/QuantConnect/Lean/blob/5f64665598b5e6119b42d11d84eb1062f12542a6/Common/Securities/SecurityPortfolioManager.cs#L733
https://github.com/QuantConnect/Lean/blob/5f64665598b5e6119b42d11d84eb1062f12542a6/Common/Securities/Option/OptionHolding.cs#L49


mchandschuh:
@jaredbroad @StefanoRaggi @AlexCatarino -- Please review/look over when you have a chance. We should get this is even though we haven't resolved the *potential* SID collision case.

PR #1510, opened 2018-01-26
Allow ZipDataCacheProvider to support >2GB data files.

theGOTOguy:
ZipDataCacheProvider cannot open data files larger than Int32.MaxValue because of the internal limits of MemoryStream.  This becomes an issue when working with level 2 data from securities with a lot of activity.  Therefore, we catch this error when it happens and instead provide a ZipInputStream that connects directly to the data file instead.

### DIFF
--- a/Engine/DataFeeds/ZipDataCacheProvider.cs
+++ b/Engine/DataFeeds/ZipDataCacheProvider.cs
@@ -87,7 +87,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
                                 lock (_zipFileSynchronizer)
                                 {
-                                    stream = CreateStream(newItem.ZipFile, entryName);
+                                    stream = CreateStream(newItem.ZipFile, entryName, filename);
                                 }
 
                                 _zipFileCache.TryAdd(filename, newItem);
@@ -108,7 +108,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                         {
                             lock (_zipFileSynchronizer)
                             {
-                                stream = CreateStream(existingEntry.ZipFile, entryName);
+                                stream = CreateStream(existingEntry.ZipFile, entryName, filename);
                             }
                         }
                         catch (Exception exception)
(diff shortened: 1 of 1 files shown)


mchandschuh:
@theGOTOguy -- thanks for the PR and finding this issue. I wasn't aware of this. I'd like to look into this a little further and evaluate the pros/cons of the old way and this way. Ideally there's just _one_ way we handle this (the best way!). Upon a first glance, _not_ using a memory stream sounds like an improvement as it would allow us to stream the file instead of forcefully reading the entire thing into memory, but I suspect there are good reasons why the code is that way.

theGOTOguy:
Absolutely.  I can give a little bit more information if it helps.

There are two possibly non-mutually-exclusive reasons that it might have been done this way in the first place.
1)  That there was a performance improvement from holding the entire file in memory at the same time.
2)  That the previous coder learned, like I did, that entry.OpenReader returns corrupt data if more than one entry is open at a time, which is clearly necessary in Lean, so loaded it into memory to work around this.

Reason 1 is a great reason to keep MemoryStream intact, but reason 2 is not. I haven't done detailed tests of the performance in both scenarios but I do know that working at all is a huge improvement over failing to load the data at all, so I'm sure that the path forward must involve ZipInputStream whether to load all of the entries or to load just enormous ones.

jaredbroad:
:) FYI it was reason 2 -- we were looking for a work around to the OpenReader corrupting data. Loading everything into a memorystream wasn't ideal but we were hoping to avoid dangling file pointers later. 

Most of the files that its opening are relatively small -- can you give an example case of such a massive file?

One of the optimizations we've been toying with is pre-synchronizing all the data and storing in massive daily files (e.g. top-500 universe file, pre-synchronized).

theGOTOguy:
Ah, well that clears things up!  I'd be happy to change this PR to use just ZipInputStrream for everything unless you actually cared to do performance optimization.

While I can't actually share the data file that I'm using to cause this error (I don't own the copyright to the proprietary data set), I can tell you that it's market depth data for CL with all expiries just dumped into the same file.  We've written a custom data source to read it in, and the market depth data occasionally goes beyond 2GB.  For instance, on 9/28/2016, the uncompressed size of the market depth data is 2.3GB.

If you're looking to reproduce the issue, you could just write a custom, dummy data source and then just dump a massive randomly-generated data sequence on it.

Issue #1511, opened 2018-01-29
Add Custom Volatility Model for python


AlexCatarino:
At the moment, custom volatility model cannot be defined in python algorithms.
An interface must be created (like for other custom modeld: FeeModel, FillModel, etc) in order to enable this feature.

jaredbroad:
Merged in #1512

PR #1512, opened 2018-01-29
Enables custom volatility models in python algorithms

AlexCatarino:
Creates a python wrapper for volatility models created in python algorithms and adds a method to the Security object to set such models.
Adds an algorithm to show how volatility models can be implemented.

Ref.: #1511


Common/Python/VolatilityModelPythonWrapper.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        /// <summary>
+        /// Gets the volatility of the security as a percentage
+        /// </summary>
+        public decimal Volatility
+        {
+            get
+            {
+                return _model.Volatility;
```
mchandschuh: Out of curiosity, why don't we need the `using (Py.GIL()){ //... }` here?

AlexCatarino: We don't need it when just reading the value of a member.

Issue #1074, opened 2017-08-17
Add Indicator Coppock Curve


LiXiang618:
Coppock Curve (Normally N1 = 10, N2 = 14, N3 = 11)

The Coppock Curve is a momentum indicator developed by Edwin “Sedge” Coppock. Coppock used monthly data to identify buying opportunities when the indicator moved from negative territory to positive territory. Although Coppock did not use it for sell signals, many technical analysts consider a cross from positive to negative territory as a sell signal.

Formula:
Coppock Curve = N1-period WMA of N2-period RoC + N3-perod RoC
WMA = Weighted moving average
RoC = Rate-of-Change = [(Close - Close n periods ago) / (Close n periods ago)] * 100

Note:
1. Coppock used monthly data to construct this indicator. Sample data files include both monthly and daily data.
2. A weighted moving average puts more weight on recent data and less weight on older data. For example, a 3-period WMA would multiply the first data point by 1, the second data point by 2 and the third data point by 3. The sum of these three numbers is then divided by 6, which is the sum of the weightings (1 + 2 + 3), to create a weighted average.

Sample Data:
[spy_coppock_curve_daily.txt](https://github.com/QuantConnect/Lean/files/1232249/spy_coppock_curve_daily.txt)
[spy_coppock_curve_monthly.txt](https://github.com/QuantConnect/Lean/files/1232248/spy_coppock_curve_monthly.txt)

Reference:
http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:coppock_curve
http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:rate_of_change_roc_and_momentum

jaredbroad:
Merged in #1513

PR #1513, opened 2018-01-30
Coppock Curve indicator implementation

Jay-Jay-D:
#1074


Indicators/CoppockCurve.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            _shortRoc.Update(input);
+            _longRoc.Update(input);
+            if (_longRoc.IsReady && _shortRoc.IsReady)
+            {
+                var rocSum = _shortRoc + _longRoc;
+                _lwma.Update(input.Time, rocSum);
+                return _lwma;
+            }
+            return decimal.Zero;
```
mchandschuh: Is there any harm in emitting the non-ready value? When possible, I think it's best that indicators emit an estimate during the non-ready phase vs a zero.

Jay-Jay-D: Yes, if the `_shortRoc` is ready but the `_longRoc` is not (ROC == 0 if not ready), the LWMA will be updated with the value of the `_shortRoc`. In the default case maybe is not important, but in case the difference between `_shortRoc` and `_longRoc` is big, the `_lwma` (the indicator itself) will be ready with many deceptive values (just the short ROC).

Also, note that the zero value will be returned until both ROC are ready, then it’ll return the `_lwma` value, even when it is not ready.

mchandschuh: Since this is an oscillating indicator, we're OK with emitting a default value of 0 in the not ready state.

Issue #1077, opened 2017-08-18
Add Indicator MACD Histogram


LiXiang618:
MACD Histogram (Normally N1 = 12, N2 = 26, N3 = 9)

Developed by Thomas Aspray in 1986, the MACD-Histogram measures the distance between MACD and its signal line (the 9-day EMA of MACD). Like MACD, the MACD-Histogram is also an oscillator that fluctuates above and below the zero line.

Formula:
MACD: (N1-day EMA - N2-day EMA) 
Signal Line: N3-day EMA of MACD
MACD Histogram: MACD - Signal Line

Note:
N1, N2, and N3 are 12, 26 and 9 respectively in the sample data.

Sample Data:
[spy_macd_histogram_12_26_9.txt](https://github.com/QuantConnect/Lean/files/1235008/spy_macd_histogram_12_26_9.txt)

Reference:
http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:macd-histogram

mchandschuh:
This could be done easily by using the existing Mac indicator which already
defines a signal line, all it needs is to perform the subtraction.

4real:
I agree it should be a property of the existing indicator.

Shamitha-R:
I've started looking into getting this indicator added. Can I be made the assignee?

AlexCatarino:
@Shamitha-R , I think you need to fork Lean so that we can assign it to you.
Anyway, consider yourself assigned for this issue!
**Thank very much!**

jaredbroad:
Merged in #1514

PR #1514, opened 2018-01-30
MACD Histogram implemented

Jay-Jay-D:
#1077

### DIFF
--- a/Indicators/MovingAverageConvergenceDivergence.cs
+++ b/Indicators/MovingAverageConvergenceDivergence.cs
@@ -43,7 +43,7 @@ namespace QuantConnect.Indicators
         /// is an oscillator that fluctuates above and below the zero line.
         /// Bullish or bearish divergences in the MACD-Histogram can alert chartists to an imminent signal line crossover in MACD.
         /// </summary>
-        public decimal Histogram { get; private set; }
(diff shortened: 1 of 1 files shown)


Indicators/MovingAverageConvergenceDivergence.cs
```diff
@@ -36,6 +38,13 @@ public class MovingAverageConvergenceDivergence : Indicator
         /// </summary>
         public IndicatorBase<IndicatorDataPoint> Signal { get; private set; }
 
+        /// <summary>
+        /// Developed by Thomas Aspray in 1986, the MACD-Histogram measures the distance between MACD and its signal line, 
+        /// is an oscillator that fluctuates above and below the zero line.
+        /// Bullish or bearish divergences in the MACD-Histogram can alert chartists to an imminent signal line crossover in MACD.
+        /// </summary>
+        public decimal Histogram { get; private set; }
```
AlexCatarino: Why is this not a `IndicatorBase<IndicatorDataPoint>` like the other variables?

Jay-Jay-D: I tried using with a `CompositeIndicator` in the first commit, but it throws error when it is reset.

But I see your point, the code will more consistent, I'll make an implementation using and `Identity` indicator for the Histogram.

PR #1521, opened 2018-02-01
Cash buying power model for GDAX

StefanoRaggi:
This PR replaces PR #1420 and is chained to PR #1506.

#### Changes from PR #1506
- The `IMarginModel` interface was renamed to a more generic `IBuyingPowerModel` with new methods independent from margin calculations. Existing margin models were also renamed to corresponding buying power models.
- Margin specific code in `QCAlgorithm` and `SecurityPortfolioManager` was moved to the  `SecurityMarginBuyingPowerModel` class (previously called `SecurityMarginModel`), specifically the methods `GetSufficientCapitalForOrder` and `CalculateOrderQuantity`.
- The `IMarginCallModel` interface was also extended with a new `GetMarginCallOrders` method. This allows margin call model logic to be entirely contained in the implementations. Margin call model specific code in `SecurityPortfolioManager.ScanForMarginCall` was moved into the `DefaultMarginCallModel` class.

#### Changes in this PR
- The new `CashBuyingPowerModel` was added to support brokerages with Cash only accounts (e.g. GDAX)
  - when calculating buying (or selling) power, it takes open orders into account
  - this model is now the default buying power model for the `Crypto` security type and for `GDAXBrokerageModel`
- The default `AccountType` for GDAX has been changed to `AccountType.Cash`, since they do not currently support margin trading.
- New `BasicTemplateCryptoAlgorithm` was added (C# and Python)

### DIFF
--- a/Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
@@ -40,7 +40,11 @@ namespace QuantConnect.Algorithm.CSharp
 
             SetTimeZone(NodaTime.DateTimeZone.Utc);
             var security = AddSecurity(SecurityType.Crypto, "BTCUSD", Resolution.Daily, Market.GDAX, false, 3.3m, true);
+
+            // The default buying power model for the Crypto security type is now CashBuyingPowerModel.
+            // Since this test algorithm uses leverage we need to set a buying power model with margin.
             security.BuyingPowerModel = new SecurityMarginBuyingPowerModel(3.3m);
+
             var con = new QuoteBarConsolidator(1);
             SubscriptionManager.AddConsolidator("BTCUSD", con);
             con.DataConsolidated += DataConsolidated;
--- a/Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
@@ -43,7 +43,7 @@ namespace QuantConnect.Algorithm.CSharp
 
             // The default buying power model for the Crypto security type is now CashBuyingPowerModel.
             // Since this test algorithm uses leverage we need to set a buying power model with margin.
-            security.BuyingPowerModel = new SecurityMarginBuyingPowerModel(3.3m);
+            security.BuyingPowerModel = new SecurityMarginModel(3.3m);
 
             var con = new QuoteBarConsolidator(1);
             SubscriptionManager.AddConsolidator("BTCUSD", con);
--- a/Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
@@ -17,6 +17,7 @@ using QuantConnect.Data.Consolidators;
 using QuantConnect.Data.Market;
 using System;
 using QuantConnect.Brokerages;
+using QuantConnect.Securities;
 
 namespace QuantConnect.Algorithm.CSharp
 {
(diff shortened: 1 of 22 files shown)


Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
```diff
@@ -39,6 +40,7 @@ public override void Initialize()
 
             SetTimeZone(NodaTime.DateTimeZone.Utc);
             var security = AddSecurity(SecurityType.Crypto, "BTCUSD", Resolution.Daily, Market.GDAX, false, 3.3m, true);
+            security.BuyingPowerModel = new SecurityMarginBuyingPowerModel(3.3m);
```
mchandschuh: Perhaps a comment here indicating why we're doing this? Or perhaps remove using the gdax brokerage model?

Common/Brokerages/DefaultBrokerageModel.cs
```diff
@@ -258,5 +257,16 @@ public virtual ISettlementModel GetSettlementModel(Security security, AccountTyp
...
+        /// <summary>
+        /// Gets a new buying power model for the security, returning the default model with the security's configured leverage.
+        /// For cash accounts, leverage = 1 is used.
+        /// </summary>
+        /// <param name="security">The security to get a buying power model for</param>
+        /// <param name="accountType">The account type</param>
+        /// <returns>The buying power model for this brokerage/security</returns>
+        public virtual IBuyingPowerModel GetBuyingPowerModel(Security security, AccountType accountType)
+        {
+            return security.BuyingPowerModel;
```
mchandschuh: I believe all of the other `GetXXXModel` methods return a new instance. Any particular reason why this one doesn't?

Common/Brokerages/IBrokerageModel.cs
```diff
@@ -146,18 +153,18 @@ public static IBrokerageModel Create(BrokerageName brokerage, AccountType accoun
...
+
                 case BrokerageName.FxcmBrokerage:
                     return new FxcmBrokerageModel(accountType);
 
                 case BrokerageName.Bitfinex:
                     return new DefaultBrokerageModel(accountType);
 
                 case BrokerageName.GDAX:
-                    return new GDAXBrokerageModel(accountType);
+                    return new GDAXBrokerageModel();
```
mchandschuh: Shouldn't we still be passing in the `accountType` here? If not, users will never get the warning if they use gdax w/ margin, and perhaps worse, if the specify gdax w/ margin it won't use margin, but would use the cash models.

Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
StefanoRaggi: I think removing the `SetBrokerageModel` call is not enough, because now the `CashBuyingPowerModel` is the default model for `Crypto` security type.
I'll add a comment explaining why this line is required in this algorithm.

Common/Brokerages/DefaultBrokerageModel.cs
StefanoRaggi: I just did it to avoid the `switch` statement on the security type, only to return the default models already declared in the constructors.
I'll add the switch for consistency with the other methods in the class.

Common/Brokerages/IBrokerageModel.cs
StefanoRaggi: As GDAX does not currently support margin trading, we have two options:
1. silently default to `CashBuyingPowerModel`, ignoring the `accountType` argument
2. throw an exception if `AccountType.Margin` is passed

I chose option 1 only to avoid forcing people to change the argument to `AccountType.Cash` in existing algorithms.

Algorithm.CSharp/FractionalQuantityRegressionAlgorithm.cs
mchandschuh: thank you :)

Issue #1385, opened 2017-12-11
Add PyTorch Pyro Machine Learning Libraries


jaredbroad:
http://pytorch.org/

http://pyro.ai/

AlexCatarino:
PR #1526 does *not* close this issue, because we need to add a test in the [PythonPackageTestAlgorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Python/PythonPackageTestAlgorithm.py).

AlexCatarino:
Closed by #1609

Issue #1369, opened 2017-12-05
Add tensorforce to supported python libraries


Custos123:
Tensorforce is a reinforcement learning library of tensorflow. http://tensorforce.readthedocs.io/en/latest/index.html

Please add to whitelist of python libraries, as this would allow reinforcement learning.

AlexCatarino:
PR #1526 does *not* close this issue, because we need to add a test in the [PythonPackageTestAlgorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Python/PythonPackageTestAlgorithm.py).

AlexCatarino:
Closed by #1609

Issue #1349, opened 2017-11-27
Add NLTK library to whitelist of python libraries


Custos123:
Hi,
it would be quite useful to be able to use NLTK for NLP purposes. Please add to whitelist of python libraries.

best regards

AlexCatarino:
Link to the [NLTK](http://www.nltk.org/) library.
We need to add it to the list of libraries in the [foundation](https://github.com/QuantConnect/Lean/blob/master/DockerfileLeanFoundation#L22) dockerfile and add a test in the [PythonPackageTestAlgorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Python/PythonPackageTestAlgorithm.py).
Here is a recent PR from @DMTSource that implements a third-party library: #1304 

It would be great if a working algorithm could also be added!

AlexCatarino:
PR #1526 does *not* close this issue, because we need to add a test in the [PythonPackageTestAlgorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Python/PythonPackageTestAlgorithm.py).

AlexCatarino:
Closed by #1609

PR #1526, opened 2018-02-02
Upgrades python support to 3.6 from 2.7

AlexCatarino:
In this pull request we propose changes that enables python 3.6 in Lean.
- Adds log to display the python version the algorithm is using.
- Fixes python algorithms that were failing because of small subtleties like leading zeroes.
- Updates pythonnet with a version compiled with python 3.6 flags

Also:
- Changes in DockerfileFoundation: we now use miniconda to manage the python environment. Took the opportunity to add NTLK (#1349), Tensorforce (#1369) and PyTorch/Pyro (#1385).
- Changes readme in Algorithm.Python to show steps to install miniconda

### DIFF
--- a/Algorithm.Python/BasicTemplateOptionsHistoryAlgorithm.py
+++ b/Algorithm.Python/BasicTemplateOptionsHistoryAlgorithm.py
@@ -69,7 +69,7 @@ class BasicTemplateOptionsHistoryAlgorithm(QCAlgorithm):
                     contract.ImpliedVolatility))
 
     def OnSecuritiesChanged(self, changes):
-        if changes == SecurityChanges.None: return
+        if changes == None: return
         for change in changes.AddedSecurities:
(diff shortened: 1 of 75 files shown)


Algorithm.Python/BasicTemplateOptionsHistoryAlgorithm.py
```diff
@@ -69,7 +69,7 @@ def OnData(self,slice):
                     contract.ImpliedVolatility))
 
     def OnSecuritiesChanged(self, changes):
-        if changes == SecurityChanges.None: return
+        if changes == None: return
```
mchandschuh: Does this evaluate properly? We could add `SecurityChanges.Count` and instead check `changes.Count == 0` ?

AlexCatarino: I think we could remove this one here, because if `changes == SecurityChanges.None`, it will loop throw empty lists below, but `changes.Count == 0` is a good solution.

PR #1531, opened 2018-02-05
Fix out-of-memory and Too Many Open Files errors when running Lean under Mono in Linux.

theGOTOguy:
Fixes out-of-memory and Too Many Open Files errors when running Lean under Mono in Linux.

Because the C# garbage collector is implemented differently in Mono under Linux than in Windows, we have found that running on very fast systems as well as running with very large data files can lead to either a Too Many Open Files error or seg-faults as the system runs out of memory.

The solution to this is two-fold:

1)  There's no reason that it should be necessary to introduce the non-determinism of periodically running CleanCache.  Instead, let's just run CleanCache when we actually need it.  This will likely be more efficient than the previous method.
2)  Limit the number of simultaneously open, cached files allowed.

It is noteworthy that even removing the MemoryStream and letting ZipInputStream handle all I/O does not fix this issue.

### DIFF
--- a/Engine/DataFeeds/ZipDataCacheProvider.cs
+++ b/Engine/DataFeeds/ZipDataCacheProvider.cs
@@ -29,10 +29,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds
     /// </summary>
     public class ZipDataCacheProvider : IDataCacheProvider
     {
+        private const int MaxFilesToCache = 10;
         private const int CacheSeconds = 10;
 
-        // ZipArchive cache used by the class
+        // ZipArchive cache used by the class.  Store the last time a file was accessed
+        // so that we can expire ones that haven't been used in a while.
         private readonly ConcurrentDictionary<string, CachedZipFile> _zipFileCache = new ConcurrentDictionary<string, CachedZipFile>();
+
         private DateTime _lastCacheScan = DateTime.MinValue;
         private readonly IDataProvider _dataProvider;
 
@@ -66,16 +69,10 @@ namespace QuantConnect.Lean.Engine.DataFeeds
             {
                 Stream stream = null;
 
-                // scan the cache once every 3 seconds
-                if (_lastCacheScan == DateTime.MinValue || _lastCacheScan < DateTime.Now.AddSeconds(-3))
-                {
-                    CleanCache();
-                }
-
                 try
                 {
                     CachedZipFile existingEntry;
-                    if (!_zipFileCache.TryGetValue(filename, out existingEntry))
+                    if (!GetFromCache(filename, out existingEntry))
                     {
                         var dataStream = _dataProvider.Fetch(filename);
 
@@ -90,7 +87,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                                     stream = CreateStream(newItem.ZipFile, entryName, filename);
                                 }
 
-                                _zipFileCache.TryAdd(filename, newItem);
+                                AddToCache(filename, newItem);
                             }
                             catch (Exception exception)
                             {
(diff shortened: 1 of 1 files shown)


jaredbroad:
Interesting fix thank you @theGOTOguy - how was the memory after this change? Was the performance greatly impacted?

theGOTOguy:
I haven't noticed any significant performance change regardless of how I handle the Zip files.  Caching them in MemoryStream, loading them directly from ZipInputStream, or uncaching them more or less frequentlly doesn't seem to cause my backtests to go significantly faster or slower, but this is just empirical observation and not in any way rigorous.  I'm sure it makes some sort of difference, but compared to the amount of time my algorithms take to crunch on any given data point it may not matter all that much.

My memory use is currently hovering right around 12GB with this fix in place.  What I find puzzling, though, is that I'm unable to actually catch the spikes in memory use that cause me to seg fault in the act.  I can sit there watching top and also have gdb running and catch the SIGSEGV the instant it happens, and yet according to top memory use never spiked over 20%.  What's more, the timing of the SIGSEGV was totally random.  Sometimes it would happen right away, sometimes halfway through the backtest, and sometimes at the very last day or not at all.  I assume this must have something to do with how the memory use for things waiting to be garbage collected is counted.

What is clear, though, is that my backtests have gone from seg faulting about half the time to never at all with this fix in place.

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -164,6 +161,46 @@ public void Dispose()
             _zipFileCache.Clear();
         }
 
+        /// <summary>
+        /// Add a file from the cache, bumping the least frequently accessed
+        /// file if necessary.
+        /// </summary>
+        private void AddToCache(string filename, CachedZipFile zipFile) {
+            // If the cache is full, we clean the cache.
+            if (_zipFileCache.Count == MaxFilesToCache) {
```
mchandschuh: Bracing style: In general, please try to follow the style of the existing code.

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -265,7 +302,15 @@ public CachedZipFile(ZipFile data, string key)
...
-            _dateCached = DateTime.Now;
+            DateCached = DateTime.Now;
+        }
+
+        /// <summary>
+        /// Method used to update the last time this file was accessed.
+        /// </summary>
+        public void Refresh()
+        {
+            DateCached = DateTime.Now;
```
mchandschuh: Prefer usage of `DateTime.UtcNow` where possible.

To be true, `DateTime.Now` documentation does state that the system local time zone is cached upon application start, so if the system time zone changed it shouldn't affect `DateTime.Now` (unless `TimeZoneInfo.ClearCachedData()` is called). That being said, I still find it safer (and more consistent) to use UTC where possible.

theGOTOguy:
Just a heads-up, this hasn't completely ended my seg fault issue.  I just had another one.  I'm going to try to update this PR to be a bullet-proof solution before addressing your comments.

It appears to me that the Mono garbage collector is simply choosing not to run until the very last possible moment, causing random seg faults when the ZipDataCacheProvider tries to load a large file when its thread fails to garbage collect in time.

theGOTOguy:
I wanted to give an update on this PR:

In order to actually stop the seg fault, in addition to taking the step in this PR that limits the number of simultaneously allowed open files I also have to force the garbage collector to run when .dispose() is called on the stream.  Simply forcing the garbage collector to run when cleanup is periodically called in the old code isn't enough.

My belief is that the mono garbage collector is only being called when memory is already critically low.  Because it runs in a separate thread, the garbage collector begins freeing memory but because we're using such large chunks at a time it is unable to complete before the memory is needed, and we seg-fault.

I'm not sure that forcing garbage collection within Lean is the right solution, because there may be other places where we could seg fault when allocating large chunks of memory and they would remain broken.

My path forward is going to be to experiment with --gc=sgen versus --gc=boehm and see if either of those helps.  Unfortunately, it takes many runs to be sure whether I've actually fixed the seg fault or whether I've just made it rarer, so it may be a while until I can make a recommendation there.

jaredbroad:
Thanks @theGOTOguy; we're doing a general review of all the LEAN critical-path pieces now (https://github.com/QuantConnect/Lean/pull/1553/files) and wrapping everything in using statements to 1) remove some memory leaks and 2) help the GC know when to dispose of the objects. My hunch is the gen-2 collection doesn't happen until its critical; so by making everything dispose in gen-1 we can save a lot of hassle later. 

What version of Mono are you testing with?

theGOTOguy:
I'm using mono 5.8.0.108, so the default collector is sgen.

jaredbroad:
LEAN foundation is on [4.6.1.5](https://github.com/QuantConnect/Lean/blob/master/DockerfileLeanFoundation#L54) - we had mono-explosion issues with later versions; the hunch was when mono moved to using open sourced MS thread stack they introduced some bugs. This might be resolved in later versions of Mono but we haven't tested.

jaredbroad:
@theGOTOguy please check out master and let us know if the issue persists for you. We discovered and fixed an issue where _all_ files were left open over backtest. https://github.com/QuantConnect/Lean/pull/1565

Possibly still worth opening the zips as streams as you've done here // capping cache to limit RAM but would be good to confirm the file handle issue is resolved.

theGOTOguy:
I'm testing at head with sgen and boehm garbage collectors now.  So far I'm not noticing a huge difference in terms of absolute memory use, but I also haven't seen any seg faults for want of memory yet.

theGOTOguy:
In 24 processor-hours of time using the sgen garbage collector I haven't managed any seg-faults at head with an algorithm which consumes roughly 1GB of data per minute in backtests.  I'll let you know if I continue to have any issue.

Thanks for the fix!  No other off-the-shelf platform we've tried comes anywhere near the performance and stability of Lean when we're pounding it with data at this rate.  This PR can probably safely be closed.

jaredbroad:
Thanks @theGOTOguy; reframed it as a memory reduction feature to look at sometime in the future.

theGOTOguy:
Scratch that, looks like it blew up again last night.

PR #1539, opened 2018-02-06
Subscription sync interface

mchandschuh:
Built on top of https://github.com/QuantConnect/Lean/pull/1538

Extracts an interface for synchronizing subscriptions which can support easily swapping out the synchronization implementation. The synchronizer implementation is expected to manage the frontier of the subscriptions provided.

### DIFF
--- a/Engine/Results/BacktestingResultHandler.cs
+++ b/Engine/Results/BacktestingResultHandler.cs
@@ -534,6 +534,14 @@ namespace QuantConnect.Lean.Engine.Results
                 Console.SetOut(debug);
                 Console.SetError(error);
             }
+            else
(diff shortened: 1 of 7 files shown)


Engine/Results/BacktestingResultHandler.cs
```diff
@@ -534,6 +534,14 @@ public void SetAlgorithm(IAlgorithm algorithm)
                 Console.SetOut(debug);
                 Console.SetError(error);
             }
+            else
+            {
+                // we need to forward Console.Write messages to the standard Log functions
```
jaredbroad: Nice catch

PR #1541, opened 2018-02-07
GDAX Downloader Failing on Scientific Notation Returned from Snapshot Service

cmoski:
Hey all,

Was testing on the GDAX downloader to grab some LTC-USD historical information. GDAX now returns volume (if fractional) in scientific notation in its JSON response.

Added  System.Globalization.NumberStyles.Float to handle the response.


jaredbroad:
Welcome to LEAN @cmoski! Does method work if not fractional? Does it return in the same format for BTC ETH as well?

cmoski:
Thank you Jared! Excited to be here -- just learning my way around. (Its literally my first time here. My background is primarily in C++, I apologize for anachronisms)

It appears that it works just fine for BTC-USD, BTC-LTC, and ETH-USD pairs through my testing.  There is one further outstanding issue, and that is GDAX requires a hyphen between the pair name (LTC-USD for example) while Lean seems to exception on the USD root for the auto-generated data directory name. 

(I'll probably submit another small patch for that issue after I test it further, renamed the directory without thinking about it but it probably would be useful to anyone running it for the first time..)

jaredbroad:
Awesome thank you for the further testing.

For your interest we separate tickers into "LEAN tickers" and "Brokerage tickers". We store/use all the tickers to a LEAN format (i.e. without the hyphen) -- and in each brokerage implementation map the LEAN format back to the brokerage (e.g. [OandaSymbolMapper](https://github.com/QuantConnect/Lean/blob/master/Brokerages/Oanda/OandaSymbolMapper.cs) class). This is so people can use the same ticker in their algorithm regardless of brokerage to make the algorithms broker agnostic. 

It sounds like we're not mapping that ticker back correctly,

PR #1544, opened 2018-02-08
Fix FineFundamental universe selection resource leak

StefanoRaggi:
The `FineFundamentalSubscriptionEnumeratorFactory` used in combination with `SingleEntryDataCacheProvider` and `DefaultDataProvider` was never disposing of the underlying file stream, leaving files open.


jaredbroad:
@theGOTOguy btw -- doing a full review of the PR you submitted and finding similar leaks everywhere. @StefanoRaggi is patching them 1 by 1 now,

PR #1547, opened 2018-02-09
Fixed EndTime/Period on QuoteBar

Worthy7:
When a consolidator is used, the produced `QuoteBar` does not have the Period/EndTime set, and thus always appears as a 1 minute bar. I believe this is exclusive to QuoteBar Consolidators.

This is my fix, I can only confirm that it fixed my problem but have tried to fix all of the aggregators too, to try and save some work for people. It needs a review from someone in the know.

### DIFF
--- a/Tests/Common/Data/QuoteBarConsolidatorTests.cs
+++ b/Tests/Common/Data/QuoteBarConsolidatorTests.cs
@@ -102,7 +102,6 @@ namespace QuantConnect.Tests.Common.Data
             Assert.AreEqual(bar3.LastBidSize, quoteBar.LastBidSize);
             Assert.AreEqual(bar4.LastAskSize, quoteBar.LastAskSize);
             Assert.AreEqual(bar1.Value, quoteBar.Value);
-            Assert.AreEqual(bar4.EndTime, quoteBar.EndTime);
         }
 
         [Test]
--- a/Tests/Common/Data/QuoteBarConsolidatorTests.cs
+++ b/Tests/Common/Data/QuoteBarConsolidatorTests.cs
@@ -102,7 +102,7 @@ namespace QuantConnect.Tests.Common.Data
             Assert.AreEqual(bar3.LastBidSize, quoteBar.LastBidSize);
             Assert.AreEqual(bar4.LastAskSize, quoteBar.LastAskSize);
             Assert.AreEqual(bar1.Value, quoteBar.Value);
-            Assert.AreEqual(time + TimeSpan.FromMinutes(4), quoteBar.EndTime);
+            Assert.AreEqual(bar4.EndTime, quoteBar.EndTime);
         }
 
         [Test]
--- a/Tests/Common/Data/QuoteBarConsolidatorTests.cs
+++ b/Tests/Common/Data/QuoteBarConsolidatorTests.cs
@@ -142,6 +142,7 @@ namespace QuantConnect.Tests.Common.Data
                 Period = period
             };
             creator.Update(bar2);
+            Assert.IsNull(quoteBar);
 
             // pushing another bar to force the fire
             var bar3 = new QuoteBar
--- a/Tests/Common/Data/QuoteBarConsolidatorTests.cs
+++ b/Tests/Common/Data/QuoteBarConsolidatorTests.cs
@@ -24,7 +24,7 @@ namespace QuantConnect.Tests.Common.Data
     public class QuoteBarConsolidatorTests
     {
         [Test]
-        public void AggregatesNewQuoteBarProperly()
+        public void AggregatesNewCountQuoteBarProperly()
         {
             QuoteBar quoteBar = null;
             var creator = new QuoteBarConsolidator(4);
@@ -34,6 +34,7 @@ namespace QuantConnect.Tests.Common.Data
             };
 
             var time = DateTime.Today;
+            var period = TimeSpan.FromMinutes(1);
             var bar1 = new QuoteBar
             {
                 Time = time,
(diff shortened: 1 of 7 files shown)


Worthy7:
Ok, one test is failing, the `GenerateDailyAndHourlyFutureDataFromMinutes` test.
The reason it is failing, is that at the end when it tries the `Tick`->`OpenInterest` consolidation for Daily data, it creates a working bar, and my patch will try to assign `EndTime` properly. Yet, because no `EndTime` property is defined on `OpenInterest`, it uses the `BaseData.EndTime` instead, which actually sets `Time`. This makes `EndTime` and `Time` always the same, so when it comes to comparison (to see if the working bar is ready to fire) it always returns false, as the distance is never over a day to the next test input bar. That's the problem, looks like a design bug, don't know how to go about fixing it.

mchandschuh:
The working bar's end time should be set to the time the working bar is supposed to be emitted. When using a period based consolidator, the end time should be the first data point's time + the requested period. When using a count based consolidator, the end time should equal the end time of the last bar in the count.

mchandschuh:
@Worthy7 -- Can we create a test that showcases the failure you speak of (in the TDD red/green/refactor style).

Also, after a quick glance, it looks like the `QuoteBarConsolidator` should be using logic similar as defined in the `TradeBarConsolidator` [here](https://github.com/QuantConnect/Lean/blob/master/Common/Data/Consolidators/TradeBarConsolidator.cs#L89) and [here](https://github.com/QuantConnect/Lean/blob/master/Common/Data/Consolidators/TradeBarConsolidator.cs#L97), although, it looks like this doesn't properly handle the count case, where the working bar's end time should be set to the end time of each new data piece, so when emitted, it has the end time of the last piece of data in the count.

Worthy7:
@mchandschuh  sorry for the delay.

Before I make a test showcase of that specific error, I want to actually confirm the point of a Tick -> OpenInterest consolidator - what is this supposed to be doing? `OpenInterest` actually inherits `Tick` so really, what is this doing? What's it's use case? If we consolidate Tick's into a Tick, it's goes to lose it's Time dimension.

mchandschuh:
@Worthy7 - Any particular reason for the *interest* in the `OpenInterestConsolidator`? This consolidator is used by some of our backend processing and it's behavior is as intended. In reality, open interest information is disseminated each morning at about 6:30AM EST by OPRA, so this consolidator is little more than a shim that allows us to process the open interest data in the same way we process other data that *does* need to be consolidated, such as price data.

Given the name of the PR, I would expect the focus to be on the obvious bug of the `QuoteBarConsolidator` not properly managing the period of the bar during consolidation as is done in the `TradeBarConsolidator`. Also, as mentioned above, both of these consolidators don't appear to properly set the working bar's end time properly in the 'count' case.

Worthy7:
@mchandschuh My interest is that this test failing may be a code smell. But you're right, it is unrelated, so instead I'll revert my changes, write up some proper tests for QuoteBar consolidators like you asked, and we'll go from there, hopefully I'll have it all done by (my) today.

Worthy7:
QuoteBar working. Tradebar is actually fine because of the +=:
`if (!IsTimeBased) workingBar.Period += data.Period;`

I'll throw the test for TradeBar on here too just for futures sake.

Worthy7:
@mchandschuh Sorry, I missed your comment about using the same logic. I'm not using the same logic as TradeBar because I was worried about if a bar is missed/skipped for some reason, like a live data network issue. I feel it's safer to use absolute values over relative values where possible, but if you prefer I can change it.

Tests/Common/Data/QuoteBarConsolidatorTests.cs
```diff
@@ -97,6 +102,86 @@ public void AggregatesNewQuoteBarProperly()
...
+                LastBidSize = 3,
+                Ask = new Bar(2.2m, 4.4m, 3.3m, 3.3m),
+                LastAskSize = 0,
+                Value = 1,
+                Period = period
+            };
+            creator.Update(bar2);
+
+            // pushing another bar to force the fire
+            var bar3 = new QuoteBar
```
mchandschuh: Should we have an `Assert.IsNull(quoteBar)` after `creator.Update(bar2)` ?

Worthy7: Yes there absolutely should be - and it fails when I included it. I decided not to bring that up here because it is a separate issue that I really don't understand. But try it yourself.
I mentioned this with Jared and gave up, because since that test will fail, that should mean that bars actually get delayed when emitted, but in live I tested it and they pop out on time. Is live data ALWAYS tick or something? So it is still delayed but only by 1 tick?  https://quantconnect.slack.com/archives/C8KNL9LEQ/p1518574661000015

Worthy7: @mchandschuh Ah sorry, I `Assert.IsNotNull(quoteBar)`. Which is what ^ is referring to.
Why Assert.IsNull(quoteBar)? Surely after 2 bars it should be emitted? I don't understand why it doesn't so I left it out.

Worthy7:
I tested some more, and I can make `TradeBarConsolidator` fail by using `TradeBarConsolidator(TimeSpan.FromMinutes(5), 3)`. My understand is this is supposed to emit after 5 minutes, or after it receives 3 inputs.

There is no nice way to handle this in the current implementation, and I don't know what the specification is so *I* cannot do this without some documentation. I'm going to make QuoteBar's just work the same way, then I'm leaving this as done.

Worthy7:
@mchandschuh You're right, but either way something breaks. We can start a separate issue about count consolidators if you like, but we need a failing test first :)

mchandschuh:
So the reason why things appear to always be delayed when using period is because the only way to *confirm* that the period has elapsed is to wait for a data point *after* the period. Now, technically, this may only really apply to tick data since you can have multiple data points at a single time. When deploying in quantconnect, we try to use the count constructor to avoid this potential delay.

Actually, this may be solvable now. Not too long ago we added an `IDataConsolidator.Scan` method. The spirit of this method is to let consolidators know that processing of data for the specified time has finished. This would allow us to emit the consolidated point exactly at the end of the period. Doing so more digging, this method appears to be properly invoked from the `AlgorithmManager` which also removes the delay.

TL;DR;  If you want to remove the assert is null from the test, add a `consolidator.Scan(time)` and that should force the working bar to be emitted.

mchandschuh:
@Worthy7 - I think with the usage of `IDataConsolidator.Scan(DateTime)` there's no bug and/or refactoring that's required. I was saying that if you want to get your expected behavior in your test then you must call `IDataConsolidator.Scan(DateTime)`. This is what happens in the algorithm manager and why consolidators aren't delayed when running the engine but appear delayed when running in your unit test.

Worthy7:
@mchandschuh Ok fair enough - I kinda disagree with the design of it, but anyway it works - not an issue :) We gonna merge this then?

PR #1548, opened 2018-02-09
Adds overloads with PyObject to Logging and Debug methods

AlexCatarino:
We enable Logging and Debug methods to accept python objects to avoid the need of calling the str method.
Those objects are safely converted into string objects.


jaredbroad:
Nice work =) That'll help a lot of people.

Issue #1482, opened 2018-01-16
Futures and custom data Python Exception 


AlexCatarino:
When running a python algorithm with custom data, we get the following exception:
```
Runtime Error: Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: The best overloaded method match for `System.Collections.Generic.KeyValuePair<QuantConnect.Symbol,QuantConnect.Data.BaseData>.KeyValuePair(QuantConnect.Symbol, QuantConnect.Data.BaseData)' has some invalid arguments
  at (wrapper dynamic-method) System.Object:CallSite.Target (System.Runtime.CompilerServices.Closure, System.Runtime.CompilerServices.CallSite, object, QuantConnect.Data.Slice)
  at QuantConnect.AlgorithmFactory.Python.Wrappers.AlgorithmPythonWrapper.OnData (QuantConnect.Data.Slice slice) [0x0005f] in <164ef4aed6104058a2ba63f623af1f30>:0 
```
it is throwing at [AlgorithmPythonWrapper.OnData L716](https://github.com/QuantConnect/Lean/blob/master/AlgorithmFactory/Python/Wrappers/AlgorithmPythonWrapper.cs#L716).

Here is the code to reproduce it:
```python
from QuantConnect.Python import PythonQuandl
from datetime import timedelta

class Algo(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2018, 1, 1)
        future = self.AddFuture(Futures.Indices.SP500EMini)
        future.SetFilter(timedelta(0), timedelta(182))              
        self.AddData(QuandlFuture, "CBOE/VIX", Resolution.Daily)

    def OnData(self, slice):
        pass

class QuandlFuture(PythonQuandl):
    def __init__(self):
        self.ValueColumnName = "vix close"
```

jaredbroad:
fixed in #1554

PR #1554, opened 2018-02-09
Creates wrapper for Slice (python custom data)

AlexCatarino:
Python algorithms with custom data requires an operation that converts a dictionary key into a attribute. In the current implementation the Slice object was converted into a python dictionary. This was not optimal, since we just need to make this conversion when the value of a key in the Slice is accessed.
This implementation proposes a wrapper for the Slice object, PythonSlice, that would just perform the operation described above when needed.
- Fixes custom data algos to reflect changes from previous commit
- Fixes issue #1482

### DIFF
--- a/Common/Python/PythonSlice.cs
+++ b/Common/Python/PythonSlice.cs
@@ -28,8 +28,29 @@ namespace QuantConnect.Python
     public class PythonSlice : Slice
     {
         private Slice _slice;
-        private static PyObject _converter;
- 
+        private static readonly PyObject _converter;
+
+        static PythonSlice()
+        {
+            // Python Data class: Converts custom data (PythonData) into a python object'''
+            _converter = PythonEngine.ModuleFromString("converter",
+                "import decimal\n" +
+
+                "class Data(object):\n" +
+                "    def __init__(self, data):\n" +
+                "        self.data = data\n" +
+                "        members = [attr for attr in dir(data) if not callable(attr) and not attr.startswith(\"__\")]\n" +
+                "        for member in members:\n" +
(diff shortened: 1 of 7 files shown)


Common/Python/PythonSlice.cs
```diff
@@ -0,0 +1,159 @@
...
+                return data;
+            }
+        }
+
+        /// <summary>
+        /// Determines whether this instance contains data for the specified symbol
+        /// </summary>
+        /// <param name="symbol">The symbol we seek data for</param>
+        /// <returns>True if this instance contains data for the symbol, false otherwise</returns>
+        public new bool ContainsKey(Symbol symbol)
```
jaredbroad: 👍

Common/Python/PythonSlice.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <returns>True if data was found, false otherwise</returns>
+        public new bool TryGetValue(Symbol symbol, out dynamic data)
+        {
+            return _slice.TryGetValue(symbol, out data);
+        }
+
+        /// <summary>
+        /// Sets the converter
+        /// </summary>
+        public static void SetConverter()
```
mchandschuh: Is this class ever valid if this is not set? Perhaps instead of forcing consumers to poke this type, the type could handle this initialization itself? I've seen a similar pattern elsewhere, where we have a `SetXXX` method that accepts no arguments. Typically, with a `SetXXX` method, we'll pass the XXX to be set. It appears this is just a hook for some required static initialization.  Static constructors are often a good solution for static initialization issues.

PR #1555, opened 2018-02-10
Internationalization of Base Account Currency

espirulina:
Allow to set base account currency from QCAlgorithm class


mchandschuh:
I'm fairly certain there exists other places in the code with a hard-coded `"USA"` string. Also, since the cashbook is initialized in the beginning, it's already created a cash entry for `"USA"` as the default currency. There may be other components that have already checked the default currency as well.

I'm also curious as to what testing has been performed around this change.

jaredbroad:
Closing for now as discussed in slack this needs to be instance based cashbook objects so we can plan for multiple brokerages and local multi-LEAN optimization.

PR #1559, opened 2018-02-12
Remove orderby timeslice create

mchandschuh:
#### Move security type ordering to SubscriptionCollection

This will change the order in which we enumerate subscriptions so we don't need to re-sort inside of `TimeSlice.Create`

#### Add Subscription.ToString override

Makes debugging much easier when we can easily see what a subscription is for

### DIFF
--- a/Engine/DataFeeds/SubscriptionCollection.cs
+++ b/Engine/DataFeeds/SubscriptionCollection.cs
@@ -119,9 +119,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
         private void SortSubscriptions()
         {
+            // it's important that we enumerate underlying securities before derivatives to this end,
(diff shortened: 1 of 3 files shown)


Engine/DataFeeds/SubscriptionCollection.cs
```diff
@@ -121,7 +121,8 @@ private void SortSubscriptions()
         {
             _subscriptionsByTickType = _subscriptions
                 .Select(x => x.Value)
-                .OrderBy(x => x.Configuration.TickType)
+                .OrderBy(x => x.Configuration.SecurityType)
+                .ThenBy(x => x.Configuration.TickType)
```
StefanoRaggi: Nice speed optimization!

PR #1562, opened 2018-02-12
Internationalization of Account Currency

espirulina:
This PR aims to have dynamic Account Currency:
- replaced _static string _ by a property on CashBook.AccountCurrency
- added AccountCurrency property to Brokerages passed in constructor
- all references are now from Cashbook **instances** instead of static class field
- some references are with helper:
                   - Portfolio.CashBook.AccountCurrency
                   - _algorithm.Portfolio.CashBook.AccountCurrency


Algorithm/QCAlgorithm.Python.cs
```diff
@@ -50,11 +50,12 @@ public void SetPandasConverter()
         /// </summary>
         /// <param name="type">Data source type</param>
         /// <param name="symbol">Key/Symbol for data</param>
+        /// <param name="accountCurrency">Account currency</param>
         /// <param name="resolution">Resolution of the data</param>
         /// <returns>The new <see cref="Security"/></returns>
-        public Security AddData(PyObject type, string symbol, Resolution resolution = Resolution.Minute)
+        public Security AddData(PyObject type, string symbol, string accountCurrency, Resolution resolution = Resolution.Minute)
```
mchandschuh: I don't think we should be updating the `AddData` methods to accept account currency. The account currency can be resolved via `Portoflio.CashBook.AccountCurrency`

Algorithm/QCAlgorithm.cs
```diff
@@ -159,6 +159,22 @@ public QCAlgorithm()
...
+        {
+            CashBook.AccountCurrency = accountCurrency;
+        }
+
+        /// <summary>
+        /// Set base account currency
+        /// </summary>
+        protected void SetAccountCurrency(string accountCurrency)
+        {
+            Portfolio.CashBook.AccountCurrency = accountCurrency;
```
mchandschuh: Due to temporal issues in initialization, we might need to initialize this _before_ the algorithm's initialize is called. Otherwise some items may be initialized with the default account currency, and then the account currency could be changed causing other items to be initialized with the newly configured account currency.

Algorithm/QCAlgorithm.cs
```diff
@@ -1458,9 +1474,9 @@ public Option AddOption(string underlying, Resolution resolution = Resolution.Mi
             }
 
             var marketHoursEntry = MarketHoursDatabase.GetEntry(market, underlying, SecurityType.Option);
-            var symbolProperties = _symbolPropertiesDatabase.GetSymbolProperties(market, underlying, SecurityType.Option, CashBook.AccountCurrency);
+            var symbolProperties = _symbolPropertiesDatabase.GetSymbolProperties(market, underlying, SecurityType.Option, Portfolio.CashBook.AccountCurrency);
             var canonicalSecurity = (Option) SecurityManager.CreateSecurity(typeof(ZipEntryName), Portfolio, SubscriptionManager,
-                marketHoursEntry.ExchangeHours, marketHoursEntry.DataTimeZone, symbolProperties, SecurityInitializer, canonicalSymbol, resolution,
+                marketHoursEntry.ExchangeHours, marketHoursEntry.DataTimeZone, symbolProperties, SecurityInitializer, canonicalSymbol, resolution, Portfolio.CashBook.AccountCurrency,
```
mchandschuh: This method already accepts the `Portoflio` object, so no need to pass the account currency explicitly. This comment applies to many other places in this PR as well.

Brokerages/Brokerage.cs
```diff
@@ -54,6 +54,11 @@ public abstract class Brokerage : IBrokerage
         /// </summary>
         public string Name { get; private set; }
 
+        /// <summary>
+        /// Base account currency
+        /// </summary>
+        public string AccountCurrency { get; set; }
```
mchandschuh: This doesn't appear to be used in this class. If individual brokerages require this value, then I think it would need to be specified via the individual brokerage factories. This might lead us to require the account currency be part of the algorithm job packet. Setting the account currency in the job packet would also remove the temporal issues during initialization.

Brokerages/GDAX/GDAXBrokerage.cs
```diff
@@ -271,18 +271,18 @@ public override List<Cash> GetCashBalance()
...
                         var rate = GetConversionRate(item.Currency);
-                        list.Add(new Cash(item.Currency.ToUpper(), item.Balance, rate));
+                        list.Add(new Cash(item.Currency.ToUpper(), item.Balance, rate, _algorithm.Portfolio.CashBook.AccountCurrency));
                     }
                     else
                     {
                         var tick = GetTick(Symbol.Create(item.Currency + "USD", SecurityType.Crypto, Market.GDAX));
 
-                        list.Add(new Cash(item.Currency.ToUpper(), item.Balance, tick.Price));
+                        list.Add(new Cash(item.Currency.ToUpper(), item.Balance, tick.Price, _algorithm.Portfolio.CashBook.AccountCurrency));
```
mchandschuh: The line above this is fetching the conversion rate to USD, but I think it should be fetching the conversion rate to the account currency?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -959,6 +959,169 @@ private IEnumerable<ContractDetails> FindContracts(Contract contract)
             return contractDetails;
         }
 
+        /// <summary>
+        /// Gets the current conversion rate into Account currency
+        /// </summary>
+        /// <remarks>Synchronous, blocking</remarks>
+        private decimal GetCurrencyConversion(string currency)
```
mchandschuh: Did we simply copy the `GetUsdConversion` method instead of upgrading it? The purpose of `GetUsdConversion` is actually to get the conversion in terms of the account currency, so I don't think we need to versions of this same method.

Common/AlphaRuntimeStatistics.cs
```diff
@@ -60,12 +60,17 @@ public class AlphaRuntimeStatistics
         /// </summary>
         public decimal MeanPopulationEstimatedAlphaValue => TotalAlphasClosed > 0 ? TotalEstimatedAlphaValue / TotalAlphasClosed : 0;
 
+        /// <summary>
+        /// Base account currency
+        /// </summary>
+        public string AccountCurrency { get; internal set; }
```
mchandschuh: Instead of having a setter, the `ToDictionary` method could simply accept the account currency symbol. This can apply to other cases as well. If a method requires the configured account currency, it is sometimes best to add it as an argument. If the class makes judicious use of the account currency, then I would recommend it be a constructor argument. Having dangling properties that require consumers to set them at some point in the future will inevitably end with someone not initializing the object properly. This is why it's best to force the issue via a constructor argument or a method argument, as it forces the consumer to make a decision.

Common/Data/UniverseSelection/Universe.cs
```diff
@@ -170,6 +170,7 @@ public IEnumerable<Symbol> PerformSelection(DateTime utcTime, BaseDataCollection
         /// <param name="algorithm">The algorithm instance</param>
         /// <param name="marketHoursDatabase">The market hours database</param>
         /// <param name="symbolPropertiesDatabase">The symbol properties database</param>
+        /// <param name="accountCurrency">Base account currency</param>
```
mchandschuh: Docs added here (and elsewhere) but no matching parameter. Also, if we have access to the entire algorithm instance, then we can access `algorithm.Portfolio.CashBook.AccountCurrency` (to [Demeter's](https://en.wikipedia.org/wiki/Law_of_Demeter) demise of course :P )

Common/Orders/OrderEvent.cs
```diff
@@ -174,7 +182,7 @@ public override string ToString()
                 : string.Format("Time: {0} OrderID: {1} Symbol: {2} Status: {3} Quantity: {4} FillPrice: {5} {6}", UtcTime, OrderId, Symbol.Value, Status, FillQuantity, FillPrice.SmartRounding(), FillPriceCurrency);
 
             // attach the order fee so it ends up in logs properly
-            if (OrderFee != 0m) message += string.Format(" OrderFee: {0} {1}", OrderFee, CashBook.AccountCurrency);
+            if (OrderFee != 0m) message += string.Format(" OrderFee: {0} {1}", OrderFee, AccountCurrency);
```
mchandschuh: There's an interesting hidden issue here, and that is that the fee might _not_ be in units of the account currency. For example, it's very common on crypto exchanges to charge fees in units of the quote currency. So if I trade +10 units of LTCBTC, I'll get charged fees in BTC and not in USD. I think we should probably fix the order fees in a separate PR before this account currency work is completed. We can rephrase the order fee calculation methods to return a cash object which would contain the quantity and units of the fee, as well as a conversion rate to the account currency. This also precludes us from piping the value everywhere as a string and instead more accurately describes the fee.

Common/Securities/Cash.cs
```diff
@@ -70,13 +70,19 @@ public class Cash
         /// </summary>
         public decimal ValueInAccountCurrency => Amount * ConversionRate;
 
+        /// <summary>
+        /// Base account currency
+        /// </summary>
+        public string AccountCurrency { get; internal set; }
```
mchandschuh: Instead of having an account currency property, it may be better to accept the account currency as an argument to the only method that requires the data.

Common/Securities/CashBook.cs
```diff
@@ -31,10 +31,10 @@ namespace QuantConnect.Securities
     public class CashBook : IDictionary<string, Cash>
     {
         /// <summary>
-        /// Gets the base currency used
+        /// Base currency used
         /// </summary>
-        public const string AccountCurrency = "USD";
-
+        public string AccountCurrency { get; set; } = "USD";
```
mchandschuh: I think if we're going to make this instance based, the it should be a required constructor argument to force callers to set it properly. This also sheds light on the initialization issue, that is to say,  I think we need to know the account currency _before_ initializing the algorithm

mchandschuh:
Closing this for now as it's gotten bigger than expected. Is there a way we can apply these changes incrementally in smaller PRs? Perhaps we can have a PR for each class using `CashBook.AccountCurrency` that refactors it to receive the value in a non-static way (ctor arg/method arg).

PR #1564, opened 2018-02-13
Move time zone conversions to data feed thread

mchandschuh:
#### Add TimeZoneOffsetProvider.ConvertToUtc

This is really the only use case for this class, converting a local time to UTC properly. Instead of enforcing this invocation pattern on all consumers, much cleaner to have the type provide this functionality.

#### Move clone and tz conversion operations to DF thread

A large percentage of synchronization time is spent performing time zone conversions. Previously, it was possible that we may compute the same tz conversion multiple times on the same piece of data until it's emitted. This change moves the time zone conversion logic onto the data feed thread right before it puts the data into enqueueable enumerator. In addition, we're not performing the clone operation at the same moment. This removes unnecessary computation while moving the necessary computation to the data feed thread to lighten the load on the synchronization/algorithm thread.

#### Performance Improvement
These changes bring the top speed of the subscription synchronizer over the 1MM ticks/second mark by shifting work to the data feed thread to prevent consuming CPU cycles on the algorithm manager thread.

#### Regression Changes
There are changes in the regression tests due to us no longer emitting an empty time slice at each time zone discontinuity. Every regression algorithm that spans a time zone discontinuity had it's regression statistics updated to reflect this change. I've confirmed that all trades/fills are identical and the discrepancy is due to benchmark sampling at the time zone discontinuities.


Common/TimeZoneOffsetProvider.cs
```diff
@@ -100,6 +100,30 @@ public long GetOffsetTicks(DateTime utcTime)
...
+        /// <summary>
+        /// Converts the specified local time to UTC. This function will advance this offset provider
+        /// </summary>
+        /// <param name="localTime">The local time to be converted to UTC</param>
+        /// <returns>The specified time in UTC</returns>
+        public DateTime ConvertToUtc(DateTime localTime)
+        {
+            // it's important to walk forward to the next time zone discontinuity
+            // to ensure a derministic read. We continue reading with the current
+            // offset until the converted value is
```
mchandschuh: dangling comment

Engine/DataFeeds/SubscriptionSynchronizer.cs
```diff
@@ -162,12 +152,12 @@ public TimeSlice Sync(IEnumerable<Subscription> subscriptions)
...
                         if (earlyBirdTicks == MaxDateTimeTicks)
                         {
-                            earlyBirdTicks = subscription.Current.EndTime.ConvertToUtc(subscription.TimeZone).Ticks;
+                            earlyBirdTicks = subscription.Current.EmitTimeUtc.Ticks;
                         }
                         else
                         {
                             // take the earliest between the next piece of data or the next tz discontinuity
-                            earlyBirdTicks = Math.Min(earlyBirdTicks, Math.Min(subscription.Current.EndTime.Ticks - currentOffsetTicks, offsetProvider.GetNextDiscontinuity()));
+                            earlyBirdTicks = Math.Min(earlyBirdTicks, subscription.Current.EmitTimeUtc.Ticks);
```
mchandschuh: stale comment

PR #1570, opened 2018-02-14
Fix GDAX error "Product not found" when submitting orders

StefanoRaggi:
GDAX API now requires uppercase symbol names, previously we were submitting REST requests with symbols in lowercase.


mark-hennessy:
Thanks!

PR #1585, opened 2018-02-16
Fix CashBuyingPowerModel.HasSufficientBuyingPowerForOrder

### LINKED ISSUE #1584: CashBuyingPower SetHoldings fails for GDAX
Please see a reproducible example in [this gist](https://gist.github.com/multidis/5b1f19411031fc32efd1ce5b2d9bbf25)

`SetHoldings` failure started happening following a recent `AccountType.Margin` to `AccountType.Cash` change. After [this commit](https://github.com/QuantConnect/Lean/commit/5b4a994aa0a8d65f7960fd74038ce13baa7a7f63) the issue will affect anyone using GDAX.

StefanoRaggi:
This method was not taking into account existing holdings for the order security.

Fixes #1584

### DIFF
--- a/Tests/Algorithm/CashModelAlgorithmTradingTests.cs
+++ b/Tests/Algorithm/CashModelAlgorithmTradingTests.cs
@@ -1,0 +1,700 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


Tests/Algorithm/CashModelAlgorithmTradingTests.cs
```diff
@@ -0,0 +1,700 @@
...
+            var actual = algo.CalculateOrderQuantity(_symbol, -0.5m);
+
+            // no shorting allowed
+            Assert.AreEqual(0, actual);
+        }
+
+        [Test]
+        public void SetHoldings_ShortToShorter_PriceRise()
+        {
+            // no initial short position with cash model
```
mchandschuh: we can delete or comment out the entire test, an empty test is of no use

jhenak:
Will this fix resolve limit orders as well?  I see that the change is limited to OrderType.Market.

StefanoRaggi:
@jhenak The bug was limited to market orders (used by SetHoldings) because it involved calculating the order quantity with iteration. This is not necessary with limit orders: 
- if buying, the quote currency is used and existing holdings for the base currency are not relevant
- if selling, we can only sell the available quantity, taking into account _open orders_ though

jhenak:
@StefanoRaggi thank you for clarifying, the situation that I encountered this utilizes exclusively limit orders and does not use SetHoldings.  I'll search through the documentation and code some more to figure out.  Perhaps the mistake is on my end.

I encountered this issue running it in Live mode (GDAX) but not when backtesting.  I do hope it fixes it,.

PR #1587, opened 2018-02-16
Add buying power model check for shorting in SetHoldings

StefanoRaggi:
Currently, when using a cash model and calling `SetHoldings` with a negative percentage (to go short), `CalculateOrderQuantity` is returning a quantity of zero as expected, so no order is submitted and no error or warning message is shown.

With this PR, the `IsShortSellingAllowed` method has been added to the `IBuyingPowerModel` interface, `SetHoldings` calls it and shows a warning message if the short position is not allowed.

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -893,6 +893,13 @@ namespace QuantConnect.Algorithm
                 return;
             }
 
+            // check if shorting allowed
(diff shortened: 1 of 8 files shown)


StefanoRaggi:
Closing this PR for now, because `IsShortSellingAllowed` belongs to `IBrokerageModel` instead of `IBuyingPowerModel`, but it will be back soon 😃

PR #1589, opened 2018-02-19
Passes inner exception to ScheduleEvent exception

AlexCatarino:
Without the inner exception, information on the user exception is lost. In python algorithm, the stack trace is completely lost.
Removes ScheduledEventExceptionMessage property as it is redundant to Message property inherited from parent class.

### DIFF
--- a/Common/Scheduling/ScheduledEvent.cs
+++ b/Common/Scheduling/ScheduledEvent.cs
@@ -248,7 +248,7 @@ namespace QuantConnect.Scheduling
 
                 // This scheduled event failed, so don't repeat the same event
                 _needsMoveNext = true;
-                throw new ScheduledEventException(ex.ToString(), ex);
+                throw new ScheduledEventException(ex.Message, ex);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Common/Scheduling/ScheduledEvent.cs
```diff
@@ -248,7 +248,7 @@ protected void OnEventFired(DateTime triggerTime)
 
                 // This scheduled event failed, so don't repeat the same event
                 _needsMoveNext = true;
-                throw new ScheduledEventException(ex.ToString());
+                throw new ScheduledEventException(ex.ToString(), ex);
```
mchandschuh: This looks like a code smell. Perhaps it should be:
```csharp
throw new ScheduledEventException(ex.Message, ex);
```

Engine/RealTime/BacktestingRealTimeHandler.cs
```diff
@@ -149,7 +149,7 @@ public void ScanPastEvents(DateTime time)
                     }
                     catch (ScheduledEventException scheduledEventException)
                     {
-                        var errorMessage = $"BacktestingRealTimeHandler.Run(): There was an error in a scheduled event {scheduledEvent.Name}. The error was {scheduledEventException.ScheduledEventExceptionMessage}";
+                        var errorMessage = $"BacktestingRealTimeHandler.Run(): There was an error in a scheduled event {scheduledEvent.Name}. The error was {scheduledEventException.Message}";
```
mchandschuh: The `Log.Error` below should be updated to accept the exception object as well, this will give us the full stack trace/ect...

PR #1594, opened 2018-02-20
Python error handling

AlexCatarino:


Common/Util/ExceptionParser.cs
```diff
@@ -0,0 +1,181 @@
...
+
+using Python.Runtime;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Util
+{
+    public class ExceptionParser : Exception
```
mchandschuh: Let's remove this inheritance here... and `ExceptionParser` is not an `Execption`.

Perhaps `ExceptionParser` can be a stand alone class with a single, state-less method (state-less methods are very easily unit tested):
``` csharp
Exception ParseException(Exception error);
```

The `ExceptionParser` can decide whether or not to invoke python/csharp parsers, or perhaps it doesn't invoke them at all if it's not a parsable exception.

Common/Interfaces/IUserExceptionParser.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <see cref="IExceptionParser"/> interface. Parser that creates an <see cref="LegibleException"/> from an <see cref="Exception"/>.
+    /// </summary>
+    public interface IExceptionParser
+    {
+        /// <summary>
+        /// Parses an <see cref="Exception"/> object into an <see cref="LegibleException"/> one
+        /// </summary>
+        /// <param name="exception"><see cref="Exception"/> object to parse into an <see cref="LegibleException"/> one.</param>
+        /// <returns>Parsed exception</returns>
+        LegibleException Parse(Exception exception);
```
mchandschuh: return type should be `Exception` -- allows implementations to 'pass' and *not* parse an exception if it doesn't meet some expectation

Common/Exceptions/PythonUserExceptionParser.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// <returns>Parsed exception</returns>
+        public Exception Parse(Exception exception)
+        {
+            var pythonException = exception as PythonException;
+            if (pythonException == null)
+            {
+                pythonException = exception.InnerException as PythonException;
+                if (pythonException == null)
+                {
+                    throw new ArgumentException("The given exception is not valid since it is of type PythonException nor its InnerException is");
```
mchandschuh: since it is **not** of type `PythonException`

Common/Exceptions/UserException.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Creates an instance of <see cref="UserException"/> : exception created by <see cref="IExceptionParser.Parse(Exception)"/>.
+        /// </summary>
+        /// <param name="message">Message that describes the current exception.</param>
+        /// <param name="stacktrace">String representation of the immediate frames on the call stack.</param>
+        public UserException(string message, string stacktrace)
+        {
+            _message = message;
+            _stacktrace = stacktrace;
```
mchandschuh: Why manually copy the stack trace? Why not accept an inner exception and let the `Exception` object manage the stack traces?

Common/Exceptions/UserException.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// Gets a string representation of the immediate frames on the call stack.
+        /// </summary>
+        public override string StackTrace => _stacktrace;
+
+        /// <summary>
+        /// Creates and returns a string representation of the current exception.
+        /// </summary>
+        /// <returns></returns>
+        public override string ToString() => $"{_message}{Environment.NewLine}{_stacktrace}";
```
mchandschuh: Is this required? The default implementation for exceptions will properly aggregate all inner exceptions and stack traces, assuming we wrap an inner exception properly.

Common/Interfaces/IUserExceptionParser.cs
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+*/
+
+using QuantConnect.Exceptions;
+using System;
+
+namespace QuantConnect.Interfaces
+{
+    /// <summary>
+    /// <see cref="IExceptionParser"/> interface. Parser that creates an <see cref="LegibleException"/> from an <see cref="Exception"/>.
```
mchandschuh: Looks like `LegibleException` was never updated. Also, `IExceptionParser` isn't used to create a specific type of exception object, but translate it to any other exception object. It's current use is to potentially generate `UserException` instances, or if it's unable to translate, it should return the original exception unchanged.

Common/Interfaces/IUserExceptionParser.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public interface IExceptionParser
+    {
+        /// <summary>
+        /// Parses an <see cref="Exception"/> object into an <see cref="LegibleException"/> one
```
mchandschuh: Reference to `LegibleException`

Engine/Setup/BacktestingSetupHandler.cs
```diff
@@ -101,12 +102,15 @@ public int MaxOrders
             }
         }
 
+        public IExceptionParser ExceptionParser { get; }
```
mchandschuh: Instead of parsing the exceptions from the set up handler, perhaps it's best to convert the `List<string> Errors>` into a `List<Exception` -- then the engine can do it's thing on the exceptions. This would remove the exception parsing dependency from the set up handlers and keep it isolated to the engine code.

Common/Exceptions/UserException.cs
AlexCatarino: I am passing the stacktrace instead of the inner exception because I am extracting the line from the python exception that is throwing.
We could have a constructor overload for that case and other for C#.

PR #1600, opened 2018-02-21
Fix duplicate AddSecurity requests

StefanoRaggi:
Previously, when adding the same security more than once, new instances were created, causing issues with consolidators and indicators.
Now, the existing security will be returned.


StefanoRaggi:
Closing as will be addressed with issue #1612

PR #1603, opened 2018-02-22
Fixes Custom Data for Python

AlexCatarino:
In order to access the custom data classes, the module containing them was added to the ObjectActivator. This was unnecessary if it wasn't a custom data algorithm.
Also, this operation would not be taken into account if the custom data class were defined after the algorithm was created: this is the case for QuantBook.
We refactor how custom data is handled: a new class was added to provide a instance creation factory that creates an instance of each python custom type.

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -34,6 +34,8 @@ namespace QuantConnect.Algorithm
 {
     public partial class QCAlgorithm
     {
+        private readonly Dictionary<IntPtr, PythonActivator> _pythonActivators = new Dictionary<IntPtr, PythonActivator>();
+
         public PandasConverter PandasConverter { get; private set; }
 
         /// <summary>
@@ -902,22 +904,26 @@ namespace QuantConnect.Algorithm
         /// <returns>Type object</returns>
         private Type CreateType(PyObject type)
         {
-            using (Py.GIL())
+            PythonActivator pythonType;
+            if (!_pythonActivators.TryGetValue(type.Handle, out pythonType))
             {
-                var an = new AssemblyName(type.Repr().Split('.')[1].Replace("\'>", ""));
-                var assemblyBuilder = AppDomain.CurrentDomain.DefineDynamicAssembly(an, AssemblyBuilderAccess.Run);
-                var moduleBuilder = assemblyBuilder.DefineDynamicModule("MainModule");
-                return moduleBuilder.DefineType(an.Name,
-                        TypeAttributes.Public |
-                        TypeAttributes.Class |
-                        TypeAttributes.AutoClass |
-                        TypeAttributes.AnsiClass |
-                        TypeAttributes.BeforeFieldInit |
-                        TypeAttributes.AutoLayout,
-                        // If the type has IsAuthCodeSet member, it is a PythonQuandl
-                        type.HasAttr("IsAuthCodeSet") ? typeof(PythonQuandl) : typeof(PythonData))
-                    .CreateType();
+                AssemblyName an;
+                using (Py.GIL())
+                {
+                    an = new AssemblyName(type.Repr().Split('\'')[1]);
+                }
+                var moduleBuilder = AppDomain.CurrentDomain
+                    .DefineDynamicAssembly(an, AssemblyBuilderAccess.Run)
+                    .DefineDynamicModule("MainModule");
+
+                pythonType = new PythonActivator(moduleBuilder.DefineType(an.Name).CreateType(), type);
+
+                ObjectActivator.AddActivator(pythonType.Type, pythonType.Factory);
+
+                // Save to prevent future additions
+                _pythonActivators.Add(type.Handle, pythonType);
             }
+            return pythonType.Type;
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Algorithm/QCAlgorithm.Python.cs
```diff
@@ -869,22 +871,20 @@ private dynamic[] GetIndicatorArray(PyObject first, PyObject second = null, PyOb
...
-                        TypeAttributes.Class |
-                        TypeAttributes.AutoClass |
-                        TypeAttributes.AnsiClass |
-                        TypeAttributes.BeforeFieldInit |
-                        TypeAttributes.AutoLayout,
-                        // If the type has IsAuthCodeSet member, it is a PythonQuandl
-                        type.HasAttr("IsAuthCodeSet") ? typeof(PythonQuandl) : typeof(PythonData))
-                    .CreateType();
+                AssemblyName an;
+                using (Py.GIL()) an = new AssemblyName(type.Repr().Split('\'')[1]);
```
mchandschuh: As a point of code style, please *always* use braces. While it may add 2 lines to the code, it makes it immeasurably easier to read.

AlgorithmFactory/Loader.cs
```diff
@@ -168,25 +168,25 @@ private bool TryCreatePythonAlgorithm(string assemblyPath, out IAlgorithm algori
                 // Import Python module
                 using (Py.GIL())
                 {
-                    Log.Trace("Loader.TryCreatePythonAlgorithm(): Importing python module " + moduleName);
+                    Log.Trace($"Loader.TryCreatePythonAlgorithm(): Python version {PythonEngine.Version}");
+                    Log.Trace($"Loader.TryCreatePythonAlgorithm(): Importing python module {moduleName}");
```
mchandschuh: nit - could combine these two trace statements into a single line:
``` csharp
Log.Trace($"Loader.TryCreatePythonAlgorithm(): Python version {PythonEngine.Version}: Importing python module {moduleName}");
```

Common/Python/PythonActivator.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <param name="value"><see cref="PyObject"/> that contains the python type</param>
+        public PythonActivator(Type type, PyObject value)
+        {
+            Type = type;
+            Factory = x =>
+            {
+                using (Py.GIL())
+                {
+                    var instance = value.Invoke();
+                    if (instance.HasAttr("ValueColumnName"))
```
mchandschuh: is there a way to check if the type derives from quandl data some how? Won't this catch *every* type with a `ValueColumnName` property?

Common/Util/ObjectActivator.cs
```diff
@@ -138,28 +124,16 @@ public static object Clone(object instanceToClone)
...
+        /// Adds method to return an instance of object
         /// </summary>
-        /// <param name="module">The algorithm python module</param>
-        public static void SetPythonModule(PyObject module)
+        /// <param name="key">The key of the method to add</param>
+        /// <param name="value">The value of the method to add</param>
+        public static void AddActivator(Type key, Func<object[], object> value)
         {
-            _activatorsByType[typeof(PythonData)] = type =>
+            if (!_activatorsByType.ContainsKey(key))
```
mchandschuh: I like this pattern much more tan the set pattern as it's more explicit. Great change.

mchandschuh: Also, what should the behavior be if `AddActivator` is called twice for the same `Type`? Should we throw? I don't think silently doing nothing is the correct approach.

AlexCatarino: In python algorithms, `AddActivator` cannot be called twice for the same type since we check for it in QCAlgorithm.CreateType.

mchandschuh:
@AlexCatarino - looks good, can we add a test that fails on `master` and passes in this branch to highlight the change

Common/Util/ObjectActivator.cs
mchandschuh: This is a public static method, anyone can call it again, but shouldn't. Can we add an `else` block that throws as it's a common pattern when adding something by key and the key already exists to throw an exception, in addition, since we don't expect this case to ever happen it would be nice if it did accidentally happen in the future that we're warned about it immediately via the exception.

Tests/Algorithm/AlgorithmAddDataTests.cs
```diff
@@ -100,6 +103,35 @@ public void CustomDataTypes_AreAddedToSubscriptions_Successfully()
...
+                var module = Py.Import("Test_CustomDataAlgorithm");
+                var qcAlgorithm = new AlgorithmPythonWrapper(module);
+
+                // Initialize contains the statements:
+                // self.AddData(Nifty, "NIFTY")
+                // self.AddData(QuandlFuture, "SCF/CME_CL1_ON", Resolution.Daily)
+                qcAlgorithm.Initialize();
+
+                var niftySubscription = qcAlgorithm.SubscriptionManager.Subscriptions.FirstOrDefault(x => x.Type.Name == "Nifty");
+                Assert.AreEqual(niftySubscription.Type.Name, "Nifty");
```
mchandschuh: I think more correctly this should be `Assert.IsNotNull(niftySubscription)`, since `FirstOrDefault` will return null if not found. Also, I try to *never* use the string type names for comparisons unless absolutely necessary. If possible, update these to check the actual type using `x.Type == typeof(Nifty)` syntax.

AlexCatarino: It is not possible to use `x.Type == typeof(Nifty)`, because `Nifty` is not a C#, it doesn't exist in QuantConnect Assembly. So maybe we could just we could just use `Assert.IsNotNull(niftySubscription)` and use `x.Symbol.Value == "NIFTY"` instead.

PR #1609, opened 2018-02-23
Upgrade LEAN Python to 3.6 from 2.7

mchandschuh:
Slightly reworked version of https://github.com/QuantConnect/Lean/pull/1526

The biggest change from #1526 is that we're now installing miniconda into `/opt/` so it's available to all users.

---
#### Description from #1526 
In this pull request we propose changes that enables python 3.6 in Lean.
- Adds log to display the python version the algorithm is using.
- Fixes python algorithms that were failing because of small subtleties like leading zeroes.
- Updates pythonnet with a version compiled with python 3.6 flags

Also included via #1526:
- Changes in DockerfileFoundation: we now use miniconda to manage the python environment. Took the opportunity to add NTLK (#1349), Tensorforce (#1369) and PyTorch/Pyro (#1385).
- Changes readme in Algorithm.Python to show steps to install miniconda

### DIFF
--- a/Algorithm.Python/readme.md
+++ b/Algorithm.Python/readme.md
@@ -6,21 +6,21 @@ QuantConnect Python Algorithm Project:
 
 1.1 Visual Studio 2017 Community edition can be downloaded free from [https://www.visualstudio.com](https://www.visualstudio.com). VS comes with Python (if you install Python Tools for Visual Studio) but it doesn't seem to add Python to the system path. We'll need to reinstall python manually. If you already have VS installed you should be able to skip this step.
 
-**2. Install Python:**
+**2. Install Python 3.6:**
 
-2.1 Use the Windows x86-64 MSI installer from [https://www.python.org/downloads/release/python-2713](https://www.python.org/downloads/release/python-2713).
+2.1 Use the Windows x86-64 MSI installer from [https://www.python.org/downloads/release/python-364/](https://www.python.org/downloads/release/python-364/).
 
 2.2 When asked to select the features to be installed, make sure you select "Add python.exe to Path"
 
-2.3 Once installed confirm the python27.dll is in place in *C:\Windows\System32*.
+2.3 Once installed confirm the python36.dll is in place in *C:\Windows\System32*.
 
-2.4 Add python27.dll location to the system path:
+2.4 Add python36.dll location to the system path:
  
 2.4.1 Right mouse button on My Computer. Click Properties.
 
 2.4.2 Click Advanced System Settings -> Environment Variables -> System Variables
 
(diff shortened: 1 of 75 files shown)


DockerfileLeanFoundation
```diff
@@ -14,35 +14,39 @@ CMD ["/sbin/my_init"]
...
-  pip install blaze cvxopt cvxpy pykalman statistics statsmodels arch copulalib && \
-  pip install xgboost theano keras tensorflow deap
+RUN apt-get update && apt-get install -y wget bzip2 python3-pip python-opengl && \
+    apt-get install -y xvfb unzip curl libxrender1 libxtst6 libxi6
+
+# Install miniconda and supported third party python packages
+ENV PATH="/opt/miniconda3/bin:${PATH}"
+RUN wget https://repo.continuum.io/miniconda/Miniconda3-4.3.31-Linux-x86_64.sh && \
+    bash Miniconda3-4.3.31-Linux-x86_64.sh -b -p /opt/miniconda3 && \
+    rm -rf Miniconda3-latest-Linux-x86_64.sh && \
```
AlexCatarino: I think we should use:
```
RUN wget https://repo.continuum.io/miniconda/Miniconda3-4.3.31-Linux-x86_64.sh -O miniconda.sh && \
    bash miniconda.sh -b -p /opt/miniconda3 && \
    rm -rf miniconda.sh && \
```
To keep the version in only one line. In this case we were deleting a non-existing file.

Algorithm.Python/readme.md
```diff
@@ -113,15 +113,28 @@ $ sudo easy-install pip
 ### Linux
 **1. Follow the installation instructions [here](https://github.com/QuantConnect/Lean#linux-debian-ubuntu)**
 
-1.1 You should see this exit successfully with a list of the algorithm statistics and 1 trade.
+**2. Install Python 3.6 with Miniconda:**
+```
+export PATH="/root/miniconda3/bin:$PATH"
+wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
+bash Miniconda3-latest-Linux-x86_64.sh -b
+rm -rf Miniconda3-latest-Linux-x86_64.sh
```
AlexCatarino: Here we should have the same commands of the dockerfile.

DockerfileLeanFoundation
mchandschuh: I actually did something similar you shared with me, but when I ran it, docker couldn't find the renamed script, so I dd it this way instead.

disruptek:
Some more missing items to add to the docs:

```python
# implement cmp()
cmp = lambda x, y: (x > y)-(x < y)

# implement basestring
try:
    basestring
except NameError:
    basestring = str
```

I had a few others but every time you guys spring this upgrade upon me, I find and fix a few more.

PR #1613, opened 2018-02-26
FXCM sample data as quote bars

Jay-Jay-D:
Sample data for FXCM at hour, minute and second resolution was actually tradebars, even when the files had the `_quote` subfix.  

This PR fixes that issue uploading quote bars for the sample FXCM data.

Related to  #1536 
Sample EURUSD minute quotebar data.
Sample NZDUSD minute quotebar data.
Sample EURUSD second quotebar data.
Sample NZDUSD second quotebar data.
Sample hourly quote bar data.


mchandschuh:
Confirmed with @Jay-Jay-D that these changes do not break existing unit/regression tests.

PR #1621, opened 2018-02-27
Refactors UserException classes

AlexCatarino:
Removes UserException class since we can use the regular Exception class instead.
Adds SyntaxError handling


Common/Exceptions/ExceptionParser.cs
```diff
@@ -43,13 +43,20 @@ public ExceptionParser()
...
+                else
+                {
+                    return _cSharpUserExceptionParser.Parse(exception);
+                }
             }
-            else
+            catch
             {
-                return _cSharpUserExceptionParser.Parse(exception);
+                return exception;
```
mchandschuh: Let's at least log the exception that was thrown here. This will help us find/debug errors in the exception parsers

Common/Exceptions/InitializeException.cs
```diff
@@ -0,0 +1,44 @@
...
+*/
+
+using System;
+
+namespace QuantConnect.Exceptions
+{
+    /// <summary>
+    /// Throw this if there is an exception in the IAlgorithm.Initialize method 
+    /// </summary>
+    class InitializeException : Exception
```
mchandschuh: Any particular reason this isn't `public` ?

Common/Exceptions/PythonUserExceptionParser.cs
```diff
@@ -13,58 +13,91 @@
...
+            var message = CreateLegibleMessage(exception.Message);
+            var errorLine = CreateErrorLine(exception.StackTrace);
+            message = $"{message}{errorLine}";
+
+            if (original.GetType() == typeof(InitializeException))
+            {
+                message = $"In the Initialize method, {message.Substring(0, 1).ToLower()}{message.Substring(1)}";
+                return new InitializeException(message, original);
+            }
+            else if (original.GetType() == typeof(ScheduledEventException))
```
mchandschuh: shame this is duplicated from the csharp type, let's not touch this now, but a note for our future selves that this can be improved.

mchandschuh:
Closing this for now, but will return in smaller pieces.

Issue #1620, opened 2018-02-26
Python Plot doesn't plot StandardDeviation indicator


Jay-Jay-D:
#### Expected Behavior

In Python, the `self.Plot` function can be used as 
```python
self.Plot('RSI', self.rsi)
```
where `self.rsi` is an indicator defined in the `Initialize` method.

#### Actual Behavior

However, if the `self.Plot` this method is called with a `StandardDeviation` indicator, it throws the following error:
`Runtime Error: QuantConnect.Scheduling.ScheduledEventException: TypeError : No method matches given arguments for Plot ---> Python.Runtime.PythonException: TypeError : No method matches given arguments for Plot at Python.Runtime.Dispatcher.Dispatch (System.Collections.ArrayList args) ` 

#### Potential Solution

As the `StandardDeviation` indicator inherits from `Variance`, the Python port cannot find the correct overload

#### Reproducing the Problem

To reproduce the error, uncomment the last line `self.Plot('indicatorSTD', self.std)` in [this algorithm](https://www.quantconnect.com/terminal/processCache/?request=embedded_backtest_ce154730f41fbcc189b376668f3bc58e.html)

#### System Information

Cloud service.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

jaredbroad:
Fixed in #1622

PR #1622, opened 2018-02-27
Adds overload to Plot method to handle indicator special cases

AlexCatarino:
#### Description
For cases where implicit conversion to decimal from indicator was not properly handled by pythonnet, we expliticly get `Current.Value` and redirect to `Plot(string,decimal)` overload.

#### Related Issue
#1620

#### Motivation and Context
Since we have included implicit conversion in our pythonnet fork, it has been able to convert Lean indicator into decimal and, consequently, it founds the right overload for the Plot method. 
Apparently it has a bug in that implementation since it couldn't resolve `StandardDeviation` that derives from `Variance` and not directly from a `IndicatorBase<T>` class.

Code provided in issue #1620 

#### Types of changes

#### Checklist:
- [ ] My change requires a change to the documentation.

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -378,11 +378,29 @@ namespace QuantConnect.Algorithm
         /// <seealso cref="Plot(string,decimal)"/>
         public void Plot(string series, PyObject pyObject)
         {
+            IIndicator<IndicatorDataPoint> indicator;
+
             using (Py.GIL())
             {
-                var value = ((dynamic)pyObject).Current.Value;
-                Plot(series, (decimal)value);
+                var pythonType = pyObject.GetPythonType();
+
+                try
+                {
+                    var type = pythonType.As<Type>();
+                    indicator = pyObject.AsManagedObject(type) as IIndicator<IndicatorDataPoint>;
+
+                    if (indicator == null)
+                    {
+                        throw new ArgumentException();
+                    }
+                }
+                catch
+                {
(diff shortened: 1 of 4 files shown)


jaredbroad:
Hi @AlexCatarino missing a test to cover this please :)
_"I have added tests to cover my changes."_

Create a regression or unit test which is _failing_ on master -- and fixed by this PR.

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -370,6 +370,21 @@ public void RegisterIndicator(Symbol symbol, PyObject indicator, IDataConsolidat
...
+        /// Plot a chart using string series name, with value.
+        /// </summary>
+        /// <param name="series">Name of the plot series</param>
+        /// <param name="pyObject">PyObject with the value to plot</param>
+        /// <seealso cref="Plot(string,decimal)"/>
+        public void Plot(string series, PyObject pyObject)
+        {
+            using (Py.GIL())
+            {
+                var value = ((dynamic)pyObject).Current.Value;
```
mchandschuh: Is it possible to to check `pyObject` for implemented interfaces to confirm it implements `IIndicator<T>`?

Concern is non-indicator `PyObject`s getting in here by accident, would like to minimally log for us and message the user if `pyObject` isn't an indicator.

What about a decimal value in python, will that bind to this method as a `PyObject`?
``` python
self.sma = SMA("SPY", 10)
self.value = sma.Current.Value
Plot("SPY-Price", self.value);
```

Issue #1628, opened 2018-02-28
Refactor ISetupHandler.Errors to List<Exception>


mchandschuh:
#### Description

In order to properly report error messages to the engine we should communicate the entire exception object instead of just a string. This will ensure stack traces and other potentially useful information embedded in the original exception isn't lost

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

mchandschuh:
Refactored in #1629

PR #1629, opened 2018-02-28
Refactor ISetupHandler.Errors to List<Exception>

mchandschuh:
#### Description

Refactors `ISetupHandler.Errors` to `List<Exception>`

#### Related Issue

#1628

#### Motivation and Context

Converting these errors into proper `Exception` objects will make them less special and able to be handled like other exceptions. In the near future, we'll be projecting system exceptions into usable messages for users and this removes the potential for losing information when using only a string.

No.

Sadly I don't think this change is covered by any existing regression tests, although I did run through it locally. It was determined to be outside of the scope of this mechanical refactoring to also improve the testability of the various setup handler implementations. I view the risk for this change as very low. We can create a separate PR to get the setup handlers under test.

#### Types of changes

- [x] Non-functional change (mechanical refactoring)

#### Checklist:

### DIFF
--- a/Engine/Engine.cs
+++ b/Engine/Engine.cs
@@ -164,7 +164,7 @@ namespace QuantConnect.Lean.Engine
                         initializeComplete = false;
                         //Get all the error messages: internal in algorithm and external in setup handler.
                         var errorMessage = String.Join(",", algorithm.ErrorMessages);
-                        errorMessage += String.Join(",", _algorithmHandlers.Setup.Errors);
+                        errorMessage += String.Join(",", _algorithmHandlers.Setup.Errors.Select(e => e.Message));
                         Log.Error("Engine.Run(): " + errorMessage);
                         _algorithmHandlers.Results.RuntimeError(errorMessage);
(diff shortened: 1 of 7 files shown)


Engine/Engine.cs
```diff
@@ -164,7 +164,7 @@ public void Run(AlgorithmNodePacket job, AlgorithmManager manager, string assemb
                         initializeComplete = false;
                         //Get all the error messages: internal in algorithm and external in setup handler.
                         var errorMessage = String.Join(",", algorithm.ErrorMessages);
-                        errorMessage += String.Join(",", _algorithmHandlers.Setup.Errors);
+                        errorMessage += String.Join(",", _algorithmHandlers.Setup.Errors.Select(e => e.Message));
```
AlexCatarino: If it only pass the `Message`, the stack trace will not be shown.

mchandschuh: Yup, this was just a mechanical refactoring and the original implementation didn't show the stack trace (since it wasn't communicated). If we want to modify it to change the stack we can do so in a separate PR.

Issue #1636, opened 2018-03-01
RegisterIndicator does not process composite indicator in python algorithms


AlexCatarino:
#### Expected Behavior
All types of indicators can be registered to automatic updates

#### Actual Behavior
Composite indicators are not being registered in python algorithms

#### Potential Solution
Add CompositeIndicator case for `RegisterIndicator(Symbol, PyObject, IDataConsolidator, PyObject)`

#### Reproducing the Problem
```python
class BasicTemplateAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.spy = self.AddEquity("SPY").Symbol
        sma = SimpleMovingAverage(30)
        delay = Delay(1)
        delayedSma = IndicatorExtensions.Of(delay, sma)
        self.RegisterIndicator(self.spy, delayedSma)
        self.PlotIndicator("DelayedSma", delayedSma)
```

#### System Information
N/A

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

AlexCatarino:
Closed by #1641

PR #1641, opened 2018-03-02
 Fixes RegisterIndicator for python algorithms

AlexCatarino:
#### Description
Refactors the `RegisterIndicator(Symbol, PyObject, IDataConsolidator, PyObject)` to check whether a C# object inherits from `IndicatorBase<T>` instead of some of its children classes (`Indicator`, `BarIndicator` and `TradeBarIndicator`).

#### Related Issue

#1636 

#### Motivation and Context

It solves the impossibility to register indicators that don't inherit from `Indicator`, `BarIndicator` and `TradeBarIndicator`, but directly from `IndicatorBase<T>`, and also from `WindowIndicator<T>`.

No.

Added unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -317,45 +317,37 @@ namespace QuantConnect.Algorithm
         /// <param name="selector">Selects a value from the BaseData send into the indicator, if null defaults to a cast (x => (T)x)</param>
         public void RegisterIndicator(Symbol symbol, PyObject indicator, IDataConsolidator consolidator, PyObject selector = null)
         {
+            IndicatorBase<IndicatorDataPoint> indicatorDataPoint;
+            IndicatorBase<IBaseDataBar> indicatorDataBar;
+            IndicatorBase<TradeBar> indicatorTradeBar;
+
+            if (indicator.TryConvert(out indicatorDataPoint))
+            {
+                Func<IBaseData, decimal> func = null;
+                selector?.TryConvert(out func);
+                RegisterIndicator(symbol, indicatorDataPoint, consolidator, func);
+                return;
+            }
+            else if (indicator.TryConvert(out indicatorDataBar))
+            {
+                Func<IBaseData, IBaseDataBar> func = null;
+                selector?.TryConvert(out func);
+                RegisterIndicator(symbol, indicatorDataBar, consolidator, func);
+                return;
+            }
+            else if (indicator.TryConvert(out indicatorTradeBar))
+            {
+                Func<IBaseData, TradeBar> func = null;
+                selector?.TryConvert(out func);
+                RegisterIndicator(symbol, indicatorTradeBar, consolidator, func);
+                return;
+            }
+
             using (Py.GIL())
             {
-                try
+                if (!indicator.HasAttr("Update"))
                 {
-                    // QuantConnect Indicators
-                    var indicatorPythonType = indicator.GetPythonType();
-                    var indicatorType = GetIndicatorBaseType(indicatorPythonType.As<Type>());
-                    var genericArgumentType = indicatorType.GetGenericArguments().FirstOrDefault();
-
-                    if (genericArgumentType == typeof(IndicatorDataPoint))
-                    {
-                        var managedIndicator = indicator.AsManagedObject(indicatorType) as IndicatorBase<IndicatorDataPoint>;
-                        var managedSelector = PythonUtil.ToFunc<IBaseData, decimal>(selector);
-                        RegisterIndicator(symbol, managedIndicator, consolidator, managedSelector);
-                    }
-                    else if (genericArgumentType == typeof(IBaseDataBar))
-                    {
-                        var managedIndicator = indicator.AsManagedObject(indicatorType) as IndicatorBase<IBaseDataBar>;
-                        var managedSelector = PythonUtil.ToFunc<IBaseData, IBaseDataBar>(selector);
-                        RegisterIndicator(symbol, managedIndicator, consolidator, managedSelector);
-                    }
-                    else if (genericArgumentType == typeof(TradeBar))
-                    {
-                        var managedIndicator = indicator.AsManagedObject(indicatorType) as IndicatorBase<TradeBar>;
-                        var managedSelector = PythonUtil.ToFunc<IBaseData, TradeBar>(selector);
-                        RegisterIndicator(symbol, managedIndicator, consolidator, managedSelector);
-                    }
-                    else
-                    {
-                        throw new ArgumentException($"QCAlgorithm.RegisterIndicator(): Unsupported indicator data type: {genericArgumentType.Name}");
-                    }
-                    return;
-                }
-                catch (InvalidCastException)
-                {
-                    if (!indicator.HasAttr("Update"))
-                    {
-                        throw new ArgumentException($"QCAlgorithm.RegisterIndicator(): Update method must be defined. Please checkout {indicator}");
-                    }
+                    throw new ArgumentException($"QCAlgorithm.RegisterIndicator(): Update method must be defined. Please checkout {indicator}");
                 }
             }
 
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -317,46 +317,46 @@ namespace QuantConnect.Algorithm
         /// <param name="selector">Selects a value from the BaseData send into the indicator, if null defaults to a cast (x => (T)x)</param>
         public void RegisterIndicator(Symbol symbol, PyObject indicator, IDataConsolidator consolidator, PyObject selector = null)
         {
-            object managedObject = null;
-
             using (Py.GIL())
             {
-                var pythonType = indicator.GetPythonType();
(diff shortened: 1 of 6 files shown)


mchandschuh:
Code looks great, but it does highlight a deficiency in our design here. Clearly the generic type parameter is causing some heartburn. I wonder if much of this could be simplified by creating a non-generic interface to be the parent of all indicators, although, we might just end up moving this type resolution code elsewhere.

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -317,46 +317,46 @@ public void RegisterIndicator(Symbol symbol, PyObject indicator, TimeSpan? resol
...
+                        var managedIndicator = indicator.AsManagedObject(indicatorType) as IndicatorBase<IndicatorDataPoint>;
+                        var managedSelector = PythonUtil.ToFunc<IBaseData, decimal>(selector);
+                        RegisterIndicator(symbol, managedIndicator, consolidator, managedSelector);
+                    }
+                    else if (genericArgumentType == typeof(IBaseDataBar))
+                    {
+                        var managedIndicator = indicator.AsManagedObject(indicatorType) as IndicatorBase<IBaseDataBar>;
+                        var managedSelector = PythonUtil.ToFunc<IBaseData, IBaseDataBar>(selector);
+                        RegisterIndicator(symbol, managedIndicator, consolidator, managedSelector);
+                    }
```
mchandschuh: Can we extract this code into a class and put that class under unit test? Any time we get into the weeds with reflection code it's very important to isolate the code. I'm not recommending we remove the integration test you've defined as it confirms it works properly from the algorithm's point of view, but I suspect we'll have issues w/ this code. For example, we talk about `TradeBar` indicators but don't support `QuoteBar`

Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs
```diff
@@ -0,0 +1,153 @@
...
+                        Assert.DoesNotThrow(() => _algorithm.RegisterIndicator(_spy, indicator, Resolution.Minute));
+                    }
+                    else if (key == typeof(TradeBar))
+                    {
+                        var indicator = (indicatorTest as CommonIndicatorTests<TradeBar>).Indicator;
+                        Assert.DoesNotThrow(() => _algorithm.RegisterIndicator(_spy, indicator, Resolution.Minute));
+                    }
+                    expected++;
+                    var actual = _algorithm.SubscriptionManager.Subscriptions.FirstOrDefault().Consolidators.Count;
+                    Assert.AreEqual(expected, actual);
```
mchandschuh: Perhaps the `expected++` belongs inside each of the type checks. We want to confirm each case has been hit. Similar comment for the below test as well.

Algorithm/QCAlgorithm.Python.cs
AlexCatarino: Agree. I am going to create the helper class with unit tests and make a separate PR. After it is reviewed and merged, I will rebase this one.

Common/Extensions.cs
```diff
@@ -1006,5 +1007,38 @@ public static string ToSafeString(this PyObject pyObject)
...
+                    result = pyObject.AsManagedObject(typeof(T)) as T;
+                }
+                catch
+                {
+                    // Do not throw or log the exception.
+                    // Return false as an exception means that the conversion could not be made.
+                }
+            }
+
+            return result != default(T);
```
mchandschuh: In cases such as this, I think it's best to have an explicit `return true` and explicit `return false` statement.

Inside the `try` block we can add `return true` following the `result = pyObject.AsManagedObject(typeof(T)) as T;` line and then this line can become `return false`.

While it may not apply in this case (or often at all) -- I prefer to avoid using `==` or `!=` on unknown types since these operators can be overloaded ([see](https://github.com/QuantConnect/Lean/blob/master/Common/Symbol.cs#L387) the `Symbol` class which equates `Symbol.Empty` and `null`!).

Also, I think it makes the code clearer by including the `return true` and `return false` statements in particular points in the control flow, since that's what really determines success -- that we were able to execute `pyObject.AsManagedObject(typeof(T))` without throwing an exception and the `result != default(T)` is a proxy check for the success of the previous method call.

Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+            _spy = _algorithm.AddEquity("SPY").Symbol;
+
+            _indicatorTestsGroups =
+                from type in GetType().Assembly.GetTypes()
+                where type.IsPublic && !type.IsAbstract
+                where
+                   typeof(CommonIndicatorTests<TradeBar>).IsAssignableFrom(type) ||
+                   typeof(CommonIndicatorTests<IBaseDataBar>).IsAssignableFrom(type) ||
+                   typeof(CommonIndicatorTests<IndicatorDataPoint>).IsAssignableFrom(type)
+                group type by type.BaseType.GetGenericArguments().FirstOrDefault();
```
mchandschuh: I love seeing you get more and more comfortable with the reflection libraries. This is great.

Hmm... which indicators should I test... I know! I'll test **all** of them :D

Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+                var key = group.Key;
+                foreach (var type in group)
+                {
+                    var indicatorTest = Activator.CreateInstance(type);
+                    if (key == typeof(IndicatorDataPoint))
+                    {
+                        using (Py.GIL())
+                        {
+                            var indicator = (indicatorTest as CommonIndicatorTests<IndicatorDataPoint>).Indicator.ToPython();
+                            Assert.DoesNotThrow(() => _algorithm.RegisterIndicator(_spy, indicator, Resolution.Minute));
```
mchandschuh: I love what's going on here but would like us to clean it up a little bit so we can use it in the future. Can we add a method to `CommonIndicatorTests<T>` to get the `PyObject`, maybe something like:
```csharp
public PyObject GetIndicatorAsPyObject()
```
which would handle getting the indicator instance and invoking `ToPython()` to convert it to `PyObject`.

With the above in place, I think this entire if/else block collapses and will make the test much clearer.

thoughts?

Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs
```diff
@@ -0,0 +1,157 @@
...
+                }
+            }
+        }
+
+        [Test, Ignore]
+        public void RegisterPythonCustomIndicatorProperly()
+        {
+            using (Py.GIL())
+            {
+                var module = PythonEngine.ModuleFromString("x",
```
mchandschuh: Does `PythonEngine` keep track of all modules names? If we reload another module named `x`, what will happen? Just thinking it may be wise to use more descriptive names here.

Tests/Common/Util/ExtensionsTests.cs
```diff
@@ -234,8 +237,8 @@ public void ConvertsDictionaryFromString()
         public void DictionaryAddsItemToExistsList()
         {
             const int key = 0;
-            var list = new List<int> {1, 2};
-            var dictionary = new Dictionary<int, List<int>> {{key, list}};
+            var list = new List<int> { 1, 2 };
+            var dictionary = new Dictionary<int, List<int>> { { key, list } };
```
mchandschuh: #### No Action Required.
When possible, try to avoid these types of white space changes. I know my settings are configured to format like the left w/out spaces following `{` -- so changes like these often just cause code churn and diffs noisy. I guess my point is, since we don't have a full-featured style guide, we shouldn't be updating existing code styles that are in line with the existing conventions.

I have a lower priority to-do items on my list for improving the R# `dotSettings` files w.r.t code style, so hopefully in the future it's as easy as telling R# to format it according to the `dotSettings` file and then we're all consistent with each other. easy peasy.

Tests/Common/Util/ExtensionsTests.cs
```diff
@@ -297,6 +300,73 @@ public void RoundsDownInTimeZone()
             Assert.AreEqual(time, roundedTime);
         }
 
+        [Test, Ignore]
+        public void TryConvertPyObject()
```
mchandschuh: This test appears to be many many tests in one. A common saying is `one assert per unit test` -- but I like to generalize this to `one *logical* assert per unit test` -- both of which are somewhat fancy ways of saying a test should only test one behavior. Maybe these cases can be extracted into `[TestCase]`s ?

Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs
AlexCatarino: Agree and made some changes. 
I couldn't remove all if/else block, because we of generics, so it is following `CommonIndicatorTests` pattern.

Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs
AlexCatarino: Totally agree. From now on, I think we should will use `Guid.NewGuid().ToString()`.

Issue #1644, opened 2018-03-02
Multiple OnEndOfDay error message


AlexCatarino:
#### Expected Behavior

When an exception occurs in a `OnEndOfDay`, algorithm should throw, log and exit.

#### Actual Behavior

When an exception occurs due to `OnEndOfDay` muitiple overloads, algorithm logs the exception and continue until the end of execution.

#### Potential Solution

Modify exception handling in `AlgorithmPythonWrapper.OnEndOfDay`, 

#### Reproducing the Problem

```python
class BasicTemplateAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2013,10, 7)
        self.SetEndDate(2013,10,11)
        self.AddEquity("SPY", Resolution.Minute)
        
    def OnData(self, data):
        pass

    def OnEndOfDay(self):
        pass
```

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #1645, opened 2018-03-02
Fixes multiple OnEndOfDay error message in python algorithms

AlexCatarino:
#### Description
The error message it was used to bypass the the error logging has been changed from `takes exactly x arguments (y given)` to `takes x positional argument but y were given`. In the previous implementation, an error in `OnEndOfDay` would be logged, but wouldn't not stop the execution.  

#### Related Issue

#1644 

#### Motivation and Context

Since python does not support method overload and `IAlgorithm` implements two overloads of `OnEndOfDay`, when this method is implemented in the python script, both overloads are called and one of them will throw an exception due the wrong number of parameters. 
This change is required because of unnecessary exception throwing. Also runtime errors that are originated by the user show stop the execution unless it has been handled by the user.

No.

Code provided in #1644 

#### Types of changes

#### Checklist:


mchandschuh:
@AlexCatarino -- is there a test we can add for this?

PR #1649, opened 2018-03-02
Fix misleading error message in CashBuyingPowerModel

### LINKED ISSUE #1648: Misleading error message in CashBuyingPowerModel
When adding a non-USD `Crypto` pair (e.g. `BTCEUR`), an internal feed is automatically added to convert the `EUR` currency to `USD` (account currency). If `SetHoldings("BTCEUR", x)` is called on a Saturday (when FX markets are closed), a misleading error message is shown.

#### Expected Behavior
Expected error message: 
_The order quantity for BTCEUR cannot be calculated: Reason: The internal cash feed required for converting EUR to USD does not have any data yet (or market may be closed)._

#### Actual Behavior
Current error message:
_The order quantity for BTCEUR cannot be calculated: Reason: The price of the BTCEUR security is zero because it does not have any market data yet. When the security price is set this security will be ready for trading._

#### Potential Solution
Update the error message.

#### Reproducing the Problem
Run a backtest for this simple algorithm:
```
    {
        private readonly Symbol _sym = QuantConnect.Symbol.Create("BTCEUR", SecurityType.Crypto, Market.GDAX);

        {
            SetStartDate(2017, 10, 7);  // Saturday
            SetEndDate(2017, 10, 8);
            SetCash(100000);

            AddCrypto("BTCEUR", Resolution.Hour);
        }
        {
            if (!Portfolio.Invested)
            {
                SetHoldings(_sym, 1);
                Debug("Purchased Stock");
            }
        }
    }
```

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
When adding a non-USD `Crypto` pair (e.g. BTCEUR), an internal feed is automatically added to convert the `EUR` currency to `USD` (account currency). If `SetHoldings("BTCEUR", x)` is called on a Saturday (when FX markets are closed), a misleading error message is shown.

#### Related Issue
Fixes #1648

#### Motivation and Context
The error message was misleading because the security price was not zero.

No

Test algorithm in #1648 

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/CashBuyingPowerModel.cs
+++ b/Common/Securities/CashBuyingPowerModel.cs
@@ -178,7 +178,18 @@ namespace QuantConnect.Securities
             var unitPrice = new MarketOrder(security.Symbol, 1, DateTime.UtcNow).GetValue(security);
             if (unitPrice == 0)
             {
-                return new GetMaximumOrderQuantityForTargetValueResult(0, $"The internal cash feed required for converting {security.QuoteCurrency.Symbol} to {CashBook.AccountCurrency} does not have any data yet (or market may be closed).");
+                if (security.QuoteCurrency.ConversionRate == 0)
+                {
+                    return new GetMaximumOrderQuantityForTargetValueResult(0, $"The internal cash feed required for converting {security.QuoteCurrency.Symbol} to {CashBook.AccountCurrency} does not have any data yet (or market may be closed).");
+                }
+
+                if (security.SymbolProperties.ContractMultiplier == 0)
+                {
+                    return new GetMaximumOrderQuantityForTargetValueResult(0, $"The contract multiplier for the {security.Symbol.Value} security is zero. The symbol properties database may be out of date.");
+                }
+
+                // security.Price == 0
+                return new GetMaximumOrderQuantityForTargetValueResult(0, $"The price of the {security.Symbol.Value} security is zero because it does not have any market data yet. When the security price is set this security will be ready for trading.");
             }
 
             // remove directionality, we'll work in the land of absolutes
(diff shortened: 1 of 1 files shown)


Common/Securities/CashBuyingPowerModel.cs
```diff
@@ -178,7 +178,7 @@ public GetMaximumOrderQuantityForTargetValueResult GetMaximumOrderQuantityForTar
             var unitPrice = new MarketOrder(security.Symbol, 1, DateTime.UtcNow).GetValue(security);
             if (unitPrice == 0)
             {
-                return new GetMaximumOrderQuantityForTargetValueResult(0, $"The price of the {security.Symbol.Value} security is zero because it does not have any market data yet. When the security price is set this security will be ready for trading.");
+                return new GetMaximumOrderQuantityForTargetValueResult(0, $"The internal cash feed required for converting {security.QuoteCurrency.Symbol} to {CashBook.AccountCurrency} does not have any data yet (or market may be closed).");
```
mchandschuh: Maybe it would be better to explicitly check that the quote currency has a non-zero conversion rate for this message. Looking at the code now, I could see 2 reasons this error would be raised, if the security price is zero or if the conversion rate security price is zero, or did I miss something?

I think this questions are some incentive to put these error messages under unit test.

StefanoRaggi: This method assumes the security price is not zero because this check is already done earlier (in `CalculateOrderQuantity`), this method will never be called if `security.Price == 0`.

mchandschuh: Very well, although I do want to caution us from thinking something *will never be called* due to the way our code is *currently* set up. It's important to view software components in isolation as it reduces logical coupling between components. For example, the engine doesn't currently call this method if `securityPrice == 0` but that doesn't mean a user can't call this method directly and/or a future refactoring could change the current call patterns.

I'm a fan of not *implementing* the same checks twice, but it is often necessary to *perform* the same checks twice. Ideally they're implemented in a single place and reused everywhere that requires those checks.

@StefanoRaggi -- I defer to your judgement regarding this comment thread, but would appreciate the addition of unit tests so that the conditions surrounding each message are well documented and insulated from potential regressions.

Issue #1676, opened 2018-03-07
Implements catch-all Python Exception Interpreter


AlexCatarino:
#### Expected Behavior

When python algorithms throw an exception that is not supported by implemented interpreters, return an exception with python stack trace.

#### Actual Behavior

No python stack trace is shown.

#### Potential Solution

Implements a catch-all interpreter that would add the internal stack trace to the message.

#### Reproducing the Problem

```python
class BasicTemplateAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2013,10,7)
        self.SetEndDate(2013,10,7)
        self.AddEquity('SPY')

    def foo(self):
        self.bar()

    def bar(self):
        x = 0
        y = 1
        z = y / x
        pass

    def OnData(self, data):
        self.foo()
        self.SetQuit(True)
```

#### System Information

QuantConnect Cloud.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #1687, opened 2018-03-13
 Implements catch-all Python Exception Interpreter

AlexCatarino:
#### Description

Implements `PythonExceptionInterpreter` that works as a "catch-all" exception intrepreter for `PythonException`. Since it can interpret more specific python exceptions that are already covered by other interpreters, its `Order` is set to `int.MaxValue` to be the last to be tested.

#### Related Issue

#1676

#### Motivation and Context

When python algorithms throw an exception that is not supported by implemented interpreters, instead of returning an exception with python stack trace, the python stacktrace were missing.

No.

Unit tests and algorithm provided in related issue

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Exceptions/PythonExceptionInterpreter.cs
+++ b/Common/Exceptions/PythonExceptionInterpreter.cs
@@ -1,0 +1,53 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


Common/Exceptions/PythonExceptionInterpreter.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Interprets the specified exception into a new exception
+        /// </summary>
+        /// <param name="exception">The exception to be interpreted</param>
+        /// <param name="innerInterpreter">An interpreter that should be applied to the inner exception.
+        /// <returns>The interpreted exception</returns>
+        public Exception Interpret(Exception exception, IExceptionInterpreter innerInterpreter)
+        {
+            var pe = (PythonException)exception;
+
+            var message = pe.Message + PythonUtil.PythonExceptionStackParser(pe.StackTrace);
```
AlexCatarino: I think we could move `PythonExceptionStackParser` from `PythonUtil` to this class since it is specific to python exception interpreters. What would be the best approach? A static method or a "normal" method? If the second, we could make all python exception interpreters inherit from `PythonExceptionInterpreter`.

mchandschuh: I kind of like it as a static method. Using inheritance solely for code reuse, in my opinion, is an anti-pattern since it limits the reuse to only sub-types. Leaving this as a static, stateless method makes it consumable anywhere and everywhere while keeping it easy to test.

Now, if we end up in a situation where we start adding options defining *how* a stack trace is parsed, I would recommend we put it into it's own class (preferably behind an interface to enable easier testing) and then use constructor dependency injection to give the implementation to all the python exception interpreters.

Issue #1699, opened 2018-03-15
Exception thrown while trying to login using VS plugin


Martin-Molinero:
#### Expected Behavior

No exception is thrown while trying to login

#### Actual Behavior

Exception "_The type initializer for 'QuantConnect.Securities.MarketHoursDatabase' threw an exception._" is thrown.

#### Potential Solution

I believe there is a state, different than the default, in which there is something funny happening with the dataFolder value

#### Reproducing the Problem

By default the issue is not present. I think deleting C:\Users\UserName\AppData\Local\Microsoft\VisualStudio\15.0_8ab640ac\privateregistry.bin caused a different state after which the issue occurs always

#### System Information

Win 10, Visual studio 2017

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

jaredbroad:
Fixed in #1708 and #1700

PR #1700, opened 2018-03-15
Fix exception thrown at login of VS Plugin

Martin-Molinero:
#### Description

Moved __MarketHoursDatabase_ _marketHoursDatabase_ from **API.cs** to **LiveTradingRealTimeHandler.cs**, since it was only used by LiveTradingRealTimeHandler. Code path throwing exception will not be executed while trying to login using the VS plugin.

#### Related Issue

https://github.com/QuantConnect/Lean/issues/1699

#### Motivation and Context

Visual Studio Plugin could not be used in some cases.

No

Manual testing
#### Types of changes

#### Checklist:

### DIFF
--- a/Api/Api.cs
+++ b/Api/Api.cs
@@ -36,7 +36,6 @@ namespace QuantConnect.Api
     public class Api : IApi
     {
         private ApiWebSocketConnection _socketConnection;
-        private static MarketHoursDatabase _marketHoursDatabase;
         private string _dataFolder;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


jaredbroad:
Note for future: The `API` class uses the data folder for the DownloadFiles() helper -- but the plugin doesn't use it at the moment. So if we ever make a "download files plugin feature" we'll have to address this data folder issue again.

mchandschuh:
A note for posterity:
It appears that with this change the real time handlers no longer require a reference to the api implementation. Also, w/ this change we remove the ability to plug-in a different market hours implementation. In the future, we may want to make this it's own pluggable component, for now I think what's here is good.

Issue #1704, opened 2018-03-15
Add Csharp and Python examples for custom indicators with custom Quandl data


jingwu74:
Add algorithm examples to show how to use custom Quandl VIX and VXV data and Indicator Extensions 
#### Checklist

- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

PR #1705, opened 2018-03-15
Add Python and Csharp Examples of CustomDataIndicatorExtensions Algorithm

jingwu74:
#### Description

Add Python and Csharp Examples of CustomDataIndicatorExtensions Algorithm
#### Related Issue
Fixes [#1704](https://github.com/QuantConnect/Lean/issues/1704)

#### Motivation and Context

Request on the forum of adding examples of using multiple custom data with the custom indicator
#### Types of changes 

#### Checklist:

<description>`

### DIFF
--- a/Algorithm.CSharp/CustomDataIndicatorExtensionsAlgorithm.cs
+++ b/Algorithm.CSharp/CustomDataIndicatorExtensionsAlgorithm.cs
@@ -34,11 +34,11 @@ namespace QuantConnect.Algorithm.CSharp
     /// <meta name="tag" content="charting" />
     public class CustomDataIndicatorExtensionsAlgorithm : QCAlgorithm
     {
-        private const string VIX = "CBOE/VIX";
-        private const string VXV = "CBOE/VXV";
-        private SimpleMovingAverage smaVIX;
-        private SimpleMovingAverage smaVXV;
-        private IndicatorBase<IndicatorDataPoint> ratio;
+        private const string _vix = "CBOE/VIX";
+        private const string _vxv = "CBOE/VXV";
+        private SimpleMovingAverage _smaVIX;
+        private SimpleMovingAverage _smaVXV;
+        private IndicatorBase<IndicatorDataPoint> _ratio;
 
         /// <summary>
         /// Initialize the data and resolution you require for your strategy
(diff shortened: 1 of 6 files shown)


Algorithm.CSharp/CustomDataIndicatorExtensionsAlgorithm.cs
```diff
@@ -0,0 +1,95 @@
...
+    /// <meta name="tag" content="indicator classes" />
+    /// <meta name="tag" content="plotting indicators" />
+    /// <meta name="tag" content="charting" />
+    public class CustomDataIndicatorExtensionsAlgorithm : QCAlgorithm
+    {
+        private const string VIX = "CBOE/VIX";
+        private const string VXV = "CBOE/VXV";
+        private SimpleMovingAverage smaVIX;
+        private SimpleMovingAverage smaVXV;
+        private IndicatorBase<IndicatorDataPoint> ratio;
```
AlexCatarino: Minor comment:
In C#, we use the underscore pattern for private global variables: 
```csharp
private const string _vix = "CBOE/VIX";
private const string _vxv = "CBOE/VXV";
private SimpleMovingAverage _smaVIX;
private SimpleMovingAverage _smaVXV;
private IndicatorBase<IndicatorDataPoint> _ratio;
```

Algorithm.CSharp/CustomDataIndicatorExtensionsAlgorithm.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    MarketOrder(VIX, 100);
+                }
+                else if (ratio < 1)
+                {
+                    Liquidate();
+                }
+                // plot our indicators 
+                Plot("SMA", smaVIX, smaVXV);
+                Plot("Ratio", ratio);
```
AlexCatarino: In the python example, you used `PlotIndicator`, should both algorithms use the same methods for consistency?

Algorithm.Python/CustomDataIndicatorExtensionsAlgorithm.py
```diff
@@ -0,0 +1,84 @@
...
+        self.PlotIndicator("Ratio", self.ratio)
+        self.PlotIndicator("Data", self.vix_sma, self.vxv_sma)
+    
+    # OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+    def OnData(self, data):
+        
+        # Wait for all indicators to fully initialize
+        if not (self.vix_sma.IsReady and self.vxv_sma.IsReady and self.ratio.IsReady): return
+        if not self.Portfolio.Invested and self.ratio.Current.Value > 1:
+            self.MarketOrder('CBOE/VIX', 100)
```
AlexCatarino: In the C# case, you used a class field instead of the string.

Algorithm.Python/MultipleSymbolConsolidationAlgorithm.py
```diff
@@ -60,7 +60,7 @@ def Initialize(self):
             self.Data[symbol] = SymbolData(forex.Symbol, BarPeriod, RollingWindowSize)
 
         # loop through all our symbols and request data subscriptions and initialize indicator
-        for symbol, symbolData in self.Data.iteritems():
+        for symbol, symbolData in self.Data.items():
```
AlexCatarino: Great catch: python 3 dictionaries use `items` instead of `iteritems`.

Algorithm.CSharp/CustomDataIndicatorExtensionsAlgorithm.cs
disruptek: I suppose if we're trying to make a point or follow a convention while calling these `private` then we should do the same with the Python code and use the PEP 8 `_` or `__` prefix for the same reason.

Issue #1692, opened 2018-03-14
Cryto security is not using the more recent trade data when there is no quote available


Jay-Jay-D:
#### Expected Behavior

Crypto security uses two sources of data, tradebars up to tick resolution and quotebars up to minute resolution. 

The expected behavior is using the latest data, independently if it is quote or tradebar.

#### Actual Behavior

If there is no quote data available, the old quote data is stale and its not using the more recent trade data.

The engine will use the old quote data for the fills, but still sending trade data into the algorithm.

#### Potential Solution

Compare the timestamp of multi data feed securities and use the most recent point when quote data is stale inside the `GetLastData` method `SecurityCache`.

#### Reproducing the Problem

The algorithm bellow shows the issue.
Given there is *no* quote data from 03-Feb-2018 up to 07-Feb-2018:
- If start date is 02-Feb-2018, then it will be flat from 03-Feb-2018 up to 07-Feb-2018
- If start date is 05-Feb-2018, then it will works fine, reading the new quote data the 07-Feb-2018

```cs
using QuantConnect.Brokerages;
using QuantConnect.Data;
using QuantConnect.Securities;

namespace QuantConnect.Algorithm.CSharp
{
    public class BasicTemplateCryptoAlgorithm : QCAlgorithm
    {
        private Security _btc;

        public override void Initialize()
        {
            SetStartDate(year: 2018, month: 2, day: 2);
            SetEndDate(year: 2018, month: 2, day: 10);
            SetCash(startingCash: 100000);

            SetBrokerageModel(BrokerageName.GDAX, AccountType.Cash);

            _btc = AddCrypto("BTCUSD", Resolution.Minute);
            var chart = new Chart("BTCUSD_Price");
            AddChart(chart);
        }

        public override void OnData(Slice data)
        {
            if (!Portfolio.Invested)
            {
                SetHoldings(_btc.Symbol, 1);
            }
            Plot("BTCUSD_Price", "BTCUSD_SlicePrice", data[_btc.Symbol].Price);
            Plot("BTCUSD_Price", "BTCUSD_SecurityClose", _btc.Price);
            Plot("BTCUSD_Price", "BTCUSD_SecurityAskClose", _btc.AskPrice);
            Plot("BTCUSD_Price", "BTCUSD_SecurityBidClose", _btc.BidPrice);
        }
    }
}

```

#### System Information

Cloud

#### Checklist

- [x] I have completely filled out this template
[...]

Issue #21, opened 2015-02-02
History Retrieval


jaredbroad:
Implement a system for retrieving historical data as an array. The QCAlgorithm helper methods should match the interface below:

Get all data from start to end; for specified symbol:
`IEnumerable<BaseData> History(string symbol, DateTime start, DateTime end);`

Overload: Get all data from *now* back a timspan period for symbol and return it as an enumerable list:
`IEnumerable<BaseData> History(string symbol, TimeSpan period);`

QCAlgorithm should have an IHistoryManager plugin point which is set by Lean Engine according to the DataFeed. The IDataFeed would provide a GetHistoryManager method which returns a IHistoryManager, and that is set in the algorithm.

IHistoryManager implementation for the FileSystemDataFeed would implement this with a physical pull from the file system for the associated dates.

jaredbroad:
Feedback from a user: This is better implemented in a "Warm Up" style -- where the user actually requests a fixed period of data before the start date and the algorithm automatically warms up / pre-populates all the arrays as it normally would.

This makes the warm up process inline with normal algorithm code: e.g.
```
SetStartDate(2014, 1, 1);
SetWarmUp(TimeSpan.FromDays(14));
```

With this example data would be pumped in from Dec 15th, but orders would only be permitted after the 1st (or once the warm up period is completed in case of live trading).

jaredbroad:
Closed in 4ae1f4f5b61

PR #1706, opened 2018-03-15
bug-1692-use-tradebar-data-when-no-quotebar-data-is-available

Jay-Jay-D:
#### Description

This PR implements the following chages in the `SecurityCache` class:
- fill-forward data is not cached.
- in case of multiple `MarketDataType`it returns the one with the latest timestamp.
- prioritize the quotebars.

#### Related Issue

#1692 

#### Motivation and Context

Before this PR, if no quote data was available, then the algorithm equity  was flat until new quote data was available.

The method for testing was using an basic template algorithm (it buys 100% of one security at the beginning) and confirm the equity follows perfectly the security close data when there is only tradebar data available. 

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/SecurityCache.cs
+++ b/Common/Securities/SecurityCache.cs
@@ -15,7 +15,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
 
@@ -104,9 +103,9 @@ namespace QuantConnect.Securities
 
             // Only cache no fill-forward data.
             if (data.IsFillForward) return;
-            
+
             // If the _dataByType dictionary is not populated or it only has one MarketDataType, just use the last value.
-            if (_lastData == null)
+            if (_dataByType.Keys.Count <= 1)
             {
                 _lastData = data;
                 _dataByType[data.GetType()] = data;
--- a/Common/Securities/SecurityCache.cs
+++ b/Common/Securities/SecurityCache.cs
@@ -95,18 +95,18 @@ namespace QuantConnect.Securities
         /// </summary>
         public void AddData(BaseData data)
         {
-            // Only cache no fill-forward data.
-            if (data.IsFillForward) return;
-
             var openInterest = data as OpenInterest;
             if (openInterest != null)
             {
                 OpenInterest = (long)openInterest.Value;
                 return;
             }
+
+            // Only cache no fill-forward data.
+            if (data.IsFillForward) return;
             
             // If the _dataByType dictionary is not populated or it only has one MarketDataType, just use the last value.
-            if (_dataByType.Keys.Count <= 1)
+            if (_lastData == null)
             {
                 _lastData = data;
                 _dataByType[data.GetType()] = data;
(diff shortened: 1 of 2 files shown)


Jay-Jay-D:
The `BasicTemplateCryptoAlgorithm` sends a trade at each hour (buy or liquidate). And all have the same value, except for the filling price of the Order #21 fired at 2016-10-07 20:00 algorithm time, equivalent to 2016-10-08 00:00 UTC. 
In order to fill that order:
- The engine uses the correspondent price, the 2016-10-07 23:59 UTC minute price.
- The last time stamp of the file `20161007_quote.zip` is 86280000, equivalent to 2016-10-07 *23:58* UTC minute price
- With the actual `SecurityCache` behavior, the 2016-10-07 23:58 UTC minute price (618.33) is fill forward as *quote* and the engine uses that price for filling the order #21.
- With the suggested change in the `SecurityCache`, the engine uses the 2016-10-07 23:59 UTC *tradebar* price (617.98) from the `20161007_trade.zip`.

That small difference explains the changes in the regression backtest statistics.

Common/Securities/SecurityCache.cs
```diff
@@ -101,10 +101,27 @@ public void AddData(BaseData data)
...
             }
 
-            _lastData = data;
-            _dataByType[data.GetType()] = data;
+            // Only cache no fill-forward data.
+            if (data.IsFillForward) return;
 
-            var tick = data as Tick;
+            // If the _dataByType dictionary is not populated or it only has one MarketDataType, just use the last value.
+            if (_dataByType.Keys.Count <= 1)
```
jaredbroad: Will need to refactor this monday but will merge for now.

Common/Securities/SecurityCache.cs
```diff
@@ -101,10 +101,27 @@ public void AddData(BaseData data)
...
+            {
+                _lastData = data;
+                _dataByType[data.GetType()] = data;
+            }
+            else
+            {
+                // If there are more then one MarketDataType then update _lastData only if is the latest, prioritizing QuoteBars
+                if (data.Time > _lastData.Time || data.Time == _lastData.Time && data.DataType == MarketDataType.QuoteBar)
```
mchandschuh: We need to check on the `EndTime` property.

Tests/RegressionTests.cs
```diff
@@ -705,19 +705,19 @@ private static TestCaseData[] GetRegressionTestParameters()
...
                 {"Profit-Loss Ratio", "0"},
                 {"Alpha", "-11.165"},
-                {"Beta", "574.85"},
-                {"Annual Standard Deviation", "0.353"},
-                {"Annual Variance", "0.125"},
-                {"Information Ratio", "-20.427"},
-                {"Tracking Error", "0.353"},
-                {"Treynor Ratio", "-0.013"},
-                {"Total Fees", "$6076.08"}
+                {"Beta", "585.081"},
```
mchandschuh: I'd be curious how these specifically changed. For example, were we previously filling using fill-forward trade data? Or perhaps we were filling using trade data and now it's filling using quote data? Changes in regression can be expected with bugfixes but I think it's critically important to be able to completely explain *why* the expected results have changed.

PR #1720, opened 2018-03-19
Apply security changes at end of time step

### LINKED ISSUE #1689: Adding multiple symbols resets option underlying asset data normalization to adjusted.
#### Expected Behavior

The price normalization mode should be raw automatically for options algorithms no matter how many underlying assets we add in the algorithm.
#### Actual Behavior

If add more than one underlying assets in options algorithms with AddEquity(), the underlying price will be the adjusted price.

#### Reproducing the Problem

The problem could be reproduced by cloning this algorithm (The algorithm will log the underlying price of IBM and AAPL from 4/24/2017 to 5/10/2017)
```

    def Initialize(self):        
        self.SetStartDate(2017,4,24)
        self.SetEndDate(2017,5,10)
        self.SetCash(500000)
        # self.stocks = ["IBM"]
        self.stocks = ["AAPL","IBM"]
        self.trade = False
        for stock in self.stocks:
            self.AddEquity(stock, Resolution.Minute)
            option = self.AddOption( stock, Resolution.Minute)
            option.SetFilter(-1, +1, timedelta(20), timedelta(30))        
        self.Schedule.On(self.DateRules.EveryDay(), self.TimeRules.BeforeMarketClose(self.stocks[0], 0), Action(self.activate))

    def OnData(self, data):
        if self.trade:
            for i in data.OptionChains:
                optionchain = list( i.Value )
                if len(optionchain) == 0: continue
                contract = optionchain[0]
                self.Log(str( contract.UnderlyingSymbol ) +" UnderlyingLastPrice: " +str(contract.UnderlyingLastPrice))
        self.trade = False
    
    def activate(self):
        self.trade = True
```
To make a comparison,  change `self.stocks = ["AAPL","IBM"]` to `self.stocks = ["IBM"]` or `        self.stocks = ["AAPL"]`, the underlying price will be Raw price automatically.

#### System Information

LEAN cloud

#### Checklist

mchandschuh:
#### Description

When calling `AddSecurity` methods, we now save the changes to a pending list to be applied at the end of the current time step. This gives the algorithm a chance to properly configure the newly created security object (normalization mode was original culprit here) *before* the data feed grabs it and creates a subscription for it. Once the DF has created a subscription for it, any further changes to the subscription properties will not take effect.

#### Related Issue

Fixes #1689 

#### Motivation and Context

Options algorithms were running with the adjust data normalization mode because by the time we configured it to use raw mode, the data feed had already grabbed the subscription and started enumerating the data source.

No.

Extensive manual testing. There is a related options regression algorithm that I have locally, but it has some issues I'm still trying to understand, but didn't want that learning exercise to prevent this bug fix from getting into `master`

#### Types of changes

#### Checklist:

>NOTE: `OptionExerciseAssignRegressionAlgorithm` is a flickering test. It appears to be caused by the ordering of fills. It passes 100% of the time when I run it by itself, but 50% when run as part of the entire regression suite. I confirmed this issue also exists on the current `master`. Also, no additional tests were added to cover this, but one is forthcoming. I'm creating an options regression algorithm that hopefully covers many of the common options issues, but it requires more time to properly validate the results.

### DIFF
--- a/Algorithm/QCAlgorithm.Universe.cs
+++ b/Algorithm/QCAlgorithm.Universe.cs
@@ -26,6 +26,13 @@ namespace QuantConnect.Algorithm
 {
     public partial class QCAlgorithm
     {
+        // save universe additions and apply at end of time step
+        // this removes temporal dependencies from w/in initialize method
+        // original motivation: adding equity/options to enforce equity raw data mode
+        private readonly object _pendingUniverseAdditionsLock = new object();
+        private readonly Dictionary<Security, UserDefinedUniverse> _pendingUserDefinedUniverseSecurityAdditions = new Dictionary<Security, UserDefinedUniverse>();
+        private readonly List<Universe> _pendingUniverseAdditions = new List<Universe>();
+
         /// <summary>
         /// Gets universe manager which holds universes keyed by their symbol
         /// </summary>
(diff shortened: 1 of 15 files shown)


mchandschuh:
fixed typo in `pendingUniverseAdditions`

PR #1728, opened 2018-03-20
Fixes AverageTrueRange indicator

### LINKED ISSUE #1722: ATR calculations do not follow J. Welles Wilder's book instructions
Like the RSI test data (as I already reported here https://github.com/QuantConnect/Lean/issues/1651), I found that the test data for the `Average True Range` is not correct either.

Here is the link to your test data:
https://github.com/QuantConnect/Lean/blob/master/Tests/TestData/spy_atr.txt

If your ATR indicator is tested against this data and successfully passes the test, it is likely that the algorithm is not based on the original ATR developed by J.W. Wilder.

Please download the test data from stockcharts. They got it right:
http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:average_true_range_atr

Take the stockcharts data and feed your indicator with it. Do the results match?

AlexCatarino:
#### Description
Since the first value of `AverageTrueRange` is the mean of `TrueRange` of the warm up period, we have introduced a `SimpleMovingAverage` that handles that calculation. Its value (`Current.Value`) is used to set the smoothing function value

#### Related Issue
Fixes #1722 

#### Motivation and Context
Review indicators that do not obey its definition.

No.

Unit tests.
We have added a new unit test for the case where `AverageTrueRange` is smoothed by Wilder's moving average. 

#### Types of changes

#### Checklist:


Indicators/AverageTrueRange.cs
```diff
@@ -116,6 +121,12 @@ protected override decimal ComputeNextValue(IBaseDataBar input)
             // compute the true range and then send it to our smoother
             TrueRange.Update(input);
             _smoother.Update(input.Time, TrueRange);
+            _sma.Update(input.Time, TrueRange);
```
mchandschuh: Can this be placed in the if block below? No need to compute after initialized.

AlexCatarino:
Replaced by #1730.

PR #1769, opened 2018-03-26
Update GDAX fill handling for market orders

### LINKED ISSUE #1751: GDAX "Insufficient buying power" error message on Liquidate
#### Expected Behavior
We should never get an _"Insufficient buying power"_ error message when calling Liquidate.

#### Actual Behavior
When a previous call to `SetHoldings` calculates a Buy quantity based on the quote currency in the account and the current ask price, the order is submitted, but is not filled completely because of insufficient liquidity on the order book at the ask price (not enough funds to buy the quantity requested at higher prices). When this happens, GDAX does not reject the buy order, it just fills as much as it can, so we end up with a series of partial fills (with a total quantity filled slightly less than the initial order quantity).
At this point the portfolio holds a quantity equal to the sum of the partial fills and the buy market order still has an order status of `PartiallyFilled`, with a (usually small) remaining quantity that will never be executed.
The current implementation of `Liquidate` includes unfilled market order quantities in the liquidation order, causing the _"Insufficient buying power"_ error, because the sell order quantity is greater than the quantity in the cash book.

#### Potential Solution
A potential solution to this issue includes these steps:
1. Ensure we set the order status to `Filled` on the last fill we receive: this would require waiting for the first _match_ (trade) message for ~~another user~~ a different order and slightly delaying the firing of market order fill events
2. ~~Change the implementation of Liquidate to include unfilled quantities in the liquidation order **only** if the order status is **not** equal to `Filled`.~~ (no need for this: if the order status is `Filled` the order will be automatically excluded)

#### Reproducing the Problem
This issue is not easily repeatable, as it requires submitting a real buy market order for a quantity larger than the quantity currently on the order book at the best ask price. 
I was able to repeat it by temporarily disabling the buying power model check in the `BrokerageTransactionHandler`, submitting a buy market order on `BTCEUR` for a size much greater than the available EUR in my account -- easier to repeat vs. waiting for the best ask size to be small enough to get multiple partial fills at higher prices in the order book.

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
Partially filled GDAX market orders are now guaranteed to transition to the final state of `Filled`, even if the total quantity filled is less than the order quantity.
Now we are (slightly) delaying market order fill events until messages for a different order are received.
Order fill event handling is performed entirely in the `"match"` message event handler.

The `"done"` message event handler was also removed since it was unused (these messages were never received because we are not subscribing to the `"full"` channel).

#### Related Issue
Fixes #1751

#### Motivation and Context
If a previous market buy order was not filled completely because of insufficient funds, the market order was never being set to status `Filled`, causing an _"Insufficient buying power"_ error when calling `Liquidate` later on.

No.

Manual testing as described in #1751.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Brokerages/GDAX/GDAXTestHelpers.cs
+++ b/Tests/Brokerages/GDAX/GDAXTestHelpers.cs
@@ -1,3 +1,18 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
 using QuantConnect.Brokerages.GDAX;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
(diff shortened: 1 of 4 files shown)


Tests/Brokerages/GDAX/GDAXTestHelpers.cs
```diff
@@ -9,6 +24,7 @@ namespace QuantConnect.Tests.Brokerages.GDAX
 {
     public class GDAXTestsHelpers
     {
+        private static readonly Symbol Btcusd = Symbol.Create("BTCUSD", SecurityType.Crypto, Market.GDAX);
```
mchandschuh: No need to change, but could alternatively use `Symbols.BTCUSD` from within the test project.

Issue #1788, opened 2018-03-27
Build custom portfolio construction model with mean variance optimization


jingwu74:
#### Expected Behavior

There should be a template for mean-variance optimization model to calculate the optimal weights for assets in the portfolio based on modern portfolio theory. 
#### Actual Behavior

There is no such custom portfolio construction model in algorithm framework.
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

AlexCatarino:
Closed by #1832.

PR #1789, opened 2018-03-27
Add mean variance optimization portfolio construction model

jingwu74:
#### Description

This custom portfolio construction model is a template of mean-variance optimization. The model calculates the optimal weights for risky assets in the portfolio based on modern portfolio theory.
#### Related Issue

Fixes [#1788](https://github.com/QuantConnect/Lean/issues/1788)
#### Motivation and Context

This custom model can be used as a template to find optimal weights in long-short strategies. The model attempts to maximize the annual return and minimize the annual variance based on the historical return. The resolution, the time interval of history price and the boundary of the weight can be changed based on the algorithm. 

The model is tested on assets with strong uptrend and downtrend. More weights would be given to uptrend stock.
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
+++ b/Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
@@ -78,7 +78,7 @@ class MeanVarianceOptimizationPortfolioConstructionModel:
         weights = PortfolioOptimization(daily_return, self.min_weight, self.max_weight).opt_portfolio() 
         # Create portfolio targets from the specified insights
         for insight in insights:
-            yield PortfolioTarget.Percent(algorithm, insight.Symbol, insight.Direction * weights[str(insight.Symbol)])
+            yield PortfolioTarget.Percent(algorithm, insight.Symbol, weights[str(insight.Symbol)])
         
     def OnSecuritiesChanged(self, algorithm, changes):
         """ 


Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
```diff
@@ -0,0 +1,136 @@
...
+        if len(self.securities) == 0:
+            return [ ]
+        
+        symbols = [insight.Symbol for insight in insights]
+        # request the daily history price of last year
+        hist = algorithm.History(symbols, self.interval, self.resolution)
+        df_price = hist['close'].unstack(level=0)
+        # calculate the daily return 
+        daily_return = (df_price / df_price.shift(1)).dropna()
+        weights = PortfolioOptimization(daily_return, self.min_weight, self.max_weight).opt_portfolio()
```
mchandschuh: Thinking out loud here...

My understanding of MPT is that it centers around the concept of *expected* returns. While it is common to use historical returns, we're actually generating predictions, some of which will have a magnitude which _is_ the expected return for the asset. I'm thinking it may make sense to use the magnitude from the insight if available as the expected return, and if not available, can fall back on historical returns as a proxy. I'm not sure how that would tie into what's here, but IMO it's worth consideration.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
```diff
@@ -0,0 +1,136 @@
...
+        hist = algorithm.History(symbols, self.interval, self.resolution)
+        df_price = hist['close'].unstack(level=0)
+        # calculate the daily return 
+        daily_return = (df_price / df_price.shift(1)).dropna()
+        weights = PortfolioOptimization(daily_return, self.min_weight, self.max_weight).opt_portfolio() 
+        weights.index = sorted([insight.Symbol.Value for insight in insights])
+        
+        # Create portfolio targets from the specified insights
+        for insight in insights:
+            yield PortfolioTarget.Percent(algorithm, insight.Symbol, insight.Direction * weights[insight.Symbol.Value])
```
mchandschuh: It's important to **never** index a dictionary using `Symbol.Value` as it's not guaranteed to be unique. Please just use the symbol object directly as the dictionary key.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
```diff
@@ -0,0 +1,136 @@
...
+                self.securities.remove(removed) 
+ 
+class PortfolioOptimization(object):
+    """ 
+    Construct the mean variance optimization model with the portfolio return
+    Args:
+       df_return(dataframe): Dataframe of porfolio return indexed by the timestamp
+       min_weight(float): The lower bounds on portfolio weights
+       max_weight(float): The upper bounds on portfolio weights
+       n(int): The number of the risk assets in the portfolio
```
mchandschuh: Love the documentation!

Algorithm.Framework/QuantConnect.Algorithm.Framework.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
mchandschuh: Please revert this change from the project file.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
jingwu74: Yes. In theory, MPT should have an expected return for each asset which is commonly the mean of historical return. Then the objectives are considered to minimize risk given a specific return. That's one of the problems MPT can solve. Another one is maximizing return for a given level of risk. Here I try to solve a numerical optimization problem to optimize both the return the variance (maximize the Sharpe ratio).

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
```diff
@@ -0,0 +1,138 @@
...
+        symbols = [insight.Symbol for insight in insights]
+        # request the daily history price of last year
+        hist = algorithm.History(symbols, self.interval, self.resolution)
+        df_price = hist['close'].unstack(level=0)
+        # calculate the daily return 
+        daily_return = (df_price / df_price.shift(1)).dropna()
+        weights = PortfolioOptimization(daily_return, self.min_weight, self.max_weight).opt_portfolio() 
+        # Create portfolio targets from the specified insights
+        for insight in insights:
+            yield PortfolioTarget.Percent(algorithm, insight.Symbol, insight.Direction * weights[str(insight.Symbol)])
```
mchandschuh: Are these weights already directional? IOW, are we potentially negating a negative weight by multiplying by the direction?

jingwu74: Right. Weights already have the direction. There's no need to time the insight.Direction.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
```diff
@@ -0,0 +1,138 @@
...
+        Args:
+            algorithm:The algorithm instance
+            insights: The insights to create portoflio targets from
+        Returns: 
+            An enumerable of portoflio targets to be sent to the execution model
+        """
+        
+        symbols = [insight.Symbol for insight in insights]
+        # request the daily history price of last year
+        hist = algorithm.History(symbols, self.interval, self.resolution)
```
AlexCatarino: Instead of making a Historical data request on every `CreateTargets` call, you could make just one in `OnSecuritiesChanged`. You would use the values to update a `SimpleMovingAverage` and `StandardDeviation` indicators for each symbol.
You would have a dictionary keyed by `Symbol` with `SymbolData` (see `MacdAlphaModel` for example) will all the variables you need to compute the weights.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstruction.py
```diff
@@ -0,0 +1,136 @@
...
+
+### <summary>
+### Provides an implementation of Mean-Variance portfolio optimization based on modern portfolio theory.
+### The target percent holdings of each security is calculated by maximizing the Sharpe ratio. 
+### The interval of weights in optimization method can be changed based on the long-short algorithm.
+### The default model uses the last three months daily price to calculate the optimal weight 
+### with the weight range from -1 to 1  
+### </summary>
+class MeanVarianceOptimizationPortfolioConstructionModel:
+    def __init__(self, resolution=Resolution.Daily, interval=63, min_weight=-1, max_weight=1):
```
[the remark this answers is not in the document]
jaredbroad: `def __init__(self, minimization_function=func, func_arguements=[], resolution=Resolution.Daily, interval=63, min_weight=-1, max_weight=1):`

Please define a input function for the minimization function; use helpers like this (https://stackoverflow.com/questions/3134829/can-i-have-two-init-functions-in-a-python-class) for creating:

`MeanVarianceOptimizationPortfolioConstructionModel.MaximizeSharpe()`
`MeanVarianceOptimizationPortfolioConstructionModel.MinimizeVolatility( target_return = 0.001 )`

Issue #1798, opened 2018-03-28
Implements python version of each framework modules/models


AlexCatarino:
#### Expected Behavior
All framework modules/models (Alphas, Execution, Portfolio, Risk and Selection) should have both C# and python versions.

#### Actual Behavior
Missing python versions.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

PR #1805, opened 2018-03-29
Adds python versions of available Alpha Models

AlexCatarino:
#### Description
- [x] Implements a framework to perform unit tests of alpha models
- [ ] Adds ConstantAlphaModel
- [ ] Adds EmaCrossAlphaModel
- [ ] Adds MacdAlphaModel
- [ ] Adds RsiAlphaModel

#### Related Issue
#1798 

#### Motivation and Context
All framework modules/models (Alphas, Execution, Portfolio, Risk and Selection) should have both C# and python versions.

Unit test.

#### Types of changes

#### Checklist:


Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
```diff
@@ -0,0 +1,265 @@
...
+        [TestCase(Language.Python)]
+        public void AddedSecuritiesTest(Language language)
+        {
+            IAlphaModel model;
+            if (!TryCreateAlphaModel(language, out model))
+            {
+                Assert.Ignore($"Ignore {GetType().Name}: Could not create {language} model.");
+            }
+
+            var changes = new SecurityChanges(AddedSecurities, RemovedSecurities);
```
mchandschuh: since we're adding `AddedSecurities` as part of initialization, I think this invocation turns into a noop. Could add another random security to the added list to ensure it turns the crank

Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
```diff
@@ -0,0 +1,265 @@
...
+
+                    var utc = request.StartTimeUtc;
+                    var end = request.EndTimeUtc;
+                    var span = request.Resolution.ToTimeSpan();
+                    var range = 0;
+
+                    while (utc < end)
+                    {
+                        var time = utc.ConvertFromUtc(sliceTimeZone);
+                        if (request.ExchangeHours.IsOpen(time, request.IncludeExtendedMarketHours))
```
mchandschuh: There's actually a method that does this in `Time.cs`:
```csharp
public static DateTime GetStartTimeForTradeBars(SecurityExchangeHours exchange, DateTime end, TimeSpan barSize, int barCount, bool extendedMarketHours)
{
    if (barSize <= TimeSpan.Zero)
    {
        throw new ArgumentException("barSize must be greater than TimeSpan.Zero", nameof(barSize));
    }

    var current = end.RoundDown(barSize);
    for (int i = 0; i < barCount;)
    {
        var previous = current;
        current = current - barSize;
        if (exchange.IsOpen(current, previous, extendedMarketHours))
        {
            i++;
        }
    }
    return current;
}
```

Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
```diff
@@ -0,0 +1,265 @@
...
+            var actualArray = allInsights.ToArray();
+            var expectedArray = ExpectedInsights().ToArray();
+
+            Assert.AreEqual(actualArray.Length, expectedArray.Length);
+
+            for (var i = 0; i < actualArray.Length; i++)
+            {
+                var actual = actualArray[i];
+                var expected = expectedArray[i];
+                Assert.True(actual.Equals(expected));
```
mchandschuh: Interesting note here, `Insight.Equals(Insight)` only compares based on the prediction properties, such as type/direction/price/magnitude/confidence -- but not on other properties, such as id/gen-time/close-time/scores/values/ect.

The thought was that users would inevitably end checking of their prediction 'has changed' and making that work. Maybe initially that was a bad idea. The reason I bring this is up is because these tests don't confirm the *time* at which these insights are emitted. I think it's very important that we also test the `GeneratedTimeUtc` value.

Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
AlexCatarino: `Time.GetStartTimeForTradeBars` is used to get the start time knowing the bar count, while we want the other way around. Maybe I could add `Time.GetBarsForPeriod` and use it instead.

Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
AlexCatarino: Added 
```csharp
Assert.AreEqual(actual.GeneratedTimeUtc, expected.GeneratedTimeUtc);
```
Now we have to change all the model to include that value and unit test `ExpectedInsights` accordingly.

mchandschuh: The generated time is set by the algorithm in `QCAlgorithmFramework.OnFrameworkData` -- recommend setting it to the current slice's end time in the runner base class.

PR #1781, opened 2018-03-26
Add rsi alpha model

mchandschuh:
#### Description

Defines a new `RsiAlphaModel` that uses Wilder's RSI as a signal source.
This PR also includes a fix for `PortofolioTarget.Percent` to properly use the buying power model when determining how much quantity to purchase. Without this change, 99% of the generated orders were failing as invalid (insufficient funds), but now they're working :)

#### Related Issue

Closes #1779 

#### Motivation and Context

This is part of a continued effort to fill out some example models in the framework ecosystem.

#### Requires Documentation Change

Yes. This should be included in documentation.

#### How Has This Been Tested?

QC production.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/ect)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [ ] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`

PR #1810, opened 2018-03-29
Add VWAP and STD execution models

### LINKED ISSUE #1784: Create VWAP Execution Model
#### Description
Create a new `VwapExectionModel`. We'll define VWAP execution as placing market orders while the security's current price is below the current VWAP. A parameter should be provided that is used to determine the maximum size of each order submitted, ideally leveraging the security's current trading volume as a measure of market depth in place of having access to order book data.

#### Checklist

mchandschuh:
#### Description

Creates two new execution models. First, `VolumeWeightedAveragePriceExecutionModel` will submit market orders while the price is more favorable than VWAP. Second, the `StandardDeviationExecutionModel` will submit market orders when the current price is at least the specified number of deviations away from the mean. Both models accept an implementation of `IOrderSizingStrategy` which is used to break up the order into multiple orders. Two implementations of `IOrderSizingStrategy` are also provided: `PercentVolumeOrderSizingStrategy` which defines a maximum order size as a specified percentage of volume from the current bar and the `TotalValueOrderSizingStrategy` which defines a maximum order size using a specified maximum order value.

#### Related Issue

Closes #1784 

#### Motivation and Context

Continuing to fill out the algorithm framework models namespace, providing some default models for users to use.

Yes! @jaredbroad -- please include these two models in the documentation.

Extensively tested in QC cloud in addition to a fair amount of local testing/debugging. The order sizing implementations are under unit test.

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue) (`PortoflioTarget.Percent` bugfix)

#### Checklist:

- [x] I have added tests to cover (**some**) my changes.  

Models are not under automated test at this time.

>NOTE: Some features were left out following review comments from https://github.com/QuantConnect/Lean/pull/1781. If agreed, I can re-add those features, including the ability to plot the underlying indicator values (VWAP/(SMA ± n*STD) and configurable resolution support for the indicators (right now they use security configured resolution). I had also considered providing support for asymmetric thresholds for buy/sell, but this was not implemented.

### DIFF
--- a/Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
+++ b/Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
@@ -15,6 +15,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using QuantConnect.Algorithm.Framework.Portfolio;
 using QuantConnect.Data.Consolidators;
 using QuantConnect.Data.UniverseSelection;
@@ -137,8 +138,11 @@ namespace QuantConnect.Algorithm.Framework.Execution
                 SymbolData data;
                 if (_symbolData.TryGetValue(removed.Symbol, out data))
                 {
-                    _symbolData.Remove(removed.Symbol);
-                    algorithm.SubscriptionManager.RemoveConsolidator(removed.Symbol, data.Consolidator);
+                    if (IsSafeToRemove(algorithm, removed.Symbol))
+                    {
+                        _symbolData.Remove(removed.Symbol);
+                        algorithm.SubscriptionManager.RemoveConsolidator(removed.Symbol, data.Consolidator);
+                    }
                 }
             }
         }
@@ -174,6 +178,15 @@ namespace QuantConnect.Algorithm.Framework.Execution
             }
 
             return false;
+        }
+
+        /// <summary>
+        /// Determines if it's safe to remove the associated symbol data
+        /// </summary>
+        private bool IsSafeToRemove(QCAlgorithmFramework algorithm, Symbol symbol)
+        {
+            // confirm the security isn't currently a member of any universe
+            return !algorithm.UniverseManager.Any(kvp => kvp.Value.ContainsMember(symbol));
         }
 
         class SymbolData
--- a/Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
+++ b/Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
@@ -1,0 +1,207 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using QuantConnect.Algorithm.Framework.Portfolio;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Indicators;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Framework.Execution
+{
+    /// <summary>
+    /// Execution model that submits orders while the current market prices is at least the configured number of standard
+    /// deviations away from the mean in the favorable direction (below/above for buy/sell respectively)
+    /// </summary>
+    public class StandardDeviationExecutionModel : IExecutionModel
+    {
+        private readonly int _period;
+        private readonly decimal _deviations;
+        private readonly Resolution _resolution;
+        private readonly PortfolioTargetCollection _targetsCollection;
+        private readonly Dictionary<Symbol, SymbolData> _symbolData;
+
+        /// <summary>
+        /// Gets or sets the maximum order value in units of the account currency.
+        /// This defaults to $20,000. For example, if purchasing a stock with a price
+        /// of $100, then the maximum order size would be 200 shares.
+        /// </summary>
+        public decimal MaximumOrderValue { get; set; } = 20 * 1000;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="StandardDeviationExecutionModel"/> class
+        /// </summary>
+        /// <param name="period">Period of the standard deviation indicator</param>
+        /// <param name="deviations">The number of deviations away from the mean before submitting an order</param>
+        /// <param name="resolution">The resolution of the STD and SMA indicators</param>
+        public StandardDeviationExecutionModel(
+            int period = 60,
+            decimal deviations = 2m,
+            Resolution resolution = Resolution.Minute
+            )
+        {
+            _period = period;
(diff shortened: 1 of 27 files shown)


mchandschuh:
fyi, this is on ice while we determine the best path forward to ensure that using these models is easy and intuitive.

mchandschuh:
Here's a view of `StandardDeviationExecutionModel` in action. The lines are:
```
Orange: SMA(60min) + 2*STD(60min)
Pink:   SPY(min)
Blue:   SMA(60min)
Green:  SMA(60min) - 2*STD(60min)
```
image
The model was asked to make a sell order for SPY at 1:00PM, but waiting until 1:08PM whenthe price moved at least 2 standard deviations away from the mean before submitting market orders. In addition, this model was configured to submit orders with a maximum order value of $20k, hence the multiple orders, one each minute:
image
Now, it wasn't done reaching it's target yet, so you can see later in the chart the price moves favorably again and we actually ended up submitting a market on open order for tomorrow at the close:
image

jaredbroad:
Note for future; normally with equities it is aiming to meet or better the VWAP for _day_ - so we could make it submit an order for any remaining quantity as a MOC order at 15.45 before MOC cut off. 

Those commit messages are great thank you. 

Whole PR TLDR; Pending problem we need to solve: given most people work/think in currency pairs (BTCUSD) - how do we intuitively & automatically create portfolios from alpha signals and execute on them.   For now handled as margin models and orders will get bounced.

Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
```diff
@@ -0,0 +1,217 @@
...
+                    // round down to even lot size
+                    orderSize -= orderSize % data.Security.SymbolProperties.LotSize;
+                    if (orderSize != 0)
+                    {
+                        algorithm.MarketOrder(symbol, Math.Sign(unorderedQuantity) * orderSize);
+                    }
+                }
+
+                // check to see if we're done with this target
+                unorderedQuantity = OrderSizing.GetUnorderedQuantity(algorithm, target);
```
jaredbroad: Note: we realize this could potentially never be 0 if you're trading cross pairs; as the portfolio.Quantity and cashbook[coin].Quantity will be incompatible. WIP until resolve how to model better.

Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
```diff
@@ -0,0 +1,217 @@
...
+            return false;
+        }
+
+        /// <summary>
+        /// Determines if it's safe to remove the associated symbol data
+        /// </summary>
+        private bool IsSafeToRemove(QCAlgorithmFramework algorithm, Symbol symbol)
+        {
+            // confirm the security isn't currently a member of any universe
+            return !algorithm.UniverseManager.Any(kvp => kvp.Value.ContainsMember(symbol));
```
jaredbroad: 👍

Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
```diff
@@ -0,0 +1,217 @@
...
+            public IDataConsolidator Consolidator { get; }
+
+            public SymbolData(QCAlgorithmFramework algorithm, Security security, int period, Resolution resolution)
+            {
+                Security = security;
+                Consolidator = algorithm.ResolveConsolidator(security.Symbol, resolution);
+                var smaName = algorithm.CreateIndicatorName(security.Symbol, "SMA" + period, resolution);
+                SMA = new SimpleMovingAverage(smaName, period);
+                var stdName = algorithm.CreateIndicatorName(security.Symbol, "STD" + period, resolution);
+                STD = new StandardDeviation(stdName, period);
```
jaredbroad: Later we can add a history call here to warm up the indicators

Algorithm.Framework/Execution/VolumeWeightedAveragePriceExecutionModel.cs
```diff
@@ -0,0 +1,282 @@
...
+
+                // fetch our symbol data containing our VWAP indicator
+                SymbolData data;
+                if (!_symbolData.TryGetValue(symbol, out data))
+                {
+                    continue;
+                }
+
+                // ensure we're receiving price data before submitting orders
+                if (data.Security.Price == 0m)
```
jaredbroad: Wonder if we can reuse the PreOrderChecks method. Wonder if these will grow over time

Algorithm.Framework/QCAlgorithmFramework.cs
```diff
@@ -91,6 +98,19 @@ public override void PostInitialize()
...
+            {
+                InsightsGenerated += (algorithm, data) => Log($"{Time}: {string.Join(" | ", data.Insights.OrderBy(i => i.Symbol.ToString()))}");
+            }
+
+            // emit warning message about using the framework with cash modelling
+            if (BrokerageModel.AccountType == AccountType.Cash)
+            {
+                Error("These models are currently unsuitable for Cash Modeled brokerages (e.g. GDAX) and may result in unexpected trades."
+                    + " To prevent possible user error we've restricted them to Margin trading. You can select margin account types with"
+                    + " SetBrokerage( ... AccountType.Margin) or substitute it for the ImmediateExecutionModel()");
```
jaredbroad: " or substitute it for the ImmediateExecutionModel()" -- this can be removed now; same issue for _all_ cash models :|

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -260,5 +257,21 @@ private void RemoveSecurityFromUniverse(Universe universe, Security member, List
...
+
+        /// <summary>
+        /// Determines if we can safely remove the security member from a universe.
+        /// We must ensure that we have zero holdings, no open orders, and no existing portfolio targets
+        /// </summary>
+        private bool IsSafeToRemove(Security member)
+        {
+            // but don't physically remove it from the algorithm if we hold stock or have open orders against it or an open target
+            var openOrders = _algorithm.Transactions.GetOrders(x => x.Status.IsOpen() && x.Symbol == member.Symbol);
+            if (!member.HoldStock && !openOrders.Any() && (member.Holdings.Target == null || member.Holdings.Target.Quantity == 0))
```
jaredbroad: Future me: `!member.HoldStock .... member.Holdings.Target.Quantity == 0` no holdings and not aiming for any more.

Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
mchandschuh: added! along w/ helper method

Algorithm.Framework/Execution/StandardDeviationExecutionModel.cs
mchandschuh: Included in that thought experiment should be consideration for what a portfolio target against a pair means without the concept of virtual positions?

Algorithm.Framework/Execution/VolumeWeightedAveragePriceExecutionModel.cs
mchandschuh: There's another flag, `IsTradable` -- currently only a public get/set and we use it to set internal feeds to false, I could see it being a little smarter and also returning false when price is zero, and perhaps other things we may check (assuming they relate directly to the security, ofc)

mchandschuh:
The `VolumeWeightedAveragePriceExecutionModel` aims to submit market orders when the price is more favorable than the current VWAP. This differs from a broker's VWAP execution in that they typically guarantee your order will fill (within some size limits) and that you'll receive at-or-better than VWAP by end of the day. This algorithm is much simpler in that it will only submit orders when the price is more favorable than the current VWAP.

Here's a view of the `VolumeWeightedAveragePriceExecutionModel`. The lines are:
```
Orange: IBM(min)
Green:  VWAP(IBM)
```
image
You can see the model executing sell orders while the price is great than the current VWAP. It then stops when the price falls, and when it rallies back above VWAP it begins selling again. Interesting to note that this is all the execution model reacting to a single $1MM IBM sell order.

PR #1832, opened 2018-04-04
Implements MeanVarianceOptimizationAlgorithm

AlexCatarino:
#### Description
This framework algorithm alpha model is  HistoricalReturnsAlphaModel and the  portfolio construction model is MeanVarianceOptimizationPortfolioConstructionModel.
This examples implements an algorithm that rebalances the portfolio according to modern portfolio theory.

#### Related Issue
#1798 
Closes #1789

#### Motivation and Context
Add more examples of framework algorithms.

No.

Running the newly created algorithm

#### Types of changes

#### Checklist:


Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.py
```diff
@@ -0,0 +1,80 @@
...
+        This is called each time the algorithm receives data for subscribed securities
+        Args:
+            algorithm: The algorithm instance
+            data: The new data available
+        Returns:
+            The new insights generated'''
+        insights = []
+        for symbol, symbolData in self.symbolDataBySymbol.items():
+            if symbolData.CanEmit:
+                insights.append(Insight(symbol, InsightType.Price, InsightDirection.Flat, self.predictionInterval, symbolData.Return, None))
```
jaredbroad: This shouldn't be flat; if its a positive return would be up / negative down.

jaredbroad: Please also update to emit variance insights for the MVO.

Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.py
```diff
@@ -0,0 +1,80 @@
...
+from datetime import timedelta
+
+class HistoricalReturnsAlphaModel:
+    '''Uses Historical returns to create insights.'''
+
+    def __init__(self, *args, **kwargs):
+        '''Initializes a new default instance of the HistoricalReturnsAlphaModel class.
+        Args:
+            resolution: The resolution of historical data'''
+        self.resolution = kwargs['resolution'] if 'resolution' in kwargs else Resolution.Daily
```
jaredbroad: The resolution should be set by the data coming into the method. Ideally this class should be agnostic

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
```diff
@@ -0,0 +1,165 @@
...
+            maximum_weight(float): The upper bounds on portfolio weights
+            target_return(float): The target portfolio return
+            optimization_method(method): Method used to compute the portfolio weights
+        """
+        self.resolution = kwargs['resolution'] if 'resolution' in kwargs else Resolution.Daily
+        self.period = kwargs['period'] if 'period' in kwargs else 63
+        self.minimum_weight = kwargs['minimum_weight'] if 'minimum_weight' in kwargs else -1
+        self.maximum_weight = kwargs['maximum_weight'] if 'maximum_weight' in kwargs else 1
+        self.target_return = kwargs['target_return'] if 'target_return' is kwargs else 2
+        self.optimization_method = kwargs['optimization_method'] if 'optimization_method' in kwargs else self.minimum_variance
```
jaredbroad: Fun :)

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
```diff
@@ -0,0 +1,165 @@
...
+            minimum_weight(float): The lower bounds on portfolio weights
+            maximum_weight(float): The upper bounds on portfolio weights
+            target_return(float): The target portfolio return
+            optimization_method(method): Method used to compute the portfolio weights
+        """
+        self.resolution = kwargs['resolution'] if 'resolution' in kwargs else Resolution.Daily
+        self.period = kwargs['period'] if 'period' in kwargs else 63
+        self.minimum_weight = kwargs['minimum_weight'] if 'minimum_weight' in kwargs else -1
+        self.maximum_weight = kwargs['maximum_weight'] if 'maximum_weight' in kwargs else 1
+        self.target_return = kwargs['target_return'] if 'target_return' is kwargs else 2
```
jaredbroad: I think this is daily?

Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.py
```diff
@@ -0,0 +1,80 @@
...
+            algorithm: The algorithm instance that experienced the change in securities
+            changes: The security additions and removals from the algorithm'''
+
+        # clean up data for removed securities
+        removed = [ x.Symbol for x in changes.RemovedSecurities ]
+        if len(removed) > 0:
+            for subscription in algorithm.SubscriptionManager.Subscriptions:
+                symbol = subscription.Symbol
+                if symbol in removed and symbol in self.symbolDataBySymbol:
+                    subscription.Consolidators.Clear()
```
mchandschuh: I don't think it's wise to clear _all_ consolidators... other models might have added consolidators and may still need them. The preferred solution is to keep track of consolidators you add and call `algorithm.SubscriptionManager.RemoveConsolidator(symbol, consolidator)`

Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.py
```diff
@@ -0,0 +1,80 @@
...
+                    self.symbolDataBySymbol.pop(symbol)
+
+        # initialize data for added securities
+        for added in changes.AddedSecurities:
+            if added.Symbol not in self.symbolDataBySymbol:
+                symbolData = SymbolData(added)
+                self.symbolDataBySymbol[added.Symbol] = symbolData
+                algorithm.RegisterIndicator(added.Symbol, symbolData.ROC, self.resolution)
```
mchandschuh: Do we want to fetch historical data to seed these need indicators?

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # initialize data for added securities
+        for added in changes.AddedSecurities:
+            if added.Symbol not in self.symbolDataBySymbol:
+                symbolData = SymbolData(added, self.period)
+                self.symbolDataBySymbol[added.Symbol] = symbolData
+
+                # seed new indicators using history request
+                history = algorithm.History(added.Symbol, self.period, self.resolution)
+                for row in history.itertuples():
```
mchandschuh: it may be a little messier, but issuing a single history request for all new ROC indicators will be much more efficient that issuing one a a time synchronously.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
```diff
@@ -0,0 +1,165 @@
...
+            algorithm: The algorithm instance that experienced the change in securities
+            changes: The security additions and removals from the algorithm'''
+
+        # clean up data for removed securities
+        removed = [ x.Symbol for x in changes.RemovedSecurities ]
+        if len(removed) > 0:
+            for subscription in algorithm.SubscriptionManager.Subscriptions:
+                symbol = subscription.Symbol
+                if symbol in removed and symbol in self.symbolDataBySymbol:
+                    self.symbolDataBySymbol.pop(symbol)
```
mchandschuh: When removing this symbol we should also remove the consolidator so that the memory can be cleared. To do this, we'll need to keep track of the consolidator instance, can use `algorithm.ResolveConsolidator` to get one that matches the configured resolution/primary tick type, and then can call `algorithm.SubscriptionManager.RemoveConsolidator(symbol, consolidator)` to clear out the events allowing the garbage collector to get rid of the attached indicators and consolidator.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
```diff
@@ -0,0 +1,165 @@
...
+        Returns: 
+            An enumerable of portoflio targets to be sent to the execution model
+        """
+        symbols = [insight.Symbol for insight in insights]
+        if len(symbols) == 0:
+            return []
+
+        for insight in insights:
+            symbolData = self.symbolDataBySymbol.get(insight.Symbol)
+            symbolData.Add(algorithm.Time, insight.Magnitude)
```
mchandschuh: Looks like this is building history of _expected_ returns for the covariance calculations -- I think we should be using historical returns.

Also might want to do a null check here on `insight.Magnitude` since it's not a required field, and throw if it is null w/ a good error message stating that the alpha model must emit a non-null `insight.Magnitude` to be used with this MVO PC model.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
```diff
@@ -0,0 +1,165 @@
...
+                history = algorithm.History(added.Symbol, self.period, self.resolution)
+                for row in history.itertuples():
+                    symbolData.ROC.Update(row.Index[1], row.close)
+
+
+    def minimum_variance(self, returns):
+        '''Minimum variance optimization method'''
+
+        # Objective function
+        fun = lambda weights: np.sqrt(np.dot(weights.T, np.dot(returns.cov(), weights)))
```
jingwu74: We could remove np.sqrt() here as the function name is minimum_variance. np.dot(weights.T, np.dot(returns.cov(), weights)) is greater or equal to zero It won't change the optimization result

Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.py
AlexCatarino: We subscribe to intraday data, but Portfolio rebalance is normally done at lower resolution.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
AlexCatarino: We use annual returns to compute the weights and consequently this is annual target return.

Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.py
AlexCatarino: No, but we should, since RateOfChange period should be configurable.

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py
jingwu74: Should we use percentage return 0.02?

PR #1872, opened 2018-04-16
Fixes RenkoConsolidator for python

### LINKED ISSUE #1742: RenkoConsolidator returns None in Python
#### Expected Behavior

`RenkoConsolidator().DataConsolidated` should return the Event Handler method in Python
#### Actual Behavior

`RenkoConsolidator` defined two `DataConsolidated` and pythonnet is picking the first that is null. 

#### Reproducing the Problem

```
class RenkoConsolidatorAlgorithm(QCAlgorithm):
    def Initialize(self):

        self.SetCash(100000)
        self.SetStartDate(2012, 1, 1)
        self.SetEndDate(2013, 1, 1)
        self.AddEquity("SPY")
        renkoClose = RenkoConsolidator(2.5)
        renkoClose.DataConsolidated += self.renkoCloseHandler
        self.SubscriptionManager.AddConsolidator("SPY", renkoClose)
        
    def renkoCloseHandler(self, sender, bar):
        self.Log("CLOSE - {0} - {1} {2}".format(str(data.Time), data.Open, data.Close))
```
#### System Information

LEAN Cloud
#### Checklist

AlexCatarino:
#### Description
- Missing `event` keyword prevented pythonnet to recognize `DataConsolidated` as a event handler.
- Adds python version of `RenkoConsolidatorAlgorithm`.

#### Related Issue
Closes #1742

#### Motivation and Context
It was not possible to attach events to renko consolidators.

No.

Existing tests and new algorithm.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Consolidators/RenkoConsolidator.cs
+++ b/Common/Data/Consolidators/RenkoConsolidator.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using Python.Runtime;
 using QuantConnect.Data.Market;
 
 namespace QuantConnect.Data.Consolidators
@@ -110,6 +111,32 @@ namespace QuantConnect.Data.Consolidators
             _volumeSelector = volumeSelector ?? (x => 0);
 
             Type = RenkoType.Classic;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RenkoConsolidator" /> class.
+        /// </summary>
+        /// <param name="barSize">The size of each bar in units of the value produced by <paramref name="selector"/></param>
+        /// <param name="selector">Extracts the value from a data instance to be formed into a <see cref="RenkoBar"/>. The default
+        /// value is (x => x.Value) the <see cref="IBaseData.Value"/> property on <see cref="IBaseData"/></param>
+        /// <param name="volumeSelector">Extracts the volume from a data instance. The default value is null which does
+        /// not aggregate volume per bar.</param>
+        /// <param name="evenBars">When true bar open/close will be a multiple of the barSize</param>
+        public RenkoConsolidator(decimal barSize, PyObject selector, PyObject volumeSelector = null, bool evenBars = true)
+            : this(barSize, evenBars)
+        {
+            if (barSize < Extensions.GetDecimalEpsilon())
+            {
+                throw new ArgumentOutOfRangeException("barSize", "RenkoConsolidator bar size must be positve and greater than 1e-28");
+            }
+
+            Func<IBaseData, decimal> selectorWrapper;
+            selector.TryConvertToDelegate(out selectorWrapper);
+            _selector = selectorWrapper ?? (x => x.Value);
+
+            Func<IBaseData, decimal> volumeSelectorWrapper;
+            volumeSelector.TryConvertToDelegate(out volumeSelectorWrapper);
+            _volumeSelector = volumeSelectorWrapper ?? (x => 0);
         }
 
         /// <summary>
--- a/Common/Data/Consolidators/RenkoConsolidator.cs
+++ b/Common/Data/Consolidators/RenkoConsolidator.cs
@@ -26,7 +26,7 @@ namespace QuantConnect.Data.Consolidators
         /// <summary>
         /// Event handler that fires when a new piece of data is produced
         /// </summary>
-        public EventHandler<RenkoBar> DataConsolidated;
+        public event EventHandler<RenkoBar> DataConsolidated;
 
         /// <summary>
         /// Event handler that fires when a new piece of data is produced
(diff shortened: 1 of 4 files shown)


Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -111,6 +113,32 @@ public RenkoConsolidator(decimal barSize, Func<IBaseData, decimal> selector, Fun
...
+            : this(barSize, evenBars)
+        {
+            if (barSize < Extensions.GetDecimalEpsilon())
+            {
+                throw new ArgumentOutOfRangeException("barSize", "RenkoConsolidator bar size must be positve and greater than 1e-28");
+            }
+
+            Func<IBaseData, decimal> selectorWrapper;
+            selector.TryConvertToDelegate(out selectorWrapper);
+            _selector = selectorWrapper ?? (x => x.Value);
```
mchandschuh: if we fail to convert it to a delegate *and* it's not null, shouldn't we throw an error? Presumably the user *thinks* their code is going to be used, but instead we'll just end up using the `Value` property.

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -111,6 +113,32 @@ public RenkoConsolidator(decimal barSize, Func<IBaseData, decimal> selector, Fun
...
+
+            Func<IBaseData, decimal> volumeSelectorWrapper;
+            volumeSelector.TryConvertToDelegate(out volumeSelectorWrapper);
+            _volumeSelector = volumeSelectorWrapper ?? (x => 0);
```
mchandschuh: I think the same goes here -- can do something like this for both of them -- and perhaps, this should be moved into a helper method that isn't a `TryConvertXXX` but is a `ConvertXXX` -- sometimes we want to throw if we can't convert and other times we don't care, but :
``` csharp
if (volumeSelector != null)
{
    if  (!volumeSelector.TryConvertToDelegate(out _volumeSelector))
    {
        throw new ArgumentException("Unable to convert parameter 'volumeSelector' to delegate type Func<IBaseData, decimal>");
    }
}
else
{
    _volumeSelector = (x => 0);
}
```

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -26,7 +27,7 @@ public class RenkoConsolidator : IDataConsolidator
         /// <summary>
         /// Event handler that fires when a new piece of data is produced
         /// </summary>
-        public EventHandler<RenkoBar> DataConsolidated;
+        public event EventHandler<RenkoBar> DataConsolidated;
```
mchandschuh: oh wow, this was a huge bug actually.

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -119,62 +147,39 @@ public RenkoConsolidator(decimal barSize, Func<IBaseData, decimal> selector, Fun
...
-            }
-        }
+        internal RenkoBar OpenRenkoBar => new RenkoBar(null, _openOn, _closeOn, _barSize, _openRate, _highRate, _lowRate, _closeRate);
 
         private void Rising(IBaseData data)
         {
             decimal limit;
 
-            while (_closeRate > (limit = (_openRate + BarSize)))
+            while (_closeRate > (limit = _openRate + _barSize))
```
mchandschuh: >NO CHANGE REQUIRED

Interesting choice to convert these usages, also somewhat surprised we didn't turn them into automatic, get-only property, since `_barSize` is read-only already.

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -351,21 +351,16 @@ public void Scan(DateTime currentLocalTime)
...
             Consolidated = consolidated;
         }
 
         /// <summary>Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.</summary>
         /// <filterpriority>2</filterpriority>
-        public void Dispose()
-        {
-            _dataConsolidatedHandler = null;
-        }
+        public void Dispose() => _dataConsolidatedHandler = null;
```
mchandschuh: Hmm, seems like an existing bug, but I think we can also set `DataConsolidated = null` here as well. As currently written, we're only clearing out event handlers on the `IDataConsolidator.DataConsolidated` event and not the well-typed `RenkoConsolidator.DataConsolidated` event.

Common/Data/Consolidators/RenkoConsolidator.cs
AlexCatarino: I left `BarSize` there, because it is a public property and removing it could break user's code.

mchandschuh: of course! I wasn't recommending removing the `BarSize` property, was just stating that it appears we don't even need a `_barSize` field, since we could define the property as:
``` csharp
public decimal BarSize { get; }
```

and then have everyone reference that property and completely remove the `_barSize` field -- but no matter, what's here is great!

mchandschuh:
Great sutff @AlexCatarino  -- if you find yourself using that tryconvert pattern again with a default value, consider making a utility method, maybe something of the form:

```csharp
public static ConvertToDelegate<T>(this PyObject pyObject, T defaultValue = null)
    where T : class
{
    if (pyObject == null)
    {
        return defaultValue;
    }

    T output;
    if (!pyObject.TryConvertToDelegate(out output)
    {
        return defaultValue;
    }

    return output;
}
```

Issue #1842, opened 2018-04-05
VS Plugin - Create tool window


Martin-Molinero:
#### Expected Behavior

Create a new tool window that allows a user to: 
- select working project
- see current/previous/on going backtests and their result status
- open and delete backtests
- link to backtest/compilation errors if any

#### Actual Behavior

N/A

#### Potential Solution

Implement tool window. 

#### Reproducing the Problem

N/A

#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #1876, opened 2018-04-17
VS Plugin - Adding new backtest ToolWindow

Martin-Molinero:
#### Description

This PR creates a new Visual Studio tool window that allows a user to:
- select working project
- see current/previous/on going backtests, for the working project, and their result status
- open and delete backtests

#### Related Issue

https://github.com/QuantConnect/Lean/issues/1842

#### Motivation and Context

Improve VS plugin value

Yes. I'm changing the VS plugin documentation.

Manually VS15 and VS17
#### Types of changes

#### Checklist:


VisualStudioPlugin/IBacktestObserver.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace QuantConnect.VisualStudioPlugin
+{
+    /// <summary>
+    /// Backtest observer interface. Used to notify backtest status executed from the IDE.
+    /// </summary>
+    internal interface IBacktestObserver
+    {
+        void NewBacktest(int projectId, Api.Backtest backtestStatus);
+        void UpdateBacktest(int projectId, Api.Backtest backtestStatus);
+        void BacktestFinished(int projectId, Api.Backtest backtestStatus);
```
mchandschuh: this may be me being nit picky :D

I love the 'eventy' sound of `BacktestFinished` -- it matches well with 'observer' -- what are your thoughts on renaming `NewBacktest` to something like `BacktestCreated` or `BacktestStarted`, and likewise for `UpdateBacktest` to `BacktestStatusUpdated`?

Another options is simply `Started`, `Updated`, and `Finished` since the 'backest' portion is already implied via the type name and the parameter of type `Api.Backtest`.

VisualStudioPlugin/SolutionExplorerMenuCommand.cs
```diff
@@ -53,6 +53,11 @@ internal sealed class SolutionExplorerMenuCommand
 
         private ProjectFinder _projectFinder;
 
+        /// <summary>
+        /// Observer for the status of backtests launched from the plugin
+        /// </summary>
+        private IBacktestObserver _backtestObserver;
```
mchandschuh: should this field be marked as read-only?

VisualStudioPlugin/readme.md
```diff
@@ -68,6 +70,19 @@ This feature allows the user to upload to QC one or more files to a target proje
...
 
+## Tool Window
+
+This utility allows the user to monitor ongoing backtests and perform different actions over previous ones.
+>Given the user selects project ‘BuyTheDip_007’ in the tool windows combo box and launches a backtest using [‘Send For Backtesting’](#SendForBacktesting) for ‘BuyTheDip_007’ project, it will display the backtests progress.
+>
+> ![alt tag](Resources/toolWindow.PNG)
+
+##### How To Open
+- From the Visual Studio IDE go to ‘View’ menu → ‘Other Windows’ → ‘QuantConnect’
```
mchandschuh: Thanks for keeping up with the documentation!

VisualStudioPlugin/IBacktestObserver.cs
Martin-Molinero: Your right! good suggestion, I'll use Backtest{event} format 👍

VisualStudioPlugin/SolutionExplorerMenuCommand.cs
Martin-Molinero: yes, good catch!

VisualStudioPlugin/SolutionExplorerMenuCommand.cs
```diff
@@ -262,20 +267,35 @@ private bool UploadFilesToServer(int selectedProjectId, IEnumerable<SelectedItem
...
+
+            var errorPresent = false;
+            while (backtestStatus.Progress < 1 && !errorPresent)
             {
-                Thread.Sleep(5000);
-                backtestStatus = api.ReadBacktest(projectId, backtestId);
+                backtestStatus = await System.Threading.Tasks.Task.Run(() =>
+                {
+                    Thread.Sleep(5000);
+                    return api.ReadBacktest(projectId, backtestId);
```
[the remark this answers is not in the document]
Martin-Molinero: your right. The suggested code works great, and using Delay seems to make the algorithm run faster!

Martin-Molinero:
I've implemented the changes requested by @mchandschuh . Also I've done the following changes:
 - Add missing try catch. Required for network or Api issues
 - I've added the option for AuthenticationCommand::Login() not to show login dialog which blocks the UI. This change solves the issue bellow. Now when created, the tool window will not show login dialog.
	> Given endpoint has no internet access, QC tool windows is open, close VS, open VS, login dialog opens and blocks solution load and for some VS bug all icons break too
 - Moved 'api.Connected' to a new task to avoid blocking UI
 - Some minor changes to documentation
 - Now when Login dialog is closed it will shown a message in the status bar.

VisualStudioPlugin/SolutionExplorerMenuCommand.cs
```diff
@@ -239,7 +258,7 @@ private bool UploadFilesToServer(int selectedProjectId, IEnumerable<SelectedItem
 
             while (compileStatus.State == Api.CompileState.InQueue)
             {
-                Thread.Sleep(2000);
+                System.Threading.Tasks.Task.Delay(2000);
```
mchandschuh: Since this task is not awaited, this method will keep on going. I'd be interested to see how many times we invoke `api.ReadCompile`  -- typically async-ness infects the entire call stack (or until someone does a `Task.Run` or similar).

Martin-Molinero: Your right! the method kept on going 😅, it queried a couple of times (in debug).
I've changed it to work with the same async logic as the backtest method does 👍 . With the change, in debug it's also queering twice.

mchandschuh: Very cool, thanks for updating. I think this is ready for merge -- we may want to browse the vs plugin project for any other usages of `Thread.Sleep` and see if w can unroll them into `async/await` like we've done here.  If it becomes common, an `async` version of the API methods might be useful and would remove all the `Task.Run` calls and make it *even faster* :)

Martin-Molinero: I've doubled checked and there are no other `Thread.Sleep` in the plugin

PR #1879, opened 2018-04-17
Multiple alpha models

### LINKED ISSUE #1862: Multiple Alpha Model support
#### Expected Behavior

Portfolio construction model needs a means of determining the source of each insight.

#### Actual Behavior

There's no means for the portfolio construction model determining the source alpha model of an insight.

#### Potential Solution

Add `string Insight.SourceModel` as a unique identifier -- reco alpha model type name + guid.
Also improve `InsightCollection` to support multiple alpha models.

#### Checklist

mchandschuh:
#### Description

Adds `Insight.SourceModel` to track where each insight object was generated. This can be set manually via constructor and/or `Insight.Price` helper, but if left null, the framework will properly resolve the correct model name. Also, a `CompositeAlphaModel` was created to support multiple alpha models running within the same framework algorithm. The composite alpha model will properly tag each generated insight with the source model's name.

#### Related Issue

Closes #1862 

Yes - @jaredbroad - we should document this new type and the `Insight.SourceModel` property. Also important to note, that users can override the 'source model' at anytime, thereby not *requiring* them to use the composite model if they don't want to, for example, multiple models inside a single alpha model implementation.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CompositeAlphaModelFrameworkAlgorithm.cs
+++ b/Algorithm.CSharp/CompositeAlphaModelFrameworkAlgorithm.cs
@@ -1,0 +1,58 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using QuantConnect.Algorithm.Framework;
+using QuantConnect.Algorithm.Framework.Alphas;
+using QuantConnect.Algorithm.Framework.Execution;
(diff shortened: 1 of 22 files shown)


Algorithm.Framework/Alphas/AlphaModelPythonWrapper.cs
```diff
@@ -24,10 +24,28 @@ namespace QuantConnect.Algorithm.Framework.Alphas
...
         private readonly dynamic _model;
 
+        public string Name
+        {
+            get
+            {
+                using (Py.GIL())
+                {
+                    // if the model defines a Name property then use that
+                    if (_model.HasAttr("Name"))
```
mchandschuh: @AlexCatarino -- what if the python model derives from a base class, does this check also check the base class? If not, perhaps I need to write a recursive `HasAttr` check -- thoughts?

Algorithm.CSharp/CompositeAlphaModelFrameworkAlgorithm.cs
```diff
@@ -0,0 +1,58 @@
...
+            SetStartDate(2013, 10, 07);
+            SetEndDate(2013, 10, 11);
+
+            // even though we're using a framework algorithm, we can still add our securities
+            // using the AddEquity/Forex/Crypto/ect methods and then pass them into a manual
+            // universe selection model using Securities.Keys
+            AddEquity("SPY");
+            AddEquity("IBM");
+            AddEquity("BAC");
+            AddEquity("AIG");
```
jaredbroad: 👍 love maintaining backwards compatibility!

mchandschuh: hehe, I knew you'd love a good example of making this easy :D

Issue #1696, opened 2018-03-14
Named arguments in python algorithms


AlexCatarino:
#### Expected Behavior

Use named arguments to specify an argument for a particular parameter by associating the argument with the parameter's name rather than with the parameter's position in the parameter list.

#### Actual Behavior

When we use named arguments, the desired values are not passed.

#### Potential Solution

Follow pythonnet's issue [265](https://github.com/pythonnet/pythonnet/issues/265)

#### Reproducing the Problem

`Log` statement should display `1`, but it `2` is shown.
```python
class BasicTemplateAlgorithm(QCAlgorithm):
    def Initialize(self):
        spy = self.AddEquity("SPY", leverage=1)
        self.Log(str(spy.Leverage))
```
#### System Information

N/A

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

Martin-Molinero:
Closed by https://github.com/QuantConnect/Lean/pull/5376

PR #1888, opened 2018-04-19
Multiple alpha models for python

### LINKED ISSUE #1887: Multiple Alpha Model support for python
#### Expected Behavior
Ability to use multiple alpha models written in python.

#### Actual Behavior
Only support C# alpha models.

#### Potential Solution
Add constructor overload to `CompositeAlphaModel` that accepts python objects.

#### Reproducing the Problem
Create a python version of `CompositeAlphaModelFrameworkAlgorithm` with python alpha models.

#### System Information
No.

#### Checklist

AlexCatarino:
#### Description
- Fixes `RsiAlphaModel.py` typo
  In the static method `Insight.Price`, `.Price` was missing.
- Refactors AlphaModelPythonWrapper.Name
  The cleanest way to get a python type name, when we know it is an instance of a class, is getting the value of the  `__class__.__name__` attribute.
- Adds constructor overload to `CompositeAlphaModel` that accepts python objects.
- Adds python version of `CompositeAlphaModelFrameworkAlgorithm`.
- Fixes python path order
  The first place python should look for scripts is not the current directory, but the directory defined at "algorithm-location".

#### Related Issue
Closes #1887 

#### Motivation and Context
Python and C# users should be able to use all the features.

See #1879.

Unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Alphas/AlphaModelPythonWrapper.cs
+++ b/Algorithm.Framework/Alphas/AlphaModelPythonWrapper.cs
@@ -41,7 +41,7 @@ namespace QuantConnect.Algorithm.Framework.Alphas
                     }
 
                     // if the model does not define a name property, use the python type name
-                    return _model.GetPythonType();
+                    return _model.__class__.__name__;
                 }
             }
         }
--- a/Algorithm.Framework/Alphas/CompositeAlphaModel.cs
+++ b/Algorithm.Framework/Alphas/CompositeAlphaModel.cs
@@ -15,6 +15,7 @@
 
 using System;
 using System.Collections.Generic;
+using Python.Runtime;
 using QuantConnect.Data;
 using QuantConnect.Data.UniverseSelection;
 using QuantConnect.Util;
@@ -41,6 +42,24 @@ namespace QuantConnect.Algorithm.Framework.Alphas
             }
 
             _alphaModels = alphaModels;
+        }
+
+        public CompositeAlphaModel(PyObject[] alphaModels)
+        {
+            if (alphaModels.IsNullOrEmpty())
+            {
+                throw new ArgumentException("Must specify at least 1 alpha model for the CompositeAlphaModel");
+            }
+
+            _alphaModels = new IAlphaModel[alphaModels.Length];
+
+            for (var i = 0; i < alphaModels.Length; i++)
+            {
+                if (!alphaModels[i].TryConvert(out _alphaModels[i]))
+                {
+                    _alphaModels[i] = new AlphaModelPythonWrapper(alphaModels[i]);
+                }
+            }
         }
 
         /// <summary>
--- a/Algorithm.Framework/Alphas/RsiAlphaModel.py
+++ b/Algorithm.Framework/Alphas/RsiAlphaModel.py
@@ -53,9 +53,9 @@ class RsiAlphaModel:
 
             if state != previous_state and rsi.IsReady:
                 if state == State.TrippedLow:
-                    insights.append(Insight(symbol, self.insightPeriod, InsightDirection.Up))
+                    insights.append(Insight.Price(symbol, self.insightPeriod, InsightDirection.Up))
                 if state == State.TrippedHigh:
-                    insights.append(Insight(symbol, self.insightPeriod, InsightDirection.Down))
+                    insights.append(Insight.Price(symbol, self.insightPeriod, InsightDirection.Down))
 
             symbolData.State = state
 
--- a/AlgorithmFactory/Loader.cs
+++ b/AlgorithmFactory/Loader.cs
@@ -158,8 +158,8 @@ namespace QuantConnect.AlgorithmFactory
             // Set the python path for loading python algorithms.
             var pythonPath = new[]
             {
-                new DirectoryInfo(Environment.CurrentDirectory).FullName,
                 pythonFile.Directory.FullName,
+                new DirectoryInfo(Environment.CurrentDirectory).FullName,
                 Environment.GetEnvironmentVariable("PYTHONPATH")
             };
 
--- a/Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
+++ b/Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
@@ -151,7 +151,10 @@ namespace QuantConnect.Tests.Algorithm.Framework.Alphas
             }
 
             var actual = model.GetModelName();
-            var expected = GetExpectedModelName(model);
+            var expected = language == Language.CSharp
+                ? GetExpectedModelName(model)
+                : (model as AlphaModelPythonWrapper).Name;
+
             Assert.AreEqual(expected, actual);
         }
 
--- a/Tests/RegressionTests.cs
+++ b/Tests/RegressionTests.cs
@@ -969,6 +969,7 @@ namespace QuantConnect.Tests
                 new AlgorithmStatisticsTestParameters("BasicTemplateFrameworkAlgorithm", basicTemplateFrameworkStatistics, Language.Python),
                 new AlgorithmStatisticsTestParameters("BasicTemplateOptionsAlgorithm", basicTemplateOptionsStatistics, Language.Python),
                 new AlgorithmStatisticsTestParameters("CustomDataRegressionAlgorithm", customDataRegressionStatistics, Language.Python),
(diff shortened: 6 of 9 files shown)


Algorithm.Framework/Alphas/AlphaModelPythonWrapper.cs
```diff
@@ -41,7 +41,7 @@ public string Name
                     }
 
                     // if the model does not define a name property, use the python type name
-                    return _model.GetPythonType();
+                    return _model.__class__.__name__;
```
mchandschuh: Awesome! Thank you for fixing, I wasn't convinced it was correct before.

Algorithm.Framework/Alphas/CompositeAlphaModel.cs
```diff
@@ -43,6 +44,24 @@ public CompositeAlphaModel(params IAlphaModel[] alphaModels)
             _alphaModels = alphaModels;
         }
 
+        public CompositeAlphaModel(PyObject[] alphaModels)
```
mchandschuh: can python make use of `params` keywords? If not it's OK since *not* having `param` likely reduces the possibility for method ambiguity.

Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
```diff
@@ -151,7 +151,10 @@ public void ModelNameTest(Language language)
             }
 
             var actual = model.GetModelName();
-            var expected = GetExpectedModelName(model);
+            var expected = language == Language.CSharp
+                ? GetExpectedModelName(model)
+                : (model as AlphaModelPythonWrapper).Name;
```
mchandschuh: for python, won't this _always_ be true? maybe the `GetExpectedModelName` implementation should check `if (model is PyObject)` or something similar?

Tests/RegressionTests.cs
```diff
@@ -969,6 +969,7 @@ private static TestCaseData[] GetRegressionTestParameters()
                 new AlgorithmStatisticsTestParameters("BasicTemplateFrameworkAlgorithm", basicTemplateFrameworkStatistics, Language.Python),
                 new AlgorithmStatisticsTestParameters("BasicTemplateOptionsAlgorithm", basicTemplateOptionsStatistics, Language.Python),
                 new AlgorithmStatisticsTestParameters("CustomDataRegressionAlgorithm", customDataRegressionStatistics, Language.Python),
+                new AlgorithmStatisticsTestParameters("CompositeAlphaModelFrameworkAlgorithm", compositeAlphaModelFrameworkAlgorithmStatistics, Language.Python),
```
mchandschuh: Great stuff, thank you!

Algorithm.Framework/Alphas/CompositeAlphaModel.cs
```diff
@@ -43,6 +44,24 @@ public CompositeAlphaModel(params IAlphaModel[] alphaModels)
...
+                throw new ArgumentException("Must specify at least 1 alpha model for the CompositeAlphaModel");
+            }
+
+            _alphaModels = new IAlphaModel[alphaModels.Length];
+
+            for (var i = 0; i < alphaModels.Length; i++)
+            {
+                if (!alphaModels[i].TryConvert(out _alphaModels[i]))
+                {
+                    _alphaModels[i] = new AlphaModelPythonWrapper(alphaModels[i]);
```
mchandschuh: I love how simple this has been boiled down to, nice code!

Algorithm.Framework/Alphas/RsiAlphaModel.py
```diff
@@ -53,9 +53,9 @@ def Update(self, algorithm, data):
 
             if state != previous_state and rsi.IsReady:
                 if state == State.TrippedLow:
-                    insights.append(Insight(symbol, self.insightPeriod, InsightDirection.Up))
+                    insights.append(Insight.Price(symbol, self.insightPeriod, InsightDirection.Up))
                 if state == State.TrippedHigh:
-                    insights.append(Insight(symbol, self.insightPeriod, InsightDirection.Down))
+                    insights.append(Insight.Price(symbol, self.insightPeriod, InsightDirection.Down))
```
mchandschuh: nice catch!

AlgorithmFactory/Loader.cs
```diff
@@ -158,8 +158,8 @@ private bool TryCreatePythonAlgorithm(string assemblyPath, out IAlgorithm algori
             // Set the python path for loading python algorithms.
             var pythonPath = new[]
             {
-                new DirectoryInfo(Environment.CurrentDirectory).FullName,
                 pythonFile.Directory.FullName,
+                new DirectoryInfo(Environment.CurrentDirectory).FullName,
```
mchandschuh: Just out of curiosity, why change the ordering here?

I know *semantically* why, so that the python directory gets preference over the current directory, but why was the reason behind this? Were we incorrectly resolving a type or some other bug raised/observed?

Tests/Algorithm/Framework/Alphas/CommonAlphaModelTests.cs
mchandschuh: Please update the python alpha models to implement the `Name` property so that it equals the c# version Name properties. While I don't expect many users to define names for their models, we still need the examples in python for how it's done, I'm hoping it's simple, if not, we may need to do a refactoring to make it easier for python users.

AlgorithmFactory/Loader.cs
AlexCatarino: Locally, VS copies the python scripts to `bin/Debug` (which is `Environment.CurrentDirectory`). When we modify the code at `Algorithm.Python` and build/rebuild, the modified code was not being copied over (I had to manually delete the old file). I tried to modify the properties to "Copy Always" or "Copy if newer", but it just didn't replace the file.

Algorithm.Framework/Alphas/AlphaModelPythonWrapper.cs
AlexCatarino: Me neither! We could have `_model.GetPythonType().ToString()` instead, but it returns something like `<class 'RsiAlphaModel.RsiAlphaModel'>` and we would have to manipulate this string. Then I look for an alternative and that solution pop out.

Algorithm.Framework/Alphas/ConstantAlphaModel.py
```diff
@@ -77,4 +91,10 @@ def ShouldEmitInsight(self, utcTime, symbol):
...
         self.insightsTimeBySymbol[symbol] = utcTime
-        return True
\ No newline at end of file
+        return True
+
+def strfdelta(tdelta):
+    d = tdelta.days
+    h, rem = divmod(tdelta.seconds, 3600)
+    m, s = divmod(rem, 60)
+    return "{}.{:02d}:{:02d}:{:02d}".format(d,h,m,s)
```
mchandschuh: awesome!

Common/Extensions.cs
```diff
@@ -1114,6 +1114,28 @@ public static bool TryConvertToDelegate<T>(this PyObject pyObject, out T result)
...
+        /// </summary>
+        /// <param name="value">Numeric value</param>
+        /// <param name="pyObject">Python object that encapsulated a Enum Type</param>
+        /// <returns>String that represents the enumerated object</returns>
+        public static string GetEnumString(this int value, PyObject pyObject)
+        {
+            Type type;
+            if (pyObject.TryConvert(out type))
+            {
+                return value.ToString().ConvertTo(type).ToString();
```
mchandschuh: Cool! Didn't know `ConvertTo(enumType)` worked like that, awesome!

AlgorithmFactory/Loader.cs
disruptek: Does this not conflict with [https://docs.python.org/3/tutorial/modules.html#the-module-search-path](url)?

Algorithm.Framework/Alphas/CompositeAlphaModel.cs
disruptek: Python can use keyword parameters.  When the keyword is omitted but a value is passed nonetheless, the keyword inherits the positional parameter.  It's bad style to not expressly use named keyword parameters when they exist, because obviously, if the position of the keywords changes with a change to the method definition, then the invoker will no longer function as intended...

AlexCatarino: @disruptek , `CompositeAlphaModel` is not a python object and we (pythonnet) have an issue on keyword/names parameters:
https://github.com/QuantConnect/Lean/issues/1696

PR #1923, opened 2018-04-26
Convert to exchange tz and reverse a day when reading fine data

### LINKED ISSUE #1905: Investigate FineFundamental EndTime inconsistencies w.r.t Coarse EndTime
#### Expected Behavior

`CoarseFundamental` and `FineFundamental` data used in the same time step should have identical `EndTime` values. The `EndTime` is the time at which the data should be emitted.

#### Actual Behavior

The `FineFundamental.EndTime` property appears to be phrased in UTC and shifted forward a day. This could be a simple bug in the `FineFundamental` data type, but could also point to a bug in selecting the correct file to open.

#### Potential Solution

I did some initial investigation and was able to confirm that the coarse data closing prices match the daily data closing prices, which means that the coarse data is emitting correctly. We'll need to verify that we're picking the correct fine file to be read and that we're assigned the correct `EndTime` to the `FineFundamental` instance. Also, fine data is in the EST/EDT (NewYork) time zone and not UTC. It looks like currently the data is being time stamped in UTC.

Interestingly enough, this time stamps aren't checked by the engine because of how we handle fine in a special manner. We process coarse normally, and then if a fine universe is being used, then we manually fetch the fine data for a specific date, but don't synchronize it, so no decisions are ever made based on the `EndTime` in question.

#### Reproducing the Problem

Run the `CoarseFineFundamentalRegressionAlgorithm` with the following edits:
``` csharp
// in CoarseSelectionFunction
var aapl = coarse.First(s => s.Symbol.Value == "AAPL");
Log($" COARSE END TIME:: {aapl.Symbol.Value}:: {aapl.EndTime:o}:: {aapl.Value}");
```
```csharp
// in FineSelectionFunction
var aapl = fine.First(s => s.Symbol.Value == "AAPL");
Log($"   FINE END TIME:: {aapl.Symbol.Value}:: {aapl.EndTime:o}:: {aapl.Value}");
```
``` csharp
// in OnData
var aapl = data.FirstOrDefault(kvp => kvp.Key.Value == "AAPL").Value;
if (aapl != null)
{
    Log($"ON DATA END TIME:: {aapl.Symbol.Value}:: {aapl.EndTime:o}:: {aapl.Value}");
}
```

The above will produce logs containing the following:
```
 Log:  COARSE END TIME:: AAPL:: 2014-04-02T00:00:00.0000000:: 541.74
 Log:    FINE END TIME:: AAPL:: 2014-04-03T04:00:00.0000000Z:: 0
 Log: ON DATA END TIME:: AAPL:: 2014-04-02T00:00:00.0000000:: 541.7400
```
Note how the fine time is in UTC (`Z`) and is shifted forward a full day. This is because the `FineFundamental` type defines itself as having a period of one day, whereas the coarse data defines itself as having an instantaneous period. The above was produced with daily resolution data.

#### System Information

All systems, discovered running LEAN locally w/ regression test suite.

#### Checklist

mchandschuh:
#### Description

The fine enumerator factor accepts a 'date'. This date is that day that the data is produced, so the emit time on that data is midnight the following day. When we go to read the fine data we can't use the
current time because that's the 'emit time' -- instead we need to back the time up a full day to find the correct start time. In addition, the fine data was being read using UTC time stamps which led to even
more confusion here -- this change resolves both issues.

In addition to the above, the security.Fundamentals property wasn't being properly set for the first time step when a security is selected due to a bug in the subscription synchronizer not clearing out the `universeData` in preparation for another loop (in the event changes != None).

Initially I was under the impression that coarse's price data should come after the security price data, but that isn't the case. Consider universe selection running at midnight on the 25th -- it's pulling data from 24->25 (date=24). When we add a security, say `AAPL`, we'll pull data from 24->25 as well and pipe it into the algorithm, then when submitting an order you'll get an execution at market open on the 25th -> no look-ahead bias confirmed.

#### Motivation and Context

Fixes #1905 

No.

Tested locally and confirmed changes with `CoarseFineFundamentalRegressionAlgorithm`.

Turns out the coarse data wasn't at fault -- the main issue here was that when reading the fine data we didn't convert from UTC->NewYork *and also* we didn't back up a day, since the 'date' parameter for the enumerator factor reads on the interval [date, date+1day] -- introducing look-ahead bias. This has been updated to properly convert to exchange time and reverse a full day so we read data for the corrct date. Manually inspected timestamps on data and ensured they're good.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/UniverseSelection.cs
+++ b/Engine/DataFeeds/UniverseSelection.cs
@@ -83,7 +83,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
                     foreach (var symbol in selectSymbolsResult)
                     {
-                        var factory = new FineFundamentalSubscriptionEnumeratorFactory(_algorithm.LiveMode, x => new[] { dateTimeUtc });
                         var config = FineFundamentalUniverse.CreateConfiguration(symbol);
 
                         var exchangeHours = _marketHoursDatabase.GetEntry(symbol.ID.Market, symbol, symbol.ID.SecurityType).ExchangeHours;
@@ -92,7 +91,9 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
                         var security = new Equity(symbol, exchangeHours, quoteCash, symbolProperties);
 
-                        var request = new SubscriptionRequest(true, universe, security, new SubscriptionDataConfig(config), dateTimeUtc, dateTimeUtc);
+                        var localStartTime = dateTimeUtc.ConvertFromUtc(exchangeHours.TimeZone).AddDays(-1);
+                        var factory = new FineFundamentalSubscriptionEnumeratorFactory(_algorithm.LiveMode, x => new[] { localStartTime });
+                        var request = new SubscriptionRequest(true, universe, security, new SubscriptionDataConfig(config), localStartTime, localStartTime);
                         using (var enumerator = factory.CreateEnumerator(request, dataProvider))
                         {
                             if (enumerator.MoveNext())
--- a/Algorithm.CSharp/CoarseFineFundamentalRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/CoarseFineFundamentalRegressionAlgorithm.cs
@@ -58,17 +58,17 @@ namespace QuantConnect.Algorithm.CSharp
             {
                 return new List<Symbol>
                 {
-                    QuantConnect.Symbol.Create("AAPL", SecurityType.Equity, "usa"),
-                    QuantConnect.Symbol.Create("AIG", SecurityType.Equity, "usa"),
-                    QuantConnect.Symbol.Create("IBM", SecurityType.Equity, "usa")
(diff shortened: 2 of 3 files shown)


Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -92,7 +91,9 @@ public SecurityChanges ApplyUniverseSelection(Universe universe, DateTime dateTi
 
                         var security = new Equity(symbol, exchangeHours, quoteCash, symbolProperties);
 
-                        var request = new SubscriptionRequest(true, universe, security, new SubscriptionDataConfig(config), dateTimeUtc, dateTimeUtc);
+                        var localStartTime = dateTimeUtc.ConvertFromUtc(exchangeHours.TimeZone).AddDays(-1);
+                        var factory = new FineFundamentalSubscriptionEnumeratorFactory(_algorithm.LiveMode, x => new[] { localStartTime });
+                        var request = new SubscriptionRequest(true, universe, security, new SubscriptionDataConfig(config), localStartTime, localStartTime);
```
StefanoRaggi: Good catch!

PR #1926, opened 2018-04-27
Lean Visualizer C# toolbox project and Python Wrapper working

### LINKED ISSUE #1835: Lean source data visualizer
#### Expected Behavior

It would comes handy having a tool for quickly visualize Lean source data, specially for data completion and quality checking.

#### Actual Behavior

There is not a tool in the toolbox to simply parse the zipped data show the results.  

#### Checklist

Jay-Jay-D:
#### Description

This PR adds a Visualizer to the ToolBox, the main functions is reading Leans zipped data files for a quick review of data quality and completion.

The new folder contains a new C# class and is in turn a Python Module. The C# side read the data and transform it into a `pandas.DataFrame`, the Python parts takes that data, plots it and saves. Also the Python module contains the CLI.

#### Related Issue

Closes #1835 

#### Motivation and Context

There is no a fast way to parse the zipped data show the results. This kind of tools speeds up the data cleaning and maintenance.

It will not required modification to the existent documentation, on the other hand, the code is documented.

All the new features in the `LeanDataReader` and the `PandasConverter` are covered by new tests.
All existing test passed. 

#### Types of changes

#### Checklist:


ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,182 @@
...
+    A copy of the config.json file will be written in the QuantConnect.ToolBox assembly folder for use of the Lean
+    Composer. The CRITICAL assumption is that if there is config.json in the QuantConnect.ToolBox assembly folder, it
+    has the correct paths, if not, there is a --reset option for force overwriting the file.
+
+    Is highly recommended to use absolute paths in the config.json.
+    """
+    def __init__(self, arguments):
+
+        self.palette = ['#f5ae29', '#657584', '#b1b9c3', '#222222']
+        self.data_file_argument = arguments['DATAFILE']
```
AlexCatarino: You don't check whether `self.data_file_argument` file exists.
You could do this check here and stop the script right away.

ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,182 @@
...
+Arguments:
+    DATAFILE   Absolute or relative path to a zipped data file to plot.
+               Optionally the zip entry file can be declared by using '#' as separator.
+
+Options:
+    -h --help                 show this.
+    -o --output file_path     path or filename for the output plot. If not declared, it will save with an
+                              auto-generated name at the default folder defined in the config.json file.
+    -s, --size height,width   plot size in pixels [default: 800,400].
+    -r, --reset               Resets it forces the writing of the config.json file in the assembly folder.
```
AlexCatarino: I don't believe this is needed. Since we set the `config.json`, if should be always good. If not, we throw an exception and edit it.

ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,182 @@
...
+                           f"Absolute path to 'data-folder' provided: {data_folder_info.resolve().absolute()}\n" +
+                           'If the issue continues even then the folder is correctly declared, please use the --reset option.')
+        assembly_folder_path = str(assemblies_folder_info.resolve().absolute())
+        # Check if config.json exist in the composer-dll-directory, if not creates it.
+        config_file = assemblies_folder_info.joinpath('config.json')
+        if not config_file.exists() or self.reset:
+            cfg_content = {'composer-dll-directory': assembly_folder_path,
+                           'data-folder': str(data_folder_info.resolve().absolute())}
+            with open(str(config_file.resolve().absolute()), 'w') as cfg:
+                json.dump(cfg_content, cfg)
```
AlexCatarino: Same first comment: the script shouldn't be rewritten.

ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,182 @@
...
+    Is highly recommended to use absolute paths in the config.json.
+    """
+    def __init__(self, arguments):
+
+        self.palette = ['#f5ae29', '#657584', '#b1b9c3', '#222222']
+        self.data_file_argument = arguments['DATAFILE']
+        self.reset = arguments['--reset']
+
+        with open('config.json', 'r') as json_data:
+            self.config = json.load(json_data)
```
AlexCatarino: Just thinking out loud...
Since arguments and config are dictionaries, why do we not merge them and add keys as we go along:
```python
self.config['DATAFILE'] = arguments['DATAFILE'] # don't create self.data_file_argument
self.config['--output'] = arguments['--output'] # or self.generate_plot_filename()
self.config['--size'] = [int(p) for p in arguments['--size'].split(',')]
```

ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        Applies the filters defined in the CLI arguments to the parsed data.
+        Not fully implemented yet, it only select the close columns right now.
+        :param df: pandas.DataFrame with all the data form the selected file.
+        :return: a filtered pandas.DataFrame.
+
+        TODO: implement column and time filters.
+        """
+        symbol = df.index.levels[0][0]
+        df = df.loc[symbol]
```
AlexCatarino: If you make this operation in `get_data`, you wouldn't need to call it again.

ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,182 @@
...
+        if not config_file.exists() or self.reset:
+            cfg_content = {'composer-dll-directory': assembly_folder_path,
+                           'data-folder': str(data_folder_info.resolve().absolute())}
+            with open(str(config_file.resolve().absolute()), 'w') as cfg:
+                json.dump(cfg_content, cfg)
+
+        # Load the visualizer
+        os.chdir(assembly_folder_path)
+        sys.path.append(assembly_folder_path)
+        AddReference("QuantConnect.ToolBox")
```
AlexCatarino: You don't need to `chdir` and add the assembly folder to path, in `AddReference` you can put `QuantConnect.ToolBox.exe` full file name.

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -13,6 +13,8 @@
     <FileAlignment>512</FileAlignment>
     <LangVersion>6</LangVersion>
     <TargetFrameworkProfile />
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
```
mchandschuh: I think this can be removed from the project file.

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -242,6 +248,15 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
     <None Include="packages.config" />
+    <Content Include="Visualizer\config.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
mchandschuh: We typically set this to 'copy if newer' so it doesn't overwrite if manually editted in bin/debug

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -289,6 +304,13 @@
   </ItemGroup>
   <ItemGroup />
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <Import Project="..\packages\QuantConnect.pythonnet.1.0.5.7\build\QuantConnect.pythonnet.targets" Condition="Exists('..\packages\QuantConnect.pythonnet.1.0.5.7\build\QuantConnect.pythonnet.targets')" />
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
+    <PropertyGroup>
+      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
+    </PropertyGroup>
```
mchandschuh: nuget warning message in the csproj file?

ToolBox/Visualizer/Visualizer.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class Visualizer : LeanDataReader
+    {
+        private readonly PandasConverter _converter = new PandasConverter();
+
+        /// <inheritdoc />
+        public Visualizer(string filepath)
+            : base(filepath)
+        { }
+
+        public PyObject ParseDataFrame() => _converter.GetDataFrame(Parse());
```
mchandschuh: it seems like this entire class exists just to call `new PandasConverted().GetDataFrame( <output from leandatareader.parse )` -- is this class needed -- or do we just need a function to create a data frame from an ienumerable of base data?

Also, it doesn't seem this class does any 'visualizing' -- so perhaps the name is just throwing me off

ToolBox/Visualizer/config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "composer-dll-directory": "../.",
+  "data-folder": "../../../../Data",
+  "default-output-folder": "../output_folder"
```
mchandschuh: Moar config files!

AlexCatarino: We can remove this and pass the values as arguments to the script.

ToolBox/Visualizer/Visualizer.cs
Jay-Jay-D: Yes it has some smell, it can be just a method in the `LeanDataReader` (Lazy class / freeloader).

ToolBox/QuantConnect.ToolBox.csproj
Jay-Jay-D: Those lines are present in all Lean `csproj` files where `pythonnet` is needed as dependency.
[Launcher](https://github.com/QuantConnect/Lean/blob/418b25eddc46813fc2855f49b06adac8dc54d148/Launcher/QuantConnect.Lean.Launcher.csproj#L143), [Algorithm](https://github.com/QuantConnect/Lean/blob/418b25eddc46813fc2855f49b06adac8dc54d148/Algorithm/QuantConnect.Algorithm.csproj#L130), [Common](https://github.com/QuantConnect/Lean/blob/418b25eddc46813fc2855f49b06adac8dc54d148/Common/QuantConnect.csproj#L613), etc.

ToolBox/Visualizer/QuantConnect.Visualizer.py
```diff
@@ -0,0 +1,156 @@
...
+
+    This class is instantiated with the dictionary docopt generates from the CLI arguments.
+
+    It contains the methods for set up and load the C# assemblies into Python. The QuantConnect.ToolBox assembly folder
+    can be declared in the module's CLI.
+    """
+    def __init__(self, arguments):
+        self.arguments = arguments
+        if not Path(self.arguments['DATAFILE'].split('#')[0]).exists():
+            raise NotImplementedError("The zipped data file doesn't exist.")
```
AlexCatarino: Would it not not be `raise FileNotFoundError(...)`?
```python
file = Path(self.arguments['DATAFILE'].split('#')[0])
if not file.existis():
    raise FileNotFoundError(f"File {file.resolve().absolute()} does not exist")
```

Tests/ToolBox/LeanDataReaderTests.cs
```diff
@@ -221,5 +222,162 @@ private void ConvertMinuteFuturesData(Symbol canonical, TickType tickType, Resol
...
+            // Assert
+            Assert.True(symbol.Equals(data.First().Symbol));
+            Assert.AreEqual(rowsInfile, data.Length);
+            Assert.AreEqual(sumValue, data.Sum(c => c.Value));
+        }
+
+
+        public static object[] OptionAndFuturesCases =
+        {
+            new object[]
```
mchandschuh: no change required -- it's nice to use the `TestCaseData` class because then you can use `SetName` to give the test case a human readable name in the test window

ToolBox/LeanDataReader.cs
```diff
@@ -48,6 +52,66 @@ public LeanDataReader(SubscriptionDataConfig config, Symbol symbol, Resolution r
...
+        /// </example>
+        public LeanDataReader(string filepath)
+        {
+            Symbol symbol;
+            DateTime date;
+            Resolution resolution;
+            string zipEntry = null;
+
+            var localDiskMapFileProvider = new LocalDiskMapFileProvider();
+            Composer.Instance.AddPart<IMapFileProvider>(localDiskMapFileProvider);
```
mchandschuh: I must have missed this in the previous review, but typically a constructor shouldn't be adding parts to the global composer instance. For example, if this constructor gets called twice we're adding two instances? It's best that this type of composer management happen at the application level and not at the component/ctor level.

Consider setting directly from python at the top of the script -- if required, can add static helper c# function to make it easier, but these two lines should be invokable via python directly.

ToolBox/LeanDataReader.cs
```diff
@@ -48,6 +52,66 @@ public LeanDataReader(SubscriptionDataConfig config, Symbol symbol, Resolution r
...
+
+            var marketHoursDataBase = MarketHoursDatabase.FromDataFolder();
+            var dataTimeZone = marketHoursDataBase.GetDataTimeZone(symbol.ID.Market, symbol, symbol.SecurityType);
+            var exchangeTimeZone = marketHoursDataBase.GetExchangeHours(symbol.ID.Market, symbol, symbol.SecurityType).TimeZone;
+
+            var tickType = LeanData.GetCommonTickType(symbol.SecurityType);
+            var fileName = Path.GetFileNameWithoutExtension(fileInfo.Name);
+            if (fileName.Contains("_"))
+            {
+                tickType = (TickType)Enum.Parse(typeof(TickType), fileName.Split('_')[1], true);
```
mchandschuh: should tick type resolution be added to `TryParsePath` ? -- we don't need to make this change now, but it's food for thought. Reasonably the tick type is part of the path, so we could add another `out` parameter w/ the tick type, defaulting to the common tick type if not present in the file name.

ToolBox/QuantConnect.ToolBox.csproj
mchandschuh: interesting!

Jay-Jay-D:
Thank you very much @AlexCatarino and @mchandschuh for all the help, the great suggestions and the reviews!

Issue #1093, opened 2017-08-27
VWAP Order - Interactive Brokers


jaredbroad:
New order type for IB - VWAP.

jaredbroad:
All or None // Fill or Kill Option on Orders

patrickstar1:
Also would be nice to have Immediate or Cancel (IOC):

https://www.interactivebrokers.com/en/index.php?f=594

jaredbroad:
Adaptive Algo
https://interactivebrokers.github.io/tws-api/ibalgos.html#gsc.tab=0

patrickstar1:
Any update on these features? can we at least have all-or-none for now? Thanks!

jaredbroad:
Not yet sorry but we'll reply here when it is done so no need to ping thread. 

Its an internal rule of QC -- bugs before features :) As long as there are red-flags we'll fix them first. If you'd like to push these features through we'd welcome a PR though!

patrickstar1:
Bugs before Features is a good policy. I really need this feature because my algo goes crazy because of the partial fills - so much that I pulled the plug on it today.

I will see if I can contribute. Otherwise looks like I have to wait for you guys to pick it up.

StefanoRaggi:
Extra comments/notes here: https://github.com/QuantConnect/Lean/pull/1203

StefanoRaggi:
There are multiple features grouped together in this issue, so we are splitting this issue in the following tasks, to keep pull requests smaller and easier to review:

- [ ] Rename enum `OrderDuration` -> `TimeInForce` and rename `Order.Duration` to `Order.TimeInForce`

- [ ] Implement `TimeInForce.Day` (DAY)
https://www.interactivebrokers.com/en/software/tws/usersguidebook/ordertypes/time_in_force_for_orders.htm

- [ ] Implement `TimeInForce.GoodTilDate` (GTD)
https://www.interactivebrokers.com/en/index.php?f=589

- [ ] Implement `TimeInForce.FillOrKill` (FOK)
https://www.interactivebrokers.com/en/index.php?f=586

- [ ] Implement `TimeInForce.ImmediateOrCancel` (IOC)
https://www.interactivebrokers.com/en/index.php?f=594

- [ ] Implement `Order.AllOrNone` (AON)
https://www.interactivebrokers.com/en/index.php?f=577

- [ ] Implement IB Algorithms with `InteractiveBrokersOrderProperties`
[...]

PR #1938, opened 2018-04-30
Add TimeInForce.Day support in backtesting and IB brokerage

StefanoRaggi:
#### Description
This PR includes some common changes to support `TimeInForce` types:
- New `ITimeInForceHandler` interface and `GoodTilCancelled` and `Day` implementations
- Base `OrderProperties` class, instantiated in `QCAlgorithm` constructor

Both unit tests and regression algorithms have been added.

This feature is currently supported only in backtesting and with Interactive Brokers brokerage.

#### Related Issue
#1093 

#### Motivation and Context
Previously we only supported `GTC` time in force.

No.

New unit tests, new regression test and IB live testing.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/TimeInForceAlgorithm.cs
+++ b/Algorithm.CSharp/TimeInForceAlgorithm.cs
@@ -55,16 +55,17 @@ namespace QuantConnect.Algorithm.CSharp
         {
             if (_gtcOrderTicket == null)
             {
-                // This order has a default time in force of GoodTilCancelled,
-                // it will never expire and will not be cancelled automatically.
+                // This order has a default time in force of GoodTilCanceled,
+                // it will never expire and will not be canceled automatically.
 
+                DefaultOrderProperties.TimeInForce = TimeInForce.GoodTilCanceled;
                 _gtcOrderTicket = LimitOrder(_symbol, 10, 160m);
             }
 
             if (_dayOrderTicket == null)
             {
                 // This order will expire at market close,
-                // if not filled by then it will be cancelled automatically.
+                // if not filled by then it will be canceled automatically.
 
                 DefaultOrderProperties.TimeInForce = TimeInForce.Day;
                 _dayOrderTicket = LimitOrder(_symbol, 10, 160m);
--- a/Algorithm.CSharp/TimeInForceAlgorithm.cs
+++ b/Algorithm.CSharp/TimeInForceAlgorithm.cs
@@ -1,0 +1,85 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using QuantConnect.Data;
+using QuantConnect.Orders;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Demonstration algorithm of time in force order settings.
+    /// </summary>
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="using quantconnect" />
+    /// <meta name="tag" content="trading and orders" />
+    public class TimeInForceAlgorithm : QCAlgorithm
+    {
+        private Symbol _symbol;
+        private OrderTicket _gtcOrderTicket;
+        private OrderTicket _dayOrderTicket;
+
+        /// <summary>
+        /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
+        /// </summary>
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 07);
+            SetEndDate(2013, 10, 11);
+            SetCash(100000);
+
+            // The default time in force setting for all orders is GoodTilCancelled (GTC),
+            // uncomment this line to set a different time in force.
+            // We currently only support GTC and DAY.
+            // DefaultOrderProperties.TimeInForce = TimeInForce.Day;
+
+            _symbol = AddEquity("SPY", Resolution.Minute).Symbol;
+        }
+
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
+        {
+            if (_gtcOrderTicket == null)
+            {
+                // This order has a default time in force of GoodTilCancelled,
+                // it will never expire and will not be cancelled automatically.
+
+                _gtcOrderTicket = LimitOrder(_symbol, 10, 160m);
+            }
+
+            if (_dayOrderTicket == null)
+            {
+                // This order will expire at market close,
+                // if not filled by then it will be cancelled automatically.
+
+                DefaultOrderProperties.TimeInForce = TimeInForce.Day;
+                _dayOrderTicket = LimitOrder(_symbol, 10, 160m);
+            }
+        }
+
+        /// <summary>
+        /// Order event handler. This handler will be called for all order events, including submissions, fills, cancellations.
+        /// </summary>
+        /// <param name="orderEvent">Order event instance containing details of the event</param>
+        /// <remarks>This method can be called asynchronously, ensure you use proper locks on thread-unsafe objects</remarks>
+        public override void OnOrderEvent(OrderEvent orderEvent)
+        {
+            Debug($"{Time} {orderEvent}");
+        }
+
+    }
+}
--- a/Algorithm.Python/TimeInForceAlgorithm.py
+++ b/Algorithm.Python/TimeInForceAlgorithm.py
@@ -1,0 +1,70 @@
+﻿# QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+# Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from clr import AddReference
+AddReference("System")
+AddReference("QuantConnect.Algorithm")
+AddReference("QuantConnect.Common")
+
+from System import *
+from QuantConnect import *
+from QuantConnect.Algorithm import *
+from QuantConnect.Orders import *
+
+### <summary>
+### Demonstration algorithm of time in force order settings.
+### </summary>
+### <meta name="tag" content="using data" />
+### <meta name="tag" content="using quantconnect" />
+### <meta name="tag" content="trading and orders" />
+class TimeInForceAlgorithm(QCAlgorithm):
+
+    # Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
+    def Initialize(self):
+
+        self.SetStartDate(2013,10,7)
+        self.SetEndDate(2013,10,11)
+        self.SetCash(100000)
+
+        # The default time in force setting for all orders is GoodTilCancelled (GTC),
+        # uncomment this line to set a different time in force.
+        # We currently only support GTC and DAY.
+        # self.DefaultOrderProperties.TimeInForce = TimeInForce.Day;
+
+        self.symbol = self.AddEquity("SPY", Resolution.Second).Symbol
+
+        self.gtcOrderTicket = None
+        self.dayOrderTicket = None
+
+    # OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+    # Arguments:
+    #    data: Slice object keyed by symbol containing the stock data
+    def OnData(self, data):
+
+        if self.gtcOrderTicket is None:
+            # This order has a default time in force of GoodTilCancelled,
+            # it will never expire and will not be cancelled automatically.
+
+            self.gtcOrderTicket = self.LimitOrder(self.symbol, 10, 160)
+
+        if self.dayOrderTicket is None:
+            # This order will expire at market close,
+            # if not filled by then it will be cancelled automatically.
+
+            self.DefaultOrderProperties.TimeInForce = TimeInForce.Day
+            self.dayOrderTicket = self.LimitOrder(self.symbol, 10, 160)
+
+    # Order event handler. This handler will be called for all order events, including submissions, fills, cancellations.
+    # This method can be called asynchronously, ensure you use proper locks on thread-unsafe objects
+    def OnOrderEvent(self, orderEvent):
+        self.Debug(f"{self.Time} {orderEvent}")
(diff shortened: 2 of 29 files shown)


Algorithm.Python/TimeInForceAlgorithm.py
```diff
@@ -0,0 +1,70 @@
...
+from QuantConnect.Algorithm import *
+from QuantConnect.Orders import *
+
+### <summary>
+### Demonstration algorithm of time in force order settings.
+### </summary>
+### <meta name="tag" content="using data" />
+### <meta name="tag" content="using quantconnect" />
+### <meta name="tag" content="trading and orders" />
+class TimeInForceAlgorithm(QCAlgorithm):
```
mchandschuh: Thanks for adding a python version too!

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -457,5 +480,27 @@ private void SetPendingOrder(Order order)
             // only save off clones!
             _pending[order.Id] = order.Clone();
         }
+
+        private ITimeInForceHandler GetTimeInForceHandler(TimeInForce timeInForce)
```
mchandschuh: instead of having this function, it might be more concise to simply initialize the dictionaries with all the handlers already mapped, and then replace calls to this function with explicit dictionary access, `_timeInForceHandlers[timeInForce]`

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <summary>
+        /// Checks if an order has expired
+        /// </summary>
+        /// <param name="order">The order to be checked</param>
+        /// <returns>Returns true if the order has expired, false otherwise</returns>
+        public bool HasOrderExpired(Order order)
+        {
+            var exchangeHours = MarketHoursDatabase
+                .FromDataFolder()
+                .GetExchangeHours(order.Symbol.ID.Market, order.Symbol, order.Symbol.SecurityType);
```
mchandschuh: Instead of accessing the `FromDataFolder()` here, I think it may be better to accept the market hours database instance in the constructor, and then who ever creates this type can provide their own market hours -- we'll pretty much always use it from the data folder, but this can make this class testable w/out having a `market-hours-database.json` file in the correct place.

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            var time = _algorithm.UtcTime.ConvertFromUtc(exchangeHours.TimeZone);
+
+            return !exchangeHours.IsOpen(time, false);
```
mchandschuh: Is this correct? Isn't it possible to move to the next day and have `IsOpen` return true again. Is it more correct to compare the order's submit date and the current date (in the security's time zone) ?

What happens with 24 hour markets? If I say time in force = day at 23:59PM - does it cancel at the stroke of midnight?

Tests/Common/Orders/TimeInForces/TimeInForceHandlerTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            var fill1 = new OrderEvent(order.Id, order.Symbol, DateTime.UtcNow, OrderStatus.PartiallyFilled, OrderDirection.Buy, order.LimitPrice, 3, 0);
+            Assert.IsTrue(handler.IsFillValid(order, fill1));
+
+            var fill2 = new OrderEvent(order.Id, order.Symbol, DateTime.UtcNow, OrderStatus.Filled, OrderDirection.Buy, order.LimitPrice, 7, 0);
+            Assert.IsTrue(handler.IsFillValid(order, fill2));
+        }
+
+        [Test]
+        public void DayTimeInForceOrderExpiresAtMarketClose()
```
mchandschuh: This assumes a couple things -- one, that we get data at market close (almost guaranteed) or that we get data before the next market open (less of a guarantee) -- see comments in the day handler impl.

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
StefanoRaggi: Good point, this might need to be different per security type:
- Forex: both FXCM and Oanda use 5 PM New York time as day close
- Crypto: GDAX does not support Day, so maybe we should use midnight UTC here?

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// Checks if an order has expired
+        /// </summary>
+        /// <param name="order">The order to be checked</param>
+        /// <returns>Returns true if the order has expired, false otherwise</returns>
+        public bool HasOrderExpired(Order order)
+        {
+            var exchangeHours = _marketHoursDatabase.GetExchangeHours(order.Symbol.ID.Market, order.Symbol, order.Symbol.SecurityType);
+
+            var orderTime = order.Time.ConvertFromUtc(exchangeHours.TimeZone);
+            var time = _algorithm.UtcTime.ConvertFromUtc(exchangeHours.TimeZone);
```
mchandschuh: This algorithm dependency seems heavy handed. It looks like we're only using the algorithm to get the current time. A common pattern in lean models is to accept the `Security` object in the method. Accepting the security object would remove the algorithm dependency *and* the market hours database dependency, since the current time and the exchange hours can both be directly resolved form the security object.

Algorithm.CSharp/TimeInForceAlgorithm.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
+        {
+            if (_gtcOrderTicket == null)
+            {
+                // This order has a default time in force of GoodTilCancelled,
+                // it will never expire and will not be cancelled automatically.
+
+                _gtcOrderTicket = LimitOrder(_symbol, 10, 160m);
```
mchandschuh: For consistency, perhaps set `DefaultOrderProperties.TimeInForce = TimeInforce.GoodTilCancelled`

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -38,6 +39,7 @@ public class BacktestingBrokerage : Brokerage
         private readonly ConcurrentDictionary<int, Order> _pending;
         private readonly object _needsScanLock = new object();
         private readonly HashSet<Symbol> _pendingOptionAssignments = new HashSet<Symbol>();
+        private readonly Dictionary<TimeInForce, ITimeInForceHandler> _timeInForceHandlers = new Dictionary<TimeInForce, ITimeInForceHandler>();
```
mchandschuh: Hopefully we can change the implementations of `ITimeInForceHandler` such that they pull all information from the method parameters leading to a parameterless constructor, and then we can inline the creating of this map here of the form:
``` csharp
private readonly Dictionary<TimeInForce, ITimeInForceHandler> _timeInForceHandlers = new Dictionary<TimeInForce, ITimeInForceHandler>
{
    { TimeInForce.GoodTilCanceled, new GoodTilCanceledTimeInForceHandler() }
};
```

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1898,6 +1891,66 @@ private static OrderType ConvertOrderType(IBApi.Order order)
...
+
+                //case TimeInForce.FillOrKill:
+                //    return IB.TimeInForce.FillOrKill;
+
+                //case TimeInForce.ImmediateOrCancel:
+                //    return IB.TimeInForce.ImmediateOrCancel;
+
+                case TimeInForce.GoodTilCancelled:
+                default:
+                    return IB.TimeInForce.GoodTillCancel;
```
mchandschuh: Something I'm just realizing. In english 'canceled' and 'cancelled' are both equally correct spellings. We already use 'canceled' with a single L in `OrderStatus.Canceled` -- so let's keep the spelling the same for all our sanity :P

Common/Brokerages/FxcmBrokerageModel.cs
```diff
@@ -122,6 +122,16 @@ public override bool CanSubmitOrder(Security security, Order order, out Brokerag
                 return IsValidOrderPrices(security, OrderType.StopLimit, stopLimit.Direction, stopLimit.StopPrice, stopLimit.LimitPrice, ref message);
             }
 
+            // validate time in force
+            if (order.TimeInForce != TimeInForce.GoodTilCancelled)
+            {
+                message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
+                    "This model does not support " + order.TimeInForce + " time in force."
```
mchandschuh: awesomely perfect and simple, great stuff!

Common/Brokerages/InteractiveBrokersBrokerageModel.cs
```diff
@@ -62,24 +62,38 @@ public override bool CanSubmitOrder(Security security, Order order, out Brokerag
         {
             message = null;
 
+            // validate security type
+            if (security.Type != SecurityType.Equity &&
+                security.Type != SecurityType.Forex &&
+                security.Type != SecurityType.Option &&
+                security.Type != SecurityType.Future)
```
mchandschuh: no change required, but typically when there's a set we're comparing against, it's often cleaner to have:
``` csharp
private static readonly HashSet<SecurityType> SupportedSecurityTypes = new HashSet<SecurityType>{ SecurityType.Equity, SecurityType.Forex, SecurityType.Option, SecurityType.Future };

if (!SupportedSecurityTypes.Contains(security.Type)
{
    // error
}
```

Common/Brokerages/InteractiveBrokersBrokerageModel.cs
```diff
@@ -62,24 +62,38 @@ public override bool CanSubmitOrder(Security security, Order order, out Brokerag
...
             //https://www.interactivebrokers.com/en/?f=%2Fen%2Ftrading%2FforexOrderSize.php
-            switch (order.SecurityType)
+            if (security.Type == SecurityType.Forex &&
+                !IsForexWithinOrderSizeLimits(order.Symbol.Value, order.Quantity, out message))
+            {
+                return false;
+            }
+
+            // validate time in force
+            if (order.TimeInForce != TimeInForce.GoodTilCancelled && order.TimeInForce != TimeInForce.Day)
```
mchandschuh: Similar comment to the above, often cleaner to define a collection of supported cases and check the collection. This makes the code read nicer and makes it easier to update the conditions in the future as we add/remove support for various things.

Common/Interfaces/IAlgorithm.cs
```diff
@@ -328,6 +328,11 @@ IFutureChainProvider FutureChainProvider
             get;
         }
 
+        /// <summary>
+        /// Gets the market hours database in use by this algorithm
+        /// </summary>
+        MarketHoursDatabase MarketHoursDatabase { get; }
```
mchandschuh: I'm reluctant about this. In theory, this is a great idea, but I have concerns on the direction of the dependency. Currently QCAlgorithm is hard-coded to fetch `FromDataFolder()` (as is the rest of the application). I love the intent behind this, making things instance based, but if we're going to make accesses to the market hours db instance based we should tackle that in a separate PR and update the solution to follow that new pattern.

Common/Interfaces/ITimeInForceHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// Handles the time in force for an order
+    /// </summary>
+    public interface ITimeInForceHandler
+    {
+        /// <summary>
+        /// Checks if an order has expired
+        /// </summary>
+        /// <param name="order">The order to be checked</param>
+        /// <returns>Returns true if the order has expired, false otherwise</returns>
+        bool HasOrderExpired(Order order);
```
mchandschuh: We spoke about adding the `Security` instance here to match existing patterns w.r.t models in lean. For example, look at the fee/fill/slippage models, all of which accept the security object in addition to the order object. This will remove the need for the algorithm and market hours db dependencies

Algorithm/QCAlgorithm.cs
```diff
@@ -72,7 +72,7 @@ public partial class QCAlgorithm : MarshalByRefObject, IAlgorithm
         /// <summary>
         /// Gets the market hours database in use by this algorithm
         /// </summary>
-        protected MarketHoursDatabase MarketHoursDatabase { get; }
+        public MarketHoursDatabase MarketHoursDatabase { get; }
```
mchandschuh: Please revert this change.

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -361,6 +385,12 @@ public void Scan()
 
                     foreach (var fill in fills)
                     {
+                        // check if the fill should be emitted
+                        if (!timeInForceHandler.IsFillValid(security, order, fill))
+                        {
+                            break;
```
mchandschuh: OK for now, but this will change, we can't assume the implementation of `IsFillValid` negates all fills in the event there are multiple fills -- when we wire up order event projections this will change, so it's ok for now.

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+            var orderTime = order.Time.ConvertFromUtc(exchangeHours.TimeZone);
+            var time = security.LocalTime;
+
+            bool expired;
+            switch (order.SecurityType)
+            {
+                case SecurityType.Forex:
+                case SecurityType.Cfd:
+                    // expires at 5 PM NewYork time
+                    var expiryTime = new DateTime(orderTime.Date.Year, orderTime.Date.Month, orderTime.Date.Day, 17, 0, 0);
```
mchandschuh: we say they expire at 5PM new york but are converting to exchange tz with no guarantee that it's new york time. Is the 5PM represented in the market hours segments? Maybe there's a more general way to define this cut off time? I don't know the answer, but worth considering.

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+                    expired = time.Date > orderTime.Date || time.ConvertTo(exchangeHours.TimeZone, TimeZones.NewYork) >= expiryTime;
```
mchandschuh: What if I submit an order at 5:01PM -- is that order good until the next day's 5PM, in which case I think the first part of this condition is incorrect.

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -39,6 +40,14 @@ public class BacktestingBrokerage : Brokerage
         private readonly object _needsScanLock = new object();
         private readonly HashSet<Symbol> _pendingOptionAssignments = new HashSet<Symbol>();
 
+        private readonly Dictionary<TimeInForce, ITimeInForceHandler> _timeInForceHandlers = new Dictionary<TimeInForce, ITimeInForceHandler>
+        {
+            { TimeInForce.GoodTilCanceled, new GoodTilCanceledTimeInForceHandler() },
+            { TimeInForce.Day, new DayTimeInForceHandler() },
+            // Custom time in force will be renamed to GTD soon and will have its own new handler
+            { TimeInForce.Custom, new GoodTilCanceledTimeInForceHandler() }
```
mchandschuh: This looks nice and clean! With the upcoming refactor this is going to change (again) such that the implementation will ride along w/ the order object -- but for now it's nice a clear what the mappings are.

Common/Orders/TimeInForces/DayTimeInForceHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+
+                    orderTime = order.Time.ConvertFromUtc(cutOffTimeZone);
+                    var expiryTime = orderTime.Date.Add(cutOffTimeSpan);
+                    if (orderTime.TimeOfDay > cutOffTimeSpan)
+                    {
+                        // order submitted after 5 PM, expiry on next date
+                        expiryTime = expiryTime.AddDays(1);
+                    }
+
+                    expired = time.ConvertTo(exchangeHours.TimeZone, cutOffTimeZone) >= expiryTime;
```
mchandschuh: as a reminder, we discussed using `exchangeHours.GetNextMarketClose(orderTime, false)` to determine when the end of 'day' should be.

mchandschuh: you're 100% correct, it won't work that way -- let's keep it as you have it. If we ever need a more dynamic value we can look into sourcing the value through the brokerage model.

PR #1939, opened 2018-04-30
Add Order.ModifiedTime

### LINKED ISSUE #1596: Adds DateTime property that reflects the last time an order was updated
The [Order](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/Order.cs) object does not have a property that represents the last time the order was updated. 
At the moment, we only have the time it was created and with this property, we would be able to get the time an order was filled or cancelled.

mchandschuh:
#### Description

Add `Order.ModifiedTime` property. Updated on each order event to the order event's time.

#### Related Issue

Closes #1596 

#### Motivation and Context

Much requested feature -- provides insight in to when the order object was last updated.

No.

Locally, also updated the `UpdateOrderRegressionAlgorithm` to confirm the modified time gets updated.

#### Types of changes

#### Checklist:


Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -894,6 +894,9 @@ private void HandleOrderEvent(OrderEvent fill)
             // set the status of our order object based on the fill event
             order.Status = fill.Status;
 
+            // set the modified time of the order to the fill's timestamp
+            order.ModifiedTime = fill.UtcTime;
```
StefanoRaggi: Isn't this called on every order status change including invalid, cancel ?
Maybe should be called `LastEventTime` ?

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -894,6 +894,35 @@ private void HandleOrderEvent(OrderEvent fill)
             // set the status of our order object based on the fill event
             order.Status = fill.Status;
 
+            // update the ticket and order after we've processed the fill, but before the event, this way everything is ready for user code
```
StefanoRaggi: This comment line is duplicated, was it meant to be moved here?

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
mchandschuh: agreed -- ended u splitting into three separate properties

PR #1943, opened 2018-05-01
Implements MaximumSectorExposureRiskManagementModel

### LINKED ISSUE #1942: Implement RiskManagement model using fundamental data
#### Expected Behavior
Provide different examples of framework models.

#### Actual Behavior
At the moment we only have one risk management model example.

#### Potential Solution
Adds a `IRiskManagementModel` that uses, for example, fundamental data.

#### Checklist

AlexCatarino:
#### Description
- Minor models fixes:
  - ConstantAlphaModel.py: sets default value `None` for magnitude and confidence;
  - BlackLittermanPortfolioConstructionModel.py: removes logging.
- Implements python version of EqualWeightingPortfolioConstructionModel;
- Implements MaximumSectorExposureRiskManagementModel:
  Provides an implementation of `IRiskManagementModel` that limits the sector exposure to the specified percentage;
- Implements SectorExposureRiskFrameworkAlgorithm:
  This algorithm and its regression test are meant to test the `MaximumSectorExposureRiskManagementModel`.

#### Related Issue
Closes #1942 

#### Motivation and Context
Provide examples of framework models

Yes. 

Regression test.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
+++ b/Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
@@ -29,6 +29,7 @@ namespace QuantConnect.Algorithm.Framework.Risk
     public class MaximumSectorExposureRiskManagementModel : IRiskManagementModel
     {
         private readonly decimal _maximumSectorExposure;
+        private readonly PortfolioTargetCollection _targetsCollection;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="MaximumSectorExposureRiskManagementModel"/> class
@@ -39,6 +40,7 @@ namespace QuantConnect.Algorithm.Framework.Risk
             )
         {
             _maximumSectorExposure = Math.Abs(maximumSectorExposure);
+            _targetsCollection = new PortfolioTargetCollection();
         }
 
         /// <summary>
@@ -50,11 +52,10 @@ namespace QuantConnect.Algorithm.Framework.Risk
         {
             var maximumSectorExposureValue = algorithm.Portfolio.TotalPortfolioValue * _maximumSectorExposure;
 
-            var targetCollection = new PortfolioTargetCollection();
-            targetCollection.AddRange(targets);
+            _targetsCollection.AddRange(targets);
 
             // Group the securities by their sector
-            var groupBySector = algorithm.Securities
+            var groupBySector = algorithm.UniverseManager.ActiveSecurities
                 .Where(x => x.Value.Fundamentals.HasFundamentalData)
                 .GroupBy(x => x.Value.Fundamentals.CompanyReference.IndustryTemplateCode);
 
@@ -70,7 +71,7 @@ namespace QuantConnect.Algorithm.Framework.Risk
                     var absoluteHoldingsValue = security.Value.Holdings.AbsoluteHoldingsValue;
 
                     IPortfolioTarget target;
-                    if (targetCollection.TryGetValue(security.Value.Symbol, out target))
+                    if (_targetsCollection.TryGetValue(security.Value.Symbol, out target))
                     {
                         absoluteHoldingsValue = security.Value.Price * Math.Abs(target.Quantity) *
                             security.Value.SymbolProperties.ContractMultiplier *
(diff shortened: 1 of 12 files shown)


Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// <param name="targets">The current portfolio targets to be assessed for risk</param>
+        public IEnumerable<IPortfolioTarget> ManageRisk(QCAlgorithmFramework algorithm, IPortfolioTarget[] targets)
+        {
+            var maximumSectorExposureValue = algorithm.Portfolio.TotalPortfolioValue * _maximumSectorExposure;
+
+            var targetCollection = new PortfolioTargetCollection();
+            targetCollection.AddRange(targets);
+
+            // Group the securities by their sector
+            var groupBySector = algorithm.Securities
```
mchandschuh: framework models should **not** loop over `algorithm.Securities` since it contains *all* securities that were ever added to the algorithm. Instead you should use `OnSecuritiesChanged` to keep a set of the current securities. You can check out `ConstantAlphaModel.OnSecuritiesChanged' for how he keeps track of current in-universe securities (we should definitely add a helper function for this). Here's the quick way of doing it:
``` csharp
NotifiedSecurityChanges.UpdateCollection(_securities, changes);
```

there's also an `UpdateDictionary` version of the method.

Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// Manages the algorithm's risk at each time step
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance</param>
+        /// <param name="targets">The current portfolio targets to be assessed for risk</param>
+        public IEnumerable<IPortfolioTarget> ManageRisk(QCAlgorithmFramework algorithm, IPortfolioTarget[] targets)
+        {
+            var maximumSectorExposureValue = algorithm.Portfolio.TotalPortfolioValue * _maximumSectorExposure;
+
+            var targetCollection = new PortfolioTargetCollection();
+            targetCollection.AddRange(targets);
```
mchandschuh: This collection type is intended to be a class level variable and not a method level variable. As a class member it maintains a complete set of all portfolio targets so you can operate against a 'full view' instead of the potentially streaming targets (which can come in one by one as alpha is generated). Have a look at `VolumeWeightedAveragePriceExecutionModel` for an example of `PortoflioTargetCollection`'s usage.

Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
```diff
@@ -0,0 +1,118 @@
...
+
+                foreach (var security in securities)
+                {
+                    var absoluteHoldingsValue = security.Value.Holdings.AbsoluteHoldingsValue;
+
+                    IPortfolioTarget target;
+                    if (targetCollection.TryGetValue(security.Value.Symbol, out target))
+                    {
+                        absoluteHoldingsValue = security.Value.Price * Math.Abs(target.Quantity) *
+                            security.Value.SymbolProperties.ContractMultiplier *
```
mchandschuh: Very smart! I think because of this you actually do need to loop over the securities collection since `OnSecuritiesChanged` will say 'removed' even if you hold stock -- in this way it's like a warning that it's been removed but if you hold stock/open orders we won't remove the subscription.

I've opened https://github.com/QuantConnect/Lean/issues/1945 to add a helper to get securities w/ active data feed subscriptions.

Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
```diff
@@ -0,0 +1,118 @@
...
+
+                        IPortfolioTarget target;
+                        if (targetCollection.TryGetValue(symbol, out target))
+                        {
+                            quantity = target.Quantity;
+                        }
+
+                        if (quantity != 0)
+                        {
+                            yield return new PortfolioTarget(symbol, quantity / ratio);
```
mchandschuh: awesome!!

Algorithm.Framework/Risk/MaximumSectorExposureRiskManagementModel.cs
AlexCatarino: Thanks. If the security is removed, `Quantity` and, consequently, `AbsoluteHoldingValue` are zero and it won't be an issue, but looping only through active subscription makes is more efficient.

Issue #1937, opened 2018-04-30
AddOptionContract() is unable to fetch the current price of the option contract


jingwu74:
#### Expected Behavior

With OptionChainProvider, the specified option contract should be available to trade immediately after AddOptionContract(). 
#### Actual Behavior

As AddOptionContract() is unable to fetch the current price of the option contract, in the option algorithms using OptionChainProvider, the filtered contract should wait until next time step to be available to trade. The algorithm returns the error message like `
Backtest Handled Error: AMZN  170616C01000000: asset price is $0. If using custom data make sure you've set the 'Value' property.` 

#### Reproducing the Problem

https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/OptionChainProviderAlgorithm.cs
#### System Information

LEAN
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

mchandschuh:
We've decided that for now you can't add contract and purchase that contract on the same time step. This is actually (somewhat) realistic as far as running in live mode is concerned. For example, if we were to add a subscription for new data and on the very next line of code query the price, there's no guarantee that we would have received a tick for that contract yet (so price would still equal zero). The alternative is to issue a history request every time we add a subscription. We actually used to do this but determined that performance cost out weighed the marginal utility. For now you'll need to add the contract on one time step and on the next time step submit the order for the contract.

PR #1946, opened 2018-05-01
Fix issue of no traded options in OptionChainProviderAlgorithm 

jingwu74:
#### Description

The OptionChainProviderAlgorithm was modified to trade options at the next time step after subscribing the data for the specified contract with AddOptionContract().
#### Related Issue

Fixes [#1937](https://github.com/QuantConnect/Lean/issues/1937)
#### Motivation and Context

The algorithm used to submit the order instantly after `AddOptionContract()`. The orders fails to fill because the data has not been received and the price of the specified contract is zero.

No

Regression test
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OptionChainProviderAlgorithm.cs
+++ b/Algorithm.CSharp/OptionChainProviderAlgorithm.cs
@@ -37,19 +37,25 @@ namespace QuantConnect.Algorithm.CSharp
     public class OptionChainProviderAlgorithm : QCAlgorithm
     {
         private Symbol _equitySymbol;
-     
+        private Symbol _optionContract = string.Empty;
+
         public override void Initialize()
         {
-            SetStartDate(2017, 06, 01);
-            SetEndDate(2017, 07, 01);
+            SetStartDate(2015, 12, 24);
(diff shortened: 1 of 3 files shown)


jingwu74:
Initializing the contract with `None` does not pass the `self.Securities.ContainsKey(self.contract)`. I got error like `Runtime Error: ArgumentNullException : Value cannot be null`.

Algorithm.CSharp/OptionChainProviderAlgorithm.cs
```diff
@@ -37,19 +37,25 @@ namespace QuantConnect.Algorithm.CSharp
...
 
         public override void OnData(Slice data)
         {
-            if (!Portfolio.Invested)
+            if (!Portfolio[_equitySymbol].Invested)
+            {
+                MarketOrder(_equitySymbol, 100);
+            }
+
+            if (!(Securities.ContainsKey(_optionContract) && Portfolio[_optionContract].Invested))
```
mchandschuh: do we need this invested condition here? This block of code is only adding the contract

PR #1947, opened 2018-05-01
Add UniverseManager.GetSecurities

### LINKED ISSUE #1945: Create helper method for getting active securities w/ data subscriptions
#### Expected Behavior

Should be easy to determine the active set of securities via a simple function call.

#### Actual Behavior

Currently one needs to aggregate `OnSecuritiesChanged` events -- and even that can be wrong if a security gets deselected from the universe but you hold stock, you'll get the removed event.

#### Potential Solution

The correct way of determining this is to loop over the universe objects and collate their membership lists.

#### Checklist

mchandschuh:
#### Description

Fetches the unique set of securities that are currently members of at least one universe. This is the set of active/tradable securities that are currently receiving data. It does not guarantee that we've received pricing data, but does guarantee that the security is a member of the universe and we've minimally added a subscription for data and have not yet removed that subscription (although it is possible for the sub to end before the security is removed -- very rare/unlikely scenario though).

#### Related Issue

Closes #1945 

#### Motivation and Context

Removes the need to manually (and often buggily) keep track of active securities via security change events. Provides an easy mechanism for looping over the current set of active securities.

Yes. @jaredbroad @jingwu74 -- we should document this as many users have asked how to do this. Until now the answer was to manage the state via security changes, but it turns out that isn't exactly correct since it's possible to receive the removed event before we remove the data feed subscription (if security holds stock and/or has open orders this happens).

Tested locally w/ debug output to confirm the set of securities.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/UniverseManager.cs
+++ b/Common/Securities/UniverseManager.cs
@@ -20,6 +20,7 @@ using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
 using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Util;
 
 namespace QuantConnect.Securities
 {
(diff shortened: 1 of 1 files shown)


mchandschuh:
reworking

Issue #1915, opened 2018-04-25
VS plugin - New action "Create Project"


Martin-Molinero:
#### Expected Behavior

New action button that enables the user to create a new project from the QuantConnect tool window.
User should be able to set project name and programming language

#### Actual Behavior

Not implemented.
#### Potential Solution

N/A
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #1948, opened 2018-05-01
VS plugin - Adding new action "Create Project"

Martin-Molinero:
#### Description

- Adding new action "Create Project" to the tool window
     - Adding popup dialog for new project
- Projects will display the associated Language as a name prefix 
- Send For Backtesting will now open backtest
- Moving ProjectNameDialog.ComboboxItem to a new file and renaming to ComboboxProjectItem
- Fix for suggested project in ProjectNameDialog. Before it suggested only with the project name causing a new project with the same name to be created. Now it will use the suggested project name to find the corresponding available project and use it.
- Tool window combobox will now auto select first project and load associated backtests
- Replacing weird positioning of labels in dialogs with padding 
- Updating documentation and images.

#### Related Issue

https://github.com/QuantConnect/Lean/issues/1915
#### Motivation and Context

Add more value to VS plugin

Yes. I'm already updating VS plugin documentation

Manually VS17 and VS15
#### Types of changes

- [x] Non-functional change (xml comments/documentation/ect)

#### Checklist:


VisualStudioPlugin/BacktestNameProvider.cs
```diff
@@ -38,18 +38,18 @@ internal static class BacktestNameProvider
...
-        private static string[] _verbs = { "Determined", "Pensive", "Adaptable", "Calculating", "Logical",
+        private static readonly string[] _verbs = { "Determined", "Pensive", "Adaptable", "Calculating", "Logical",
            "Energetic", "Creative", "Smooth", "Calm", "Hyper-Active",
            "Measured", "Fat", "Emotional", "Crying", "Jumping",
            "Swimming", "Crawling", "Dancing", "Focused", "Well Dressed",
            "Retrospective", "Hipster", "Square", "Upgraded", "Ugly",
            "Casual", "Formal", "Geeky", "Virtual", "Muscular",
            "Alert", "Sleepy" };
-        private static Random _random = new Random();
+        private static readonly Random _random = new Random();
```
mchandschuh: thanks!

VisualStudioPlugin/NewProjectDialog.xaml.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        /// <summary>
+        /// User provided project programming laguange
+        /// </summary>
+        public Language NewProjectLanguage { get; private set; }
+
+        /// <summary>
+        /// Supported programming languages. This is to filter enum Language, which has unsupported languages.
+        /// </summary>
+        private static string[] _supportedLanguages = { "FSharp", "CSharp", "Python" };
```
mchandschuh: no change required, but curious why choose to use strings here instead of the `Language` enum directly?
Also, this member can be marked as `readonly`.

VisualStudioPlugin/ComboBoxProjectItem.cs
```diff
@@ -0,0 +1,48 @@
...
+        public ComboboxProjectItem(int id, string name, Language language)
+        {
+            Id = id;
+            Name = name;
+            Language = language;
+            var memInfo = typeof(Language).GetMember(Language.ToString());
+            var attr = memInfo[0].GetCustomAttributes(false).OfType<EnumMemberAttribute>().FirstOrDefault();
+            if (attr != null)
+            {
+                _namePrefix = attr.Value + " - ";
```
mchandschuh: what if `attr` is null? I think an `else { _namePrefix = Language.ToString(); }` might make sense here.

VisualStudioPlugin/NewProjectDialog.xaml.cs
Martin-Molinero: Your right! Using `Language` would of been a cleaner approach. My brain just wasn't contributing when I thought this through. I'll see to it in the next patch 👌

PR #1982, opened 2018-05-09
Changes Insight.Group return type to IEnumerable<Insight> from Guid

### LINKED ISSUE #1981: Change Insight.Group return type to IEnumerable<Insight> from Guid
#### Expected Behavior
Get `IEnumerable<Insight>` from `Insight.Group(params Insight[])` call.

#### Actual Behavior
`Insight.Group(params Insight[])` returns the Guid that was assigned to the group members.

#### Potential Solution
Change return type.

#### Checklist

AlexCatarino:
#### Description
Changes `Insight.Group` return type to `IEnumerable<Insight>` from `Guid`.

#### Related Issue
Closes #1981

Unit test.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Algorithm/Framework/Alphas/Insight.cs
+++ b/Common/Algorithm/Framework/Alphas/Insight.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using Newtonsoft.Json;
 using QuantConnect.Algorithm.Framework.Alphas.Serialization;
 
@@ -201,7 +202,7 @@ namespace QuantConnect.Algorithm.Framework.Alphas
         /// Creates a new, unique group id and sets it on each insight
         /// </summary>
         /// <param name="insights">The insights to be grouped</param>
-        public static Guid Group(params Insight[] insights)
+        public static IEnumerable<Insight> Group(params Insight[] insights)
         {
             if (insights == null)
             {
@@ -217,9 +218,9 @@ namespace QuantConnect.Algorithm.Framework.Alphas
                 }
 
                 insight.GroupId = groupId;
-            }
 
-            return groupId;
+                yield return insight;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Common/Algorithm/Framework/Alphas/Insight.cs
```diff
@@ -217,11 +218,17 @@ public static Guid Group(params Insight[] insights)
...
-            return groupId;
+                yield return insight;
+            }
         }
 
+        /// <summary>
+        /// Creates a new, unique group id and sets it on each insight
+        /// </summary>
+        /// <param name="insight">The insight to be grouped</param>
+        public static IEnumerable<Insight> Group(Insight insight) => Group(new[] { insight });
```
mchandschuh: Is this necessary? The `params` modifier should do this by default? Is this added for python convenience?

Common/Algorithm/Framework/Alphas/Insight.cs
```diff
@@ -217,11 +218,17 @@ public static Guid Group(params Insight[] insights)
                 }
 
                 insight.GroupId = groupId;
-            }
 
-            return groupId;
+                yield return insight;
```
mchandschuh: I don't think this should be lazy. Let's remove the `yield return` syntax, otherwise users are definitely going to forget to materialize the list, for example, the existing pairs trading example doesn't materialize `Insight.Group` so I assume that is now broken by this change.

Common/Algorithm/Framework/Alphas/Insight.cs
AlexCatarino: Yes. This is due to pythonnet inability to solve this:
```python
Insight.Group(insight)
```
as it says that it cannot find an overload. 
I could take a look at pythonnet and solve this issue.

mchandschuh: But is that required? a Group method for one insight? I think all other places we use C# `params` we use python `[]` array notation. I'm not against adding this, just wondering if it adds much value, that's all.

PR #1997, opened 2018-05-10
Port C# demonstration algorithms to python

### LINKED ISSUE #1610: Port C# Demonstration Algorithms->Py
These algorithms should be ported to Python to match their C# equivalents:

https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/IndicatorSuiteAlgorithm.cs

https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/RenkoConsolidatorAlgorithm.cs

https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/CoarseUniverseTop5DollarVolumeAlgorithm.cs

https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/DisplacedMovingAverageRibbon.cs

AlexCatarino:
#### Description
The following algorithms were ported but requested regression tests were missing.
- RenkoConsolidatorAlgorithm
- CoarseFundamentalTop5Algorithm
- DisplacedMovingAverageRibbon

#### Related Issue
Closes #1610 

#### Motivation and Context
Provide matching examples in C# and python.

No.

Added regression tests.

#### Types of changes
- [c] New feature (non-breaking change which adds functionality)

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CoarseFundamentalTop5Algorithm.cs
+++ b/Algorithm.CSharp/CoarseFundamentalTop5Algorithm.cs
@@ -66,7 +66,7 @@ namespace QuantConnect.Algorithm.CSharp
         //Data Event Handler: New data arrives here. "TradeBars" type is a dictionary of strings so you can access it by symbol.
         public void OnData(TradeBars data)
         {
-            Console.WriteLine($"OnData({UtcTime:o}): Keys: {string.Join(", ", data.Keys.OrderBy(x => x))}");
+            Log($"OnData({UtcTime:o}): Keys: {string.Join(", ", data.Keys.OrderBy(x => x))}");
 
             // if we have no changes, do nothing
             if (_changes == SecurityChanges.None) return;
@@ -97,7 +97,7 @@ namespace QuantConnect.Algorithm.CSharp
 
         public override void OnOrderEvent(OrderEvent fill)
         {
-            Console.WriteLine($"OnOrderEvent({UtcTime:o}):: {fill}");
+            Log($"OnOrderEvent({UtcTime:o}):: {fill}");
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Algorithm.CSharp/CoarseFundamentalTop5Algorithm.cs
```diff
@@ -66,7 +66,7 @@ public static IEnumerable<Symbol> CoarseSelectionFunction(IEnumerable<CoarseFund
         //Data Event Handler: New data arrives here. "TradeBars" type is a dictionary of strings so you can access it by symbol.
         public void OnData(TradeBars data)
         {
-            Console.WriteLine($"OnData({UtcTime:o}): Keys: {string.Join(", ", data.Keys.OrderBy(x => x))}");
+            Log($"OnData({UtcTime:o}): Keys: {string.Join(", ", data.Keys.OrderBy(x => x))}");
```
mchandschuh: Ironically, these were purposefully `Console.WriteLine` statements. Using `Console.WriteLine` from within the regression tests pipes it directly to `Log.Trace` so they're written synchronously with other logging statements. If we use `Log`, then those messages get queued and then dequeued in the result handler a few seconds later, making debugging some data feed issues much harder.

*Also*, since `Console.WriteLine` pipes to `algorithm.Log` in normal circumstances (running locally or in QC cloud), I think it's fine to leave as `Console.WriteLine`.

Another also, the `forward-console-messages` configuration value is what controls this piping behavior. `true` forwards the messages to `algorithm.Log` while `false` sends them directly to the screen (IMO, as you would expect from a `Console.WriteLine` statement)

Algorithm.CSharp/DisplacedMovingAverageRibbon.cs
```diff
@@ -46,7 +46,7 @@ public override void Initialize()
             SetStartDate(2009, 01, 01);
             SetEndDate(2015, 01, 01);
 
-            AddSecurity(SecurityType.Equity, "SPY", Resolution.Minute);
+            AddSecurity(SecurityType.Equity, "SPY", Resolution.Daily);
```
mchandschuh: Thanks for the commit message, at first I was like, whyyy :)

for posterity: this was modified to daily so we can run the tests using the data available in the open source.

PR #1999, opened 2018-05-11
Remove disposed security data from TimeSlice

### LINKED ISSUE #1604: Data for recently removed securities shows up in OnData
In the time step that universe selection occurs, it is possible to get data for securities that were just removed via universe selection. This is due to sometimes synchronizing the data before the universe selection logic has been applied, so we've already dequeued the data.

See https://github.com/QuantConnect/Lean/pull/1444 for a proposed solution. Ideally, we could address this structurally, that is, perhaps by making universe selection always happen before we synchronize other data at that time step. This would preclude us from needing to manage extra state about pending removals and such.

mchandschuh:
#### Description

If we pull data and on the same time step that security gets removed,
we can still get that data in OnData(Slice) even though it was removed.
This change filters out removed securities by tracking a reference to
the subscription's disposed flag. Another change was made to wait until
the end of the time step to dispose of subscriptions.

#### Related Issue

Closes #1604 

#### Motivation and Context

When a security is removed from a universe we should stop receiving data for that security.

No.

Regression tests updated to confirm slice data matches active securities.

#### Types of changes

#### Checklist:

- [x] All new and existing tests passed. (those that pass on master)

### DIFF
--- a/Engine/DataFeeds/DataFeedPacket.cs
+++ b/Engine/DataFeeds/DataFeedPacket.cs
@@ -1,11 +1,11 @@
 /*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -17,6 +17,7 @@
 using System.Collections.Generic;
 using QuantConnect.Data;
 using QuantConnect.Securities;
+using QuantConnect.Util;
 
 namespace QuantConnect.Lean.Engine.DataFeeds
 {
@@ -25,7 +26,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
     /// </summary>
     public class DataFeedPacket
     {
-        private readonly List<BaseData> _data;
+        private readonly IReadOnlyRef<bool> _isRemoved;
 
         /// <summary>
         /// The security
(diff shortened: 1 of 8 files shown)


Engine/DataFeeds/Subscription.cs
```diff
@@ -189,7 +188,22 @@ public void Dispose()
         /// <filterpriority>2</filterpriority>
         public override int GetHashCode()
         {
-            return Configuration.Symbol.GetHashCode();
+            return Configuration.GetHashCode();
```
StefanoRaggi: Nice bug fix! 👍

mchandschuh:
This has been cleaned up and verified in the QC cloud using the following algorithm. Confirmed the algorithm throws the exception on current `master` and passes with this branch.

```
    public class UniverseAlgorithm : QCAlgorithm
    {
    	private int securityCount = 50;
        private int onDataCalls = 0;
        private int onSecuritiesChangedCalls = 0;
        {
            SetStartDate(2017, 10, 02);
            SetEndDate(2017, 10, 4);

            AddUniverse(coarse => {
                return (from c in coarse
                    orderby c.DollarVolume descending
                    select c.Symbol).Take(securityCount);
            });
        }

        {
            if (data.Count != securityCount)
            {
                throw new Exception("Test failed. Excepted 50 data points.");
            }
            onDataCalls++;
            Log($"OnData {data.Count} symbols at {data.Time}");
        }

        public override void OnSecuritiesChanged(SecurityChanges changes)
        {
            onSecuritiesChangedCalls++;
            Log($"Added {changes.AddedSecurities.Count}. Removed {changes.RemovedSecurities.Count}");
        }

        {
            Log($"{onDataCalls} OnData Calls {onSecuritiesChangedCalls} OnSecuritiesChanged Calls");
        }
    }
```

Engine/DataFeeds/TimeSlice.cs
```diff
@@ -158,6 +158,12 @@ public class TimeSlice
             // ensure we read equity data before option data, so we can set the current underlying price
             foreach (var packet in data)
             {
+                // filter out packets for removed subscriptions
+                if (packet.IsSubscriptionRemoved)
```
jaredbroad: Does this ever get hit now?

mchandschuh: This gets hit during universe selection algorithms. The algorithm listed above on current `master` has some time steps that emit data for 60 securities instead of 50. After this change, each time step emits data for 50 securities as expected. The underlying cause was enumerating the data feed subscription before performing universe selection. I had considered using sorting to enforce an order to the enumeration but felt this required less coupling/coordination between data feed and universe selection mechanics.

Engine/DataFeeds/DataFeedPacket.cs
```diff
@@ -78,11 +77,13 @@ public DataFeedPacket(Security security, SubscriptionDataConfig configuration)
...
         /// internally and NOT copied.</param>
-        public DataFeedPacket(Security security, SubscriptionDataConfig configuration, List<BaseData> data)
+        /// <param name="isSubscriptionRemoved">Reference to whether or not the subscription has since been removed, defaults to false</param>
+        public DataFeedPacket(Security security, SubscriptionDataConfig configuration, List<BaseData> data, IReadOnlyRef<bool> isSubscriptionRemoved = null)
         {
             Security = security;
             Configuration = configuration;
-            _data = data;
+            Data = data;
+            _isRemoved = isSubscriptionRemoved ?? Ref.Create(false);
```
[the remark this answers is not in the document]
mchandschuh: `Ref` is a custom type created to allow us to pass references to value types. A more common implementation is `Box<T>` which would maintain it's own storage. `Ref<T>` is defined in such a way that it doesn't maintain it's own storage and instead uses functions to get/set a value. Originally written to enable a shared fill forward increment, as the fill forward time span is maintained by the data feeds but required in all enumerator stacks and potentially updated each time a subscription is added or removed from the feed.

Issue #996, opened 2017-06-28
[Crypto] Enable Currency Abbreviations with More than 3 Characters


lnaie:
In the crypto-currency market there are coins with more than 3 chars. Check this out: https://coinmarketcap.com/all/views/all/.

I'm not sure if it's already supported deep down in the pipeline, but it should.

jaredbroad:
Great point thank you @lnaie

lnaie:
Could somebody pin point the work that has to be done and maybe the files/abstractions affected by this change?

viliwonka:
I have already done this. Should I push the changes?

jaredbroad:
@viliwonka That would be awesome! I've made a branch for you "crypto-symbol-length" -- if possible can you PR it there for review

viliwonka:
Ok, let me work this one out, never worked with branches.

viliwonka:
Ok excuses for me being complete noob with GitHub here, but I wasn't able to open PR with committing code, I will continue with it tomorrow.

PR #2022, opened 2018-05-21
Crypto symbol length bug fixes

viliwonka:
Now instead of having max 4 chars per currency code, which was hard-coded, it has const now, that can be changed if need arises. 

Also this PR fixes the EnsureCurrencyDataFeed bug, for example SENTETH didn't got ETHUSD subscription & Lean reported missing SENTUSD pair instead.

#### Description

For fixing fixed length currency code, Currencies.cs, Cash.cs, CashTests.cs, ForexTests.cs were modified.

For EnsureCurrencyDataFeed bug, Cash.cs EnsureCurrencyDataFeed function was main target.
Forex.cs had code style fixes and has received 1 new function. ForexTests.cs have new tests.

#### Related Issue

#996 

#### Motivation and Context

It solves 2 things:
1. enables currency codes with more than 4 chars
2. fixed bug in EnsureCurrencyDataFeed for new cryptocurrencies

In Currencies.cs, two new consts were added, that should be in documentation:

    public const int MaxCharactersPerCurrencyCode = 6;

and

    public const int MaxCharactersPerCurrencyPair = 2*MaxCharactersPerCurrencyCode;

In Forex.cs, one new function has been added:

    public static string CurrencyPairDual(string currencyPair, string knownSymbol)

By running Unit tests and template algorithms in backend mode.  

Adding `AddCrypto("SENTETH");` into algorithm results in backtest wanting 2 symbols, "SENTETH" and "ETHUSD", instead of "SENTUSD".
 
#### Types of changes

- [x] Non-functional change (xml comments/documentation/ect)

#### Checklist:

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -66,9 +66,10 @@ namespace QuantConnect.Securities
         {
             get
             {
-                return ConversionRateSecurity.Select(conSec => conSec.RateSecurity.Symbol).ToList();
+                return (ConversionRateSecurity == null? new List<Symbol>() : ConversionRateSecurity.Select(conSec => conSec.RateSecurity.Symbol).ToList());
             }
         }
+        
         /// <summary>
         /// Gets the security used to apply conversion rates.
         /// If this cash represents the account currency, then null is returned.
@@ -303,12 +304,11 @@ namespace QuantConnect.Securities
             // Example #1: RENUSD doesn't exist, but there is RENETH  and ETHUSD,  from which we can calculate RENUSD
             // Example #2: RENUSD doesn't exist, but there is RENUSDT and USDTUSD, from which we can calculate RENUSD
 
-            // Finds pairs that contain REN such as RENBTC and RENETH
+            // Make a copy
             var existingPotentials = Currencies.CryptoCurrencyPairs.Select(x => x);
-            //.Where(cryptoPair => cryptoPair.Contains(this.Symbol));
-
-            // Order secondaryPotentials by whenever they are already contained in securities.
-            // This is must; if you add RENETH, it will use ETHUSD as a conversion pair for USD, and not BTCUSD
+           
+            // Order secondaryPotentials by whenever they are already contained in securities object.
+            // This is must; if you use AddCrypto(RENETH), currency conversion will use ETHUSD as a conversion pair for USD, and not BTCUSD
             existingPotentials = existingPotentials
             .OrderByDescending(cryptoPair =>
             {
@@ -340,8 +340,6 @@ namespace QuantConnect.Securities
                     // ETH
                     string secondCode = Forex.Forex.CurrencyPairDual(mainPair, this.Symbol);
 
-                    Log.Trace($"Found second code {secondCode} in mainPair {mainPair}");
-
                     bool mainInvert = mainPair.IndexOf(this.Symbol) != 0;
 
                     // ETHUSD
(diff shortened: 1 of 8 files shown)


StefanoRaggi:
@viliwonka 
Thanks for expanding the symbol length, I think there is still an problem with  `AddCrypto("SENTETH")` though:

The `Cash.ConversionRate` property is updated when receiving data for the internal feed and represents the conversion rate from the cash symbol (SENT) to the **account currency** (for now always USD);

If `SENTUSD` or `USDSENT` pairs are not available, we probably need an **indirect conversion**, calculating the SENT->USD conversion rate using prices from these two pairs (`ETHUSD` and `SENTETH`),
that should be added both as internal feeds by `EnsureCurrencyDataFeed` (as you did in this PR 👍)

Perhaps `Cash.ConversionRateSecurity` may need to become a `List<Security>` ?

viliwonka:
@StefanoRaggi 
Hi back! Thanks for review.

Yes, I had suspicion there should be conversion done for Cashbook.AccountCurrency-less pairs somewhere.

Any idea where the code for indirect conversion should be located & modified/added?

And yes, I had same idea about Cash.ConversionRateSecurity turning into List<Security> type, would make sense in multi-path conversion.

viliwonka:
@StefanoRaggi 
Hey, sorry for late pinging, but there it is - conversion to list.

mchandschuh:
Please remove the recent changes for maintaining a list of conversion rate securities. I'd like to merge the crypto symbol length changes and we can work on designing multi-legged conversion rates in a separate PR as a new feature.

viliwonka:
@mchandschuh Ok, done!

viliwonka:
Travis build has failed, but it's due to some keyserver error, not actual compile errors / unit test failures.

mchandschuh:
@viliwonka -- I think you may have reverted the wrong change. I was referring to the changes that change the conversion rate security into a list of conversion rate securities. I know this will crypto pairs that aren't immediately convertible into the account currency inoperable, but supporting multi-legged conversion is something we need to think carefully about -- my biggest concern is that there are often multiple paths. I think the multi-legged currency conversion logic should be handled in a separate feature PR. This PR should stay focused on supporting variable length crypto symbols.

viliwonka:
@mchandschuh :man_facepalming:  Yeah, I knew exactly what you meant, I mis-understood the "revert commit" part, I thought it would rollback through history. My bad, I created a nice mess here.

Common/Securities/Cash.cs
```diff
@@ -251,6 +252,74 @@ SecurityChanges changes
...
                 }
+
+
+            }
+
+
+            // No direct conversion rate pair is found, check for secondary conversion pair
+            // Common to cryptocurrencies where there are no direct pairings with USD, but there is intermediary such as BTC or ETH
+            // Example #1: RENUSD doesn't exist, but there is RENETH  and ETHUSD,  from which we can calculate RENUSD
+            // Example #2: RENUSD doesn't exist, but there is RENUSDT and USDTUSD, from which we can calculate RENUSD
```
mchandschuh: Looks like the code below (and this comment) are attempting to solve the indirect (multi-legged) conversion case. Please remove.

Common/Securities/Forex/Forex.cs
```diff
@@ -170,16 +171,44 @@ public static void DecomposeCurrencyPair(string currencyPair, out string baseCur
...
+            string CurrencyB = null;
+
+            DecomposeCurrencyPair(currencyPair, out CurrencyA, out CurrencyB);
+
+            if (CurrencyA == knownSymbol)
+            {
+                return CurrencyB;
+            }
+            else
+            if (CurrencyB == knownSymbol)
```
mchandschuh: code style - please keep `else if` on a single line.

viliwonka: Ok, on it fixing both.

viliwonka:
@mchandschuh Here, hope all is well. Don't forget to make branch for indirect conversion :).

mchandschuh:
For the next phase:
Branch: [feature-2060-multi-leg-currency-conversion](https://github.com/QuantConnect/Lean/tree/feature-2060-multi-leg-currency-conversion)
Issue: #2060 Support indirect/multi-leg currency conversions ( A->{Bridge Currencies}->B

PR #2024, opened 2018-05-21
Fixes EqualWeightingPortfolioConstructionModel logic

### LINKED ISSUE #2023: Review EqualWeightingPortfolioConstructionModel logic
#### Expected Behavior
Equal weighting portfolio construction models (C# and Python) allocate all cash to the stocks who have insights in universe.

#### Actual Behavior
Equal weighting portfolio construction models (C# and Python) allocate only the share that each symbol when a new insight from that algorithm arrives. 

#### Potential Solution
Review the logic.

#### Checklist

AlexCatarino:
#### Description
`EqualEeightingPortfolioConstructionModel` (C# and Python) allocates all cash to the stocks who have insights in universe.
- Fixes regression tests to reflect the model logic change
- Fixes imports in python algorithms to use python models when available

#### Related Issue
Closes #2023 

#### Motivation and Context
Keep reasonable logic in our example models.

No.

Regression tests and manually verifying that the model behaves accordingly to the designed logic. For instance, when three insights arrive at difference time steps, all cash will be allocated to the symbol (S1) of that insight, then 50% of equity will be allocated to second symbol (S2) and 50% to S1 and finally 33% of equity will be allocated to the third symbol, 33% to S2 and 33% to S1. 

#### Types of changes

#### Checklist:


jaredbroad:
Almost there- I think the criteria for including the insight should also be that the insight has not expired. e.g If the insight prediction ended 3 months ago its not relevant for the portfolio today.

We should record the end-time for an insight (when its created) and use it to determine who should still be in the portfolio.

Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.cs
```diff
@@ -48,16 +50,24 @@ public override IEnumerable<IPortfolioTarget> CreateTargets(QCAlgorithmFramework
...
             }
 
-            if (_securities.Count == 0)
+            if (insights.Length == 0)
             {
                 return Enumerable.Empty<IPortfolioTarget>();
             }
+            
+            // Get symbols that have emit insights and still in the universe
+            var symbols = _insightCollection.Select(x => x.Symbol).Distinct().ToList();
```
mchandschuh: will this contain any symbols from `_removedSymbols` ? I'm thinking that removed symbols shouldn't count towards the denominator of equal weighting. Also, this has approach may have unexpected consequences when considering alpha models that don't emit all insights at one time (such an an EMA cross). For example, we have a universe with 500 securities and we get our first EMA cross. With this logic, we'll give that security 100% weighting. 20 minutes later our next security can experience an EMA cross and then we'll give him 50%, and so on. This was the original reason this looks at the universe size for the denominator instead of the count of unique securities that have had insights generated.

mchandschuh:
Regarding expired insights, can use `Insight.CloseTimeUtc` to determine when an insight's period has officially ended. This time isn't simply the insight's generated time + the period, but instead attempts to take into account market hours and weekends.

PR #2032, opened 2018-05-24
Add TimeInForce.GoodTilDate (GTD) support in backtesting and IB brokerage

StefanoRaggi:
#### Description
This PR includes some refactoring to make it easier to add new time in forces (with custom settings, such as `Expiry` property in `GoodTilDate`):

- Renamed `TimeInForce` enum to `TimeInForceType`
- Added new `TimeInForce` abstract class
- Removed `ITimeInForceHandler` implementations, existing code moved to the derived `TimeInForce` classes
- Removed `Order.DurationValue` property, replaced by `GoodTilDateTimeInForce.Expiry`
- `BacktestingBrokerage` no longer requires a dictionary of time in force handlers, it now directly calls methods on `TimeInForce` objects

Unit tests have been added and example usage has been added to the existing `TimeInForceAlgorithm`.

This feature is currently supported only in backtesting and with Interactive Brokers brokerage.

#### Related Issue
#1093

#### Motivation and Context
Good Til Date (GTD) time in force was not currently supported

No.

New unit tests, updated regression test and IB live testing.

#### Types of changes

#### Checklist:


Algorithm.CSharp/TimeInForceAlgorithm.cs
```diff
@@ -70,6 +73,15 @@ public override void OnData(Slice data)
                 DefaultOrderProperties.TimeInForce = TimeInForce.Day;
                 _dayOrderTicket = LimitOrder(_symbol, 10, 160m);
             }
+
+            if (_gtdOrderTicket == null)
+            {
+                // This order will expire on October 10th at market close,
+                // if not filled by then it will be canceled automatically.
+
+                DefaultOrderProperties.TimeInForce = new GoodTilDateTimeInForce(new DateTime(2013, 10, 10));
```
mchandschuh: I'm wondering if it makes more sense to define `GoodTilDateTimeInForce` using a `TimeSpan` instead of an actual `DateTime`. Using a time span I think is more natural and also doesn't require it to be updated every day. Also, from an implementation standpoint it's easily supported since the order object has the submit time on it, so can easily compute the date we should cancel. Thoughts?

mchandschuh: Also, since this is under regression test, maybe we should have a case where GTD is cancelled _and_ GTD gets executed. Also, let's confirm that other TIF have both sides exercised in the regression algorithm, otherwise we're only testing half of the implementation via regression.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1917,6 +1942,28 @@ private static TimeInForce ConvertTimeInForce(string timeInForce)
...
+            // NOTE: we currently ignore the time zone in this method for a couple of reasons:
+            // - TZ abbreviations are ambiguous and unparsable to a unique time zone
+            //   see this article for more info:
+            //   https://codeblog.jonskeet.uk/2015/05/05/common-mistakes-in-datetime-formatting-and-parsing/
+            // - IB seems to also have issues with Daylight Saving Time zones
+            //   Example: an order submitted from Europe with GoodTillDate property set to "20180524 21:00:00 UTC"
+            //   when reading the open orders, the same property will have this value: "20180524 23:00:00 CET"
+            //   which is incorrect: should be CEST (UTC+2) instead of CET (UTC+1)
+
+            // We can ignore this issue, because the method is only called by GetOpenOrders.
```
mchandschuh: great comment above about IB's mishandling of time zones, perhaps add an additional blurb to the bottom line here. _why_ is it OK that it's only called by `GetOpenOrders`? -- my hunch is that it's because we only call `GetOpenOrders` during live trading, which means we won't be simulating TIF and instead will rely on brokerages to apply TIF properly.

Common/Orders/OrderJsonConverter.cs
```diff
@@ -194,5 +194,41 @@ private static Order CreateOrder(OrderType orderType, JObject jObject)
...
+                {
+                    var expiry = jObject["DurationValue"].Value<DateTime>();
+                    return new GoodTilDateTimeInForce(expiry);
+                }
+
+                return (TimeInForceType)value;
+            }
+
+            var timeInForceType = (TimeInForceType)timeInForce["Type"].Value<int>();
+            switch (timeInForceType)
```
mchandschuh: maybe this switch statement can be removed/simplified by ensuring that we're serializing the `$type` to JSON. This will reduce coupling within the system and prevent potential errors from changes/additions to the TIF implementations.

Common/Orders/TimeInForce.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// </summary>
+        public static implicit operator TimeInForce(TimeInForceType type)
+        {
+            switch (type)
+            {
+                case TimeInForceType.GoodTilCanceled:
+                    return new GoodTilCanceledTimeInForce();
+
+                case TimeInForceType.Day:
+                    return new DayTimeInForce();
```
mchandschuh: Instead of creating a new object, can we use the existing `static readonly` instances already defined? IIRC, all of these supported implementations are stateless.

Common/Orders/TimeInForce.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <param name="left">The left operand</param>
+        /// <param name="right">The right operand</param>
+        /// <returns>True if both symbols are equal, otherwise false</returns>
+        public static bool operator ==(TimeInForce left, TimeInForce right)
+        {
+            if (ReferenceEquals(left, right)) return true;
+            if (ReferenceEquals(left, null)) return false;
+            if (ReferenceEquals(right, null)) return false;
+
+            return left.Type == right.Type;
```
mchandschuh: This isn't 100% correct. Consider two GTD implementations with different end dates (or different time spans if we make that change). Best bet is to `return left.Equals(right)` and let the implementations decide how to resolve equality. Similar comment for all equality methods in this class.

Common/Orders/OrderTypes.cs
```diff
@@ -56,11 +56,10 @@ public enum OrderType
         OptionExercise
     }
 
-
     /// <summary>
-    /// Time In Force - defines the length of time over which an order will continue working before it is canceled
+    /// Time In Force Type - defines the length of time over which an order will continue working before it is canceled
     /// </summary>
-    public enum TimeInForce
+    public enum TimeInForceType
```
mchandschuh: Is this enum still required? I think the only place it's used anymore is type checks which can be replaced by actual type checks vs enum value checks. The other place it's currently used is the implicit operator, but I don't think that one is required since users can use `TimeInForce.Day` instead of `TimeInForceType.Day` -- the concern is having two very similar things that are potentially different.

Tests/Common/Orders/TimeInForces/TimeInForceTests.cs
```diff
@@ -0,0 +1,560 @@
...
+
+        [Test]
+        public void GtdTimeInForceCompareProperly()
+        {
+            var tif1 = new GoodTilDateTimeInForce(new DateTime(2018, 5, 21));
+            var tif2 = new GoodTilDateTimeInForce(new DateTime(2018, 5, 21));
+            var tif3 = new GoodTilDateTimeInForce(new DateTime(2018, 6, 6));
+
+            Assert.IsTrue(tif1 == tif2);
+            Assert.IsFalse(tif1 != tif2);
```
mchandschuh: I suspect this test will unexpectedly fail if changed to:
``` csharp
TimeInForce tif1 = new GoodTilDateTimeInForce(new DateTime(2018, 5, 21));
TimeInForce tif2 = new GoodTilDateTimeInForce(new DateTime(2018, 5, 21));
TimeInForce tif3 = new GoodTilDateTimeInForce(new DateTime(2018, 6, 6));

Assert.IsTrue(tif1 == tif2);
Assert.IsFalse(tif1 != tif2);
```

This is because the static equals operator does not support polymorphic/dynamic dispatch, so it will bind to the equals operator defined in `TimeInForce` (the one statically known) instead of the one defined in `GoodTilDateTimeInForce` -- see comments in `TimeInForce` regarding equality implementation.

Common/Orders/TimeInForce.cs
```diff
@@ -0,0 +1,159 @@
...
+            return other != null && Type.Equals(other.Type);
+        }
+
+        /// <summary>
+        /// Serves as a hash function for a particular type.
+        /// </summary>
+        /// <returns>A hash code for the current <see cref="TimeInForce"/>.</returns>
+        public override int GetHashCode()
+        {
+            return Type.GetHashCode();
```
mchandschuh: The more I think about equality implementations the more I think these should be defined as abstract or not defined at all in this class (`Equals` and `GetHashCode`) and the operators `==` and `!=` should pipe through to the `Equals` method to enable proper dynamic dispatch. Thoughts?

Algorithm.CSharp/TimeInForceAlgorithm.cs
mchandschuh: Another reason here, I'm thinking of eventually supporting syntax like:
``` csharp
DefaultOrderProperties.TimeInForce = TimeSpan.FromDays(10);
```
via an implicit operator which would require the data held to be a time span value instead of an absolute date time value.

Common/Orders/OrderTypes.cs
StefanoRaggi: 👍  Agree, the `TimeInForceType` enum is not really needed any more -- I'll remove it

Common/Orders/TimeInForce.cs
StefanoRaggi: The idea here was to have a common default implementation of equality methods (to avoid duplicating these methods in derived classes with no additional properties -- such as GTC and DAY) , but since I'm going to remove the `TimeInForceType Type` property, this will be changed to `left.Equals(right)`.

Common/Orders/TimeInForce.cs
StefanoRaggi: Sure, this was an oversight

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: > it's because we only call GetOpenOrders during live trading, which means we won't be simulating TIF and instead will rely on brokerages to apply TIF properly.
This is exactly the reason -- I'll add this comment to make it more clear

Common/Orders/OrderJsonConverter.cs
StefanoRaggi: I'll try to handle serialization/deserialization with `$type`, although we'll still need a switch on a few numeric values (0,1,2) for backwards-compatibility.

Common/Orders/TimeInForce.cs
StefanoRaggi: Agree on the abstract equality methods and dynamic dispatch, if possible I'd try leaving the `==` and `!=` in the base class (to avoid duplication)

Tests/Common/Orders/TimeInForces/TimeInForceTests.cs
StefanoRaggi: Aren't these asserts currently passing in the PR?
All equality methods and operators have been overridden in the `GoodTilDateTimeInForce` -- will be refactored shortly

Algorithm.CSharp/TimeInForceAlgorithm.cs
StefanoRaggi: The only issue we would have changing `GTD` from `DateTime Expiry` to `TimeSpan Duration` is that not all brokerage API have the order submission time available, so we would not be able to get back the `Duration` value in `GetOpenOrders`.

IB and FXCM do not have order creation/submission time in their get order API methods while OANDA and GDAX have this field.

IB, FXCM and OANDA have the expiry date/time field, GDAX doesn't have it but in its `GTT` time in force only allows fixed durations of 1 minute, 1 hour or 1 day.

Maybe we need to have both properties?

PR #2057, opened 2018-05-31
Add price and split factors to CoarseFundamental class

### LINKED ISSUE #2055: Add price and split factors to CoarseFundamental class
Currently the `CoarseFundamental.Price` property is `Raw` and this causes indicator problems with splits and dividends.

To solve the issue, we should add both price and split factors to the coarse fundamental `CSV` data files and in the `CoarseFundamental` class, so the adjusted price can be easily calculated.

These properties should be added to `CoarseFundamental`:
``` C#
        /// <summary>
        /// Gets the price factor for the given date
        /// </summary>
        public decimal PriceFactor { get; set; }

        /// <summary>
        /// Gets the split factor for the given date
        /// </summary>
        public decimal SplitFactor { get; set; }

        /// <summary>
        /// Gets the combined factor used to create adjusted prices from raw prices
        /// </summary>
        public decimal PriceScaleFactor => PriceFactor * SplitFactor;

        /// <summary>
        /// Gets the split and dividend adjusted price
        /// </summary>
        public decimal AdjustedPrice => Price * PriceScaleFactor;
```

#### Reproducing the Problem
This is a simple daily universe algorithm with a price jump on the AAPL split:
``` C#
    {
        private static readonly Symbol Apple = QuantConnect.Symbol.Create("AAPL", SecurityType.Equity, Market.USA);

        {

            SetStartDate(2014, 06, 5);
            SetEndDate(2014, 06, 11);
            SetCash(50000);

            AddUniverse(coarse =>
            {
                var cf = coarse.FirstOrDefault(x => x.Symbol == Apple);
                if (cf != null)
                {
                    Console.WriteLine($"Symbol:{cf.Symbol.Value} EndTime:{cf.EndTime} RawPrice:{cf.Price}");
                }

                return new List<Symbol> {Apple};
            });
        }
    }
```

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- Updated Coarse Generator to save two extra columns in the output CSV files: 
`PriceFactor` and `SplitFactor`
- Added new properties to `CoarseFundamental`: 
`PriceFactor`, `SplitFactor`, `PriceScaleFactor` and `AdjustedPrice`
- Added new regression test algorithm: `CoarseUniverseSplitRegressionAlgorithm`

#### Related Issue
Closes #2055 

#### Motivation and Context
Currently only the raw (not adjusted for splits and dividends) price is included in coarse data, so it is impossible to handle splits and dividends properly in coarse universes.

No.

- Full run of Coarse generator with all daily data
- New regression test for AAPL split (June 2014)

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/CoarseUniverseGenerator/Program.cs
+++ b/ToolBox/CoarseUniverseGenerator/Program.cs
@@ -172,6 +172,9 @@ namespace QuantConnect.ToolBox.CoarseUniverseGenerator
             }
             var fineFundamentalFolder = Path.Combine(fundamentalDirectoryInfo.FullName, "fine");
 
+            var mapFileProvider = new LocalDiskMapFileProvider();
+            var factorFileProvider = new LocalDiskFactorFileProvider(mapFileProvider);
+
             // open up each daily file to get the values and append to the daily coarse files
             foreach (var file in Directory.EnumerateFiles(dailyFolder))
             {
@@ -264,8 +267,15 @@ namespace QuantConnect.ToolBox.CoarseUniverseGenerator
                             // check if symbol has fine fundamental data for the current date
                             var hasFundamentalDataForDate = date >= firstFineSymbolDate;
 
-                            // sid,symbol,close,volume,dollar volume,has fundamental data
-                            var coarseFileLine = sid + "," + symbol + "," + close + "," + volume + "," + Math.Truncate(dollarVolume) + "," + hasFundamentalDataForDate;
+                            // get price and split factors from factor files
+                            var leanSymbol = new Symbol(sid, symbol);
+                            var factorFile = factorFileProvider.Get(leanSymbol);
+                            var factorFileRow = factorFile?.GetScalingFactors(date);
+                            var priceFactor = factorFileRow?.PriceFactor ?? 1m;
+                            var splitFactor = factorFileRow?.SplitFactor ?? 1m;
+
+                            // sid,symbol,close,volume,dollar volume,has fundamental data,price factor,split factor
+                            var coarseFileLine = $"{sid},{symbol},{close},{volume},{Math.Truncate(dollarVolume)},{hasFundamentalDataForDate},{priceFactor},{splitFactor}";
 
                             StreamWriter writer;
                             if (!writers.TryGetValue(coarseFile, out writer))
--- a/ToolBox/QuantConnect.ToolBox.csproj
+++ b/ToolBox/QuantConnect.ToolBox.csproj
@@ -36,7 +36,7 @@
     <LangVersion>6</LangVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <StartupObject>QuantConnect.ToolBox.KaikoDataConverter.Program</StartupObject>
+    <StartupObject>QuantConnect.ToolBox.CoarseUniverseGenerator.Program</StartupObject>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="DotNetZip, Version=1.10.1.0, Culture=neutral, PublicKeyToken=6583c7c814667745, processorArchitecture=MSIL">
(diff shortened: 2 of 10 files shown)


ToolBox/CoarseUniverseGenerator/Program.cs
```diff
@@ -264,8 +267,15 @@ public static ICollection<string> ProcessDailyFolder(string dailyFolder, string
...
                             var hasFundamentalDataForDate = date >= firstFineSymbolDate;
 
-                            // sid,symbol,close,volume,dollar volume,has fundamental data
-                            var coarseFileLine = sid + "," + symbol + "," + close + "," + volume + "," + Math.Truncate(dollarVolume) + "," + hasFundamentalDataForDate;
+                            // get price and split factors from factor files
+                            var leanSymbol = new Symbol(sid, symbol);
+                            var factorFile = factorFileProvider.Get(leanSymbol);
+                            var factorFileRow = factorFile?.GetScalingFactors(date);
+                            var priceFactor = factorFileRow?.PriceFactor ?? 1m;
+                            var splitFactor = factorFileRow?.SplitFactor ?? 1m;
```
mchandschuh: gotta love the null coalescing operator! ( `?.` )

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -36,7 +36,7 @@
     <LangVersion>6</LangVersion>
   </PropertyGroup>
   <PropertyGroup>
-    <StartupObject>QuantConnect.ToolBox.KaikoDataConverter.Program</StartupObject>
+    <StartupObject>QuantConnect.ToolBox.CoarseUniverseGenerator.Program</StartupObject>
```
mchandschuh: Not sure if this change was intentional or matters, don't think we have a 'default' startup program

Algorithm.CSharp/CoarseUniverseSplitRegressionAlgorithm.cs
```diff
@@ -0,0 +1,110 @@
...
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This is a regression algorithm to test the price and split factors in coarse fundamental data
+    /// </summary>
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="universes" />
+    /// <meta name="tag" content="coarse universes" />
+    /// <meta name="tag" content="regression test" />
+    public class CoarseUniverseSplitRegressionAlgorithm : QCAlgorithm
```
mchandschuh: Wondering if a full regression algorithm is a little heavy here. It seems most of this test is confirming that we're able to parse out these new values and could be written as a unit test against the `CoarseFundamental` type using string literals for input lines to be parsed. The nice part about going the unit test route is that they're super fast and fail travis.

thoughts?

ToolBox/QuantConnect.ToolBox.csproj
StefanoRaggi: `QuantConnect.ToolBox` is the only project in the LEAN solution with multiple `Main` methods (programs) and VS stores the currently selected startup object/program in the `StartupObject` tag of the `csproj` file. 

Although there is a `(Not set)` option in the VS project properties for this tag, selecting it will cause this build error:
```
error CS0017: Program has more than one entry point defined. Compile with /main to specify the type that contains the entry point.
```
I think in the past we have always set this value to the last updated program in the project.

To get rid of this issue, I think we should move each program to its own project.
This would also be a nice thing, every program would not be compiled to the same `QuantConnect.ToolBox.exe` filename :)

jaredbroad: Or Toolbox could have a master project which can call the other more like class libraries; removing all main methods and concentrating it into one. 

`toolbox.exe --app=YahooDownloader SYMBOLS RESOLUTION FROMDATE TODATE`

StefanoRaggi: New issue #2065 for this

PR #2102, opened 2018-06-11
Improves Extensions.TryConvert to avoid test by exception

### LINKED ISSUE #2091: Indicator consolidator throws error in python algorithm
#### Expected Behavior

Python consolidator should work with indicator 
#### Actual Behavior

Using indicator in the consolidator throws the error 

`During the algorithm initialization, the following exception has occurred: SystemError : <bound method 'RegisterIndicator'> returned a result with an error set
  at Initialize` 

#### Reproducing the Problem

```from datetime import timedelta
    def Initialize(self):      
        self.SetStartDate(2013,10,07)  
        self.SetEndDate(2013,10,11)    
        self.SetCash(100000)           
        self.AddForex("EURUSD", Resolution.Minute)

        fourHourConsolidator = QuoteBarConsolidator(timedelta(minutes=240))
        fourHourConsolidator.DataConsolidated += self.FourHourBarHandler
        self.SubscriptionManager.AddConsolidator("EURUSD", fourHourConsolidator)
        
        self.fourHourIchimoku = IchimokuKinkoHyo("EURUSD", 9, 26, 26, 52, 26, 26)
        self.RegisterIndicator("EURUSD", self.fourHourIchimoku, fourHourConsolidator, None)

    def OnData(self, data):
            self.SetHoldings("EURUSD", 1)
           
    def FourHourBarHandler(self, sender, bar):
        self.Log(str(self.fourHourIchimoku.Current))                        
```
#### Checklist

AlexCatarino:
#### Description
Instead of letting `PyObject.AsManagedObject` throw an exception because the types do not match, we check whether the target type is assignable from the python object type.

#### Related Issue
Closes #2091 

No.

Algorithm provided in #2091.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -1050,9 +1050,23 @@ namespace QuantConnect
             {
                 try
                 {
+                    if (typeof(T) == typeof(Type))
+                    {
+                        result = pyObject.As<Type>() as T;
+                        return true;
+                    }
+
+                    var pythonType = pyObject.GetPythonType();
+                    var csharpType = pythonType.As<Type>();
+
+                    if (!typeof(T).IsAssignableFrom(csharpType))
+                    {
+                        return false;
+                    }
+
                     result = pyObject.AsManagedObject(typeof(T)) as T;
 
-                    if (typeof(T) == typeof(Type) || result is IEnumerable)
+                    if (result is IEnumerable)
                     {
                         return true;
                     }
@@ -1060,7 +1074,7 @@ namespace QuantConnect
                     // If the PyObject type and the managed object names are the same,
                     // pyObject is a C# object wrapped in PyObject, in this case return true
                     // Otherwise, pyObject is a python object that subclass a C# class.
-                    string name = (pyObject.GetPythonType() as dynamic).__name__;
+                    string name = (pythonType as dynamic).__name__;
                     return name == result.GetType().Name;
                 }
                 catch


mchandschuh:
Do we need some unit tests to cover this guy?

disruptek:
Is it really safe enough to merely compare the `__name__` when what we're trying to test for is identity?

AlexCatarino:
@mchandschuh We have [unit tests](https://github.com/QuantConnect/Lean/blob/master/Tests/Algorithm/AlgorithmRegisterIndicatorTests.cs#L30) that were and are passing.
@disruptek, this code is meant to tackle particular case: when a python class inherit from a C# class, "its" type is the C# type when, in fact, it is a python type.

disruptek:
I get it, but the `__name__` property is hardly immutable or unique...

mchandschuh:
@disruptek -- is there an alternative solution you see here?

Common/Extensions.cs
```diff
@@ -1050,17 +1050,31 @@ public static bool TryConvert<T>(this PyObject pyObject, out T result)
             {
                 try
                 {
+                    if (typeof(T) == typeof(Type))
+                    {
+                        result = pyObject.As<Type>() as T;
+                        return true;
+                    }
+
+                    var pythonType = pyObject.GetPythonType();
```
disruptek: This call constitutes a round-trip into Python to get the type as a PyObject which we don't even use in the following two tests.

I still maintain that the code makes no sense, but at least it could be a little faster at calculating the wrong answer.

PR #2120, opened 2018-06-15
Fix non-deterministic query in QCAlgorithm.GetSubscription

### LINKED ISSUE #2119: BasicTemplateCryptoAlgorithm regression failing
#### Expected Behavior
`BasicTemplateCryptoAlgorithm` backtest should always return the same results.

#### Actual Behavior
`BasicTemplateCryptoAlgorithm` returns different results for a backtest with the Test runner vs LEAN launcher.

#### Reproducing the Problem
Run the `BasicTemplateCryptoAlgorithm` with LEAN launcher: **10 trades** (correct)
Run the `BasicTemplateCryptoAlgorithm` with Test runner: **5 trades** (incorrect)

#### System Information
Windows 10

#### Checklist

StefanoRaggi:
#### Description
`QCAlgorithm.GetSubscription` now searches for the first matching subscription in an enumerable sorted by `Symbol` and `TickType`.

#### Related Issue
Closes #2119 

#### Motivation and Context
`BasicTemplateCryptoAlgorithm` is returning different results for a backtest with the Test runner vs LEAN launcher.

No.

`BasicTemplateCryptoAlgorithm` is now consistently returning the same results.
Previously, when running in the test runner, the last 5 orders were never submitted because the EMA values were always zero (due to the wrong consolidators being selected).

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -1332,12 +1332,15 @@ namespace QuantConnect.Algorithm
             SubscriptionDataConfig subscription;
             try
             {
+                // deterministic ordering is required here
+                var subscriptions = SubscriptionManager.Subscriptions.OrderBy(x => x.TickType);
+
                 // find our subscription to this symbol
-                subscription = SubscriptionManager.Subscriptions.FirstOrDefault(x => x.Symbol == symbol && (tickType == null || tickType == x.TickType));
+                subscription = subscriptions.FirstOrDefault(x => x.Symbol == symbol && (tickType == null || tickType == x.TickType));
                 if (subscription == null)
                 {
                     // if we can't locate the exact subscription by tick type just grab the first one we find
-                    subscription = SubscriptionManager.Subscriptions.First(x => x.Symbol == symbol);
+                    subscription = subscriptions.First(x => x.Symbol == symbol);
                 }
             }
             catch (InvalidOperationException)


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1332,12 +1332,17 @@ public SubscriptionDataConfig GetSubscription(Symbol symbol, TickType? tickType)
             SubscriptionDataConfig subscription;
             try
             {
+                // deterministic ordering is required here
+                var subscriptions = SubscriptionManager.Subscriptions
+                    .OrderBy(x => x.Symbol)
+                    .ThenBy(x => x.TickType);
```
mchandschuh: since we're filtering by `symbol` below, I don't think we need to perform an order by based on symbol. Please test removing the order by symbol to see if it still works as expected

StefanoRaggi: You're right, order by symbol is not needed. Updated PR

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1332,12 +1332,15 @@ public SubscriptionDataConfig GetSubscription(Symbol symbol, TickType? tickType)
             SubscriptionDataConfig subscription;
             try
             {
+                // deterministic ordering is required here
+                var subscriptions = SubscriptionManager.Subscriptions.OrderBy(x => x.TickType);
```
Martin-Molinero: I wonder if it would be more performant to filter first by (x => x.Symbol == symbol). I image it will probably depend on each case.

jaredbroad:
This was a bad one. Do we know how it arrived & (other than regression) how to prevent it in the future?

Issue #2065, opened 2018-06-01
Make Toolbox single project with application as command line argument


StefanoRaggi:
#### Expected Behavior
We expect to be able to run any toolbox application from the same single executable file.

#### Actual Behavior
`QuantConnect.ToolBox` is the only project in the LEAN solution with multiple `Main` methods (programs) and VS stores the currently selected startup object/program in the `StartupObject` tag of the `csproj` file.

Although there is a `(Not set)` option in the VS project properties for this tag, selecting it will cause this build error:
```
error CS0017: Program has more than one entry point defined. Compile with /main to specify the type that contains the entry point.
```

#### Potential Solution
1. Toolbox could have a master project which can call the other more like class libraries; removing all main methods and concentrating it into one.
```
toolbox.exe --app=YahooDownloader SYMBOLS RESOLUTION FROMDATE TODATE
```

2. Another potential solution could be to have one project per toolbox application, each one with its own different executable file name.

#### Reproducing the Problem
If we need to run a different application from the one currently selected as _Startup Object_, we currently need to change the setting and rebuild the project.

#### System Information
n/a

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #2166, opened 2018-06-27
Make toolbox a single application wrapping all the tools

Martin-Molinero:
#### Description

- Decoupled the Lean command line argument parser into the new `ApplicationParser.cs`, so it can be reused at the new `ToolboxArgumentParser.cs`
- Adding ` CommandLineOption.cs` class that will keep information about a specific command line option
- Wrapped the 17 tools under `ToolBox/Program.cs`. Shared/unified as many parameters as possible, they can be found at the new `Configuration/ToolboxArgumentParser.cs` with their corresponding help/description.
- Modified each tool as least as possible, removing argument parsing and now receiving parameters directly.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2065
#### Motivation and Context

Simplify the usage of the different tools

N/A

Manually
#### Types of changes

#### Checklist:


Configuration/CommandLineOption.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Command option type
+        /// </summary>
+        public CommandOptionType Type { get; }
+
+        /// <summary>
+        /// Command option description
+        /// </summary>
+        public string Description { get; }
```
mchandschuh: no change required - General style of the project places properties at the top. I *think* it's generally like this:

private mutable state
private state
(generally should *never* have public fields, but they do occur sometimes in the code-base)
public properties
ctors (ordered by accessibility)
public methods
abstract methods
protected methods
private methods

This will be included in a R# `dotSettings` file *one-day* when I get some free time

ToolBox/Program.cs
```diff
@@ -0,0 +1,190 @@
...
+                        PrintMessageAndExit("ERROR: Unrecognized --app value");
+                        break;
+                }
+            }
+        }
+        private static IList<string> DictionaryObjectToList(object dictionaryObject)
+        {
+            IList<string> list = new List<string>();
+            var dictionary = dictionaryObject as Dictionary<string, object>;
+            foreach (var item in dictionary)
```
mchandschuh: no change required, but perhaps much more simply, `return dictionary.Keys.ToList()`

ToolBox/Program.cs
```diff
@@ -0,0 +1,190 @@
...
+        private static void PrintMessageAndExit(string message = "")
+        {
+            if (!message.IsNullOrEmpty())
+            {
+                Console.WriteLine("\n" + message);
+                Console.WriteLine("Use the '--h' parameter for more information");
+            }
+            Console.WriteLine("\nPress any key to quit");
+            Console.ReadLine();
+            Environment.Exit(0);
```
mchandschuh: If this is an error condition we should return a non-zero exit code to aid any potential automation use-cases

ToolBox/Program.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        private static string GetParameterOrExit(IReadOnlyDictionary<string, object> optionsObject, string parameter)
+        {
+            if (!optionsObject.ContainsKey(parameter))
+            {
+                PrintMessageAndExit("ERROR: REQUIRED parameter --" + parameter + "= is missing");
```
mchandschuh: This one is _definitely_ an error case and should end up with a non-zero exit code.

Configuration/CommandLineOption.cs
Martin-Molinero: I'll change it ;)

ToolBox/Program.cs
Martin-Molinero: Your right! I'll just lose this method and use `dictionary.Keys.ToList()` approach

ToolBox/Program.cs
Martin-Molinero: I'll add `exitCode` as a parameter for `PrintMessageAndExit`

mchandschuh:
Awesome stuff, thanks @Martin-Molinero!

PR #2176, opened 2018-06-28
Fixes ambigous call bug for Download method

### LINKED ISSUE #2030: Download() ambiguous with a single argument
#### Expected Behavior

`Download("urlstring")` should work in C# as in the [docs example](https://www.quantconnect.com/docs/algorithm-reference/importing-custom-data#Importing-Custom-Data-Loading-Reference-Data)

#### Actual Behavior

ERROR: the call is ambiguous, 2nd argument is required to discern between IEnumerable and PyObject.

#### Potential Solution

Make Python method have a more specific signature?

We could of course pass some unimportant header as a part of the second argument, but it seems uncouth to instantiate a dummy enumerable.

#### Reproducing the Problem

`Download("urlstring")` in any basic C# algorithm, then try compiling/building an executable.

#### System Information

Mono on Ubuntu Linux, but that should not matter.

#### Checklist

AlexCatarino:
#### Description
Removes the `QCAlgorithm.Download` method that accepts a `PyObject` and modify the remaining overload to accept a string. With this setup, this method can be called by both languages.

#### Related Issue
Closes #2030 

No.

Before this change, C# QCAlgorithm would not compile if this method was included.
Test manually the method win C# and python with and without setting the second argument.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -2056,11 +2056,28 @@ namespace QuantConnect.Algorithm
         /// The resource to download is specified as a <see cref="string"/> containing the URI.
         /// </summary>
         /// <param name="address">A string containing the URI to download</param>
+        /// <returns>The requested resource as a <see cref="string"/></returns>
+        public string Download(string address) => Download(address, Enumerable.Empty<KeyValuePair<string, string>>());
+
+        /// <summary>
+        /// Downloads the requested resource as a <see cref="string"/>.
+        /// The resource to download is specified as a <see cref="string"/> containing the URI.
+        /// </summary>
+        /// <param name="address">A string containing the URI to download</param>
+        /// <param name="headers">Defines header values to add to the request</param>
+        /// <returns>The requested resource as a <see cref="string"/></returns>
+        public string Download(string address, IEnumerable<KeyValuePair<string, string>> headers) => Download(address, headers, null, null);
+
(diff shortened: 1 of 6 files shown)


jaredbroad:
Can you please make 2 example algorithms (C# and Py) using the Download() to download a CSV file; parse it and using it in the user's algorithm as say a list of assets to assign equal weight?

Possibly with the header arg specified to give a complex example. 

Should / can this be covered with a test?

Algorithm/QCAlgorithm.cs
```diff
@@ -2056,18 +2057,18 @@ public void SetStatus(AlgorithmStatus status)
         /// The resource to download is specified as a <see cref="string"/> containing the URI.
         /// </summary>
         /// <param name="address">A string containing the URI to download</param>
-        /// <param name="headers">Defines header values to add to the request</param>
+        /// <param name="headers">Defines header values to add to the request (json format)</param>
         /// <param name="userName">The user name associated with the credentials</param>
         /// <param name="password">The password for the user name associated with the credentials</param>
         /// <returns>The requested resource as a <see cref="string"/></returns>
-        public string Download(string address, IEnumerable<KeyValuePair<string, string>> headers = null, string userName = null, string password = null)
+        public string Download(string address, string headers = null, string userName = null, string password = null)
```
mchandschuh: Isn't this a breaking change?

PR #2190, opened 2018-07-02
Parallelize reading of fine fundamental data

### LINKED ISSUE #2189: Parallelize reading of fine fundamental data
#### Expected Behavior

Universes that use the fine fundamental data should not be exponentially slower than w/out fine data.

#### Actual Behavior

Fine universe algorithms are incredibly slow.

#### Potential Solution

Read fine fundamental data in parallel to improve performance.

#### Reproducing the Problem

Run an algorithm using the fine fundamental data and observe the performance slow down as more fine data is requested.

#### Checklist

mchandschuh:
#### Description

Read fine fundamental data in parallel

#### Related Issue

Closes #2189 

#### Motivation and Context

Faster is better!

No.

Ran benchmarks in the cloud and observed performance increases proportional to the number of fine symbols selected. Using 10 fine symbols for a year the run time dropped from ~115 seconds to ~80 seconds.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/UniverseSelection.cs
+++ b/Engine/DataFeeds/UniverseSelection.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using QuantConnect.Data;
 using QuantConnect.Data.UniverseSelection;
 using QuantConnect.Interfaces;
@@ -81,7 +82,9 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                     var fineCollection = new BaseDataCollection();
                     var dataProvider = new DefaultDataProvider();
 
-                    foreach (var symbol in selectSymbolsResult)
+                    // use all available threads, the entire system is waiting for this to complete
+                    var options = new ParallelOptions{MaxDegreeOfParallelism = Environment.ProcessorCount};
+                    Parallel.ForEach(selectSymbolsResult, options, symbol =>
                     {
                         var config = FineFundamentalUniverse.CreateConfiguration(symbol);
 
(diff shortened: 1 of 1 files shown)


Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -81,7 +82,9 @@ public SecurityChanges  ApplyUniverseSelection(Universe universe, DateTime dateT
                     var fineCollection = new BaseDataCollection();
                     var dataProvider = new DefaultDataProvider();
 
-                    foreach (var symbol in selectSymbolsResult)
+                    // use all available threads, the entire system is waiting for this to complete
+                    var options = new ParallelOptions{MaxDegreeOfParallelism = Environment.ProcessorCount};
+                    Parallel.ForEach(selectSymbolsResult, options, symbol =>
```
jaredbroad: 😎

Martin-Molinero: Beautiful! just out of curiosity, did you test using `> Environment.ProcessorCount`? Like `Environment.ProcessorCount * 1.5`, Maybe there is room for even more juice

mchandschuh: I did not try but I can. I wouldn't expect that to help since all of these reads are synchronous -- if we were doing async reads I would think we could squeeze a little more out of it

mchandschuh: @Martin-Molinero -- tried 1.5x and didn't notice and improvements (statistically insignificant delay was actually observed) -- hunch is that it's just extra overhead and the threads are simply competing each other

Issue #2240, opened 2018-07-12
SubscriptionCollection is sorted mulltiple times affecting performance


Martin-Molinero:
#### Expected Behavior

`SubscriptionCollection` is only sorted the required amount of times.
#### Actual Behavior

`SubscriptionCollection` is sorted each time a subscription is added and removed inside multiple loops **on the algorithm thread making it slower**.
#### Potential Solution

Review and refactor `SubscriptionCollection` sorting mechanism.
#### Reproducing the Problem

Execute an algorithm with multiple subscriptions. The following image shows the thread running the algorithm spending 48% (`RemoveSubscription` and `AddSubscription`) of its time sorting. Using SSD and 6 symbols.
![imagen](https://user-images.githubusercontent.com/18473240/42659055-71b27abc-85fd-11e8-842b-8171a284308d.png)

#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2242, opened 2018-07-13
Improve subscription sorting, FillForwardResolution, mechanism to improve performance

Martin-Molinero:
#### Description

- Moving `_fillForwardResolution` field, from both `FileSystemDataFeed.cs` and `LiveTradingDataFeed.cs` into `SubscriptionCollection.cs`. This avoids duplication and more importantly adds control on how many times the `_fillForwardResolution` is calculated. Also, **this avoids having to sort each time we calculate** the `_fillForwardResolution`, since now we will use the ConcurrentDictionaries keys. Adding new unit tests
- Adding a new flied `_sortingSubscriptionRequired` to `SubscriptionCollection`, which allows a jealous approach for calls on `SortSubscriptions()`. The sorting will only happen if there were subscriptions changes and when the `public IEnumerator<Subscription> GetEnumerator()` is called. Adding new unit tests

- Both changes above generate a **230% time improvement** in the execution of the test algorithm (using SSD and 6 symbols), at 43k data points per second, versus 19k data points per second on master. Performance improvement should be bigger when using more symbols. New algorithm thread runner performance call tree follows, master call tree at https://github.com/QuantConnect/Lean/issues/2240
imagen

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2240
#### Motivation and Context

Improve performance

N/A

Test algorithm, unit and regression tests.
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/SubscriptionCollection.cs
+++ b/Engine/DataFeeds/SubscriptionCollection.cs
@@ -14,11 +14,13 @@
  *
 */
 
+using System;
 using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Data;
+using QuantConnect.Util;
 
 namespace QuantConnect.Lean.Engine.DataFeeds
 {
@@ -28,6 +30,8 @@ namespace QuantConnect.Lean.Engine.DataFeeds
     public class SubscriptionCollection : IEnumerable<Subscription>
     {
         private readonly ConcurrentDictionary<SubscriptionDataConfig, Subscription> _subscriptions;
+        private bool _sortingSubscriptionRequired;
+        private readonly Ref<TimeSpan> _fillForwardResolution;
 
         // some asset types (options, futures, crypto) have multiple subscriptions for different tick types,
         // we keep a sorted list of subscriptions so we can return them in a deterministic order
@@ -40,6 +44,8 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         {
             _subscriptions = new ConcurrentDictionary<SubscriptionDataConfig, Subscription>();
             _subscriptionsByTickType = new List<Subscription>();
+            var ffres = Time.OneMinute;
+            _fillForwardResolution = Ref.Create(() => ffres, res => ffres = res);
         }
 
         /// <summary>
@@ -60,12 +66,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// <returns>True if the subscription is successfully added, false otherwise</returns>
         public bool TryAdd(Subscription subscription)
         {
-            if (!_subscriptions.TryAdd(subscription.Configuration, subscription))
-                return false;
-
-            SortSubscriptions();
-
-            return true;
+            if (_subscriptions.TryAdd(subscription.Configuration, subscription))
+            {
+                UpdateFillForwardResolution(FillForwardResolutionOperation.AfterAdd, subscription.Configuration);
+                _sortingSubscriptionRequired = true;
+                return true;
+            }
+            return false;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Engine/DataFeeds/SubscriptionCollection.cs
```diff
@@ -117,17 +126,70 @@ IEnumerator IEnumerable.GetEnumerator()
...
+        /// Gets and updates the fill forward resolution by checking specified subscription configurations and
+        /// selecting the smallest resoluton not equal to tick
+        /// </summary>
+        private void UpdateFillForwardResolution(bool adding, SubscriptionDataConfig configuration)
+        {
+            // Due to performance implications let's be jealous in updating the _fillForwardResolution
+            if (ValidateFillForwardResolution(configuration) &&
+                (
+                    (adding  // we are adding
+                    && configuration.Resolution.ToTimeSpan() != _fillForwardResolution.Value) // check if the new timeSpan is different
```
mchandschuh: I think we can even do a _little_ better here. First, let's use `configuration.Increment` which is the time span equivalent of resolution as well, second I think we only need to update the fill forward resolution if the added subscription's increment is smaller.

Engine/DataFeeds/SubscriptionCollection.cs
```diff
@@ -117,17 +126,70 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                ||
+                    (!adding // We are removing
+                    && configuration.Resolution.ToTimeSpan() == _fillForwardResolution.Value // True: We are removing the resolution we were using
+                    && _subscriptions.All(x => x.Key.Resolution != configuration.Resolution))) // False: there is at least another one equal, no need to update
+                )
+            {
+                var configurations = adding ? _subscriptions.Keys.Concat(new[] { configuration }) : _subscriptions.Keys;
```
mchandschuh: I think the name `adding` here is misleading. I think we only need to append the requested `configuration` is via `UpdateAndGetFillForwardResolution` where the consumer is basically saying, I'm 'about' the add this to the collection, so start using this one now, but via calls from `TryAdd` above, the specified `configuration` will already be in `_subscriptions.Keys` (methinks)

Engine/DataFeeds/SubscriptionCollection.cs
```diff
@@ -117,17 +126,70 @@ IEnumerator IEnumerable.GetEnumerator()
...
+                _fillForwardResolution.Value = configurations.Where(ValidateFillForwardResolution)
+                                                             .Select(x => x.Resolution)
+                                                             .Distinct()
```
mchandschuh: I know this is existing code that has been transplanted, but looking at it now I don't think we actually need this `Distinct` call.

Engine/DataFeeds/SubscriptionCollection.cs
Martin-Molinero: I'll use `configuration.Increment` 👍 
Right, I thought the same. Problem is, the _default_ fill forward enumerator resolution might be smaller than the users resolution. So we should add logic to detect when the `_subscriptions` collection is empty for overwriting the _smaller_ logic, and that stopped me from doing so since its an operation with awful performance, what do you think?

Engine/DataFeeds/SubscriptionCollection.cs
Martin-Molinero: yes your right! I'll make it more specific and clear

Engine/DataFeeds/SubscriptionCollection.cs
Martin-Molinero: I added the `Distinct` call :P. Your right, we don't need it, but I believed it would be more performant since we would remove duplicates. What do you think?

Engine/DataFeeds/SubscriptionCollection.cs
mchandschuh: good point, I think the cost savings probably aren't worth the added complexity.

Engine/DataFeeds/SubscriptionCollection.cs
mchandschuh: `Distinct` under the cover creates a new set, loops through all entries pushing them into the set, and ones that are successfully added move on to the next iterator -- my hunch is that it's extra work we don't need to do -- I doubt the savings will be statistically relevant since we've already minimized how often this method is called -- would probably only be noticeable if we continuously added/removed subscriptions with different resolutions requiring the fill forward resolution to be constantly updated.

PR #2244, opened 2018-07-15
Features 1998 & 2219: portfolio optimization implementation

### LINKED ISSUE #1998: Implements C# version of BlackLittermanPortfolioConstructionModel
#### Expected Behavior
C# and Python implementations of all models.

#### Actual Behavior
Missing Black Litterman implementation in C#.

#### Potential Solution
Port to C#: 
 https://github.com/QuantConnect/Lean/blob/master/Algorithm.Framework/Portfolio/BlackLittermanPortfolioConstructionModel.py

#### Checklist 

### LINKED ISSUE #2219: Implements C# version of MeanVarianceOptimizationPortfolioConstructionModel
#### Expected Behavior
C# and Python implementations of all models.

#### Actual Behavior
Missing Mean Variance Optimization implementation in C#.

#### Potential Solution
Port to C#: [MeanVarianceOptimizationPortfolioConstructionModel.py](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.py)

#### Note
Please look at the note left in #1998, since it tries to solve the same issue: find a method to compute the minimum of the objective function. 

#### Checklist

wildart:
Implementation of Mean-Variance & Black-Litterman portfolio optimizations.

#### Description
Implementation of Mean-Variance & Black-Litterman portfolio optimizations using Alglib2 library NLP & QP solvers.

QP solver used for MVO, NLP solver for BL.

#### Related Issue
Closes #1998
Closes #2219

Python and C# programs can successfully use QuantConnect.Algorithm.Framework library to access portfolio optimization classes. Results between Python and C# implementation are similar.

#### Types of changes

#### Checklist:


Algorithm.Framework/Portfolio/SymbolData.cs
```diff
@@ -0,0 +1,71 @@
...
+using QuantConnect.Data;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Indicators;
+
+namespace QuantConnect.Algorithm.Framework.Portfolio
+{
+    /// <summary>
+    /// Contains data specific to a symbol required by this model
+    /// </summary>
+    public class SymbolData
```
AlexCatarino: `SymbolData` is a "generic" name we use in many models, so a more specific name would be welcome. For example, `RateOfChangeSymbolData` or `ReturnSymbolData`.

Algorithm.Python/MeanVarianceOptimizationAlgorithm.py
```diff
@@ -24,8 +24,10 @@
 from QuantConnect.Algorithm.Framework.Execution import *
 from QuantConnect.Algorithm.Framework.Risk import *
 from QuantConnect.Algorithm.Framework.Selection import *
-from Alphas.HistoricalReturnsAlphaModel import *
-from Portfolio.MeanVarianceOptimizationPortfolioConstructionModel import *
+#from Alphas.HistoricalReturnsAlphaModel import *
+#from Portfolio.MeanVarianceOptimizationPortfolioConstructionModel import *
+from QuantConnect.Algorithm.Framework.Alphas import *
+from QuantConnect.Algorithm.Framework.Portfolio import *
```
AlexCatarino: Since you have commented out the imports from python models, this example will use the C# models. When possible, we should use python models in python algorithms.

Algorithm.CSharp/BlackLittermanOptimizationAlgorithm.cs
```diff
@@ -0,0 +1,106 @@
...
+            // Forex, CFD, Equities Resolutions: Tick, Second, Minute, Hour, Daily.
+            // Futures Resolution: Tick, Second, Minute
+            // Options Resolution: Minute Only.
+            
+            // set algorithm framework models
+            SetUniverseSelection(new CoarseFundamentalUniverseSelectionModel(CoarseSelector));
+            SetAlpha(new CompositeAlphaModel(
+                new HistoricalReturnsAlphaModel(lookback, Resolution.Daily),
+                new RsiAlphaModel()));
+            SetPortfolioConstruction(new BlackLittermanPortfolioConstructionModel(new MaxSharpeRatioPortfolio()));
```
AlexCatarino: In the python version, `BlackLittermanPortfolioConstructionModel`'s first parameter is optional. We think that the default value should be this  (`MaxSharpeRatioPortfolio`) optimization method. 

Also the field variable `_symbols` should have the **private** access modifier.

Algorithm.CSharp/MeanVarianceOptimizationAlgorithm.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            // Find more symbols here: http://quantconnect.com/data
+            // Forex, CFD, Equities Resolutions: Tick, Second, Minute, Hour, Daily.
+            // Futures Resolution: Tick, Second, Minute
+            // Options Resolution: Minute Only.
+            
+            // set algorithm framework models
+            SetUniverseSelection(new CoarseFundamentalUniverseSelectionModel(CoarseSelector));
+            SetAlpha(new HistoricalReturnsAlphaModel(lookback, Resolution.Daily));
+            SetPortfolioConstruction(new MeanVarianceOptimizationPortfolioConstructionModel(new MeanVariancePortfolio()));
```
AlexCatarino: In the python version, `MeanVarianceOptimizationPortfolioConstructionModel`'s first parameter is optional. We think that the default value should be this (`MeanVariancePortfolio`) optimization method.

Also the field variable `_symbols` should have the **private** access modifier.

Algorithm.Python/BlackLittermanOptimizationAlgorithm.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def coarseSelector(self, coarse):
+        # Drops SPY after the 8th
+        last = 3 if self.Time.day > 8 else len(self.symbols)
+
+        return self.symbols[0:last]
+
+    def OnOrderEvent(self, orderEvent):
+        if orderEvent.Status == OrderStatus.Filled:
+            self.Debug(orderEvent.ToString())
```
AlexCatarino: No change required, just a note: in python, we don't need to use C# `.ToString()` method. We can use `str(obj)` of `f"{obj}"` or even just pass the object: `self.Debug(orderEvent)` since Lean will convert C# managed objects from python into string.

Algorithm.Framework/Portfolio/BlackLittermanPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,310 @@
...
+                .PushThrough(bar =>
+                {
+                    ReturnsSymbolData symbolData;
+                    if (_symbolDataDict.TryGetValue(bar.Symbol, out symbolData))
+                    {
+                        symbolData.ROC.Update(bar.EndTime, bar.Value);
+                        if (symbolData.IsReady)
+                        {
+                            var values = symbolData.Window.Select(x => x.Value).ToArray();
+                            algorithm.Log(" ### " + symbolData.Symbol.Value + string.Join(",", values));
```
AlexCatarino: In the previous peer-review, I forgot to ask to change this `Log` call for `Debug` call.

Algorithm.Framework/Portfolio/BlackLittermanPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,310 @@
...
+                    var A = invCov.Add(PTomega.Dot(P));
+                    var B = invCov.Dot(Π).Add(PTomega.Dot(Q));
+                    Π = A.Inverse().Dot(B);
+                }
+            }
+
+            // The optimization method processes the data frame
+            double[] W;
+            _optimization.SetCovariance(Σ);
+            _optimization.SetBounds(_minimumWeight, _maximumWeight);
```
AlexCatarino: Since these bounds don't change during runtime, they should be set in the optimization class constructor along with targetReturn.

Algorithm.Framework/Portfolio/Optimization.cs
```diff
@@ -0,0 +1,256 @@
...
+    {
+        public double[,] _cov;
+        public double[] _x0;
+        public double[] _scale;
+        public double _lower;
+        public double _upper;
+        public List<double[]> _constraints;
+        public List<int> _constraintTypes;
+
+        public double _targetReturn;
```
AlexCatarino: Do these (or, at least, all of these) public fields need to be public?

wildart:
Currently, the optimization model default value is **null**, and depending on a construction model the optimization algo is set accordinly in the model constructor. That is the only way with optional parameters in C# as they must be a compile time constants, i.e. not objects.

jaredbroad:
Very close thanks @wildart, I think Alex means: in python the args are:

`self.SetPortfolioConstruction(MeanVarianceOptimizationPortfolioConstructionModel(optimization_method = self.maximum_sharpe_ratio))`

All the other models we've made are "drop-in" compatible as it reduces documentation and difficulty of understanding for the consumers. If possible can you please make it have identical parameter args; so in this case, the python function is taking a function passed in from local scope, not a custom class interface+implementation.

`optimization_method = self.maximum_sharpe_ratio`

Where the function takes `returns` and returns `weights`

@AlexCatarino looks like the `opt` is not used; can we remove that from the python version?

Algorithm.Framework/Portfolio/PortfolioConstructionModel.cs
```diff
@@ -12,12 +12,24 @@
...
+using System.Linq;
+using Accord.Math;
 using QuantConnect.Algorithm.Framework.Alphas;
 using QuantConnect.Data.UniverseSelection;
 
 namespace QuantConnect.Algorithm.Framework.Portfolio
 {
+    public static class Tools
+    {
+        public static TValue GetValueOrDefault<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, TKey key, TValue defaultValue)
```
mchandschuh: This is an overly generic method that may be better placed in `QuantConnect.Common\Util\LinqExtensions.cs`

mchandschuh: This is an overly generic function that is perhaps better place in `QuantConnect.Common\Util\LinqExtensions.cs`

Algorithm.Framework/Portfolio/MeanVarianceOptimizationPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,181 @@
...
+                        continue;
+                    }
+                    data.Add(algorithm.Time, (decimal)insight.Magnitude.Value);
+                }
+            }
+
+            // Get symbols' returns
+            var returns = GetReturns(from s in symbols join sd in _symbolDataDict on s equals sd.Key select sd.Value.Returns);
+            // Calculate rate of returns
+            var rreturns = returns.Apply(e => System.Math.Pow(1.0 + e, 252.0) - 1.0);
```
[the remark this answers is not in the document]
jaredbroad: :+1: No problem can merge as is give python shares same defect. We'll fix both later to read the tradable dates.

Algorithm.Framework/Portfolio/PortfolioConstructionModel.cs
```diff
@@ -42,5 +54,17 @@ public virtual IEnumerable<IPortfolioTarget> CreateTargets(QCAlgorithmFramework
         public virtual void OnSecuritiesChanged(QCAlgorithmFramework algorithm, SecurityChanges changes)
         {
         }
+
+        public static double[,] GetReturns(IEnumerable<Dictionary<DateTime, double>> returnsByDate)
```
mchandschuh: Not sure if this function belongs here. Specifically, it's not *getting* returns, but its massaging the provided returns into a specific structure according to specific rules. Maybe simply changing the name would make this more descriptive? Also, some XML documentation might help here as well.

EDIT: Looks like this may find a better home on `ReturnsSymbolData` ?

Algorithm.Framework/Portfolio/ReturnsSymbolData.cs
```diff
@@ -0,0 +1,65 @@
...
+                {
+                    Window.Add(updated);
+                }
+            };
+        }
+
+        public void Add(DateTime time, decimal value)
+        {
+            var item = new IndicatorDataPoint(Symbol, time, value);
+            Window.Add(item);
```
mchandschuh: Does `Window` need to be public? Since we have an `Add` method and `IsReady` and `Returns` properties, makes me think `Window` should be private

Algorithm.Framework/QuantConnect.Algorithm.Framework.csproj
```diff
@@ -123,6 +140,10 @@
       <Project>{73fb2522-c3ed-4e47-8e3d-afad48a6b888}</Project>
       <Name>QuantConnect.Indicators</Name>
     </ProjectReference>
+    <ProjectReference Include="..\packages\alglib2.dll.3.10.4\Alglib2.csproj">
```
mchandschuh: Interesting... this nuget package comes with the entire source code? Does this **have** to be a project reference?

Algorithm.Python/MeanVarianceOptimizationAlgorithm.py
```diff
@@ -68,7 +68,7 @@ def coarseSelector(self, coarse):
 
     def OnOrderEvent(self, orderEvent):
         if orderEvent.Status == OrderStatus.Filled:
-            self.Debug(orderEvent.ToString())
+            self.Debug(orderEvent)
```
mchandschuh: was this intentional/required?

jaredbroad:
More research needed... GPL v2 (http://www.alglib.net/download.php) 

"Despite our best efforts, the FSF has never considered the Apache License to be compatible with GPL version 2, citing the patent termination and indemnification provisions as restrictions not present in the older GPL license. The Apache Software Foundation believes that you should always try to obey the constraints expressed by the copyright holder when redistributing their work."
https://www.apache.org/licenses/GPL-compatibility.html

IANAL but it might mean not even authorized to use for personal use. Will contact AlgLib and see.

QuantConnect.Lean.sln
```diff
@@ -48,7 +48,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "QuantConnect.Algorithm.Pyth
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "QuantConnect.ToolBox", "ToolBox\QuantConnect.ToolBox.csproj", "{AC9A142C-B485-44D7-91FF-015C22C43D05}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "QuantConnect.VisualStudioPlugin", "VisualStudioPlugin\QuantConnect.VisualStudio17Plugin.csproj", "{5326A9FB-0270-4647-8C43-20C5607DB529}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "QuantConnect.VisualStudio17Plugin", "VisualStudioPlugin\QuantConnect.VisualStudio17Plugin.csproj", "{5326A9FB-0270-4647-8C43-20C5607DB529}"
```
Martin-Molinero: Is this change required? 
I believe changes on `QuantConnect.Lean.sln` should also be applied to `QuantConnect.Lean.VS15.sln`

Algorithm.Python/MeanVarianceOptimizationAlgorithm.py
wildart: @AlexCatarino suggested that `self.Debug(orderEvent)` is the way, see above.

Algorithm.Python/BlackLittermanOptimizationAlgorithm.py
wildart: @mchandschuh has some doubts about it.

Algorithm.Framework/Alphas/HistoricalReturnsAlphaModel.cs
```diff
@@ -57,7 +57,7 @@ public class HistoricalReturnsAlphaModel : AlphaModel
         /// <param name="algorithm">The algorithm instance</param>
         /// <param name="data">The new data available</param>
         /// <returns>The new insights generated</returns>
-        public IEnumerable<Insight> Update(QCAlgorithmFramework algorithm, Slice data)
+        public override IEnumerable<Insight> Update(QCAlgorithmFramework algorithm, Slice data)
```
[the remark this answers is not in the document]
wildart: yep

wildart:
@jameschch A solution to licencing issue could be solved by downloading GPL library before building project for personal use. As for commercial purposes you would need a proper license.

QuantConnect.Lean.sln
wildart: QuantConnect.Lean.VS15.sln references QuantConnect.VisualStudio15Plugin project.

wildart:
I added project conditional building for alglib if the dll is it present in a `Vendor` directory. This would allow to avoid GPL contamination of the project, no GPL code is included, while providing dynamic linking.

wildart:
> Also, is the SetCovariance method absolutely required? Maybe there's another way we can achieve this, perhaps the optimization method accepts it as an argument? Or since covariance is a pretty standard definition, maybe just a utility function that computes it that all optimizers can invoke to compute it where required.

@mchandschuh Currently `PortfolioConstructionModel` takes optimization method as an object, it would be impossible to create setup the covariance at the construction time. One way of getting around is to pass a type of optimization method to some factory function that would construct a new optimization method object, but then it would be hard to create the optimization method with the specific parameters, like a target return value.

wildart:
I added an addition optimization implementations based on Accord Goldfarb-Idnani QP solver. Although Aglib is more precise and stable, you would not have to worry about licencing issues. So right now, default solver is Accord, and if you decided to use in Alglib just put library in `Vendor` folder and recompile.

Any other review issues?

Algorithm.Framework/Portfolio/ReturnsSymbolData.cs
wildart: `Window` is private now.

jaredbroad:
Hey @wildart! On the home stretch! There's been some intense brainstorming on this pull request here at QC HQ. Thank you for your patience as we've been iterating on the final solution. 

 1. Thank you for the Accord optimizer implementation! That is perfect. I spoke with Serge at AlgLib and he disagreed that the linker would avoid GPL contamination, and just for sake of simplicity I agree and think we should just strip it out and use the accord implementation only. We _love_ the implementation and idea -- perhaps its something we can circle back to if Alglib's license changes.

2. Please rebase; we defined a base interface `IPortfolioOptimizer` which should serve both optimizers identified so far. It would be nice if you could please refactor your solution to use it. This also removes the SetCovariance() method -- "the LEAN way" would be to handle it internally so all the optimizers could be directly API Compatible. Hopefully, the python examples should be nice and clear to follow now as they also implement the interface.

We serve a c# and python audience so to minimize the documentation and maximize the ability of people to communicate we aim for the API to be "drop in compatible" meaning they follow _exactly_ the same naming conventions, parameter ordering etc. 

3. Once ready for final PR; please strip out all debug/logging. All our other models are all silent once they're shipped unless there is an error in which case you can make noise :smiley:. 

4. In python we have this method:
```
def WarmUpIndicators(self, history):
    for tuple in history.itertuples():
        self.roc.Update(tuple.Index, tuple.close)
```
But in C# you should use `history.PushThrough` to update your roc indicator. This was a helper we made to make indicators easier.

Alex recommended you start with a new branch and copy your changes manually across. It might help minimize the conflicts.

Issue #2197, opened 2018-07-03
Delisted symbol data being sent through OnData(Slice)


Martin-Molinero:
#### Expected Behavior

Delisted symbol should not be sent through `OnData(Slice)` after it has been delisted and user informed.
#### Actual Behavior

On the day a minute symbol is delisted, `OnData(Slice)` contains all the minute data from the previous day. Log [log_and_algorithm.zip](https://github.com/QuantConnect/Lean/files/2159636/log_and_algorithm.zip)
#### Potential Solution

Review code to find underlying issue.
#### Reproducing the Problem

The algorithm to reproduce the issue: [log_and_algorithm.zip](https://github.com/QuantConnect/Lean/files/2159636/log_and_algorithm.zip). The issue happens on the 1/16/2016 12:00:00 AM, which is a Saturday.

#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2250, opened 2018-07-16
Fix for delisted subscription data being sent

Martin-Molinero:
#### Description

- Moving the `if (_delisted) { break; }` above the `while (_subscriptionFactoryEnumerator.MoveNext())` loop. Why? The `_subscriptionFactoryEnumerator` isn't useful, when created, we detect that the subscription is delisted through `CheckForDelisting()` cause `date > delistingDate`. But we still need to loop once more to send the `new Delisting` event, but after we send it we are done and we can break (not going through `_subscriptionFactoryEnumerator.MoveNext()`, since it will start reading a New zip file that's after the delisting date). We still need to create the `_subscriptionFactoryEnumerator` since that's the way we detect that the subscription has to be delisted.
- Replacing a couple of lines with an already defined method with the same behavior. Method: https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/SubscriptionDataReader.cs#L740
- Modified `Common/Data/Slice.cs` so that the enumerator does not send null data to the algorithm
- Adding a regression test. It will throw an exception if we receive any delisted data (fails on master). It will also throw if we didn't recieve the warning and delisted events. ***requires data not present in the open source repository and will pass in that case**.
     - Modified `Engine/AlgorithmManager.cs` to set the algorithms state to `completed`, it was being left as `running`
     - Modified `Tests/AlgorithmRunner.cs` to assert the state of the algorithm and if not `completed` fail.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2197
#### Motivation and Context

Delisted symbol should not be sent through OnData(Slice) after it has been delisted and user informed.

N/A

Unit and regression tests. Test algorithm at https://github.com/QuantConnect/Lean/issues/2197
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OptionDelistedDataRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionDelistedDataRegressionAlgorithm.cs
@@ -1,0 +1,121 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


mchandschuh:
> NOTE: Adding regression test for this

mchandschuh:
> NOTE: This caused other regression failures to be noticed. @Martin-Molinero is creating issues for the newly discovered failing regression tests.

Issue #2249, opened 2018-07-16
When launching LEAN locally, command-line argument "--config" does not work


mblouin02:
#### Expected Behavior
When specifying the command-ling argument "--config <path to config file>" when launching Quantconnect.Lean"Launcher.exe, the data from the specified config file should be used.

#### Actual Behavior
When specifying the a local config file as an argument, we get a log that claims that this file will be used for the configuration. However, values from the default config file location are used instead.

#### Potential Solution
I actually diagnosed this issue. It comes from the fact that, in Config.cs, the "Settings" singleton is instantiated before the main program gets the chance to read the command-line arguments. In other words, the config file is read when the static variables get created, which happens before program execution.

The solution would be to add "Config.Reset()" after the program read command-line arguments, in order to re-read the settings from the correct config file.

#### Reproducing the Problem
1- Copy ".\Launcher\config.json" to another location (ex: C:\Launcher\config.json).
2- Rename the new file config_UseThisOne.json
3- In config_UseThisOne.json, change "environment" to "live-paper" instead of "backtesting"
4- In Visual Studio, add a break point in Program.cs, line 58 (var liveMode = Config.GetBool("live-mode");)
5- Launch in debug mode, by passing the argument "--config <path to config_UseThisOne.json> file"
6- Step-over the line, so that "live-mode" gets populated
7- Notice that the variable liveMode is false, it should be true.

8- Now, stop the program.
9- In ".\Launcher\config.json", change "environment" to "live-paper" instead of "backtesting"
10- Repeat steps 5-7, notice that the value is now properly "true"

#### System Information
- Windows 10

#### Checklist
- [X] I have completely filled out this template
- [X] I have confirmed that this issue exists on the current `master` branch
- [X] I have confirmed that this is not a duplicate issue by searching [issues]
- [X] I have provided detailed steps to reproduce the issue

jaredbroad:
Thanks @mblouin02 sounds broken -- we'll look into it further.

mblouin02:
Hi Jared, 

See this pull request, it should fix the issue: https://github.com/QuantConnect/Lean/pull/2251

StefanoRaggi:
Closed by #2251

PR #2251, opened 2018-07-16
Fix broken --config command-line argument

mblouin02:
#### Description
Reset the configuration settings at the beginning of program
execution, so that, when passing a custom --config argument
to the program, values from that file are read instead of the
default config file.

#### Related Issue
https://github.com/QuantConnect/Lean/issues/2249

#### Motivation and Context
Config file argument was not used in the program.

Reproduced the steps 1-7 described in https://github.com/QuantConnect/Lean/issues/2249, and confirmed that the value is now correctly read.

#### Types of changes

#### Checklist:
- [N/A] I have added tests to cover my changes.
- [X] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`

### DIFF
--- a/Configuration/Config.cs
+++ b/Configuration/Config.cs
@@ -57,6 +57,7 @@ namespace QuantConnect.Configuration
             if (cliArguments.ContainsKey("config"))
             {
                 SetConfigurationFile(cliArguments["config"] as string);
+                Reset();
             }
 
             var jsonArguments = JsonConvert.DeserializeObject(JsonConvert.SerializeObject(cliArguments));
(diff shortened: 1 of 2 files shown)


Martin-Molinero:
Hi @mblouin02!
Thank you for the issue and PR. Let me suggest a minor change. I believe it would be better to move the `Reset()` call inside the if statement at https://github.com/QuantConnect/Lean/blob/master/Configuration/Config.cs#L57. That way, if another command line argument is send along with the `--config` it will also be taken into account.

mblouin02:
Hi @Martin-Molinero,

I did the change that you recommended. Feel free to comment.

Issue #2204, opened 2018-07-04
OandA Fires duplicate events occasionally


Worthy7:
#### Expected Behavior
Only one event

#### Actual Behavior
Sometimes two of the exact same event

#### Potential Solution
No idea.

#### Reproducing the Problem
Cannot easily reproduce.

#### System Information
Here is a screenshot of the orders, complete with BrokerId ticket
![image](https://user-images.githubusercontent.com/4762399/42256569-1f8deb16-7f8d-11e8-9368-dd5541f0116b.png)

Here is my output log:

> 20180704 13:16:02.177 Trace:: BrokerageTransactionHandler.OrderStatusChanged(): Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Submitted BrokerId: 1162
> 20180704 13:16:02.178 Trace:: LiveTradingResultHandler.OrderEvent(): Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Submitted
> 20180704 13:16:02.181 Trace:: BrokerageTransactionHandler.OrderStatusChanged(): Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Filled Quantity: 2136000 FillPrice: 1.16646 USD Message: Oanda Fill Event BrokerId: 1162
> 20180704 13:16:02.181 Trace:: BrokerageTransactionHandler.OrderStatusChanged(): Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Filled Quantity: 2136000 FillPrice: 1.16646 USD Message: Oanda Fill Event BrokerId: 1162
> 20180704 13:16:02.189 Trace:: Debug: New Order Event: Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Submitted
> 20180704 13:16:02.261 Trace:: LiveTradingResultHandler.OrderEvent(): Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Filled Quantity: 2136000 FillPrice: 1.16646 USD Message: Oanda Fill Event
> 20180704 13:16:02.272 Trace:: Debug: New Order Event: Time: 2018/07/04 4:16:02 OrderID: 2 Symbol: EURUSD Status: Filled Quantity: 2136000 FillPrice: 1.16646 USD Message: Oanda Fill Event
[...]

PR #2254, opened 2018-07-17
Fix OANDA fires duplicate market events

Martin-Molinero:
#### Description

- Adding `Lock` to `OandaRestApiBase.cs` so both V1 and V20 api use it. To avoid depending on `order.Status` Oanda needs to keep track internally of the market orders and their status., adding new collection `ConcurrentDictionary<int, OrderStatus>` for this. Adding another test.
     - `PlaceOrder()` => Take lock, api.PlaceOrder(), add to dictionary only if order is Partially filled, release lock, inform user.
     - `Callback()` => Takes lock, get order, release lock. If order was not removed from the dictionary (means its already fully filled), pass, else inform user.
     - Moving ` OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, orderFee) { Status = OrderStatus.Submitted });` outside lock scope.
- For both api V1 and V20, on `PlaceOrder()` will use `OrderStatus.PartiallyFilled` or `OrderStatus.Filled` based on `(order.AbsoluteQuantity - marketOrderFillQuantity)`
- On `OnTransactionDataReceived()` in api V20, removed line where we set order status directly.
- On `OandaBrokerageTests.cs` overwrote `OrderParameters` to avoid unsupported `StopLimitOrder` tests.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2204
#### Motivation and Context

OANDA Fires duplicate events

N/A

Oanda brokerage tests using trial account
#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/Oanda/OandaRestApiV1.cs
+++ b/Brokerages/Oanda/OandaRestApiV1.cs
@@ -135,70 +135,77 @@ namespace QuantConnect.Brokerages.Oanda
                 { "units", Convert.ToInt32(order.AbsoluteQuantity).ToString() }
             };
 
+            const int orderFee = 0;
+            var marketOrderFillQuantity = 0;
+            var marketOrderRemainingQuantity = 0;
+            decimal marketOrderFillPrice;
+            var marketOrderStatus = OrderStatus.Filled;
+            order.PriceCurrency = SecurityProvider.GetSecurity(order.Symbol).SymbolProperties.QuoteCurrency;
             PopulateOrderRequestParameters(order, requestParams);
 
-            var postOrderResponse = PostOrderAsync(requestParams);
-            if (postOrderResponse == null)
-                return false;
-
-            // if market order, find fill quantity and price
-            var marketOrderFillPrice = 0m;
-            if (order.Type == OrderType.Market)
+            lock (Locker)
             {
+                var postOrderResponse = PostOrderAsync(requestParams);
+                if (postOrderResponse == null)
+                    return false;
+                // Market orders are special, due to the callback not being triggered always, if the order was filled,
+                // find fill quantity and price and inform the user
+                if (postOrderResponse.tradeOpened != null && postOrderResponse.tradeOpened.id > 0)
+                {
+                    if (order.Type == OrderType.Market)
+                    {
+                        marketOrderFillQuantity = postOrderResponse.tradeOpened.units;
+                    }
+                    else
+                    {
+                        order.BrokerId.Add(postOrderResponse.tradeOpened.id.ToString());
+                    }
+                }
+
+                if (postOrderResponse.tradeReduced != null && postOrderResponse.tradeReduced.id > 0)
+                {
+                    if (order.Type == OrderType.Market)
+                    {
+                        marketOrderFillQuantity = postOrderResponse.tradeReduced.units;
+                    }
+                    else
+                    {
+                        order.BrokerId.Add(postOrderResponse.tradeReduced.id.ToString());
+                    }
+                }
+
+                if (postOrderResponse.orderOpened != null && postOrderResponse.orderOpened.id > 0)
+                {
+                    if (order.Type != OrderType.Market)
+                    {
+                        order.BrokerId.Add(postOrderResponse.orderOpened.id.ToString());
+                    }
+                }
+
+                if (postOrderResponse.tradesClosed != null && postOrderResponse.tradesClosed.Count > 0)
+                {
+                    marketOrderFillQuantity += postOrderResponse.tradesClosed
+                        .Where(trade => order.Type == OrderType.Market)
+                        .Sum(trade => trade.units);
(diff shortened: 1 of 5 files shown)


Brokerages/Oanda/OandaRestApiV20.cs
```diff
@@ -157,56 +156,51 @@ public override bool PlaceOrder(Order order)
         {
             const int orderFee = 0;
             ApiResponse<InlineResponse201> response;
+            var request = GenerateOrderRequest(order);
 
-            lock (_locker)
+            // send Submitted order event
+            order.PriceCurrency = SecurityProvider.GetSecurity(order.Symbol).SymbolProperties.QuoteCurrency;
+            lock (Locker)
```
StefanoRaggi: Good catch with the missing locks!

Brokerages/Oanda/OandaRestApiV20.cs
```diff
@@ -349,16 +351,14 @@ private void OnTransactionDataReceived(string json)
                     var transaction = obj.ToObject<OrderFillTransaction>();
 
                     Order order;
-                    lock (_locker)
+                    lock (Locker)
                     {
                         order = OrderProvider.GetOrderByBrokerageId(transaction.OrderID);
                     }
                     if (order != null)
```
StefanoRaggi: Should we move this `if` block inside the `lock` block as well?

Brokerages/Oanda/OandaRestApiV20.cs
```diff
@@ -349,16 +351,14 @@ private void OnTransactionDataReceived(string json)
...
                     {
                         if (order.Type != OrderType.Market || order.Status != OrderStatus.Filled)
                         {
-                            order.Status = OrderStatus.Filled;
-
```
StefanoRaggi: Nice clean up :)

Brokerages/Oanda/OandaRestApiV1.cs
```diff
@@ -134,74 +134,72 @@ public override bool PlaceOrder(Order order)
                 { "instrument", SymbolMapper.GetBrokerageSymbol(order.Symbol) },
```
StefanoRaggi: `OandaRestApiV1`, is deprecated since May 2018 and don't think the endpoints are even working any more:
https://developer.oanda.com/rest-live/introduction/

Maybe we could remove the class completely?

Brokerages/Oanda/OandaRestApiV20.cs
Martin-Molinero: :) thanks!

Brokerages/Oanda/OandaRestApiV20.cs
Martin-Molinero: I wasn't 100% sure, because the methods that call the `api` won't free the lock until they are finished and the callback will be triggered _after_ the call to the `api`, so the thread processing the async response will wait at line354.
_But_ if a response for a previous operation comes in and we call another operation on the same order then both threads will be working on the same order 🙉.
Summary: I'll expand the `lock` block 👍

Brokerages/Oanda/OandaRestApiV1.cs
Martin-Molinero: Yes your right, I'll verify if we are not using it, if so, I'll remove it 👍

Martin-Molinero: I've requested Oanda to confirm that the v1 REST API isn't used for any case/users. I think for this PR I'll leave it, I suspect there are changes in the cloud required too

mchandschuh: we can remove V1 support following confirmation from Oanda and in a separate PR

Brokerages/Oanda/OandaRestApiV1.cs
```diff
@@ -134,74 +134,72 @@ public override bool PlaceOrder(Order order)
...
-                    FillPrice = marketOrderFillPrice,
-                    FillQuantity = marketOrderFillQuantity * Math.Sign(order.Quantity)
-                });
+                    var remainingQuantity = Convert.ToInt32(order.AbsoluteQuantity - marketOrderFillQuantity);
+                    // set order status based on remaining quantity
+                    var status = remainingQuantity > 0 ? OrderStatus.PartiallyFilled : OrderStatus.Filled;
+                    // if market order, find fill quantity and price
+                    var marketOrderFillPrice = Convert.ToDecimal(postOrderResponse.price);
+                    // if market order, also send Filled order event
+                    OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, orderFee)
```
mchandschuh: `OnOrderEvent` will invoke the algorithm's `OrderEvent` event handler, and if that handler submits an order, is it possible to end up in a dead-lock? Locking over user code seems risky to me

Issue #2252, opened 2018-07-17
Integration with Tiingo custom data


StefanoRaggi:
We should support `Tiingo` services as custom data:
- EOD Price data: https://api.tiingo.com/services/tiingo/daily
- News data: https://api.tiingo.com/services/tiingo/news

API:
- https://api.tiingo.com/docs/tiingo/daily
- https://api.tiingo.com/docs/tiingo/news

mchandschuh:
Merged in #2255

PR #2255, opened 2018-07-17
Integration with Tiingo daily data

StefanoRaggi:
#### Description
This PR adds `Tiingo` daily prices as custom data.

#### Related Issue
#2252 
(does not close the issue because News data is not included in this PR)

Possibly we could add the new `TiingoDailyData` type in the documentation, in addition to `Quandl`

Backtesting and Live testing with basic algorithm using `AddData<TiingoDailyData>()`

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/Tiingo/TiingoDailyData.cs
+++ b/Common/Data/Custom/Tiingo/TiingoDailyData.cs
@@ -1,0 +1,172 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using QuantConnect.Data.Market;
+using QuantConnect.Data.UniverseSelection;
+
+namespace QuantConnect.Data.Custom.Tiingo
+{
+    /// <summary>
+    /// Tiingo daily price data
+    /// https://api.tiingo.com/docs/tiingo/daily
+    /// </summary>
+    public class TiingoDailyData : TradeBar
+    {
+        private readonly ConcurrentDictionary<string, DateTime> _startDates = new ConcurrentDictionary<string, DateTime>();
+
+        /// <summary>
+        /// The end time of this data. Some data covers spans (trade bars) and as such we want
+        /// to know the entire time span covered
+        /// </summary>
+        public override DateTime EndTime
+        {
+            get { return Time + Period; }
+            set { Time = value - Period; }
+        }
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


Common/Data/Custom/Tiingo/TiingoDailyData.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// <summary>
+        /// Returns true if the Tiingo API token has been set.
+        /// </summary>
+        public static bool IsAuthTokenSet { get; private set; }
+
+        /// <summary>
+        /// Sets the Tiingo API token.
+        /// </summary>
+        /// <param name="authToken">The Tiingo API token</param>
+        public static void SetAuthToken(string authToken)
```
jaredbroad: To keep naming convention same as Quandl please use "SetAuthCode" or provide an overload which passes through here as SetAuthToken.

Common/Data/Custom/Tiingo/TiingoDailyData.cs
```diff
@@ -0,0 +1,189 @@
...
+        ///     Instance of the T:BaseData object generated by this line of the CSV
+        /// </returns>
+        public override BaseData Reader(SubscriptionDataConfig config, string content, DateTime date, bool isLiveMode)
+        {
+            var list = JsonConvert.DeserializeObject<List<TiingoDailyData>>(content);
+
+            foreach (var item in list)
+            {
+                item.Symbol = config.Symbol;
+                item.Time = item.Date;
```
mchandschuh: cool, looks like we're properly handling start/end times of the bar here -- I'm assuming that `item.Date` means the item contains daily data for that date

Common/Data/Custom/Tiingo/TiingoDailyData.cs
```diff
@@ -0,0 +1,189 @@
...
+using Newtonsoft.Json;
+using QuantConnect.Data.UniverseSelection;
+
+namespace QuantConnect.Data.Custom.Tiingo
+{
+    /// <summary>
+    /// Tiingo daily price data
+    /// https://api.tiingo.com/docs/tiingo/daily
+    /// </summary>
+    public class TiingoDailyData : BaseData
```
mchandschuh: Should this derive from `TradeBar` so it can be used by indicators dependent on OHCLV data (without needing to specify a trade bar selector)?

Common/Data/Custom/Tiingo/TiingoDailyData.cs
```diff
@@ -72,31 +73,31 @@ public override DateTime EndTime
         /// The actual (not adjusted) open price of the asset on the specific date
         /// </summary>
         [JsonProperty("open")]
-        public decimal Open { get; set; }
+        public new decimal Open { get; set; }
```
mchandschuh: yea, I guess I didn't think about how we'll add the `JsonProperty` attributes to the base class properties -- I think these `get/set` implementations **must** patch through to the `base`:

``` csharp
[JsonProperty("open")]
public new decimal Open
{
    get { return base.Open; }
    set { base.Open = value; }
}
```

otherwise we'll have separate storage, one for this who see this as `TradeBar` and another for those who see this as `TiingoDailyData`

mchandschuh: Another option is to make open/high/low/close/volume/period all overrideable -- then you can decorate them directly without the layer of indirection. This also would enable this same behavior for any users attempting to subclass -- thoughts?

Common/Data/Custom/Tiingo/TiingoDailyData.cs
```diff
@@ -60,7 +61,7 @@ public override DateTime EndTime
         /// <summary>
         /// Gets a time span of one day
         /// </summary>
-        public TimeSpan Period => QuantConnect.Time.OneDay;
+        public new TimeSpan Period => QuantConnect.Time.OneDay;
```
mchandschuh: I don't think this will play well with `TradeBar.EndTime`

PR #2269, opened 2018-07-18
Refactors SetHolding to use MarketOnOpen for daily data

### LINKED ISSUE #2210: SetHoldings use MarketOnOpen for Daily Data
#### Expected Behavior
SetHoldings smart enough to use appropriate order type if intraday data isn't available. 

#### Actual Behavior
Constant warnings about conversion to MarketOnOpen when using SetHoldings on Daily.

#### Potential Solution
 - Abstract QCAlgorithm.Trading L204 -> 213 into Method.
 - Use Method QCAlgorithm.Trading L922 to get right order type.
```
//Only place trade if we've got > 1 share to order.
var quantity = CalculateOrderQuantity(symbol, percentage) - marketOrdersQuantity;
if (Math.Abs(quantity) > 0)
{
     MarketOrder(symbol, quantity, false, tag);
}
```

#### Reproducing the Problem
Run equities daily SetHoldings algorith,

#### Checklist

AlexCatarino:
#### Description
The logic to select the right market order type is moved to `SetHoldings`. If the market is closed or there is only daily data, a market on open order is placed.

#### Related Issue
Closes #2210

#### Motivation and Context
LEAN users were confused when inundated by messages saying their market orders were being converted to `MarketOnOpenOrders` when they were using `SetHoldings`. To make the method more intelligent we implemented logic which automatically sends `MarketOnOpen` orders when the exchange is open. We considered factoring in the daily-subscription information but discounted it when realized it is perfectly reasonable to have a daily subscription and submit market orders which maintains the current behaviour. @mchandschuh rightly pointed out that MOO orders would actually be more accurate as they'd fill for the next datapoint; however to maintain the current models we've elected to only convert to `MarketOnOpenOrder` when the exchange is closed.

No.

Manually running BasicTemplateAlgorithm with Daily resolution.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -198,6 +198,20 @@ namespace QuantConnect.Algorithm
         public OrderTicket MarketOrder(Symbol symbol, decimal quantity, bool asynchronous = false, string tag = "")
         {
             var security = Securities[symbol];
+
+            // check the exchange is open before sending a market order, if it's not open
+            // then convert it into a market on open order
+            if (!security.Exchange.ExchangeOpen)
+            {
+                var mooTicket = MarketOnOpenOrder(security.Symbol, quantity, tag);
+                var anyNonDailySubscriptions = security.Subscriptions.Any(x => x.Resolution != Resolution.Daily);
+                if (mooTicket.SubmitRequest.Response.IsSuccess && !anyNonDailySubscriptions)
+                {
+                    Debug("Converted OrderID: " + mooTicket.OrderId + " into a MarketOnOpen order.");
+                }
+                return mooTicket;
+            }
+
             var request = CreateSubmitOrderRequest(OrderType.Market, security, quantity, tag, DefaultOrderProperties?.Clone());
 
             // If warming up, do not submit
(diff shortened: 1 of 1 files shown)


Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -198,20 +198,6 @@ public OrderTicket MarketOrder(Symbol symbol, double quantity, bool asynchronous
         public OrderTicket MarketOrder(Symbol symbol, decimal quantity, bool asynchronous = false, string tag = "")
         {
             var security = Securities[symbol];
-
-            // check the exchange is open before sending a market order, if it's not open
-            // then convert it into a market on open order
-            if (!security.Exchange.ExchangeOpen)
```
mchandschuh: I don't think we want to remove this -- just make `SetHoldings` issue the correct order type.

Consider an algorithm with daily data. Right now submitting `MarketOrder` just works as they get converted to `MarketOnOpenOrder` calls -- if we remove this, the `MarketOrder` will fail due to market being closed (always closed w/ daily data!)

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -922,7 +909,18 @@ into ticket
             var quantity = CalculateOrderQuantity(symbol, percentage) - marketOrdersQuantity;
             if (Math.Abs(quantity) > 0)
             {
-                MarketOrder(symbol, quantity, false, tag);
+                var anyNonDailySubscriptions = security.Subscriptions.Any(x => x.Resolution != Resolution.Daily);
+
+                //Check whether the exchange is open and intraday data is available to send a market order. 
+                //If not, send a market on open order instead
+                if (security.Exchange.ExchangeOpen && anyNonDailySubscriptions)
```
mchandschuh: nice check, this makes sense to me. if exchange is open and we have intraday data, submit `MarketOrder` -- otherwise do `MarketOnOpen` -- love it.

Issue #2156, opened 2018-06-24
Logging.WhoCalledMe.GetMethodName NullReferenceException


DummyCo:
Hi there! While trying to acquire historical data from _Oanda_ with next parameters:
`./QuantConnect.ToolBox EURUSD Minute 20100101 20180101`
I've got next exception after a couple of execution hours:

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at QuantConnect.Logging.WhoCalledMe.GetMethodName(Int32 frame) in \Lean\Logging\WhoCalledMe.cs:line 36
   at QuantConnect.ToolBox.OandaDownloader.Program.Main(String[] args) in \Lean\ToolBox\OandaDownloader\Program.cs:line 102
```

#### Potential Solution
Looks like _new StackFrame(frame + 1).GetMethod()_ returns null somehow and actual exception is lost because of error in logging pipeline.

#### Reproducing the Problem
No stable attempts to reproduce

#### System Information
Windows 10

Thank you!

mchandschuh:
Interesting. This method should only be invoked via the log handler. I notice that the log handler isn't in that stack. Given the stack it makes sense, since +1 would put it above main. This MIGHT be the jit inlining the log method, but unlikely, I don't think it will inline interface methods. Given the stack it looks like the method was called directly from main which it isn't designed to do. If we can't reproduce this then I suggest we close it.

mchandschuh:
Did you make any edits to the code? If not, then we may need to add attributes to suppress jit inlining.

DummyCo:
@mchandschuh no changes made

mchandschuh:
Roger that.

Let's add attributes to suppress method inlining on all methods that call into WhoCalledMe.

PR #2322, opened 2018-07-27
Suppressing inlining on methods that call WhoCalledMe

Martin-Molinero:
#### Description

- Adding `[MethodImpl(MethodImplOptions.NoInlining)]` so no inlining occurs, in the two methods that call `WhoCalledMe()`
   - `public static void Error()` at `Log.cs`
   - `public void GetMethodNameTest()`
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2156
#### Motivation and Context

Get the correct method for the error message.

N/A

Unit and regression tests.

Manual tests, using ToolBox release build:
Stack trace is:
- Program.Main()
- OandaDownloaderProgram.OandaDownloader() -> Exception throwing

Before:

> 20180727 17:28:10.023 ERROR:: Program.Main():  System.Exception: Pepe grillo
   at QuantConnect.ToolBox.OandaDownloader.OandaDownloaderProgram.OandaDownloader(IList`1 tickers, String resolution, DateTime startDate, DateTime endDate) in C:\Users\Poli\Documents\Bolsa\QuantConnect\MyLean\Lean\ToolBox\OandaDownloader\OandaDownloaderProgram.cs:line 43

After:

> 20180727 17:31:56.382 ERROR:: OandaDownloaderProgram.OandaDownloader():  System.Exception: Pepe grillo
   at QuantConnect.ToolBox.OandaDownloader.OandaDownloaderProgram.OandaDownloader(IList`1 tickers, String resolution, DateTime startDate, DateTime endDate) in C:\Users\Poli\Documents\Bolsa\QuantConnect\MyLean\Lean\ToolBox\OandaDownloader\OandaDownloaderProgram.cs:line 43

#### Types of changes

#### Checklist:

### DIFF
--- a/Logging/Log.cs
+++ b/Logging/Log.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -15,10 +15,11 @@
 
 using System;
 using System.Collections;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading;
 
-namespace QuantConnect.Logging 
+namespace QuantConnect.Logging
 {
     /// <summary>
     /// Logging management class.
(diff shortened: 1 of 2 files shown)


Tests/Common/Util/WhoCalledMeTests.cs
```diff
@@ -22,6 +23,7 @@ namespace QuantConnect.Tests.Common.Util
     public class WhoCalledMeTests
     {
         [Test]
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
mchandschuh: Is this required?

Logging/Log.cs
```diff
@@ -104,15 +106,15 @@ public static void Error(Exception exception, string message = null, bool overri
         /// <summary>
         /// Log trace
         /// </summary>
-        public static void Trace(string traceText, bool overrideMessageFloodProtection = false) 
-        { 
-            try 
+        public static void Trace(string traceText, bool overrideMessageFloodProtection = false)
```
mchandschuh: I think we need the no inlining attribute here and on debug as well.

Martin-Molinero: Sorry about these changes, they were automatic, just deleting spaces.
The only two places that `WhoCalledMe.GetMethodName()` are used, is in `Error()` and in the test

Tests/Common/Util/WhoCalledMeTests.cs
Martin-Molinero: Currently, the test passes so no, we don't need it. But in theory yes, it could start inlining and make the test fail

PR #2324, opened 2018-07-30
Exclude warmup period from sampling data

### LINKED ISSUE #2279: CAGR Formula Includes Warmup
#### Expected Behavior
CAGR / Statistics should include tradable period. Warm up is essentially a history request and shouldn't impact the statistics results -- or potentially even the charting period.

#### Actual Behavior
Charts flatline during warm up (because you're not allowed to trade) and the statistics skewed because they use the equity period for the years calculation.

#### Potential Solution
 - Stop plotting of the equity chart during the warm up period. This would fix both the skewed statistics and the weird looking charts.
https://github.com/QuantConnect/Lean/blob/master/Common/Statistics/Statistics.cs#L206

#### Reproducing the Problem
```
{
    private Symbol _spy = QuantConnect.Symbol.Create("SPY", SecurityType.Equity, Market.USA);
    {
        SetStartDate(2010, 07, 18); 
        SetEndDate(DateTime.Now - TimeSpan.FromDays(1)); 
        SetCash(100000); 
        SetWarmup(1000); 
        AddEquity("SPY", Resolution.Daily); 
    }
 
    {
        if (IsWarmingUp) return;
        if (!Portfolio.Invested)
        {
            SetHoldings(_spy, 1);
            Debug("Purchased Stock");
        }
    }
}
```
image

#### Checklist

AlexCatarino:
#### Description
Do not add samples to equity and benchmark charts during warm up.

#### Related Issue
Closes #2279

#### Motivation and Context
Get correct and consistent statistics. When we add samples to the equity and benchmark charts during the warm up, the statistics calculations include the warm up time.
Also, do not show warm up flat line when plotting the equity curve.

No.

Algorithm provided in issue #2279. 
Results with and without `SetCash(100000);` match.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/Results/BacktestingResultHandler.cs
+++ b/Engine/Results/BacktestingResultHandler.cs
@@ -642,6 +642,12 @@ namespace QuantConnect.Lean.Engine.Results
         /// <param name="value">Value for the chart sample.</param>
         public void Sample(string chartName, string seriesName, int seriesIndex, SeriesType seriesType, DateTime time, decimal value, string unit = "$")
         {
+            // Sampling during warming up period skews statistics
+            if (_algorithm.IsWarmingUp)
+            {
+                return;
+            }
+
             lock (_chartLock)
             {
                 //Add a copy locally:
--- a/Algorithm.CSharp/WarmupConversionRatesRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/WarmupConversionRatesRegressionAlgorithm.cs
@@ -49,7 +49,10 @@ namespace QuantConnect.Algorithm.CSharp
             {
                 throw new Exception("Conversion rate for BTC should not be zero.");
             }
-
+            if (!Portfolio.Invested && !IsWarmingUp)
+            {
+                SetHoldings("BTCUSD", 1);
+            }
             Log($"BTC current price: {Securities["BTCUSD"].Price}");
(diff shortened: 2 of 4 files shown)


Engine/Results/BacktestingResultHandler.cs
```diff
@@ -642,6 +642,12 @@ public void RuntimeError(string message, string stacktrace = "")
         /// <param name="value">Value for the chart sample.</param>
         public void Sample(string chartName, string seriesName, int seriesIndex, SeriesType seriesType, DateTime time, decimal value, string unit = "$")
         {
+            // Sampling during warming up period skews statistics
+            if (_algorithm.IsWarmingUp)
```
mchandschuh: this change didn't impact the regression statistics? If not, then I question the assumption that sampling during warm up skews statistics.

AlexCatarino: It didn't. 
We have three types of algorithm with warm up:
1. Those that are not regression algorithm;
2. Those that do not have trades;
3. Those that warm up period is less than one day.
So the regression tests didn't catch the change.

mchandschuh: Let's add a regression test whose statistics change because of this PR -- confirm by running it on master and observing different statistics than on this PR's branch.

Issue #2060, opened 2018-06-01
Support indirect/multi-leg currency conversions ( A->{Bridge Currencies}->B )


mchandschuh:
#### Expected Behavior

The system is able to handle currency pairs that are not directly convertible into the account currency

#### Actual Behavior

The system is unable to perform currency conversions if it can't find a direct pair leading back to the account currency. On crypto exchanges, it's normal fare for currencies/tokens to not be directly convertible to fiat of any kind, for example, on binance *no* coins are directly convertible to fiat (USDT doesn't count :) ). In such cases we need to move through a 'bridge' currency, such as BTC and then can use a BTCUSD rate (maybe even a BTCUSDT if it's the best available), in order to price securities in terms of dollars (or whatever your account currency may be).

This can get unwieldy if you consider the above example but for an account currency of EUR. We now would require three 'legs', an additional one to move from USD (or USDT) to EUR, unless of course the exchange provides a BTCEUR pair of it's own to skip the USD middleman.

#### Potential Solution

> The below may have spiraled into stream of consciousness but I think the comments shed some light on some potential issues/solutions/paths forward so it's left unaltered.

I think to start we should list the various concerns and cases. The main issue I see here is that it's potentially unbounded (the number of potential paths from A->...>B) and may also be temporal, IOW, if there's multiple paths how do you choose which path is the right one (and be consistent!), so things like using market cap (or volume) to give preference to particular currencies to act as 'bridge' currencies can lead to different paths depending on market movement (which might actually be the right thing to do?).

Another thought is first limiting the unbounded aspect (say max of 3 hops? A -> Bridge1 -> Bridge2 -> B, maybe 4?) and could compute a range of potential conversion rates and pull the mean (would require some work to how we pull conversion rates)... I think in all cases the fact that we currently use the security's current price for the conversion rate should be decoupled so that there's a component that provides conversion rates (perhaps `IConversionRateProvider`) and he just so happens to be initialized with some securities, so a refactoring step could be creating this abstraction and implementing for the existing direct case A->B and then expand on that for cases where we require multiple hops.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

viliwonka:
I was thinking about this before.

We are dealing with graph with bidirectional connections.

[...]

PR #2390, opened 2018-08-10
Replace Fixer API with QC API for GDAX FX conversion rates

### LINKED ISSUE #2383: GDAX Brokerage Relying on Fixer
#### Expected Behavior
Users of LEAN would be able to use the GDAX brokerage or the QC API to get the current price of FX assets.

#### Actual Behavior
Relying on rate limited external brokerage service.

#### Potential Solution
Review the GDAX brokerage for an easy option; if none; replace the fixer call with a QC-API/prices call. 
https://github.com/QuantConnect/Lean/blob/master/Api/Api.cs#L592

#### Reproducing the Problem
https://github.com/QuantConnect/Lean/blob/master/Brokerages/GDAX/GDAXBrokerage.Messaging.cs#L95

#### Checklist

StefanoRaggi:
#### Description
The Fixer API has been replaced by the new QC Pricing API in the `GDAXBrokerage`.

#### Related Issue
Closes #2383 

#### Motivation and Context
Remove the dependency on an external service.

No.

A live GDAX algorithm using BTCEUR (which requires the EURUSD conversion rate) has been tested locally.

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
@@ -14,7 +14,6 @@
 */
 
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
 using QuantConnect.Orders;
@@ -61,9 +60,10 @@ namespace QuantConnect.Brokerages.GDAX
         private int _pendingLeanMarketOrderId;
 
         /// <summary>
-        /// Rest client used to call missing conversion rates
+        /// QC API client used to fetch missing conversion rates
         /// </summary>
-        public IRestClient RateClient { get; set; }
+        public IApi RateClient { get; set; } = new Api.Api();
+
         #endregion
 
         /// <summary>
@@ -86,13 +86,17 @@ namespace QuantConnect.Brokerages.GDAX
         /// <param name="apiSecret">api secret</param>
         /// <param name="passPhrase">pass phrase</param>
         /// <param name="algorithm">the algorithm instance is required to retreive account type</param>
-        public GDAXBrokerage(string wssUrl, IWebSocket websocket, IRestClient restClient, string apiKey, string apiSecret, string passPhrase, IAlgorithm algorithm)
+        /// <param name="userId">QC user id</param>
+        /// <param name="userToken">QC user token</param>
+        public GDAXBrokerage(string wssUrl, IWebSocket websocket, IRestClient restClient, string apiKey, string apiSecret, string passPhrase, IAlgorithm algorithm,
+            int userId, string userToken)
             : base(wssUrl, websocket, restClient, apiKey, apiSecret, Market.GDAX, "GDAX")
         {
             FillSplit = new ConcurrentDictionary<long, GDAXFill>();
             _passPhrase = passPhrase;
             _algorithm = algorithm;
-            RateClient = new RestClient("http://data.fixer.io/api/latest?base=usd&access_key=26a2eb9f13db3f14b6df6ec2379f9261");
+
+            RateClient.Initialize(userId, userToken, "");
 
             WebSocket.Open += (sender, args) =>
             {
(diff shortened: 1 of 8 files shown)


Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -61,9 +60,10 @@ internal enum GdaxEndpointType { Public, Private }
         private int _pendingLeanMarketOrderId;
 
         /// <summary>
-        /// Rest client used to call missing conversion rates
+        /// QC API client used to fetch missing conversion rates
         /// </summary>
-        public IRestClient RateClient { get; set; }
+        public IApi RateClient { get; set; } = new Api.Api();
```
mchandschuh: Instead of requiring the entire API object (and also instead of 'newing' it up in this file, let's define an abstraction, something like `IConversionRateProvider` and accept it as a dependency in the constructor.

This will also remove the need to accept the qc user id and the qc user token in the constructor :)

There was a mock of this interface done on the multi-currency conversion PR #2060 

That interface was more focused on currencyA->currencyB conversions, no reason it can't be modified to accept currencies A and B in the `GetConversionRate()` method.

Brokerages/QCPriceProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Linq;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Brokerages
+{
+    /// <summary>
+    /// An implementation of <see cref="IPriceProvider"/> which uses QC API to fetch price data
+    /// </summary>
+    public class QCPriceProvider : IPriceProvider
```
mchandschuh: This may be me being pedantic, but pretty much all implementations in LEAN are `QC` -- can we call this `ApiPriceProvider` to be a little more descriptive?

Brokerages/QCPriceProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        private readonly IApi _api = new Api.Api();
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QCPriceProvider"/> class
+        /// </summary>
+        /// <param name="userId">The QC user id</param>
+        /// <param name="userToken">The QC user token</param>
+        public QCPriceProvider(int userId, string userToken)
```
mchandschuh: Please provide an overload constructor that accepts the `IApi` instance w/out userid/usertoken so that it works with any `IApi` instance that has implemented the `ReadPrices` method.

Common/Interfaces/IPriceProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Provides access to price data for a given asset
+    /// </summary>
+    public interface IPriceProvider
+    {
+        /// <summary>
+        /// Gets the latest price for a given asset
+        /// </summary>
+        /// <param name="symbol">The symbol</param>
+        /// <returns>The latest price</returns>
+        decimal GetLastPrice(Symbol symbol);
```
mchandschuh: awesome! thanks!

Brokerages/ApiPriceProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApiPriceProvider"/> class
+        /// </summary>
+        /// <param name="userId">The QC user id</param>
+        /// <param name="userToken">The QC user token</param>
+        public ApiPriceProvider(int userId, string userToken)
+        {
+            _api = new Api.Api();
+            _api.Initialize(userId, userToken, "");
+        }
```
Martin-Molinero: What do you think if the `ApiPriceProvider` constructors check _`api.ApiConnection.Connected` and if false throw?

StefanoRaggi: Other `IApi` implementations might not have an `ApiConnection` property, I think it would require a new method in the `IApi` interface, e.g. `IsInitialized`

Issue #2392, opened 2018-08-12
Implement Bitfinex brokerage


Adalyat:
#### Expected Behavior

Support Bitfinex cryptocurrency trading platform using on API v1.1 (stable)

#### Actual Behavior
Currently Lean doesn't support Bitfinex brokerage

#### Checklist

- [v] I have completely filled out this template
- [v] I have confirmed that this issue exists on the current `master` branch
- [v] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

StefanoRaggi:
Closed by PR #2393

PR #2393, opened 2018-08-12
Bitfinex brokerage implementation

Adalyat:
Implement Bitfinex cryptocurrency trading platform using API v1.1

#### Description
BitfinexBrokerage is a websocket based brokerage

#### Related Issue
#2392 

#### Motivation and Context
Cryptocurrency trading is quite poor in Lean because of GDAX is very limited.

Current implementation requires changes in documentation for IDataDownloader as it supports new arguments values.

I've tested it manually using simple algorithm to placing/update/cancel orders. Also unit tests covering new code created. 

BitfinexBrokerageTests requires a valid apikey and apisecret to test it over production

#### Types of changes

#### Checklist:


Brokerages/Bitfinex/BitfinexBrokerage.Utility.cs
```diff
@@ -0,0 +1,380 @@
...
+                }
+            }
+
+            //restore starting candles
+            while (candles[0].Timestamp - start > resolution)
+            {
+                candles.Insert(0, new Messages.Candle(candles[0].Timestamp - resolution, candles[0].Open));
+            }
+
+            //restore starting candles
```
StefanoRaggi: Comment should be "restore ending candles"

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,374 @@
...
+                Log.Trace("BitfinexBrokerage.CancelOrder(): Unable to cancel order without BrokerId.");
+                return false;
+            }
+
+            LockStream();
+            var endpoint = GetEndpoint("order/cancel/multi");
+            var payload = new JsonObject();
+            payload.Add("request", endpoint);
+            payload.Add("nonce", GetNonce().ToString());
+            payload.Add("order_ids", order.BrokerId.Select(i => long.Parse(i)));
```
StefanoRaggi: LEAN does not support multi-cancel, is there a reason for not using the single cancel endpoint ?
`/v1/order/cancel`

Brokerages/Bitfinex/BitfinexSymbolMapper.cs
```diff
@@ -0,0 +1,210 @@
...
...TH","DAIUSD","DAIBTC","DAIETH","CFIUSD","CFIBTC","CFIETH","AGIUSD","AGIBTC","AGIETH","BFTUSD","BFTBTC","BFTETH","MTNUSD","MTNBTC","MTNETH","ODEUSD","ODEBTC","ODEETH","ANTUSD","ANTBTC","ANTETH","DTHUSD","DTHBTC","DTHETH","MITUSD","MITBTC","MITETH","STJUSD","STJBTC","STJETH","XLMUSD","XLMEUR","XLMJPY","XLMGBP","XLMBTC","XLMETH","XVGUSD","XVGEUR","XVGJPY","XVGGBP","XVGBTC","XVGETH","BCIUSD","BCIBTC","MKRUSD","MKRBTC","MKRETH","VENUSD","VENBTC","VENETH","KNCUSD","KNCBTC","KNCETH","POAUSD","POABTC","POAETH","LYMUSD","LYMBTC","LYMETH","UTKUSD","UTKBTC","UTKETH","VEEUSD","VEEBTC","VEEETH","DADUSD","DADBTC","DADETH","ORSUSD","ORSBTC","ORSETH","AUCUSD","AUCBTC","AUCETH","POYUSD","POYBTC","POYETH","FSNUSD","FSNBTC","FSNETH","CBTUSD","CBTBTC","CBTETH","ZCNUSD","ZCNBTC","ZCNETH","SENUSD","SENBTC","SENETH","NCAUSD","NCABTC","NCAETH","CNDUSD","CNDBTC","CNDETH","CTXUSD","CTXBTC","CTXETH","PAIUSD","PAIBTC","SEEUSD","SEEBTC","SEEETH","ESSUSD","ESSBTC","ESSETH","ATMUSD","ATMBTC","ATMETH","HOTUSD","HOTBTC","HOTETH","DTAUSD","DTABTC","DTAETH","IQXUSD","IQXBTC","IQXEOS","WPRUSD","WPRBTC","WPRETH","ZILUSD","ZILBTC","ZILETH","BNTUSD","BNTBTC","BNTETH","ABSUSD","ABSETH","XRAUSD","XRAETH","MANUSD","MANETH"
```
StefanoRaggi: For better readability, let's break this single line list into multiple lines, 10 symbols per line should be good enough

Brokerages/Bitfinex/BitfinexSymbolMapper.cs
```diff
@@ -0,0 +1,210 @@
...
+                return false;
+
+            return KnownSymbolStrings.Contains(brokerageSymbol);
+        }
+
+        /// <summary>
+        /// Checks if the currency is supported by Bitfinex
+        /// </summary>
+        /// <returns>True if Bitfinex supports the currency</returns>
+        public bool IsKnownBrokerageCurrency(string currency)
```
StefanoRaggi: Nice idea adding the list of currencies to the symbol mapper. 
Only comment is on the naming of the method, since `IsKnownBrokerageCurrency` could lead people to think it includes also cryptos, perhaps could be renamed to `IsKnownFiatCurrency` ?

Common/Brokerages/BitfinexBrokerageModel.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// Initializes a new instance of the <see cref="BitfinexBrokerageModel"/> class
+        /// </summary>
+        /// <param name="accountType">The type of account to be modelled, defaults to
+        /// <see cref="AccountType.Margin"/></param>
+        public BitfinexBrokerageModel(AccountType accountType = AccountType.Margin)
+            : base(accountType)
+        {
+        }
+
+        public override IFeeModel GetFeeModel(Security security)
```
StefanoRaggi: Please add XML comment for public methods

Common/Orders/Fees/BitfinexFeeModel.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                // limit order posted to the order book, 0% maker fee
+                fee = MakerFee;
+            }
+
+            // get order value in account currency
+            var unitPrice = order.Direction == OrderDirection.Buy ? security.AskPrice : security.BidPrice;
+            if (order.Type == OrderType.Limit)
+            {
+                // limit order posted to the order book, 0% maker fee
```
StefanoRaggi: Please remove the `0%` from the comment

Data/symbol-properties/symbol-properties-database.csv
```diff
@@ -250,3 +250,273 @@ gdax,BTCEUR,crypto,BTCEUR,EUR,1,0.01,0.00000001
...
+bitfinex,ZILETH,crypto,ZILETH,ETH,1,0.00001,256
+bitfinex,BNTUSD,crypto,BNTUSD,USD,1,0.00001,6
+bitfinex,BNTBTC,crypto,BNTBTC,BTC,1,0.00001,6
+bitfinex,BNTETH,crypto,BNTETH,ETH,1,0.00001,6
+bitfinex,ABSUSD,crypto,ABSUSD,USD,1,0.00001,392
+bitfinex,ABSETH,crypto,ABSETH,ETH,1,0.00001,392
+bitfinex,XRAUSD,crypto,XRAUSD,USD,1,0.00001,138
+bitfinex,XRAETH,crypto,XRAETH,ETH,1,0.00001,138
+bitfinex,MANUSD,crypto,MANUSD,USD,1,0.00001,30
+bitfinex,MANETH,crypto,MANETH,ETH,1,0.00001,30
```
StefanoRaggi: Our crypto users will be very happy, 260+ new symbols 👍

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,509 @@
...
+                if (order == null)
+                {
+                    order = _algorithm.Transactions.GetOrderByBrokerageId(brokerId);
+                    if (order == null)
+                    {
+                        // not our order, nothing else to do here
+                        return;
+                    }
+                }
+
```
StefanoRaggi: Shouldn't we emit a `TickType.Trade` tick here ?

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,509 @@
...
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Linq;
+using System.Net;
+using QuantConnect.Logging;
+using QuantConnect.Interfaces;
+using System.Collections.Concurrent;
+using QuantConnect.Util;
+using Newtonsoft.Json.Linq;
+using com.sun.corba.se.impl.protocol.giopmsgheaders;
```
StefanoRaggi: Is this `using` statement used?

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,509 @@
...
+        /// <param name="restUrl">rest api url</param>
+        /// <param name="apiKey">api key</param>
+        /// <param name="apiSecret">api secret</param>
+        /// <param name="algorithm">the algorithm instance is required to retreive account type</param>
+        public BitfinexBrokerage(string wssUrl, string restUrl, string apiKey, string apiSecret, IAlgorithm algorithm)
+            : base(wssUrl, new WebSocketWrapper(), new RestClient(restUrl), apiKey, apiSecret, Market.Bitfinex, "Bitfinex")
+        {
+            _algorithm = algorithm;
+            _securityProvider = algorithm?.Portfolio;
+            RateClient = new RestClient("http://data.fixer.io/api/latest?base=usd&access_key=26a2eb9f13db3f14b6df6ec2379f9261");
```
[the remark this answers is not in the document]
Adalyat: Got it, my intention was to make a separate PR for this migration. Will do that in current PR

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,509 @@
...
+{
+    public partial class BitfinexBrokerage
+    {
+        private const string ApiVersion = "v1";
+        private readonly IAlgorithm _algorithm;
+        private readonly ISecurityProvider _securityProvider;
+        private readonly ConcurrentQueue<WebSocketMessage> _messageBuffer = new ConcurrentQueue<WebSocketMessage>();
+        private readonly object channelLocker = new object();
+        private volatile bool _streamLocked;
+        internal enum BitfinexEndpointType { Public, Private }
```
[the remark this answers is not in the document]
Adalyat: Yeah, actually it's not used right now. But still left it in the code to outline private vs public request. Agreed with you as we can add it later if necessary

Brokerages/Bitfinex/BitfinexBrokerage.Utility.cs
```diff
@@ -0,0 +1,380 @@
...
+        }
+
+        private static decimal GetOrderPrice(Order order)
+        {
+            switch (order.Type)
+            {
+                case OrderType.Limit:
+                    return ((LimitOrder)order).LimitPrice;
+                case OrderType.Market:
+                    return 1;
```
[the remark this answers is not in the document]
Adalyat: I tried zero but Bitfinex refuses for price = 0. It should be any positive number. I used 1 :)

Brokerages/Bitfinex/BitfinexBrokerage.cs
Adalyat: Hmm. Lean allows to associate multiple brokerage orders to a single lean order (order.BrokerId is List). Am I right in thinking that we need close all associated brokerage orders if lean order closed?

Brokerages/Bitfinex/BitfinexSymbolMapper.cs
Adalyat: Okay, will rename

Brokerages/Bitfinex/BitfinexBrokerage.Utility.cs
```diff
@@ -0,0 +1,380 @@
...
+            const int maxAttempts = 10;
+            var attempts = 0;
+            IRestResponse response;
+
+            do
+            {
+                _restRateLimiter.WaitToProceed();
+                response = RestClient.Execute(request);
+                // 429 status code: Too Many Requests
+            } while (++attempts < maxAttempts && (int)response.StatusCode == 429);
```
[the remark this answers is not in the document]
Adalyat: It's relevant for Bitfinex too, but i'm okay if you want to check body separately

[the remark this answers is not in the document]
StefanoRaggi: Great, code 429 is returned so no changes required

Brokerages/Bitfinex/BitfinexBrokerage.Utility.cs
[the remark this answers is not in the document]
StefanoRaggi: This is a good enough reason :) Perhaps we can just add a comment line here

Brokerages/Bitfinex/BitfinexBrokerage.cs
StefanoRaggi: I believe the BrokerId list was implemented for Tradier brokerage who would generate two orders when crossing the zero line (e.g. I am long 1 and try to go short by selling 2). Does Bitfinex behave in the same way?
No changes are required, was just curious about the reason

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
Adalyat: Exactly, you are right. Implemented.

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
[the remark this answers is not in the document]
Adalyat: Your PR helped me :) Implemented

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,522 @@
...
+                    else if (channel == 0)
+                    {
+                        string term = token[1].ToObject<string>();
+                        switch (term.ToLower())
+                        {
+                            case "oc":
+                                OnOrderClose(token[2].ToObject<string[]>());
+                                return;
+                            case "tu":
+                                OnTradeExecuted(token[2].ToObject<string[]>());
```
StefanoRaggi: Is `channel == 0` only for our trades? Do we need to handle `"tu"` messages also if `channel != 0` ?

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,522 @@
...
+                    {
+                        case "subscribed":
+                            OnSubscribe(token.ToObject<Messages.OrderBookSubscription>());
+                            return;
+                        case "unsubscribed":
+                            OnUnsubscribe(token.ToObject<Messages.OrderBookUnsubscribing>());
+                            return;
+                        case "info":
+                        case "ping":
+                            return;
```
StefanoRaggi: We might want to add `"auth"` to the list of known events

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,522 @@
...
+                {
+                    order = _algorithm.Transactions.GetOrderByBrokerageId(brokerId);
+                    if (order == null)
+                    {
+                        // not our order, nothing else to do here
+                        return;
+                    }
+                }
+                Trade trade = Trade.FromTU(entries);
+                EmitTradeTick(order, trade);
```
StefanoRaggi: I think we should move this line before the order checks, otherwise we'll only emit Trade ticks for our fills.

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,522 @@
...
+        {
+            lock (TickLocker)
+            {
+                Ticks.Add(new Tick
+                {
+                    Value = trade.Price,
+                    Time = trade.Time,
+                    Symbol = _symbolMapper.GetLeanSymbol(trade.Pair),
+                    TickType = TickType.Trade,
+                    Quantity = trade.Amount
```
StefanoRaggi: As much as I would love to have _signed quantity_ in Trade ticks, LEAN does not support this for now (volume based indicators currently assume positive values), we would probably need to add the `TradeDirection` to the `Tick` class.
For now I recommend we use `Math.Abs(trade.Amount)` here.

Common/Brokerages/BitfinexBrokerageModel.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Provides Bitfinex fee model
+        /// </summary>
+        /// <param name="security"></param>
+        /// <returns></returns>
+        public override IFeeModel GetFeeModel(Security security)
+        {
+            return new BitfinexFeeModel();
+        }
+    }
```
StefanoRaggi: We might also want to override `DefaultMarkets` here, since the default market for `SecurityType.Crypto` is `Market.GDAX` and this would require algorithms to pass `Market.Bitfinex` to all `AddCrypto` calls in `Initialize`.

Maybe something similar to this:
``` C#
        public override IReadOnlyDictionary<SecurityType, string> DefaultMarkets {  get; } = GetDefaultMarkets();

        private static IReadOnlyDictionary<SecurityType, string>  GetDefaultMarkets()
        {
            var map = DefaultMarketMap.ToDictionary();
            map[SecurityType.Crypto] = Market.Bitfinex;
            return map.ToReadOnlyDictionary();
        }
```

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
Adalyat: Do mean that we still need Emit Trade Tick for order which is created outside LEAN?

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -0,0 +1,522 @@
...
+                    continue;
+                }
+
+                WebSocket.Send(JsonConvert.SerializeObject(new
+                {
+                    @event = "subscribe",
+                    channel = "book",
+                    pair = _symbolMapper.GetBrokerageSymbol(symbol)
+                }));
+            }
```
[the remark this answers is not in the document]
Adalyat: >> I think we also need to subscribe to the "trades" channel to receive all trade ticks.
All trade ticks for all subscribe symbols or only for our orders?

>> Subscribe is called twice by LEAN
Currently bitfinex handle it internally and just send error message = "dup". Not sure that we need to control it on our side.

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
StefanoRaggi: We should emit _Order Fill Events_ only for orders handled by LEAN and this is handled fine in the current code.

We should emit _Trade Ticks_ for **all** market trades, not only ours.

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
Adalyat: Yeah, channel == 0 only for our trades.

>Do we need to handle "tu" messages also if channel != 0
If we want to we need to subscribe to trade channel of specific symbol. Do we want to emit trade tick or all subscribed symbols or only for specific order?

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
Adalyat: Okay

Common/Brokerages/BitfinexBrokerageModel.cs
Adalyat: I remember Jared mentioned about this, but I thought that it's managed internally, but not on current brokerage implementation. Will do.

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,378 @@
...
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
+                throw new Exception($"BitfinexBrokerage.GetCashBalance: request failed: [{(int)response.StatusCode}] {response.StatusDescription}, Content: {response.Content}, ErrorMessage: {response.ErrorMessage}");
+            }
+
+            var availableWallets = JsonConvert.DeserializeObject<Messages.Wallet[]>(response.Content)
+                .Where(WalletFilter(_algorithm.BrokerageModel.AccountType));
+            foreach (var item in availableWallets)
+            {
+                if (item.Available > 0)
```
StefanoRaggi: I believe we should use `item.Amount` here (and in the two usages below) instead of `item.Available`, to include amounts reserved for open orders.

Common/Brokerages/BitfinexBrokerageModel.cs
```diff
@@ -0,0 +1,63 @@
...
+        public override IReadOnlyDictionary<SecurityType, string> DefaultMarkets { get; } = GetDefaultMarkets();
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BitfinexBrokerageModel"/> class
+        /// </summary>
+        /// <param name="accountType">The type of account to be modelled, defaults to <see cref="AccountType.Margin"/></param>
+        public BitfinexBrokerageModel(AccountType accountType = AccountType.Margin)
+            : base(accountType)
+        {
+        }
```
StefanoRaggi: Since `BitfinexBrokerage` will support both Cash and Margin accounts, we should set the `BuyingPowerModel` depending on the `AccountType`:

``` C#
        public override IBuyingPowerModel GetBuyingPowerModel(Security security)
        {
            return AccountType == AccountType.Cash
                ? (IBuyingPowerModel) new CashBuyingPowerModel()
                : new SecurityMarginModel(3.3m);
        }
```

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -0,0 +1,378 @@
...
+                if (response.StatusCode != HttpStatusCode.OK)
+                {
+                    throw new Exception($"BitfinexBrokerage.GetHistory: request failed: [{(int)response.StatusCode}] {response.StatusDescription}, Content: {response.Content}, ErrorMessage: {response.ErrorMessage}");
+                }
+
+                var candles = JsonConvert.DeserializeObject<object[][]>(response.Content)
+                    .Select(entries => new Messages.Candle(entries))
+                    .ToList();
+
+                RestoreMissedCandles(candles, startMTS, endMTS, resolutionInMS);
```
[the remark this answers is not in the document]
Adalyat: Just clarify this. Not sure it's really necessary, but on small resolution bitfinex miss some with no trades within interval, so data doesn't look continuty.

Just give you example (returns 58 bars instead of 60): https://api.bitfinex.com/v2/candles/trade:1m:tBTCUSD/hist?limit=1000&sort=1&start=1533150000000&end=1533153540000

you can see that there is no candles for 
1533152220000 (1 August 2018 г., 19:37:00)
1533153000000 (1 August 2018 г., 19:50:00)

If you think that Lean handle this itself, i'm happy to remove this trick.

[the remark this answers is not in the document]
jaredbroad: `Candles with Volume = 0 are not returned from service` - These should not be returned; LEAN automatically fills forward by default and we allow them to disable that with `fillForward:false`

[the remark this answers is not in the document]
Adalyat: Shall I remove my own implementation? or leave it as it is?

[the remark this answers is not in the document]
StefanoRaggi: Unfortunately we'll have to remove the call to `RestoreMissedCandles` and the method itself

voidale:
@Adalyat awesome job man this is a little hidden gem in Lean right now, wondering what can that new bitfinex supported exchange do? 

-Does it support Margin trading? can we short now? 
- Can we trade BTCETH? e.g. no USD default currency?

Thanks!

gelinger777:
also please note following . SetWarmup behaves wrong when you have subscribed to the minute bars but have chosen some indicator on different resolution (Hour) . So  to make algo warm up for Tema i had to set WarmUp tp 9600 approximately.

Issue #2378, opened 2018-08-09
Add raw price to factor file entries


mchandschuh:
#### Expected Behavior

Factor files contain enough data to be able to self update when provided a split or dividend event.

#### Actual Behavior

Factor files missing raw price for each entry date

#### Potential Solution

Add raw price for each entry date

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

mchandschuh:
In addition to linked issues, see 57affafdc2d1b359ecf91cab902c87f7e74e8dee for updates to the factor files in this repo.

PR #2397, opened 2018-08-13
Implement 'live' factor files

mchandschuh:
#### Description

Adds reference price to dividend/split
Adds GetSplitsAndDividends to FactorFile
Adds Apply methods to FactorFile/FactorFileRow
Updates factor files to include reference prices

#### Related Issue

See #2378 

#### Motivation and Context

Embed reference prices in factor files so we can apply dividend/split events and produce updated factors.

Perhaps.

Extensively tested via unit tests.

#### Types of changes

#### Checklist:


Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -36,12 +40,20 @@ public class FactorFileRow
         /// <summary>
         /// Gets the price factor associated with this data
         /// </summary>
-        public decimal PriceFactor { get; private set; }
+        public decimal PriceFactor
+        {
+            get { return _priceFactor.RoundToSignificantDigits(7); }
```
jaredbroad: Just to preserve as much significance as possible for the triple leveraged ETF's deep splitting cases lets remove rounding.

Common/Data/Auxiliary/FactorFile.cs
```diff
@@ -208,26 +211,127 @@ public bool HasSplitEventOnNextTradingDay(DateTime date, out decimal splitFactor
...
+            var combinedData = GetSplitsAndDividends(data[0].Symbol, exchangeHours).Concat(data)
+                .OrderByDescending(d => d.Time.Date);
 
+            foreach (var datum in combinedData)
+            {
+                var dividend = datum as Dividend;
+                if (dividend != null)
+                {
+                    var nextEntry = lastEntry.Apply(dividend, exchangeHours);
+                    factorFileRows.Add(nextEntry);
```
jaredbroad: A possible case of dropped splits if the splits and dividends occur on the same date it seems?

Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -36,12 +40,20 @@ public class FactorFileRow
...
+            set { _priceFactor = value; }
+        }
 
         /// <summary>
         /// Gets te split factored associated with the date
```
jaredbroad: Spelling

Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -98,6 +129,122 @@ public static IEnumerable<FactorFileRow> Read(string permtick, string market, ou
...
+        public Dividend GetDividend(FactorFileRow futureFactorFileRow, Symbol symbol, SecurityExchangeHours exchangeHours)
+        {
+            // find previous trading day
+            var previousTradingDay = exchangeHours.GetNextTradingDay(Date);
+
+            return Dividend.Create(
+                symbol,
+                previousTradingDay,
+                ReferencePrice,
+                _priceFactor / futureFactorFileRow._priceFactor
```
jaredbroad: Possible divide by zero error

Common/Securities/SecurityExchangeHours.cs
```diff
@@ -335,5 +335,37 @@ private bool IsTimeAfterEarlyClose(DateTime localDateTime)
...
+
+        /// <summary>
+        /// Gets the previous trading day
+        /// </summary>
+        /// <param name="date">The date to start searching at</param>
+        /// <returns>The previous trading day</returns>
+        public DateTime GetPreviousTradingDay(DateTime date)
+        {
+            date = date.AddDays(-1);
+            while (!IsDateOpen(date))
```
jaredbroad: Might be nice to name the variables "localDate" for clarity in exchange timezone (i think?)

Data/equity/usa/factor_files/aig.csv
```diff
@@ -1,69 +1,69 @@
...
+19980901,0.7233058,10.66100,77.38
+19981201,0.7238269,10.66100,94.13
+19990302,0.7242731,10.66100,112.88
+19990601,0.7246296,10.66100,113.94
+19990730,0.7249906,10.66100,118
+19990831,0.7249906,13.333300,92.69
+19991129,0.7253816,13.333300,105.65
+20000229,0.7257351,13.333300,87.19
+20000530,0.7261453,13.333300,113.25
+20000728,0.7264628,13.333300,129.56
```
jaredbroad: Possibly should be normalized?

mchandschuh: I'll update the generator to normalize and refresh these files.

Common/Data/Market/Dividend.cs
```diff
@@ -23,6 +23,25 @@ namespace QuantConnect.Data.Market
     /// </summary>
     public class Dividend : BaseData
     {
+        /// <summary>
+        /// Gets the dividend payment
+        /// </summary>
+        public decimal Distribution
+        {
+            get { return Math.Round(Value, 2); }
```
Martin-Molinero: Would rounding in the `set` cover the `get` case?

Data/equity/usa/factor_files/aig.csv
mchandschuh: I noticed that *some* of the entries had trailing zeros, either way it's nice to not have them, saves on space :)

Common/Data/Auxiliary/FactorFile.cs
```diff
@@ -208,26 +220,137 @@ public bool HasSplitEventOnNextTradingDay(DateTime date, out decimal splitFactor
...
+        /// Gets all of the splits and dividends represented by this factor file
+        /// </summary>
+        /// <param name="symbol">The symbol to ues for the dividend and split objects</param>
+        /// <param name="exchangeHours">Exchange hours used for resolving the previous trading day</param>
+        /// <returns>All splits and diviends represented by this factor file in chronological order</returns>
+        public List<BaseData> GetSplitsAndDividends(Symbol symbol, SecurityExchangeHours exchangeHours)
+        {
+            var dividendsAndSplits = new List<BaseData>();
+            var futureFactorFileRow = SortedFactorFileData.Last().Value;
+            for (var i = SortedFactorFileData.Count - 2; i >= 0 ; i--)
```
Martin-Molinero: Not sure if I'm right, but if using `Count - 1` shouldn't you be able to remove lines `260` and `276`?

Common/Data/Auxiliary/FactorFile.cs
```diff
@@ -208,26 +220,137 @@ public bool HasSplitEventOnNextTradingDay(DateTime date, out decimal splitFactor
...
+            if (data.Count == 0)
+            {
+                return this;
+            }
+
+            var factorFileRows = new List<FactorFileRow>();
+            var lastEntry = SortedFactorFileData.Last().Value;
+            factorFileRows.Add(lastEntry);
+
+            var combinedData = GetSplitsAndDividends(data[0].Symbol, exchangeHours).Concat(data)
```
Martin-Molinero: `data` will/can have `Split` and `Dividend` in it?

Common/Data/Auxiliary/FactorFile.cs
```diff
@@ -208,26 +220,137 @@ public bool HasSplitEventOnNextTradingDay(DateTime date, out decimal splitFactor
...
+        /// <returns>An enumerable of lines representing this factor file</returns>
+        public IEnumerable<string> ToCsvLines()
+        {
+            if (FactorFileMinimumDate != null)
+            {
+                var min = SortedFactorFileData.First().Value;
+                yield return $"{FactorFileMinimumDate:yyyyMMdd},{min.PriceFactor},{min.SplitFactor}";
+            }
+
+            foreach (var kvp in SortedFactorFileData.OrderBy(kvp => kvp.Value.Date))
```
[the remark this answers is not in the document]
mchandschuh: I'll test this, but I'm pretty sure you're right that the sorted collection guarantees enumeration order -- would save a couple cycles/enumerations 👍

Common/Data/Market/Dividend.cs
mchandschuh: I debated this myself, the issue is that it *can* be set directly via the `Value` property which isn't virtual... in reality, this should be updated to hold it's own storage and not use the `Value` property as it can be confusing

Common/Data/Auxiliary/FactorFile.cs
mchandschuh: what would I pass to the `GetDividend` method if I haven't initialized `futureFactorFileRow` first? I figured doing it this way removed an `if` statement from the loop detecting the first iteration which has the year 2050 entry.

Common/Data/Auxiliary/FactorFile.cs
mchandschuh: a single `data` element will be either a `Split` or a `Dividend` -- but it is possible to have a split and dividend on the same date (or at least, this implementation supports it properly). The entire list of data elements can have both splits and dividends

Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -89,15 +106,150 @@ public static IEnumerable<FactorFileRow> Read(string permtick, string market, ou
...
+            // this instance must be chronologically at or in front of the dividend
+            // this is because the factors are defined working from current to past
+            if (Date < previousTradingDay)
+            {
+                throw new ArgumentException($"Factor file row date '{Date:yyy-MM-dd}' is before dividend previous trading date '{previousTradingDay.Date:yyyy-MM-dd}'.");
+            }
+
+            // pfi - new price factor pf(i+1) - this price factor D - distribution C - previous close
+            // pfi = pf(i+1) * (C-D)/C
+            var priceFactor = PriceFactor * (dividend.ReferencePrice - dividend.Distribution) / dividend.ReferencePrice;
```
StefanoRaggi: ✔️

Common/Data/Auxiliary/FactorFile.cs
Martin-Molinero: Your right! I analyzed the code wrong 👍

PR #2299, opened 2018-07-23
Revert Benchmark to SPY

wildart:
Reverted benchmark to SPY.

#### Description
Set benchmark symbol to SPY.

#### Related Issue
#2260

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [ ] I have added tests to cover my changes. 
- [ ] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`

PR #2403, opened 2018-08-15
Move benchmark security to its own universe + Set default benchmark to SPY

### LINKED ISSUE #2153: SetBenchmark Silent Default to Equity
#### Expected Behavior
Ideally the benchmark is resolved as BTCUSD not Equity.
```
SetBenchmark("BTCUSD");
AddCrypto("BTCUSD", Resolution.Minute);
```

#### Actual Behavior
BTCUSD added as Equity; and the AddCrypto line ignored. It was quite confusing as in this case the price was always $0 as there is no BTCUSD Equity. 

#### Potential Solution
Show a message about the Equity assumption? Scan market hours db to get security type? 

#### Checklist

### LINKED ISSUE #2260: Revert Benchmark to SPY
#### Expected Behavior
Benchmark follows index.

#### Actual Behavior
2% absolute return used as benchmark; causing weird beta values for most people.

#### Potential Solution
Revert to using daily SPY as benchmark.

#### Reproducing the Problem
n/a

#### Checklist 

Community confirmed; the monkeys won.
2018-07-17 17_25_47-slack - quantconnect

### LINKED ISSUE #2381: SetBenchmark called in BasicTemplateFrameworkAlgorithm changes number of trades
#### Expected Behavior
We expect a `SetBenchmark` call in any algorithm to have no effect on the number of trades placed.

#### Actual Behavior
`SetBenchmark` changes the number of trades placed by `BasicTemplateFrameworkAlgorithm`.

#### Potential Solution
The benchmark security (which has `Leverage=1`) is being used for trades instead of the security selected by the universe selection model.

#### Reproducing the Problem
Add the following line anywhere in the `Initialize` method of  `BasicTemplateFrameworkAlgorithm`:
```
    SetBenchmark("SPY");
```
Run the algorithm: it will place **19** trades instead of **3**.

#### Checklist

StefanoRaggi:
#### Description
- Added new `BenchmarkUniverse` to handle the benchmark security and subscription separately from the algorithm ones
- Updated both data feeds to handle the `CollectionChanged` event for this universe
- Added a new `CustomUniverseWithBenchmarkRegressionAlgorithm`
- Updated regression stats for a couple of algorithms using `SetBenchmark`
- Set default benchmark to SPY and updated all regression statistics

#### Related Issue
Closes #2381
Closes #2260 
Closes #2299 
Closes #2153 

#### Motivation and Context
Due to the current implementation of `SetBenchmark` allowing reuse of the uninitialized benchmark security, many algorithms were behaving incorrectly (incorrect leverage used and/or no data streamed).

No.

New regression algorithm + existing algorithms with the additional `SetBenchmark` call.

#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.Universe.cs
```diff
@@ -419,6 +419,30 @@ public void AddUniverse(SecurityType securityType, string name, Resolution resol
...
+
+            Universe universe;
+            if (!UniverseManager.TryGetValue(universeSymbol, out universe))
+            {
+                var objectType = SubscriptionManager.LookupSubscriptionConfigDataTypes(security.Type, security.Resolution, false).First().Item1;
+                var marketHoursDbEntry = MarketHoursDatabase.GetEntry(security.Symbol.ID.Market, security.Symbol.Value, security.Type);
+                var dataTimeZone = marketHoursDbEntry.DataTimeZone;
+                var exchangeTimeZone = marketHoursDbEntry.ExchangeHours.TimeZone;
+                var config = new SubscriptionDataConfig(objectType, universeSymbol, security.Resolution, dataTimeZone, exchangeTimeZone, false, false, true, isFilteredSubscription: false);
+                universe = new BenchmarkUniverse(config);
```
mchandschuh: gorgeous!

Algorithm/QCAlgorithm.Universe.cs
```diff
@@ -419,6 +419,30 @@ public void AddUniverse(SecurityType securityType, string name, Resolution resol
...
+
+                UniverseManager.Add(universeSymbol, universe);
+            }
+
+            var benchmarkUniverse = universe as BenchmarkUniverse;
+            benchmarkUniverse?.AddBenchmarkSecurity(UtcTime, security);
```
mchandschuh: we should remove the `?` here -- if it's not a `BenchmarkUniverse` then I think it's correct to throw, perhaps something like this:
``` csharp
if (benchmarkUniverse == null)
{
    throw new Exception($"Expected BenchmarkUniverse with symbol {universeSymbol} but received {univerese.GetType().FullName}");
}
benchmarkUniveres.AddBenchmarkSecurity(UtcTime, security);
```

Common/Data/UniverseSelection/BenchmarkUniverse.cs
```diff
@@ -0,0 +1,107 @@
...
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Data.UniverseSelection
+{
+    /// <summary>
+    /// Represents the universe containing only the selected benchmark security.
+    /// This universe keeps the benchmark security and its subscription separate
+    /// from the algorithm securities. Universe selection will fire only once,
+    /// when the benchmark security is added to the universe.
```
mchandschuh: awesome doc! thank you -- will help us later :)

Common/Data/UniverseSelection/BenchmarkUniverse.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        /// <summary>
+        /// Event fired when a symbol is added or removed from this universe
+        /// </summary>
+        public event NotifyCollectionChangedEventHandler CollectionChanged;
+
+        /// <summary>
+        /// Gets the settings used for subscriptions added for this universe (unused in this universe)
+        /// </summary>
+        public override UniverseSettings UniverseSettings => null;
```
mchandschuh: Perhaps a comment why we're setting this to null -- and does this cause potential null ref exceptions elsewhere in the code -- and if not, maybe a little doc on *why* this is safe.

Common/Data/UniverseSelection/BenchmarkUniverse.cs
```diff
@@ -0,0 +1,107 @@
...
+            return Securities.Select(x => x.Value.Security).FirstOrDefault(x => x.Symbol == symbol);
+        }
+
+        /// <summary>
+        /// Adds the specified symbol to this universe
+        /// </summary>
+        /// <param name="utcTime">The current UTC time</param>
+        /// <param name="security">The symbol to be added to this universe</param>
+        /// <returns>True if the symbol was added, false if it was already present</returns>
+        public bool AddBenchmarkSecurity(DateTime utcTime, Security security)
```
mchandschuh: Perhaps `SetBenchmarkSecurity` -- when worded this way, and usage of `Securities` makes it look like this universe can support *multiple* securities as the benchmark.

Common/Data/UniverseSelection/BenchmarkUniverse.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            if (AddMember(utcTime, security))
+            {
+                OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Add, security.Symbol));
```
mchandschuh: Awesome usage of `INotifyCollectionChanged` -- hooks right into the existing infrastructure set up for `UserDefinedUniverse` 👍

Common/Data/UniverseSelection/BenchmarkUniverse.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /// <summary>
+        /// Event invocator for the <see cref="CollectionChanged"/> event
+        /// </summary>
+        /// <param name="e">The notify collection changed event arguments</param>
+        protected virtual void OnCollectionChanged(NotifyCollectionChangedEventArgs e)
+        {
+            var handler = CollectionChanged;
+            handler?.Invoke(this, e);
```
mchandschuh: no change required, but interestingly, when you do `instance?.Method()` the compiler expands it to:
``` csharp
var local = instance;
if (instance != null)
{
    instance.Method();
}
```
The compiler does this to avoid potential race conditions with setting `instance` to null after the null check but before the method invocation, so the saving to a local is redundant (and probably optimized away by the compiler anyway :) ).

Now, in the old days (before the null navigation operation `?.`) this was absolutely necessary given the potential for race conditions modifying the underlying storage (`MulticastDelegate`) to be null to clear the list of event handlers, but is no longer needed.

Common/Data/UniverseSelection/Universe.cs
```diff
@@ -206,6 +206,17 @@ public virtual Security CreateSecurity(Symbol symbol, IAlgorithm algorithm, Mark
...
         }
 
+        /// <summary>
+        /// Returns an existing security for this universe
+        /// </summary>
+        public virtual Security GetSecurity(Symbol symbol, IAlgorithm algorithm)
+        {
+            // by default return the security from the algorithm security collection
+            Security security;
+            algorithm.Securities.TryGetValue(symbol, out security);
```
mchandschuh: This wasn't initially how I saw this. I was originally thinking `GetSecurity` would *always* return a non-null security, creating if necessary. Setting it up like this does absolutely work but also provides an opportunity for a discontinuity between `GetSecurity` and `CreateSecurity` -- for example, `GetSecurity` could feasibly return a different security than the one that was created using `CreateSecurity` -- I'll ping you in slack to hash this one out -- I'm think it'll be good as is but want to cover our bases.

Common/Data/UniverseSelection/UserDefinedUniverse.cs
```diff
@@ -106,44 +106,7 @@ public UserDefinedUniverse(SubscriptionDataConfig configuration, UniverseSetting
...
-                    sid = SecurityIdentifier.GenerateCrypto(ticker, market);
-                    break;
-
-                case SecurityType.Commodity:
-                default:
-                    throw new NotImplementedException("The specified security type is not implemented yet: " + securityType);
-            }
-
-            return new Symbol(sid, ticker);
+            return CreateSymbol(ticker, securityType, market);
```
mchandschuh: nice consolidation of code duplication!

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -337,7 +337,10 @@ private Subscription CreateUniverseSubscription(SubscriptionRequest request)
             var config = request.Configuration;
 
             // define our data enumerator
-            var enumerator = GetEnumeratorFactory(request).CreateEnumerator(request, _dataProvider);
+            var factory = GetEnumeratorFactory(request);
+            if (factory == null) return null;
```
mchandschuh: maybe a comment here about why we may get a null enumerator for a subscription request -- if I'm understanding this correctly, it's essentially saying that we're *not* going to act on this subscription request.

Algorithm.CSharp/CustomUniverseWithBenchmarkRegressionAlgorithm.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </summary>
+        public override void Initialize()
+        {
+            UniverseSettings.Resolution = Resolution.Daily;
+
+            SetStartDate(2014, 4, 2);  //Set Start Date
+            SetEndDate(2014, 4, 6);    //Set End Date
+            SetCash(100000);           //Set End Date
+
+            AddUniverse("my-universe", x => new List<string> { "SPY" });
```
mchandschuh: Can we make this have return empty for the first couple of days and then add it -- I want to confirm that there isn't a race condition between this universe and the benchmark universe in creating the security object which determines which one makes it into the algorithm's security dictionary.

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -238,10 +237,14 @@ public SecurityChanges  ApplyUniverseSelection(Universe universe, DateTime dateT
...
-                if (!pendingAdditions.TryGetValue(symbol, out security) && !_algorithm.Securities.TryGetValue(symbol, out security))
+                if (!pendingAdditions.TryGetValue(symbol, out security))
                 {
-                    security = universe.CreateSecurity(symbol, _algorithm, _marketHoursDatabase, _symbolPropertiesDatabase);
-                    pendingAdditions.Add(symbol, security);
+                    security = universe.GetSecurity(symbol, _algorithm);
+                    if (security == null)
+                    {
+                        security = universe.CreateSecurity(symbol, _algorithm, _marketHoursDatabase, _symbolPropertiesDatabase);
+                        pendingAdditions.Add(symbol, security);
```
mchandschuh: RE: line 270
> additions.Add(security);

Should we be sending `SecuritiesChanged` events for the benchmark? Maybe a check for internals? I'm guessing in general we should be sending these events for any internal subscriptions (maybe a separate PR issue) -- but may cause issues in `OnSecuritiesChanged` if someone does something with the security and it's really the benchmark security and not the universe generated security.

StefanoRaggi: My initial thought was that internal/benchmark security additions should never fire `OnSecurityChanged`.
When `ApplyUniverseSelection` is called for the benchmark universe, the `SecurityChanges` return value is not currently used.

Common/Data/UniverseSelection/Universe.cs
StefanoRaggi: I did it this way to ensure the existing behavior in `UniverseSelection.ApplyUniverseSelection` is unchanged, just by moving the `TryGetValue` call into `Universe.GetSecurity` -- at least for this PR didn't want to change too much :)

Common/Data/UniverseSelection/BenchmarkUniverse.cs
StefanoRaggi: Good point!
The only potential null refs I can see are these two:
- in `Universe.CanRemoveMember`: should never be called, but solved by adding the override in `BenchmarkUniverse` 
- in `BacktestingSetupHandler.GetMaximumRuntime`

Thinking about it, might be safer to create an instance with default values (even if they're not actually used) ?

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -360,12 +367,37 @@ private Subscription CreateUniverseSubscription(SubscriptionRequest request)
...
+                    {
+                        if (args.Action != NotifyCollectionChangedAction.Add) return;
+
+                        var items = args.NewItems;
+                        if (items == null) return;
+
+                        var symbol = items.OfType<Symbol>().FirstOrDefault();
+                        if (symbol == null) return;
+
+                        var collection = new BaseDataCollection(_algorithm.UtcTime, symbol);
```
Martin-Molinero: I believe that if `Symbol` is of security type `future` or `option`, `collection` should be of type `FuturesChainUniverseDataCollection` or `OptionChainUniverseDataCollection`, else will throw.
Check this guy https://github.com/QuantConnect/Lean/blob/master/Engine/AlgorithmManager.cs#L811

mchandschuh:
Closing for now while we work on the [Data Feed Architecture Streamling Project](https://github.com/QuantConnect/Lean/projects/5). This change highlighted some deficiencies in the architecture -- we'll come back to this following architecture improvements.

PR #2439, opened 2018-08-23
Split dividend improvements

mchandschuh:
#### Description

* Add FactorFileRow.ToCsv(source) and fix Apply(Dividend) bug

    The time check in FactorFileRow.Apply(Dividend) was not using the previous
    trading date for comparison.

	Added a 'source' parameter to FactorFileRow.ToCsv() for tracking each row's
    data source.

* Make Dividend.ReferencePrice public set

    This is to be consistent with the other properties in Dividend

* Adds MapFile.FirstTicker

    The first ticker and first date are most commonly used to properly create
    security identifiers. Currently resolution of the first ticker is strewn
    about in various places and often times it doesn't take into account the
    case of an empty map file, where the first (and last) ticker are assumed
    to be the permtick after removing the '.<number>' specifier at the end.
    It's important to also not remove '.<letter>' -- as some securities are
    named as such.

* Add FactorFile.MostRecentFactorChange, bugfix for empty files

    Empty factor files (whether a single 2050 line or no lines), by definition don't
    have any splits or dividends.

    FactorFile.MostRecentFactorChange yields the maximum date that isn't the 2050 date

* Add FactorFile.MostRecentFactorChange, bugfix for empty files

    Empty factor files (whether a single 2050 line or no lines), by definition don't
    have any splits or dividends.

    FactorFile.MostRecentFactorChange yields the maximum date that isn't the 2050 date

#### Related Issue

See #2378 

#### Motivation and Context

Provides improvements to splits/dividends required by the new factor file generation process.

No.

Unit tests added for new features and bug fixes.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Auxiliary/FactorFile.cs
+++ b/Common/Data/Auxiliary/FactorFile.cs
@@ -50,6 +50,12 @@ namespace QuantConnect.Data.Auxiliary
         public DateTime? FactorFileMinimumDate { get; set; }
 
         /// <summary>
+        /// Gets the most recent factor change in the factor file
+        /// </summary>
+        public DateTime MostRecentFactorChange => SortedFactorFileData.Reverse()
+            .FirstOrDefault(kvp => kvp.Key != Time.EndOfTime).Key;
+
+        /// <summary>
         /// Gets the symbol this factor file represents
         /// </summary>
         public string Permtick { get; }
@@ -257,6 +263,12 @@ namespace QuantConnect.Data.Auxiliary
         public List<BaseData> GetSplitsAndDividends(Symbol symbol, SecurityExchangeHours exchangeHours)
         {
             var dividendsAndSplits = new List<BaseData>();
+            if (SortedFactorFileData.Count == 0)
+            {
+                Log.Trace($"{symbol} has no factors!");
+                return dividendsAndSplits;
+            }
+
             var futureFactorFileRow = SortedFactorFileData.Last().Value;
             for (var i = SortedFactorFileData.Count - 2; i >= 0 ; i--)
             {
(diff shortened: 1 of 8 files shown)


Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -180,15 +180,15 @@ public FactorFileRow Apply(Split split, SecurityExchangeHours exchangeHours)
                 throw new ArgumentException("Unable to apply split with reference price of zero.");
             }
 
+            var previousTradingDay = exchangeHours.GetPreviousTradingDay(split.Time);
+
             // this instance must be chronologically at or in front of the split
             // this is because the factors are defined working from current to past
-            if (Date < split.Time.Date)
+            if (Date < previousTradingDay)
```
jaredbroad: Was this a bug fix? Curious where the motivation for it came from?

Common/Data/Auxiliary/FactorFile.cs
```diff
@@ -49,6 +49,12 @@ public class FactorFile : IEnumerable<FactorFileRow>
         /// </remarks>
         public DateTime? FactorFileMinimumDate { get; set; }
 
+        /// <summary>
+        /// Gets the most recent factor change in the factor file
+        /// </summary>
+        public DateTime MostRecentFactorChange => SortedFactorFileData.Reverse()
+            .FirstOrDefault(kvp => kvp.Key != Time.EndOfTime).Key;
```
jaredbroad: Will (Default).Key be null exception?

mchandschuh: `KeyValuePair<TKey, TValue>` is a value type, so no null possible, in the case where's `SoratedFactFileData` is empty, this will return `default(DateTime)`

Common/Data/Auxiliary/FactorFileRow.cs
mchandschuh: the date in the factor file is the trading day _before_ the split/dividend event. I found this when writing a test to apply a split/dividend on the same day.

Issue #2453, opened 2018-08-27
Create new DataManager class


Martin-Molinero:
#### Expected Behavior

In the context of project https://github.com/QuantConnect/Lean/projects/5
- Create DataManager which will manage and own all subscriptions
#### Potential Solution

Implement DataManager class
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2455, opened 2018-08-27
Create DataManager

Martin-Molinero:
#### Description

- Adding a new `DataManager` class which will own the two current subscription containers, for the `SubscriptionManager` and the `DataFeed`. **The goal is to centralize all the subscription data**.
    - Adding a new interface for each consumer, `IDataFeedSubscriptionManager` and `IAlgorithmSubscriptionManager`
- Adjusting unit tests initialization
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2453
#### Motivation and Context

Project https://github.com/QuantConnect/Lean/projects/5

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/SubscriptionManager.cs
+++ b/Common/Data/SubscriptionManager.cs
@@ -33,12 +33,12 @@ namespace QuantConnect.Data
     {
         private readonly IAlgorithmSettings _algorithmSettings;
         private readonly TimeKeeper _timeKeeper;
-        private IDataManager _dataManager;
+        private IAlgorithmSubscriptionManager _subscriptionManager;
 
         /// <summary>
         /// Returns an IEnumerable of Subscriptions
         /// </summary>
-        public IEnumerable<SubscriptionDataConfig> Subscriptions => _dataManager.SubscriptionManagerSubscriptions;
+        public IEnumerable<SubscriptionDataConfig> Subscriptions => _subscriptionManager.SubscriptionManagerSubscriptions;
 
         /// <summary>
         /// Flags the existence of custom data in the subscriptions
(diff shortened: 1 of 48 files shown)


Common/Data/SubscriptionManager.cs
```diff
@@ -128,13 +136,13 @@ public SubscriptionDataConfig Add(Type dataType, TickType tickType, Symbol symbo
             var newConfig = new SubscriptionDataConfig(dataType, symbol, resolution, dataTimeZone, exchangeTimeZone, fillDataForward, extendedMarketHours, isInternalFeed, isCustomData, isFilteredSubscription: isFilteredSubscription, tickType: tickType);
 
             //Add to subscription list: make sure we don't have this symbol:
-            if (_subscriptions.ContainsKey(newConfig))
+            if (_dataManager.SubscriptionManagerContainsKey(newConfig))
```
mchandschuh: NOTE: These method namings are purposeful so that as we continue this refactoring/redesign we're very aware of what/where things are/coming from. When finalized, we'll go back through and take naming/API considerations into account, until then, it's scaffolding

Common/Interfaces/IDataManager.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace QuantConnect.Interfaces
+{
+    /// <summary>
+    /// DataManager interface will manage the subscriptions for both the DataFeeds and the SubscriptionManager
+    /// </summary>
+    public interface IDataManager
+    {
+        /// <summary>
+        /// Gets all the current data config subscriptions that are being processed
```
mchandschuh: This is a little misleading -- really it's all subscriptions added to the subscription manager -- comment of a similar color for the docs below as well.

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -67,14 +67,16 @@ public IEnumerable<Subscription> Subscriptions
         /// <summary>
         /// Initializes the data feed for the specified job and algorithm
         /// </summary>
-        public void Initialize(IAlgorithm algorithm, AlgorithmNodePacket job, IResultHandler resultHandler, IMapFileProvider mapFileProvider, IFactorFileProvider factorFileProvider, IDataProvider dataProvider)
+        public void Initialize(IAlgorithm algorithm, AlgorithmNodePacket job, IResultHandler resultHandler,
+                               IMapFileProvider mapFileProvider, IFactorFileProvider factorFileProvider,
+                               IDataProvider dataProvider, SubscriptionCollection subscriptionCollection)
```
mchandschuh: Perhaps instead of sending in the `SubscriptionCollection` we should be sending in the `DataManager` -- or perhaps `IDataManager` ?

_Eventually_ the storage should be hidden (encapsulation) and the required behaviors exposed.

Martin-Molinero: good point! I've updated it to receive the new `IDataFeedSubscriptionManager`

Common/Interfaces/IDataManager.cs
Martin-Molinero: I've updated the docs to be more explicit 👍. Also we will now have an interface for each consumer.

PR #2480, opened 2018-09-07
New Broker Support for Alpaca

umitanuki:
#### Description
This is to add new broker support for Alpaca. We followed the broker integration document instruction and the client SDK is placed as source rather than package install. Alpaca API supports WebSocket streaming for account and trade updates, which we use here.

#### Motivation and Context
Alpaca is a new broker with algo trading focus and free commission supporting US equities. Adding alternative to Interactive Brokers should give benefits to the Lean users.

The broker support is aligned to other broker supports, and the description for this should be placed next to them.

We have been testing this for a week now with live API and everything has been running well. If you need other type of testing, we are very happy to help and provide the support.

#### Types of changes

#### Checklist:
- [ ] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`
Not sure about the last one what we should do. Let me know if any change is required.


smartchris84:
Hi,

Thanks for having a chance to participate in the one of the greatest open source community.

I'm having some troubles understanding the error logs from the travis-ci, which wonderfully builds and tracks errors.

As you can see, the Build #7595 had only 1 error and I committed the fix `e60e4c8` shortly after(the last commit in this PR).

Then the travis build #7596 shows numerous errors, all of them showing the Lean path problem with bitfinex (I don't know why, sorry in advance). Which I can't understand at all.

The last commit was very simple, actually several lines of changes in two files, which has nothing to do bitfinex or so.

Did anyone have an experience with the above matter? Or Do I miss something critical?

Personally, I'd like to have Travis build the PR once again but it's beyond my control.

I may commit another nominal fix to make Travis build again but I think it's the last resort.

Please help me for this problem.

Thanks again for your patience and support.

Regards.

Brokerages/Alpaca/Markets/Enums/AccountStatus.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System.Runtime.Serialization;
```
StefanoRaggi: I'm assuming these files were sourced from here:
https://github.com/alpacahq/alpaca-trade-api-csharp
I think we should add a `readme.md` file in the `Lean/Brokerages/Alpaca/` folder including the GitHub URL above, just for reference and licensing.

Brokerages/Alpaca/PricingStreamSession.cs
```diff
@@ -0,0 +1,96 @@
...
+		public void StartSession()
+		{
+			_shutdown = false;
+
+			_client = _api.GetNatsClient();
+
+			_runningTask = Task.Run(() =>
+			{
+				_client.Open();
+                _client.QuoteReceived += QuoteReceived;
```
StefanoRaggi: I think we also need to attach an event handler for the `TradeReceived` event here, so we can receive both trades and quotes.

Common/Brokerages/AlpacaBrokerageModel.cs
```diff
@@ -0,0 +1,134 @@
...
+			{
+				message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
+					$"The {nameof(AlpacaBrokerageModel)} does not support {security.Type} security type."
+				);
+
+				return false;
+			}
+
+			// validate order type
+			if (order.Type != OrderType.Limit && order.Type != OrderType.Market && order.Type != OrderType.StopLimit)
```
StefanoRaggi: We should add `OrderType.StopMarket` and `OrderType.MarketOnOpen` here.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1140 @@
...
+			{
+				order = OrderProvider.GetOrderByBrokerageId(trade.Order.OrderId.ToString());
+			}
+			if (order != null)
+			{
+				Orders.OrderStatus status;
+                // Market orders are special: if the order was not in 'PartiallyFilledMarketOrders', means
+                // we already sent the fill event with OrderStatus.Filled, else it means we already informed the user
+                // of a partiall fill, or didn't inform the user, so we need to do it now
+                if (tradeEvent == "FILL")
```
StefanoRaggi: Is `trade.Event` set to `"fill"` for both partial and total fills?

umitanuki:
@StefanoRaggi Thanks for reviewing it! We'll work on it.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1140 @@
...
+					stopPrice = ((StopMarketOrder)order).StopPrice;
+					break;
+
+				case Orders.OrderType.StopLimit:
+					type = Markets.OrderType.StopLimit;
+					stopPrice = ((StopLimitOrder)order).StopPrice;
+					limitPrice = ((StopLimitOrder)order).LimitPrice;
+					break;
+				default:
+					throw new NotSupportedException("The order type " + order.Type + " is not supported.");
```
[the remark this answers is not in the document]
smartchris84: Thanks, will check now.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1140 @@
...
+			if (!order.BrokerId.Any())
+			{
+				Log.Trace("AlpacaBrokerage.CancelOrder(): Unable to cancel order without BrokerId.");
+				return false;
+			}
+
+			foreach (var orderId in order.BrokerId)
+			{
+				var res = restClient.DeleteOrderAsync(new Guid(orderId)).Result;
+				OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, 0, "Alpaca Cancel Order Event") { Status = Orders.OrderStatus.CancelPending });
```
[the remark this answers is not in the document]
smartchris84: As I'm new to QC framework, I was not aware of that.

I'll remove that part.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1140 @@
...
+
+		/// <summary>
+		/// Starts streaming prices for a list of instruments
+		/// </summary>
+		public void StartPricingStream(List<string> instruments)
+		{
+            Log.Trace("Start Pricing Stream");
+			_ratesSession = new PricingStreamSession(this, instruments);
+			_ratesSession.QuoteReceived += new Action<Markets.IStreamQuote>(OnPricingDataReceived);
+			_ratesSession.StartSession();
```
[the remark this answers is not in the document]
smartchris84: The `TradeReceived` event is separated for clarity and another handler is attached to it.

As it uses separate websocket endpoint and thus the api, the logic is as it is.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1140 @@
...
+
+		/// <summary>
+		/// Returns true if this brokerage supports the specified symbol
+		/// </summary>
+		private static bool CanSubscribe(Symbol symbol)
+		{
+			// ignore unsupported security types
+			if (symbol.ID.SecurityType != SecurityType.Equity)
+				return false;
+
```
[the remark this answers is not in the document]
smartchris84: Actually, I was confused about the strange subscriptions. Thanks!

Brokerages/Alpaca/Environment.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <summary>
+        /// A stable environment; recommended for testing with your fxTrade Practice account and your personal access token.
+        /// </summary>
+        Practice,
+
+        /// <summary>
+        /// A stable environment; recommended for production-ready code to execute with your fxTrade account and your personal access token.
+        /// </summary>
+        Trade
+    }
```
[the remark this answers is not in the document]
smartchris84: It should be removed, just I don't know how to remove from the commit.

Sorry.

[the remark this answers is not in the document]
StefanoRaggi: You can remove the file with `git rm filename` followed by a `git commit`

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
StefanoRaggi: Please correct me if I'm wrong, but the endpoint used in `SockClient.cs` is only used for streaming events for the account (order fills and account updates) and the endpoint used in `NatsClient.cs` is used for pricing data (including events for both trades and quotes).
For the latter, I think both `SubscribeQuote` and `SubscribeTrade` should be called in `PricingStreamSession` and both `QuoteReceived` and `TradeReceived` should be handled.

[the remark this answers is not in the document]
StefanoRaggi: Just to clarify, quote ticks with `TickType.Quote` should be emitted in the `QuoteReceived` event handler and trade ticks with `TickType.Trade` should be emitted in the `TradeReceived` event handler.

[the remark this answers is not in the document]
smartchris84: That is right, that's why `StartTransactionStream` and `StartPricingStream` functions are separated.

Brokerages/Alpaca/Environment.cs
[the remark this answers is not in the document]
smartchris84: Thanks.

Brokerages/Alpaca/Markets/Enums/AccountStatus.cs
umitanuki: We don't mind adding readme.md under Alpaca, but just wanted to make sure we understand the request correctly. I don't see any other broker code under `Brokerages` directory that has readme.md with the original SDK code. Do you still want it for Alpaca specifically, or you are ok with the latest change that just adds the link to the original SDK repository in the source comment?

StefanoRaggi: @umitanuki The repo link in the source header comment is totally fine, thanks.

umitanuki: 👍 Thanks for confirming

Brokerages/Alpaca/AlpacaBrokerageFactory.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            tradingMode = tradingMode.ToLower();
+            if (!tradingMode.Equals("live") && !tradingMode.Equals("paper"))
+            {
+                // if the trading mode is invalid, do not proceed further
+                throw new Exception("Available trading mode: paper/live");
+            }
+
+            var brokerage = new AlpacaBrokerage(algorithm.Transactions, algorithm.Portfolio, keyId, secretKey, tradingMode);
```
jaredbroad: Can you put the trading mode into the `BrokerageData` parameters and remove the base url? Given the baseUrl is hard coded in the constructor it looks like it shouldn't be in `BrokerageData`.

smartchris84: I guess I've put the trading mode into the BrokerageData and removed the baseUrl from it.

Did I do sth wrong?

Please let me know.  Thanks.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+            var orderByBrokerageId = OrderProvider.GetOrderByBrokerageId(id);
+            if (orderByBrokerageId != null)
+            {
+                qcOrder.Id = orderByBrokerageId.Id;
+            }
+
+            if (order.ExpiredAt != null)
+            {
+                qcOrder.Properties.TimeInForce = Orders.TimeInForce.GoodTilDate(order.ExpiredAt.Value);
```
mchandschuh: Awesome!!

Brokerages/Alpaca/AlpacaBrokerage.cs
```diff
@@ -0,0 +1,310 @@
...
+namespace QuantConnect.Brokerages.Alpaca
+{
+    /// <summary>
+    /// Alpaca Brokerage implementation
+    /// </summary>
+    [BrokerageFactory(typeof(AlpacaBrokerageFactory))]
+    public class AlpacaBrokerage : Brokerage, IDataQueueHandler
+    {
+        //private readonly OandaSymbolMapper _symbolMapper = new OandaSymbolMapper();
+        private readonly AlpacaApiBase _api;
```
mchandschuh: Why does `AlpacaApiBase` derive from `Brokerage` *and* `AlpacaBrokerage` derive from `Brokerage`? I think only `AlpacaBrokerage` should derive from `Brokerage` as the `AlpacaApiBase` isn't actually an `IBrokerage` w/ an associated factory. Perhaps what was intended was to have `AlpacaBrokerage` derive from `AlpacaApiBase` ?

mchandschuh: Also noting that many methods here simply pass through to `AlpacaApiBase` it looks like this type should probably derive from `AlpacaApiBase` -- thoughts?

Brokerages/Alpaca/Markets/Enums/AccountStatus.cs
mchandschuh: I think we'll need a statement in each file that was sourced from the `alpaca-trade-api-csharp` repository, preferably referencing the commit hash that they were pulled from, something like (with the correct commit hash):
``` csharp
// Sourced from:  https://github.com/alpacahq/alpaca-trade-api-csharp/tree/45cf16dc1c3ea631ed13f1bd9adee23058ae0bbb
```

I saw another file had the following header:
```
/*
 * The official C# API client for alpaca brokerage
 * https://github.com/alpacahq/alpaca-trade-api-csharp
*/
```
Just want to make sure consumers of this repository are aware of each and every file that is sourced from a separate repository for licensing purposes.

Common/Brokerages/AlpacaBrokerageModel.cs
```diff
@@ -0,0 +1,133 @@
...
+            {
+                message = new BrokerageMessageEvent(BrokerageMessageType.Warning, "NotSupported",
+                    $"The {nameof(AlpacaBrokerageModel)} does not support {order.Type} order type."
+                );
+
+                return false;
+            }
+
+            // validate time in force
+            if (order.TimeInForce != TimeInForce.GoodTilCanceled && order.TimeInForce != TimeInForce.Day)
```
mchandschuh: I'm not sure these equality checks will work as expected, perhaps type checks are intended? Accompanying unit tests for the brokerage model would help to alleviate this concern.

Common/Brokerages/AlpacaBrokerageModel.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        /// <summary>
+        /// Gets a new fee model that represents this brokerage's fee structure
+        /// </summary>
+        /// <param name="security">The security to get a fee model for</param>
+        /// <returns>The new fee model for this brokerage</returns>
+        public override IFeeModel GetFeeModel(Security security)
+        {
+            return new ConstantFeeModel(0m);
```
mchandschuh: Who doesn't love zero fees! Awesome!

Launcher/config.json
```diff
@@ -1,302 +1,323 @@
...
+      "setup-handler": "QuantConnect.Lean.Engine.Setup.BrokerageSetupHandler",
+      "result-handler": "QuantConnect.Lean.Engine.Results.LiveTradingResultHandler",
+      "data-feed-handler": "QuantConnect.Lean.Engine.DataFeeds.LiveTradingDataFeed",
+      "real-time-handler": "QuantConnect.Lean.Engine.RealTime.LiveTradingRealTimeHandler",
+      "transaction-handler": "QuantConnect.Lean.Engine.TransactionHandlers.BrokerageTransactionHandler",
+      "history-provider": "BrokerageHistoryProvider"
+    },
+
+    // defines the 'live-alpaca' environment
+    "live-alpaca": {
```
mchandschuh: awesome! 👍

Tests/app.config
```diff
@@ -20,7 +20,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="WebSocket4Net" publicKeyToken="eb4e154b696bf72a" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-0.14.1.0" newVersion="0.14.1.0" />
+        <bindingRedirect oldVersion="0.0.0.0-0.15.2.11" newVersion="0.15.2.11" />
```
mchandschuh: IIRC, the GDAX brokerage depends on this as well and I thought we had memory leaks when trying to upgrade it. We'll need to perform some stability/memory leak testing with GDAX brokerage before deploying this live.

@jaredbroad  -- do you remember this as well or am I mis-remembering?

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+
+        /// <summary>
+        /// Gets the list of available tradable instruments/products from Alpaca
+        /// </summary>
+        public List<string> GetInstrumentList()
+        {
+            try
+            {
+                CheckRateLimiting();
+                var response = restClient.ListAssetsAsync().Result;
```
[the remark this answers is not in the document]
smartchris84: I saw the resource you pointed. And they say when the context is needed, you can't use ConfigureAwait, and most of the cases here, we need the context.

And also, the msdn requires the use of async keyword in all async method, does it apply here?
If so, then I have to rename all the methods which doesn't make any sense.

Can you please point out how to deal with these?

Thanks.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+                        FillQuantity = Convert.ToInt32(trade.Order.Quantity)
+                    });
+                }
+                else if (tradeEvent == "CANCELED")
+                {
+                    OnOrderEvent(new OrderEvent(order, DateTime.UtcNow, 0, "Alpaca Cancel Order Event") { Status = Orders.OrderStatus.Canceled });
+                }
+                else if (tradeEvent == "ORDER_CANCEL_REJECTED")
+                {
+
```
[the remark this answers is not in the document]
smartchris84: It happened during the test, and I think it's useful to log this rare occurence.

Brokerages/Alpaca/AlpacaBrokerage.cs
smartchris84: Well, there is no need to derive AlpacaApiBase from AlpacaBrokerage. It was done early in the implementation and I'll remove that.

Common/Brokerages/AlpacaBrokerageModel.cs
smartchris84: Well, the validation of every parameter is needed in any brokerage, I think.

And I saw it on every brokerage here in QC.

Tests/app.config
smartchris84: Well, we have a sslProtocol issue with WebSocket4Net 14.1.0.

Actually, when we're trying to connect to websocket server, the server requires the specfic sslProtocols and I have found no way to configure it in WebSocket4Net 14.1.0.

If you can help me with this, I'll happy to use 14.1.0.

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
mchandschuh: > and most of the cases here, we need the context.

Where in this method (or elsewhere) do we depend on the originally captured [`SynchronizationContext`](https://msdn.microsoft.com/en-us/magazine/gg598924.aspx)?

It's very rare in library (dll) code that you require the original captured `SynchronizationContext`. `SynchronizationContext` is used to execute a delegate on the _original_ thread where the task was launched from. This is important when dealing with UI applications, or really any application who's `Main` is designed as an event message loop. The `SynchronizationContext` allows you to [`Post`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.synchronizationcontext.post?redirectedfrom=MSDN&view=netframework-4.7.2#System_Threading_SynchronizationContext_Post_System_Threading_SendOrPostCallback_System_Object_) a delegate to be executed on the original thread to do things such as setting a text field on a UI element. I did a quick search of the entire code base as of this PR and found no usages of `SynchronizationContext`.


I would go one step further and bury the `ConfigureAwait(false)` in the `AlpacaApiBase` implementation, and then this consuming code can, without fear of deadlock, use `Task.Result`.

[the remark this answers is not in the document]
mchandschuh: I've posted a PR https://github.com/QuantConnect/Lean/pull/2561 providing extension methods for safely resolving a task synchronously.

Brokerages/Alpaca/AlpacaBrokerage.cs
mchandschuh: I'm suggesting that `AlpacaBrokerage : AlpacaApiBase` -- then many of the pass-through methods in `AlpacaBrokerage` can simply be deleted (along with the `_api` member variable.

Tests/app.config
mchandschuh: I've made a note to research this further. Thank you for the additional information.

Common/Brokerages/AlpacaBrokerageModel.cs
mchandschuh: Those models may have bugs too :/

Consider the following failing unit test:

``` csharp
[Test]
public void NewDayTimeInForceEqualsStaticTimeInForceDay()
{
    var order = new MarketOrder();
    order.Properties.TimeInForce = new DayTimeInForce();
    Assert.AreEqual(order.TimeInForce, TimeInForce.Day);
}
```

Brokerages/Alpaca/AlpacaBrokerage.cs
mchandschuh: That being said, I agree that there's no reason for `AlpacaApiBase` to subclass `Brokerage` and prefer the compositional approach this PR uses (exception for the extra bit of inheritance by `AlpacaApiBase : Brokerage`)

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
mchandschuh: PR #2561 has been merged. If you refresh your `origin` and rebase you'll be able to use the new extension methods.

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
mchandschuh: Agreed, let's log it include data about the `tradeEvent`

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+            try
+            {
+                CheckRateLimiting();
+                var response = restClient.ListAssetsAsync().Result;
+
+                return response.Select(x => x.Symbol).ToList();
+            }
+            catch (Exception e)
+            {
+                if (e.InnerException != null)
```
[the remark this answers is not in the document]
smartchris84: As we're using Task for async operations, all exceptions are aggregated into a AggregateException.

In order to get precise exception, one must use InnerException. I guess the probable exception may be HttpException, like connection problem.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+        /// <param name="symbol">The symbol</param>
+        /// <param name="startTimeUtc">The starting time (UTC)</param>
+        /// <param name="endTimeUtc">The ending time (UTC)</param>
+        /// <param name="resolution">The requested resolution</param>
+        /// <param name="requestedTimeZone">The requested timezone for the data</param>
+        /// <returns>The list of bars</returns>
+        public IEnumerable<QuoteBar> DownloadQuoteBars(Symbol symbol, DateTime startTimeUtc, DateTime endTimeUtc, Resolution resolution, DateTimeZone requestedTimeZone)
+        {
+            var startUtc = startTimeUtc.ToString("yyyy-MM-ddTHH:mm:ssZ");
+            var endUtc = endTimeUtc.ToString("yyyy-MM-ddTHH:mm:ssZ");
```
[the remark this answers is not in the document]
smartchris84: According to Polygon API, the endpoint for QuoteBars doesn't require end date and the problem above doesn't exist here.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+                }
+
+                try
+                {
+                    while (!_cancellationTokenSource.IsCancellationRequested)
+                    {
+                        TimeSpan elapsed;
+                        lock (LockerConnectionMonitor)
+                        {
+                            elapsed = TimeSpan.FromSeconds(10); // DateTime.UtcNow - LastHeartbeatUtcTime;
```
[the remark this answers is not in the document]
mchandschuh: I've verified this has been resolved. ty.

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
mchandschuh: this works perfectly, well, thank you:
``` csharp
var task = restClient.ListAssetsAsync();
var response = Extensions.SynchronouslyAwaitTaskResult(task);
```
another way to write this is:
``` csharp
var response = restClient.ListAssetsAsync().SynchronouslyAwaitTaskResult();
```

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+        }
+
+        /// <summary>
+        /// Stops streaming transactions for the active account
+        /// </summary>
+        public void StopTransactionStream()
+        {
+            if (_eventsSession != null)
+            {
+                _eventsSession.StopSession();
```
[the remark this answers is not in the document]
mchandschuh: confirmed fix, ty.

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
mchandschuh: Can we also include in the log `JsonConvert.SerializeObject(trade)` -- this will ensure that if/when it happens we'll also have some context

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+
+        /// <summary>
+        /// Stops streaming prices for all instruments
+        /// </summary>
+        public void StopPricingStream()
+        {
+            if (_ratesSession != null)
+            {
+                Log.Trace("End Pricing Stream");
+                _ratesSession.StopSession();
```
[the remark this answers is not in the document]
mchandschuh: confirmed fix. ty.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+        {
+            LastHeartbeatUtcTime = DateTime.UtcNow;
+            var symbol = Symbol.Create(quote.Symbol, SecurityType.Equity, Market.USA);
+            var time = quote.Time;
+
+            // live ticks timestamps must be in exchange time zone
+            DateTimeZone exchangeTimeZone;
+            if (!_symbolExchangeTimeZones.TryGetValue(key: symbol, value: out exchangeTimeZone))
+            {
+                exchangeTimeZone = MarketHoursDatabase.FromDataFolder().GetExchangeHours(Market.USA, symbol, SecurityType.Equity).TimeZone;
```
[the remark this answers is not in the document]
mchandschuh: confirmed fix, ty

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+            while (true)
+            {
+                List<IHistoricalQuote> asList = new List<IHistoricalQuote>();
+                try
+                {
+                    CheckRateLimiting();
+                    var newBars = restClient.ListHistoricalQuotesAsync(symbol.Value, startTime, offsets).Result;
+                    asList = newBars.Items.ToList();
+
+                    // The first item in the HistoricalQuote is always 0 on BidPrice, so ignore it.
```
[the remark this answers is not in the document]
mchandschuh: this isn't holding anything up, but I am still very curious.

Only reference to this endpoint in the [alpaca api docs](https://docs.alpaca.markets/) is [here](https://docs.alpaca.markets/#tag/polygon) and doesn't really describe the endpoint.

Brokerages/Alpaca/AlpacaBrokerage.cs
mchandschuh: the new class hierarchy looks better, less simple patching of calls through, resolving.

Brokerages/Alpaca/AlpacaBrokerage.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        #region IBrokerage implementation
+
+        /// <summary>
+        /// Returns true if we're currently connected to the broker
+        /// </summary>
+        public override bool IsConnected
+        {
+            get { return IsConnected; }
```
mchandschuh: This causes a stack overflow exception -- please run all the tests to confirm recent changes haven't broken anytihng.

Brokerages/Alpaca/Markets/Enums/AccountStatus.cs
mchandschuh: Confirmed all files under `./Brokers/Alpaca/Markets` (86 in total) contain the following header:
```
/*
 * The official C# API client for alpaca brokerage
 * Sourced from: https://github.com/alpacahq/alpaca-trade-api-csharp/commit/161b114b4b40d852a14a903bd6e69d26fe637922
*/
```

Brokerages/Alpaca/AlpacaBrokerage.cs
smartchris84: Okay, let me check.

Brokerages/Alpaca/AlpacaApiBase.cs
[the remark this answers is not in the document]
mchandschuh: agreed, using `Log.Error(Exception)` will ensure that all exceptions are completely logged (including aggregates/inners/etc). There are still several cases of us not logging errors (do a search for `catch` and confirm that all catch blocks use `Log.Error(Exception)`, and if not, please provide a comment as to why that error is expected/not important and doesn't need to be logged.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+        public Dictionary<string, Tick> GetRates(List<string> instruments)
+        {
+            try
+            {
+                CheckRateLimiting();
+                var response = restClient.ListQuotesAsync(instruments).Result;
+                return response
+                    .ToDictionary(
+                        x => x.Symbol,
+                        x => new Tick { BidPrice = x.BidPrice, AskPrice = x.AskPrice }
```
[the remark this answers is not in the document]
mchandschuh: Looks like this was updated to use `TickType.Trade` but we're invoking `ListQuotesAsync`, so perhaps `TickType.Quote` is more appropriate here.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -0,0 +1,1219 @@
...
+
+                    // The first item in the HistoricalQuote is always 0 on BidPrice, so ignore it.
+                    asList.RemoveAt(0);
+
+                    offsets = asList.Last().TimeOffset;
+                    if (DateTimeHelper.FromUnixTimeMilliseconds(offsets) < startTimeUtc) continue;
+                }
+                catch (Exception e)
+                {
+                    if (e.InnerException != null && e.InnerException.Message.Contains("ticks"))
```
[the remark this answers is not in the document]
mchandschuh: appears this case has been removed, resolving as OBE.

Brokerages/Alpaca/AlpacaApiBase.cs
```diff
@@ -723,9 +729,10 @@ public IEnumerable<TradeBar> DownloadTradeBars(Symbol symbol, DateTime startTime
                 }
                 catch (Exception e)
                 {
+                    Log.Error(e);
                     if (e.InnerException != null)
                         OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Warning, 100, e.InnerException.Message));
-                    throw;
+                    throw e;
```
StefanoRaggi: I believe when re-throwing exceptions we should always use `throw;` instead of `throw e;`, since the latter resets the call stack to the current location and the original stack trace would be lost.

smartchris84: Oh I see.

Tests/app.config
smartchris84: I've looked at the GDAX implementation and it seems like they used WebSocketSharp instead of WebSocket4Net.

mchandschuh:
Closing as this PR is completely contained via https://github.com/QuantConnect/Lean/pull/2640

Issue #2513, opened 2018-09-17
Performing Cash Sync is executed multiple times


Martin-Molinero:
#### Expected Behavior

- `PerformingCashSync()` should only be executed once in the happy case, which is: when there haven't been fills in a 20 seconds window since the last sync.
#### Actual Behavior

- `PerformingCashSync()` is called multiple times in a 10 second window, due to `_lastSyncTimeTicks` being set 10 seconds later by the `Task.Delay`
#### Potential Solution

Set `_lastSyncTimeTicks` when setting `_syncedLiveBrokerageCashToday` to `true`
#### Reproducing the Problem

A live algorithm, each morning
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2519, opened 2018-09-18
Fix so we perform cash sync only once

Martin-Molinero:
#### Description

- Will now set `_lastSyncTimeTicks` right away
- Testing:
    - Lowering some protection level from `private` to `protected`, and adding `CurrentTimeUtc`, so I could unit test.
    - Adding unit tests which reproduce the issue and cover expected behavior
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2513
#### Motivation and Context

- Performing Cash Sync is executed multiple times

N/A

Unit and regression tests, live deployment
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
+++ b/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
@@ -137,7 +137,7 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
             // we don't need to do this today because we just initialized/synced
             _resultHandler = resultHandler;
             _syncedLiveBrokerageCashToday = true;
-            _lastSyncTimeTicks = DateTime.UtcNow.Ticks;
+            _lastSyncTimeTicks = CurrentTimeUtc.Ticks;
 
             _brokerage = brokerage;
             _brokerage.OrderStatusChanged += (sender, fill) =>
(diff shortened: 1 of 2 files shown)


Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -1086,19 +1086,24 @@ private void HandlePositionAssigned(OrderEvent fill)
...
         /// Gets the date of the last sync (New York time zone)
         /// </summary>
-        private DateTime LastSyncDate
+        protected DateTime LastSyncDate
         {
             get { return new DateTime(Interlocked.Read(ref _lastSyncTimeTicks)).ConvertFromUtc(TimeZones.NewYork).Date; }
         }
 
+        /// <summary>
+        /// Gets current time UTC. This is here to facilitate testing
```
mchandschuh: Thank you for the comment describing _why_ it's here :)

mchandschuh: might want to add a similar comment via `<remarks></remarks>` to the other two members who are now protected (and one other is virtual) -- presumably for the sole purpose for supporting unit tests.

no action require -- just me commenting --, taking hidden members and making them visible _just_ to support unit testing is usually indicative of a missing component/class/abstraction that would have public members. In this case it would be some 'thing' who's responsibility it is to manage the brokerage cash synchronization process.

Issue #2551, opened 2018-09-27
Add new ISubscriptionDataConfigBuilder 


Martin-Molinero:
#### Expected Behavior

- The new interface will be implemented by `DataManager` exposed through `SubscriptionManager` and consumed by `SecurityManager.CreateSecurity` and `SubscriptionManager`. In a next PR, its intended for it to also be consumed by `Universe.GetSubscriptionRequests()`
#### Actual Behavior

- `Universe.GetSubscriptionRequests()` does not have enough information to successfully create new SubscriptionRequests.
#### Potential Solution

- Refactor
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2552, opened 2018-09-27
Adding new ISubscriptionDataConfigService

Martin-Molinero:
#### Description

- Adding new `ISubscriptionDataConfigService` implemented by `DataManager`, exposed by
`SubscriptionManager`
- `SecurityManager` and `SubscriptionManager` will use new interface.
In a next `PR`, its intended for it to also be consumed by `Universe.GetSubscriptionRequests()`, since it does not have enough data/context to successfully create a new `SubscriptionDataConfig`
- Moving from `SubscriptionManager` into `DataManager`:
     - `LookupSubscriptionConfigDataTypes` implementation
     - `HasCustomData` , `TimeKeeper` and `AvailableDataTypes`

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2551
#### Motivation and Context

`Universe.GetSubscriptionRequests()` does not have enough data/context to successfully create a new `SubscriptionDataConfig`, this new interface will be provided as a parameter

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/SubscriptionManager.cs
+++ b/Common/Data/SubscriptionManager.cs
@@ -29,13 +29,12 @@ namespace QuantConnect.Data
     /// </summary>
     public class SubscriptionManager
     {
-        private readonly TimeKeeper _timeKeeper;
         private IAlgorithmSubscriptionManager _subscriptionManager;
 
         /// <summary>
-        ///     Instance that implements <see cref="ISubscriptionDataConfigBuilder" />
+        ///     Instance that implements <see cref="ISubscriptionDataConfigService" />
         /// </summary>
-        public ISubscriptionDataConfigBuilder SubscriptionDataConfigBuilder => _subscriptionManager;
+        public ISubscriptionDataConfigService SubscriptionDataConfigService => _subscriptionManager;
 
         /// <summary>
         ///     Returns an IEnumerable of Subscriptions
@@ -45,24 +44,16 @@ namespace QuantConnect.Data
         /// <summary>
         ///     Flags the existence of custom data in the subscriptions
         /// </summary>
-        public bool HasCustomData { get; set; }
+        public bool HasCustomData
+        {
+            get { return _subscriptionManager.HasCustomData; }
+            set { _subscriptionManager.HasCustomData = value; }
+        }
 
         /// <summary>
         ///     The different <see cref="TickType" /> each <see cref="SecurityType" /> supports
         /// </summary>
-        public Dictionary<SecurityType, List<TickType>> AvailableDataTypes { get; }
-
-        /// <summary>
-        ///     Initialise the Generic Data Manager Class
-        /// </summary>
-        /// <param name="timeKeeper">The algorithm's time keeper</param>
-        public SubscriptionManager(TimeKeeper timeKeeper)
-        {
-            _timeKeeper = timeKeeper;
-
-            // Initialize the default data feeds for each security type
-            AvailableDataTypes = DefaultDataTypes();
-        }
+        public Dictionary<SecurityType, List<TickType>> AvailableDataTypes => _subscriptionManager.AvailableDataTypes;
 
         /// <summary>
         ///     Get the count of assets:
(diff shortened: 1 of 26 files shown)


mchandschuh:
### NOTE: DO NOT MERGE -- PENDING INVESTIGATION INTO REGRESSION FAILURE

Common/Data/SubscriptionManager.cs
```diff
@@ -26,132 +25,129 @@
...
-        /// <param name="timeKeeper">The algorithm's time keeper</param>
-        public SubscriptionManager(TimeKeeper timeKeeper)
+        public bool HasCustomData
         {
-            _timeKeeper = timeKeeper;
-
-            // Initialize the default data feeds for each security type
-            AvailableDataTypes = DefaultDataTypes();
+            get { return _subscriptionManager.HasCustomData; }
+            set { _subscriptionManager.HasCustomData = value; }
```
mchandschuh: Should this be settable here?

Common/Securities/SecurityManager.cs
```diff
@@ -315,25 +315,15 @@ protected virtual void OnCollectionChanged(NotifyCollectionChangedEventArgs chan
...
-            if (!subscriptionDataTypes.Any())
-            {
-                throw new ArgumentNullException(nameof(subscriptionDataTypes), "At least one type needed to create security");
-            }
-
             // add the symbol to our cache
             if (addToSymbolCache) SymbolCache.Set(symbol.Value, symbol);
 
             // Add the symbol to Data Manager -- generate unified data streams for algorithm events
+            var configs = subscriptionManager.SubscriptionDataConfigService.Add(symbol, resolution, fillDataForward,
```
mchandschuh: awesome!

mchandschuh: we'll have to come back later and see what method args we can clean up -- yay! cleaning dependencies!

Engine/DataFeeds/DataManager.cs
```diff
@@ -19,39 +19,45 @@
...
     /// <summary>
-    /// DataManager will manage the subscriptions for both the DataFeeds and the SubscriptionManager
+    ///     DataManager will manage the subscriptions for both the DataFeeds and the SubscriptionManager
     /// </summary>
     public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionManager, IDataManager
     {
-        private readonly IDataFeed _dataFeed;
         private readonly IAlgorithmSettings _algorithmSettings;
+        private readonly IDataFeed _dataFeed;
+        private readonly MarketHoursDatabase _marketHoursDatabase = MarketHoursDatabase.FromDataFolder();
```
mchandschuh: Perhaps for another PR, but this should probably be a ctor arg

Common/Data/SubscriptionManager.cs
Martin-Molinero: The setter is used by this guy https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.cs#L1720 which is using it as a hack for when removing a security. I've opened a new card in the project to address removing `SubscriptionDataConfigs` and updating properties/consumers  https://github.com/QuantConnect/Lean/projects/5#card-13408875

Common/Securities/SecurityManager.cs
Martin-Molinero: Totally agree, I've also opened a new card for this https://github.com/QuantConnect/Lean/projects/5#card-13409112 . I'm thinking we could even add an overload which does not create the `SubscriptionDataConfigs` at all.

Common/Data/SubscriptionManager.cs
mchandschuh: RE: https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.cs#L1720
This might not be needed anymore? When the universe creates a subscription, does he go through the new factory?

Common/Interfaces/ITimeKeeper.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Interface implemented by <see cref="TimeKeeper"/>
+    /// </summary>
+    public interface ITimeKeeper
+    {
+        /// <summary>
+        /// Adds the specified time zone to this time keeper
+        /// </summary>
+        /// <param name="timeZone"></param>
+        void AddTimeZone(DateTimeZone timeZone);
```
mchandschuh: While this is the only value required by consumers of `ITimeKeeper` currently, perhaps it makes sense to also include the `GetLocalTimeKeeper` method and the `UtcTime` property

Engine/DataFeeds/DataManager.cs
```diff
@@ -60,12 +66,18 @@ public DataManager(IDataFeed dataFeed, UniverseSelection universeSelection, IAlg
         #region IAlgorithmSubscriptionManager
 
         /// <summary>
-        /// Gets all the current data config subscriptions that are being processed for the SubscriptionManager
+        ///     Flags the existence of custom data in the subscriptions
+        /// </summary>
+        public bool HasCustomData { get; set; }
```
mchandschuh: Eventually this `set` should become private.

Common/Interfaces/ITimeKeeper.cs
Martin-Molinero: Good idea, and I'll make `SecurityManager` consume the new interface too 👍

Common/Data/SubscriptionManager.cs
Martin-Molinero: Context around the `HasCustomData` flag: 
> This flag will be used to speed python algorithms execution, since it avoids a wrapping operation that is only required when there is custom data in python algorithms.

The setter is required **only** in the case a security is removed. If the security had a custom data subscription we will want to update the flag.
When universes create a subscription it will go through the new factory, call stack looks like: `SecurityManager.CreateSecurity` -> `subscriptionManager.SubscriptionDataConfigService.Add()`. In the next PR universes, will directly be calling `SubscriptionDataConfigService.Add()`. `SubscriptionDataConfigService.Add` is updating the `HasCustomData` flag internally.

I'm thinking maybe we could move `HasCustomData` directly to the `DF`, but either way I believe, we should tackle the fact we are not removing `SDC` from the `SubscriptionManager` (chat for another PR)

Issue #831, opened 2017-04-05
Parallel Runner Threading Issue


jaredbroad:
User reported issue:

```
2017-04-05T16:43:06.6406003Z ERROR:: ParallelRunnerController.ProcessHoldQueue():  System.ArgumentNullException: Value cannot be null.
  at System.Threading.SemaphoreSlim.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
  at System.Collections.Concurrent.BlockingCollection`1.TryTakeWithNoTimeValidation(T& item, Int32 millisecondsTimeout, CancellationToken cancellationToken, CancellationTokenSource combinedTokenSource)
  at System.Collections.Concurrent.BlockingCollection`1.<GetConsumingEnumerable>d__68.MoveNext()
  at QuantConnect.Util.ParallelRunnerController.ProcessHoldQueue(CancellationToken token) in Lean\Common\Util\ParallelRunner.cs:line 120
```
Error shown after the performance summary approximately 1/10 times,

Backtesting on Windows 7 on a Intel Core i7-6700 CPU @ 3.40 GHz. 30 tickers with 30 minute bars.

jaredbroad:
Fixed in PR #903

PR #2562, opened 2018-09-28
Dispose queues after consumers - Improve regression tests speed

Martin-Molinero:
#### Description

- Removing `sleep` and disposing queues after consumers. I believe this should also solve the original issue while saving time.
> For 100 regression tests, this reduces execution time by 16 minutes (100 * 10 sec / 60)
- Will first `Join` with 1-second timeout and `Abort` if it failed (tests showed `Abort` was never called). Adding try-catch just for robustness.
- Using `DisposeSafely` versus `Dispose`. 
- Adding check for `_waitHandle` to verify the handle is not already closed.
#### Related Issue

https://github.com/QuantConnect/Lean/issues/831
#### Motivation and Context

Faster is better

N/A

Executed unit and regressions tests (multiple times) and verified logs
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Util/ParallelRunner.cs
+++ b/Common/Util/ParallelRunner.cs
@@ -16,7 +16,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -172,13 +171,14 @@ namespace QuantConnect.Util
                     worker.DisposeSafely();
                 }
 
-                if (_holdQueue != null) _holdQueue.Dispose();
-                if (_processQueue != null) _processQueue.Dispose();
+                _holdQueue?.DisposeSafely();
+                _processQueue?.DisposeSafely();
 
-                if (_waitHandle != null)
+                // if IsClosed means its already disposed
+                if (_waitHandle != null && !_waitHandle.SafeWaitHandle.IsClosed)
                 {
                     _waitHandle.Set();
-                    _waitHandle.Dispose();
+                    _waitHandle.DisposeSafely();
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


Common/Util/ParallelRunner.cs
```diff
@@ -152,19 +152,16 @@ public void Dispose()
         {
             lock (_sync)
             {
-                if (_holdQueue != null) _holdQueue.Dispose();
-                if (_processQueue != null) _processQueue.Dispose();
-
-                // Wait for _holdQueue disposal be completed
-                Thread.Sleep(10000);
-
                 if (_processQueueThread != null && _processQueueThread.IsAlive) _processQueueThread.Abort();
```
StefanoRaggi: Love the sleep removal 👍
 
While we're touching this `Dispose` method, we could also try to terminate more gracefully, perhaps replacing the `Thread.Abort` with `Thread.Join` or something else?

Martin-Molinero: Good idea, I'll add a combination of both (`Join` with timeout) 👍

Common/Util/ParallelRunner.cs
```diff
@@ -152,19 +152,29 @@ public void Dispose()
...
+                        Log.Error(err);
+                    }
+                }
 
                 foreach (var worker in _workers)
                 {
                     worker.DisposeSafely();
                 }
 
+                if (_holdQueue != null) _holdQueue.Dispose();
```
mchandschuh: Please use the `DisposeSafely()` extension method when performing disposal -- helps to manage logging/error handling of exceptions thrown during disposal.

Common/Util/ParallelRunner.cs
mchandschuh: Great ideas guys, love it!

PR #2570, opened 2018-10-01
Refactor fill models to not depend on fee models

### LINKED ISSUE #2567: Refactor Fill Model Invocation
#### Expected Behavior

Fill models don't have dependency on fee models

#### Actual Behavior

Fill models are expected to invoke fee models directly.

#### Potential Solution

Invoke fill and fee models separately and compose the result. In order to maintain backwards compatibility, confirm that the fee hasn't been computed before computing and applying the fee.

#### Checklist

mchandschuh:
#### Description

Invocation of `IFeeModel` has been moved from the `IFillModel` implementations (and `IOptionExerciseModel` to the `BacktestingBrokerage` (or any other invoker of `IFillModel`). `BacktestingBrokerage` has also been updated to invoke `IFeeModel` for orders that have been completely filled, which is in-line with the current implementations of `IFillModel`.

We may want to remove the restriction of only invoking fee models on completely filled orders to support partial fees (common in crypto exchanges) as well as order submission fees (less common, but some brokerages charge these). For this PR, maintaining existing functionality was the primary concern as this is only a refactor of existing method call hierarchies. 

#### Related Issue

Closes #2567 

#### Motivation and Context

A recent attempt to improve the implementation of `IFeeModel` revealed a dependency issue between `IFillModel` and `IFeeModel`. Separating the invocation of `IFeeModel` from those who implement `IFillModel` will resolve this dependency issue. The assumption here is that user models should be directly invoked from LEAN infrastructure rather than from other models. By maintaining this invocation pattern, we free ourselves from requiring user models to manage dependencies that _other_ user models may require (now or in the future). 

YES! We should update our documentation regarding custom `IFillModel` implementations to use the `FillModel` base class as well as to **not** compute an order fee.

These changes are extensively covered by unit tests. Existing `ImmediateFillModelTests` were updated to confirm zero fees were returned.

#### Types of changes

- [x] Mechanical refactor

#### Checklist:

Existing test coverage more than sufficient for these changes


mchandschuh:
Shame on me for using travis as a test runner :/

Investigating unexpected test failures.

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -353,6 +353,20 @@ public void Scan()
...
                             }
+
+                            // invoke fee models for completely filled order events
+                            foreach (var fill in fills)
+                            {
+                                if (fill.Status == OrderStatus.Filled)
+                                {
+                                    // this check is provided for backwards compatibility of older user-defined fill models
+                                    // that may be performing fee computation inside the fill model w/out invoking the fee model
+                                    if (fill.OrderFee == 0m)
```
Martin-Molinero: What are your thoughts about checking whether or not `fill` is the new base class `FillModel` or a derivative of it, versus checking for order fee 0? Maybe using `IsAssignableFrom`

mchandschuh: great idea, since it's definitely possible that a custom `IFillModel` implemented zero fees explicitly and in contradiction to the configured `IFeeModel` -- I'll make this change 👍

mchandschuh: after more thought, I'm not sure this will be great as it would cause a bug if a user directly implements the `IFillModel` interface, for example, in python. Now, we could also check for the python type (which doesn't subclass `FillModel`), but I think for now checking zero is sufficient. I'll add a note that we should remove the check in 6 months

Issue #2571, opened 2018-10-01
Refactor Universe.GetSubscriptionRequests - Obsolete Universe.CreateSecurity


Martin-Molinero:
#### Expected Behavior

- Each `Universe` should be able to independently create its own desired `SubscriptionDataConfigs`, through calls to `Universe.GetSubscriptionRequests()`, and it also should be independent of using `Security.Subscriptions`.
- `Universes` will select the desired `Symbols`, the system will create the corresponding `Securities` and the `Universe` will be queried, for each `Security`, what are the desired `SDC`
#### Actual Behavior

- `Universe.GetSubscriptionRequests()` cannot create new `SubscriptionDataConfigs` and uses `Security.Subscriptions`.
- `Universe.CreateSecurity` is not compatible with a multiple `Universe` scenario.
#### Potential Solution

- Refactor `Universe.GetSubscriptionRequests()`
- Obsolete `Universe.CreateSecurity()`
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2572, opened 2018-10-02
Refactor Universe.GetSubscriptionRequests - Obsolete Universe.CreateSecurity

Martin-Molinero:
#### Description

- 1st commit:
   - Obsolete `Universe.CreateSecurity`, `Universe.SetSecurityInitializer`,
`Universe.GetSubscriptionRequests(Security security, DateTime currentTimeUtc, DateTime maximumEndTimeUtc)`
   - Adding new `GetSubscriptionRequests()` overload that will receive
instance which implementes `ISubscriptionDataConfigService`
   - `UserDefinedUniverse` and default `Universe` will stop using `Security.Subscriptions`
> `ManualUniverse` will continue to use `Security.Subscriptions` for now.

- 2nd commit:
   - `AddToUserDefinedUniverse()` will receive configuration settings as parameters.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2571
#### Motivation and Context

Unit and regression tests
#### Types of changes

#### Checklist:


Algorithm.Framework/Selection/ManualUniverse.cs
```diff
@@ -28,6 +29,7 @@ namespace QuantConnect.Algorithm.Framework.Selection
     /// </summary>
     public class ManualUniverse : UserDefinedUniverse
     {
+        /// <inheritdoc />
```
StefanoRaggi: I don't think our doc generator supports the `inheritdoc` XML tag yet

Martin-Molinero: Good catch, I'll remove the `inheritdoc` XML tag

Algorithm.Framework/Selection/ManualUniverse.cs
```diff
@@ -28,6 +29,9 @@ namespace QuantConnect.Algorithm.Framework.Selection
     /// </summary>
     public class ManualUniverse : UserDefinedUniverse
     {
+        /// <summary>
+        /// Creates a new instance of the ManualUniverse
```
mchandschuh: not need to change, in future please use the xml reference -- this provides linking behavior.
```
<see cref="ManualUniverse"/>
```

Algorithm/QCAlgorithm.Universe.cs
```diff
@@ -429,19 +434,26 @@ public void AddUniverse(SecurityType securityType, string name, Resolution resol
...
+        /// <param name="isExtendedMarketHours">ExtendedMarketHours send in data from 4am - 8pm, not used for FOREX</param>
+        /// <param name="isInternalFeed">Setting this flag to true will prevent the data from being sent into the algorithm's OnData methods</param>
+        private void AddToUserDefinedUniverse(Security security, Resolution resolution, decimal leverage, string market, bool isFillDataForward, bool isExtendedMarketHours, bool isInternalFeed)
         {
             // if we are adding a non-internal security which already has an internal feed, we remove it first
             Security existingSecurity;
             if (Securities.TryGetValue(security.Symbol, out existingSecurity))
             {
-                if (!security.IsInternalFeed() && existingSecurity.IsInternalFeed())
+                if (!isInternalFeed && existingSecurity.IsInternalFeed())
```
mchandschuh: awesome, getting rid of those `security.Property` references

Common/Data/UniverseSelection/FuturesChainUniverse.cs
```diff
@@ -32,31 +31,50 @@ public class FuturesChainUniverse : Universe
...
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FuturesChainUniverse"/> class
+        /// </summary>
+        /// <param name="future">The canonical future chain security</param>
+        /// <param name="universeSettings">The universe settings to be used for new subscriptions</param>
+        /// <param name="subscriptionManager">The subscription manager used to return available data types</param>
+        public FuturesChainUniverse(Future future,
+            UniverseSettings universeSettings,
+            SubscriptionManager subscriptionManager)
```
mchandschuh: is this `SubscriptionManager` required here? it seems out of place? Perhaps we only need certain functions available via an interface?

mchandschuh: I'm seeing that the previous ctor required this as well -- I think it's worth checking to see if this class only requires a subset and perhaps another PR we can modify this class to depend on an interface which provides the behavior -- this is all in an effort to remove algorithm layer objects as dependencies to various lean engine components, such as universes.

Also interesting that the options chain universe doesn't require `SubscriptionManager` -- I would have expected them to have the same dependencies.

Common/Data/UniverseSelection/OptionChainUniverse.cs
```diff
@@ -42,25 +41,48 @@ public class OptionChainUniverse : Universe
...
+            Option = option;
+            _underlyingSymbol = new[] { Option.Symbol.Underlying };
             _universeSettings = universeSettings;
             _liveMode = liveMode;
         }
 
+        /// <summary>
+        /// The canonical option chain security
+        /// </summary>
+        public readonly Option Option;
```
mchandschuh: for public members, please use properties, in this case, it can be get-only:
``` csharp
public Option Option { get; }
```

This same comment applies to the `Future` class

Common/Data/UniverseSelection/Universe.cs
```diff
@@ -227,6 +243,33 @@ public virtual IEnumerable<SubscriptionRequest> GetSubscriptionRequests(Security
...
+            DateTime currentTimeUtc,
+            DateTime maximumEndTimeUtc,
+            ISubscriptionDataConfigService subscriptionService)
+        {
+
+            var result = subscriptionService.Add(security.Symbol,
+                                                 UniverseSettings.Resolution,
+                                                 UniverseSettings.FillForward,
+                                                 UniverseSettings.ExtendedMarketHours);
+            return result.Select(config => new SubscriptionRequest(isUniverseSubscription: false,
```
mchandschuh: no need to change, but I tend to prefer left-aligned code as it prevents us ending up too far to the right
``` csharp
var result = subscriptionService.Add(
	security.Symbol,
	UniverseSettings.Resolution,
	UniverseSettings.FillForward,
	UniverseSettings.ExtendedMarketHours
);
	
return result.Select(config => new SubscriptionRequest(
	isUniverseSubscription: false,
	universe: this,
	security: security,
	configuration: config,
	startTimeUtc: currentTimeUtc,
	endTimeUtc: maximumEndTimeUtc
));
```

I find this style leads to more alignment between constructs, vs aligning with a parenthesis leads to various things not being aligned (I know, I'm picky with code formatting :P )

Common/Data/UniverseSelection/UserDefinedUniverse.cs
```diff
@@ -65,13 +70,14 @@ public override UniverseSettings UniverseSettings
...
         /// <param name="symbols">The initial set of symbols in this universe</param>
+        [Obsolete("This constructor is obsolete because SecurityInitializer is obsolete and will not be used.")]
         public UserDefinedUniverse(SubscriptionDataConfig configuration, UniverseSettings universeSettings, ISecurityInitializer securityInitializer, TimeSpan interval, IEnumerable<Symbol> symbols)
             : base(configuration, securityInitializer)
         {
             _interval = interval;
             _symbols = symbols.ToHashSet();
             _universeSettings = universeSettings;
-            _selector = time => _symbols;
+            _selector = time => _subscriptionDataConfigs.Select(x => x.Symbol).Union(_symbols);
```
mchandschuh: Was this `Union(_symbols)` required? Perhaps in the event a user implements a function that just returns symbols and we wouldn't have the requisite `SubscriptionDataConfig` instances? perhaps a comment here will suffice to make it clearer.

Or we can place the comment in the non-obsoleted constructor.

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -240,15 +241,28 @@ public SecurityChanges  ApplyUniverseSelection(Universe universe, DateTime dateT
                 Security security;
                 if (!pendingAdditions.TryGetValue(symbol, out security) && !_algorithm.Securities.TryGetValue(symbol, out security))
                 {
-                    security = universe.CreateSecurity(symbol, _algorithm, _marketHoursDatabase, _symbolPropertiesDatabase);
+                    security = SecurityManager.CreateSecurity(_algorithm.Portfolio, _algorithm.SubscriptionManager,
```
mchandschuh: no need to change -- I think this formatting showcases my earlier comment on preferring left-aligned code vs parenthesis aligned code. On the diff there's plenty of code scrolling off to the right, multiple parameters on each line, etc, just making it a little harder on the eyes

mchandschuh: also, a note for posterity, this is temporary code here as soon we'll replace all invocations of `SecurityManager.CreateSecurity` with an instance based approach, perhaps something like, `ISecurityService.CreateSecurity`

Algorithm.Framework/Selection/ManualUniverse.cs
Martin-Molinero: I'll add it, also I'll add the obsolete and new constructor 👍

Algorithm/QCAlgorithm.Universe.cs
Martin-Molinero: Yeah! 👍  were are getting there

Common/Data/UniverseSelection/FuturesChainUniverse.cs
Martin-Molinero: Your right! `SubscriptionManager` is not required, I'll remove it in the new constructor. I left it because I did not want to add more changes to this PR, but since you mention it I'll change it.
Totally agree about the interfaces 👍

Common/Data/UniverseSelection/Universe.cs
Martin-Molinero: No worries, I'll re format it :)

Common/Data/UniverseSelection/UserDefinedUniverse.cs
Martin-Molinero: I'll add a new comment explaining it 👍

Engine/DataFeeds/UniverseSelection.cs
Martin-Molinero: I'll reformat it 👍

Issue #2575, opened 2018-10-02
HistoryRequest using Resolution.Tick is fill forwared causing infitine loop situation


Martin-Molinero:
#### Expected Behavior

- No infinite loop situation
#### Actual Behavior

- HistoryRequest using Resolution.Tick is fill forwared causing infitine loop situation 
#### Potential Solution

- `Resolution.Tick` should never be fill forwared
#### Reproducing the Problem

```
from datetime import timedelta

class TickAlgorithm(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2018, 1, 1) 
        self.SetEndDate(2018, 1, 3) 
        self.SetCash(50000) 
        self.AddEquity("SPY", Resolution.Hour)
        self.flag = True 

    def OnData(self, data):
        if data.ContainsKey("SPY") and self.flag:
            hist = self.History(["SPY"], timedelta(seconds=2), Resolution.Tick)
            if not hist.empty:
                self.Debug(str(hist))
                self.flag = False
```
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2577, opened 2018-10-02
Add check to avoid FillForward for Resolution.Tick

Martin-Molinero:
#### Description

- Adding two checks to avoid setting FillForward resolution for
Resolution.Tick History requests
- Adding unit tests which reproduce the original issue

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2575
#### Motivation and Context

HistoryRequest using Resolution.Tick is fill forwared causing infitine loop situation

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/HistoryRequest.cs
+++ b/Common/Data/HistoryRequest.cs
@@ -16,9 +16,7 @@
 
 using System;
 using NodaTime;
-using QuantConnect.Data.Market;
 using QuantConnect.Securities;
-using QuantConnect.Util;
 
 namespace QuantConnect.Data
 {
@@ -27,6 +25,8 @@ namespace QuantConnect.Data
     /// </summary>
     public class HistoryRequest
     {
+        private Resolution? _fillForwardResolution;
+
         /// <summary>
         /// Gets the start time of the request.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.History.cs
```diff
@@ -608,7 +608,10 @@ private IEnumerable<HistoryRequest> CreateDateRangeHistoryRequests(IEnumerable<S
 
                     // apply overrides
                     var res = GetResolution(x, resolution);
-                    if (fillForward.HasValue) request.FillForwardResolution = fillForward.Value ? res : null;
+                    if (fillForward.HasValue)
+                    {
+                        request.FillForwardResolution = (fillForward.Value && resolution != Resolution.Tick) ? res : null;
```
mchandschuh: Wondering if this should be handled in the algorithm layer or should be handled via the `IHistoryProvider` implementation, or maybe even better, the `HistoryRequest` constructor could manage this -- and then it's in a single spot.

Martin-Molinero: Good idea! The properties are being set outside of the constructor, so I'll add an internal check to the `HistoryRequest.FillForwardResolution { get; }`. Why not the `{ set }`? Because it depends on the value of `HistoryRequest.Resolution`, but that value can change after the `FillForwardResolution` was set.

Common/Data/HistoryRequest.cs
```diff
@@ -53,9 +53,20 @@ public class HistoryRequest
...
-        /// Gets the requested fill forward resolution, set to null for no fill forward behavior
+        /// Gets the requested fill forward resolution, set to null for no fill forward behavior.
+        /// Will always return null when Resolution is set to Tick.
         /// </summary>
-        public Resolution? FillForwardResolution { get; set; }
+        public Resolution? FillForwardResolution
+        {
+            get
+            {
+                return Resolution == Resolution.Tick ? null : _fillForwardResolution;
```
mchandschuh: this is great, love it, ty

Issue #2495, opened 2018-09-12
Remove bridge from LiveTradingDataFeed - DF thread


Martin-Molinero:
#### Expected Behavior

In the context of project https://github.com/QuantConnect/Lean/projects/5
- `LiveTradingDataFeed` should yield return Slices directly from the enumerator, as done in the `FileSystemDataFeed`, so this logic can be extracted and shared.
- `DataFeed` does not need a running thread.
#### Actual Behavior

- `LiveTradingDataFeed` has a separate thread sending the slices to a bridge
#### Potential Solution

Remove `LiveTradingDataFeed` bridge and `DF` threads.
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2594, opened 2018-10-09
Removing DF thread - LTDF bridge

Martin-Molinero:
#### Description

- Removing `LiveTradingDataFeed` bridge. `LTDF` will yield return `Slices`
directly from the enumerator, as done in the `FileSystemDataFeed`. The objective is for
both to share the same logic and for it to be extracted in next PR.
- Removing now unnecessary `DataFeed` running thread
- Adding `if (_isStopping) { break; }` statement for `BaseDataExchange`. This is to avoid the thread to continue looping over the enumerators when stopping. Makes sense when there are many securities.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2495
#### Motivation and Context

N/A

Unit, regression and live trading tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/LiveTradingDataFeed.cs
+++ b/Engine/DataFeeds/LiveTradingDataFeed.cs
@@ -61,7 +61,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         private BaseDataExchange _customExchange;
         private SubscriptionCollection _subscriptions;
         private CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
-        private BusyBlockingCollection<TimeSlice> _bridge;
         private UniverseSelection _universeSelection;
         private DateTime _frontierUtc;
 
@@ -111,13 +110,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds
             _exchange.AddEnumerator(DataQueueHandlerSymbol, GetNextTicksEnumerator());
             _subscriptions = subscriptionManager.DataFeedSubscriptions;
 
-            _bridge = new BusyBlockingCollection<TimeSlice>();
             _universeSelection = subscriptionManager.UniverseSelection;
 
             // run the exchanges
             Task.Run(() => _exchange.Start(_cancellationTokenSource.Token));
             Task.Run(() => _customExchange.Start(_cancellationTokenSource.Token));
 
+            IsActive = true;
             // wire ourselves up to receive notifications when universes are added/removed
             var start = _timeProvider.GetUtcNow();
             algorithm.UniverseManager.CollectionChanged += (sender, args) =>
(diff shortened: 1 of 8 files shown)


Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -700,7 +627,61 @@ private IEnumerator<BaseData> GetNextTicksEnumerator()
...
+                    break;
+                }
+                // check for cancellation
+                if (_cancellationTokenSource.IsCancellationRequested) break;
+                // emit on data or if we've elapsed a full second since last emit or there are security changes
+                if (timeSlice.SecurityChanges != SecurityChanges.None || timeSlice.Data.Count != 0 || _frontierUtc >= nextEmit)
+                {
+                    yield return timeSlice;
+                    // force emitting every second
+                    nextEmit = _frontierUtc.RoundDown(Time.OneSecond).Add(Time.OneSecond);
```
mchandschuh: no need to change -- simply noting that this appears to be an _old_ bug (probably has my name all over it :D )

consider `_frontierUtc` is one tick below the round second, `nextEmit` will end up being only one tick ahead of `_frontierUtc` -- now, this is unlikely to happen in non-tick algorithms since data is always emitted on rounded seconds, but for algorithms with many tick subscriptions this could become a source of undue spinning

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -700,7 +627,61 @@ private IEnumerator<BaseData> GetNextTicksEnumerator()
...
+                }
+                // take a short nap
+                Thread.Sleep(1);
+            }
+            if (shouldSendExtraEmptyPacket)
+            {
+                // send last empty packet list before terminating,
+                // so the algorithm manager has a chance to detect the runtime error
+                // and exit showing the correct error instead of a timeout
+                nextEmit = _frontierUtc.RoundDown(Time.OneSecond).Add(Time.OneSecond);
```
mchandschuh: Thanks for the comment here -- does backtesting require something similar? Is this a design flaw in the synchronizer or algorithm manager for requiring this extra last point to 'turn the crank' to pick up any runtime errors?

Also curious how you encountered/noticed this guy

Tests/Engine/DataFeeds/CustomLiveDataFeedTests.cs
```diff
@@ -277,16 +276,6 @@ public void RemoteDataDoesNotIncreaseNumberOfSlices()
                 var request = new SubscriptionRequest(false, null, algorithm.Securities[symbol], config, startDate, Time.EndOfTime);
                 feed.AddSubscription(request);
             }
-
-            var feedThreadStarted = new ManualResetEvent(false);
-            Task.Factory.StartNew(() =>
-            {
-                feedThreadStarted.Set();
```
mchandschuh: awesome!

Engine/DataFeeds/LiveTradingDataFeed.cs
Martin-Molinero: He was already there before, just had to do some refactor due to the limitation of not being able to `yield return` inside a `try - catch` statement

Tests/Engine/DataFeeds/CustomLiveDataFeedTests.cs
Martin-Molinero: we are getting there! 😎

Issue #1418, opened 2017-12-20
Investigate currency unit discrepancies in various models


mchandschuh:
#### IOrderFeeModel
The `IOrderFeeModel` documentation clearly states that order fees are to be returned in unit of the account currency. Likewse, the `Order.GetValue(Security)` documentation states the same requirement. Upon a quick glance, it appears no `IOrderFeeModel` implementations actually guarantee that we convert to the account currency before returning the fees. This is paramount in accurate modeling of any securities with a non-account currency quote currency (mostly forex/cfd/crypto).

#### ISecurityMarginModel
The security margin model is used in such a way that it requires the results from certain methods to be in the same units under similar conditions. Specifically, the `GetInitialMarginRequiredForOrder` and `GetMarginRemaining` are expected to be in the same units as they're directly compared in `SecurityTransactionManager.GetSufficientCapitalForOrder`. This hidden coupling provides an argument for moving sufficient capital checks *into* the margin models themselves, which would make the units dependency between methods much more explicit while also potentially reducing the overall interface size.

#### The Big Picture
The issue of managing currency *units* in the code is becoming increasingly burdensome and error prone. I'm beginning to think that such implementations should return a `Cash` object (or similar) which would indicate an amount, the currency unit used, as well as a conversion to the account currency. This would remove the ambiguity while also allowing implementations to require fees be paid with certain currencies. I think this new pattern could effectively be applied anywhere we're returning opaque price data from models. Off the top of my head this includes `IOrderFeeModel` and `ISecurityMarginModel` -- possibly even `IMarginCallModel`.

##### Honorable Mention
Another issue noted was that margin calculations always seem to view all currencies as equivalent, but many brokerages aren't as kind as IB and actually require you to have the proper currencies available. Still, even other brokerages require paying fees in specific currencies (sometimes the quote currency and other times the account currency). Our model interface definitions and invocation patterns should be robust enough to support these cases, preferably by burying the technical details in the implementations of specific models themselves, thereby preventing the leaky abstraction.

jaredbroad:
Closing as 99% complete. Missing fees on currency conversions for internal transactions.

PR #2595, opened 2018-10-09
Add ICurrencyConverter dependencies to Security objects

mchandschuh:
#### Description

The scope of this change is small but the breadth is large, which is why I've decided to submit this as a standalone PR so it doesn't create noise for future changes that involve more involved logic/refactoring.

This change adds two new `ICurrencyConverter` implementations, an `IdentityCurrencyConverter` (currently unused), and an `ErrorCurrencyConverter`. The identity implementation returns the input `CashAmount` untouched unless the specified amount is not in units of the account currency, in which case it throws. This is a sensible default value that can be used in situations where we're guaranteed to be working in units of the account currency. This type is purposefully not used in LEAN infrastructure but is provided in the inevitable case that it's required to maintain backwards compatibility. The `ErrorCurrencyConverter` is intended to be used everywhere we assume that it won't be invoked, as it's implementation throws an exception immediately on every call. This implementation has widespread use among the exist unit tests and also in certain contexts where we create temporary security objects, for example, history requests, that won't require it's services. If in the future they do require these conversion services, we'll need to properly pipe the required algorithm dependencies into those providers.

#### Related Issue

See #1418 - this is but a small step towards satisfying the goals set out in this issue.

#### Motivation and Context

A (conceptually) small change to embed the `ICurrencyConverter` dependency privately inside security objects as it was identified as a place that requires it (for example, `TotalCloseProfit()`). This is being PR'd separately to avoid noise in future changes that have impact on behavior and logic.

No.

Wrote unit tests for new classes, as well as existing coverage via regression and existing unit tests.

#### Types of changes

- [x] New feature (non-breaking change which adds functionality) [on the way to a feature]

#### Checklist:

### DIFF
--- a/Common/Securities/ErrorCurrencyConverter.cs
+++ b/Common/Securities/ErrorCurrencyConverter.cs
@@ -1,0 +1,49 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="ICurrencyConverter"/> for use in
+    /// tests that don't depend on this behavior.
+    /// </summary>
+    public class ErrorCurrencyConverter : ICurrencyConverter
+    {
+        /// <summary>
+        /// Provides access to the single instance of <see cref="ErrorCurrencyConverter"/>.
+        /// This is done this way to ensure usage is explicit.
+        /// </summary>
+        public static ICurrencyConverter Instance = new ErrorCurrencyConverter();
+
+        private ErrorCurrencyConverter()
+        {
+        }
+
+        /// <summary>
+        /// Converts a cash amount to the account currency
+        /// </summary>
+        /// <param name="cashAmount">The <see cref="CashAmount"/> instance to convert</param>
+        /// <returns>A new <see cref="CashAmount"/> instance denominated in the account currency</returns>
+        public CashAmount ConvertToAccountCurrency(CashAmount cashAmount)
+        {
+            throw new InvalidOperationException($"This method purposefully throws as a proof that a " +
+                $"test does not depend on {nameof(ICurrencyConverter)}.If this exception is encountered, " +
+                $"it means the test DOES depend on {nameof(ICurrencyConverter)} and should be properly " +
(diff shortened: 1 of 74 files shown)


Common/Securities/ErrorCurrencyConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public class ErrorCurrencyConverter : ICurrencyConverter
+    {
+        public static ICurrencyConverter Instance = new ErrorCurrencyConverter();
+
+        private ErrorCurrencyConverter()
+        {
+        }
+
+        public CashAmount ConvertToAccountCurrency(CashAmount cashAmount)
```
StefanoRaggi: XML comments are missing here.

Common/Securities/ErrorCurrencyConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="ICurrencyConverter"/> for use in
+    /// tests that don't depend on this behavior.
+    /// </summary>
+    public class ErrorCurrencyConverter : ICurrencyConverter
```
StefanoRaggi: Nice idea 👍

QuantConnect.Lean.sln.DotSettings
```diff
@@ -1,3 +1,6 @@
-<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
     <s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String>
+	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_AFTER_INVOCATION_LPAR/@EntryValue">True</s:Boolean>
+	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_ARGUMENTS_STYLE/@EntryValue">CHOP_IF_LONG</s:String>
+	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_BEFORE_INVOCATION_RPAR/@EntryValue">True</s:Boolean>
```
StefanoRaggi: Love the new style settings ❤️

mchandschuh: yah! these were initially an accident, but then I realized we should all be using the same style settings anyways, so going to start porting my settings to the team-shared settings -- basically as I run into them, more coming to a `DotSettings` file near you! :)

Common/Securities/ErrorCurrencyConverter.cs
mchandschuh: thanks! great way to confirm we _don't_ depend on it, super useful in isolating tests that need it and those that dont

Common/Securities/ErrorCurrencyConverter.cs
mchandschuh: will remedy. ty.

Issue #2596, opened 2018-10-10
Duplicate effort between ManualUniverse and UserDefinedUniverse


Martin-Molinero:
#### Expected Behavior

Users subscription requests are added just once.
#### Actual Behavior

There is a duplicate effort/situation between `ManualUniverse` and `UserDefinedUniverse` (Addxxx calls). Example is the `CompositeAlphaModelFrameworkAlgorithm.cs` or `ConvertToFrameworkAlgorithm.cs`. Both universes are trying to add the same SubscriptionRequest
#### Potential Solution

Review and refactor.
#### Reproducing the Problem

`CompositeAlphaModelFrameworkAlgorithm.cs` or `ConvertToFrameworkAlgorithm.cs`
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2598, opened 2018-10-10
Fix ManualUniverse duplicate effort

Martin-Molinero:
#### Description

- Removed usages of `algorithm.Securities.key` as a parameter for the
`ManualUniverseSelectionModel()` since those securities, added through
`AddXXXX` calls will be managed by the `UserDefinedUniverse`. This was
causing both Universes to try to add the same subscription requests
- Adding new empty constructor for `ManualUniverseSelectionModel`,
required for Python
- ManualUniverse will return any existing SDC for the symbol (when security.subscriptions is removed it will be able to return empty). This is for maintaining existing behavior and preventing breaking changes: Specifically motivated by usages of `Algorithm.Securities.Keys` as constructor parameter of the `ManualUniverseSelectionModel`, since those `Symbols` added by `Addxxx()` calls will already be managed by the UserDefinedUniverse
- Making some format modifications to aling with used Lean formatting

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2596
#### Motivation and Context

This change helps in the effort of stop using `Security.Subscriptions`

Adding a note to the constructors of `ManualUniverseSelectionModel()`

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Selection/ManualUniverse.cs
+++ b/Algorithm.Framework/Selection/ManualUniverse.cs
@@ -27,6 +27,8 @@ namespace QuantConnect.Algorithm.Framework.Selection
     /// Defines a universe as a set of manually set symbols. This differs from <see cref="UserDefinedUniverse"/>
     /// in that these securities were not added via AddSecurity.
     /// </summary>
+    /// <remarks>Incompatible with multiple <see cref="Universe"/> selecting the same <see cref="Symbol"/>.
+    /// with different <see cref="SubscriptionDataConfig"/>. More information <see cref="GetSubscriptionRequests"/></remarks>
     public class ManualUniverse : UserDefinedUniverse
     {
         /// <summary>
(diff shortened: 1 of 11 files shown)


Algorithm.Framework/Selection/ManualUniverseSelectionModel.cs
```diff
@@ -39,7 +39,17 @@ public class ManualUniverseSelectionModel : UniverseSelectionModel
         /// Initializes a new instance of the <see cref="ManualUniverseSelectionModel"/> class using the algorithm's
         /// security initializer and universe settings
         /// </summary>
-        /// <param name="symbols">The symbols to subscribe to</param>
+        public ManualUniverseSelectionModel()
```
mchandschuh: This constructor isn't really needed since there's the `params Symbol[]` constructor which would gladly accept no arguments. That being said, I do like the explicitness of this.

Engine/DataFeeds/DataManager.cs
```diff
@@ -66,18 +66,26 @@ public DataManager(IDataFeed dataFeed, UniverseSelection universeSelection, IAlg
...
         public IEnumerable<SubscriptionDataConfig> SubscriptionManagerSubscriptions =>
             _subscriptionManagerSubscriptions.Select(x => x.Key);
 
         /// <summary>
-        ///     Gets existing or adds new <see cref="SubscriptionDataConfig" />
+        /// Gets a list of all registered <see cref="SubscriptionDataConfig"/> for a given <see cref="Symbol"/>
+        /// </summary>
+        public List<SubscriptionDataConfig> GetSubscriptionDataConfigs(Symbol symbol)
+        {
+            return SubscriptionManagerSubscriptions.Where(x => x.Symbol == symbol).ToList();
```
mchandschuh: just opinions :)

In general, I usually prefer to just return the enumerable as moves the `ToList()` decision to the caller. For example, if the caller immediately wants to group the result of this it will end up looping over the result set twice. No need to change anything here, and sometimes, it's really nice to return the already materialized list instead of an enumerable, which is really just a definition and not a collection. A really good example where you always want to return a list is if there's synchronization issues or if the underlying collection doesn't permit modification during enumeration. The underlying storage collection here, IIRC, is a `ConcurrentDictionary` so he doesn't really care either way.

Engine/DataFeeds/DataManager.cs
```diff
@@ -175,10 +183,18 @@ public int SubscriptionManagerCount()
             var result = (from subscriptionDataType in dataTypes
                 let dataType = subscriptionDataType.Item1
                 let tickType = subscriptionDataType.Item2
-                select new SubscriptionDataConfig(dataType, symbol, resolution, marketHoursDbEntry.DataTimeZone,
-                    exchangeHours.TimeZone, fillForward, extendedMarketHours,
-                    isInternalFeed, isCustomData,
-                    isFilteredSubscription: isFilteredSubscription, tickType: tickType,
+                select new SubscriptionDataConfig(
```
mchandschuh: I don't know about you, but this formatting makes it _soo_ much easier to actually read and see all of the parameters. Thanks!

Algorithm.Framework/Selection/ManualUniverse.cs
```diff
@@ -60,18 +62,26 @@ public class ManualUniverse : UserDefinedUniverse
...
-                    security: security,
-                    configuration: new SubscriptionDataConfig(config, isInternalFeed: false),
-                    startTimeUtc: currentTimeUtc,
-                    endTimeUtc: maximumEndTimeUtc
-                )
-            );
+            // ManualUniverse will return any existing SDC for the symbol, else will create new, using universe settings.
+            // This is for maintaining existing behavior and preventing breaking changes: Specifically motivated
+            // by usages of Algorithm.Securities.Keys as constructor parameter of the ManualUniverseSelectionModel.
+            // Symbols at Algorithm.Securities.Keys added by Addxxx() calls will already be added by the UserDefinedUniverse.
```
mchandschuh: Thank you for adding this note. Future generations (or maybe future us) will thank you greatly :)

Issue #2608, opened 2018-10-15
Create new Synchronizer - TimeSlice streamer


Martin-Molinero:
#### Expected Behavior

- Create new class `Synchronizer` which will provide the shared implementation of `IEnumerable<TimeSlice> StreamData()` for both `LiveTrading` and `FileSystem` `DataFeeds`
#### Actual Behavior

- Both `FileSystemDF` and `LiveTradingDF` have duplicated code
#### Potential Solution

- Refactor, extract and consolidate existing TimeSlice streaming code.
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2609, opened 2018-10-16
Creating new Synchronizer - TimeSlice streamer

Martin-Molinero:
#### Description

- Creating new `Synchronizer` which will consolidate and combine `TimeSlice`
streaming for both live and backtests modes. Will live in front of the
`DataManager`
- `SubscriptionSynchronizer` will be the `FrontierTimeProvider` exposed,
and owned, by the new `Synchronizer`
- Updating existing tests

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2608
#### Motivation and Context

Both live and backtest modes will now share the same TimeSlice streaming loop.

N/A

Unit tests, regression tests and live deployments
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/IDataFeedTimeProvider.cs
+++ b/Engine/DataFeeds/IDataFeedTimeProvider.cs
@@ -17,13 +17,18 @@
 namespace QuantConnect.Lean.Engine.DataFeeds
 {
     /// <summary>
-    /// IDataManager is the engines view of the Data Manager.
+    /// Reduced interface which exposes required <see cref="ITimeProvider"/> for <see cref="IDataFeed"/> implementations
     /// </summary>
-    public interface IDataManager
+    public interface IDataFeedTimeProvider
     {
         /// <summary>
-        /// Get the universe selection instance
+        /// Continuous UTC time provider
         /// </summary>
-        UniverseSelection UniverseSelection { get; }
+        ITimeProvider TimeProvider { get; }
+
+        /// <summary>
+        /// Time provider which returns current UTC frontier time
+        /// </summary>
+        ITimeProvider FrontierTimeProvider { get; }
     }
 }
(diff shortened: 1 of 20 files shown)


Engine/DataFeeds/IDataFeedTimeProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    /// <summary>
+    /// Reduced interface which exposes required <see cref="ITimeProvider"/> for <see cref="IDataFeed"/> implementations
+    /// </summary>
+    public interface IDataFeedTimeProvider
+    {
+        /// <summary>
+        /// Continuous UTC time provider
+        /// </summary>
+        ITimeProvider TimeProvider { get; }
```
mchandschuh: interesting? who outside the DF would need access to this?

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -495,13 +481,13 @@ private Subscription CreateUniverseSubscription(SubscriptionRequest request)
                             args.Action == NotifyCollectionChangedAction.Add ? args.NewItems :
                             args.Action == NotifyCollectionChangedAction.Remove ? args.OldItems : null;
 
-                        if (items == null || _frontierUtc == DateTime.MinValue) return;
+                        if (items == null || _frontierTimeProvider.GetUtcNow() == DateTime.MinValue) return;
```
mchandschuh: let's extract `_frontierTimeProvider.GetUtcNow()` into a local variable so we're not invoking it three times here

Engine/DataFeeds/IDataFeedTimeProvider.cs
Martin-Molinero: Nobody else, only the `DF` and today actually only the `LiveTradingDataFeed`. Sending it directly through the `IDataFeed.Initialize()`

Engine/DataFeeds/LiveTradingDataFeed.cs
Martin-Molinero: Whops! Shame on me, I'll update it 👍

PR #2643, opened 2018-10-24
Feature: Maximum Portfolio Drawdown

dd-w:
#### Description

Adds RiskManagementModel for Maximum Portfolio Drawdown

#### Motivation and Context

Limits potential loss impact of misbehaving Algorithm

Regression Test

#### Types of changes

#### Checklist:

- [ ] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`

### DIFF
--- a/Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
+++ b/Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
@@ -1,0 +1,78 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System.Collections.Generic;
+using QuantConnect.Algorithm.Framework;
+using QuantConnect.Algorithm.Framework.Alphas;
+using QuantConnect.Algorithm.Framework.Execution;
+using QuantConnect.Algorithm.Framework.Portfolio;
+using QuantConnect.Algorithm.Framework.Risk;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Show cases how to use the <see cref="CompositeRiskManagementModel"/> to define
+    /// </summary>
+    public class MaximumPortfolioDrawdownFrameworkAlgorithm : QCAlgorithmFramework, IRegressionAlgorithmDefinition
+    {
+        public override void Initialize()
+        {
+            // Set requested data resolution
+            UniverseSettings.Resolution = Resolution.Minute;
+
+            SetStartDate(2013, 10, 07);  //Set Start Date
+            SetEndDate(2013, 10, 11);    //Set End Date
+            SetCash(100000);             //Set Strategy Cash
+
+            // set algorithm framework models
+            SetUniverseSelection(new ManualUniverseSelectionModel(QuantConnect.Symbol.Create("SPY", SecurityType.Equity, Market.USA)));
+            SetAlpha(new ConstantAlphaModel(InsightType.Price, InsightDirection.Up, System.TimeSpan.FromMinutes(20), 0.025, null));
+            SetPortfolioConstruction(new EqualWeightingPortfolioConstructionModel());
+            SetExecution(new ImmediateExecutionModel());
+
+            // define risk management model as a composite of several risk management models
+            SetRiskManagement(new CompositeRiskManagementModel(
+                new MaximumDrawdownPercentPortfolio(0.01m)
+            ));
+        }
+
+        /// <summary>
+        /// This is used by the regression test system to indicate if the open source Lean repository has the required data to run this algorithm.
+        /// </summary>
+        public bool CanRunLocally { get; } = true;
+
+        /// <summary>
+        /// This is used by the regression test system to indicate which languages this algorithm is written in.
+        /// </summary>
+        public Language[] Languages { get; } = { Language.CSharp };
+
+        /// <summary>
+        /// This is used by the regression test system to indicate what the expected statistics are from running the algorithm
+        /// </summary>
+        public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>
+        {
+            {"Total Trades", "2"},
+            {"Compounding Annual Return", "-90.716%"},
+            {"Drawdown", "1.600%"},
+            {"Expectancy", "-1"},
+            {"Net Profit", "-1.030%"},
+            {"Loss Rate", "100%"},
+            {"Win Rate", "0%"},
+            {"Profit-Loss Ratio", "0"},
+        };
+    }
+}
--- a/Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
+++ b/Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
@@ -1,0 +1,108 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using System.Collections.Generic;
+using QuantConnect.Algorithm.Framework.Portfolio;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Framework.Risk
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="IRiskManagementModel"/> that limits the drawdown
+    /// per holding to the specified percentage
+    /// </summary>
+    public class MaximumDrawdownPercentPortfolio : RiskManagementModel
+    {
+        private readonly decimal _maximumDrawdownPercent;
+        private decimal _startingValue;
+        private bool _initialised = false;
+        private string _email;
+        private string _phoneNumber;
+
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MaximumDrawdownPercentPortfolio"/> class
+        /// </summary>
+        /// <param name="maximumDrawdownPercent">The maximum percentage drawdown allowed for algorithm portfolio
+        /// compared with starting value, defaults to 5% drawdown</param>
+        /// <param name="email">Optional - Email address for Algorithm Termination Notifications</param>
+        /// <param name="phoneNumber">Optional - Phone Number for Algorithm Termination SMS Notifications</param>
+        public MaximumDrawdownPercentPortfolio(
+            decimal maximumDrawdownPercent = 0.05m, string email = "", string phoneNumber = ""
+            )
+        {
+            _maximumDrawdownPercent = -Math.Abs(maximumDrawdownPercent);
+            _email = email;
+            _phoneNumber = phoneNumber;
+        }
+
+        /// <summary>
+        /// Manages the algorithm's risk at each time step
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance</param>
+        /// <param name="targets">The current portfolio targets to be assessed for risk</param>
+        public override IEnumerable<IPortfolioTarget> ManageRisk(QCAlgorithmFramework algorithm, IPortfolioTarget[] targets)
+        {
+            if(!_initialised)
+            {
+                _startingValue = algorithm.Portfolio.TotalPortfolioValue; // Set inital portfolio value
+                _initialised = true;
+            }
+
+            var pnl = TotalDrawdownPercent(algorithm.Portfolio);
+            if (pnl < _maximumDrawdownPercent)
+            {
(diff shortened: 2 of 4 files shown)


Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
```diff
@@ -0,0 +1,78 @@
...
+        public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>
+        {
+            {"Total Trades", "2"},
+            {"Compounding Annual Return", "-90.716%"},
+            {"Drawdown", "1.600%"},
+            {"Expectancy", "-1"},
+            {"Net Profit", "-1.030%"},
+            {"Loss Rate", "100%"},
+            {"Win Rate", "0%"},
+            {"Profit-Loss Ratio", "0"},
```
AlexCatarino: Did this algorithm pass the regression test with only part of the expected statistics?

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// compared with starting value, defaults to 5% drawdown</param>
+        /// <param name="email">Optional - Email address for Algorithm Termination Notifications</param>
+        /// <param name="phoneNumber">Optional - Phone Number for Algorithm Termination SMS Notifications</param>
+        public MaximumDrawdownPercentPortfolio(
+            decimal maximumDrawdownPercent = 0.05m, string email = "", string phoneNumber = ""
+            )
+        {
+            _maximumDrawdownPercent = -Math.Abs(maximumDrawdownPercent);
+            _email = email;
+            _phoneNumber = phoneNumber;
```
AlexCatarino: Despite the fact that it is optional, the pattern that we are following does not include notifications. An user can create a new class that inherits from this one and add the notification helpers if he wants. Or he can look for orders event in `OnOrdersEvent` and send the notification from there.

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        private void KillAlgorithm(QCAlgorithmFramework algorithm)
+        {
+            // Liqudate
+            algorithm.Log($"Liquidating All Holdings");
+            algorithm.Liquidate();
+
+            // Stop algorithm
+            algorithm.Quit("Terminated Algorithm due to Drawdown");
```
AlexCatarino: The framework models in Lean should not terminate the algorithm. I think that liquidating all (or just some) positions should be enough.

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                _startingValue = algorithm.Portfolio.TotalPortfolioValue; // Set inital portfolio value
+                _initialised = true;
+            }
+
+            var pnl = TotalDrawdownPercent(algorithm.Portfolio);
+            if (pnl < _maximumDrawdownPercent)
+            {
+                var message = $"Maximum Portfolio Drawdown Reached - Drawdown: {pnl.SmartRounding()}%, Drawdown Limit: {_maximumDrawdownPercent.SmartRounding()}%";
+                algorithm.Log(message);
```
AlexCatarino: The pattern that we are following for framework models does not log messages.

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
dd-w: I can do that but my issues is I don't know what the Order Execution Model will do. It might be expecting to manage all the orders and if I Liquidate in the Risk Model, it might mess up the execution model - so that's why I called Quit, to avoid the Execution Model doing anything unpredictable.

Alternatively I could just change all the Targets to 0 quantity but then it's relying on the Execution Model to do the orders which take some time... 

Which approach would you prefer?

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
dd-w: Can you please tell me where I can see this "pattern"? 

Can I use a different Logging method? Maybe "Debug" or "Error"? Also there is a "DebugMode" flag can I use that? It's all quite confusing as to what the recommended way is.

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
dd-w: For me the ideal implementation would be for the user to specify their contact details (Email, SMS or both) for "Critical Messages" or maybe just "Errors" and then in the Framework model the developer can output message to this logging level.

AlexCatarino: We have been trying to separate concerns with the framework models. The risk model, for instance, will just create `PortfolioTarget` objects that reduce the Quantity of each asset.

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
AlexCatarino: The right approach follows the separation of concerns and rely on the execution model to liquidate the positions, so you should change all the targets to zero.
Currently, you are using `Liquidate` method. What if the users disables it?
```csharp
    Settings.LiquidateEnabled = false;
}
```
and pick an execution model that does not use `Liquidate()`?

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
AlexCatarino: You can see the pattern by expecting the other models. :-)
We would rather not having logging. We just log exceptions, for example:
```csharp
if (insight.Magnitude == null) {
    algorithm.SetRunTimeError(new ArgumentNullException("BlackLittermanOptimizationPortfolioConstructionModel does not accept \'null\' as Insight.Magnitude. Please make sure your Alpha Model is generating Insights with the Magnitude property set."));
}
```
because the code will exit and let the user know what he had done wrong (in the case above, used an invalid alpha model).

jaredbroad:
Awesome PR @dd-w thank you for diving into the framework!

We're building a marketplace soon which will let you post highly customized/feature rich models like this - which might not be a fit for _everyone_. The requests Alex is asking for is to make it the most applicable to all users.

The marketplace will let you sell and list for free modules like this that people can plug into their algorithms. The alerts then would be a feature; and the quitting of the algorithm would be a design of that module which users could give you feedback on.

Its up to you - if interested to merge into LEAN; I'd suggest making it conform to LEAN design by removing the liquidate / logging etc. If you'd like to keep it intact, we'll be building the marketplace in the next quarter.

dd-w:
@jaredbroad thanks for the feedback and I look forward to the support for modular models.
I would like to contribute back to LEAN and hopefully expand it's capabilities. The issues you described (Liquidate / Logging) are just because I didn't know what the best practises or design patterns are, do you have this information that you can share with potential contributors?

I think for development being able to output debug messages is very helpful. Even if they are disabled by default. Is there a capability to do this currently?

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
dd-w: OK that makes sense, I'll remove it

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+
+        private decimal TotalDrawdownPercent(SecurityPortfolioManager portfolio)
+        {
+            var currentValue = portfolio.TotalPortfolioValue;
+            return (currentValue / _startingValue) - 1.0m;            
+        }
+
+        private decimal TotalDrawdownValue(SecurityPortfolioManager portfolio)
```
AlexCatarino: I think you can remove this method since it is not being used.

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            var pnl = TotalDrawdownPercent(algorithm.Portfolio);
+            if (pnl < _maximumDrawdownPercent)
+            {
+                foreach(var target in targets)
+                    yield return new PortfolioTarget(target.Symbol, 0);
+            }
+        }
+
+        private decimal TotalDrawdownPercent(SecurityPortfolioManager portfolio)
```
AlexCatarino: I would rename the method to `GetTotalDrawdownPercent` and accept a decimal (algorithm.Portfolio.TotalPortfolioValue):
```csharp 
private decimal GetTotalDrawdownPercent(decimal currentValue)
{
    return (currentValue / _startingValue) - 1.0m;
}
```

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,77 @@
...
+using System;
+using System.Collections.Generic;
+using QuantConnect.Algorithm.Framework.Portfolio;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Framework.Risk
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="IRiskManagementModel"/> that limits the drawdown
+    /// of the portfolio to the specified percentage - it terminates the algorithm if this limt is reached
```
AlexCatarino: You can remove the " - it terminates the algorithm if this limt is reached" part. ;-)

dd-w:
@AlexCatarino thanks for the comments. I look forward to the Contributing Readme update - it will be very helpful.

Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
```diff
@@ -0,0 +1,78 @@
...
+using QuantConnect.Algorithm.Framework.Execution;
+using QuantConnect.Algorithm.Framework.Portfolio;
+using QuantConnect.Algorithm.Framework.Risk;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Show cases how to use the <see cref="CompositeRiskManagementModel"/> to define
```
mchandschuh: Looks like some copy pasta here in the doc :)

Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
```diff
@@ -0,0 +1,78 @@
...
+            SetCash(100000);             //Set Strategy Cash
+
+            // set algorithm framework models
+            SetUniverseSelection(new ManualUniverseSelectionModel(QuantConnect.Symbol.Create("SPY", SecurityType.Equity, Market.USA)));
+            SetAlpha(new ConstantAlphaModel(InsightType.Price, InsightDirection.Up, System.TimeSpan.FromMinutes(20), 0.025, null));
+            SetPortfolioConstruction(new EqualWeightingPortfolioConstructionModel());
+            SetExecution(new ImmediateExecutionModel());
+
+            // define risk management model as a composite of several risk management models
+            SetRiskManagement(new CompositeRiskManagementModel(
```
mchandschuh: Since this algorithm is showcasing the `MaximumDrawdownPercentPortfolio` risk model, perhaps we don't need to add the extra layer of using the `CompositeRiskManagementModel` here

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (pnl < _maximumDrawdownPercent)
+            {
+                foreach(var target in targets)
+                    yield return new PortfolioTarget(target.Symbol, 0);
+            }
+        }
+
+        private decimal GetTotalDrawdownPercent(decimal currentValue)
+        {
+            return (currentValue / _startingValue) - 1.0m;
```
mchandschuh: This appears to be a 'global' drawdown, meaning from the time the algorithm first launched. Do we need a way to 'reset' the `currentValue` ? For example, consider if an algorithm runs for a year it may be up 100% w/ this model set to a 5% max drawdown and then from there drop 50% and this check still wouldn't trigger a liquidate -- is that intended? If this _is_ intended, perhaps some extra documentation describing that.

I think the ideal here would be to change `currentValue` to be maximum value, so on each iteration check if `TotalPortfolioValue` is greater than the previous and reset it. We could also set it each time we perform a global liquidate to set it to the new lower threshold value.

thoughts?

Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
dd-w: oh sorry

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
dd-w: Yes, I like that implementation. To me it sounds like a Trailing Stop Loss rather than an Absolute Stop Loss. Perhaps a user would use both Risk Management Models - with the Absolute Stop Loss set to a lower drawdown than the Trailing? If so, I could rename this implementation to make it clear it's an Absolute Stop Loss rather than Trailing - then implement Trailing at a later time?

Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
mchandschuh: In general, it's best to have a regression test focused on a particular feature/model instead of piling them into a single one. This helps ensure that in isolation the model works as expected _and_ that the test actually exercises the model's logic

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.cs
mchandschuh: Naming and documentation are key! So long as it's properly named/documented I'm happy with whichever route you choose 👍

dd-w: @mchandschuh I've added a constructor flag so both options can be used and documented it. Is this fine or should it really be two separate classes?

Algorithm.CSharp/MaximumPortfolioDrawdownFrameworkAlgorithm.cs
dd-w: @mchandschuh that makes sense, it just seemed to me that the Algos in `QuantConnect.Algorithm.CSharp` were meant as examples for users and not really there for test coverage. So I was hesitate to create unnecessary bulk.

I've updated the regression test to cover the two usages of `MaximumDrawdownPercentPortfolio` or should there be separate regression tests?

dd-w:
Thanks @mchandschuh!

Issue #2627, opened 2018-10-19
Move AddSubscription and RemoveSubscription from IDataFeed into DataManager 


Martin-Molinero:
#### Expected Behavior

- `AddSubscription()` and `RemoveSubscription()` logic is shared and not duplicate in `FileSystem` and `LiveTrading` data feeds.
- `DataManager` is completly in front of `IDataFeed`
#### Actual Behavior

- Logic is duplicated
#### Potential Solution

- Refactor
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2647, opened 2018-10-26
Move DataManager in front of DataFeed

Martin-Molinero:
#### Description

- Completly move `DataManager` in front of `DataFeed`. Specifically
`AddSubscription()` and `RemoveSubscription()` implementations. Also
moving `IDataFeed.Subscriptions`

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2627
#### Motivation and Context

- Make IDataFeed implementation specific
- Avoid logic duplication, improvement can be seen deletions > additions

N/A

Unit, regression and live deployment tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/DataManager.cs
+++ b/Engine/DataFeeds/DataManager.cs
@@ -164,8 +164,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
             Log.Debug($"DataManager.AddSubscription(): Added {request.Configuration}." +
                 $" Start: {request.StartTimeUtc}. End: {request.EndTimeUtc}");
-            DataFeedSubscriptions.TryAdd(subscription);
-            return true;
+            return DataFeedSubscriptions.TryAdd(subscription);
         }
 
         /// <summary>
@@ -206,9 +205,10 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                 }
                 subscription.Dispose();
                 Log.Debug($"DataManager.RemoveSubscription(): Removed {configuration}");
+                return true;
             }
 
-            return true;
+            return false;
         }
 
         #endregion
--- a/Engine/DataFeeds/DataManager.cs
+++ b/Engine/DataFeeds/DataManager.cs
@@ -17,9 +17,11 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Data.Auxiliary;
+using QuantConnect.Data.UniverseSelection;
 using QuantConnect.Interfaces;
 using QuantConnect.Logging;
 using QuantConnect.Securities;
@@ -48,16 +50,66 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         public DataManager(
             IDataFeed dataFeed,
             UniverseSelection universeSelection,
-            IAlgorithmSettings algorithmSettings,
+            IAlgorithm algorithm,
             ITimeKeeper timeKeeper,
             MarketHoursDatabase marketHoursDatabase)
         {
             _dataFeed = dataFeed;
             UniverseSelection = universeSelection;
-            _algorithmSettings = algorithmSettings;
+            UniverseSelection.SetDataManager(this);
+            _algorithmSettings = algorithm.Settings;
             AvailableDataTypes = SubscriptionManager.DefaultDataTypes();
             _timeKeeper = timeKeeper;
             _marketHoursDatabase = marketHoursDatabase;
+
+            var liveStart = DateTime.UtcNow;
+            // wire ourselves up to receive notifications when universes are added/removed
+            algorithm.UniverseManager.CollectionChanged += (sender, args) =>
+            {
+                switch (args.Action)
+                {
+                    case NotifyCollectionChangedAction.Add:
+                        foreach (var universe in args.NewItems.OfType<Universe>())
+                        {
+                            var config = universe.Configuration;
+                            var start = algorithm.LiveMode ? liveStart : algorithm.UtcTime;
+
+                            var end = algorithm.LiveMode ? Time.EndOfTime
+                                : algorithm.EndDate.ConvertToUtc(algorithm.TimeZone);
+
+                            Security security;
+                            if (!algorithm.Securities.TryGetValue(config.Symbol, out security))
+                            {
+                                // create a canonical security object if it doesn't exist
+                                security = new Security(
+                                    _marketHoursDatabase.GetExchangeHours(config),
+                                    config,
+                                    algorithm.Portfolio.CashBook[CashBook.AccountCurrency],
+                                    SymbolProperties.GetDefault(CashBook.AccountCurrency),
+                                    algorithm.Portfolio.CashBook
+                                 );
+                            }
+                            AddSubscription(
+                                new SubscriptionRequest(true,
+                                    universe,
+                                    security,
+                                    config,
+                                    start,
+                                    end));
+                        }
+                        break;
+
+                    case NotifyCollectionChangedAction.Remove:
+                        foreach (var universe in args.OldItems.OfType<Universe>())
+                        {
+                            RemoveSubscription(universe.Configuration);
+                        }
+                        break;
+
+                    default:
+                        throw new NotImplementedException("The specified action is not implemented: " + args.Action);
+                }
+            };
         }
 
         #region IDataFeedSubscriptionManager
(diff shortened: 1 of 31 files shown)


Engine/DataFeeds/DataManager.cs
```diff
@@ -48,16 +50,66 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
...
             UniverseSelection universeSelection,
-            IAlgorithmSettings algorithmSettings,
+            IAlgorithm algorithm,
             ITimeKeeper timeKeeper,
             MarketHoursDatabase marketHoursDatabase)
         {
             _dataFeed = dataFeed;
             UniverseSelection = universeSelection;
-            _algorithmSettings = algorithmSettings;
+            UniverseSelection.SetDataManager(this);
```
mchandschuh: I would typically perform this type of initialization from the call site as it would become more obvious if someone's setting it twice. Also, it's typically good to 'wire' up components outside of the component itself -- although, holy wars have been fought over the concepts of self-registration, but my position is that it's not the components job to decide it's registrations/wirings

Engine/DataFeeds/DataManager.cs
```diff
@@ -67,6 +119,98 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Adds a new <see cref="Subscription"/> to provide data for the specified security.
+        /// </summary>
+        /// <param name="request">Defines the <see cref="SubscriptionRequest"/> to be added</param>
+        /// <returns>True if the subscription was created and added successfully, false otherwise</returns>
+        public bool AddSubscription(SubscriptionRequest request)
```
mchandschuh: This is great -- subscription management is more of a `DataManager` concern than an `IDataFeed` concern.

Engine/DataFeeds/DataManager.cs
```diff
@@ -67,6 +119,98 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
...
+        {
+            if (DataFeedSubscriptions.Contains(request.Configuration))
+            {
+                // duplicate subscription request
+                return false;
+            }
+
+            var subscription = _dataFeed.CreateSubscription(request);
```
mchandschuh: eventually it _may_ make sense to have an `ISubscriptionFactory` with two implementations, live and backtest, further removing subscription management responsibilities from the data feed -- when we get to that point I suspect the data feeds will be able to collapse into each other :)

Engine/DataFeeds/DataManager.cs
```diff
@@ -67,6 +119,98 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
...
+            {
+                Log.Trace($"DataManager.AddSubscription(): Unable to add subscription for: {request.Configuration}");
+                // subscription will be null when there's no tradeable dates for the security between the requested times, so
+                // don't even try to load the data
+                return false;
+            }
+
+            Log.Debug($"DataManager.AddSubscription(): Added {request.Configuration}." +
+                $" Start: {request.StartTimeUtc}. End: {request.EndTimeUtc}");
+            DataFeedSubscriptions.TryAdd(subscription);
```
mchandschuh: should we be handling the `false` case here? perhaps this is as simple as `return DataFeedSubscriptions.TryAdd(subscription)` to ensure the return value of this function is correct

Engine/DataFeeds/DataManager.cs
```diff
@@ -67,6 +119,98 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
...
+        /// Removes the <see cref="Subscription"/>, if it exists
+        /// </summary>
+        /// <param name="configuration">The <see cref="SubscriptionDataConfig"/> of the subscription to remove</param>
+        /// <returns>True if the subscription was successfully removed, false otherwise</returns>
+        public bool RemoveSubscription(SubscriptionDataConfig configuration)
+        {
+            // remove the subscription from our collection, if it exists
+            Subscription subscription;
+
+            if (DataFeedSubscriptions.TryGetValue(configuration, out subscription))
```
mchandschuh: Should the `else` condition here be `return false` since we didn't remove it?

Engine/DataFeeds/DataManager.cs
```diff
@@ -67,6 +119,98 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
...
+                    return false;
+                }
+
+                if (!DataFeedSubscriptions.TryRemove(configuration, out subscription))
+                {
+                    Log.Error($"DataManager.RemoveSubscription(): Unable to remove {configuration}");
+                    return false;
+                }
+
+                _dataFeed.RemoveSubscription(subscription);
```
mchandschuh: In `AddSubscription` we only interact with the `DataFeedSubscriptions` but here we're removing directly from `IDataFeed` -- why the inconsistency? Also, should be handling the return value from `RemoveSubscription` in the event it wasn't removed from the DF?

Engine/DataFeeds/FileSystemDataFeed.cs
```diff
@@ -89,53 +81,9 @@ public IEnumerable<Subscription> Subscriptions
...
-                        break;
-
-                    default:
-                        throw new NotImplementedException("The specified action is not implemented: " + args.Action);
-                }
-            };
         }
 
-        private Subscription CreateSubscription(SubscriptionRequest request)
+        private Subscription InternalCreateSubscription(SubscriptionRequest request)
```
mchandschuh: Maybe instead of `InternalCreateSubscription` this is `CreateDataSubcription` which is different from `CreateUniverseSubscription` ? Just thinking of ways to keep the verbiage between the two options consistent and descriptive (universe vs data subscriptions)

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -114,116 +106,40 @@ public bool IsActive
...
         }
 
         /// <summary>
         /// Removes the subscription from the data feed, if it exists
         /// </summary>
-        /// <param name="configuration">The configuration of the subscription to remove</param>
-        /// <returns>True if the subscription was successfully removed, false otherwise</returns>
-        public bool RemoveSubscription(SubscriptionDataConfig configuration)
+        /// <param name="subscription">The subscription to remove</param>
+        public void RemoveSubscription(Subscription subscription)
```
mchandschuh: Interesting to note that the removal semantics are different for backtesting and live -- no need to change as part of this PR (or maybe at all), but worth thinking of ideas for how a subscription can clean himself up -- perhaps live subscriptions implement a `Dispose` method that performs the behavior required here in `RemoveSubscription` -- this would end up (methinks) with removing `IDataFeed.RemoveSubscription`

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -44,18 +44,28 @@ public class UniverseSelection
...
         }
 
+        /// <summary>
+        /// Sets the data manager
+        /// </summary>
+        public void SetDataManager(IDataFeedSubscriptionManager dataManager)
+        {
+            if (_dataManager != null)
+            {
+                throw new Exception("UniverseSelection.SetDataManager(): can only be set once");
```
mchandschuh: Nice -- an earlier comment made note of this possibility.

Engine/Results/BacktestingResultHandler.cs
```diff
@@ -129,10 +129,9 @@ public BacktestingResultHandler()
         /// <param name="job">Algorithm job packet for this result handler</param>
         /// <param name="messagingHandler">The handler responsible for communicating messages to listeners</param>
         /// <param name="api">The api instance used for handling logs</param>
-        /// <param name="dataFeed"></param>
         /// <param name="setupHandler"></param>
         /// <param name="transactionHandler"></param>
-        public virtual void Initialize(AlgorithmNodePacket job, IMessagingHandler messagingHandler, IApi api, IDataFeed dataFeed, ISetupHandler setupHandler, ITransactionHandler transactionHandler)
+        public virtual void Initialize(AlgorithmNodePacket job, IMessagingHandler messagingHandler, IApi api, ISetupHandler setupHandler, ITransactionHandler transactionHandler)
```
mchandschuh: Awesome!!! Love seeing parameter counts get reduced :P

Engine/DataFeeds/DataManager.cs
Martin-Molinero: Yes, good point. I'll modify it 👍

Engine/DataFeeds/DataManager.cs
Martin-Molinero: Yes, good point. I'll modify it 👍

Engine/DataFeeds/DataManager.cs
Martin-Molinero: Both while adding and removing we interact with both `DataFeedSubscriptions` and `IDataFeed`. While adding we call `_dataFeed.CreateSubscription()` and `DataFeedSubscriptions.TryAdd()`. The `DF` does not handle the Subscription at this point anymore. Its used as a cleanup process, which only is used in the `LiveTradingDataFeed`

Engine/DataFeeds/FileSystemDataFeed.cs
Martin-Molinero: Good idea 👍 I'll change it

Engine/DataFeeds/LiveTradingDataFeed.cs
Martin-Molinero: Good idea! That would make it look even cleaner

Engine/DataFeeds/DataManager.cs
mchandschuh: very awesome, a future improvement could be implementing the live-specific clean up as a dispose method and that _should_ end up by removing `RemoveSubscription` from `IDataFeed` which _should_ completely de-couple the data feeds from subscription manage 🎆

PR #2652, opened 2018-10-28
Convert from CRLF to LF for GDAXDownloader

dd-w:
Produced from git autocrlf


mchandschuh:
Any particular reason this change is needed? It looks like most (maybe all?) of the files in this repo (except any `.sh` scripts) use windows line ending conventions of `CRLF` -- if we do want to use unix style `LF` line ending convention than we should convert the entire repository in one fell swoop.

@jaredbroad -- what's your preference here? `CRLF` or `LF` line endings? Perhaps we should also update the `CONTRIBUTING.md` to specify how to configure `git` to ensure line endings remain consistent.

dd-w:
This is the only file that kept causing showing as changed with my Git client (GitKraken). I assumed it was autocrlf that was changing the file. 

Does anyone else have a similar issue?

jaredbroad:
Thanks for PR @dd-w -- suggest we keep it as the bulk of repo for now like @mchandschuh mentioned.

I'm guessing your IDE/git is converting it as you work on the files

dd-w:
@jaredbroad @mchandschuh I looked into it further and from what I can tell, in this repo on Github the only file that has `CRLF` line ending is `GDAXDownloader.cs`. I could only verify this by downloading the raw files from Github - not by checking out the repo as Git will automatically converts to `CRLF` for me.

mchandschuh:
Great point, I foolishly checked the files on my local machine.

PR #2654, opened 2018-10-28
Add Trailing Stop Loss Risk Management Model

dd-w:
#### Description

Provides an implementation of a trailing stop risk management model

#### Motivation and Context

It's useful for users

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Risk/TrailingStopRiskManagementModel.cs
+++ b/Algorithm.Framework/Risk/TrailingStopRiskManagementModel.cs
@@ -1,0 +1,93 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


Algorithm.Framework/Risk/TrailingStopRiskManagementModel.cs
```diff
@@ -0,0 +1,89 @@
...
+                // Check for new highs and update - set to tradebar high
+                if (_trailingHighs[symbol] < security.High)
+                {
+                    _trailingHighs[symbol] = security.High;
+                    continue;
+                }
+
+                // Check for securities past the drawdown limit
+                var securityHigh = _trailingHighs[symbol];
+                var drawdown = (security.Low / securityHigh) - 1m;
```
mchandschuh: Is comparing against `security.Low` correct here? Consider we're using daily data and there's a really bad intraday low, but we close the day higher? I'm thinking it makes more sense to compare against the current `security.Price`  -- thoughts?

dd-w: Yes, I thought about this and because it's a risk management model I though it best to cover worst case scenario. I agree that the past data is also not really relevant, so in that case `security.Price` makes sense. Should I make this change?

jaredbroad: Please keep it as low; its _triggering_ the trailing stop; so is still relevant. The delay in the application is a fault of using long-term data like that. We shouldn't accept higher drawdowns because of choice of bars?

An extreme case would be weekly bars; during the flash crash week. I would expect a trailing stop to trigger in this case?

PR #2657, opened 2018-10-29
Adds ContainsKey check in MacdAlphaModel

### LINKED ISSUE #2656: Addition of existing key to dictionary in MacdAlphaModel
#### Expected Behavior
Framework models should not throw exceptions.

#### Actual Behavior
In `MacdAlphaModel.OnSecuritiesChanged`, when a security is added more than once to the model, it tries to add it to a dictionary that has an existing key.

#### Potential Solution
Add `ContainsKey` check.

#### Reproducing the Problem
Executes `CustomFrameworkModelsAlgorithm` in the cloud with a different `EndDate`. 

#### System Information
No.

#### Checklist

AlexCatarino:
#### Description
In `MacdAlphaModel.OnSecuritiesChanged`, a missing `ContainsKey` is not preventing a second key addition to a dictionary.

#### Related Issue
Closes #2656

#### Motivation and Context
Framework models should not throw.

No.

Executes `CustomFrameworkModelsAlgorithm` in the cloud with a different `EndDate`.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Alphas/MacdAlphaModel.cs
+++ b/Algorithm.Framework/Alphas/MacdAlphaModel.cs
@@ -111,6 +111,10 @@ namespace QuantConnect.Algorithm.Framework.Alphas
         {
             foreach (var added in changes.AddedSecurities)
             {
+                if (_symbolData.ContainsKey(added.Symbol))
+                {
+                    continue;
+                }
                 _symbolData.Add(added.Symbol, new SymbolData(algorithm, added, _fastPeriod, _slowPeriod, _signalPeriod, _movingAverageType, _resolution));
             }
 
@@ -121,6 +125,7 @@ namespace QuantConnect.Algorithm.Framework.Alphas
                 {
                     // clean up our consolidator
                     algorithm.SubscriptionManager.RemoveConsolidator(data.Security.Symbol, data.Consolidator);
+                    _symbolData.Remove(removed.Symbol);
                 }
             }
         }


Algorithm.Framework/Alphas/MacdAlphaModel.cs
```diff
@@ -111,6 +111,10 @@ public override void OnSecuritiesChanged(QCAlgorithmFramework algorithm, Securit
         {
             foreach (var added in changes.AddedSecurities)
             {
+                if (_symbolData.ContainsKey(added.Symbol))
```
jaredbroad: Can you give some details about why this appears to be adding a second time?

AlexCatarino: When the security is removed from the universe, we don't remove the `SymbolData`:
```csharp
foreach (var removed in changes.RemovedSecurities)
{
    SymbolData data;
    if (_symbolData.TryGetValue(removed.Symbol, out data))
    {
        // clean up our consolidator
        algorithm.SubscriptionManager.RemoveConsolidator(data.Security.Symbol, data.Consolidator);
    }
}
```
Now I realized that, if the dictionary contains the key, we need to add the consolidator again, so we should use `TryRemove` below.

mchandschuh:
fyi, it may still be possible to receive multiple added securities events -- this definitely used to hold true for multiple universes selecting the same security (the original intent of the event was notifying that a universe added the security, not that the algorithm added the security). We would certainly want to confirm this behavior, so it may be nice to leave the containskey check in there for robustness?

Issue #2616, opened 2018-10-17
Track Universes requesting each Subscription


Martin-Molinero:
#### Expected Behavior

-  Track `Universes` requesting each `Subscription`. This is a step forward in having multiple `Universe` requesting the same `SubscritpionDataConfig`. In the case both `Universe` A and B request the same `Subscription` and during execution `Universe` B removes the request, it should still exist due to `Universe` A
#### Actual Behavior

- Currently Lean is no tracking which universe requested each subscription.
#### Potential Solution

- Refactor
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2662, opened 2018-10-31
 Subscriptions will track Universe requests

Martin-Molinero:
#### Description

- The class `Subscription` will internally track each `Universe`
`SubscriptionRequest` added or removed. **This PR is not updating the 'UtcStartTime' and 'UtcEndTime' of the `Subscription`**, they will maintain there initial value.
- Adding regression test in which two different `Universe` request the
same `SubscriptionDataConfig` and one of them removes/adds it in a
toggle fashion **(fails on current master)**
- `UniverseSelection` pending removals will also be tracked by
`Universe`
   - Extracting pending removals logic from `UniverseSelection` class into
a new helper class `PendingRemovalsManager`. This new class will keep
track of the `universes` requesting to remove a security. Adding unit tests
- `UniverseDecorator` will overwrite the `Universe` member of
`SubscriptionsRequests` at `GetSubscriptionRequests()`. This is due to
`this != this,Universe`. Behavior covered by existing regression tests.
- Adding `Subscription` unit tests covering expected behavior

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2616
#### Motivation and Context

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/Subscription.cs
+++ b/Engine/DataFeeds/Subscription.cs
@@ -38,14 +38,9 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// <summary>
         /// Gets the universe for this subscription
         /// </summary>
-        public IEnumerable<Universe> Universe => _subscriptionRequests
+        public IEnumerable<Universe> Universes => _subscriptionRequests
             .Where(x => x.Universe != null)
             .Select(x => x.Universe);
-
-        /// <summary>
-        /// Gets true if there is at least one subscription requests in this subscription
-        /// </summary>
-        public bool HasSubscriptionRequests => _subscriptionRequests.Any();
 
         /// <summary>
         /// Gets the security this subscription points to
@@ -95,7 +90,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// <summary>
         /// Gets whether or not this subscription has been removed from its parent universe
         /// </summary>
-        public IReadOnlyRef<bool> RemovedFromUniverse { get; private set; }
+        public IReadOnlyRef<bool> RemovedFromUniverse { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Subscription"/> class with a universe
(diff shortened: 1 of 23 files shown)


Engine/DataFeeds/Subscription.cs
```diff
@@ -32,11 +33,19 @@ public class Subscription : IEnumerator<SubscriptionData>
...
         private bool _removedFromUniverse;
         private readonly IEnumerator<SubscriptionData> _enumerator;
+        private List<SubscriptionRequest> _subscriptionRequests;
 
         /// <summary>
         /// Gets the universe for this subscription
         /// </summary>
-        public Universe Universe { get; }
+        public IEnumerable<Universe> Universe => _subscriptionRequests
+            .Where(x => x.Universe != null)
```
mchandschuh: as a note -- the only case of a null universe should be subscriptions added for the sole purpose of securities added via `EnsureCurrencyDataFeeds`

Engine/DataFeeds/Subscription.cs
```diff
@@ -91,37 +100,76 @@ public class Subscription : IEnumerator<SubscriptionData>
...
+            if (universe == null)
+            {
+                var result = _subscriptionRequests;
+                _subscriptionRequests = new List<SubscriptionRequest>();
+                return result.Where(x => x.Universe != null)
+                    .Select(x => x.Universe);
+            }
+
+            _subscriptionRequests.RemoveAll(x => x.Universe == universe);
+            // TODO this might update the 'UtcStartTime' and 'UtcEndTime' of this subscription
```
mchandschuh: Luckily start is always now and end is always end of algo, although support for dynamic end values may be useful in the future, so definitely good to be considering this.

Engine/DataFeeds/SubscriptionSynchronizer.cs
```diff
@@ -147,7 +149,7 @@ public TimeSlice Sync(IEnumerable<Subscription> subscriptions)
                                 : packetBaseDataCollection.Data;
 
                             BaseDataCollection collection;
-                            if (universeData.TryGetValue(subscription.Universe, out collection))
+                            if (universeData.TryGetValue(subscription.Universe.First(), out collection))
```
mchandschuh: should we be performing this for each universe? if we want to affirm only one universe in the list, let's use `Single()`.
Also, just now realizing the naming, perhaps in a stand-alone PR, can rename `Subscription.Universe` to `Subscription.Universes` to make it clear that it's a list.

PR #2673, opened 2018-11-06
Prevents negative Variance

### LINKED ISSUE #2672: No check for negative value for Variance indicator
#### Expected Behavior
No negative value for `Variance`.

#### Actual Behavior
`Variance` indicator yields negative values.

#### Potential Solution
Add `Math.Max` to ensure Variance indicator doesn't return negative values.

#### Checklist

AlexCatarino:
#### Description
Prevents negative variance (mathematically impossible) in `Variance` indicator.
Since variance is used to calculate standard deviation, we found that negative variance yield `double.NaN` and, consequently, an arithmetic overflow when we convert `NaN` into `decimal`.

#### Related Issue
Closes #2672

#### Motivation and Context
Fix bugs in indicators 

No.

Existing unit and regression tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/Variance.cs
+++ b/Indicators/Variance.cs
@@ -78,7 +78,8 @@ namespace QuantConnect.Indicators
                 _rollingSumOfSquares -= removedValue * removedValue;
             }
 
-            return meanValue2 - meanValue1 * meanValue1;
+            // Ensure non-negative variance
+            return System.Math.Max(0m, meanValue2 - meanValue1 * meanValue1);
         }
 
         /// <summary>


Indicators/Variance.cs
```diff
@@ -78,7 +78,8 @@ protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint>
                 _rollingSumOfSquares -= removedValue * removedValue;
             }
 
-            return meanValue2 - meanValue1 * meanValue1;
+            // Ensure non-negative variance
+            return System.Math.Max(0m, meanValue2 - meanValue1 * meanValue1);
```
Martin-Molinero: From my understanding, Variance should never be negative since it squares the difference. So, I'm wondering if the result being negative is a result of this progressive algorithm carrying a precision error? or how is it we ended up with a negative number for the Variance? I think it would be nice to confirm/note this

AlexCatarino: Right, it is a matter of precision error. It led `meanValue1 * meanValue1` being greater than `meanValue2` by a very small negative number: `-2e-25`.

Issue #2634, opened 2018-10-23
FillModel relies on configurations properties of Security


Martin-Molinero:
#### Expected Behavior

- FillModel does not rely on configurations properties of Security
#### Actual Behavior

- FillModel uses `Security.Subscriptions` and `Security.IsExtendedMarketHours`.
   - Usage of `Security.Subscriptions` was incorporated as a fix for another issue https://github.com/QuantConnect/Lean/pull/1078. Underlying issue is still there, data in security cache might not always be trusted.
#### Potential Solution

- Refactor
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2683, opened 2018-11-08
Refactor Fill Models - v2

Martin-Molinero:
#### Description

The objective of this PR is to remove usages of `Security.Subscriptions` and
`Security.IsExtendedMarketHours` properties by `FillModels` maintaining retro-compatibility.

System will always call new `Fill(FillModelContext context)` which will detect if the user is overriting old `xxxxFill(Security asset, MarketOrder order)` and call it insted. 
> If users fill model implementation inherited `FillModel` and calls `base.xxxxFill(Security asset, MarketOrder order)`, `FillModel` will now send the call to new `xxxxFillImplementation(FillModelContext context)`, not `Fill(FillModelContext context)` (else we would loop).

- Making `IFillModel` a class
- Adding new `FillModelContext` that will hold `Security`, `Order` and
`ISubscriptionDataConfigProvider`
- Adding new `InterfaceFillModel` who will replace old `IFillModel`.
New `Fill()` method will have new `FillModelContext` as only parameter
- Adding unit tests showcasing retro compatibility.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2634
#### Motivation and Context

yes, link https://www.quantconnect.com/docs/algorithm-reference/reality-modelling#Reality-Modelling-Fill-Models

> The Fill Models implement the `IFillModelInterface` interface. => `OrderEvent Fill(FillModelContext context)`

> Users are suggested to override base class `FillModel`, which allows users to overwrite a specific order type fill method, as `OrderEvent MarketFill(FillModelContext context)`. Versus having to handle all order types when directly implemeting `IFillModelInterface` interface, unless this is desired. => `OrderEvent Fill(FillModelContext context)`

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/Backtesting/BacktestingBrokerage.cs
+++ b/Brokerages/Backtesting/BacktestingBrokerage.cs
@@ -20,6 +20,7 @@ using System.Linq;
 using QuantConnect.Interfaces;
 using QuantConnect.Logging;
 using QuantConnect.Orders;
+using QuantConnect.Orders.Fills;
 using QuantConnect.Securities;
 using QuantConnect.Securities.Option;
 
@@ -318,34 +319,22 @@ namespace QuantConnect.Brokerages.Backtesting
                     {
                         //Model:
                         var model = security.FillModel;
-
                         //Based on the order type: refresh its model to get fill price and quantity
                         try
                         {
                             switch (order.Type)
                             {
                                 case OrderType.Limit:
-                                    fills = new[] { model.LimitFill(security, order as LimitOrder) };
-                                    break;
-
                                 case OrderType.StopMarket:
-                                    fills = new[] { model.StopMarketFill(security, order as StopMarketOrder) };
-                                    break;
-
                                 case OrderType.Market:
-                                    fills = new[] { model.MarketFill(security, order as MarketOrder) };
-                                    break;
-
                                 case OrderType.StopLimit:
-                                    fills = new[] { model.StopLimitFill(security, order as StopLimitOrder) };
-                                    break;
-
                                 case OrderType.MarketOnOpen:
-                                    fills = new[] { model.MarketOnOpenFill(security, order as MarketOnOpenOrder) };
-                                    break;
-
                                 case OrderType.MarketOnClose:
-                                    fills = new[] { model.MarketOnCloseFill(security, order as MarketOnCloseOrder) };
+                                    var context = new FillModelContext(
+                                        security,
+                                        order,
+                                        Algorithm.SubscriptionManager.SubscriptionDataConfigService);
+                                    fills = new[] { model.Fill(context) };
                                     break;
 
                                 case OrderType.OptionExercise:
--- a/Brokerages/Backtesting/BacktestingBrokerage.cs
+++ b/Brokerages/Backtesting/BacktestingBrokerage.cs
@@ -322,25 +322,18 @@ namespace QuantConnect.Brokerages.Backtesting
                         //Based on the order type: refresh its model to get fill price and quantity
                         try
                         {
-                            switch (order.Type)
+                            if (order.Type == OrderType.OptionExercise)
                             {
-                                case OrderType.Limit:
-                                case OrderType.StopMarket:
-                                case OrderType.Market:
-                                case OrderType.StopLimit:
-                                case OrderType.MarketOnOpen:
-                                case OrderType.MarketOnClose:
-                                    var context = new FillModelContext(
-                                        security,
-                                        order,
-                                        Algorithm.SubscriptionManager.SubscriptionDataConfigService);
-                                    fills = new[] { model.Fill(context) };
-                                    break;
-
-                                case OrderType.OptionExercise:
-                                    var option = (Option)security;
-                                    fills = option.OptionExerciseModel.OptionExercise(option, order as OptionExerciseOrder).ToArray();
-                                    break;
+                                var option = (Option)security;
+                                fills = option.OptionExerciseModel.OptionExercise(option, order as OptionExerciseOrder).ToArray();
+                            }
+                            else
+                            {
+                                var context = new FillModelContext(
+                                    security,
+                                    order,
+                                    Algorithm.SubscriptionManager.SubscriptionDataConfigService);
+                                fills = new[] { model.Fill(context) };
                             }
 
                             // invoke fee models for completely filled order events
--- a/Common/Orders/Fills/FillModel.cs
+++ b/Common/Orders/Fills/FillModel.cs
@@ -14,8 +14,6 @@
 */
 
 using System;
-using System.Linq;
-using QuantConnect.Data.Market;
 using QuantConnect.Securities;
 
 namespace QuantConnect.Orders.Fills
@@ -23,6 +21,7 @@ namespace QuantConnect.Orders.Fills
     /// <summary>
     /// Provides a base class for all fill models
     /// </summary>
+    /// <remarks>Users FillModels should inherit IFillModel</remarks>
     public class FillModel : IFillModel
     {
         /// <summary>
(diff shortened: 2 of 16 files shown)


Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -318,34 +319,22 @@ public virtual void Scan()
...
-                                    fills = new[] { model.StopLimitFill(security, order as StopLimitOrder) };
-                                    break;
-
                                 case OrderType.MarketOnOpen:
-                                    fills = new[] { model.MarketOnOpenFill(security, order as MarketOnOpenOrder) };
-                                    break;
-
                                 case OrderType.MarketOnClose:
-                                    fills = new[] { model.MarketOnCloseFill(security, order as MarketOnCloseOrder) };
+                                    var context = new FillModelContext(
```
mchandschuh: in order to better support polymorphism, we should make the fill model the default and do:
``` csharp
if (order.Type == OrderType.OptionExercise) // do opt exercise
else { // invoke fill model for everything else }
```
As it is now, if we (or a user) added a new order type it, changes would be needed here in order to support it, just so it goes to the fill model. This should happen automatically.

Common/Orders/Fills/FillModel.cs
```diff
@@ -31,41 +30,23 @@ public class FillModel : IFillModel
         /// <param name="asset">Security asset we're filling</param>
         /// <param name="order">Order packet to model</param>
         /// <returns>Order fill information detailing the average price and quantity filled.</returns>
-        /// <seealso cref="SecurityTransactionModel.StopMarketFill"/>
-        /// <seealso cref="SecurityTransactionModel.LimitFill"/>
+        [Obsolete("This was left for retro compatibility, see new MarketFill(FillModelContext context)")]
```
mchandschuh: I think we can leave these methods non-obsolete. We're only removing them from the `IFillModel` definition. No need to remove them from `FillModel`. It basically _moves_ the switch statement from `BacktestingBrokerage` into `FillModel.Fill`

Common/Orders/Fills/IFillModel.cs
```diff
@@ -13,61 +13,799 @@
...
 using QuantConnect.Securities;
 
 namespace QuantConnect.Orders.Fills
 {
     /// <summary>
-    /// Represents a model that simulates order fill events
+    /// Represents a base model that simulates order fill events
     /// </summary>
-    public interface IFillModel
+    public abstract class IFillModel : InterfaceFillModel
```
mchandschuh: cool, little switch-a-roo -- let's leave good documentation so future devs know _why_ it'slike this, because it does look strange, but there's good reasoning.

Common/Orders/Fills/IFillModel.cs
```diff
@@ -13,61 +13,799 @@
...
-        /// <param name="order">Order to update</param>
+        /// <param name="context">A context object containing the security and order</param>
         /// <returns>Order fill information detailing the average price and quantity filled.</returns>
-        OrderEvent MarketFill(Security asset, MarketOrder order);
+        protected virtual OrderEvent MarketFill(FillModelContext context)
+        {
+            if (!_checkedMarketFill)
+            {
+                _checkedMarketFill = true;
+                SetFunctionCall(context.Order.Type);
```
mchandschuh: this might be less confusing in this method if it's set up as `_marketFill = CreateFunctionCall(context.Order.Type)` vs having the method set the member and then using it a few lines later. This is a 'hidden assignment' from this context. The function, can be a _pure_ function and it will improve understandability here.

Common/Orders/Fills/InterfaceFillModel.cs
```diff
@@ -1,33 +1,32 @@
...
+namespace QuantConnect.Orders.Fills
 {
     /// <summary>
-    /// Security transaction model interface for QuantConnect security objects
+    /// Represents an interface that simulates order fill events
     /// </summary>
-    /// <seealso cref="EquityTransactionModel"/>
-    /// <seealso cref="ForexTransactionModel"/>
-    public interface ISecurityTransactionModel : IFillModel, IFeeModel, ISlippageModel
+    /// <remarks>Users should override base class <see cref="IFillModel"/></remarks>
```
mchandschuh: Perhaps we should make the 'guidance' to override `FillModel` and not `IFillModel`, but we should _support_ `IFillModel` for backwards compatibility. Otherwise, `IFillModel` is a sore thumb. Consider the other classes that have a dedicated base class, it's always `I<Model>` for the interface and then implement from the base class `<Model>`

Martin-Molinero: Good point, for consistency it makes sense for us to suggest they subclass `FillModel` (it actually works correctly either way for us). I'll update this 👍

Common/Orders/Fills/InterfaceFillModel.cs
```diff
@@ -1,33 +1,32 @@
...
 {
     /// <summary>
-    /// Security transaction model interface for QuantConnect security objects
+    /// Represents an interface that simulates order fill events
     /// </summary>
-    /// <seealso cref="EquityTransactionModel"/>
-    /// <seealso cref="ForexTransactionModel"/>
-    public interface ISecurityTransactionModel : IFillModel, IFeeModel, ISlippageModel
+    /// <remarks>Users should override base class <see cref="FillModel"/></remarks>
+    public interface InterfaceFillModel
```
StefanoRaggi: Just a minor comment: maybe this interface could be named `IFillModelInterface` or `IBaseFillModel` for naming convention consistency?
https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/8bc1fexb(v=vs.71)

Martin-Molinero: Whoops! Good idea about the naming convention, I'll rename to `IFillModelInterface` 😅

Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -318,40 +319,21 @@ public virtual void Scan()
...
-
-                                case OrderType.OptionExercise:
-                                    var option = (Option)security;
-                                    fills = option.OptionExerciseModel.OptionExercise(option, order as OptionExerciseOrder).ToArray();
-                                    break;
+                                var context = new FillModelContext(
+                                    security,
+                                    order,
+                                    Algorithm.SubscriptionManager.SubscriptionDataConfigService);
+                                fills = new[] { model.Fill(context) };
```
mchandschuh: Sexy! Love it! User defined custom orders here we come!

Hmm... I almost wonder if we should take this opportunity to make this return a context object -- I'm thinking about support multiple fills at the same time step which will become important when we start managing L2 depth data. With L2 depth data we can get _really_ accurate fill models by eating through the book, and each price level you 'eat' would be a separate fill, so it's possible to have multiple fills at the same time step. Figured it may be worthwhile to account for this now while we're doing the semi-breaking changes.

Common/Orders/Fills/FillModelContext.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Gets the <see cref="Order"/>
+        /// </summary>
+        public Order Order { get; }
+
+        /// <summary>
+        /// Gets the <see cref="SubscriptionDataConfig"/> provider
+        /// </summary>
+        public ISubscriptionDataConfigProvider ConfigProvider { get; }
```
mchandschuh: Awesome! Dependencies made easy :)

Common/Orders/Fills/IFillModel.cs
```diff
@@ -13,61 +13,741 @@
...
 namespace QuantConnect.Orders.Fills
 {
     /// <summary>
-    /// Represents a model that simulates order fill events
+    /// Represents a base model that simulates order fill events.
     /// </summary>
-    public interface IFillModel
+    /// <remarks>Originally <see cref="IFillModel"/> was an interface but due to
+    /// maintain compatibility is was change to be an abstract class and
+    /// <see cref="IFillModelInterface"/> was added. See GH issue 2634.
```
mchandschuh: ty for providing reasoning -- w/out this future us would be like... whyyy!

Common/Orders/Fills/IFillModelInterface.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    /// <remarks>Users should override base class <see cref="FillModel"/></remarks>
+    public interface IFillModelInterface
+    {
+        /// <summary>
+        /// Return an order event with the fill details
+        /// </summary>
+        /// <param name="context">A context object containing the security and order</param>
+        /// <returns>Order fill information detailing the average price and quantity filled.</returns>
+        OrderEvent Fill(FillModelContext context);
```
mchandschuh: Might make sense to give `IFillModelInterface.Fill` a dedicated response object as was done w/ various other methods converted to the context pattern. The idea being that the dedicated return type allows it to evolve. Also, the dedicated return type is little more than a wrapper object to maintain compile-time compatibility. This could make it easier to support multiple fills at a single time step or other (unknown) additions to the return in the future.

Martin-Molinero:
Closing this PR due to issues with the Python side. Will open a new PR.

Issue #2635, opened 2018-10-23
Remove unused SubscriptionDataConfigs


Martin-Molinero:
#### Expected Behavior

- When the configuration is no longer used the `SubscriptionDataConfig` should be removed from the `DataManager`/`SubscriptionManager`
#### Actual Behavior

- SubscriptionDataConfigs are being added to the SubscriptionManager/DataManager but they are NOT being removed, even when a security is removed or when the configuration is removed from the DataFeed. Also, depending properties (as HasCustomData)/Consumers (as TimeKeeper) should be updated.
#### Potential Solution

- Refactor
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2684, opened 2018-11-08
Remove unused SubscriptionDataConfig

Martin-Molinero:
#### Description

- After removing a `Subscription` entirely, we will now also remove the
matching `SubscriptionDataConfig` from the `SubscriptionManager`
- Adding a new regression test, which fails in `master`
- Fixing previous PR change that reduced log level from `Trace` to `Debug` when adding
and removing a `Subscription`, for live mode
- Fixing a bug in the `UserDefinedUniverse` where calling `RemoveMember`
would cause the `SubscriptionDataConfig` to be re added to the
`SubscriptionManager` (not the `DF`). Found this through added
regression test. Fixed by calling new `private RemoveAndKeepTrack()` from both
 `RemoveMember()` and `Remove()` (has no impact today in `master`)
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2635
#### Motivation and Context

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/UniverseSelection/UserDefinedUniverse.cs
+++ b/Common/Data/UniverseSelection/UserDefinedUniverse.cs
@@ -228,13 +228,8 @@ namespace QuantConnect.Data.UniverseSelection
         /// <returns>True if the symbol was removed, false if the symbol was not present</returns>
         public bool Remove(Symbol symbol)
         {
-            var toBeRemoved = _subscriptionDataConfigs.Where(x => x.Symbol == symbol).ToList();
-            var removedSymbol = _symbols.Remove(symbol);
-
-            if (toBeRemoved.Any() || removedSymbol)
+            if (RemoveAndKeepTrack(symbol))
             {
-                _subscriptionDataConfigs.RemoveWhere(x => x.Symbol == symbol);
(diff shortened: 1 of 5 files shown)


Common/Data/UniverseSelection/UserDefinedUniverse.cs
```diff
@@ -322,5 +317,37 @@ protected virtual void OnCollectionChanged(NotifyCollectionChangedEventArgs e)
...
+            }
+            return false;
+        }
+
+        private bool RemoveAndKeepTrack(Symbol symbol)
+        {
+            var toBeRemoved = _subscriptionDataConfigs.Where(x => x.Symbol == symbol).ToList();
+            var removedSymbol = _symbols.Remove(symbol);
+
+            if (toBeRemoved.Any() || removedSymbol)
```
mchandschuh: could save a couple cycles here by doing:
``` csharp
if (removedSymbol || toBeRemoved.Any())
{
    //
}
```
This will cause us to only evaluate the LINQ expression _if_ `removedSymbols == false`

Martin-Molinero: Good catch 👍

Issue #1924, opened 2018-04-27
Fees being applied in USD with Cypto pairs in BaseAccount currency in EUR


espirulina:
#### Expected Behavior

When using AddCrypto("ETHEUR") on a EUR Base account currency, we expect the fees being applied in EUR 

#### Actual Behavior

The fees are being applied in USD

#### Reproducing the Problem

```
using System;
using System.Linq;
using QuantConnect.Brokerages;
using QuantConnect.Data;
using QuantConnect.Indicators;
using QuantConnect.Orders;

namespace QuantConnect.Algorithm.CSharp
{
    public class TestFeesInEuroBaseAccountCurency : QCAlgorithm
    {
        private Symbol symbol;

        public override void Initialize()
        {
            SetStartDate(2018, 1, 1); // Set Start Date
            SetEndDate(2018, 1, 2); // Set End Date

            SetCash("USD", 0, 1.23m);
            SetCash("EUR", 1000, 1.23m);
            SetCash("BTC", 0m, 7300m);
            SetCash("ETH", 0m, 400m);

            symbol = AddCrypto("ETHEUR", Resolution.Daily, Market.GDAX).Symbol;

            AddSecurity(SecurityType.Crypto, symbol, Resolution.Daily);
            Securities[symbol].FeeModel = new GDAXFeeModel();

            SetBrokerageModel(BrokerageName.GDAX, AccountType.Cash);

        }


        public override void OnData(Slice data)
        {
        	if (!Portfolio.Securities["ETHEUR"].Invested)
            	SetHoldings(symbol,1);
        }

        public override void OnOrderEvent(OrderEvent orderEvent)
        {
            Debug("OrderFee: " + orderEvent.OrderFee + " FillPriceCurrency " + orderEvent.FillPriceCurrency);
            Debug("EUR: " + Portfolio.CashBook["EUR"].Amount + " USD: " + Portfolio.CashBook["USD"].Amount + " ETH: " +Portfolio.CashBook["ETH"].Amount);
            
            
        }

    }
}
```

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

jaredbroad:
Thanks for the report @espirulina.

Issue #2741, opened 2018-12-07
Update GDAX fee currency


Martin-Molinero:
#### Expected Behavior

- GDAX fees are in quote currency
#### Actual Behavior

- GDAX fees are in account currency (USD)
#### Potential Solution

- Refactor GDAX fee model and brokerage implementation to use quote currency as fee currency.
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

Martin-Molinero:
Closed by https://github.com/QuantConnect/Lean/pull/2742

PR #2740, opened 2018-12-07
Add OrderFee to OrderEvent

Martin-Molinero:
#### Description

> This PR is on top of PR https://github.com/QuantConnect/Lean/pull/2725
- Replacing `decimal` for `OrderFee` at the `OrderEvent` class.
    - Adding implicit conversion from `OrderFee` to `decimal` for backwards compatibility.
    - Also adding `ToString()` implementation for `OrderFee` for backwards compatibility.
- Adding `FeeModelNotUsingAccountCurrency` regression test
- Adding unit tests for `CashBuyingPowerModel`, `SecurityPortfolioModel` and `SecurityMarginModel`
with non account currency fees
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2722
#### Motivation and Context

https://github.com/QuantConnect/Lean/projects/6
#### Requires Documentation Change

N/A
#### How Has This Been Tested?

Unit and regression tests
#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`

PR #2742, opened 2018-12-07
Update GDAX order fee currency

Martin-Molinero:
#### Description

> This PR is on top of PR https://github.com/QuantConnect/Lean/pull/2740
- GDAX `FeeModel` and live implementation will now use quote currency as fee currency. Adding unit
test.

> BETA of `BasicTemplateCryptoAlgorithm` changed because the EUR USD conversion rate at the end of the execution was slightly different (smaller) than the one when the trade happened.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/1924 and https://github.com/QuantConnect/Lean/issues/2741
#### Motivation and Context

https://github.com/QuantConnect/Lean/projects/6

N/A

Unit and regression tests
> Live implementation is tested by `GDAXBrokerageTests.OnMessageFillTest` unit test, added assert over currency value.
#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.Messaging.cs
@@ -28,6 +28,7 @@ using RestSharp;
 using System.Text.RegularExpressions;
 using QuantConnect.Logging;
 using QuantConnect.Orders.Fees;
+using QuantConnect.Securities;
 using QuantConnect.Util;
 
 namespace QuantConnect.Brokerages.GDAX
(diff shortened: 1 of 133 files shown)


Brokerages/GDAX/GDAXBrokerage.Messaging.cs
```diff
@@ -414,7 +420,9 @@ private void EmitFillOrderEvent(Messages.Matched message, Symbol symbol, GDAXFil
             var fillPrice = message.Price;
             var fillQuantity = direction == OrderDirection.Sell ? -message.Size : message.Size;
             var isMaker = order.BrokerId[0] == message.MakerOrderId;
-            var orderFee = GetFillFee(symbol, fillPrice, fillQuantity, isMaker);
+            var orderFee = new OrderFee(new CashAmount(
+                GetFillFee(symbol, fillPrice, fillQuantity, isMaker),
+                symbol.Value.Substring(3)));
```
StefanoRaggi: This line assumes the _base_ currency has a fixed length of 3 characters. 
Since Coinbase is incrementally adding new symbols with 4-char base currencies and 4-char quote currencies (e.g. `MANA/USDC`, `LOOM/USDC`), I think it might be worth to move this logic to private methods such as `GetBaseCurrency(string ticker)` and `GetQuoteCurrency(string ticker)`.

For now we might just use a hardcoded list of available quote currencies (USD, EUR, GBP, BTC, USDC), in the future we could fetch this information directly from the brokerage at startup time.

Martin-Molinero: Good point! 
After digging a bit more on this I decided to use `order.PriceCurrency` or, if empty, `_algorithm.Securities[symbol].SymbolProperties.QuoteCurrency` (as done in other brokers implementation).
Regarding the fixed 3 char lenght, it seems to be scattered among othe places too e.g. https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityService.cs#L66 https://github.com/QuantConnect/Lean/blob/master/Common/Securities/Cash.cs#L90 which would require more changes, so I'm thinking of openning a new issue to address it separatly.

Martin-Molinero: New issue at https://github.com/QuantConnect/Lean/issues/2744

Issue #2753, opened 2018-12-13
Add Non-USD Account Currency Support to Brokerages


Martin-Molinero:
#### Expected Behavior

- Brokerages implementations support non-usd account currencies
#### Actual Behavior

- Review the different `Brokerages` implementations. For example `GetCashBalances()` conversion rates are being calculated as if account currency was in USD (FXCM, GDAX)
#### Potential Solution

- Maybe send in account currency as a parameter for the `GetCashBalances()` method if this was the only method requiring it.
- What if `Cash` class doesn't know about conversion rates at all?
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #2786, opened 2018-12-27
Non-Usd brokerage support

Martin-Molinero:
#### Description

- `GetCashBalance()` will return a `List<CashAmount>`, will not need to
set conversion rates, which requires knowing what the account currency is.
    - Removing now unused brokerage `GetConversionRate` implementation (used internally, not present in the interface/base brokerage)
- `Global.Holding` conversion rate field, will be set to 1 for `List<Holding> GetAccountHoldings()`, since it isn't being used and requires knowing what the account currency is.
- Adding equality operators for `CashAmount`. Adding unit tests.
- Adding trace log for `BaseSetupHandler.SetupCurrencyConversions()`
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2753
#### Motivation and Context

https://github.com/QuantConnect/Lean/projects/6

N/A

Unit and regression tests, live deployment
#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -2051,6 +2051,7 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
                 Quantity = e.Position,
                 AveragePrice = Convert.ToDecimal(e.AverageCost) / multiplier,
                 MarketPrice = Convert.ToDecimal(e.MarketPrice),
+                ConversionRate = 1m, // this will be overwritten when GetAccountHoldings is called to ensure fresh values
                 CurrencySymbol = currencySymbol
             };
         }
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -2051,7 +2051,6 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
                 Quantity = e.Position,
                 AveragePrice = Convert.ToDecimal(e.AverageCost) / multiplier,
                 MarketPrice = Convert.ToDecimal(e.MarketPrice),
-                ConversionRate = 1m, // this will be overwritten when GetAccountHoldings is called to ensure fresh values
                 CurrencySymbol = currencySymbol
             };
         }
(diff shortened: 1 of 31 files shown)


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1035,179 +1011,6 @@ private IEnumerable<ContractDetails> FindContracts(Contract contract)
...
-
-                    // be sure to unwire our history handler as well
-                    _client.HistoricalData -= clientOnHistoricalData;
-                    _client.HistoricalDataEnd -= clientOnHistoricalDataEnd;
-                    _client.Error -= clientOnError;
-
-                } while (pacingViolation);
-            }
-
-            return inverted ? 1 / rate : rate;
```
mchandschuh: I bet deleting this mess felt great!

Common/Securities/CashAmount.cs
```diff
@@ -47,5 +47,42 @@ public CashAmount(decimal amount, string currency)
...
+            return !Equals(lhs, rhs);
+        }
+
+        /// <summary>
+        /// Used to compare two <see cref="CashAmount"/> instances.
+        /// Useful to compare against the default instance
+        /// </summary>
+        /// <param name="obj">The other object to compare with</param>
+        /// <returns>True if <see cref="Currency"/> and <see cref="Amount"/> are equal</returns>
+        public override bool Equals(object obj)
```
mchandschuh: Recommend also implementing `GetHashCode` - - in general, when overriding `Equals` on an immutable type you also want to override `GetHashCode`. This is especially true for struct types.

Also, ReSharper has a helper that generates the equality member functions by placing a cursor on the type, going to the ReSharper menu, then Edit, then Generate Code and selecting equality members from the pop-up. It will create a great hash function for you that guarantees a decent distribution and will also create the idiomatic equals which includes implementing `IEquatable<T>`.

Engine/Setup/BaseSetupHandler.cs
```diff
@@ -92,6 +93,9 @@ public static class BaseSetupHandler
                     cash.Update(data);
                 }
             });
+
+            Log.Trace("BaseSetupHandler.SetupCurrencyConversions(): " +
+                $"{string.Join(" | ", algorithm.Portfolio.CashBook.Values.Select(x => x.ToString()))}");
```
mchandschuh: Aww, you didn't like the nifty `CashBook.ToString()` overload, prints it all out as a table :)

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
Martin-Molinero: Yeah! :D

Common/Securities/CashAmount.cs
Martin-Molinero: Nice, thanks for the tip!

Engine/Setup/BaseSetupHandler.cs
Martin-Molinero: Whoops! Your right, the table looks much better, I'll replace it 👍

Issue #2824, opened 2019-01-13
Create Random Data Generator


jaredbroad:
### Background
To test algorithms locally data doesn't need to be fully real. Just look like it and be able to run a full backtest. This is especially important for benchmarking where you need a substantial, realistic dataset to fully test an optimization.

### Goal

Design a C# QuantConnect.ToolBox class which generates random data in the QC format. If re-using LEAN classes this should actually be a simple project:

 - Use [McNaster CLI](https://github.com/natemcmaster/CommandLineUtils) args reader to get the target data:
 - - Start Date? End Date?
 - - Which type of data? (default -> equity) Equity, Forex, Cfd, Options or Futures?
 - - If Equity -> Include Coarse Universe Data? (default->yes)
 - - Which resolution of data? (default->minute) Tick, Second, Minute, Hour, Daily?
 - - Which market should we target? (default-> equity->usa, forex->oanda)
 - - Sparse or Dense Data? (default->dense)
 - - How many assets to generate?

Generate the tick type of data (Trade/Quote) depending on the consumer request. Equity (trade only), Forex-CFD (quote only), Future-Option (Trade/Quote/OpenInterest). 

Dense data means a tick at least every target resolution step. Sparse meaning gaps between resolution steps (sufficient to trigger use of LEAN fill forward, e.g. target minute data; generate ticks every rand()*4 minutes)

For Options input check the resolution; options can only do minute data for now.Stop the processing and show an error if another resolution set.

#### Symbol Generator
Create a fake symbol generator function feeding it the input parameters above; yield new symbols. 

#### Tick Generator 
1) Now the target is set; use a random walk to generate the series of prices for a given tick type, date range, and symbol object. The prices function should create an enumerable of ticks (type tick or trade).
2) Use the target settings to space out the time if sparse data is requested.
3) Enumerate each data-type (trade-quote) outside the tick generator, passing in the input tick-type.

#### Consolidate
1) Create consolidators for the resolution requested. If coarse-universe flag set; make sure a daily consolidator is in the consolidator collection.
1.1) For Trade ticks use a TradeTickConsolidator; QuoteTicks -> QuoteTickConsolidator.
2) Enumerate the tick generator; pipe into the appropriate consolidator. 
3) Write the data to file using the `LeanDataWriter` class. 

#### Loop All Symbols
Once all the above is done, you've successfully written the ticks to consolidated files, loop over all the fake symbols generated by the Symbol Generator and create N-fake datasets.

[...]

PR #2828, opened 2019-01-15
#2824 Add ToolBox RandomDataGenerator

mchandschuh:
#### Description

Creates a new toolbox project: `RandomDataGenerator` which satisfies the requirements as laid out in #2824. A piece missing from those requirements is properly emitting underlying equity data in tandem with options data. I've defined an `IRandomValueDataGenerator` which produces all of the random values and it supports specifying a seed value  to permit deterministic results between multiple runs. I added an enum, `DataDensity` with values `Dense`, `Sparse` and `VerySparse`. I felt the need for `VerySparse` was large given how thin some options and even some equities can trade, plus having less than a trade tick a day is an important test case for `Lean`. The `RandomValueGenerator` as implemented uses a simplified model of [Browning motion](https://en.wikipedia.org/wiki/Brownian_motion) where we simply advance the current value by +/- random value. Also, too keep the ticks more realistic, they too are advanced at random values capped by the `DataDensity` parameter. All settings are configurable via command line and extensive output is provided for users to help make using the program easier, including solid reporting of parse errors and invalid configuration values. I also threw in some color on the command line to keep your eyes pleased :)

A note on performance: The generation of data is 100% lazy-fed into consolidators where it is _non_ lazily aggregated into lists per consolidator. These consolidated lists are then passed off to the `LeanDataWriter` to produce the correct output files in `Lean` format. While it _isn't_ parallelized, it would be very easy to make it parallel as the processing of each `symbol` does not share any mutable state with the processing of other symbols, so we could simply throw it in a `Parallel.ForEach` and it'll be happy (be sure to use things like `Interlocked.Increment` when updating the `count` for progress) I didn't want to make this PR more complicated than necessary, but if desired, I'm happy to add it -- would only take a few minutes.

A note on options data -- this PR has a _serious_ shortcoming that I raised but did not address here, although it shouldn't take too much time to remedy, and that's when producing `Option` data we should be producing the `Equity` data at the same time, or perhaps, produce the equity data, and then use those prices to guide the options prices... this may require just a tad more though, since if we want it to be realistic we should probably be running black scholes on the equity data to resolve volatility, etc and back-compute realistic options prices instead of _randomized_ prices for options -- that way we don't get wierd/nan values for the greeks.

#### Related Issue

See #2824 

#### Motivation and Context

#2824 aims to democratize the testing of `Lean` by empowering the open source community to generate large test data sets that are realistic enough to perform performance benchmarking of `Lean` while not requiring any data licenses, etc.

Documentation _could_ be added regarding usage of the application, although I'm not sure if it's standard practice to add usage documentation for the `ToolBox` projects -- I'll defer to @jaredbroad on this one.

The `RandomValueGenerator` is _extensively_ tested by units tests with almost 100% test coverage. Since this is the meat of the application, I felt it necessary to provide confidence that the generated data is consistent with the various data types used and that the generated data meets the specifications outlined in #2824 .

#### Types of changes

#### Checklist:

#### Sample Data Output
image

### DIFF
--- a/Configuration/ToolboxArgumentParser.cs
+++ b/Configuration/ToolboxArgumentParser.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using Microsoft.Extensions.CommandLineUtils;
 
 namespace QuantConnect.Configuration
@@ -23,13 +24,16 @@ namespace QuantConnect.Configuration
                                                      + "/YahooDownloader or YDL/AlgoSeekFuturesConverter or ASFC/AlgoSeekOptionsConverter or ASOC"
                                                      + "/IVolatilityEquityConverter or IVEC/KaikoDataConverter or KDC/NseMarketDataConverter or NMDC"
                                                      + "/QuantQuoteConverter or QQC/CoarseUniverseGenerator or CUG\n"
+                                                     + "RandomDataGenerator or RDG"
                                                      + "Example 1: --app=DDL\n"
-                                                     + "Example 2: --app=NseMarketDataConverter\n"),
+                                                     + "Example 2: --app=NseMarketDataConverter\n"
+                                                     + "Example 3: --app=RDG"),
                 new CommandLineOption("tickers", CommandOptionType.MultipleValue, "[REQUIRED ALL downloaders (except QBDL)] "
                                                                                   + "--tickers=SPY,AAPL,etc"),
                 new CommandLineOption("resolution", CommandOptionType.SingleValue, "[REQUIRED ALL downloaders (except QBDL, CDL) and IVolatilityEquityConverter,"
                                                                                    + " QuantQuoteConverter] *Not all downloaders support all resolutions. Send empty for more information.*"
-                                                                                   + " CASE SENSITIVE: --resolution=Tick/Second/Minute/Hour/Daily/All"),
+                                                                                   + " CASE SENSITIVE: --resolution=Tick/Second/Minute/Hour/Daily/All" +Environment.NewLine+
+                                                                                   "[OPTIONAL for RandomDataGenerator - same format as downloaders, Options only support Minute"),
                 new CommandLineOption("from-date", CommandOptionType.SingleValue, "[REQUIRED ALL downloaders] --from-date=yyyyMMdd-HH:mm:ss"),
                 new CommandLineOption("to-date", CommandOptionType.SingleValue, "[OPTIONAL for downloaders] If not provided 'DateTime.UtcNow' will "
                                                                                 + "be used. --to-date=yyyyMMdd-HH:mm:ss"),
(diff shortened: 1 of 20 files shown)


ToolBox/RandomDataGenerator/ConsoleLeveledOutput.cs
```diff
@@ -0,0 +1,39 @@
...
+            Info = Console.Out;
+            Warn = new FuncTextWriter(line =>
+            {
+                Console.ForegroundColor = ConsoleColor.Yellow;
+                Console.WriteLine(line);
+                Console.ResetColor();
+            });
+            Error = new FuncTextWriter(line =>
+            {
+                Console.ForegroundColor = ConsoleColor.Red;
```
mchandschuh: I'm such a sucker for colored output :P

Configuration/ToolboxArgumentParser.cs
```diff
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using Microsoft.Extensions.CommandLineUtils;
 
 namespace QuantConnect.Configuration
```
Jay-Jay-D: Missing doc entry for `randomdatagenerator` [here](https://github.com/mchandschuh/Lean/blob/32f7704b18351f1f7ef6a4f2b7e7e4712ce02ebf/Configuration/ToolboxArgumentParser.cs#L26)

ToolBox/RandomDataGenerator/RandomDataGeneratorSettings.cs
```diff
@@ -0,0 +1,263 @@
...
+            TickType[] tickTypes;
+            Resolution resolution;
+            DataDensity dataDensity;
+            SecurityType securityType;
+            DateTime startDate, endDate;
+
+            // --start
+            if (!DateTime.TryParseExact(startDateString, DateFormat.EightCharacter, null, DateTimeStyles.None, out startDate))
+            {
+                output.Error.WriteLine($"Required parameter --start was incorrectly formatted. Please specify in yyyy-MM-dd format. Value provided: '{startDateString}'");
```
Jay-Jay-D: Inconsistency between the documentation and the functionality. If the date is formatted as `yyyy-MM-dd` it is not parsed.
imagen

QuantConnect.Lean.sln.DotSettings
```diff
@@ -1,8 +1,12 @@
...
     <s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String>
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/INDENT_INVOCATION_PARS/@EntryValue">INSIDE</s:String>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_AFTER_INVOCATION_LPAR/@EntryValue">True</s:Boolean>
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_ARGUMENTS_STYLE/@EntryValue">CHOP_IF_LONG</s:String>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/WRAP_BEFORE_INVOCATION_RPAR/@EntryValue">True</s:Boolean>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateInstanceFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="_" Suffix="" Style="aaBb" /&gt;</s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
```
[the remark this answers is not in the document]
mchandschuh: Great catch! This commit was only supposed to include the addition of `Quant` to the team-shared dictionary. I've removed these values, although, after doing [some](https://resharper-support.jetbrains.com/hc/en-us/community/posts/206653075--sln-DotSettings-file-keeps-being-changed) [research](https://stackoverflow.com/a/29110662/1582922), it looks like it's ReSharper noting that it _has_ applied some settings and it shouldn't update it in the future... but it still seems suspect to me whether or not they should be shared, so I've removed them from this PR.

Tests/ToolBox/RandomDataGenerator/RandomValueGeneratorTests.cs
```diff
@@ -0,0 +1,350 @@
...
+        {
+            var min = new DateTime(2000, 01, 01);
+            var max = min.AddDays(-1);
+            Assert.Throws<ArgumentException>(() =>
+                randomValueGenerator.NextDate(min, max, dayOfWeek: null)
+            );
+        }
+
+        [Test]
+        public void NextDateTime_ThrowsArgumentExceptionException_WhenRangeIsTooSmallToProduceDateTimeOnRequestedDayOfWeek()
```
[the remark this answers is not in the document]
mchandschuh: Nice spot -- will fix 👍

ToolBox/RandomDataGenerator/RandomDataGeneratorSettings.cs
mchandschuh: yup, yup, it was changed to be in-sync with the other date format -- yet again --another amazing find, love the thorough review!! I will update the documentation. I'll also check the error validation messages and confirm they have the correct format, if they mention it as well

mchandschuh:
RE: equity-option -- recommend generating equity underlying data and then looping through that to generate options data for N contracts -- would be super cool to back out reasonable options prices based on black scholes + small % randomization to mimick market inefficiencies

PR #2842, opened 2019-01-20
Bug 2840 random data generator error trade quote same time

### LINKED ISSUE #2840: RandomDataGenerator Fails to Create Futures Data
#### Expected Behavior

I am able to generate futures data

#### Actual Behavior

I am not able to generate futures data. The exception message reads: 

```
Unhandled Exception: System.ArgumentException: The provided maximum percent deviation must be a postive number
   at QuantConnect.ToolBox.RandomDataGenerator.RandomValueGenerator.NextPrice(SecurityType securityType, String market, Decimal referencePrice, Decimal maximumPercentDeviati
on) in C:\Users\gsal\Projects\lean\ToolBox\RandomDataGenerator\RandomValueGenerator.cs:line 70
```
On line 63 in `TickGenerator.cs`, there is a call to `NextTick` with the following parameters: 
```C#
var nextQuote = _random.NextTick(symbol, next, TickType.Quote, nextTrade.Value, 0m);
```
Following the method call, we get brought to the file `RandomValueGenerator.cs` and method `NextTick`. On line 208, `NextTick` is called:
```C#
var next = NextPrice(symbol.SecurityType, symbol.ID.Market, previousValue, maximumPercentDeviation);
```
In the method `NextPrice`, there exists a check to ensure that `maximumPercentDeviation` cannot be a value below zero.
```C#
if (maximumPercentDeviation <= 0)
{
    throw new ArgumentException("The provided maximum percent deviation must be a postive number");
}
```

However, given that the exception will be thrown for all values equal to or less than zero, our original invocation with `maximumPercentDeviation = 0m` will always trigger this exception.

#### Potential Solution

Change the `maximumPercentDevation` check to not include 0:

```C#
if (maximumPercentDeviation < 0)
{
    throw new ArgumentException("The provided maximum percent deviation must be a postive number");
}
```
#### Reproducing the Problem

Run the following command:

```
.\QuantConnect.ToolBox.exe --app=rdg --from-date="20131006" --to-date="20131013" --security-type=Future --market=usa --symbol-count=1 --data-density=VerySparse --resolution=Second
```

#### System Information

Windows 10

#### Checklist

mchandschuh:
#### Description
BUG #2840: Prevent emitting trade&quote in same time step

This side steps the issue reported in #2840 by removing the desire to pass
zero for the maximum deviation. The previous issue was that we were leaning
on the trade ticks to produce the series variation and then trying to fit
quote ticks around a previously generated trade tick. This solution permits
both quote and trade ticks to produce variations and prevents the generator
from emitting both at the same time step.

A new parameter, --quote-trade-ratio, determines the relative density of each.
For example, a quote trade ratio of 1 means equal trade and quote ticks, whereas
a value of 2 means twice as many quote ticks as trade ticks.

This change also removes special treatment regarding the downsizing of the
requested deviation for quote ticks. If the consumer wants to limit the
deviation of quote ticks than the consumer can make that decision, but the
random value generator should simply follow instructions like a good little
boy.

#### Related Issue

Closes #2840

#### Motivation and Context

The `RandomDataGenerator` was throwing exceptions for all security types requiring both trade and quote ticks.

Perhaps -- added additional optional parameter `--quote-trade-ratio` defaulting to unity.

Ran futures data and confirmed quote and trade files written.

#### Types of changes

#### Checklist:


ToolBox/TickAggregator.cs
```diff
@@ -44,7 +44,13 @@ protected TickAggregator(Resolution resolution, TickType tickType)
         /// </summary>
         public List<BaseData> Flush()
         {
-            return new List<BaseData>(Consolidated) { Consolidator.WorkingData as BaseData };
+            var data = new List<BaseData>(Consolidated);
+            if (Consolidator.WorkingData == null)
```
jaredbroad: This bit is odd; possible just need to invert the check to `!=`

mchandschuh: Yes, this is certainly odd -- thanks for spotting -- I've performed a `--fixup` on the `HEAD` commit of this PR fixing the offending conditional.

PR #2867, opened 2019-01-28
Fix bitfinex order event status filled

voidale:
OrderEvent for status filled is not being fired because filled amounts are not being accumulated and compared to the right value.

Looking into the API documentation "oc" event is called for order "canceled" and "executed" added a case for executed to call final "EmitFillOrder" to send a "Filled" order event.


voidale:
Taking a second look that "executed" event might never happen as it's unclear if it's just for V2/V1 API's. And we might need to accumulate amounts just as GDAX has it and use the "tu" event.

StefanoRaggi:
@voidale Thank you for reporting the bug!
As you correctly pointed out, quantities for partial fills need to be accumulated to update the order status properly.
The proposed fix has been implemented in #2879 by @Adalyat

Issue #2859, opened 2019-01-24
Improve Random Data Generator to Generate Mapping and Factor Files


jaredbroad:
#### Expected Behavior
LEAN [Random Data Generator](https://github.com/QuantConnect/Lean/tree/master/ToolBox/RandomDataGenerator) Stock delistings, mappings, splits and dividend payments would be randomly generated. 

#### Actual Behavior
Currently stocks never get delisted, and all start at the same time.

#### Potential Solution

- Randomize the start date and length of the company lifespan so not all companies start with data. Assume 5% IPO after the state date specified in the generator so if requesting only 5 stocks it still generates a pretty solid dataset.

- Generate a random dividend profit/loss every quarter (including no dividend), include this in the data generated.

- Generate symbol rename(mapping) events to handle renaming of the stocks. This may require generating a new ticker as well since there will be new company name (JP Morgan -> "JP Morgan Chase"). 

- Have 1% of companies get delisted within a year.

- Include a minimum lifespan of 6months (realistically companies rarely IPO then disappear within 6mo).

- Once the backtest is completed, write these map and factor files to disk for all the symbols. 

- Map and Factor files are always saved with the _current_ ticker. Previous tickers do not 

So the generator should write the final `newname.csv` to disk. The interim names should _not_ have a factor/map generated. The old companies with the same symbol are written as `name.1.csv`

Full documentation here:
 - https://www.quantquote.com/docs/TickMAP_Overview.pdf
 - https://www.quantquote.com/docs/QuantQuote_Minute.pdf
Highly recommend reading all the documentation, its not a simple concept.  

Example Real Map files:
 - https://github.com/QuantConnect/Lean/tree/master/Data/equity/usa/map_files

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

jaredbroad:
Closed by #2868

PR #2868, opened 2019-01-28
RandomDataGenerator Mapping, Delisting, Splits, and Dividends Generator

gsalaz98:
Create mapping, IPO, delisting, splits, and dividends random generation

#### Description

Created random-seed argument for rdg in order to let the user get deterministic output

Update documentation in 'FactorFileRow.cs' to accurately reflect factor
file structure

Update CSV generation for FactorFile so that it uses FactorFileRow's CSV
generator

Add FinancialCalendar to make it easier to implement logic regarding
financial quarters

Add mapping events to RandomDataGenerator

Update MapFileRow ToCsv method to correctly emit the ticker as lowercase

Fixed bug in FactorFile where we would get the same initial data point twice
when we converted it to CSV

Create new method to convert a MapFile to CSV

Create new method to write MapFile to disk as CSV

Add unit test to test for successful CSV generation in MapFile

Add new files to project

Add FinancialCalendar unit tests

Create new class to handle generation of dividends, splits, and maps

#### Related Issue

#2859 
#### Motivation and Context

Allows the user to generate data for backtest benchmarking and testing

N/A

Add new unit tests to cover some changes

#### Types of changes

#### Checklist:


Jay-Jay-D:
Seems pretty solid PR Gerardo.

However, I cannot test its behavior, the CLI is not working as expected.

image

Also, I'd suggest to explicit the default values for the optional arguments.

Common/Data/Auxiliary/MapFile.cs
```diff
@@ -131,6 +143,16 @@ public static MapFile Read(string permtick, string market)
             return new MapFile(permtick, MapFileRow.Read(permtick, market));
         }
 
+        /// <summary>
+        /// Writes the map file to a CSV file
+        /// </summary>
+        /// <param name="market">The market to save the MapFile to</param>
+        public void WriteToCsv(string market) 
+        {
+            var filePath = GetMapFilePath(Permtick, market);
```
Jay-Jay-D: Maybe we can heck if the folder exist first.

ToolBox/RandomDataGenerator/DividendSplitMapGenerator.cs
```diff
@@ -0,0 +1,203 @@
...
+        /// </summary>
+        /// <param name="tickHistory"></param>
+        public void GenerateSplitsDividends(IEnumerable<Tick> tickHistory)
+        {
+            var previousMonth = -1;
+            var monthsTrading = 0;
+
+            var hasRename = _randomValueGenerator.NextBool(30.0);
+            var hasSplits = _randomValueGenerator.NextBool(15.0);
+            var hasDividends = _randomValueGenerator.NextBool(60.0);
```
Jay-Jay-D: Is there any reason for those values? 

Maybe they can be parameterized?

Jay-Jay-D:
Great Job @gsalaz98!

Just a couple of details.

Common/Data/Auxiliary/MapFile.cs
gsalaz98: What kind of behavior should we be aiming for here? I'm thinking about writing an error message to the log and returning with no write attempted if the directory can't be found

Jay-Jay-D: I think that with a simple `Directory.Create(path)` is enough. If there is no folder it'll be created, if it exists, nothing happens

PR #2905, opened 2019-02-13
Orderticketbranch

simonsonjack:
Made updated to OrderTickDemoAlgorithm after an issue noticed by a community member. Changes align setting Limit Price in Limit Orders and Stop Limit Orders according to best/industry standard practice. Additionally, removed ShareClassMeanReversionAlpha since it is not properly implemented and should be published with updates once Alpha is deployed properly.


jaredbroad:
https://www.investopedia.com/university/intro-to-order-types/limit-orders.asp

>Limit orders let you enter and exit trades with precision; however, you have to enter them correctly to ensure they accomplish the goal of improving price – that is, to get a specified price or better. It’s important to be on the right side of market, otherwise you’ll get filled at the market price. When placing a buy limit order, specify a price that’s at or below the current bid; for a sell limit order, specify a price that’s at or above the current ask. 

Many brokerages don't allow placing limit orders on the wrong side of the book. You *must* place it below the price to be accepted.

PR #2953, opened 2019-02-25
Handle unexpected IB disconnected state in cash sync during reset period

StefanoRaggi:
#### Description
Occasionally the disconnection due to the IB reset period might last much longer than expected during weekends (even up to the cash sync time at 7:45 AM NY). 

System log inspection revealed that the IB disconnect notification message (Error code 1100) was  received once per minute up until cash sync time. Algorithms were then terminated because of the disconnected state detected after a connection attempt.

This behavior has been observed only twice in the last months:
- Saturday, Jan 19th, 2019
- Saturday, Feb 23rd, 2019

When this happens, now we do not try to reconnect (since this would fail anyway) but we return the existing balances instead.

#### Motivation and Context
Live IB algorithms terminated due to connection failure at cash sync time.

No.

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -435,15 +435,26 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
 
             if (!IsConnected)
             {
-                Log.Trace("InteractiveBrokersBrokerage.GetCashBalance(): not connected, connecting now");
-                Connect();
+                if (IsWithinScheduledServerResetTimes())
+                {
+                    // Occasionally the disconnection due to the IB reset period might last
+                    // much longer than expected during weekends (even up to the cash sync time).
+                    // In this case we do not try to reconnect (since this would fail anyway)
+                    // but we return the existing balances instead.
+                    Log.Trace("InteractiveBrokersBrokerage.GetCashBalance(): not connected within reset times, returning existing balances");
+                }
+                else
+                {
+                    Log.Trace("InteractiveBrokersBrokerage.GetCashBalance(): not connected, connecting now");
+                    Connect();
+                }
             }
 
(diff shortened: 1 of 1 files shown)


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -435,15 +435,26 @@ public override List<CashAmount> GetCashBalance()
...
             {
-                Log.Trace("InteractiveBrokersBrokerage.GetCashBalance(): not connected, connecting now");
-                Connect();
+                if (IsWithinScheduledServerResetTimes())
+                {
+                    // Occasionally the disconnection due to the IB reset period might last
+                    // much longer than expected during weekends (even up to the cash sync time).
+                    // In this case we do not try to reconnect (since this would fail anyway)
+                    // but we return the existing balances instead.
+                    Log.Trace("InteractiveBrokersBrokerage.GetCashBalance(): not connected within reset times, returning existing balances");
```
Martin-Molinero: `IsWithinScheduledServerResetTimes()` will return true on `Saturdays` and from 11:00pm->1:30am, right?
I'm not sure how will this new `if` statement, cover for this case `...much longer than expected during weekends (even up to the cash sync time)...`? from what I understand it will go through `Connect();` as before

StefanoRaggi: I was considering a disconnection from e.g. Friday night at 11 pm to Saturday morning at 7:45 am as `much longer than expected`.
At 7:45, with current master branch code, we are expecting to be able to connect successfully (and it seems to not be always the case).
With this change, we are able to still be disconnected until Monday at 7:45 without killing the algorithm and hopefully, we'll be able to connect then 😃

jaredbroad:
Acceptable to delay synchronizing cash until outside of the server reset times.

Issue #2820, opened 2019-01-11
Daily and Hour Resolution History Requests Slow


Martin-Molinero:
#### Expected Behavior
- Hour resolution history request data throughput should be similar to other resolutions.

#### Actual Behavior
- Hour resolution history request is an order of magnitude slower than for example a minute resolution history request. This is because the entire hour data for each symbol is stored **in a single file** causing a big overhead for just fetching one data point (vs separated daily zips for Minute Resolution)

#### Potential Solution
- Maybe we could jump around in the file without actually iterating through every line by using a binary search algorithm + [File.Seek](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.seek?view=netframework-4.5.2).

#### Reproducing the Problem
- Execute an algorithm performing a `History(_symbol, 1, Resolution.Hour);` on each OnData.

#### System Information
N/A
#### Checklist 
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

jaredbroad:
#### Step 2: Caching, and only _reading_ the history once from disk. 
Definitely the fastest idea from @gsalaz98; but will need some smarts to prevent memory explosion. Cache the results of the history call and only parse it from disk once. Imagine `History` asking for 1 minute bar - every minute of the day; it will parse the minute zip file 390x per day. Caching the instances parsed will reduce work done 390x (less overhead for returning relevant rows and caching).

jaredbroad:
Caching sounds worth a shot. We might need to be smart about who gets cached though. The daily data is about 6GB in memory so can't cache it all.  

IMHO the key issue isn't parallel or parsing the data -- its the design of the data feed which disposes of the entire enumerator stack each day - even for files which have more than 1-day of data. This causes the feed to need to start from scratch each day, re-enumerating the whole daily file each day.

stan4git:
PR is out.
https://github.com/QuantConnect/Lean/pull/2899

6GBs is specifically for shared data web service(remote data source). Not for standalone LEAN instance.

On interesting finding during the performance profile is - Each LEAN backtest instance spins up about 16 user threads, two of which are taking about 90% of the time/computation while others doing very little work.

[...]

PR #2975, opened 2019-02-28
Cache raw data points

Martin-Molinero:
#### Description

- `TextSubscriptionDataSourceReader` will now cache raw price mode data
points for each requested source file, per data type.
   - Caching will be handled by the `MemoryCache` class. Max physical memory usage will be 70%. 
Cache entry should be evicted if it has not been accessed in 5 minutes.
   - Adding unit tests
- `GetLastKnownPrice()` will not `RoundDown` end time. This was causing
it to fail to fetch any data point in some cases.

### CLOUD performance testing
```
        {
            SetStartDate(2016, 01, 01);
            SetEndDate(2018, 01, 01);
            SetCash(10000);
            _symbol = AddEquity("SPY").Symbol;
        }
        public override void OnEndOfDay()
        {
            var dailyHistory = History(_symbol, 1, Resolution.Daily/Hour).First();
        }
```

**PR Resolution.Hour**
4.62 seconds at 42k data points per second. Processing total of 196,141 data points.
4.38 seconds at 45k data points per second. Processing total of 196,141 data points.
4.86 seconds at 40k data points per second. Processing total of 196,141 data points.
**PR Resolution.Daily**
5.13 seconds at 38k data points per second. Processing total of 196,141 data points.
4.27 seconds at 46k data points per second. Processing total of 196,141 data points.
4.02 seconds at 49k data points per second. Processing total of 196,141 data points.
**MASTER Resolution.Hour**
82.92 seconds at 2k data points per second. Processing total of 196,141 data points.
82.48 seconds at 2k data points per second. Processing total of 196,141 data points.
85.68 seconds at 2k data points per second. Processing total of 196,141 data points.
**MASTER Resolution.Daily**
16.73 seconds at 12k data points per second. Processing total of 196,141 data points.
17.46 seconds at 11k data points per second. Processing total of 196,141 data points.
15.71 seconds at 12k data points per second. Processing total of 196,141 data points.

**Result: ~20x improvement for `Resolution.Hour`. ~4x `Resolution.Daily`**

### Performance Analysis
**MASTER Resolution.Daily**
imagen

**MASTER Resolution.Hour**
imagen

Both `Resolution.Daily` and `Resolution.Hour` data files are organized in 1 zip file per symbol (for the entire data history). **Each** history request has to open, read and parse this only data file. The objective of this PR is to cache and reuse this raw data.

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2820
#### Motivation and Context

Daily and Hour Resolution History Requests were adding significant overhead to overall performance.

N/A

Unit and regression tests. Performance testing in the cloud. Live deployment. Local testing
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
+++ b/Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
@@ -155,7 +155,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                     $"Type is: {cacheItem.Value.GetType()}");
             }
             // Find the first data point 10 days (just in case) before the desired date
-            // and subtract one item (just in case there wasn't a time gap and data.Time is after _date)
+            // and subtract one item (just in case there was a time gap and data.Time is after _date)
             var index = cache.FindIndex(data => data.Time.AddDays(10) > _date);
             index = index > 0 ? (index - 1) : 0;
             foreach (var data in cache.Skip(index))
--- a/Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
+++ b/Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
@@ -15,11 +15,16 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.ComponentModel;
+using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Interfaces;
 using QuantConnect.Lean.Engine.DataFeeds.Transport;
 using QuantConnect.Util;
+using System.Runtime.Caching;
+using QuantConnect.Data.Fundamental;
+using QuantConnect.Data.UniverseSelection;
 
 namespace QuantConnect.Lean.Engine.DataFeeds
 {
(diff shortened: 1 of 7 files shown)


Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
```diff
@@ -34,7 +39,16 @@ public class TextSubscriptionDataSourceReader : ISubscriptionDataSourceReader
...
         private readonly SubscriptionDataConfig _config;
+        private bool _shouldCacheDataPoints;
         private readonly IDataCacheProvider _dataCacheProvider;
+        private static readonly MemoryCache BaseDataSourceCache = new MemoryCache("BaseDataSourceCache",
+            // Cache can use up to 70% of the installed physical memory
+            new NameValueCollection{ { "physicalMemoryLimitPercentage", "70"} });
+        private static readonly CacheItemPolicy CachePolicy = new CacheItemPolicy
+        {
+            // Cache entry should be evicted if it has not been accessed in given span of time:
+            SlidingExpiration = TimeSpan.FromMinutes(5)
```
stan4git: LRU instead of time-based caching might make more sense here.

Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
```diff
@@ -69,6 +83,8 @@ public TextSubscriptionDataSourceReader(IDataCacheProvider dataCacheProvider, Su
             _config = config;
             _isLiveMode = isLiveMode;
             _factory = (BaseData) ObjectActivator.GetActivator(config.Type).Invoke(new object[] { config.Type });
+            _shouldCacheDataPoints = !_config.IsCustomData && _config.Resolution >= Resolution.Hour
```
stan4git: Why not cache higher frequency data?

Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
Martin-Molinero: Added this sliding expiration policy to avoid reaching the memory limit in a long running algorithm. I believe LRU will be used when the memory limit is reached.

Engine/DataFeeds/TextSubscriptionDataSourceReader.cs
Martin-Molinero: Both `Resolution.Daily` and `Resolution.Hour` data files are organized in 1 zip file per symbol (for the entire data history). Higher frequency data (`Minute`, `Second`, `Tick`) are organized in daily files. So there is a big difference in potential performance improvement. So as an initial approach tackling `Resolution.Daily` and `Resolution.Hour` seems like a good idea, also having in mind the potential RAM usage too.

PR #3000, opened 2019-03-13
Support custom currency length

### LINKED ISSUE #996: [Crypto] Enable Currency Abbreviations with More than 3 Characters

### LINKED ISSUE #2874: Custom Account Currenty has a limitation on its length
#### Expected Behavior

Custom Account currency should accept any custom currency with different length. For example, Binance uses USDT (pegged currency) as account curreny.

#### Actual Behavior

I setup an Account curreny using
`SetAccountCurrency("USDT");`
And get an ArgumentException "Cash symbols must be exactly 3 characters."

#### Reproducing the Problem

Use any Binance account to replicate the problem.

#### Checklist

Adalyat:
#### Description

Parse crypto symbols using symbol properties db. Using entries from this db we can identify base and quote currencies for each symbol.

#### Related Issue

Closes #996
Closes #2874

#### Motivation and Context

It allows us to support new brokerages with different currencies including stable coins that can have length different from 3 (i.e. USDT, USDC etc) 

Autotests; manual testing

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -1388,10 +1388,10 @@ namespace QuantConnect
         }
 
         /// <summary>
-        /// Safely remove ending
+        /// Returns a new string in which specified ending in the current instance is removed.
         /// </summary>
-        /// <param name="s"></param>
-        /// <param name="ending"></param>
+        /// <param name="s">original string value</param>
+        /// <param name="ending">the string to be removed</param>
         /// <returns></returns>
         public static string RemoveFromEnd(this string s, string ending)
         {
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -1386,5 +1386,23 @@ namespace QuantConnect
         {
             task.ConfigureAwait(false).GetAwaiter().GetResult();
         }
+
+        /// <summary>
+        /// Safely remove ending
+        /// </summary>
+        /// <param name="s"></param>
+        /// <param name="ending"></param>
+        /// <returns></returns>
+        public static string RemoveFromEnd(this string s, string ending)
+        {
+            if (s.EndsWith(ending))
+            {
+                return s.Substring(0, s.Length - ending.Length);
+            }
+            else
+            {
+                return s;
+            }
+        }
     }
 }
--- a/Common/Securities/Crypto/Crypto.cs
+++ b/Common/Securities/Crypto/Crypto.cs
@@ -19,6 +19,7 @@ using QuantConnect.Orders.Fees;
 using QuantConnect.Orders.Fills;
 using QuantConnect.Orders.Slippage;
 using QuantConnect.Securities.Forex;
+using System;
 
 namespace QuantConnect.Securities.Crypto
 {
@@ -58,9 +59,15 @@ namespace QuantConnect.Securities.Crypto
             Holdings = new CryptoHolding(this, currencyConverter);
 
             // decompose the symbol into each currency pair
-            string baseCurrencySymbol, quoteCurrencySymbol;
-            Forex.Forex.DecomposeCurrencyPair(config.Symbol.Value, out baseCurrencySymbol, out quoteCurrencySymbol);
-            BaseCurrencySymbol = baseCurrencySymbol;
+            string quoteCurrencySymbol = symbolProperties.QuoteCurrency;
+            if (config.Symbol.Value.EndsWith(quoteCurrencySymbol))
+            {
+                BaseCurrencySymbol = config.Symbol.Value.RemoveFromEnd(quoteCurrencySymbol);
+            }
+            else
+            {
+                throw new InvalidOperationException($"symbol doesn't end with {quoteCurrencySymbol}");
+            }
         }
 
         /// <summary>
@@ -93,9 +100,15 @@ namespace QuantConnect.Securities.Crypto
             Holdings = new CryptoHolding(this, currencyConverter);
 
             // decompose the symbol into each currency pair
-            string baseCurrencySymbol, quoteCurrencySymbol;
-            Forex.Forex.DecomposeCurrencyPair(symbol.Value, out baseCurrencySymbol, out quoteCurrencySymbol);
-            BaseCurrencySymbol = baseCurrencySymbol;
+            string quoteCurrencySymbol = symbolProperties.QuoteCurrency;
+            if (symbol.Value.EndsWith(quoteCurrencySymbol))
+            {
+                BaseCurrencySymbol = symbol.Value.RemoveFromEnd(quoteCurrencySymbol);
+            }
+            else
+            {
+                throw new InvalidOperationException($"symbol doesn't end with {quoteCurrencySymbol}");
+            }
         }
 
         /// <summary>
--- a/Common/Securities/Crypto/Crypto.cs
+++ b/Common/Securities/Crypto/Crypto.cs
@@ -59,15 +59,9 @@ namespace QuantConnect.Securities.Crypto
             Holdings = new CryptoHolding(this, currencyConverter);
 
             // decompose the symbol into each currency pair
-            string quoteCurrencySymbol = symbolProperties.QuoteCurrency;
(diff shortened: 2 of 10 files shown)


Common/Extensions.cs
```diff
@@ -1304,5 +1304,23 @@ public static void SynchronouslyAwaitTask(this Task task)
         {
             task.ConfigureAwait(false).GetAwaiter().GetResult();
         }
+
+        /// <summary>
+        /// Safely remove ending
+        /// </summary>
+        /// <param name="s"></param>
+        /// <param name="ending"></param>
+        /// <returns></returns>
```
Martin-Molinero: This method is missing some of the documentation

Tests/Common/Securities/Cryptos/CryptoTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    [TestFixture]
+    public class CryptoTests
+    {
+        private static TimeKeeper TimeKeeper
+        {
+            get { return new TimeKeeper(DateTime.Now, new[] { TimeZones.NewYork }); }
+        }
+
+        [Test]
+        [TestCase("BTCUSD", "USD")]
```
Martin-Molinero: Maybe it would be good to add a test with some crypto whose quote currency isn't `USD`

Common/Securities/Crypto/Crypto.cs
```diff
@@ -58,9 +59,15 @@ public Crypto(SecurityExchangeHours exchangeHours, Cash quoteCurrency, Subscript
...
-            BaseCurrencySymbol = baseCurrencySymbol;
+            string quoteCurrencySymbol = symbolProperties.QuoteCurrency;
+            if (config.Symbol.Value.EndsWith(quoteCurrencySymbol))
+            {
+                BaseCurrencySymbol = config.Symbol.Value.RemoveFromEnd(quoteCurrencySymbol);
+            }
+            else
+            {
+                throw new InvalidOperationException($"symbol doesn't end with {quoteCurrencySymbol}");
+            }
```
Martin-Molinero: Maybe we could move this into its own method, similar to `Forex.DecomposeCurrencyPair` and add some unit tests for it, verifying different Crypto pair cases

Common/Securities/SymbolPropertiesDatabase.cs
```diff
@@ -35,6 +35,34 @@ private SymbolPropertiesDatabase(IReadOnlyDictionary<SecurityDatabaseKey, Symbol
...
+
+        /// <summary>
+        /// Check whether symbol properties exists for the specified market/symbol/security-type
+        /// </summary>
+        /// <param name="market">The market the exchange resides in, i.e, 'usa', 'fxcm', ect...</param>
+        /// <param name="symbol">The particular symbol being traded (Symbol class)</param>
+        /// <param name="securityType">The security type of the symbol</param>
+        public bool ContainsKey(string market, Symbol symbol, SecurityType securityType)
+        {
+            var stringSymbol = symbol == null ? string.Empty :
```
Martin-Molinero: There is a `public static MarketHoursDatabase.GetDatabaseSymbolKey()` that handles this logic, maybe it would be better to directly call it instead

Martin-Molinero:
Please verify that for Crypto the new `Decomposer` is used instead of the `Forex.DecomposeCurrencyPair`/or the old way, example https://github.com/QuantConnect/Lean/blob/master/Brokerages/Oanda/OandaBrokerage.cs#L155 should be replaced too, else will cause runtime errors

Adalyat:
> Please verify that for Crypto the new `Decomposer` is used instead of the `Forex.DecomposeCurrencyPair`/or the old way, example https://github.com/QuantConnect/Lean/blob/master/Brokerages/Oanda/OandaBrokerage.cs#L155 should be replaced too, else will cause runtime errors

I've checked all `Forex.DecomposeCurrencyPair` references and can't see where I need to replace it including what you mentioned. Firstly, the line you highlighted under `SecurityType.Forex` condition so `Forex.DecomposeCurrencyPair` can be applied successfully here, isn't it?
Secondly, `Crypto.DecomposeCurrencyPair` will depend on SecurityProperties to get quoteCurrency properly, that's why it's difficult to use it everywhere.

Common/Securities/Crypto/Crypto.cs
Adalyat: I'm okay with that. Does it mean that this new method should accept symbol and symbolProperties? The main problem having symbol props is strong restriction and usually, to my knowledge, its not available everywhere

Martin-Molinero:
> I've checked all Forex.DecomposeCurrencyPair references and can't see where I need to replace it including what you mentioned. Firstly, the line you highlighted under SecurityType.Forex condition so Forex.DecomposeCurrencyPair can be applied successfully here, isn't it?
Secondly, Crypto.DecomposeCurrencyPair will depend on SecurityProperties to get quoteCurrency properly, that's why it's difficult to use it everywhere.

Firstly -> Your right! sorry about that.
Secondly -> Agree, that's why I was mentioning its important to verify/detect specifically where _everywhere_ is (_if any case_), so we can find a way around it, maybe even finding another source of the quote/base currency -> for example the `Crypto/Forex` `Security` have the data (see `IBaseCurrencySymbol`).

Common/Securities/Crypto/Crypto.cs
Martin-Molinero: Agree, replied extensively in https://github.com/QuantConnect/Lean/pull/3000#issuecomment-475625604

Common/Securities/SecurityService.cs
```diff
@@ -61,13 +62,17 @@ public class SecurityService : ISecurityService
...
-            if (symbol.ID.SecurityType == SecurityType.Forex || symbol.ID.SecurityType == SecurityType.Crypto)
+            if (symbol.ID.SecurityType == SecurityType.Forex)
             {
                 defaultQuoteCurrency = symbol.Value.Substring(3);
             }
 
-            var symbolProperties = _symbolPropertiesDatabase.GetSymbolProperties(symbol.ID.Market, symbol, symbol.ID.SecurityType, defaultQuoteCurrency);
+            if (symbol.ID.SecurityType == SecurityType.Crypto && !_symbolPropertiesDatabase.ContainsKey(symbol.ID.Market, symbol, symbol.ID.SecurityType))
+            {
+                throw new ArgumentException($"symbol can't be found in csv: {symbol.Value}");
```
jaredbroad: Please update to "Symbol can't be found in the Symbol Properties Database:" + the unit test case

Adalyat: Done

Common/Securities/Crypto/Crypto.cs
```diff
@@ -111,5 +112,25 @@ public Crypto(Symbol symbol, SecurityExchangeHours exchangeHours, Cash quoteCurr
...
+        public static void DecomposeCurrencyPair(Symbol symbol, SymbolProperties symbolProperties, out string baseCurrency, out string quoteCurrency)
+        {
+            quoteCurrency = symbolProperties.QuoteCurrency;
+            if (symbol.Value.EndsWith(quoteCurrency))
+            {
+                baseCurrency = symbol.Value.RemoveFromEnd(quoteCurrency);
+            }
+            else
+            {
+                throw new InvalidOperationException($"symbol doesn't end with {quoteCurrency}");
```
jaredbroad: Perhaps to maintain backward compatibility for if not in SymbolDB? 
`Forex.Forex.DecomposeCurrencyPair(symbol.Value, out baseCurrency, out quoteCurrency);`

jaredbroad:
Merging but suggest we think about rejigging this implementation from IF branches to layered attempts:

1) Use the separator character in ticker string if the user provides one.
2) Use Symbol DB if we find it in the DB.
3) Fall back on using the current 6 character split if 1,2 fail to provide viable options.

PR #3002, opened 2019-03-14
Fix Time Zone Bug in SubscriptionDataReader

StefanoRaggi:
#### Description
In a few places, `SubscriptionDataReader` was incorrectly assuming the `TradeableDates` were in `ExchangeTimeZone` instead of `DataTimeZone`. This issue was found specifically with CFD assets which have the exchange time zone **ahead** of the data time zone.

- `SubscriptionDataReader` has been updated to fix the bug mentioned above
- A new `CfdTimeZonesRegressionAlgorithm` has been added
- The order event logging in `BasicTemplateFuturesAlgorithm` has been removed (almost doubling the speed for this regression test)

#### Motivation and Context
Incorrect dates used when backtesting CFD assets which have the exchange time zone ahead of the data time zone.

No.

New regression test included.

#### Types of changes

#### Checklist:


Engine/DataFeeds/SubscriptionDataReader.cs
```diff
@@ -372,7 +372,7 @@ public bool MoveNext()
 
                     // if we move past our current 'date' then we need to do daily things, such
                     // as updating factors and symbol mapping
-                    if (instance.EndTime.Date > _tradeableDates.Current)
+                    if (instance.EndTime.ConvertTo(_config.ExchangeTimeZone, _config.DataTimeZone).Date > _tradeableDates.Current)
```
Martin-Molinero: Nice! This could actually mean we were losing some data points too, right? Depending on the diff between the data and the exchange time zone, we might have created a new enumerator before time

StefanoRaggi: Yes, exactly 👍

PR #3070, opened 2019-04-11
Fix uncorrelated universe

HalldorAndersen:
#### Description
A bug fix for the Uncorrelated Universe Selection Model.

#### Related Issue

The Uncorrelated Universe Selection Model retrieves history for newly added symbols to the universe. Some symbols might have short data history which might result in an error.

#### Motivation and Context

History for all new symbols is retrieved at once and combined in one DataFrame.

This fix fills the DataFrame NA values for symbols that have short data history with the latest price when security has too few data points (a rare occurrence). Further, this fix allows for ranking all securities on dollar volume and without changing the remaining code structure.

Tested in the cloud using ConstantAlphaModel and EmergingMarketsHFIAlphaModel.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Selection/UncorrelatedUniverseSelectionModel.py
+++ b/Algorithm.Framework/Selection/UncorrelatedUniverseSelectionModel.py
@@ -76,11 +76,11 @@ class UncorrelatedUniverseSelectionModel(FundamentalUniverseSelectionModel):
             if not data.IsReady:
                 newSymbols.append(symbol)
 
-        # Warm up the dictionary objects of selected symbols and benchmark that do not have enought data
+        # Warm up the dictionary objects of selected symbols and benchmark that do not have enough data
         if len(newSymbols) > 1:
             history = algorithm.History(newSymbols, self.historyLength, Resolution.Daily)
-            if not (history.empty or len(history.index) < self.historyLength):
-                history = history.close.unstack(level=0).dropna()
+            if not history.empty:
+                history = history.close.unstack(level=0).fillna(method='bfill')
                 for symbol in newSymbols:
                     self.cache[symbol].Warmup(history)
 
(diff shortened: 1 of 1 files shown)


Algorithm.Framework/Selection/UncorrelatedUniverseSelectionModel.py
```diff
@@ -76,12 +76,13 @@ def SelectCoarse(self, algorithm, coarse):
...
 
-        # Warm up the dictionary objects of selected symbols and benchmark that do not have enought data
+        # Warm up the dictionary objects of selected symbols and benchmark that do not have enough data
         if len(newSymbols) > 1:
             history = algorithm.History(newSymbols, self.historyLength, Resolution.Daily)
-            history = history.close.unstack(level=0)
-            for symbol in newSymbols:
-                self.cache[symbol].Warmup(history)
+            if not history.empty:
+                history = history.close.unstack(level=0).fillna(method='bfill')
```
AlexCatarino: We cannot use backward fill, `bfill`, to warm up indicators, because it will introduce look-ahead bias. Instead, we should use forward fill, `ffill`, and drop all remaining `nan`:
```python
history = history.close.unstack(level=0).ffill().dropna()
```

AlexCatarino: We cannot use backward fill, `bfill`, to warm up indicators, because it will introduce look-ahead bias. Instead, we should use forward fill, `ffill`, and drop all remaining `nan`:
```python
history = history.close.unstack(level=0).ffill().dropna()
```

HalldorAndersen: Using dropna() reduces the number of rows in the DataFrame when data history is not available for all symbols, and it will not solve the problem because we need a complete history to calculate the Z-score.

It is unlikely that securities with values filled using 'bfill' will score high on Z-score as there is no fluctuation in correlation until data becomes available.

AlexCatarino: We really should not use `bfill`.
If we don't have a complete history, maybe we should discard the symbol altogether (especially if it is unlikely to have high Z-score for it). 
Alternatively, we drop the `nan` from the symbol and only discard the symbol if x% was dropped. In this case, we will drop the same dates from the benchmark history as well to have arrays with the same size.

HalldorAndersen: There is no need to use `ffill` or `bfill`  as SymbolData returns 0 if it is not ready. Only condition needed to resolve the problem is `if not history.empty`.

Issue #2991, opened 2019-03-08
Update Futures Expiry functions and symbol properties


Jay-Jay-D:
#### Expected Behavior

All futures must be modeled correctly.

#### Actual Behavior

For available futures, only 33 have an expiry function defined and 29 have properties defined in the [symbol-properties-database](https://github.com/QuantConnect/Lean/blob/master/Data/symbol-properties/symbol-properties-database.csv#L87).

#### Potential Solution

Define all future's expiry functions and properties. 

#### Reproducing the Problem

N/A.

#### System Information

N/A.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [ ] I have provided detailed steps to reproduce the issue

Jay-Jay-D:
The values for `ZT` in the [`symbol-properties-database.csv`](https://github.com/QuantConnect/Lean/blob/5db98c8f532d8c6aab2998aeaecf3e1644bb44fa/Data/symbol-properties/symbol-properties-database.csv#L93) should be updated to correctly model bond futures behavior.

Martin-Molinero:
Believe we can close this one now. The [symbol-properties-database](https://github.com/QuantConnect/Lean/blob/master/Data/symbol-properties/symbol-properties-database.csv#L87) has 189 futures at the moment 🚀

PR #3094, opened 2019-04-17
Add Various Symbols To Futures Database (Part 3 / 11)

gsalaz98:
#### Description

Adds symbols AGA, AJL, AJS, AJY, AKL, AKZ, ANE, APS, AR0, ARE, AUP, AVZ, AW, AYV, AYX, AZ1, B0, B7H, BCF, BIO, BK, BOO, BR7 to futures expiration functions, market hours database, symbol-properties-database

Corrects wrong entry for symbol ZT in symbol-properties-database

Adds missing holidays or early closes for energies and fx in mhdb
#### Related Issue

#2991 
#### Motivation and Context

Ongoing completion efforts of the futures database

No

New unit tests
#### Types of changes

#### Checklist:


jaredbroad:
Thank you again for the PR @gsalaz98, please see if you can tidy up the test exceptions: https://api.travis-ci.org/v3/job/521068249/log.txt

Data/symbol-properties/symbol-properties-database.csv
```diff
@@ -84,13 +84,14 @@ usa,[*],option,,USD,100,0.01,1
...
 
+usa,BCF,future,,USD,50,0.25,1
 usa,ZC,future,,USD,50,0.25,1
 usa,ZW,future,,USD,50,0.25,1
 usa,ZS,future,,USD,50,0.25,1
 usa,ZM,future,,USD,100,0.1,1
 usa,ZL,future,,USD,600,0.01,1
 usa,ZO,future,,USD,50,0.25,1
-usa,ZT,future,,USD,2000,0.0078125,1
+usa,ZT,future,,USD,2000,0.00390625,1
```
Jay-Jay-D: Can you please describe how you got this value?

gsalaz98:
The source of all the new holidays and early closes I added were mainly Thanksgiving and Independence Day additions. I discovered that I had an incomplete list of holidays and early closes after browsing through the holiday spreadsheets and documents manually while implementing the Equity, Bloomberg, and Black Sea Corn groups.

As for the repeating holidays, (mainly 7/31 and 8/31), these holidays are respectively: Eid al-Adha (Hari Raya Haji) in Singapore (ending on 7/31/2020) and London Summer Bank Holidays. As for the reason why some symbols exhibit these holidays, I don't know completely, but I suspect it probably has to do with the symbol's origin (e.g. Singapore Gasoil (Platts) vs. Low Sulphur Gasoil Futures "AGA" closing for Eid al-Adha, but not other symbols). 

I did inquire about odd closing hours for the FX group, and received a message from CME stating that because the settlement bank doesn't accept deposits on a certain date (Columbus), the symbol has an exception for its Last Trade Date. Maybe in this scenario, it might be a similar case as the FX group, but with different holidays.

Another thing I want to point out: there are Good Friday holidays included in some symbols for 2024 (3/29/2024). We might be able to infer that CME will declare that Good Friday a holiday, consistent with previous published holidays. But as I've told Jared, there might exist inconsistencies with predicting future holidays (such as changed early closing hours), and the only way to know for sure is to look at the holiday calendar once it's been published by CME.

Data/symbol-properties/symbol-properties-database.csv
gsalaz98: From the sample data you've provided, I've verified that this new entry in symbol-properties-database is the correct value.

I did some digging into bonds and its terminology, and found out that "Face value at maturity of $200,000" means the par value of $200,000 in bonds, where par value is usually $100. With that being said, the par value of the bond is $2000 USD ($200000 / $100), which is our contract multiplier.

Following the instructions on the CME page, I verified that the tick value is in fact 1/256 because the sample data you've provided has minimum increments of 1/256ths. Not only that, if we multiply the Contract Multiplier and Minimum Price Variation, we should have the USD value per contract ($2000 * (1/256) = $7.8125). In addition, if the price of the future goes up by $1, the underlying value would have gone up by $2000 (per contract). Because the minimum price increment is 1/256 (i.e. 0.00390625), minimum_price_variation should be 0.00390625

Jay-Jay-D:
Thank you very much, Gerardo!

Issue #3123, opened 2019-04-24
Predefined Constituent List Universes


Martin-Molinero:
#### Expected Behavior

- Users can add to there algorithm `Universes` with pre-selected `Symbols`, for example, `QC500`, `Aviation Industry`, `Bank Universe`, `Russel3000` etc..
#### Actual Behavior

N/A
#### Potential Solution

N/A
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3141, opened 2019-04-26
Adding Constituents Universes

Martin-Molinero:
#### Description

- Adding base `ConstituentsUniverse` and `ConstituentsUniverseData`
- Adding new `Symbol.None` and `SecurityIdentifier.None` implementation
- Adding Py and C# regression test algorithms `ConstituentsUniverseRegressionAlgorithm`
- Adding C# `ConstituentsUniverseDataGeneratorAlgorithm` that will generate the data for the new regression test algorithm.
- Fixing bug in `UniverseSelection`, it wasn't removing pending to be
removed securities unless the universe selection changed
- Fix bug in live mode custom universe,
`BaseDataCollectionAggregatorEnumerator` returning false was causing the
subscription to be removed completely
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3123
#### Motivation and Context

This new feature will allow users to select a pre defined changing set of Symbols, for example `Banking Stocks`. Allowing an improvement in performance, since the computation required to perform the symbol selection will not be required.

Yes

Unit and regression tests
#### Types of changes

#### Checklist:


Common/Data/UniverseSelection/ConstituentsUniverseData.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <returns>Instance of the T:BaseData object generated by this line of the CSV</returns>
+        public override BaseData Reader(SubscriptionDataConfig config, string line, DateTime date, bool isLiveMode)
+        {
+            try
+            {
+                var csv = line.Split(',');
+                var preselected = new ConstituentsUniverseData
+                {
+                    Symbol = new Symbol(SecurityIdentifier.Parse(csv[0]), csv[1]),
+                    Time = isLiveMode ? date.AddDays(-1) : date
```
mchandschuh: interesting discrepancy between live and backtesting -- I don't see something similar in `CoarseFundamental` which makes me nervous. Is this `date.AddDays(-1)` really highlighting a bug in the data generator algorithm which is saving to the incorrect date?

Common/SecurityIdentifier.cs
```diff
@@ -751,6 +756,7 @@ public override bool Equals(object obj)
         public override string ToString()
         {
             var props = EncodeBase36(_properties);
+            props = props.IsNullOrEmpty() ? "0" : props;
```
mchandschuh: nit - `EncodeBase64` never returns null, faster to just check `props.Length == 0 ? "0" : props` -- also curious as to why this was needed and does it affect anything else? This _will_ affect serialization of `Symbol.Empty`, but not sure if we depend on that serialization anywhere. I can't think of a case where we send it down the pipe via ticker plant or history, etc, so this is likely safe, but wanted to bring attention to it.

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -474,7 +474,7 @@ private Subscription CreateUniverseSubscription(SubscriptionRequest request)
 
                 var factory = new LiveCustomDataSubscriptionEnumeratorFactory(_timeProvider);
                 var enumeratorStack = factory.CreateEnumerator(request, _dataProvider);
-                enumerator = new BaseDataCollectionAggregatorEnumerator(enumeratorStack, config.Symbol);
+                enumerator = new BaseDataCollectionAggregatorEnumerator(enumeratorStack, config.Symbol, liveMode:true);
```
mchandschuh: I wholeheartedly approve of the `liveMode:true` syntax -- it's nice to give names to `bool` parameters 👍

Algorithm.CSharp/ConstituentsUniverseDataGeneratorAlgorithm.cs
```diff
@@ -0,0 +1,152 @@
...
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Algorithm used to generate daily data on disk for the different <see cref="ConstituentsUniverse"/>.
+    /// How to use me? search for 'CHANGE ME'
+    /// 1- change the universe name
+    /// 2- change the desired start and end date to generate
+    /// 3- define the universe selection to use
+    /// Data will be generated in:
+    /// 'Globals.DataFolder\SecurityType\Market\universes\Resolution\{UniverseName}\{dates:yyyyMMdd}.csv'
```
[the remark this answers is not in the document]
Martin-Molinero: Good idea, I'll add a note 👍

Common/Data/UniverseSelection/ConstituentsUniverseData.cs
mchandschuh: Something just seems off about this, but if you're confident that things are consistent that than works for me. Also, this would _not_ produce a bug, as the application of these dates are completely within the silo of universe data being generated -- meaning -- so long as the type that generates the data and the type that reads the data agree on some 'convention' for specifying the time/end time, then we'll get the right result.  The only concern is we might have discrepancies between various 'silos', such as `CoarseFundamental` vs this constitutent data.

Common/SecurityIdentifier.cs
mchandschuh: RE: try parse will fail -- interesting, I would have expected this to require us to update the parsing and not the serialization logic.

mchandschuh:
@Martin-Molinero -- I think this is good to go, but it will need to be rebased before merge.

PR #3171, opened 2019-05-05
Fix infinite loop in `InsightsCollection.CopyTo`

### LINKED ISSUE #3168: JSON serialization of ordered InsightCollection causes stack overflow exception
#### Expected Behavior
Ordering an `InsightCollection` before serializing it works.

#### Actual Behavior
Ordering an `InsightCollection` and subsequently JSON-serializing it produces a stack overflow exception (VS reports it as that, the below is from the Terminal):

```
[ERROR] FATAL UNHANDLED EXCEPTION: at Newtonsoft.Json.JsonSerializer.SerializeInternal (Newtonsoft.Json.JsonWriter, object, System.Type) [0x00253] in :0, at Newtonsoft.Json.JsonSerializer.Serialize (Newtonsoft.Json.JsonWriter, object, System.Type) [0x00000] in :0, at Newtonsoft.Json.JsonConvert.SerializeObjectInternal (object, System.Type, Newtonsoft.Json.JsonSerializer) [0x00028] in :0, at Newtonsoft.Json.JsonConvert.SerializeObject (object, System.Type, Newtonsoft.Json.Formatting, Newtonsoft.Json.JsonSerializerSettings) [0x0000e] in :0, at Newtonsoft.Json.JsonConvert.SerializeObject (object, Newtonsoft.Json.Formatting, Newtonsoft.Json.JsonSerializerSettings) [0x00000] in :0, at Newtonsoft.Json.JsonConvert.SerializeObject (object, Newtonsoft.Json.Formatting) [0x00000] in :0, at QuantConnec
```

#### Potential Solution
Just ordering the object works, so it must be something that happens when during JSON serialization.

#### Reproducing the Problem
_Set `EqualWeightingPortfolioConstructionModel._insightCollection` to `public`._

```c#
using Newtonsoft.Json;

{
    public class ParticleHorizontalProcessor : QCAlgorithm
    {
        {
            SetStartDate(2019, 5, 1);
            SetCash(100000);
            SetPortfolioConstruction(new EqualWeightingPortfolioConstructionModel());
            SetAlpha(new ConstantAlphaModel(InsightType.Price, InsightDirection.Up, TimeSpan.FromDays(1)));
            UniverseSettings.Resolution = Resolution.Hour;
            var symbols = new[] { QuantConnect.Symbol.Create("EURUSD", SecurityType.Forex, Market.Oanda) };
            SetUniverseSelection( new ManualUniverseSelectionModel(symbols) );
        }

        {
            var portConst = PortfolioConstruction as EqualWeightingPortfolioConstructionModel;
            var sorted = portConst._insightCollection.OrderBy( x => x.GeneratedTimeUtc );
            var test = JsonConvert.SerializeObject( sorted );
            Debug(test);
        }
    }
}
```

#### System Information
Happens on both QC terminal and Win x64.

#### Checklist

gsalaz98:
#### Description

Fixes infinite loop in call to `CopyTo` in order to allow list creation and serialization.

This fix may cause a performance degradation.
#### Related Issue

Closes #3168 
#### Motivation and Context

Allows for LINQ queries on `InsightCollection`

No

Unable to be tested; Halting problem prevents testing of stack overflows
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Algorithm/Framework/Alphas/InsightCollection.cs
+++ b/Common/Algorithm/Framework/Alphas/InsightCollection.cs
@@ -104,13 +104,7 @@ namespace QuantConnect.Algorithm.Framework.Alphas
         public void CopyTo(Insight[] array, int arrayIndex)
         {
             // Avoid calling `ToList` on insights to avoid potential infinite loop (issue #3168)
-            var copy = new List<Insight>(Count);
-
-            foreach (var insight in this)
-            {
-                copy.Add(insight);
-            }
-            copy.CopyTo(array, arrayIndex);
+            Array.Copy(_insights.SelectMany(kvp => kvp.Value).ToArray(), 0, array, arrayIndex, Count);
         }
 
         /// <summary>Removes the first occurrence of a specific object from the <see cref="T:System.Collections.Generic.ICollection`1" />.</summary>
--- a/Common/Algorithm/Framework/Alphas/InsightCollection.cs
+++ b/Common/Algorithm/Framework/Alphas/InsightCollection.cs
@@ -103,7 +103,13 @@ namespace QuantConnect.Algorithm.Framework.Alphas
         /// <exception cref="T:System.ArgumentException">The number of elements in the source <see cref="T:System.Collections.Generic.ICollection`1" /> is greater than the available space from <paramref name="arrayIndex" /> to the end of the destination <paramref name="array" />.</exception>
         public void CopyTo(Insight[] array, int arrayIndex)
         {
-            var copy = this.ToList();
+            // Avoid calling `ToList` on insights to avoid potential infinite loop (issue #3168)
+            var copy = new List<Insight>(Count);
+
+            foreach (var insight in this)
+            {
+                copy.Add(insight);
+            }
             copy.CopyTo(array, arrayIndex);
         }
 
--- a/Tests/Algorithm/Framework/Alphas/InsightCollectionTests.cs
+++ b/Tests/Algorithm/Framework/Alphas/InsightCollectionTests.cs
@@ -1,0 +1,113 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Linq;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using QuantConnect.Algorithm.Framework.Alphas;
+
+namespace QuantConnect.Tests.Algorithm.Framework.Alphas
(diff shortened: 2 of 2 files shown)


Common/Algorithm/Framework/Alphas/InsightCollection.cs
```diff
@@ -103,7 +103,13 @@ public bool ContainsKey(Symbol symbol)
...
         public void CopyTo(Insight[] array, int arrayIndex)
         {
-            var copy = this.ToList();
+            // Avoid calling `ToList` on insights to avoid potential infinite loop (issue #3168)
+            var copy = new List<Insight>(Count);
+
+            foreach (var insight in this)
+            {
+                copy.Add(insight);
+            }
```
Martin-Molinero: Good catch!
I think it would be nice if we added a unit test that stacked overflowed before and now passes. Maybe could directly use `AddRange(this)`.

gsalaz98: I attempted to write a unit test for this bug, but unfortunately, didn't end up working out.

Here's the unit test I wrote (or at least what I remember of it):

```csharp
using System;
using System.Linq;
using NUnit.Framework;
using QuantConnect.Algorithm.Framework.Alphas;

namespace QuantConnect.Tests.Algorithm.Framework.Alphas
{
    [TestFixture]
    public class InsightCollectionTests
    {
        [Test]
        public static void InsightCollectionShouldBeAbleToBeConvertedToListWithoutStackOverflow()
        {
            var aapl = Symbol.Create("AAPL", SecurityType.Equity, "usa");
            var insightCollection = new InsightCollection();

            insightCollection.Add(new Insight(aapl, new TimeSpan(1, 0, 0, 0), InsightType.Price, InsightDirection.Up)
            {
                CloseTimeUtc = new DateTime(2019, 1, 1),
            });
            insightCollection.Add(new Insight(aapl, new TimeSpan(1, 0, 0, 0), InsightType.Volatility, InsightDirection.Up)
            {
                CloseTimeUtc = new DateTime(2019, 1, 2),
            });
            
            Assert.DoesNotThrow(() => insightCollection.OrderBy(x => x.GeneratedTimeUtc).ToList());
        }
    }
}
```

Stack overflows aren't able to be caught with traditional patterns (try and catch). I'd like your insight as to how we can test this if possible @Martin-Molinero . I don't know how this test would behave in CI if it fails by SO, so I'm trying to stay clear from that route for the meanwhile.

Martin-Molinero: The unit test looks good to me, stack overflows in `master` and passes in this `PR`.
ACK `AddRange`: right, it shouldn't use `this`! this works: `copy.AddRange(_insights.SelectMany(kvp => kvp.Value));`
Also I checked the `List` implementation and it uses `Array.Copy`, this seems to also work `Array.Copy((Array)list.ToArray(), 0, (Array)array, arrayIndex, list.Count);`

gsalaz98: I believe the method can be simplified to the following:
```csharp
public void CopyTo(Insight[] array, int arrayIndex)
{
    // Avoid calling `ToList` on insights to avoid potential infinite loop (issue #3168)
    Array.Copy(_insights.Values.ToArray(), 0, array, arrayIndex, Count);
}
```

Because `this.ToArray()` also causes a stack overflow, I believe we must source the data from within `_insights` if we are to get rid of the foreach loop.

I'll include the unit tests in the next commit. It may report bad information in CI if the test fails.

One last question: is there a reason to use `SelectMany` over just accessing the `Values` member of `_insights`?

Martin-Molinero: res `SelectMany`: I believe its required because `_insights` is a collection of `List<Insight>`.
I'm thinking we might want to add another unit test to verify the behavior of `CopyTo` is as expected

PR #3175, opened 2019-05-08
Replace IBController with QuantConnect.IBAutomater

StefanoRaggi:
#### Description
The IB brokerage implementation has been updated to use the new QC IBAutomater:
- https://github.com/QuantConnect/IBAutomater
- https://www.nuget.org/packages/QuantConnect.IBAutomater/

Important:
- The IBAutomater only supports IB Gateway v974 or later versions.
- The IB TWS application is not supported.

#### Motivation and Context
The IBController license (GPL v3) is compatible with the LEAN license (Apache 2.0) when used for mere linkage but to avoid debate we decided to implement our own pluggable, nuget distributable package which is more reliable and deterministic. 

The new package aims allow integration and serve as an Apache 2.0 licensed functional equivalent. 

References to IBController in LEAN installation for IB usage will have to be removed.

Live trading with IB accounts both locally and in the cloud.

#### Types of changes

#### Checklist:


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -201,6 +227,22 @@ public InteractiveBrokersBrokerage(IAlgorithm algorithm, IOrderProvider orderPro
...
+
+            // start IB Gateway
+            _ibAutomater = new IBAutomater.IBAutomater(twsDirectory, ibVersion, userName, password, tradingMode, port);
+            _ibAutomater.OutputDataReceived += OnIbAutomaterOutputDataReceived;
+            _ibAutomater.ErrorDataReceived += OnIbAutomaterErrorDataReceived;
+            _ibAutomater.Exited += OnIbAutomaterExited;
+            _ibAutomater.Start(false);
+
+            // wait a few seconds for IB to start up
+            _ibAutomaterInitializeEvent.WaitOne(TimeSpan.FromSeconds(30));
```
Martin-Molinero: Wondering what happens if it timed out and there were no errors, `CheckIbAutomaterErrors()` didn't throw?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1135,10 +1151,10 @@ public void ResetGatewayConnection()
             Disconnect();
 
             Log.Trace("InteractiveBrokersBrokerage.ResetGatewayConnection(): Stopping IB Gateway...");
-            InteractiveBrokersGatewayRunner.Stop();
+            _ibAutomater.Stop();
 
             Log.Trace("InteractiveBrokersBrokerage.ResetGatewayConnection(): Restarting IB Gateway...");
-            InteractiveBrokersGatewayRunner.Restart();
+            _ibAutomater.Start(false);
```
Martin-Molinero: Shouldn't we reset `_ibAutomaterInitializeEvent` before staring and waiting for it to be set? Also maybe call `CheckIbAutomaterErrors();`? Same doubt about it timing out.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: The 30-second wait was previously required for TWS (which had a longer startup time than IB Gateway), now we don't support TWS anymore so this timeout could be shortened.
In this PR I added the initialize event to catch the two handled errors (existing session and security dialog) immediately, without waiting the full 30 seconds. 
Currently, the reset event is only set in those two cases: maybe I could try setting the event upon successful initialization.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Good point! 
`CheckIbAutomaterErrors();` should be called here as well, I'll update and retest this.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
Martin-Molinero: > maybe I could try setting the event upon successful initialization.

That would be awesome. Also I think we should handle somehow the case: that if past the 30 seconds it didn't connect and we didn't detect expected errors?

StefanoRaggi: Actually, the real connection is not done by the IBAutomater (it only inputs credentials for login and changes a few settings in the IB Gateway GUI). The 30 seconds delay has always been there only to give enough time for the TWS/IBG process to be loaded and ready.
The real connection to the IB C# API is done later on in the `Connect` method.

For IBAutomater, other things that could go wrong are unhandled dialogs/popups, which will now be much easier to notice and fix, since now every window event (opened, closed, activated) is now logged in the LEAN syslog.

Issue #3183, opened 2019-05-10
Remove BacktestingTransactionHandler thread


Martin-Molinero:
#### Expected Behavior

- To enable functional python debugging, we want all user code to be executed by the same thread. This requires removing the `BacktestingTransactionHandler` thread, which isn't needed.
#### Actual Behavior

- `BacktestingTransactionHandler.Run()` is executed by a dedicated thread
#### Potential Solution

- Remove `BacktestingTransactionHandler` thread
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3184, opened 2019-05-10
Removing BacktestingTransactionHandler thread

Martin-Molinero:
#### Description

- For backtesting, the algorithm thread will sequentially process pending
order requests
- Updating `UpdateOrderRegressionAlgorithm` that suffered of a race
condition: it supposed `OnOrderEvent` would be called after the call to
`Transactions.AddOrder()` was finished

`Performance RegressionAlgorithm -> 5.4K trades`
`master`
31s
29.8s
26.9s
27.1s
`PR`
29s
29s
26.5s
26.4s

*Results show there isn't a negative impact in performance execution times, the PR is sightly faster.

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3183
#### Motivation and Context

This is required to allow functional debugging in python, specifically adding breakpoints during a debugging session.

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Interfaces/IBusyCollection.cs
+++ b/Common/Interfaces/IBusyCollection.cs
@@ -1,0 +1,76 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace QuantConnect.Interfaces
+{
+    /// <summary>
+    /// Interface used to handle items being processed and communicate busy state
+    /// </summary>
+    /// <typeparam name="T">The item type being processed</typeparam>
+    public interface IBusyCollection<T> : IDisposable
+    {
+        /// <summary>
+        /// Gets a wait handle that can be used to wait until this instance is done
+        /// processing all of it's item
+        /// </summary>
+        WaitHandle WaitHandle { get; }
+
+        /// <summary>
+        /// Gets the number of items held within this collection
+        /// </summary>
+        int Count { get; }
+
+        /// <summary>
+        /// Returns true if processing, false otherwise
+        /// </summary>
+        bool IsBusy { get; }
+
+        /// <summary>
+        /// Adds the items to this collection
+        /// </summary>
+        /// <param name="item">The item to be added</param>
+        void Add(T item);
+
+        /// <summary>
+        /// Adds the items to this collection
+        /// </summary>
+        /// <param name="item">The item to be added</param>
+        /// <param name="cancellationToken">A cancellation token to observer</param>
+        void Add(T item, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Marks the collection as not accepting any more additions
+        /// </summary>
+        void CompleteAdding();
+
+        /// <summary>
+        /// Provides a consuming enumerable for items in this collection.
+        /// </summary>
+        /// <returns>An enumerable that removes and returns items from the collection</returns>
+        IEnumerable<T> GetConsumingEnumerable();
+
+        /// <summary>
+        /// Provides a consuming enumerable for items in this collection.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token to observer</param>
+        /// <returns>An enumerable that removes and returns items from the collection</returns>
+        IEnumerable<T> GetConsumingEnumerable(CancellationToken cancellationToken);
+    }
+}
--- a/Common/Util/BusyCollection.cs
+++ b/Common/Util/BusyCollection.cs
@@ -14,6 +14,7 @@
  *
 */
 
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
@@ -29,6 +30,7 @@ namespace QuantConnect.Util
     {
         private readonly ConcurrentQueue<T> _collection = new ConcurrentQueue<T>();
         private readonly ManualResetEventSlim _processingCompletedEvent = new ManualResetEventSlim(true);
+        private bool _completedAdding;
 
         /// <summary>
         /// Gets a wait handle that can be used to wait until this instance is done
@@ -62,9 +64,19 @@ namespace QuantConnect.Util
         /// <param name="cancellationToken">A cancellation token to observer</param>
         public void Add(T item, CancellationToken cancellationToken)
         {
-            // we're adding work to be done, mark us as busy
-            _processingCompletedEvent.Reset();
-            _collection.Enqueue(item);
+            if (_completedAdding)
+            {
+                throw new InvalidOperationException("Collection has already been marked as not " +
+                    $"accepting more additions, see {nameof(CompleteAdding)}");
+            }
+
+            // locking to avoid race condition with GetConsumingEnumerable()
+            lock (_processingCompletedEvent)
+            {
+                // we're adding work to be done, mark us as busy
+                _processingCompletedEvent.Reset();
+                _collection.Enqueue(item);
+            }
         }
 
         /// <summary>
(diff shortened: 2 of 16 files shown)


Common/Interfaces/IBusyCollection.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
+using System.Threading;
+
+namespace QuantConnect.Interfaces
+{
+    /// <summary>
+    /// Interface used to handle items being processed and communicate busy state
+    /// </summary>
+    /// <typeparam name="T">The item type being processed</typeparam>
+    public interface IBusyCollection<T> : IDisposable
```
mchandschuh: Smart! Let's us test some of these complex/multi-threaded interactions in a _hopefully_ more deterministic way.

Common/Util/BusyCollection.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <param name="cancellationToken">A cancellation token to observer</param>
+        /// <returns>An enumerable that removes and returns items from the collection</returns>
+        public IEnumerable<T> GetConsumingEnumerable(CancellationToken cancellationToken)
+        {
+            T item;
+            while (_collection.TryDequeue(out item))
+            {
+                yield return item;
+            }
+            _processingCompletedEvent.Set();
```
mchandschuh: hmm... this is tough. it's small, but there exists a race condition between [enqueuing](https://github.com/QuantConnect/Lean/pull/3184/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.csproj#diff-ef82eecda24fe8fdf993e6c251ad18f7R67) a new piece of data and resetting the handle here.

Consider we're adding an item, we flip the switch to say we're currently processing, then that thread stops running, another thread processing `GetConsumingEnumerable` had just emptied the `_collection` and now sets the completed event but there's still another item in the queue.

I could see something like this to fix it:
```
// replacing impl of Add(T, CancellationToken)
lock (sync)
{
    // we're adding work to be done, mark us as busy
    _processingCompletedEvent.Reset();
    _collection.Enqueue(item);
}

// replacing line 91
lock (sync)
{
    if (!_collection.TryPeek(out var _))
    {
        _processingCompletedEvent.Set();
    }
}
```

All that being said, the surface area of this potential race is exceedingly slim. This would serialize additions to the collection, but if I recall correctly, we're not pumping massive numbers of items into these collections, also, contention on the lock should usually be non-existent.

Common/Util/BusyCollection.cs
```diff
@@ -0,0 +1,111 @@
...
+        public void Dispose()
+        {
+            _collection.Clear();
+            _processingCompletedEvent.Dispose();
+        }
+
+        /// <summary>
+        /// Marks the collection as not accepting any more additions
+        /// </summary>
+        public void CompleteAdding()
```
mchandschuh: In order to maintain behavior with the other implementation, subsequent calls to `Add` should fail after calling `CompleteAdding`.

See my good ol' friend [Liskov](https://en.wikipedia.org/wiki/Liskov_substitution_principle) for more information.

Engine/TransactionHandlers/BacktestingTransactionHandler.cs
```diff
@@ -67,5 +75,31 @@ public override void ProcessAsynchronousEvents()
             _brokerage.SimulateMarket();
             _brokerage.Scan();
         }
+
+        /// <summary>
+        /// For backtesting we will process the order ourselves
+        /// </summary>
+        /// <param name="ticket">The <see cref="OrderTicket"/> expecting to be processed</param>
+        protected override void WaitForOrderToBeProcess(OrderTicket ticket)
```
mchandschuh: typo? `WaitForOrderToBeProcessed` **ed** suffix for past tense

Engine/TransactionHandlers/BacktestingTransactionHandler.cs
```diff
@@ -67,5 +75,31 @@ public override void ProcessAsynchronousEvents()
...
+        {
+            // we process the order request our selves
+            Run();
+
+            if (!ticket.OrderSet.WaitOne(0))
+            {
+                // this shouldn't happen, but just in case...
```
mchandschuh: perhaps describe what this means? If I'm reading it correctly, it means that the ticket's order property was not set yet. (nice usage of wait handles 👍 )

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -143,7 +145,7 @@ public virtual void Initialize(IAlgorithm algorithm, IBrokerage brokerage, IResu
             {
                 throw new ArgumentNullException("brokerage");
             }
-
+            _orderRequestQueue = new BusyBlockingCollection<OrderRequest>();
```
mchandschuh: maybe a note saying why this impl uses the thread approach -- I'm assuming it's because we want live algorithms to have this run on a separate thread as is the current behavior -- but for the future generations :)

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -267,6 +275,20 @@ public OrderTicket AddOrder(SubmitOrderRequest request)
             return ticket;
         }
 
+        /// <summary>
+        /// For backtesting we will process the order ourselves
+        /// </summary>
+        /// <param name="ticket">The <see cref="OrderTicket"/> expecting to be processed</param>
+        protected virtual void WaitForOrderToBeProcess(OrderTicket ticket)
```
mchandschuh: Perhaps this could be named a little better -- when I hear 'order processed' it makes me think the order has reached a completed state, but the checks below seem to imply that LEAN has done initial processing/submission of order but it might not yet be filled. Perhaps something like `WaitForOrderSubmission` -- (naming is hard!)

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -485,7 +507,7 @@ public List<Order> GetOpenOrders(Func<Order, bool> filter = null)
         /// <summary>
         /// Primary thread entry point to launch the transaction thread.
         /// </summary>
-        public void Run()
+        protected void Run()
```
mchandschuh: This comment is for line 514

VERY COOL! When reading the top I was wondering how it was going to work for the non-threaded version, i.e, how does it break the loop, but duh! it breaks the loop when it's done processing. Gorgeous 👍

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -503,8 +525,11 @@ public void Run()
                 _algorithm.RunTimeError = err;
             }
 
-            Log.Trace("BrokerageTransactionHandler.Run(): Ending Thread...");
-            IsActive = false;
+            if (_processingThread != null)
+            {
+                Log.Trace("BrokerageTransactionHandler.Run(): Ending Thread...");
+                IsActive = false;
```
mchandschuh: Do we keep `IsActive = false` for the non-threaded version? Perhaps the bigger question is what does `IsActive` imply? Does it imply that it can handle requests (contract) or does it imply a thread is running (implementation) -- I try to stick to contractual definitions vs implementation definitions as they're usually more geared towards the desired behavior vs how that behavior was implemented.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -703,6 +728,18 @@ public void Exit()
                 Log.Error("BrokerageTransactionHandler.Exit(): Exceed timeout: " + (int)(timeout.TotalSeconds) + " seconds.");
             }
             _cancellationTokenSource.Cancel();
+            if (_processingThread != null)
+            {
+                if (_processingThread.IsAlive)
+                {
+                    _processingThread.Abort();
```
mchandschuh: Should we set `IsActive = false` after we call `_processingThread.Abort()` ?

Tests/Algorithm/AlgorithmLiveTradingTests.cs
```diff
@@ -48,8 +48,7 @@ public void SetHoldingsTakesIntoAccountPendingMarketOrders()
             var transactionHandler = new BrokerageTransactionHandler();
 
             transactionHandler.Initialize(algorithm, brokerage, new LiveTradingResultHandler());
-            new Thread(transactionHandler.Run) { IsBackground = true }.Start();
-            Thread.Sleep(2000);
+            Thread.Sleep(250);
```
mchandschuh: hmm, `Thread.Sleep(x)` always has a certain [smell](https://en.wikipedia.org/wiki/Code_smell) to me. I don't think this is important to tackle in this PR, but more for just my philosophy and how one could change it -- We could define an `WaitHandle ITransactionHandler.Initialized` and then just wait for init to complete.

Consider someone setting up a SQL database or maybe they need to fetch remote files in the initializer -- yes, esoteric examples, and simply food for throught

Tests/Common/Securities/CashBuyingPowerModelTests.cs
```diff
@@ -61,7 +61,6 @@ public void Initialize()
             _transactionHandler = new BacktestingTransactionHandler();
             _brokerage = new BacktestingBrokerage(_algorithm);
             _transactionHandler.Initialize(_algorithm, _brokerage, new TestResultHandler());
-            new Thread(_transactionHandler.Run) { IsBackground = true }.Start();
```
mchandschuh: love it!

PR #3194, opened 2019-05-14
Add template algorithms

simonsonjack:
Two demonstration algorithms showing how to incorporate Python into a C# algorithm and vice versa

#### Description

Outputs two debug statements that print the value of sin(10) in both languages.

#### Motivation and Context

A user asked about this in the forum and it seemed like it might interest more people. Publishing it here would give people the skeleton code to work with.

Run locally and in the terminal, verified matching results.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/BasicPythonIntegrationTemplateAlgorithm.cs
+++ b/Algorithm.CSharp/BasicPythonIntegrationTemplateAlgorithm.cs
@@ -1,0 +1,56 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using QuantConnect.Data;
+using Python.Runtime;
+namespace QuantConnect.Algorithm.CSharp
+{
+    public class BasicPythonIntegrationTemplateAlgorithm : QCAlgorithm
+    {
+
+        public override void Initialize()
+        {
+            SetStartDate(2018, 11, 10);  //Set Start Date
+            SetEndDate(2018, 12, 1);
+            SetCash(100000);             //Set Strategy Cash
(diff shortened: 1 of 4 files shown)


Algorithm.CSharp/BasicPythonIntegrationTemplateAlgorithm.cs
```diff
@@ -0,0 +1,56 @@
...
+            SetCash(100000);             //Set Strategy Cash
+            AddEquity("SPY", Resolution.Minute);
+
+        }
+
+        private decimal ComputeSin(decimal value)
+        {
+            using (Py.GIL())
+            {
+                dynamic np = Py.Import("numpy");
```
AlexCatarino: I would suggest that we create a class field for numpy:
```csharp
private dynamic _numpy
```
assign it in Initialize:
```csharp
{
    // other initilization statements
    using (Py.GIL())
    {
        _numpy = Py.Import("numpy");
    }
}
```
and finally use it in the method:
```csharp
private decimal ComputeSin(decimal value)
{
    using (Py.GIL())
    {
        return (decimal)_numpy.sin(value);
    }
}
```
to avoid provinding an example where we try to import the same Python multiple times.

Algorithm.CSharp/BasicPythonIntegrationTemplateAlgorithm.cs
```diff
@@ -0,0 +1,56 @@
...
+using Python.Runtime;
+namespace QuantConnect.Algorithm.CSharp
+{
+    public class BasicPythonIntegrationTemplateAlgorithm : QCAlgorithm
+    {
+
+        public override void Initialize()
+        {
+            SetStartDate(2018, 11, 10);  //Set Start Date
+            SetEndDate(2018, 12, 1);
```
AlexCatarino: For consistency, we should keep the same date range in both algorithms.

Algorithm.Python/BasicCSharpIntegrationTemplateAlgorithm.py
```diff
@@ -0,0 +1,44 @@
...
+class BasicCSharpIntegrationTemplateAlgorithm(QCAlgorithm):
+
+    def Initialize(self):
+        self.SetStartDate(2013,10, 7)  #Set Start Date
+        self.SetEndDate(2013,10,11)    #Set End Date
+        self.SetCash(100000)           #Set Strategy Cash
+        
+        self.AddEquity("SPY", Resolution.Second)
+        self.Debug(f'According to Python, the value of sin(10) is {np.sin(10)}')
+        self.Debug(f'According to C#, the value of sin(10) is {Math.Sin(10)}')
```
AlexCatarino: In the C# version, these `Debug` statements appear in `OnData` method.

Algorithm.CSharp/BasicPythonIntegrationTemplateAlgorithm.cs
simonsonjack: I made the change in the python algo, they'll be the same now

Issue #3074, opened 2019-04-11
Warm Up Indicator Helper


AlexCatarino:
#### Expected Behavior
Be able to easily warm up indicators using QCAlgorithm API.

#### Actual Behavior
In order to warm up indicators, users need to make a history request call and use them to update the indicators manually. 

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

PR #3195, opened 2019-05-15
Extends NormalizedAverageTrueRange and TrueRange Indicators With IIndicatorWarmUpPeriodProvider

AlexCatarino:
#### Description
Extends `NormalizedAverageTrueRange` and `TrueRange` indicators with `IIndicatorWarmUpPeriodProvider`.

#### Related Issue
#3074 

#### Motivation and Context
See #3074 

Unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/NormalizedAverageTrueRange.cs
+++ b/Indicators/NormalizedAverageTrueRange.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


Indicators/NormalizedAverageTrueRange.cs
```diff
@@ -47,17 +47,19 @@ public class NormalizedAverageTrueRange : BarIndicator
...
-        public override bool IsReady
-        {
-            get { return Samples > _period; }
-        }
+        public override bool IsReady => Samples > _period;
+
+        /// <summary>
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public int WarmUpPeriod => _period + 1;
```
mchandschuh: I didn't think of this on the previous PR, but I think these definitions might be better phrased against the dependency, for example, here we could say `=> _smoother.WarmUpPeriod` -- and going even further, could make `SimpleMovingAverage.WarmUpPeriod => RollingSum.WarmUpPeriod` (which would require `RollingWindow<T> : IWarmUpPeriodProvider`

---

#### History of +1 warm up periods

My reasoning here comes down to this infamous `+ 1` that often appears and IIRC, it was a _very_ [early](b3de52091) decision made due to `RollingWindow<T>.MostRecentlyRemoved` throwing an exception if we haven't removed any items yet. Thinking back on that decision now it seems silly and could potentially lead to variations in results vs different platforms, especially in the indicators that use a different formula for the warmup period (such as, technically, wilders).

I'm not recommending a change here, but simply wanted to share some knowledge on why this is like this (SMA10 requires 11 points to be `IsReady==true`).

@jaredbroad @StefanoRaggi @Martin-Molinero - tagged for knowledge transfer :) feel free to ping me to discuss further

AlexCatarino: When we have an indicator that inherits from `WindowIndicator<IndicatorDataPoint>`, like SMA, we get the following `WarmUpPeriod`:
```csharp
public int WarmUpPeriod => Period;
```
So, we would have:
```csharp
public class RollingWindow<T> : IReadOnlyWindow<T>, IIndicatorWarmUpPeriodProvider
{
    public int WarmUpPeriod => Size;
}
```
and
```csharp
public abstract class WindowIndicator<T> : IndicatorBase<T>, IIndicatorWarmUpPeriodProvider
{
    public int WarmUpPeriod => Period;   // since public int Period => _window.Size;
}
```

However, this particular case is different, because IsReady is defined as:
```csharp
public override bool IsReady => Samples > _period;
```
where `IsReady` in `WindowIndicator`/`RollingWindow` is a greater **or equal** condition.

Issue #3156, opened 2019-05-01
Infinitely Polling Custom Data URL When Return Null


jaredbroad:
#### Expected Behavior
Should ping custom data URL's on the resolution step with max 30min, and when no data wait for the next step to poll again.

#### Actual Behavior
Returning null triggered immediate re-polling, causing LEAN to hit rate limit warnings for custom data APIs. 

#### Potential Solution
Review custom data enumerator stack; perhaps there should be another return option:
 `null` - immediately retry?  `BaseData.None` - wait for next resolution step?  

#### Reproducing the Problem
Return null from USEnergyInformation.cs instead of new BaseDataCollection. 

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

erikbengtson:
Hi, In TextSubscriptionDataSourceReader.Read I added the following the check, it fixes my issue. That means, if the current BaseData implementation does not return a value, it skips and goes to the next enumerator. does this fix makes sense? at least I've been using this fix for a while. 


`                        if(instance == null)
                        {
                            yield break;
                        }

                        if (instance != null && instance.EndTime != default(DateTime))
                        {`

PR #3218, opened 2019-05-21
Fix infinite live rest custom data polling

Martin-Molinero:
#### Description

- Adding `IStreamReader.RateLimit { get; }` specifying if calls to
`ReadLine()` should be rate limited by the source readers.
- Adding three failling unit tests in `master`
    - Live rest custom data: returning null, throwing an exception, returning old data will not infinitely poll

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3156
#### Motivation and Context

Fix infinite live custom data rest polling

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Interfaces/IStreamReader.cs
+++ b/Common/Interfaces/IStreamReader.cs
@@ -41,6 +41,6 @@ namespace QuantConnect.Interfaces
         /// <summary>
         /// Gets whether or not this stream reader should be rate limited
         /// </summary>
-        bool RateLimit { get; }
+        bool ShouldBeRateLimited { get; }
     }
 }
--- a/Common/Interfaces/IStreamReader.cs
+++ b/Common/Interfaces/IStreamReader.cs
@@ -1,11 +1,11 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


Common/Interfaces/IStreamReader.cs
```diff
@@ -34,8 +34,13 @@ public interface IStreamReader : IDisposable
...
         /// <summary>
-        /// Gets the next line/batch of content from the stream 
+        /// Gets the next line/batch of content from the stream
         /// </summary>
         string ReadLine();
+
+        /// <summary>
+        /// Gets whether or not this stream reader should be rate limited
+        /// </summary>
+        bool RateLimit { get; }
```
StefanoRaggi: Just my personal preference, this property name sounds better as `ShouldBeRateLimited`

Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
```diff
@@ -115,7 +115,10 @@ public IEnumerable<BaseData> Read(SubscriptionDataSource source)
                     catch (Exception err)
                     {
                         OnReaderError(raw, err);
-                        continue;
+                        if (!reader.RateLimit)
+                        {
+                            continue;
+                        }
```
StefanoRaggi: The current implementations of `RateLimit` only return `true` in live mode, so L130 would never be reached with `instances == null`. Currently not a bug, but just to be proactive, maybe better add a null check on `instances` before L130?

Tests/Engine/DataFeeds/LiveTradingDataFeedTests.cs
```diff
@@ -485,6 +488,63 @@ public void HandlesManyCustomDataSubscriptions()
             Assert.IsTrue(emittedData);
         }
 
+        [TestCase(FileFormat.Csv)]
+        [TestCase(FileFormat.Collection)]
+        public void RestCustomDataReturningNullDoNotInfinitelyPoll(FileFormat fileFormat)
```
StefanoRaggi: A small typo: `DoNot` -> `DoesNot`

Tests/Engine/DataFeeds/LiveTradingDataFeedTests.cs
```diff
@@ -967,4 +1027,65 @@ protected override ITimeProvider GetTimeProvider()
             return _timeProvider;
         }
     }
+
+    internal class TestCustomData : BaseData
```
StefanoRaggi: Nice tests! 👍

Common/Interfaces/IStreamReader.cs
Martin-Molinero: Your right 😄 I'll rename it

Tests/Engine/DataFeeds/LiveTradingDataFeedTests.cs
Martin-Molinero: Thanks 👍

Engine/DataFeeds/CollectionSubscriptionDataSourceReader.cs
Martin-Molinero: Okay, I'll add the null check just in case 👍

PR #3245, opened 2019-05-29
Implement CoinApiSymbolMapper

### LINKED ISSUE #3244: CoinApi bitfinex symbols needs to be mapped
#### Expected Behavior

Data files generated with `ToolBox.CoinApiDataConverter` should have consistent symbols with `BitfinexBrokerage`

#### Actual Behavior

`ToolBox.CoinApiDataConverter` generate the symbols using CoinApi symbology. 

#### Potential Solution

Implement an instance of `ISymbolMapper` for CoinApi Bitfinex

#### Reproducing the Problem

Use `ToolBox.CoinApiDataConverter` with crypto like Dash, Iota, Tehther, etc.

#### System Information

N/A

#### Checklist

Jay-Jay-D:
#### Description

Implement an instance of `ISymbolMapper` for CoinApi Bitfinex

#### Related Issue

Closes #3244 

#### Motivation and Context

Data files generated with ToolBox.CoinApiDataConverter should have consistent symbols with BitfinexBrokerage.

No.

Added simple tests.

#### Types of changes

#### Checklist:


Tests/QuantConnect.Tests.csproj
```diff
@@ -480,6 +480,7 @@
     <Compile Include="Symbols.cs" />
     <Compile Include="TestExtensions.cs" />
     <Compile Include="Indicators\LinearWeightedMovingAverageTests.cs" />
+    <Compile Include="ToolBox\CoinApiBitfinexSymbolMapperTest.cs" />
```
jaredbroad: Please call the class `CoinApiSymbolMapper`

ToolBox/CoinApiDataConverter/CoinApiBitfinexSymbolMapper.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <param name="market">The market</param>
+        /// <param name="expirationDate">Expiration date of the security(if applicable)</param>
+        /// <param name="strike">The strike of the security (if applicable)</param>
+        /// <param name="optionRight">The option right of the security (if applicable)</param>
+        /// <returns>
+        /// A new Lean Symbol instance
+        /// </returns>
+        public Symbol GetLeanSymbol(string brokerageSymbol, SecurityType securityType, string market, DateTime expirationDate = default(DateTime), decimal strike = 0, OptionRight optionRight = OptionRight.Call)
+        {
+            var pair = brokerageSymbol.Split('_');
```
jaredbroad: Sanity checks / error checks for if pair.length != 2 etc

jaredbroad: The input should be the symbol_id from coinApi:
"symbol_id": "DERIBIT_OPT_BTC_USD_190329_10000_C",

https://docs.coinapi.io/#list-all-symbols

ToolBox/CoinApiDataConverter/CoinApiBitfinexSymbolMapper.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// </summary>
+        /// <param name="symbol">A Lean symbol instance</param>
+        /// <returns>
+        /// The brokerage symbol
+        /// </returns>
+        public string GetBrokerageSymbol(Symbol symbol)
+        {
+            var leanSymbol = symbol.Value;
+
+            var pair = new[] { leanSymbol.Substring(0, 3), leanSymbol.Substring(3, 3) };
```
Martin-Molinero: Maybe aligned with the changes of the PR https://github.com/QuantConnect/Lean/pull/3000 we could use the `SymbolPropertiesDatabase` to get the quote and base currency? @jaredbroad what do you guys think?

ToolBox/CoinApiDataConverter/CoinApiBitfinexSymbolMapper.cs
```diff
@@ -0,0 +1,99 @@
...
+
+namespace QuantConnect.ToolBox.CoinApiDataConverter
+{
+    /// <summary>
+    /// Mapper of CoinApi Bitfinex symbols with more than 3 characters.
+    /// </summary>
+    /// <seealso cref="QuantConnect.Brokerages.ISymbolMapper" />
+    public class CoinApiBitfinexSymbolMapper : ISymbolMapper
+    {
+        private HashSet<Tuple<string, string>> _mappedSymbols = new HashSet<Tuple<string, string>>
```
Martin-Molinero: I believe this collection could be `static`?
Maybe to obtain O(1) lookup for both mapping operations, we could use two dictionaries, one of which is lazy initialized as an inverted version of the other, thoughts?
```
new Lazy<Dictionary<string, string>>(() => _mappedSymbols.ToDictionary(kp => kp.Value, kp => kp.Key));
```

ToolBox/CoinApiDataConverter/CoinApiBitfinexSymbolMapper.cs
StefanoRaggi: Since here we don't need all symbol properties (tick sizes, tick increments, etc.) but only the ticker mapping, I think it's easier to maintain with a simple REST API call.
I just pushed an update for this.

ToolBox/CoinApi/CoinApiSymbolMapper.cs
```diff
@@ -0,0 +1,178 @@
...
+            { Market.Bitfinex, "BITFINEX" }
+        };
+        private static readonly Dictionary<string, string> MapExchangeIdsToMarkets =
+            MapMarketsToExchangeIds.ToDictionary(x => x.Value, x => x.Key);
+
+        private static readonly Dictionary<string, Dictionary<string, string>> CoinApiToLeanCurrencyMappings =
+            new Dictionary<string, Dictionary<string, string>>
+            {
+                {
+                    Market.Bitfinex,
```
Martin-Molinero: The dictionary is missing `Market.GDAX`, wondering if the mapping is straight, based on the `ConvertCoinApiCurrencyToLeanCurrency` implementation? Maybe we could add some unit test for `Market.GDAX` too.
Or if it is yet to be implemented, we can throw some error?

StefanoRaggi: The GDAX entry is not present in the dictionary because there is no mapping required for currencies (base and quote currency symbols for CoinAPI and GDAX are the same)

ToolBox/CoinApi/CoinApiSymbolMapper.cs
```diff
@@ -0,0 +1,178 @@
...
+            return exchangeId;
+        }
+
+        private void LoadSymbolMap(string[] exchangeIds)
+        {
+            var list = string.Join(",", exchangeIds);
+
+            using (var wc = new WebClient())
+            {
+                var url = $"{RestUrl}/v1/symbols?filter_symbol_id={list}&apiKey={_apiKey}";
```
jaredbroad: For future design consideration, this would have been better without the external dependencies. This causes rate limits when applied massively in parallel. 

The pattern for this problem in LEAN is the `HashSet<string> KnownSymbolStrings = new HashSet<string>`. The pattern isn't perfect, but it also doesn't fail if you run it 100x =)

PR #3250, opened 2019-05-30
Add IDataQueueHandler implementation for CoinAPI

### LINKED ISSUE #3249: Add CoinApiDataQueueHandler implementation
#### Expected Behavior
New feature.

#### Actual Behavior
Not implemented.

#### Checklist

StefanoRaggi:
#### Description
- Added new `IDataQueueHandler` implementation for CoinAPI: `CoinApiDataQueueHandler`
- Added missing `OnClose` event handler in base `WebSocketWrapper` class
- Added new `IConnectionHandler` interface with `DefaultConnectionHandler` implementation

**Note**: this PR depends on #3245 (for `CoinApiSymbolMapper`)

#### Related Issue
Closes #3249 

#### Motivation and Context
New data feed for Crypto.

No.

Local algorithm testing + full market feed testing for CoinBase and Bitfinex.

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/CoinApi/CoinApiDataQueueHandler.cs
+++ b/ToolBox/CoinApi/CoinApiDataQueueHandler.cs
@@ -16,7 +16,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -56,6 +55,8 @@ namespace QuantConnect.ToolBox.CoinApi
         private DateTime _nextHelloMessageUtcTime = DateTime.MinValue;
 
         private List<string> _subscribedExchanges = new List<string>();
+
+        private readonly Dictionary<Symbol, Tick> _previousQuotes = new Dictionary<Symbol, Tick>();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CoinApiDataQueueHandler"/> class
@@ -307,7 +308,7 @@ namespace QuantConnect.ToolBox.CoinApi
 
                             lock (_locker)
                             {
-                                _ticks.Add(new Tick
+                                var tick = new Tick
                                 {
                                     Symbol = _symbolMapper.GetLeanSymbol(quote.SymbolId, SecurityType.Crypto, string.Empty),
                                     Time = quote.TimeExchange,
@@ -316,7 +317,17 @@ namespace QuantConnect.ToolBox.CoinApi
                                     BidPrice = quote.BidPrice,
                                     BidSize = quote.BidSize,
                                     TickType = TickType.Quote
-                                });
+                                };
+
+                                // only emit quote ticks if bid price or ask price changed
+                                Tick previousQuote;
+                                if (!_previousQuotes.TryGetValue(tick.Symbol, out previousQuote) ||
+                                    tick.AskPrice != previousQuote.AskPrice ||
+                                    tick.BidPrice != previousQuote.BidPrice)
+                                {
+                                    _previousQuotes[tick.Symbol] = tick;
+                                    _ticks.Add(tick);
+                                }
                             }
 
                             _connectionHandler.KeepAlive(quote.TimeExchange);
--- a/ToolBox/CoinApi/CoinApiDataQueueHandler.cs
+++ b/ToolBox/CoinApi/CoinApiDataQueueHandler.cs
@@ -337,7 +337,15 @@ namespace QuantConnect.ToolBox.CoinApi
 
         private void OnReconnectRequested(object sender, EventArgs e)
         {
-            _webSocket.Connect();
+            if (!_webSocket.IsOpen)
+            {
+                _webSocket.Connect();
+            }
+
+            if (!_webSocket.IsOpen)
+            {
+                return;
+            }
 
             if (_subscribedExchanges.Count > 0)
             {
--- a/ToolBox/CoinApi/CoinApiDataQueueHandler.cs
+++ b/ToolBox/CoinApi/CoinApiDataQueueHandler.cs
@@ -52,6 +52,9 @@ namespace QuantConnect.ToolBox.CoinApi
         private DateTime _lastSubscribeRequestUtcTime = DateTime.MinValue;
         private bool _subscriptionsPending;
 
+        private readonly TimeSpan _minimumTimeBetweenHelloMessages = TimeSpan.FromSeconds(5);
+        private DateTime _nextHelloMessageUtcTime = DateTime.MinValue;
+
         private List<string> _subscribedExchanges = new List<string>();
 
         /// <summary>
@@ -104,7 +107,7 @@ namespace QuantConnect.ToolBox.CoinApi
                 Log.Trace($"CoinApiDataQueueHandler.Subscribe(): {string.Join(",", symbolsToSubscribe.Select(x => x.Value))}");
 
                 // CoinAPI requires at least 5 seconds between subscription requests so we need to batch them
-                _subscribedSymbols = symbolsToSubscribe.Union(_subscribedSymbols).ToHashSet();
+                _subscribedSymbols = symbolsToSubscribe.Concat(_subscribedSymbols).ToHashSet();
 
                 ProcessSubscriptionRequest();
             }
@@ -153,21 +156,11 @@ namespace QuantConnect.ToolBox.CoinApi
         /// <param name="markets">List of LEAN markets (exchanges) to subscribe</param>
         public void SubscribeMarkets(List<string> markets)
         {
-            if (markets.Count == 0) return;
-
             Log.Trace($"CoinApiDataQueueHandler.SubscribeMarkets(): {string.Join(",", markets)}");
 
             _subscribedExchanges = markets.ToList();
 
-            var message = JsonConvert.SerializeObject(new HelloMessage
-            {
-                ApiKey = _apiKey,
-                Heartbeat = true,
-                SubscribeDataType = new[] { "trade", "quote" },
-                SubscribeFilterSymbolId = markets.Select(x => _symbolMapper.GetExchangeId(x)).ToArray()
-            });
-
-            _webSocket.Send(message);
+            SendHelloMessage(markets.Select(x => _symbolMapper.GetExchangeId(x)));
 
             _connectionHandler.EnableMonitoring(true);
         }
@@ -179,7 +172,7 @@ namespace QuantConnect.ToolBox.CoinApi
             _lastSubscribeRequestUtcTime = DateTime.UtcNow;
             _subscriptionsPending = true;
 
-            Task.Run(() =>
+            Task.Run(async () =>
             {
                 while (true)
                 {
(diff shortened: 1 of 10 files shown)


ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -0,0 +1,360 @@
...
+                {
+                    DateTime requestTime;
+                    List<Symbol> symbolsToSubscribe;
+                    lock (_lockerSubscriptions)
+                    {
+                        requestTime = _lastSubscribeRequestUtcTime.Add(_subscribeDelay);
+                        symbolsToSubscribe = _subscribedSymbols.ToList();
+                    }
+
+                    if (DateTime.UtcNow > requestTime)
```
Martin-Molinero: Seems there will always be a `_subscribeDelay` delay for the first subscriptions in the task, due to lines 179 + 190? Maybe it could set `_lastSubscribeRequestUtcTime` only _after_ it finishes each subscription request, line 201, and initialized at `MinValue` (& removing line 179)?

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -0,0 +1,360 @@
...
+                            if (_subscribedSymbols.Count == symbolsToSubscribe.Count)
+                            {
+                                // no more subscriptions pending, task finished
+                                _subscriptionsPending = false;
+                                break;
+                            }
+                        }
+                    }
+
+                    Thread.Sleep(200);
```
Martin-Molinero: I think we might be able to sleep the exact amount we need, so `DateTime.UtcNow > requestTime` is true?

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -0,0 +1,360 @@
...
+            });
+
+            _webSocket.Send(message);
+
+            _connectionHandler.EnableMonitoring(true);
+        }
+
+        private void OnMessage(object sender, WebSocketMessage e)
+        {
+            //Log.Trace(e.Message);
```
Martin-Molinero: This probably was used while testing 😄, should be removed?

ToolBox/CoinApi/DefaultConnectionHandler.cs
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        /// <summary>
+        /// Notifies the connection handler that new data was received
+        /// </summary>
+        public void KeepAlive()
+        {
+            lock (_lockerConnectionMonitor)
+            {
+                _lastDataReceivedTime = DateTime.UtcNow;
```
Martin-Molinero: Being picky about performance here: doing `DateTime.UtcNow` on each new data point might add a significant overhead. Idea: what if we set a flag/event here and have a timer or maybe the `_connectionMonitorThread` check on it every `_maximumIdleTimeSpan`: if not set-> connection lost, else clear flag

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -0,0 +1,360 @@
...
+            {
+                var symbolsToUnsubscribe = (from symbol in symbols
+                                            where _subscribedSymbols.Contains(symbol)
+                                            select symbol).ToList();
+                if (symbolsToUnsubscribe.Count == 0)
+                    return;
+
+                Log.Trace($"CoinApiDataQueueHandler.Unsubscribe(): {string.Join(",", symbolsToUnsubscribe.Select(x => x.Value))}");
+
+                // CoinAPI requires at least 5 seconds between subscription requests so we need to batch them
```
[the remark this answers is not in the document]
StefanoRaggi: Yeah, the comment is confusing 😃.
I'll try to explain the issue better:
CoinAPI web sockets API only have one Hello message (where we need to pass all the symbols/markets to be subscribed): https://docs.coinapi.io/#hello

I quickly found out that if you send multiple Hello message quickly, they will simply stop sending data with no error.
I was able to get this message only a couple of times and this is where my 5-second comment comes from:
```{"message":"Between resubscriptions (hello messages) must pass minimum 5 seconds.","type":"error"}```

So we cannot send a Hello message for each Subscribe request but we must group all the symbols together and submit a single Hello message for all symbols.
This is where the 250ms `_subscribeDelay` comes into play: when we don't receive a new subscribe message for 250ms, we assume that the list of symbols is complete and we can submit the request.

A similar logic was implemented for subscribe requests in the Oanda brokerage, although for a different reason.

File: ToolBox/CoinApi/CoinApiDataQueueHandler.cs (hunk 1 above)
Comment (StefanoRaggi) [reply ToolBox/CoinApi/CoinApiDataQueueHandler.cs]: The delay is intentional and expected as explained in the previous comment.

File: ToolBox/CoinApi/CoinApiDataQueueHandler.cs (hunk 2 above)
Comment (StefanoRaggi) [reply ToolBox/CoinApi/CoinApiDataQueueHandler.cs]: The 200ms is not related to the 250ms delay mentioned above, this value was chosen only to minimize CPU usage in this loop.

File: ToolBox/CoinApi/CoinApiDataQueueHandler.cs (hunk 3 above)
Comment (StefanoRaggi) [reply ToolBox/CoinApi/CoinApiDataQueueHandler.cs]: Sure, will remove 👍

File: ToolBox/CoinApi/DefaultConnectionHandler.cs (hunk 4 above)
Comment (StefanoRaggi) [reply ToolBox/CoinApi/DefaultConnectionHandler.cs]: Good point! 
We could also remove the `DateTime.UtcNow` call by passing in the UTC timestamp of the new tick received, thoughts?

File: ToolBox/CoinApi/CoinApiDataQueueHandler.cs (hunk 5 above)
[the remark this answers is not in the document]
Comment (Martin-Molinero) [reply ToolBox/CoinApi/CoinApiDataQueueHandler.cs]: Cool, I get it now 👍 .
The only comment is that after the initial batch of subscriptions are sent, I believe the implementation is not guaranteeing/waiting the 5 seconds if a new subscription request is sent after `_subscribeDelay`, right? Maybe it could keep the state somehow

File: ToolBox/CoinApi/DefaultConnectionHandler.cs (hunk 4 above)
Comment (Martin-Molinero) [reply ToolBox/CoinApi/DefaultConnectionHandler.cs]: Yes, I think that should do the trick too! 👍 it wouldn't require a timezone conversion right? Those are terribly expensive 😄

Comment (StefanoRaggi) [reply ToolBox/CoinApi/DefaultConnectionHandler.cs]: Right, raw data timestamps are UTC for CoinAPI -- should be the same for all other live data feeds 😄

File: ToolBox/CoinApi/CoinApiDataQueueHandler.cs (hunk 5 above)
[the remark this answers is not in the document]
Comment (StefanoRaggi) [reply ToolBox/CoinApi/CoinApiDataQueueHandler.cs]: Seems a bit of an edge case, like an algorithm playing with `AddSecurity` and `RemoveSecurity` at Tick or Second resolutions, but should be handled anyway. I'll track the time of the last Hello message and guarantee at least 5 seconds between requests.

File: ToolBox/CoinApi/CoinApiDataQueueHandler.cs [hunk 6]
```diff
@@ -0,0 +1,360 @@
...
+                var symbolsToSubscribe = (from symbol in symbols
+                                          where !_subscribedSymbols.Contains(symbol) && CanSubscribe(symbol)
+                                          select symbol).ToList();
+                if (symbolsToSubscribe.Count == 0)
+                    return;
+
+                Log.Trace($"CoinApiDataQueueHandler.Subscribe(): {string.Join(",", symbolsToSubscribe.Select(x => x.Value))}");
+
+                // CoinAPI requires at least 5 seconds between subscription requests so we need to batch them
+                _subscribedSymbols = symbolsToSubscribe.Union(_subscribedSymbols).ToHashSet();
```
mchandschuh: nit - no need to change - could use `Concat` here since `ToHashSet` will perform the set union as well.

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
mchandschuh: In general, we shouldn't be calling `Thread.Sleep` within a task object as it will block a thread pool thread. You can use the task constructs, such as `await Task.Delay(200).ConfigureAwait(false)` instead (will need to add `async` before the lambda.

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -0,0 +1,360 @@
...
+
+            _connectionHandler.EnableMonitoring(true);
+        }
+
+        private void ProcessSubscriptionRequest()
+        {
+            if (_subscriptionsPending) return;
+
+            _lastSubscribeRequestUtcTime = DateTime.UtcNow;
+            _subscriptionsPending = true;
```
mchandschuh: At first this looks like a potential race condition, but this entire method is covered by `lock(_lockerSubsriptions` via the `Subscribe`/`Unsubscribe` methods.

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
mchandschuh: This loop is intended to only run until we've processed all the subscriptions, correct? Also, the only reason we _wouldn't_ be able to process all of the subscriptionsis due to the 5 second change via `DateTime.UtcNow > requestTime` -- or perhaps I'm missing how this is intended to operate. If the purpose of the loop is to continually check until the required 5 seconds has passed, then I agree w/ Martin that we could sleep (or delay, see other comment) until the required amount of time has passed instead of napping for ~200ms only to wake up and take another nap because of the the request time checks.

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -0,0 +1,360 @@
...
+                                    TickType = TickType.Quote
+                                });
+                            }
+
+                            _connectionHandler.KeepAlive();
+                            break;
+                        }
+
+                    // not a typo :)
+                    case "hearbeat":
```
mchandschuh: haha, really! nice comment, but that's super strange that their api sends it back as `hearbeat` -- maybe we can add `heartbeat` as well just in case they decide to fix it:

``` csharp
case "hearbeat":
case "heartbeat":
```

ToolBox/CoinApi/DefaultConnectionHandler.cs
```diff
@@ -0,0 +1,196 @@
...
+
+            _connectionMonitorThread.Start();
+
+            waitHandle.WaitOne();
+        }
+
+        /// <summary>
+        /// Enables/disables monitoring of the connection
+        /// </summary>
+        public void EnableMonitoring(bool isEnabled)
```
mchandschuh: Function is called `EnableMonitoring` but it looks like it's intended to enable _or_ disable monitoring, however, if monitoring is already enabled then we won't disable monitor (`_isEnabled == true && isEnabled == false` seems like a broken case)?

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
StefanoRaggi: The purpose of this task is to group subscription requests together because we can't send Hello messages before 5 seconds have passed since the previous message. We don't want to wait 5 seconds after the last subscription request is received so we submit the request if at least 250ms has elapsed since the last request.

Note: the check for 5-second minimum elapsed time between Hello messages is not implemented yet. 
I'm working on it now and will also change the logic to wait the correct amount of time.

PR #3262, opened 2019-06-03
Coin api duplicated symbols fix

Jay-Jay-D:
#### Description

This PR solves an issue with CoinApi symbols for Bitfinex, it turns out that raw data folders of older dates have duplicated data for symbols with four letters e.g. `DASH`.

Also changes the logic of `CoinApiSymbolMapper`, now making request to `CoinApi` API is the default behavior, but it can be configured to read a local file. This is useful for large paralellized bat processes.  

#### Related Issue

N/A.

#### Motivation and Context

The duplicated symbol issue generates error when two threads tries to write the same file.

The `CoinApiSymbolMapper` issue hit request limits when processing processing CoinApi data in parallel.

No.

Local run.

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/CoinApi/CoinApiSymbolMapper.cs
+++ b/ToolBox/CoinApi/CoinApiSymbolMapper.cs
@@ -15,11 +15,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net;
 using Newtonsoft.Json;
 using QuantConnect.Brokerages;
 using QuantConnect.Configuration;
+using QuantConnect.Logging;
 
 namespace QuantConnect.ToolBox.CoinApi
 {
@@ -31,7 +33,8 @@ namespace QuantConnect.ToolBox.CoinApi
     {
         private const string RestUrl = "https://rest.coinapi.io";
         private readonly string _apiKey = Config.Get("coinapi-api-key");
-
+        private readonly bool _useLocalSymbolList = Config.GetBool("coinapi-use-local-symbol-list", false);
+        private readonly FileInfo _coinApiSymbolsListFile = new FileInfo("CoinApiSymbols.json");
         // LEAN market <-> CoinAPI exchange id maps
         private static readonly Dictionary<string, string> MapMarketsToExchangeIds = new Dictionary<string, string>
         {
@@ -63,6 +66,7 @@ namespace QuantConnect.ToolBox.CoinApi
 
         // map LEAN symbols to CoinAPI symbol ids
         private Dictionary<Symbol, string> _symbolMap = new Dictionary<Symbol, string>();
+
 
         /// <summary>
         /// Creates a new instance of the <see cref="CoinApiSymbolMapper"/> class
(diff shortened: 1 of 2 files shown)


ToolBox/CoinApi/CoinApiSymbolMapper.cs
```diff
@@ -31,7 +33,8 @@ public class CoinApiSymbolMapper : ISymbolMapper
     {
         private const string RestUrl = "https://rest.coinapi.io";
         private readonly string _apiKey = Config.Get("coinapi-api-key");
-
+        private readonly bool _useLocalSymbolList = Config.GetBool("coinapi-use-local-symbol-list", false);
+        private readonly FileInfo _coinApiSymbolsListFile = new FileInfo("CoinApiSymbols.json");
```
mchandschuh: Should this file be sourced from configuration instead of hard-coded?

ToolBox/CoinApi/CoinApiSymbolMapper.cs
```diff
@@ -137,27 +141,38 @@ public string GetExchangeId(string market)
         private void LoadSymbolMap(string[] exchangeIds)
         {
             var list = string.Join(",", exchangeIds);
+            var json = string.Empty;
 
-            using (var wc = new WebClient())
+            if (_useLocalSymbolList && _coinApiSymbolsListFile.Exists && _coinApiSymbolsListFile.CreationTimeUtc.Date < DateTime.UtcNow.Date.AddDays(-1))
```
mchandschuh: Am I reading this correct? We want to use the local file if it is _older_ than yesterday? Also, using a file time to determine the age of data may not be the best -- wondering if it would be better to put the time the data was generated into the file instead? Consider a file being a week old and then we copy it to a new machine and it would show as new (methinks?).

StefanoRaggi: Aside from the inverted file date check, a couple of alternatives to using the file timestamp:
1. adding the date in the filename, e.g. `CoinApiSymbols.yyyyMMdd.json`
2. reading the last date from the data itself (in the `data_end` field), this is the file format:
```
[
  {
    "symbol_id": "COINBASE_SPOT_BTC_USD",
    "exchange_id": "COINBASE",
    "symbol_type": "SPOT",
    "asset_id_base": "BTC",
    "asset_id_quote": "USD",
    "data_start": "2015-01-14",
    "data_end": "2019-05-29",
    "data_quote_start": "2015-05-17T00:51:36.2870000Z",
    "data_quote_end": "2019-05-29T22:57:31.6924043Z",
    "data_orderbook_start": "2015-05-17T00:51:36.2870000Z",
    "data_orderbook_end": "2019-05-29T19:02:17.4261189Z",
    "data_trade_start": "2015-01-14T16:07:05.0000000Z",
    "data_trade_end": "2019-05-29T23:05:46.5590000Z",
    "volume_1hrs": 421.426615810,
    "volume_1hrs_usd": 3681568.97,
    "volume_1day": 12742.682746710,
    "volume_1day_usd": 111319654.79,
    "volume_1mth": 603712.812420610,
    "volume_1mth_usd": 5274015151.04
  },
  ...
]
```

ToolBox/CoinApiDataConverter/CoinApiDataConverter.cs
```diff
@@ -80,9 +80,18 @@ public bool Run()
...
-            Parallel.ForEach(_rawDataFolder.EnumerateFiles("*.gz"),(file, loopState) =>
+
+            var fileToProcess = _rawDataFolder.EnumerateFiles("*.gz")
+                .DistinctBy(
+                    x =>
+                    {
+                        var parts = x.Name.Split('-').Take(2);
+                        return string.Join("-", parts);
+                    }
+                );
```
StefanoRaggi: 👍

ToolBox/CoinApi/CoinApiSymbolMapper.cs
Jay-Jay-D: My bad, not enough coffee that afternoon.

ToolBox/CoinApi/CoinApiSymbolMapper.cs
```diff
@@ -31,7 +33,8 @@ public class CoinApiSymbolMapper : ISymbolMapper
     {
         private const string RestUrl = "https://rest.coinapi.io";
         private readonly string _apiKey = Config.Get("coinapi-api-key");
-
+        private readonly bool _useLocalSymbolList = Config.GetBool("coinapi-use-local-symbol-list", false);
```
StefanoRaggi: `Config.GetBool` defaults to `false`, so we can simplify:
``` C#
private readonly bool _useLocalSymbolList = Config.GetBool("coinapi-use-local-symbol-list");
```

ToolBox/CoinApi/CoinApiSymbolMapper.cs
StefanoRaggi: Agree we should give this option with a default value:
``` C#
private readonly FileInfo _coinApiSymbolsListFile = new FileInfo(
    Config.Get("coinapi-default-symbol-list-file", "CoinApiSymbols.json"));
```

ToolBox/CoinApi/CoinApiSymbolMapper.cs
StefanoRaggi: Thinking more about it, I'm wondering if we even need the file date check 🤔 

When the `CoinApiSymbolMapper` is used in the `CoinApiDataQueueHandler` or in an occasional converter run, there will only be a single instance, so API rate limits are not an issue.

When it's used for massively parallel processing with the `CoinApiDataConverter`, the date check could return false for all processes and we wouldn't guarantee rate limits won't be hit. Perhaps for these scenarios, the file could be pre-downloaded (once a day) and the symbol mapper would just read the file with no date checks?

``` C#
if (_useLocalSymbolList)
{
    if (!_coinApiSymbolsListFile.Exists) 
    {
        throw new Exception($"File not found: {_coinApiSymbolsListFile.FullName}");
    }
}
```

PR #3264, opened 2019-06-04
IQFeed Downloader Program

aytacasan:
#### Description
IQFeed data downloader program.

#### Related Issue
Currently, there is no data downloader for IQFeed in LEAN.

#### Motivation and Context
I need data downloader for backtesting and see there is other users requested it.

I don't think so but if there is a list of downloaders maybe this downloader must be add to that list too.

I do some tests manually. Download data and check is it same as with provider's data.

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/IQFeedDownloader/IQFeedDataDownloader.cs
+++ b/ToolBox/IQFeedDownloader/IQFeedDataDownloader.cs
@@ -1,0 +1,115 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Securities;
+using QuantConnect.ToolBox.IQFeed;
+
+namespace QuantConnect.ToolBox.IQFeedDownloader
+{
+    /// <summary>
+    /// IQFeed Data Downloader class 
+    /// </summary>
+    public class IQFeedDataDownloader : IDataDownloader
+    {
+        private readonly IQConnect _iqConnect;
+        private readonly HistoryPort _historyPort;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="IQFeedDataDownloader"/> class
+        /// </summary>
+        /// <param name="username">IQFeed user name</param>
+        /// <param name="password">IQFeed password</param>
+        /// <param name="productName">IQFeed product name</param>
+        /// <param name="productVersion">IQFeed product version</param>
+        public IQFeedDataDownloader(string userName, string password, string productName, string productVersion)
+        {
+            _iqConnect = new IQConnect(productName, productVersion);
+            _iqConnect.Launch(new IQCredentials(userName, password, true, true));
+
+            _historyPort = new HistoryPort(new IQFeedDataQueueUniverseProvider(), 0, 0);
+            _historyPort.Connect();
+            _historyPort.SetClientName("History");
+        }
+
+        /// <summary>
+        /// Get historical data enumerable for a single symbol, type and resolution given this start and end time (in UTC).
+        /// </summary>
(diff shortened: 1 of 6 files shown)


ToolBox/IQFeedDownloader/IQFeedDataDownloader.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// <param name="symbol"></param>
+        /// <param name="ticks"></param>
+        /// <param name="resolution"></param>
+        /// <returns></returns>
+        internal static IEnumerable<TradeBar> AggregateTicks(Symbol symbol, IEnumerable<Tick> ticks, TimeSpan resolution)
+        {
+            return (from t in ticks
+                    group t by t.Time.RoundDown(resolution) into g
+                    select new TradeBar
+                    {
```
mchandschuh: might want to explicitly set the `Period` property of trade bar to the provided `resolution` -- this will ensure the correct `EndTime` will be set. In LEAN, all data is time synced by the end time, so it turns out to be most important to get that time correct.

aytacasan:
When i look at LEAN's sample data under data folder (for example lean\data\equity\usa\tick\spy) i see there is miliseconds, lastprice, quantity, exchange, scalecondition, suspicious columns.

What is that mean that characters P, T, @ etc. under exchange and scalecondition? Are they market makers codes or something like that?

What is the source of suspicious column, what is the criterias that data is suspicious or not?

Other tick downloaders (for example fxcm's or dukascopy's data downloaders) don't write quantity. Reason is they don't provide us that column or is there any other reason?

From a perspective of iqfeed data downloader, there is no way to write that columns because iqfeed does not provide us. So is there any penalty about that when we do backtest without these columns?

My other problem is performance. When i try to download tick data with that socket command AAPL Tick 20190520 20190524 via iqfeed's sample application that they provide their cusotmers. Process take a time only 45 seconds. But with our downloader it's take a time 15 minutes. Difference is so huge. Is there any idea why this is like that?

ToolBox/IQFeedDownloader/IQFeedDataDownloader.cs
aytacasan: Before develop this downloader, i work on iqfeed api, especially on their api examples. After a while i see actually lean already have almost all required codes for accomplish this task. Because i don't want to duplicate many lines of code, i do small changes current code and use them. For example used current HistoryPort implementation for request history from iqfeed's history socket port. After decide to support write all resolutions with a single command, just like HistoryPort i try to follow up current (similar) codes and check other donwloaders and see they are aggregating ticks without set EndTime property. I did not encounter any problems about that. But of course it should be examined in more detail.

mchandschuh:
Quantity/trade size is not available for forex data, that's why they don't write it. If a value is not provided, such as sale condition or suspicious, then we can't write it to the csv and that's OK. We minimally need the timestamp and the price along. For tick resolution data we'll also need the tick type (quote/trade).

IIRC, the suspicious field is provided by the data provider and is used to indicate a tick that may have been reported to the tape in delayed fashion _or_ is a trade that executed on a dark pool. IOW, it likely contains a price that you would not be able to execute against during normal trading on the exchange.

RE: Performance -- I can't opine without seeing logs or something else to help direct to where all time is spent.

ToolBox/IQFeedDownloader/IQFeedDataDownloader.cs
mchandschuh: I think we should still verify the correctness of this to ensure the trade bars are creating with the correct end times as that's what LEAN uses for time synchronization.

mathpaquette:
@jaredbroad are you open to integrate https://github.com/mathpaquette/IQFeed.CSharpApiClient in Lean. Maybe I can revisit the implementation?

Issue #3281, opened 2019-06-07
Add Cash and CashBook Update event


Martin-Molinero:
#### Expected Behavior

- `Cash` and `CashBook` classes should provide an `Update` event, which gets invoked when the instances are updated
#### Actual Behavior

- `Cash` and `CashBook` have to be pulled/verified to detected any change
#### Potential Solution

- Add new `public event EventHandler Updated`
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3282, opened 2019-06-07
Adding Updated event for Cash and CashBook

Martin-Molinero:
#### Description

- Adding new `Updated` event for `Cash` and `CashBook`. Adding unit
tests.
- `CashBook.Update` event will also be called for updates in the `Cash`
instances he holds.

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3281
#### Motivation and Context

- `Cash` and `CashBook` have to be pulled/verified to detected any change

N/A

Unit tests and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -288,9 +288,7 @@ namespace QuantConnect.Securities
 
         private void OnUpdate()
         {
-            // copy for thread safety
-            var handler = Updated;
-            handler?.Invoke(this, EventArgs.Empty);
+            Updated?.Invoke(this, EventArgs.Empty);
         }
     }
 }
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -29,10 +29,17 @@ namespace QuantConnect.Securities
     /// </summary>
     public class Cash
     {
+        private decimal _conversionRate;
         private bool _isBaseCurrency;
         private bool _invertRealTimePrice;
 
         private readonly object _locker = new object();
+
+        /// <summary>
+        /// Event fired when this instance is updated
+        /// <see cref="AddAmount"/>, <see cref="SetAmount"/>, <see cref="Update"/>
+        /// </summary>
+        public event EventHandler Updated;
 
         /// <summary>
         /// Gets the symbol of the security required to provide conversion rates.
@@ -61,7 +68,18 @@ namespace QuantConnect.Securities
         /// <summary>
         /// Gets the conversion rate into account currency
         /// </summary>
-        public decimal ConversionRate { get; internal set; }
+        public decimal ConversionRate
+        {
+            get
+            {
+                return _conversionRate;
+            }
+            internal set
+            {
+                _conversionRate = value;
+                OnUpdate();
+            }
+        }
 
         /// <summary>
         /// The symbol of the currency, such as $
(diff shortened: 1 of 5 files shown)


Common/Securities/CashBook.cs
```diff
@@ -408,5 +409,39 @@ protected virtual void OnCashAdded(Cash cash)
...
+        private void OnCashUpdate(object sender, EventArgs eventArgs)
+        {
+            OnUpdate();
+        }
+
+        private void OnUpdate()
+        {
+            // copy for thread safety
+            var handler = Updated;
+            handler?.Invoke(this, EventArgs.Empty);
```
StefanoRaggi: I think the copy is no longer necessary when using the `?.` operator:
https://codeblog.jonskeet.uk/2015/01/30/clean-event-handlers-invocation-with-c-6/

`Updated?.Invoke(this, EventArgs.Empty);` should be safe

Martin-Molinero: Right but since we are using `.Net 4.5.2` I think we are using C# 5? 
https://www.tutorialsteacher.com/csharp/csharp-version-history (I wasn't able to quickly find a more official link 😅 )

StefanoRaggi: All Lean projects are configured to build with C# 6 compiler (which is supported with VS 2015 and later versions). We have NET 4.5.2 as target framework version, but we could build with any C# compiler version, also C#7 which have new syntax features, but this would require at least VS 2017.

Martin-Molinero: Awesome, I'll update this then 👍

Engine/DataFeeds/CurrencySubscriptionDataConfigManager.cs
```diff
@@ -53,7 +53,13 @@ public class CurrencySubscriptionDataConfigManager
             ISecurityService securityService,
             IBrokerageModel brokerageModel)
         {
-            cashBook.CashAdded += (sender, cash) => _ensureCurrencyDataFeeds = true;
+            cashBook.Updated += (sender, updateType) =>
+            {
+                if (updateType == CashBook.UpdateType.Added)
```
jaredbroad: ❤️

Common/Securities/Cash.cs
```diff
@@ -29,11 +29,18 @@ namespace QuantConnect.Securities
...
         private bool _isBaseCurrency;
         private bool _invertRealTimePrice;
 
         private readonly object _locker = new object();
 
+        /// <summary>
+        /// Event fired when this instance is updated
+        /// <see cref="AddAmount"/>, <see cref="SetAmount"/>, <see cref="Update"/>
+        /// </summary>
+        public event EventHandler Updated;
```
mchandschuh: no need to change :P -- I was originally thinking this would have some information in the event args, such as old value/new value - this would enable future code to be written incrementally and always ensure we're not re-computing things. For example, in this case consumers are likely to use the `ValueInAccountCurrency`, so multiple consumers would yield extra evaluations of the property.

Engine/DataFeeds/CurrencySubscriptionDataConfigManager.cs
mchandschuh: very nice! 👍

PR #3285, opened 2019-06-09
Created new method. To set configuration value at specified json path.

IlshatGaripov:
#### Description
The method allows to set the value in the configuration at the specified path. This can be useful for passing algorithm parameters to a program. For example, Config.SetValue("parameters.intrinio-username", "igaripov")

#### Motivation and Context
I wrote a program where I passed various parameters to the agorithm programmatically. While there is a function to read a value from composite path. Example - Config.GetValue<decimal>("parameters.sma26") there is no natural method to set values to a voluntary config path. I implemented the functionality with this commit.

I ran few exmplaes .

#### Types of changes

#### Checklist:

### DIFF
--- a/Configuration/Config.cs
+++ b/Configuration/Config.cs
@@ -182,6 +182,44 @@ namespace QuantConnect.Configuration
         }
 
         /// <summary>
+        /// Sets new configuration value at the specified path. If some of the properties
+        /// that constitute a path do not exist method will create them.
+        /// </summary>
+        /// <param name="path">Json value path. A path could be multi-levelled (separated by commas).</param>
+        /// <param name="value">Value to set.</param>
+        public static void SetValue(string path, string value)
+        {
+            // value token at given path
+            var token = Settings.Value.SelectToken(path);
+
+            // if do not exist - we need to create the missing properties
+            if (token == null)
+            {
+                // Going down a multi-level path we will create the missing properties
+                // Current json object will hold the inner JObject that resides by the current path
+                var currentJObject = Settings.Value;
+                foreach (var word in path.Split('.'))
+                {
+                    if (currentJObject[word] == null)
+                    {
+                        currentJObject.Add(word, new JObject());
+                    }
+
+                    // Move an object one step down
+                    currentJObject = (JObject)currentJObject[word];
+                }
+
+                // Now as we made sure the token for given path does exist - populate the token with given value
+                currentJObject.Replace(value);
+            }
+            else
+            {
+                // If path token does exist - just populate with given value
+                token.Replace(value);
+            }
+        }
+
+        /// <summary>
         /// Get a boolean value configuration setting by a configuration key.
         /// </summary>
         /// <param name="key">String value of the configuration key.</param>
--- a/Configuration/Config.cs
+++ b/Configuration/Config.cs
@@ -159,63 +159,59 @@ namespace QuantConnect.Configuration
         }
 
         /// <summary>
-        /// Sets a configuration value. This is really only used to help testing. The key heye can be
-        /// specified as {environment}.key to set a value on a specific environment
+        /// Sets a configuration value. If isEnvironmentContext is set to true (default) key can be
+        /// specified as {environment}.key to set a value on a specific environment. When false,
+        /// key is considered the full path to the meaning object.
         /// </summary>
         /// <param name="key">The key to be set</param>
         /// <param name="value">The new value</param>
-        public static void Set(string key, string value)
+        /// <param name="isEnvironmentContext">Flag indicates whether to understand key as specific environment key</param>
+        public static void Set(string key, string value, bool isEnvironmentContext = true)
         {
-            JToken environment = Settings.Value;
-            while (key.Contains("."))
+            if (isEnvironmentContext)
             {
-                var envName = key.Substring(0, key.IndexOf("."));
-                key = key.Substring(key.IndexOf(".") + 1);
-                var environments = environment["environments"];
-                if (environments == null)
+                JToken environment = Settings.Value;
+                while (key.Contains("."))
                 {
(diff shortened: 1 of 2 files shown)


Doggie52:
What's the difference between this and the already existing `Config.Set( "config", "value" )`?

IlshatGaripov:
@Doggie52  this will raise an error Config.Set( "parameters.intrinio-username", "igaripov" ) but Config.Set( "parameters-intrinio-username", "igaripov" )

Doggie52:
Apologies - missed the `parameters` bit.

What you need is the following:
```c#
SetParameters( new Dictionary<string, string> { { "intrinio-username", "igaripov" } } );
```

IlshatGaripov:
@Doggie52 thanks! I didn't know of existance of SetParameters. I set parameters programmatically before the Engine even launched. So the algorithm is not yet even existent. Anyways, new method is not the thing impossible to do the things without.. I can quite cope with Config.Set( "some-unique-param-identified", "value" ) . but I thought ok but how would I set the value at any desired path, I wrote the method and committed as this perhaps could be of any future use .. as it brings additional fuctionality to the lib..

Configuration/Config.cs
```diff
@@ -181,6 +181,44 @@ public static void Set(string key, string value)
             environment[key] = value;
         }
 
+        /// <summary>
+        /// Sets new configuration value at the specified path. If some of the properties
+        /// that constitute a path do not exist method will create them.
+        /// </summary>
+        /// <param name="path">Json value path. A path could be multi-levelled (separated by commas).</param>
+        /// <param name="value">Value to set.</param>
+        public static void SetValue(string path, string value)
```
mchandschuh: How will this behave regarding multiple loan environments? I suppose they would be handled transparently by simply providing a path that is environment aware.

IlshatGaripov: It does not affect environments - yes - would just help to store the value at precise path provided. If path does not exist this will create it. I thought this could be helpful to merge some logically related variables within a single Json object. For example.. to set this 
  "parameters": {
    "intrinio-username": "",
    "intrinio-password": "",
    "ema-fast": 10,
    "ema-slow": 20
  },

I could call 

Config.SetValue("parameters.intrinio-username", "");
Config.SetValue("parameters.ema-slow", "20");
etc. 

Current Set will throw an exception if I try set any of above. This line will throw an error:
environment[key] = value;

And this is likely because JsonNet indexer would not let to set value with complex path but works fine with simple. This would be ok:
Config.Set("alpaca-trading-mode", "paper")

mchandschuh:
Regarding the deficiency in `Config.Set` -- perhaps improving that to properly handle nested objects would be the correct way forward? Also, unit tests would be great to add for this kind of change.

IlshatGaripov:
Hi! Have made the changes as suggested. Combined both methods in one. I introduced a new variable, a flag, that tells how to treat the key - as {environment}.key or literally like a full path to an object. Does it sound good? Thanks.

Configuration/Config.cs
```diff
@@ -159,26 +159,60 @@ public static JToken GetToken(string key)
...
         /// </summary>
         /// <param name="key">The key to be set</param>
         /// <param name="value">The new value</param>
-        public static void Set(string key, string value)
+        /// <param name="isEnvironmentContext">Flag indicates whether to understand key as specific environment key</param>
+        public static void Set(string key, string value, bool isEnvironmentContext = true)
         {
-            JToken environment = Settings.Value;
-            while (key.Contains("."))
+            if (isEnvironmentContext)
```
mchandschuh: Is there a reason to have this flag? I suppose the only reason would be to support keys that have a dot in them. Do we have such a case, and if so, I would recommend against configuration keys that have periods in them due to the ambiguity that can arise.

Tests/Configuration/ConfigTests.cs
```diff
@@ -37,6 +36,20 @@ public void SetRespectsEnvironment()
             Assert.AreNotEqual(betaMode, betaMode2);
         }
 
+        [Test]
+        public void SetNestedObject()
```
mchandschuh: I think we'll want to include more tests, minimally testing both values of the `isEnvironmentContext` flag, but my vote would be to remove the flag or provide reasoning as to why we need to forms of using `Config.Set` -- unless I'm missing something, I think it can simply be removed.

Perhaps there is special handling regarding environments? My understanding was that we handled it just the same as nested objects, such as `environments.back-testing.my-setting`

IlshatGaripov: Thank you  @mchandschuh  for taking time to review all this. Well I agree isEnvironmentContext could be redundant. May I please clarify one thing..  As I could judge by the code of initial Config.Set it would expect that any assignment that has dot separator in its key specifier, i.e nested object, to have relation to one or another environment or not? Simply saying assigment to **environments.back-testing.my-setting** would be done by calling Config.Set(back-testing.my-setting, value) ?

IlshatGaripov: I made a small test for initial Set. How would it behave if we want to assign a value to environment that config.json has to information of?  Example:
Config.Set("mega-backtesting.setup-handler", "3285-testing");
I realized it will throw an exception. Here come a commit next .

IlshatGaripov:
This small change in initial Config.Set will make the things work:
line183:  environment = environments[envName] = new JObject();
I applied the fix and redid the test.  -> commit follows

jaredbroad:
Closing as a negligible overall difference to well-designed algorithms and may encourage poor patterns. Suggest a better way set parameters for algorithms is the job.Parameters system.

IlshatGaripov:
@jaredbroad job.Parameters as it designed now also reads these values from Config. And there is not method to way to set them programatically. So I created such method. that's how the descussion began =)

PR #3288, opened 2019-06-10
Fix Live coarse universe carrying over securities from previous day

### LINKED ISSUE #3287: Live coarse universe may include securities from previous day
#### Expected Behavior
After Live coarse universe selection runs, we expect algorithm.ActiveSecurities to include only the securities selected (plus any security with open positions and orders).

#### Actual Behavior
Depending on the time Live coarse universe selection runs, securities which were selected previously may be kept in the universe (even if there are no open positions/orders).

#### Potential Solution
Update the default value of `QCAlgorithm.UniverseSettings.MinimumTimeInUniverse` from `TimeSpan.FromDays(1)` to `TimeSpan.Zero`.

#### Reproducing the Problem
Run a Coarse universe selection algorithm in Live mode for a week, logging the symbols selected by the coarse selection function and the symbols in `algorithm.ActiveSecurities` each day. 
On some days, securities from the previous day will be carried over to the next day.

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- The default value of `QCAlgorithm.UniverseSettings.MinimumTimeInUniverse` has been changed from `TimeSpan.FromDays(1)` to `TimeSpan.Zero`.

#### Related Issue
Closes #3287 

#### Motivation and Context
Live algorithms using Coarse universe selection were including symbols selected on the previous day.
This is caused by the fact that in Live mode Coarse universe selection is not triggered **exactly** at the same time every day, but it can happen several minutes before (or after) the 24-hour offset from the time of the previous day.

No.

New unit test included.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Universe.cs
+++ b/Algorithm/QCAlgorithm.Universe.cs
@@ -373,6 +373,8 @@ namespace QuantConnect.Algorithm
         /// <param name="selector">Defines an initial coarse selection</param>
         public void AddUniverse(Func<IEnumerable<CoarseFundamental>, IEnumerable<Symbol>> selector)
         {
+            CheckLiveCoarseUniverseSettings();
+
             AddUniverse(new CoarseFundamentalUniverse(UniverseSettings, SecurityInitializer, selector));
         }
 
@@ -384,6 +386,8 @@ namespace QuantConnect.Algorithm
         /// <param name="fineSelector">Defines a more detailed selection with access to more data</param>
         public void AddUniverse(Func<IEnumerable<CoarseFundamental>, IEnumerable<Symbol>> coarseSelector, Func<IEnumerable<FineFundamental>, IEnumerable<Symbol>> fineSelector)
         {
+            CheckLiveCoarseUniverseSettings();
+
             var coarse = new CoarseFundamentalUniverse(UniverseSettings, SecurityInitializer, coarseSelector);
 
             AddUniverse(new FineFundamentalFilteredUniverse(coarse, fineSelector));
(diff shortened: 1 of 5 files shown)


Algorithm/QCAlgorithm.Universe.cs
```diff
@@ -542,6 +546,22 @@ private void ConfigureUnderlyingSecurity(Security security)
...
+        /// </summary>
+        private void CheckLiveCoarseUniverseSettings()
+        {
+            if (_liveMode && UniverseSettings.MinimumTimeInUniverse >= TimeSpan.FromDays(1))
+            {
+                Debug("Warning: Coarse universe selection in Live mode does not always happen at the same time. " +
+                      "The current setting of UniverseSettings.MinimumTimeInUniverse (one day or more) could cause unexpected universe members. " +
+                      "For this reason the setting has been updated to TimeSpan.Zero.");
+
+                UniverseSettings.MinimumTimeInUniverse = TimeSpan.Zero;
```
mchandschuh: So it's now _impossible_ to set this to be greater than 1 day? I imagine some subset of users depend on the min time setting? I suppose the other way for a user to enforce a min time would be to manually track it in a local subscriptions list to feed back into the coarse selector. Is there a way to default the value to zero for coarse but permit it to be overridable?

Algorithm/QCAlgorithm.cs
```diff
@@ -153,7 +153,7 @@ public QCAlgorithm()
             // universe selection
             UniverseManager = new UniverseManager();
             Universe = new UniverseDefinitions(this);
-            UniverseSettings = new UniverseSettings(Resolution.Minute, 2m, true, false, TimeSpan.FromDays(1));
+            UniverseSettings = new UniverseSettings(Resolution.Minute, 2m, true, false, TimeSpan.Zero);
```
mchandschuh: I suspect this would be a breaking change for many user algorithms?

jaredbroad:
Per discussion need to think about this in a broader context to see if we can solve it neatly. Agree with @mchandschuh 's comment changing the CoarseFundamental Universe setting for the time in universe is OK;

Other options discussed were 
 - Changing the refresh frequency of the universe to remove these redundant securities more often than 1 day; 
 - Setting a scheduled time for the universe to run; say 8am each day, for outside of coarse time variance.
 - Setting a dynamic "minimum in universe" period for the Coarse based on when the live feed arrives.

The other current live universe timing issue is the option daily poll at midnight for the new contract list is happening when IB is closed -- resulting in any option live subscription requests failing and killing LEAN. This could also be handled with a scheduled universe event.

PR #3309, opened 2019-06-14
Add a virtual property get in LiveSynchronizer

StefanoRaggi:
#### Description
- This new property in `LiveSynchronizer` enables overriding of the hardcoded `500ms` value in unit tests.

#### Related Issue
n/a

#### Motivation and Context
Enables custom configurations for additional live trading unit tests.

No.

Live trading data feed unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/LiveSynchronizer.cs
+++ b/Engine/DataFeeds/LiveSynchronizer.cs
@@ -31,6 +31,11 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         private readonly AutoResetEvent _newLiveDataEmitted = new AutoResetEvent(false);
 
         /// <summary>
+        /// Maximum time to wait for new live data before synchronizing the data feed subscriptions
+        /// </summary>
+        protected virtual TimeSpan NewLiveDataTimeout { get; } = TimeSpan.FromMilliseconds(500);
+
+        /// <summary>
         /// Initializes the instance of the Synchronizer class
         /// </summary>
         public override void Initialize(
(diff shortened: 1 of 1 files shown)


Engine/DataFeeds/LiveSynchronizer.cs
```diff
@@ -30,6 +30,11 @@ public class LiveSynchronizer : Synchronizer
     {
         private readonly AutoResetEvent _newLiveDataEmitted = new AutoResetEvent(false);
 
+        /// <summary>
+        /// Maximum time to wait for new live data before synchronizing the data feed subscriptions
+        /// </summary>
+        protected virtual TimeSpan NewLiveDataTimeout { get; } = TimeSpan.FromMilliseconds(500);
```
mchandschuh: Do we always create a sub class for the unit tests? If so, then disregard this suggestion, but thinking it would be easier to just make it a public get/set.

StefanoRaggi: We already have a subclass in the `LiveTradingDataFeed` tests (`TestableLiveSynchronizer`),
which I was thinking to change to something like this:

``` C#
    internal class TestableLiveSynchronizer : LiveSynchronizer
    {
        private readonly ITimeProvider _timeProvider;

        protected override TimeSpan NewLiveDataTimeout { get; }

        public TestableLiveSynchronizer(ITimeProvider timeProvider = null, TimeSpan? newLiveDataTimeout = null)
        {
            _timeProvider = timeProvider ?? new RealTimeProvider();
            NewLiveDataTimeout = newLiveDataTimeout ?? base.NewLiveDataTimeout;
        }

        protected override ITimeProvider GetTimeProvider()
        {
            return _timeProvider;
        }
    }
```
Then, only in specific tests:
``` C#
var synchronizer = new TestableLiveSynchronizer(timeProvider, TimeSpan.FromMilliseconds(50));
```

mchandschuh: Awesome sauce

PR #3314, opened 2019-06-14
Adding PredicateTimeProvider for live mode

Martin-Molinero:
#### Description

- Adding new ~`StepTimeProvider`~ `PredicateTimeProvider` that will advance time based on a
desired custom evaluator.
   - **Note**: this class could potentially also accept a `TimeSpan` to use as time step, limiting the emission of data selection (or any data) to intervals. This was initially implemented for the `Member Time In Universe` issue but the `Rounding` solution made it unrequired, so was removed.
- Live trading `Options`, `Futures` and `Coarse` data will use a
`FrontierAwareEnumerator` + ~`StepTimeProvider`~ `PredicateTimeProvider`. This will allow holding
the selection data until its the desired time: `time.Hour < 23 && time.Hour > 5 && time.DayOfWeek != DayOfWeek.Saturday` -> from 6AM to 22PM except Saturday
- `Universe.CanRemoveMember` will round the members time in the universe
based on the `UniverseSettings.MinimumTimeInUniverse`
- `LiveCoarseUniverseTests` taken from previous PR, @StefanoRaggi author. https://github.com/QuantConnect/Lean/pull/3288/files#diff-2156b0d39801199b451ce096dd5b2a75

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3287 and `Option daily poll at midnight for the new contract list is happening when IB is closed -- resulting in any option live subscription requests failing and killing LEAN.`

#### Motivation and Context

- Live coarse universe may include securities from previous day

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/UniverseSelection/Universe.cs
+++ b/Common/Data/UniverseSelection/Universe.cs
@@ -29,6 +29,10 @@ namespace QuantConnect.Data.UniverseSelection
     /// </summary>
     public abstract class Universe : IDisposable
     {
+        /// <summary>
+        /// Used to round the members time in universe <see cref="CanRemoveMember"/>, this is
+        /// done because we can not guarantee exact selection time in live mode, see GH issue 3287
+        /// </summary>
         private TimeSpan? _minimumTimeInUniverseRoundingInterval;
 
         /// <summary>
@@ -160,6 +164,7 @@ namespace QuantConnect.Data.UniverseSelection
             {
                 if (_minimumTimeInUniverseRoundingInterval == null)
                 {
+                    // lets set _minimumTimeInUniverseRoundingInterval once
                     _minimumTimeInUniverseRoundingInterval = UniverseSettings.MinimumTimeInUniverse;
                     AdjustMinimumTimeInUniverseRoundingInterval();
                 }
--- a/Common/Data/UniverseSelection/Universe.cs
+++ b/Common/Data/UniverseSelection/Universe.cs
@@ -29,6 +29,8 @@ namespace QuantConnect.Data.UniverseSelection
     /// </summary>
     public abstract class Universe : IDisposable
     {
+        private TimeSpan? _minimumTimeInUniverseRoundingInterval;
+
         /// <summary>
         /// Gets a value indicating that no change to the universe should be made
         /// </summary>
@@ -156,8 +158,15 @@ namespace QuantConnect.Data.UniverseSelection
             Member member;
             if (Securities.TryGetValue(security.Symbol, out member))
             {
+                if (_minimumTimeInUniverseRoundingInterval == null)
+                {
+                    _minimumTimeInUniverseRoundingInterval = UniverseSettings.MinimumTimeInUniverse;
+                    AdjustMinimumTimeInUniverseRoundingInterval();
+                }
+
                 var timeInUniverse = utcTime - member.Added;
-                if (timeInUniverse >= UniverseSettings.MinimumTimeInUniverse)
+                if (timeInUniverse.Round(_minimumTimeInUniverseRoundingInterval.Value)
+                    >= UniverseSettings.MinimumTimeInUniverse)
                 {
                     return true;
                 }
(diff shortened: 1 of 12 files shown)


Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -577,6 +578,22 @@ private IEnumerator<BaseData> GetNextTicksEnumerator()
...
+        /// using a <see cref="StepTimeProvider"/> that will advance time based on the provided
+        /// function
+        /// </summary>
+        /// <remarks>Won't advance time if now: bigger than 23pm, less than 6am or Saturday</remarks>
+        private IEnumerator<BaseData> GetConfiguredFrontierAwareEnumerator(IEnumerator<BaseData> enumerator,
+            TimeZoneOffsetProvider tzOffsetProvider)
+        {
+            var stepTimeProvider = new StepTimeProvider(_frontierTimeProvider,
+                // advance time if before 23pm or after 5am and not on Saturdays
+                time => time.Hour < 23 && time.Hour > 5 && time.DayOfWeek != DayOfWeek.Saturday);
```
mchandschuh: These times/days of weeks seem very much coupled to particular security types. The documentation states _what_ it does but not _why_ it does it.

Engine/DataFeeds/StepTimeProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// Gets the current utc time step
+        /// </summary>
+        public DateTime GetUtcNow()
+        {
+            var utcNow = _underlyingTimeProvider.GetUtcNow();
+
+            // we check if we should advance time based on the provided custom step evaluator
+            if (_customStepEvaluator(utcNow))
```
mchandschuh: just some thoughts -- This type's name implies it performs some step-wise logic, but it appears that the 'step-wise-ness' is actually a function of the provided `customStepEvaluator` -- it almost seems like its more of a `PredicateTimeProvider` where if the given predicate returns true it updates the time and if it returns false it does not update the time

Common/Data/UniverseSelection/Universe.cs
```diff
@@ -29,6 +29,8 @@ namespace QuantConnect.Data.UniverseSelection
     /// </summary>
     public abstract class Universe : IDisposable
     {
+        private TimeSpan? _minimumTimeInUniverseRoundingInterval;
```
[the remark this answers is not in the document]
Martin-Molinero: I'll add a comment for him.
`What it aims to support`:
I'm using it to round the members `timeInUniverse`, why? **as a fix** for the fact that we can't guarantee the **exact** selection time, example `UniverseSettings.MinimumTimeInUniverse` could be 1 day and `timeInUniverse` 0.99 day. See related issue. 
Even if we used the new `StepTimeProvider` to set up a schedule, say advance time once a day, it seems error prone and weak to potential race conditions.

Made him nullable to avoid calling `AdjustMinimumTimeInUniverseRoundingInterval()` on each call to `CanRemoveMemeber()`, so I'm adding this field to hold the value and setting him just once. Can't use `UniverseSettings.MinimumTimeInUniverse` for rounding directly.

Let me know what you think :)

Engine/DataFeeds/LiveTradingDataFeed.cs
Martin-Molinero: Good point! I'll add _why_ it does it

Engine/DataFeeds/StepTimeProvider.cs
Martin-Molinero: True! Originally I implemented him as a step time provider, for the `timeInUniverse` fix, with a custom predicate for the other related issue but since the rounding was needed anyway I decided to remove the stepping on an interval.
I'll update the name 👍

Common/Data/UniverseSelection/Universe.cs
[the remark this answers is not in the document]
mchandschuh: Yup, this is what I figured it was used for -- I just want to make sure we capture the _why_ behind things as it will be incredibly helpful in the future. I think this approach makes sense given that variability of invocation time

👍

PR #3316, opened 2019-06-18
Update FxcmBrokerage for TLS 1.2 support

StefanoRaggi:
#### Description
- Added support for TLS 1.2 to `FxcmBrokerage`
- Updated `QuantConnect.Fxcm.dll` from https://github.com/QuantConnect/FXCM/pull/1
- Updated the IKVM libraries from `v7.3.4830.1` to `v8.1.5717.0` (last version available).

#### Related Issue
https://github.com/QuantConnect/FXCM/pull/1

#### Motivation and Context
These changes are required to connect and receive data properly since the **TLS 1.2 server upgrade on 6/16/2019**. This also required upgrading the IKVM libraries for C#/Java interop.

#### Requires Documentation Change
No.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [ ] I have added tests to cover my changes. 
- [ ] All new and existing tests passed.
- [ ] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #3318, opened 2019-06-18
Update FxcmBrokerage for TLS 1.2 support (resubmit)

StefanoRaggi:
- Resubmit of #3316 (reverted because of missing IKVM modules in Launcher folder)


Brokerages/QuantConnect.Brokerages.csproj
```diff
@@ -45,96 +45,92 @@
     <Reference Include="CSharpAPI">
       <HintPath>InteractiveBrokers\CSharpAPI.dll</HintPath>
     </Reference>
-    <Reference Include="IKVM.OpenJDK.Beans, Version=7.2.4630.5, Culture=neutral, PublicKeyToken=13235d27fcbfff58, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\packages\IKVM-WithExes.7.3.4830.1\tools\IKVM.OpenJDK.Beans.dll</HintPath>
+    <Reference Include="IKVM.AWT.WinForms, Version=8.1.5717.0, Culture=neutral, PublicKeyToken=13235d27fcbfff58, processorArchitecture=MSIL">
```
mchandschuh: I presume these are the 'latest' versions that you manually converted? I forgot that FXCM connected via java code. Just wondering how we can document what was done here to produce these binaries so someone can follow a similar process in the future. Perhaps adding some details to the commit message would be sufficient

Brokerages/Fxcm/FxcmBrokerage.cs
```diff
@@ -102,6 +103,9 @@ public FxcmBrokerage(IOrderProvider orderProvider, ISecurityProvider securityPro
 
             HistoryResponseTimeout = 5000;
             MaximumHistoryRetryAttempts = 1;
+
+            // FXCM requires TLS 1.2 since 6/16/2019
+            ServicePointManager.SecurityProtocol |= SecurityProtocolType.Tls12;
```
mchandschuh: This may be more naturally phrased as part of a static (run-once) constructor

Brokerages/app.config
```diff
@@ -0,0 +1,59 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <runtime>
```
mchandschuh: This is a new file, and since the brokerages project is a library and not an exe, I think this file is not necessary

Jupyter/app.config
```diff
@@ -0,0 +1,59 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
mchandschuh: Similar comment here, wonder if this is required.

Launcher/QuantConnect.Lean.Launcher.csproj
```diff
@@ -45,6 +45,90 @@
     <Reference Include="DynamicInterop, Version=0.7.4.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\DynamicInterop.0.7.4\lib\net40\DynamicInterop.dll</HintPath>
     </Reference>
+    <Reference Include="IKVM.AWT.WinForms, Version=8.1.5717.0, Culture=neutral, PublicKeyToken=13235d27fcbfff58, processorArchitecture=MSIL">
```
mchandschuh: Interestingly, it looks like this wasn't required before. Has something changed?

Brokerages/QuantConnect.Brokerages.csproj
mchandschuh: See https://github.com/QuantConnect/FXCM/pull/1 for update to FXCM libraries.

mchandschuh:
Forcing issue linkage with FXCM changes: https://github.com/QuantConnect/FXCM/pull/1

Brokerages/app.config
StefanoRaggi: These binding redirects were added automatically by VS 2017 when I updated the NuGet reference to the new version of IKVM.
But you're right, they are not needed at all -- thankfully we don't use multiple versions of IKVM in LEAN 😄 
I will remove them.

Jupyter/app.config
StefanoRaggi: Not needed, will remove 👍

Brokerages/QuantConnect.Brokerages.csproj
StefanoRaggi: The only extra step required (after building a new version of `QuantConnect.Fxcm.dll` in https://github.com/QuantConnect/FXCM) is to manually copy the DLL in the `Brokerages/Fxcm` folder and push it to the LEAN repo (as done in this PR). 
Maybe in the future, we could create a NuGet package for it.

Launcher/QuantConnect.Lean.Launcher.csproj
StefanoRaggi: Looking at the IKVM version history on NuGet (https://www.nuget.org/packages/IKVM/), we upgraded from v7.3 to v8.1 so the new version includes 2 years of changes (as well as a few new modules). I don't exactly know which is the minimal set of DLLs required for LEAN usage and the dependency tree, but since debugging MEF loader errors is quite painful and time-consuming, I opted for the safe solution for now -- including all of the references added automatically by the NuGet package manager.

Issue #2865, opened 2019-01-27
OnEndOfDay not firing for Framework algos


Worthy7:
#### Expected Behavior
Fires

#### Actual Behavior
Doesn't fire

#### Potential Solution
Don't know

#### Reproducing the Problem
Add 
```
        public override void OnEndOfDay()
        {
            base.OnEndOfDay();
            Debug("EOD");
        }
```
To the `BasicTemplateFrameworkAlgorithm`

#### Checklist

- [ ] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

Issue #2625, opened 2018-10-19
Setting Multiple Scheduled Events to Same Time Does Not Trigger in Deterministic Order


arshamawan:
#### Expected Behavior
We want the Scheduled Events, which fire at the same time, to activate in the same order in both online BTs and local.

#### Actual Behavior
They can happen the other way round.

mchandschuh:
The underlying cause of this is that the scheduled events are stored in a concurrent dictionary which has non-deterministic enumeration behavior. We could add an index to each entry and sort them before invoking the handlers.

PR #3331, opened 2019-06-20
Fix OnEndOfDay not fired - Undeterministic ScheduledEvent triggering

Martin-Molinero:
#### Description

- Adding `IRealTimeHandler.OnSecuritiesChanged()` will be used to update
the `OnEndOfDay` security related scheduled events
- Adding `BaseRealTimeHandler.cs` to reduce code duplication in the
`Backtesting` and `LiveTrading` `RealTimeHandlers`
- Adding CSharp and Python regression tests
- Deprecating `OnEndOfDay()` callback because of two reasons, mainly
because Python does not support two methods with the same name, but also
because different assets have different market close times.
- `ScheduledEvents` set at the same time will now be deterministic. Adding unit test (_which fails in  master_)

**Performance tests StatefulCoarseUniverseSelectionBenchmark 2017 - 2018 - 250 securities (250 scheduled events added in this PR)**
`PR`
33.72 seconds at 57k data points per second. Processing total of 1,925,403 data points.
33.38 seconds at 58k data points per second. Processing total of 1,925,403 data points.
33.56 seconds at 57k data points per second. Processing total of 1,925,403 data points.
imagen

`master`
32.12 seconds at 60k data points per second. Processing total of 1,925,403 data points.
32.15 seconds at 60k data points per second. Processing total of 1,925,403 data points.
31.74 seconds at 61k data points per second. Processing total of 1,925,403 data points.

imagen

**Note: there is a ~5% performance hit related to triggering the 250 security related `OnEndOfDay` scheduled events.**
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2625
Closes https://github.com/QuantConnect/Lean/issues/2865
#### Motivation and Context

Fix bugs

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -325,6 +325,8 @@ namespace QuantConnect.Lean.Engine
                             security.IsTradable = false;
                         }
                     }
+
+                    realtime.OnSecuritiesChanged(timeSlice.SecurityChanges);
                 }
 
                 //Update the securities properties: first before calling user code to avoid issues with data
--- a/Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
@@ -1,0 +1,160 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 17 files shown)


jaredbroad:
Thank you Martin, we should make something which detects if the consumer uses the event handler. Few people do so it doesn't make sense to scan it each time.

Engine/AlgorithmManager.cs
```diff
@@ -325,6 +325,8 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                             security.IsTradable = false;
                         }
                     }
+
+                    realtime.OnSecuritiesChanged(timeSlice.SecurityChanges);
```
jaredbroad: 👍

Issue #3296, opened 2019-06-11
Custom Data (BaseData) With Relationship To Tickers Does Not Support MapFiles


gsalaz98:
#### Expected Behavior

Consider the following: 

What would happen if `ENRN` gets delisted on 2001-12-27T00:00:00 UTC and another company takes over the ticker on 2019-01-01T00:00:00 UTC, which then undergoes a renaming event on 2019-02-01T00:00:00 UTC and the ticker is now named `GRR`? 

Additionally, what would happen if on 2019-03-01T00:00:00 UTC, another company takes over the ticker `ENRN`?

The intended behavior is that we would receive updates for `GRR`, and not `ENRN` after the renaming event. The user would have to re-add `ENRN` in order to get updates for ticker `ENRN` after the ticker had been renamed.

#### Actual Behavior

The user will keep receiving updates for the symbol `ENRN` after the symbol has been renamed from `ENRN` to `GRR`, and not be able to track the rename event of the ticker. In addition, if we wanted to access historical `GRR` data, we would load data for another company with the ticker `GRR` from the past, even though the two companies would have no association with each other whatsoever.

#### Potential Solution

Modify `BaseData` and all classes that derive from `BaseData` to include a field whether the data instance should be mapped. This ensures consistency in the codebase, and opens up the possibility for new asset classes and custom data to easily implement the mapping system.

#### Reproducing the Problem

An example algorithm that demonstrates the issue and goes into depth regarding the issue at hand can be found at [here](https://github.com/gsalaz98/LeanInternalsDocumentation/blob/master/LEAN%20Data%20and%20DataFeeds%20-%20Custom%20Data%20Mapping%20Issue/00%20-%20Introduction.md)

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3338, opened 2019-06-21
Add Mapfile Support For Custom Data Sources

gsalaz98:
#### Description

Implements the ability to use map files for custom data. 

This means that you can now apply rename events to custom data and be alerted when the symbol changes, and have the new updated symbol track the symbol at a given time T.

An example:
```
      AOL     ----------------------->     TWX      

   2003-10-15                           2003-10-16
```

Previously, if you added your symbol as "TWX" (the current stock ticker), you would only receive data points for "TWX" before 2003-10-16. But now, when you add "TWX", through the map file system, we load data for "AOL" before 2003-10-16, and rename the symbol on 2003-10-16 to "TWX" in your backtest containing custom data. 

This feature is opt-in via the `SubscriptionDataConfig.MapFileTypes` list. You can add your class to this list in order to enable mapping for your source of custom data

#### Related Issue

#3296 
#### Motivation and Context

Fulfills a requirement for sources of custom data based on equities

Yes. The tutorial teaching how to add custom data should consider adding a section detailing how to add a custom data class to the supported mapping (rename event) types.

New regression tests and unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
@@ -13,13 +13,13 @@
  * limitations under the License.
 */
 
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Data.Custom.SEC;
-using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
-using QuantConnect.Orders.Fees;
 
 namespace QuantConnect.Algorithm.CSharp
 {
@@ -38,8 +38,7 @@ namespace QuantConnect.Algorithm.CSharp
     {
         private Symbol _symbol;
         private bool _changedSymbol;
-        private bool _properSymbolBeforeRename;
-        private bool _properSymbolAfterRename;
+        private Dictionary<DateTime, string> _tickers = new Dictionary<DateTime, string>();
 
         /// <summary>
         /// Ticker we use for testing
@@ -75,19 +74,18 @@ namespace QuantConnect.Algorithm.CSharp
 
             foreach (var report in slice.Get<SECReport8K>())
             {
-                if (!_properSymbolBeforeRename)
+                var ticker = report.Key.Value;
+                var date = Time.Date;
+
+                if (date == new DateTime(2001, 1, 26) || date == new DateTime(2003, 10, 22))
                 {
-                    _properSymbolBeforeRename = report.Key.Value == "AOL" && Time < new DateTime(2003, 10, 16);
-                }
-                if (!_properSymbolAfterRename)
-                {
-                    _properSymbolAfterRename = report.Key.Value == "TWX" && Time >= new DateTime(2003, 10, 16);
+                    _tickers[date] = ticker;
                 }
 
-                Log($"{Time} - Received 8-K report for {report.Key.Value}");
+                Log($"{Time} - Received 8-K report for {ticker}");
             }
         }
-        
+
         /// <summary>
         /// Final step of the algorithm
         /// </summary>
@@ -97,13 +95,19 @@ namespace QuantConnect.Algorithm.CSharp
             {
                 throw new Exception("The ticker did not rename throughout the course of its life even though it should have");
             }
-            if (!_properSymbolBeforeRename)
+
+            var expectedTickers = new Dictionary<DateTime, string>
             {
-                throw new Exception("The SEC report data never renamed to its old ticker");
-            }
-            if (!_properSymbolAfterRename)
+                { new DateTime(2001, 1, 26), "AOL" },
+                { new DateTime(2003, 10, 22), "TWX" },
+            };
+
+            // Check for dictionary equality: https://stackoverflow.com/a/3804852
+            if (_tickers.Count != expectedTickers.Count && _tickers.Except(expectedTickers).Any())
             {
-                throw new Exception("The SEC report data never renamed back to its present-day ticker");
+                Log($"Found: {JsonConvert.SerializeObject(_tickers, Formatting.None)}");
+                Log($"Expected: {JsonConvert.SerializeObject(expectedTickers, Formatting.None)}");
+                throw new Exception("SEC data event tickers do not match test case");
             }
         }
 
--- a/Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
@@ -1,0 +1,146 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Data.Custom.SEC;
+using QuantConnect.Data.Market;
+using QuantConnect.Interfaces;
+using QuantConnect.Orders.Fees;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm demonstrating use of map files with custom data
+    /// </summary>
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="custom data" />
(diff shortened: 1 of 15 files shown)


Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
```diff
@@ -0,0 +1,146 @@
...
+            if (slice.SymbolChangedEvents.ContainsKey(_symbol))
+            {
+                // Check to see if it was renamed on the 16th
+                _changedSymbol = Time.Date == new DateTime(2003, 10, 16);
+                Log($"{Time} - Ticker changed from: {slice.SymbolChangedEvents[_symbol].OldSymbol} to {slice.SymbolChangedEvents[_symbol].NewSymbol}");
+            }
+
+            foreach (var report in slice.Get<SECReport8K>())
+            {
+                if (!_properSymbolBeforeRename)
```
mchandschuh: This condition would allow the test to be wrong _and then_ right, and the flag will still be set as passing. Another option is to add these flags to a list and in `OnEndOfAlgorithm` confirm that all of them are correct. Another way is to add just `report.Key.Value` to a list (or dictionary) w/ the dates and perform the checks there.

Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// This is used by the regression test system to indicate which languages this algorithm is written in.
+        /// </summary>
+        public Language[] Languages { get; } = { Language.CSharp, Language.Python };
+
+        /// <summary>
+        /// This is used by the regression test system to indicate what the expected statistics are from running the algorithm
+        /// </summary>
+        public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>
+        {
+            {"Total Trades", "0"},
```
mchandschuh: In general, I think it's best to create at least one trade... I forget _why_ this was the previous guidance, but I do remember it being the recommendation at one point in time.

Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
```diff
@@ -0,0 +1,146 @@
...
+        /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
+        /// </summary>
+        public override void Initialize()
+        {
+            SetStartDate(2001, 1, 1);
+            SetEndDate(2003, 12, 31);
+            SetCash(100000);
+
+            // AOL renames to TWX in 2003
+            _symbol = AddData<SECReport8K>(Ticker, Resolution.Daily).Symbol;
```
mchandschuh: I think it makes sense to add some of the other custom data types to this regression test so we have better coverage, especially since some of the other types are much older, and so they may behave differently than expected.

Common/SecurityIdentifier.cs
```diff
@@ -350,6 +350,35 @@ public SecurityIdentifier(string symbol, ulong properties, SecurityIdentifier un
...
+        /// <summary>
+        /// This implementation uses the configured <see cref="IMapFileProvider"/> via the <see cref="Composer.Instance"/>
+        /// </summary>
+        /// <param name="symbol">The symbol as it is known today</param>
+        /// <param name="market">The market</param>
+        /// <param name="securityType">The security type of the symbol</param>
+        /// <param name="mapSymbol">Specifies if symbol should be mapped using map file provider</param>
+        /// <param name="mapFileProvider">Specifies the IMapFileProvider to use for resolving symbols, specify null to load from Composer</param>
+        /// <returns>A new <see cref="SecurityIdentifier"/> representing the specified symbol today</returns>
+        public static SecurityIdentifier GenerateWithFirstDate(string symbol, string market, SecurityType securityType, bool mapSymbol = true, IMapFileProvider mapFileProvider = null)
```
mchandschuh: We might want to make this private. I view this 'first date' behavior as an internal implementation detail of the `SecurityIdentifier` and as such, shouldn't be publicly exposed unless absolutely necessary.

Common/SecurityIdentifier.cs
```diff
@@ -350,6 +350,35 @@ public SecurityIdentifier(string symbol, ulong properties, SecurityIdentifier un
...
+        {
+            var firstDate = DefaultDate;
+            Func<DateTime, string, string, SecurityIdentifier> GenerateSid;
+
+            switch (securityType)
```
mchandschuh: This is cute, but seems unnecessary. This check could be moved to the end of the method and be replaced w/ the actual invocation of the method instead of having to store it in a `Func<DateTime, string, string SecurityIdentifier>`

In fact, this entire `switch` statement can be moved to the end of the method and each `case` would turn into `return GenerateXXX( params )`. I think this would be more explicit and remove this layer of indirection.

Engine/DataFeeds/Enumerators/Factories/CorporateEventEnumeratorFactory.cs
```diff
@@ -80,34 +80,17 @@ public static class CorporateEventEnumeratorFactory
             MapFileResolver mapFileResolver)
         {
             var mapFileToUse = new MapFile(config.Symbol.Value, new List<MapFileRow>());
+            var hasUnderlying = config.Symbol.HasUnderlying;
 
-            // load up the map and factor files for equities
-            if (!config.IsCustomData && config.SecurityType == SecurityType.Equity)
+            // load up the map and factor files for equities, options, and custom data
+            if (config.UsesMapFiles)
```
mchandschuh: love it! 👍

mchandschuh:
@gsalaz98  - This is awaiting response to the existing review comments.

Engine/DataFeeds/Enumerators/Factories/CorporateEventEnumeratorFactory.cs
gsalaz98: This might be a bit problematic in that Options will be allowed through the gate, even though in the past it seemed to be only for equities. Either we're introducing a bug unknowingly, or fixing one. I'd like to review this and the following conditions further to ensure that the current implementation was intended.

`Engine/DataFeeds/SubscriptionDataReader.cs ` - Lines 225-226 
`Engine/HistoricalData/SubscriptionDataReaderHistoryProvider.cs` - Line 111

Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
gsalaz98: I agree, but I don't think that any of the other data sources bundled with LEAN at the moment have support for mapping except SEC and Psychsignal (which we can't include)

mchandschuh: Very well, I suppose I was thinking we could add a quandl type that could be mapped or just define a new custom data type specifically for this regression algorithm that does the AOL->TWX mapping. It's always nice for users to be able to see how to use a feature if they need to implement their own.

Common/SecurityIdentifier.cs
mchandschuh: Yea, in general you want to test a type's public API so that it's internals can change w/out having to update everyone else. If you find yourself _needing_ to test internal logic, it's usually a [code smell](https://en.wikipedia.org/wiki/Code_smell) indicating that the type is too complicated and some of the internal logic that requires testing can be extracted into it's own type complete with it's own public API.

Engine/DataFeeds/Enumerators/Factories/CorporateEventEnumeratorFactory.cs
mchandschuh: Great catch!!

Options are a whole 'nother beast. There is an `OptionRenameRegressionAlgorithm` which seems to imply that we've thought about this in the past.

This makes me somewhat concerned about potentially other locations that may suffer from a similar situation.

mchandschuh:
> Still a few more issues I managed to catch during my own code review. I'd like some feedback on the `SecurityType.Option` conditional statements that existed before @mchandschuh - git blame shows you as the author of two of these blocks, another one Stefano, and another user by the name of `quant1729`.

What are the commit ids of all of these? I can look into them.

gsalaz98:
Here's a link to the test algorithm I used to generate the following results:

https://gist.github.com/gsalaz98/4c131b1884955732f57f72a54c578275

And here are the differing History requests that get generated as a result of our changes:

* Master: https://gist.github.com/gsalaz98/13ae6ca14459fc1bcf47f0a121c7ade4
* This PR (rebased with master as well; similar output): https://gist.github.com/gsalaz98/39e4092beab48ce3698790f7c9a2325a

In master, when we submit a history request using the algorithm above, we get a few errors describing how we failed to load data for `FOXA` on `2013-06-28` even though `FOXA` didn't exist until `2013-06-29` or later. This most likely means that because we don't apply any sort of mapping, we will fail to load data for `NWSA` if a history request is submitted on the day of the rename. That is my understanding of the situation. If I'm wrong, please feel free to correct me.

I'm unsure which is the intended behavior, but some behavior we're not testing changed as a result of this PR. If I had to take a guess, I'd say this PR fixes a bug in History requests for options data since we previously were not applying any sort of mapping to options symbols, but we were for equities.

Let me know your thoughts y'all. CC @Martin-Molinero @StefanoRaggi since git blame shows you two have worked on `SubscriptionDataReaderHistoryProvider` before

jaredbroad:
Sperceeded by #3521

PR #3341, opened 2019-06-24
Add missing GDAX currency pairs

voidale:
Added missing currencies (XRP,EOS,XLM,ETC,ZRX)


voidale:
@StefanoRaggi any chance for review?

StefanoRaggi:
@voidale Thanks for the PR!
I'm just wondering if we could also add the USDC pairs, as PR #3000 has already been merged?

@mchandschuh For live trading, it would definitely be nice to have the updated list from symbols from the brokerage, also including the symbol properties (such as minimum lot size, minimum price increment, etc.). 
Since this information is also required in backtesting, maybe we could have some automated tool to periodically refresh the symbol properties database from the brokerage -- out of scope for this PR :)

PR #3344, opened 2019-06-25
Implements QCAlgorith.IsTraining and QCAlgorithm.Train

### LINKED ISSUE #3319: Custom Timeouts for Training Models
With the rise of AI it is becoming more common to do expensive computations in a strategy which might take a while. LEAN currently has a limit of 10 minutes per loop of the algorithm designed to prevent an algorithm locking up and crashing without user awareness.

We need a way to train these models for longer periods.

#### Expected Behavior
Some mechanism for training models -- with a maximum timeout controlled by the Controls class. 

#### Actual Behavior
No safe way to do computations longer than 10min. 

#### Potential Solution
Helper method:
QCA.Train( action, defaultTimeout = 30min, callback=null, synchronous=true)

or "semaphore" flag for 1x increases in the timeout:
QCA.LoopTimeout = TimeSpan.FromMinutes(30);

#### Reproducing the Problem
Run any backtest with long processing intra-loop. 

#### Checklist 

AlexCatarino:
#### Description
`QCAlgorithm.IsTraining` is a flag that signals that work is being performed by the `QCAlgorithm.Train`. The `AlgorithmManager` is aware of this special algorithm state and use a larger loop timeout set by `Controls`.

#### Related Issue
Closes #3319 

#### Motivation and Context
With the rise of AI it is becoming more common to do expensive computations in a strategy which might take a while. LEAN currently has a limit of 10 minutes per loop of the algorithm designed to prevent an algorithm locking up and crashing without user awareness.

Yes. We need to add a section that explains its usage.

Unit tests and example algorithm.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/TrainExampleAlgorithm.cs
+++ b/Algorithm.CSharp/TrainExampleAlgorithm.cs
@@ -1,0 +1,69 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using QuantConnect.Data;
+using System;
+using System.Threading;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This example shows how we can execute a method that takes longer than Lean timeout limit
+    /// This feature is useful for algorithms that train models
+    /// </summary>
+    /// <meta name="tag" content="using quantconnect" />
+    public class TrainExampleAlgorithm : QCAlgorithm
+    {
+        /// <summary>
+        /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
+        /// </summary>
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 07);  //Set Start Date
+            SetEndDate(2013, 10, 11);    //Set End Date
+
+            AddEquity("SPY");
+
+            Schedule.On(
+                DateRules.EveryDay("SPY"),
+                TimeRules.AfterMarketOpen("SPY", 10),
+                () => {
+                    Train(
+                        SleepTraining,
+                        TimeSpan.FromSeconds(20),
+                        () => { Debug($"Callback called at {Time:O}"); }
+                    );
+                }
+            );
+        }
+
+        private void SleepTraining()
+        {
+            Thread.Sleep(10000);
+            Debug($"Portfolio Invested: {Portfolio.Invested}.");
+        }
+
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
+        {
+            if (Portfolio.Invested) return;
+
+            SetHoldings("SPY", 1);
+        }
+    }
+}
--- a/Algorithm.CSharp/TrainExampleAlgorithm.cs
+++ b/Algorithm.CSharp/TrainExampleAlgorithm.cs
@@ -15,7 +15,6 @@
 
 using QuantConnect.Data;
 using System;
-using System.Threading;
 
 namespace QuantConnect.Algorithm.CSharp
 {
@@ -26,6 +25,8 @@ namespace QuantConnect.Algorithm.CSharp
     /// <meta name="tag" content="using quantconnect" />
     public class TrainExampleAlgorithm : QCAlgorithm
     {
+        private string _piString;
+
         /// <summary>
         /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
         /// </summary>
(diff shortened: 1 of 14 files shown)


Tests/Common/Util/ExtensionsTests.cs
```diff
@@ -771,6 +772,47 @@ public void DecimalTruncateTo3DecimalPlaces()
...
+
+            // Callback throws exception
+            var task3 = action.TimeoutAfterAndContinue(TimeSpan.FromMilliseconds(2000), divideByZeroAction);
+            task3.ContinueWith(x =>
+                {
+                    Assert.IsTrue(x.IsFaulted);
+                    Assert.AreEqual(x.Exception?.InnerException?.GetType(), typeof(DivideByZeroException));
+                }
+            );
+            Assert.Throws<AggregateException>(() => task3.Wait());
```
Martin-Molinero: I think a nice unit test policy is to divide the different assertions into specific unit tests

Algorithm.CSharp/TrainExampleAlgorithm.cs
```diff
@@ -0,0 +1,120 @@
...
+using System;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This example shows how we can execute a method that takes longer than Lean timeout limit
+    /// This feature is useful for algorithms that train models
+    /// </summary>
+    /// <meta name="tag" content="using quantconnect" />
+    public class TrainExampleAlgorithm : QCAlgorithm
```
Martin-Molinero: Nice example algorithm! I think it would be nice to add an algorithm using this feature into the regression algorithms, `IRegressionAlgorithmDefinition`, so we test it continuously

Algorithm/QCAlgorithm.cs
```diff
@@ -2144,6 +2149,44 @@ public string Download(string address, IEnumerable<KeyValuePair<string, string>>
...
+        public void Train(Action action, TimeSpan? timeout = null, Action callback = null, bool synchronous = true)
+        {
+            IsTraining = true;
+
+            if (synchronous)
+            {
+                action();
+                callback?.Invoke();
+                IsTraining = false;
+                return;
```
Martin-Molinero: I think there might be a small race condition until we finish running this time loop, where the `CurrentTimeStepElapsed` is re freshed and the the `Isolator` checking on it?

Algorithm/QCAlgorithm.cs
```diff
@@ -2144,6 +2149,44 @@ public string Download(string address, IEnumerable<KeyValuePair<string, string>>
...
+        /// In synchronous mode, execute the work and the callback in sequence while <see cref="IsTraining"/> flag is true
+        /// to signal the algorithm is in training mode.
+        /// In asynchronous mode, queues the specified work to run on the thread pool
+        /// that must complete within a specified time interval.
+        /// If the work completes within the interval, a new action is invoked if defined.
+        /// </summary>
+        /// <param name="action">The work to execute asynchronously</param>
+        /// <param name="timeout">The time span to wait before completing the returned task, or TimeSpan.FromMilliseconds(-1) to wait indefinitely.</param>
+        /// <param name="callback">The work to execute synchronously after the first work is completed</param>
+        /// <param name="synchronous">Queue the work synchronously (true). Otherwise we'll block until it is completed</param>
```
Martin-Molinero: Wondering if for backtesting it would make sense **not** to run `synchronous`? Maybe could check `QCAlgorithm.LiveMode` flag and throw?

Common/Extensions.cs
```diff
@@ -1445,5 +1445,39 @@ public static string RemoveFromEnd(this string s, string ending)
...
+                    {
+                        throw new TimeoutException(
+                            $"The operation has timed out after {delay.Value.TotalMinutes:F2} minutes"
+                        );
+                    }
+                }
+
+                timeoutCancellationTokenSource.Cancel();
+
+                await task;
```
Martin-Molinero: Wondering if at this stage wouldn't `task` already be completed?

Algorithm/QCAlgorithm.cs
mchandschuh: I would think backtesting you would especially want synchronous execution, otherwise by the time it's done training, months worth of time may have passed. I was thinking live mode you would prefer asynchronous training so you could continue to perform other tasks while training is running.

Common/Extensions.cs
```diff
@@ -1445,5 +1445,39 @@ public static string RemoveFromEnd(this string s, string ending)
...
+        /// Queues the specified work to run on the thread pool that must complete within a specified time interval.
+        /// If the work completes within the interval, a new action is invoked if defined.
+        /// </summary>
+        /// <param name="action">The work to execute asynchronously</param>
+        /// <param name="delay">The time span to wait before completing the returned task, or TimeSpan.FromMilliseconds(-1) to wait indefinitely.</param>
+        /// <param name="continuationAction">The work to execute synchronously after the first work is completed</param>
+        /// <exception cref="TimeoutException">The exception that is thrown when the time allotted for the asynchronously work has expired.</exception>
+        public static async Task TimeoutAfterAndContinue(this Action action, TimeSpan? delay, Action continuationAction = null)
+        {
+            using (var timeoutCancellationTokenSource = new CancellationTokenSource())
```
mchandschuh: You can provide the delay to the `CancellationTokenSource` constructor, and after the delay has passed, it will automatically flag it as cancellation requested. This would also have the benefit of removing the `delayTask`

Common/Extensions.cs
```diff
@@ -1445,5 +1445,39 @@ public static string RemoveFromEnd(this string s, string ending)
...
+            {
+                var task = Task.Run(action, timeoutCancellationTokenSource.Token);
+                var delayTask = Task.Delay(delay ?? TimeSpan.FromMilliseconds(-1), timeoutCancellationTokenSource.Token);
```
mchandschuh: Instead of using `TimeSpan.FromMilliseconds(-1)`, prefer usage of `Timeout.InfiniteTimeSpan` as it's more explicit as to the intention. Upon first reading, it looked like a bug (negative time span).

Common/Extensions.cs
mchandschuh: Correct, since we did `await Task.WhenAll(task, delayTask)`

AlexCatarino: If the task is completed within timeout, it may have faulted or been canceled. So we need to re-await the task so that any exceptions/cancellation is rethrown. Ref.: https://stackoverflow.com/a/11191070
I will add a comment to document the re-wait.

Issue #3347, opened 2019-06-26
Split SEC Data Converter and Improve Logging


gsalaz98:
#### Expected Behavior

We should only output issues regarding the parsing experience

The SEC Data Downloader and Converter should be separate ToolBox applications.
#### Actual Behavior

Multiple duplicate "SEC form type XYZ is not supported at this time" errors are outputted

The SEC Data Downloader and Converter are a single ToolBox application
#### Potential Solution

Remove unsupported form type error message
#### Reproducing the Problem

Run SEC Data Downloader/Converter ToolBox application

#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

gsalaz98:
Closed by #3348

PR #3348, opened 2019-06-26
Improves SEC Logging Experience and Splits SEC Downloader/Converter

gsalaz98:
#### Description

Split SECDataDownloader and SECDataConverter to two ToolBox applications

Improve logging experience

Move data unpacking folder to `Path.GetTempPath()` to prevent pollution of raw data folder

Add detection for &lt;PAPER&gt; submissions

Continue instead of returning on &lt;CONFIRMING-COPY&gt; entries

#### Related Issue

#3347 
#### Motivation and Context

Improve data processing experience

Yes - updated ToolBox --help output

Ran downloader and converter locally
#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/SECDataDownloader/SECDataConverter.cs
+++ b/ToolBox/SECDataDownloader/SECDataConverter.cs
@@ -24,8 +24,6 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
-using ICSharpCode.SharpZipLib.GZip;
-using ICSharpCode.SharpZipLib.Tar;
 using Newtonsoft.Json;
 using QuantConnect.Data;
 using QuantConnect.Data.Auxiliary;
@@ -183,9 +181,9 @@ namespace QuantConnect.ToolBox.SECDataDownloader
                 }
                 
                 // SEC data list contains a null value in the ticker.txt file
-                if (!string.IsNullOrEmpty(ticker))
+                if (!string.IsNullOrWhiteSpace(ticker))
                 {
-                    symbol.Add(tickerCik[0]);
+                    symbol.Add(ticker);
                 }
             }
 
@@ -200,11 +198,11 @@ namespace QuantConnect.ToolBox.SECDataDownloader
                 List<string> symbol;
                 if (!CikTicker.TryGetValue(companyCik, out symbol))
                 {
-                    symbol = new List<string>() { companyTicker };
+                    symbol = new List<string>();
                     CikTicker[companyCik] = symbol;
                 }
                 // Add null check just in case data comes malformed
-                else if (!symbol.Contains(companyTicker) && !string.IsNullOrEmpty(companyTicker))
+                if (!symbol.Contains(companyTicker) && !string.IsNullOrWhiteSpace(companyTicker))
                 {
                     symbol.Add(companyTicker);
                 }
@@ -216,7 +214,7 @@ namespace QuantConnect.ToolBox.SECDataDownloader
             {
                 if (Holidays.Contains(processingDate))
                 {
-                    Log.Trace("SECDataConverter.Process(): File is missing but it was expected. Nothing to do.");
+                    Log.Trace("SECDataConverter.Process(): File is missing, but we expected it to be missing. Nothing to do.");
                     return;
                 }
                 throw new Exception($"SECDataConverter.Process(): Raw data {remoteRawData} not found. No process can be done.");
@@ -376,7 +374,7 @@ namespace QuantConnect.ToolBox.SECDataDownloader
                         if (!tickerMapFile.Any()) continue;
 
                         var tickerReports = Reports.GetOrAdd(
-                            tickerMapFile.GetMappedSymbol(processingDate),
+                            ticker,
                             _ => new ConcurrentDictionary<DateTime, List<ISECReport>>()
                         );
                         var reports = tickerReports.GetOrAdd(
--- a/ToolBox/SECDataDownloader/SECDataConverter.cs
+++ b/ToolBox/SECDataDownloader/SECDataConverter.cs
@@ -150,7 +150,6 @@ namespace QuantConnect.ToolBox.SECDataDownloader
         /// </summary>
         /// <param name="rawSource">Source of raw data</param>
         /// <param name="destination">Destination of formatted data</param>
-        /// <param name="tickerFolder">Known ticker folder (should be low resolution tick folder)</param>
         public SECDataConverter(string rawSource, string destination)
         {
             RawSource = rawSource;
@@ -160,8 +159,7 @@ namespace QuantConnect.ToolBox.SECDataDownloader
         /// <summary>
         /// Converts the data from raw format (*.nz.tar.gz) to json files consumable by LEAN
         /// </summary>
-        /// <param name="startDate">Starting date to start process files</param>
-        /// <param name="endDate">Ending date to stop processing files</param>
+        /// <param name="processingDate">Date to process SEC filings for</param>
         public void Process(DateTime processingDate)
         {
             // Process data into dictionary of CIK -> List{T} of tickers
@@ -367,7 +365,8 @@ namespace QuantConnect.ToolBox.SECDataDownloader
                     }
 
                     // Default to company CIK if no known ticker is found.
-                    // If we don't find a known equity in our list, the equity is probably not worth our time
+                    // If the equity is not does not resolve to a map file or
+                    // it is not found in the map files, we skip writing it.
                     foreach (var ticker in tickers)
                     {
                         var tickerMapFile = mapFileResolver.ResolveMapFile(ticker, processingDate);
@@ -380,6 +379,7 @@ namespace QuantConnect.ToolBox.SECDataDownloader
                         // Map the current ticker to the ticker it was in the past using the map file system
                         var mappedTicker = tickerMapFile.GetMappedSymbol(processingDate);
                         
+                        // If no suitable date is found for the symbol in the map file, we skip writing the data
                         if (string.IsNullOrEmpty(mappedTicker))
                         {
                             Log.Trace($"SECDataConverter.Process(): {processingDate} - Failed to find mapped symbol for ticker: {ticker}");
@@ -430,7 +430,6 @@ namespace QuantConnect.ToolBox.SECDataDownloader
         /// </summary>
         /// <param name="reports">List of SEC Report objects</param>
         /// <param name="ticker">Symbol ticker</param>
-        /// <param name="destination">Folder to write the reports to</param>
         public void WriteReport(List<ISECReport> reports, string ticker)
         {
             var report = reports.First();
@@ -461,7 +460,6 @@ namespace QuantConnect.ToolBox.SECDataDownloader
         /// </summary>
         /// <param name="report">SEC report <see cref="BaseData"/> instance</param>
         /// <param name="companyCik">Company CIK to use to lookup filings for</param>
-        /// <remarks>This method caches the results on a per company basis (by CIK), so subsequent lookups for the publication date of the same equity by CIK will be fast</remarks>
         public void GetPublicationDate(ISECReport report, string companyCik)
         {
             Dictionary<string, DateTime> companyPublicationDates;
--- a/ToolBox/SECDataDownloader/SECDataConverter.cs
+++ b/ToolBox/SECDataDownloader/SECDataConverter.cs
@@ -371,10 +371,23 @@ namespace QuantConnect.ToolBox.SECDataDownloader
                     foreach (var ticker in tickers)
                     {
                         var tickerMapFile = mapFileResolver.ResolveMapFile(ticker, processingDate);
-                        if (!tickerMapFile.Any()) continue;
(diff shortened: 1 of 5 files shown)


ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -66,19 +60,18 @@ public class SECDataConverter
...
+        private DirectoryInfo _tickerFolder;
 
-        public SECDataConverter(string rawSource, string destination, string knownEquityFolder)
+        public SECDataConverter(string rawSource, string destination, string tickerFolder)
         {
             RawSource = rawSource;
             Destination = destination;
-            
+
+            _tickerFolder = new DirectoryInfo(tickerFolder);
```
mchandschuh: Is there any concern with this directory not existing?

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -100,7 +93,7 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
                         symbol.Add(tickerCik[0]);
                     }
                 );
-            
+
```
mchandschuh: Unrelated to this PR, but this constructor does _a lot_ of work. Further, it uses a pattern I would try to stay away from where we use the `ToList()` LINQ method simply so we can use the `ForEach` method instead of using `for each`. The difference is bigger than you may expect. The `ToList()` method will obviously require complete enumeration of the underlying enumerable along with the requisite allocations. Give the `rankAndFileData` may not be large, this may not be critical. In addition, since we're reading from disk, we can enumerate the lines one by one instead of using `Split('\n')` -- _further_ we're not guaranteed that splitting on `\n` characters is sufficient. Windows will use CRLF `\n\r` as line break, so all of these lines would have  an initial `\r` character.

Instead, it's better practice to stream the file line-by-line using something like this:
``` csharp
foreach (var line in File.ReadLines(Path.Combine(RawSource, "cik-ticker-mappings-rankandfile.txt"))
{
    var tickerInfo = line.Split('|');
    // process  as before
}
```

This removes a couple of issues, the biggest of which is using `string.Split` on an entire file's worth of data. The `Split` function is notoriously memory inefficient as well as potentially _incorrect_ (windows v linux), especially when streaming it from disk using `File.ReadLines` is simpler, more efficient, and _arguably_ more readable. This also removes the unnecessary allocation of the intermediate list just to make the `ForEach` method available _and also_ completely removes the need for putting the entire file in memory.

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -123,189 +116,189 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
...
+            Log.Trace($"SECDataConverter.Process(): Copying raw data locally...");
+
+            var localRawData = remoteRawData.CopyTo(Path.Combine(Path.GetTempPath(), remoteRawData.Name));
+            var extractDataPath = Path.Combine(Path.GetTempPath(), formattedDate);
+
+            Log.Trace($"SECDataConverter.Process(): Extract raw data...");
+
+            using (var data = localRawData.OpenRead())
+            {
+                using (var archive = TarArchive.CreateInputTarArchive(new GZipInputStream(data)))
```
mchandschuh: I would have expected the `Compression` library to support `.tar.gz`/`.tgz` files. We may want to add this functionality to the `Compression.cs` class so we don't end up duplicating this code elsewhere.

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -123,189 +116,189 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
...
+                    Directory.CreateDirectory(extractDataPath);
+                    archive.ExtractContents(extractDataPath);
+
+                    Log.Trace($"SECDataConverter.Process(): Extracted SEC data to path {extractDataPath}");
+                }
+            }
+
+            var tickerList = _tickerFolder.EnumerateDirectories().AsParallel()
+                .Where(d => d.EnumerateFiles($"{formattedDate}*").Any())
+                .Select(d => d.Name);
```
mchandschuh: OK, so this might be unexpected, but this `tickerList` _enumerable_ is used in exactly one location, [line 266](https://github.com/QuantConnect/Lean/pull/3348/files#diff-415f3d17a5d03708870daca21d68f5e5R266) which is invoked inside a loop _and_ inside a `Parallel.ForEach` block (that's n^2 invocations. Add that to the fact that this is an _enumerable_ and each time you call `tickerList.Contains` it will re-enumerate the enumerable until it finds a matching element, and if no element is found, then it will enumerate the entire enumerable, _including_  enumeration of `EnumerateFiles` defines on [line 154](https://github.com/QuantConnect/Lean/pull/3348/files#diff-415f3d17a5d03708870daca21d68f5e5R154). The easy way to remedy this is to call `ToHashSet()` at the end which will cause the enumerable to be materialized. _Further_, any time you're using a collection _only_ to call it's `Contains` method, you should be using a `HashSet<T>` as it provides `O(1)` (constant time) lookups, vs `O(n)` for `IEnumerable<T>`.

So this loop w/ the contains call is currently executing on `O(n^3)`  !!!!!

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -123,189 +116,189 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
...
-                    // Only process files within start and end bounds
-                    if (currentDate < startDate || currentDate > endDate)
-                    {
-                        return;
-                    }
+                    // We need to escape any nested XML to ensure our deserialization happens smoothly
+                    var parsingText = false;
 
-                    using (var data = File.OpenRead(rawFile))
+                    foreach (var line in File.ReadLines(rawReportFilePath))
```
mchandschuh: Nice usage of `File.ReadLines` 👍 to stream the file line-by-line instead of sucking the whole thing up into RAM.

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -123,189 +116,189 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
...
+
+                    ISECReport report;
+                    try
+                    {
+                        report = factory.CreateSECReport(xmlText.ToString());
+                    }
+                    // Ignore unsupported form types for now
+                    catch (DataException)
+                    {
+                        return;
```
mchandschuh: Do we want to log this case at least? This same comment applies for some other exception handlers as well.

Thanks for providing the comment regarding the source of the `DateException` 👍

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -123,189 +116,189 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
...
+
+                    // First filer listed in SEC report is usually the company listed on stock exchanges
+                    var companyCik = report.Report.Filers.First().CompanyData.Cik;
+
+                    // Some companies can operate under two tickers, but have the same CIK.
+                    // Don't bother continuing if we don't find any tickers for the given CIK
+                    List<string> tickers;
+                    if (!CikTicker.TryGetValue(companyCik, out tickers))
+                    {
+                        return;
```
mchandschuh: Again, logging here may be nice, especially if this program will eventually be run as part of a daily update process -- it's nice to catch major decisions (such as not processing an item) in the logs so that in the event something unexpected happens elsewhere in the stack, we're able to figure out what this program did with that data.

Also, if the reason for not logging is that it's _too much_ -- we can hold these kinds of things in RAM and at the end write a summary to the log of what the result of processing each item was (or to some other summary specific files, though my preference would be the standard log file)

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -123,189 +116,189 @@ public SECDataConverter(string rawSource, string destination, string knownEquity
...
+                    }
+
+                    try
+                    {
+                        // There can potentially not be an index file present for the given CIK
+                        GetPublicationDate(report, companyCik);
```
mchandschuh: Are there any other side effects required by invoking this function, or are we really just checking if:
``` csharp
var indexFilePath = Path.Combine(RawSource, "indexes", $"{companyCik}.json");
if (File.Exists(indexFilePath))
{
    throw new Exception($"Submission not found for CIK: {cik}");
}
```

If we're doing this just to check if the file exists, let's just check it  directly and _not_ throw the exception. Exceptions should be reserved for exceptional cases and [not for control flow](https://web.archive.org/web/20140430044213/http://c2.com/cgi-bin/wiki?DontUseExceptionsForFlowControl).

ToolBox/SECDataDownloader/SECDataConverter.cs
gsalaz98: The raw amount of CIKs included with the SEC reports is in my opinion way too much to even contain in RAM. I'd say that for a good chunk of each day, there are companies that aren't listed on exchanges that submit filings with the SEC, and are useless in our use case.

On another note, we're operating at very high RAM capacity, and we don't want to incur any more memory usage. If we had infinite resources, I'd for sure consider logging every event. But given that this PR's branch name has `reduce-log-spam-secdataconverter`, I guess you can see where I was going with this 😛

gsalaz98:
This is finally ready for production. @Jay-Jay-D - please review my changes on Monday when you have the time.

I did some testing with 2001-01-26 for TWX remapping to AOL (i.e. writing files to `AOL/*` when we have initial ticker TWX), and it works perfectly 😄 

Note: We need PR #3338 to go through in order for this to finally go into production with my most recent commit.

Thank you for the review @mchandschuh , you've been really helpful 👍

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -65,247 +144,283 @@ public class SECDataConverter
...
-                        // tickerCik[0] = symbol, tickerCik[1] = CIK
-                        // Note that SEC tickers come lowercase, so we don't have to alter the ticker
-                        var cikFormatted = tickerCik[1].PadLeft(10, '0');
+                        var newTextLine = line;
+                        var currentTagName = GetTagNameFromLine(newTextLine);
+
+                        // This tag is present rarely in SEC reports, but is unclosed without value when encountered.
+                        // Verified by searching with ripgrep for "CONFIRMING-COPY"
+                        //
+                        // Sometimes, ASSIGNED-SIC contains no value and causes errors. Check to make sure that when
```
mchandschuh: Thanks for the _excellent_ documentation of these cases. 👍

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -65,247 +144,283 @@ public class SECDataConverter
...
-                );
 
-            KnownEquities = Directory.GetFiles(knownEquityFolder)
-                .Select(x => Path.GetFileNameWithoutExtension(x).ToLower())
-                .ToList();
-        }
+                        // Don't encode the opening tag
+                        if (currentTagName == "TEXT")
+                        {
+                            parsingText = true;
```
mchandschuh: nit - more succinctly as `parsingText = currentTagName == "TEXT"`

gsalaz98: I wish I could do so, but we need to ensure that when this advances to the next line, that the boolean value remains unchanged unless the tag encountered == `</TEXT>`. 

By doing as you've said, on the next line following a `<TEXT>` tag, the condition `parsingText` would be set to false, and we'd stop escaping the inputs.

ToolBox/SECDataDownloader/SECDataConverter.cs
```diff
@@ -17,39 +17,118 @@
...
-
+        
         /// <summary>
-        /// Max file size in bytes we're willing to parse. Default is 50Mb
+        /// Whitelist of dates to not throw errors on if the file is missing.
+        /// Most of the dates are either Veterans Day or Columbus Day, but
+        /// sometimes there exists random days where the SEC doesn't publish reports
         /// </summary>
-        public int MaxFileSize = 50000000;
+        public HashSet<DateTime> Holidays = new HashSet<DateTime>
```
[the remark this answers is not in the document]
gsalaz98: Thanks for the catch, updating it to check for USHolidays as well in my latest commit 👍

gsalaz98:
> 
> 
> Approving, but think we might want to make [this](https://github.com/QuantConnect/Lean/pull/3348/files#diff-415f3d17a5d03708870daca21d68f5e5R360) call to `GetPublicationDate` more explicit as to the intent. The comment says it's checking for a potentially corrupted index which is detected down a bit from this method's invocation and might be more explicit if the `File.Exists` check is performed in line here instead.

Sometimes, the file contains a binary blob before the initial JSON starting tag due to errors in either tar archiving or corruption during the download process. We can't know that it's corrupted up until we try to deserialize the file and it throws an error. 

I've moved the File.Exists check a few lines above the try...catch statement to stop using exceptions for control flow. I'd like to discuss more about this, since I'm having some troubles understanding what you're suggesting in the comment quoted above.

Issue #3213, opened 2019-05-20
Implement Psychsignal Custom Data


gsalaz98:
#### Expected Behavior

We should be able to use psychsignal custom data in algorithms and convert backfill data into a format usable by Lean
#### Actual Behavior

Does not exist
#### Potential Solution

Implement the custom data reader and converter
#### Reproducing the Problem

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [ ] I have provided detailed steps to reproduce the issue

gsalaz98:
Closed by #3349

PR #3349, opened 2019-06-26
Adds Psychsignal Data Downloader/Converter and Custom Data Implementation With Example Algorithm

gsalaz98:
#### Description

Adds psychsignal as an alternative data source. This includes an example algorithm in both python and C#, psychsignal data downloader/converters, and custom data implementation.
#### Related Issue

#3213 
#### Motivation and Context

Custom data efforts

Maybe - PsychSignal will be built-in to LEAN with this PR

Local testing, no new unit tests 😕 - Maybe can make the `PsychSignalSentimentAlgorithm` a regression test.
#### Types of changes

#### Checklist:


Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <returns></returns>
+        public override SubscriptionDataSource GetSource(SubscriptionDataConfig config, DateTime date, bool isLiveMode)
+        {
+            return new SubscriptionDataSource(
+                Path.Combine(
+                    Globals.DataFolder,
+                    "alternative",
+                    "psychsignal",
+                    $"{config.Symbol.Value.ToLower()}",
+                    $"{date:yyyyMMdd}.zip#{date:yyyyMMdd}.csv"
```
Martin-Molinero: I _think_ that if the zip just has one entry specifying it is not required https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/ZipDataCacheProvider.cs#L205

Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <returns>
+        ///     Instance of the T:BaseData object containing psychsignal specific data
+        /// </returns> 
+        public override BaseData Reader(SubscriptionDataConfig config, string line, DateTime date, bool isLiveMode)
+        {
+            try
+            {
+                var csv = line.Split(',');
+
+                var timestamp = new DateTime(date.Year, date.Month, date.Day).AddMilliseconds(
```
Martin-Molinero: I believe you could directly create the new `DateTime` instance with the milliseconds specified, avoiding an extra `DateTime` instance creation

Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
```diff
@@ -0,0 +1,137 @@
...
+ *
+*/
+
+using System;
+using System.Globalization;
+using System.IO;
+
+namespace QuantConnect.Data.Custom.PsychSignal
+{
+    public class PsychSignalSentimentData : BaseData
```
Martin-Molinero: Maybe could add some documentation on the `PsychSignalSentimentData` class, what it is, who is producing it, maybe a link?

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+                if (date != _date)
+                {
+                    Dispose();
+
+                    _date = date;
+                    _tempPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+                    _writer = new StreamWriter(_tempPath);
+                }
+
+                _writer.WriteLine(ToCsv(timestamp, csv.Skip(3)));
```
Martin-Molinero: Maybe would be nice to mention why its skipping 3?

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+            /// <summary>
+            /// Moves the temporary file containing data to the final path, deleting
+            /// any existing file to avoid conflicts when moving
+            /// </summary>
+            private void MoveTempFile()
+            {
+                var tickerDirectory = Path.Combine(_destinationDirectory.FullName, _ticker);
+                var writePath = Path.Combine(tickerDirectory, $"{_date:yyyyMMdd}.csv");
+
+                Directory.CreateDirectory(tickerDirectory);
```
Martin-Molinero: Maybe to reduce effort the `tickerDirectory` could be created at construction time?

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+            /// </summary>
+            /// <param name="ticker"></param>
+            /// <param name="date"></param>
+            /// <param name="destinationDirectory"></param>
+            public TickerData(string ticker, DateTime date, DirectoryInfo destinationDirectory)
+            {
+                _date = date;
+                _ticker = ticker;
+                // Using win32 Path.GetTempFileName can cause filename collisions
+                _tempPath = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
```
Martin-Molinero: Wondering why we need a temp intermediate file? Why can't be directly write into the final destination file? Might be worth documenting it, sorry if I missed it!

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// we will lower the <paramref name="endDateUtc" /> by one hour in order
+        /// to make sure that we only download complete, non-changing dataa
+        /// </summary>
+        /// <param name="startDateUtc">Starting date</param>
+        /// <param name="endDateUtc">Ending date</param>
+        public void Download(DateTime startDateUtc, DateTime endDateUtc)
+        {
+            if (startDateUtc < endDateUtc.AddDays(-15))
+            {
+                throw new ArgumentException("The starting date can only be at most 15 days from now");
```
Martin-Molinero: > "The starting date can only be at most 15 days **from now"**

I think it should say from end date?

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,149 @@
...
+                {
+                    // Psychsignal imposes very strict rate limits
+                    Thread.Sleep(10000);
+
+                    try
+                    {
+                        using (var client = new WebClient())
+                        {
+                            client.DownloadFile($"{_baseUrl}/replay?apikey={_apiKey}&update=1m&sources={_dataSource}&from={startDateUtc:yyyyMMddHH}&format=csv", rawDataPathTemp);
+                            File.Move(rawDataPathTemp, rawDataPath);
```
Martin-Molinero: Same question here about using temp intermediate paths, do we need this?

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+                {
+                    Compression.Zip(dataFile, deleteOriginal: true);
+                    Log.Trace($"PsychSignalDataConverter.CompressData(): Successfully compressed: {dataFile}");
+                }
+            }
+        }
+        
+        /// <summary>
+        /// Handle to a file so that we don't have to open and close it every time we want
+        /// to write to a file. This helps us speed up time spent processing massively.
```
Martin-Molinero: Nice! 👍

Algorithm.CSharp/PsychSignalSentimentRegressionAlgorithm.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    /// This example algorithm shows how to import and use psychsignal sentiment data.
+    /// </summary>
+    /// <meta name="tag" content="strategy example" />
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="custom data" />
+    /// <meta name="tag" content="psychsignal" />
+    /// <meta name="tag" content="sentiment" />
+    public class PsychSignalSentimentRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
```
Martin-Molinero: Love new regression algorithms! I think if we could add a couple of data points and set `CanRunLocally` to true it would be awesome

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
gsalaz98: Because writing a file is a process that can be interrupted, you can easily end up with corrupted data. However, because file moves are in most cases atomic operations, we will only end up with a final working version of the file.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
gsalaz98: PsychSignal only allows you to reference data from 15 hours from the current time

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
mchandschuh: another option is to create in a static constructor, or provide a static flag so we're not creating it every time we create a new base data type.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
mchandschuh: Do we want to ensure we delete the temporary file?

Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        /// <summary>
+        /// Bullish intensity as reported by psychsignal
+        /// </summary>
+        public decimal BullIntensity;
```
mchandschuh: I think the prevailing style in the code base is to use properties:
``` csharp
public decimal BullIntensity { get; set; }
```

Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
mchandschuh: yup, should be able to do:
``` csharp
var timestamp = date.AddMilliseconds(Convert.ToDouble(csv[0]));
```

mchandschuh: also, do we want to add `CultureInfo.InvariantCulture` to ensure compatibility?

Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
```diff
@@ -0,0 +1,137 @@
...
+                var csv = line.Split(',');
+
+                var timestamp = new DateTime(date.Year, date.Month, date.Day).AddMilliseconds(
+                    Convert.ToDouble(csv[0])
+                );
+                var bullIntensity = Convert.ToDecimal(csv[1], CultureInfo.InvariantCulture);
+                var bearIntensity = Convert.ToDecimal(csv[2], CultureInfo.InvariantCulture);
+                var bullMinusBear = Convert.ToDecimal(csv[3], CultureInfo.InvariantCulture);
+                var bullScoredMessages = Convert.ToInt32(csv[4]);
+                var bearScoredMessages = Convert.ToInt32(csv[5]);
```
mchandschuh: Why not include `CultureInfo.InvariantCulture` here?

Tests/ToolBox/PsychSignalDataTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace QuantConnect.Tests.ToolBox
+{
+    [TestFixture, Ignore("This test requires raw PsychSignal data")]
+    public class PsychSignalDataTests
+    {
+        [Test]
+        public void FileHourMatchesDataTime()
+        {
+            var rawPath = Path.Combine(Globals.DataFolder, "alternative", "psychsignal", "raw-psychsignal");
```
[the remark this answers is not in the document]
mchandschuh: and after adding a small amount of data, we can remove the `Ignore` attribute on [line 27](https://github.com/QuantConnect/Lean/pull/3349/files#diff-4ab3507ecefeacfaaab4d8b0b4863905R27)

Tests/ToolBox/PsychSignalDataTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            
+            foreach (var file in Directory.GetFiles(rawPath, "*.csv", SearchOption.TopDirectoryOnly).ToList())
+            {
+                var fileSplit = file.Split('_');
+
+                var hour = Convert.ToInt32(fileSplit[fileSplit.Length - 4]);
+
+                // Read one line of the file, and compare the hour of the day to the hour on the file
+                var line = File.ReadLines(file).ToList().Last().Split(',');
```
mchandschuh: This `ToList()` is not required. If we're just interested in the last line then we can keep it as `IEnumerable` so we don't need to read the _entire_ file into memory. With this file type it may be inconsequential, but with larger files it can cause an issue.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using QuantConnect.Logging;
+using QuantConnect.Util;
+
+namespace QuantConnect.ToolBox.PsychSignalDataConverter
+{
+    public class PsychSignalDataConverter
```
mchandschuh: This class may benefit from some class-level documentation.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+        /// <param name="knownTickerFolder">Directory where we source the ticker list</param>
+        public PsychSignalDataConverter(string sourceDirectory, string destinationDirectory, string knownTickerFolder)
+        {
+            _rawSourceDirectory = new DirectoryInfo(sourceDirectory);
+            _destinationDirectory = new DirectoryInfo(destinationDirectory);
+
+            _fileHandles = new Dictionary<string, TickerData>();
+            _knownTickers = Directory.GetFiles(knownTickerFolder, "*.zip").Select(Path.GetFileNameWithoutExtension).ToHashSet();
+
+            _destinationDirectory.Create();
```
mchandschuh: I'm torn on things like this in constructors. On one hand, it's nice to encapsulate things like this inside the class, but on the other hand, it's  typically best to avoid potentially exception-raise actions inside constructors. No need to change, just thinking (writing) out loud :)

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+        {
+            var file = File.ReadLines(sourceFilePath.FullName);
+
+            foreach (var line in file)
+            {
+                var csv = line.Split(',');
+                var ticker = csv[1].ToLower();
+                DateTime timestamp;
+
+                if (csv[0] == "SOURCE" || !_knownTickers.Contains(ticker) || !DateTime.TryParseExact(csv[2], @"yyyy-MM-dd\THH:mm:ss\Z", CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out timestamp))
```
mchandschuh: nit - this may be a little more readable if each condition is on it's own line:
``` csharp
if (csv[0] == "SOURCE"
  || !_knownTickers.Contains(ticker)
  || !DateTime.TryParseExact(csv[2], @"yyyy-MM-dd\THH:mm:ss\Z", CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out timestamp))
```

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+
+        /// <summary>
+        /// Flushes all open <see cref="StreamWriter"/> instances.
+        /// Should be called once the data has finished processing
+        /// </summary>
+        public void FlushAll()
+        {
+            foreach (var handle in _fileHandles.Values)
+            {
+                handle.Dispose();
```
mchandschuh: The method name `FlushAll()` and it's behavior _dispose_ all seems inconsistent. Also, after disposing all of the handles, we don't remove them  from the dictionary. If this is intended to simply be a flush operation, then lets change this to call a new method, `TickerData.Flush` and update `PsychSignalDataConverter` to implement `IDisposable` and that `Dispose` implementation can perform final disposal of the streams.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,270 @@
...
+            /// </summary>
+            /// <param name="timestamp">Event timestamp in UTC</param>
+            /// <param name="csv">CSV enumerable</param>
+            public void Append(DateTime timestamp, IEnumerable<string> csv)
+            {
+                var date = timestamp.Date;
+
+                if (date != _date)
+                {
+                    Dispose();
```
mchandschuh: Disposal is intended to be permanent, non-reversible action taken against on object. Perhaps `Dispose` here is really a `Recycle`. In general, once `Dispose` is called on an object, any further calls to the object have undefined behavior, but typically they will raise an `ObjectDisposedException`.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+            
+            // PsychSignal paginates data by hour
+            for (; startDateUtc < endDateUtc; startDateUtc = startDateUtc.AddHours(1))
+            {
+                var rawDataPath = Path.Combine(_rawDataDestination, $"{startDateUtc:yyyyMMdd_HH}_{_dataSource}.csv");
+                var rawDataPathTemp = Path.Combine(Path.GetTempPath(), $"{startDateUtc:yyyyMMdd_HH}_{_dataSource}.csv.tmp");
+                
+                // Don't download files we already have
+                if (File.Exists(rawDataPath))
```
mchandschuh: Is there any concern regarding new data in these files? I recall the data is updated hourly? -- also looks like the hours are included in the file path. Is there any chance of downloading a _partial_ hour? For example, consider the current time is 10:30AM, is it possible to request 10AM->11AM  and only get the first 30 minutes? If so, this check will prevent us from downloading the full hour's worth of data when it's available.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,149 @@
...
+                {
+                    continue;
+                }
+                
+                // Retry in case a download failed
+                for (var retries = 0; retries < MaxRetries; retries++)
+                {
+                    // Psychsignal imposes very strict rate limits
+                    Thread.Sleep(10000);
```
mchandschuh: Consider using the `RateGate` type to handle rate limits instead of sleeping the current thread.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,149 @@
...
+                            break;
+                        }
+                        if (response.StatusCode == HttpStatusCode.NotFound)
+                        {
+                            Log.Error("PsychSignalDataDownloader.Download(): Received an HTTP 404. Continuing...");
+                            break;
+                        }
+                        if (response.StatusCode == (HttpStatusCode) 429)
+                        {
+                            Log.Trace("PsychSignalDataDownloader.Download(): We are being rate limited. Retrying...");
```
mchandschuh: Would the sleep be better placed here instead of at the top of the loop? Also, for documentation purposes, it may be helpful to include the API's rat limiting rules.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
gsalaz98: Yes, it is possible to download a partial hour. However, try to make sure that doesn't happen with the following lines:

https://github.com/QuantConnect/Lean/blob/42980c5094999f0f4ca7d2f3f61ec97b7af95efb/ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs#L79-L82

But on second thought, it's a bit naive to assume that the top bound will always be `DateTime.UtcNow` - I'll probably have to do some refactoring here to avoid that issue. Thanks for pointing this out :smile:

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
Martin-Molinero: Right, my comment wasn't that clear but I was pointing at the fact that the code seems to asume `endDateUtc` is ~now? Maybe it should directly check start date against `UtcNow.AddDays(-15)`

Common/Data/Custom/PsychSignal/PsychSignalSentimentData.cs
gsalaz98: Agreed, but I did this for another reason: The cloning of the object would not successfully clone members who were implemented as properties. I didn't know this before I implemented the class, but just commenting here in case someone ever stumbles across their custom data variables being null. The solution to that problem is to implement the clone method for the custom data source.

Tests/ToolBox/PsychSignalDataTests.cs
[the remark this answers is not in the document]
gsalaz98: I'm unsure if we can include any test data at all unfortunately. 

@jaredbroad - would be nice to confirm if we could do this.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
gsalaz98: It seemed overkill to me; A simple thread sleeping is in my opinion easier to follow than creating a rategate instance and using it in a single-threaded context.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
gsalaz98: Every request, no matter if it's successful or not, requires that it be sent `n` seconds after the previous request, in which this case I've set `n` to 10 seconds. 

They don't document their API. I had to figure out an appropriate rate limit myself through trial and error

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
mchandschuh: For now it's single threaded :) but _also_ the `RateGate` is very explicit as to what the allowables rates are and is accompanied by all the documentation and battle-tested usages. The naive thread-sleeping ignores various things like processing time and downloading time, etc, leading this process to take a _little_ bit (agreed, likely negligibly) longer.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
mchandschuh: oh wow, don't you just _love_ undocumented APIs! 👍 

Still thinking a `RateGate` would be more explicit as to the allowable rate limits (empirically set at 1 call per 10 seconds ?)

Tests/ToolBox/PsychSignalDataTests.cs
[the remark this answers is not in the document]
mchandschuh: @gsalaz98 -- I checked with @jaredbroad and sadly we can't include this sample data. Also, let's move downloading of the raw data to somewhere else, and preferably clean/delete it following the test, perhaps using a `[SetUp]` and `[TearDown]` nunit attributes.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
mchandschuh: there's a comment here saying we're ignoring the first three entries but no reasoning as to _why_ we're ignoring the first three lines. I suspect headers and perhaps copyrights/file information?

Algorithm.CSharp/PsychSignalSentimentRegressionAlgorithm.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    Liquidate(_symbol);
+                }
+            }
+        }
+
+        /// <summary>
+        /// This is used by the regression test system to indicate if the open source Lean repository has the required data to run this algorithm.
+        /// </summary>
+        public bool CanRunLocally { get; } = false;
```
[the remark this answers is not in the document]
gsalaz98: The data I added is just dummy data with no data, just placeholders. I can discuss more about my reasoning why in Slack if you wish, but it's unfortunate that we can't add data and run this regression algorithm locally.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,376 @@
...
+            Log.Trace($"PsychSignalDataConverter.Convert(): Finished converting {sourceFilePath.FullName}");
+        }
+
+        /// <summary>
+        /// Filter raw files to be inclusive within the date range specified and converts those files.
+        /// </summary>
+        /// <param name="startDateUtc">Date to start parsing files from</param>
+        /// <param name="endDateUtc">Date to stop parsing files</param>
+        /// <remarks>Inclusive on the lower bound, and exclusive on the upper bound</remarks>
+        public void ConvertFrom(DateTime startDateUtc, DateTime endDateUtc)
```
Jay-Jay-D: In general the data processing unit is a single date. Therefore, implemet a convert method for a single date, maybe an overload?

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,376 @@
...
+        /// <param name="sourceDirectory">Directory to source our raw data from</param>
+        /// <param name="destinationDirectory">Directory to write formatted data to</param>
+        /// <param name="knownTickerFolder">Directory where we source the ticker list</param>
+        public PsychSignalDataConverter(string sourceDirectory, string destinationDirectory, string knownTickerFolder)
+        {
+            _rawSourceDirectory = new DirectoryInfo(sourceDirectory);
+            _destinationDirectory = new DirectoryInfo(destinationDirectory);
+
+            _fileHandles = new Dictionary<string, TickerData>();
+            _knownTickers = Directory.GetDirectories(knownTickerFolder)
```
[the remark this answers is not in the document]
gsalaz98: I don't think so. I'll update it to just check for the existence of a map file in my upcoming commit

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,448 @@
...
+                .ToList();
+
+            var fileCount = files.Count();
+            var i = 0;
+
+            foreach (var rawFile in files)
+            {
+                i++;
+                Log.Trace($"PsychSignalDataConverter.ConvertFrom(): Reading file {rawFile.Name} (file {i}/{fileCount})");
+                Log.Trace($"PsychSignalDataConverter.ConvertFrom(): Begin converting {rawFile.Name}");
```
mchandschuh: nit -- the double trace here seems redundant.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,448 @@
...
+                            return false;
+                        }
+
+                        return fileDate >= startDateUtc && fileDate < endDateUtc;
+                    }
+                )
+                .OrderBy(x => DateTime.ParseExact(x.Name.Substring(0, 11), "yyyyMMdd_HH", CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal))
+                .ToList();
+
+            var fileCount = files.Count();
```
mchandschuh: nit - should use `Count` property instead of `Enumerable.Count()` -- that being said -- the implementation of `Enumerable.Count()` will actually perform a cast to see if it implements `ICollection` and then use the `Count` property of that. If it _doesn't_ implement `ICollection`, the `Enumerable.Count()` method would actually loop over everything to compute the count.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,448 @@
...
+
+        /// <summary>
+        /// Converts gzipped backfill data. Reads all *.gz files in the raw directory and attempts to convert them
+        /// </summary>
+        public void ConvertHistoricalData()
+        {
+            foreach (var archive in _rawSourceDirectory.GetFiles("*.gz", SearchOption.TopDirectoryOnly))
+            {
+                Log.Trace($"PsychSignalDataConverter.ConvertBackfill(): Begin converting historical data for file: {archive.FullName}");
+                using (var archiveStream = StreamProvider.ForExtension(".gz").Open(archive.FullName).First())
```
mchandschuh: nit - if we're assuming there's _exactly_ one of something, we should be using `Single()` instead of `First()` so that if our assumption is wrong, we'll get an error.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,448 @@
...
+                .ToList();
+
+            var i = 0;
+            var fileCount = files.Count();
+
+            foreach (var rawFile in files)
+            {
+                i++;
+                Log.Trace($"PsychSignalDataConverter.ConvertFrom(): Reading file {rawFile.Name} (file {i}/{fileCount})");
+                Log.Trace($"PsychSignalDataConverter.ConvertDirectory(): Begin converting {rawFile.Name}");
```
mchandschuh: nit - double trace seems redundant.

ToolBox/PsychSignalDataConverter/PsychSignalDataConverter.cs
```diff
@@ -0,0 +1,448 @@
...
+        /// </summary>
+        private void CompressData()
+        {
+            if (_disposedValue)
+            {
+                throw new ObjectDisposedException("PsychSignalDataConverter has already been disposed");
+            }
+
+            Log.Trace("PsychSignalDataConverter.CompressData(): Begin compressing PsychSignal data");
+            var timer = Stopwatch.StartNew();
```
mchandschuh: `Stopwatch` is great for things that operate on sub-second intervals -- typically for anything that's interacting with I/O, using a `var start = DateTime.UtcNow` is a little more lightweight.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,170 @@
...
+            Directory.CreateDirectory(_rawDataDestination);
+
+            if (startDateUtc < now.AddDays(-15))
+            {
+                throw new ArgumentException("The starting date can only be at most 15 days from now");
+            }
+
+            // Makes sure we only get final, non-changing data by checking if the end date is greater than
+            // or equal to the current time and setting it to an hour before the current time if the condition is met
+            if (nowHour <= new DateTime(endDateUtc.Year, endDateUtc.Month, endDateUtc.Day, endDateUtc.Hour, 0, 0))
```
mchandschuh: awesome, thanks for adding this check 👍  -- style comment -- in general, I find it's best to keep the code logically matching the the comments. The comment says we're checking if the end date is greater than or equal to the current time but the code is checking if the current time is less than or equal to the end date. These are (of course) semantically the same, but does require the reader to perform the flip in their head and could lead to confusion.

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,170 @@
...
+            var totalHours = endDateUtc.Subtract(startDateUtc).TotalHours;
+            var percentage = 1 / totalHours;
+            Log.Trace("PsychSignalDataDownloader.Download(): Begin downloading raw data");
+
+            // PsychSignal paginates data by hour. Note that it is possible to retrieve non-complete data if the requested hour
+            // is the same as the current hour or greater than the current hour.
+            for (var i = 1; startDateUtc < endDateUtc; startDateUtc = startDateUtc.AddHours(1), i++)
+            {
+                var rawDataPath = Path.Combine(_rawDataDestination, $"{startDateUtc:yyyyMMdd_HH}.csv");
+                var rawDataPathTemp = Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid().ToString()}.csv.tmp");
```
mchandschuh: fyi, when using string interpolation, you don't need to call `ToString()`, so this is equally written as:
``` csharp
$"{Guid.NewGuid()}.csv.tmp"
```
And if you want to pass a format specifier, such as `N` to remove dashes, or when stringifying a date time, `O` to use ISO8601, you can use:
``` csharp
"${Guid.NewGuid():N}.csv.tmp"
```

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
mchandschuh: note to  future readers -- this check is now safe given the protections against downloading partial data provided [here](https://github.com/QuantConnect/Lean/pull/3349/files#diff-4195c27ad597116ef70315139cfaf2caR87).

ToolBox/PsychSignalDataConverter/PsychSignalDataDownloader.cs
```diff
@@ -0,0 +1,170 @@
...
+                {
+                    Log.Trace($"PsychSignalDataDownloader.Download(): File already exists: {rawDataPath}");
+                    continue;
+                }
+
+                // Retry in case a download failed
+                for (var retries = 0; retries < MaxRetries; retries++)
+                {
+                    // Set a max timeout of ten seconds
+                    _apiRateGate.WaitToProceed(10000);
```
mchandschuh: nit -- the idea of using the rate gate was to remove the details of the rate limiting from elsewhere in the code -- no need to change, but the rate gate should be able to handle these details all on it's own.

Issue #3346, opened 2019-06-26
Create Alpha Streams Reality Models


simonsonjack:
#### Expected Behavior

Build realistic fee, slippage, and margin models for institutional clients to be used in AlphaStreams. Per an institutional client, these would act in the following ways:

**Fee Model**

- Equities
    - Long positions on margin _pays_ 1-month LIBOR rate + 40bps (basis points) on USD notional (approx 0.024 + 0.004 per annum)
    - Short positions on margin _receive_ 1-month Libor - 40bps on USD notional (approx 0.024 - 0.004 per annum)
    - I.e., a balanced long/short position only pays 2x 40bps on USD notional
- Forex
    - If trading directly with the prime broker, the cost is baked into the bid/ask spread
    - If paying the PB to place trades, the commission will be about 0.000002 * USD notional value

- Futures/Options
    - $0.40 comission per contract + $0.10 per contract for clearing

**Borrowing/Financing Model**

- Equities
    - Easy to borrow (highly liquid) stocks can be modeled at 0
    - Difficult to borrow stocks can pay anything >= 0% per annum (need to make a reasonable estimate/sliding scale based on volume)

- Forex
    - A position 1000 long in EURUSD (with rate 1.2, e.g.) would get the PB's EUR credit rate on 1000 EUR and you pay the PB's debit rate on 1200 USD.
    - Both are based on an overnight rate + spread, which will need to be investigated for best model. Spread assumption can be ~40 bps.

- Futures/Options
    - Need to borrow from PB a certain percentage of notional value to exchange as margin
    - This can be modeled at the cash debit rate (overnight rate + 40bps spread)

This will require getting PB FX data history, common margin requirements for options/futures, and common overnight rates charged by PBs.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

simonsonjack:
Fee Model class PR made. Once completed, the borrowing power model will need to be completed.

PR #3351, opened 2019-06-27
Alpha Streams Fee Model

simonsonjack:
#### Description

Created a Fee Model to be used for Alpha Streams submissions. The fee model reflects the general fees that funds will pay based on security type.
Equities -- 40 basis points + 1-month Libor (~2.4%) applied to the total notional value of the trade in USD
Forex -- 0.02 basis points applied to the total notional value of the trade in USD
Options/Futures -- $0.40 commission + $0.10 clearing fee per contract

#### Related Issue

This resolves part of issue #3346 . 

#### Motivation and Context

This will allow us to apply a universal, relevant fee model to all Alpha submissions.

Unit tests created and added to the QuantConnect.Tests project. Run in Windows 10.
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Orders/Fees/AlphaStreamsFeeModel.cs
+++ b/Common/Orders/Fees/AlphaStreamsFeeModel.cs
@@ -1,0 +1,163 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using QLNet;
+using QuantConnect.Data.Fundamental;
+using QuantConnect.Orders.Fills;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Orders.Fees
+{
+    public class AlphaStreamsFeeModel : FeeModel
+    {
+
+        private readonly decimal _forexCommissionRate = 0.000002m;
+        private decimal _liborRate;
+
+        private readonly Dictionary<string, EquityFee> _equityFee =
+            new Dictionary<string, EquityFee>();
+
+        private readonly Dictionary<string, CashAmount> _futureFee =
+            //                                                               Commission plus clearing fee
+            new Dictionary<string, CashAmount> { { Market.USA, new CashAmount(0.4m + 0.1m, "USD") } };
+
+        private readonly Dictionary<string, CashAmount> _optionFee =
+            //                                                               Commission plus clearing fee
+            new Dictionary<string, CashAmount> { { Market.USA, new CashAmount(0.4m + 0.1m, "USD") } };
+
+        public AlphaStreamsFeeModel(decimal liborRate = 0.024m)
+        {
+            _liborRate = Math.Abs(liborRate);
+            _equityFee.Add(Market.USA, new EquityFee("USD", 0.004m + _liborRate, 0));
+        }
+        /// <summary>
+        /// Gets the order fee associated with the specified order. This returns the cost
+        /// of the transaction in the account currency
+        /// </summary>
+        /// <param name="parameters">A <see cref="OrderFeeParameters"/> object
+        /// containing the security and order</param>
+        /// <returns>The cost of the order in units of the account currency</returns>
+        public override OrderFee GetOrderFee(OrderFeeParameters parameters)
+        {
+            var order = parameters.Order;
+            var security = parameters.Security;
+
+            // Option exercise for equity options is free of charge
+            if (order.Type == OrderType.OptionExercise)
+            {
+                var optionOrder = (OptionExerciseOrder)order;
+
+                if (optionOrder.Symbol.ID.SecurityType == SecurityType.Option &&
+                    optionOrder.Symbol.ID.Underlying.SecurityType == SecurityType.Equity)
+                {
+                    return OrderFee.Zero;
+                }
+            }
+
+            decimal feeResult;
+            string feeCurrency;
+            var market = security.Symbol.ID.Market;
+            switch (security.Type)
+            {
+                case SecurityType.Forex:
+                    // get the total order value in the account currency
+                    var totalOrderValue = order.GetValue(security);
+                    var fee = Math.Abs(_forexCommissionRate * totalOrderValue);
+                    feeResult = Math.Max(0, fee);
+                    // Forex fees are all in USD
+                    feeCurrency = Currencies.USD;
+                    break;
+
+                case SecurityType.Option:
+                    CashAmount optionsFeeRatePerContract;
+                    if (!_optionFee.TryGetValue(market, out optionsFeeRatePerContract))
+                    {
+                        throw new Exception($"AlphaStreamsFeeModel(): unexpected future Market {market}");
+                    }
+                    feeResult = order.AbsoluteQuantity * optionsFeeRatePerContract.Amount;
+                    feeCurrency = optionsFeeRatePerContract.Currency;
+                    break;
+
+                case SecurityType.Future:
+                    if (market == Market.Globex || market == Market.NYMEX
+                        || market == Market.CBOT || market == Market.ICE
+                        || market == Market.CBOE || market == Market.NSE)
+                    {
+                        // just in case...
+                        market = Market.USA;
+                    }
+
+                    CashAmount futuresFeeRatePerContract;
+                    if (!_futureFee.TryGetValue(market, out futuresFeeRatePerContract))
+                    {
+                        throw new Exception($"AlphaStreamsFeeModel(): unexpected future Market {market}");
+                    }
+                    feeResult = order.AbsoluteQuantity * futuresFeeRatePerContract.Amount;
+                    feeCurrency = futuresFeeRatePerContract.Currency;
+                    break;
+
+                case SecurityType.Equity:
+                    EquityFee equityFee;
+                    if (!_equityFee.TryGetValue(market, out equityFee))
+                    {
+                        throw new Exception($"AlphaStreamsFeeModel(): unexpected equity Market {market}");
+                    }
+                    
+                    //Per trade notional value fees
+                    var tradeFee = equityFee.FeePerTrade * order.GetValue(security);
+
+                    if (tradeFee < equityFee.MinimumFee)
+                    {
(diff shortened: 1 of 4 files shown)


simonsonjack:
The main thing that I think is missing is how we can update the LIBOR rate. We don't want Alphas to be using a fee model with the current 1-month LIBOR (~2.4%) if it jumps up or down a significant amount. I can look for a reliable source for this data and then we can modify the class so that it is updated on a rolling basis.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+using System;
+using System.Collections.Generic;
+using QLNet;
+using QuantConnect.Data.Fundamental;
+using QuantConnect.Orders.Fills;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Orders.Fees
+{
+    public class AlphaStreamsFeeModel : FeeModel
```
mchandschuh: We might want to add some class-level documentation here, describing the motivation/aims of this implementation as well as some of the details.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+    {
+
+        private readonly decimal _forexCommissionRate = 0.000002m;
+        private decimal _liborRate;
+
+        private readonly Dictionary<string, EquityFee> _equityFee =
+            new Dictionary<string, EquityFee>();
+
+        private readonly Dictionary<string, CashAmount> _futureFee =
+            //                                                               Commission plus clearing fee
```
mchandschuh: Why have the comment pushed all the way out to the right margin? I think this just looks odd and doesn't seem to follow the style in the rest of the code base.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+            var order = parameters.Order;
+            var security = parameters.Security;
+
+            // Option exercise for equity options is free of charge
+            if (order.Type == OrderType.OptionExercise)
+            {
+                var optionOrder = (OptionExerciseOrder)order;
+
+                if (optionOrder.Symbol.ID.SecurityType == SecurityType.Option &&
+                    optionOrder.Symbol.ID.Underlying.SecurityType == SecurityType.Equity)
```
mchandschuh: So what happens if this condition isn't true? Maybe more importantly, _when_ is this condition not true? Should we log something? I think _in general_ exercising options don't incur the usual fees. If this can be confirmed, then I think this conditional is somewhat superfluous.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+                    break;
+
+                case SecurityType.Option:
+                    CashAmount optionsFeeRatePerContract;
+                    if (!_optionFee.TryGetValue(market, out optionsFeeRatePerContract))
+                    {
+                        throw new Exception($"AlphaStreamsFeeModel(): unexpected future Market {market}");
+                    }
+                    feeResult = order.AbsoluteQuantity * optionsFeeRatePerContract.Amount;
+                    feeCurrency = optionsFeeRatePerContract.Currency;
```
mchandschuh: I suppose this is where a non-equity option exercise would end up.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+
+            decimal feeResult;
+            string feeCurrency;
+            var market = security.Symbol.ID.Market;
+            switch (security.Type)
+            {
+                case SecurityType.Forex:
+                    // get the total order value in the account currency
+                    var totalOrderValue = order.GetValue(security);
+                    var fee = Math.Abs(_forexCommissionRate * totalOrderValue);
```
mchandschuh: nit - if `order.GetValue` has the potential to be negative, then I think it makes more sense to place the `Math.Abs` around that instead of around this multiplication operation.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+        private decimal _liborRate;
+
+        private readonly Dictionary<string, EquityFee> _equityFee =
+            new Dictionary<string, EquityFee>();
+
+        private readonly Dictionary<string, CashAmount> _futureFee =
+            //                                                               Commission plus clearing fee
+            new Dictionary<string, CashAmount> { { Market.USA, new CashAmount(0.4m + 0.1m, "USD") } };
+
+        private readonly Dictionary<string, CashAmount> _optionFee =
```
mchandschuh: If the reason for the right aligned comment is so it appears near the numbers in `new CashAmount(...)`, perhaps formatting as this would help:

``` csharp
private readonly Dictionary<string, CashAmount> _optionFee = new Dictionary<string, CashAmount>
{
    // Commission plus clearing fee
    { Market.USA, new CashAmount(0.4m + 0.1m, "USD") }
};
```

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+            //                                                               Commission plus clearing fee
+            new Dictionary<string, CashAmount> { { Market.USA, new CashAmount(0.4m + 0.1m, "USD") } };
+
+        public AlphaStreamsFeeModel(decimal liborRate = 0.024m)
+        {
+            _liborRate = Math.Abs(liborRate);
```
mchandschuh: As suggested in the main discussion area, this may be better phrased as a `Security` object that can be updated via `BaseData` -- this would also require the accompanying `BaseData` definition. As the LIBOR rate changes, we'll be able to send updates so the algorithms get the fresh value. We would also need a system for requesting the current LIBOR rate for when an algorithm first initializes (unless we wanted to pulse the data every minute or so down the data feed). Another bonus of phrasing this as a `BaseData` implementation is the ability to properly model changing LIBOR rate in backtesting.

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+        private readonly Dictionary<string, EquityFee> _equityFee =
+            new Dictionary<string, EquityFee>();
+
+        private readonly Dictionary<string, CashAmount> _futureFee =
+            //                                                               Commission plus clearing fee
+            new Dictionary<string, CashAmount> { { Market.USA, new CashAmount(0.4m + 0.1m, "USD") } };
+
+        private readonly Dictionary<string, CashAmount> _optionFee =
+            //                                                               Commission plus clearing fee
+            new Dictionary<string, CashAmount> { { Market.USA, new CashAmount(0.4m + 0.1m, "USD") } };
```
mchandschuh: > Options/Futures -- $0.40 commission + $0.10 clearing fee per contract

Is the commission a per trade or per contract number? If so, I'm curious why the distinction is made both here and in the code regarding commission+clearing (same for futures).

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        public AlphaStreamsFeeModel(decimal liborRate = 0.024m)
+        {
+            _liborRate = Math.Abs(liborRate);
+            _equityFee.Add(Market.USA, new EquityFee("USD", 0.004m + _liborRate, 0));
```
mchandschuh: Are there any concerns about only supporting US equities?

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+                    feeCurrency = optionsFeeRatePerContract.Currency;
+                    break;
+
+                case SecurityType.Future:
+                    if (market == Market.Globex || market == Market.NYMEX
+                        || market == Market.CBOT || market == Market.ICE
+                        || market == Market.CBOE || market == Market.NSE)
+                    {
+                        // just in case...
+                        market = Market.USA;
```
mchandschuh: if we're going to do all of this, and since the dictionary has exactly one value, I'm wondering of the benefit of using the dictionary at all -- why not handle it like forex, or conversely, why isn't forex handled like all the others?

Common/Orders/Fees/AlphaStreamsFeeModel.cs
```diff
@@ -0,0 +1,163 @@
...
+                    throw new ArgumentException($"Unsupported security type: {security.Type}");
+            }
+
+            return new OrderFee(new CashAmount(
+                feeResult,
+                feeCurrency));
+        }
+
+        /// <summary>
+        /// Helper class to handle IB Equity fees
```
mchandschuh: If this specific the IB, then perhaps naming it as `IBEquityFee` or `InteractiveBrokersEquityFee` maybe be more descriptive.

Tests/Common/Orders/Fees/AlphaStreamsFeeModelTests.cs
```diff
@@ -0,0 +1,198 @@
...
+using QuantConnect.Securities.Cfd;
+using QuantConnect.Securities.Forex;
+using QuantConnect.Securities.Future;
+using QuantConnect.Securities.Option;
+using QuantConnect.Tests.Common.Securities;
+
+namespace QuantConnect.Tests.Common.Orders.Fees
+{
+    [TestFixture]
+    class AlphaStreamsFeeModelTests
```
mchandschuh: I think in most cases, we define test classes as public classes.

Tests/Common/Orders/Fees/AlphaStreamsFeeModelTests.cs
```diff
@@ -0,0 +1,198 @@
...
+                    new MarketOrder(security.Symbol, 1000, DateTime.UtcNow)
+                )
+            );
+
+            Assert.AreEqual(Currencies.USD, fee.Value.Currency);
+            Assert.AreEqual(1000 * 0.50m, fee.Value.Amount);
+        }
+
+        [Test]
+        public void USAOptionFee()
```
mchandschuh: style - feel free to ignore, but please consider more descriptive names - I tend to prefer test method names that are descriptive as to what case is being tested and what the expected result is. Reading this test name, `USAOptionFee`, all I know is that it's test option fees of the USA market, but I have no idea what the expected result is or the logic behind the result. I think something like, `GetOrderFee_Is50BasisPointsInUSD_WhenSecuityTypeIsOption` -- the usage of underscores is a recent personal style change aimed at making the test names more readable. I've lately been generally following a pattern of `<method>_<result>_when<case>` as it makes reading any test logs extremely explicit as to what was tested (`<classname>.<method>``), under what conditions `when<case>`, and the expected result `<result>`.

simonsonjack:
@mchandschuh I appreciate the thorough feedback. This is one of my first major LEAN contributions in C#, and I took the IB fee model for inspiration and replaced some of the guts to write this. Thus, my negligence in commenting properly shows in some places. I'll address changes on Monday and hopefully review more with @Martin-Molinero and others to find a good way of implementing the LIBOR rate so that it can be updated as the true rate is rather than being statically defined at the beginning of an Alpha deployment.

Issue #3078, opened 2019-04-12
Missing Import in api.py Cause a Runtime Error in download_data Method


AlexCatarino:
#### Expected Behavior
Be able to download data from QuantConnect using a Python script.

#### Actual Behavior
`NameError: name 'requests' is not defined` exception due to a missing import.

#### Potential Solution
Replaces:
`from requests import Request`
for
`from requests import get, Request`

#### Reproducing the Problem
```python
from datetime import datetime
from quantconnect.api import Api
api = Api(job-user-id, api-access-token)
api.download_data('AUDJPY', 'Forex', 'fxcm', 'second', datetime(2017, 4, 5) , '20170405_quote.zip')
```
#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #3353, opened 2019-06-27
 Reduce memory usage in Api.download_data

tipabu:
#### Description

Requests [has some sample code](https://2.python-requests.org//en/master/user/quickstart/#raw-response-content) in its docs for streaming a response to a file; follow the recommended pattern.

> In general, however, you should use a pattern like this to save what is being streamed to a file:
>
> ``` python
> with open(filename, 'wb') as fd:
>     for chunk in r.iter_content(chunk_size=128):
>         fd.write(chunk)
> ```

#### Related Issue

#3078 - Missing Import in api.py Cause a Runtime Error in download_data Method

#### Motivation and Context
Previously, the entire response would be read and buffered in memory before any of it was written to disk.

Now, stream the response and write down data as it's received.

As a drive-by, make the thing actually work.

N/A

#### Types of changes

#### Checklist:

### DIFF
--- a/PythonToolbox/quantconnect/api.py
+++ b/PythonToolbox/quantconnect/api.py
@@ -13,7 +13,7 @@
 
 from datetime import datetime
 from json import dumps
-from requests import Request
+from requests import get as requests_get, Request
 from time import mktime
 from quantconnect import ApiConnection
 
@@ -468,7 +468,7 @@ class Api:
         # Make sure the link was successfully retrieved
         if link['success']:
             # download and save the data
-            request = requests.get(link['link'])
+            request = requests_get(link['link'])
             with open(fileName + '.zip', "wb") as code:
                 code.write(request.content)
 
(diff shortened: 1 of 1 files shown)


PythonToolbox/quantconnect/api.py
```diff
@@ -13,7 +13,7 @@
 
 from datetime import datetime
 from json import dumps
-from requests import Request
+from requests import get as requests_get, Request
```
mchandschuh: was there a naming collision here that required the alias? if not, my preference would be to not provide the alias as they tend to make code less readable as it requires one to determine what the alias means. That being said, the chosen alias is _extremely_ explicit, so the aforementioned statement really doesn't apply. I certainly like the naming convention of `<object>_<method>` for aliases as it's almost like it wasn't aliased at all 👍

PythonToolbox/quantconnect/api.py
```diff
@@ -468,8 +468,9 @@ def download_data(self, symbol, securityType, market, resolution, date, fileName
         # Make sure the link was successfully retrieved
         if link['success']:
             # download and save the data
-            request = requests.get(link['link'])
+            request = requests_get(link['link'], stream=True)
```
mchandschuh: I love how such a simple change can make a difference. Was any profiling using to show to show the improvement. I agree that the referenced documentation is reason enough to support this change 👍

Issue #3354, opened 2019-06-27
Scheduled Events Crash with 0-Timespan Delta


jaredbroad:
#### Expected Behavior
Throw an exception when the user tries to schedule an event for every 0 seconds (infinite).

#### Actual Behavior
Tries to calculate infinite scheduled events and runs out of ram

#### Potential Solution
Check time value is > 0 

#### Reproducing the Problem
Configure TimeRule for 0 minutes;

#### Checklist 
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3355, opened 2019-06-28
Validate TimeRules.Every TimeSpan

Martin-Molinero:
#### Description

- Will verify `TimeSpan` is not zero or less, adding unit tests.

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3354
#### Motivation and Context

Improve user experience

N/A

Added unit tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Common/Scheduling/TimeRulesTests.cs
+++ b/Tests/Common/Scheduling/TimeRulesTests.cs
@@ -33,7 +33,7 @@ namespace QuantConnect.Tests.Common.Scheduling
         {
             var rules = GetTimeRules(TimeZones.Utc);
             var rule = rules.At(TimeSpan.FromHours(12));
-            var times = rule.CreateUtcEventTimes(new[] {new DateTime(2000, 01, 01)});
+            var times = rule.CreateUtcEventTimes(new[] { new DateTime(2000, 01, 01) });
 
             int count = 0;
             foreach (var time in times)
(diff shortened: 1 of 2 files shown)


Tests/Common/Scheduling/TimeRulesTests.cs
```diff
@@ -183,11 +183,36 @@ public void ExtendedMarketCloseWithDelta()
...
                 count++;
-                Assert.AreEqual(TimeSpan.FromHours((20 + 5 - .5)%24), time.TimeOfDay);
+                Assert.AreEqual(TimeSpan.FromHours((20 + 5 - .5) % 24), time.TimeOfDay);
             }
             Assert.AreEqual(1, count);
         }
 
+        [TestCase(0)]
+        [TestCase(-1)]
+        public void EveryValidatesTimeSpan(int timeSpanMinutes)
```
mchandschuh: I tend to prefer phrasing that states what the validation is, something like `EveryThrowsArgumentExceptionWhenIntervalIsLessThanOrEqualToZero` -- it's very wordy, and recently I've been inserting underscores, so something like <method>_<result>_when<case> formatting, so `Every_ThrowsArgumentException_WhenIntervalIsLessthanOrEqualToZero` -- the benefit of this explicitness is two fold, firstly, tests of documentation are a good thing, and secondly, if the tests fail just the name of the failing test usually provides enough information to determine what might have happened (assuming recent changes broke the test). The underscore pattern isn't really used in this code base, so not sure I would recommend switching now unless it's a team decision.

PR #2588, opened 2018-10-08
Exclude OptionChainProviderTests from Travis build

StefanoRaggi:
#### Description
The `OptionChainProvider` unit tests have been excluded from the Travis build.

#### Related Issue
Closes #2587 

#### Motivation and Context
Randomly failing unit tests.

#### Requires Documentation Change
No.

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description` or `feature-<issue#>-<description>`

PR #3365, opened 2019-07-03
Add retry logic to LiveOptionChainProvider.GetOptionContractList

### LINKED ISSUE #3364: LiveOptionChainProvider.GetOptionContractList Failure
#### Expected Behavior
`LiveOptionChainProvider.GetOptionContractList` should try to download the option chain a few times, before reporting the exception and terminating the algorithm.

#### Actual Behavior
In `LiveOptionChainProvider.GetOptionContractList` any web failure or timeout causes the algorithm to be terminated immediately with a runtime error. 
The most common error seen is this one:
```
System.Net.WebException: Error getting response stream (ReadDoneAsync2): ReceiveFailure
```

#### Potential Solution
Add retry logic in `LiveOptionChainProvider.GetOptionContractList`

#### Reproducing the Problem
n/a

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- The `LiveOptionChainProvider.GetOptionContractList` method has been updated to retry failed API calls up to a maximum of 5 attempts.
- `OptionChainProvider` unit tests have been re-enabled for Travis build (previously disabled in #2588) 

#### Related Issue
Closes #3364 

#### Motivation and Context
Prevent algorithm termination after a single Web request failure.

No.

- Existing unit tests with forced failure on the first attempt
- Existing unit tests with forced failure on all attempts
- Existing unit tests with a real failure (timeout + server unreachable) -- got lucky there :)

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/LiveOptionChainProvider.cs
+++ b/Engine/DataFeeds/LiveOptionChainProvider.cs
@@ -17,7 +17,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Threading;
 using QuantConnect.Interfaces;
+using QuantConnect.Logging;
 
 namespace QuantConnect.Lean.Engine.DataFeeds
 {
@@ -27,6 +29,8 @@ namespace QuantConnect.Lean.Engine.DataFeeds
     /// </summary>
     public class LiveOptionChainProvider : IOptionChainProvider
     {
+        private const int MaxDownloadAttempts = 5;
+
         /// <summary>
         /// Static constructor for the <see cref="LiveOptionChainProvider"/> class
         /// </summary>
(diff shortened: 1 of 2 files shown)


Engine/DataFeeds/LiveOptionChainProvider.cs
```diff
@@ -50,7 +54,27 @@ public IEnumerable<Symbol> GetOptionContractList(Symbol symbol, DateTime date)
...
+            do
+            {
+                try
+                {
+                    Log.Trace($"LiveOptionChainProvider.GetOptionContractList(): Fetching option chain for {symbol.Value} [Attempt {attempt}]");
+
+                    contracts = FindOptionContracts(symbol.Value);
+                    break;
+                }
+                catch (Exception exception)
```
Martin-Molinero: Wondering if we want to catch any type of exception or maybe better just the web request one?
Also if we fail do download the `contracts` 5 times, wouldn't we want to finally throw?

StefanoRaggi: Good points! 
An exception which is not a `WebException` could be caused by updated content/format by OCC and should be reported immediately, not only logged silently.
If 5 download attempts fail, it should definitely throw.
Will fix both.

Issue #3363, opened 2019-07-03
ZipDataCacheProvider dispose race conditon


Martin-Molinero:
#### Expected Behavior

- `ZipDataCacheProvider` suffers no race condition
#### Actual Behavior

- `ZipDataCacheProvider` `CleanCache()` called every 3 seconds could dispose https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/ZipDataCacheProvider.cs#L187 -> https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/ZipDataCacheProvider.cs#L326 of an object being used by another thread, causing it to fail to read the zip entry data at `CreateStream()`

```
20190630 19:19:08.499 ERROR:: master-8e51649-bf69f17: ZipDataCacheProvider.Fetch(): Inner try/catch System.ObjectDisposedException: Cannot access a disposed object.
20190630 19:19:08.499 ERROR:: master-8e51649-bf69f17: Object name: 'Stream has been closed'.
20190630 19:19:08.499 ERROR:: master-8e51649-bf69f17:   at System.IO.FileStream.get_Position () [0x0000d] in <b0e1ad7573a24fd5a9f2af9595e677e7>:0
20190630 19:19:08.499 ERROR:: master-8e51649-bf69f17:   at Ionic.Zip.ZipEntry.SetFdpLoh () [0x00006] in <7bc1b4baa0c9431b8115554519fcef22>:0
20190630 19:19:08.500 ERROR:: master-8e51649-bf69f17:   at Ionic.Zip.ZipEntry.get_FileDataPosition () [0x0000a] in <7bc1b4baa0c9431b8115554519fcef22>:0
20190630 19:19:08.500 ERROR:: master-8e51649-bf69f17:   at Ionic.Zip.ZipEntry.ExtractAndCrc (System.IO.Stream archiveStream, System.IO.Stream targetOutput, System.Int16 compressionMethod, System.Int64 compressedFileDataSize, System.Int64 uncompressedSize) [0x00002] in <7bc1b4baa0c9431b8115554519fcef22>:0
[...]

PR #3366, opened 2019-07-04
Fix dispose race condition at ZipDataCacheProvider

Martin-Molinero:
#### Description

- `ZipDataCacheProvider.CleanCache()` will not dispose or remove entry if its being used
   - In the edge case, an entry is disposed before securing the lock on it, the entry will be re created and added.
- Reduce `DateTime.UtcNow` usages
- Just one thread will execute `CleanCache()` at the same time

### Tests
 `BasicTemplateOptionTradesAlgorithm 2018 to 2018-6. Reducing time in the cache to 0.1 seconds to reproduce the issue more frequently`

`master`
- 7 exceptions, didn't finish running, algorithm thread died/dead locked?
- 3 exceptions
- 3 exceptions
- 9 Exceptions

`pr - no exceptions`
151.66 seconds at 880k data points per second. Processing total of 133,435,554 data points.
149.40 seconds at 893k data points per second. Processing total of 133,435,554 data points.
146.89 seconds at 908k data points per second. Processing total of 133,435,554 data points.
143.47 seconds at 930k data points per second. Processing total of 133,435,554 data points.

`pr` using 10 seconds cache timeout
78.93 seconds at 1691k data points per second. Processing total of 133,435,554 data points.
77.19 seconds at **1729k data points per second**. Processing total of 133,435,554 data points.

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3363
#### Motivation and Context

Make `ZipDataCacheProvider` more reliable

N/A

Regression and unit tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/ZipDataCacheProvider.cs
+++ b/Engine/DataFeeds/ZipDataCacheProvider.cs
@@ -19,6 +19,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using QuantConnect.Logging;
 using System.Linq;
+using System.Threading;
 using Ionic.Zip;
 using Ionic.Zlib;
 using QuantConnect.Interfaces;
@@ -35,7 +36,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
 
         // ZipArchive cache used by the class
         private readonly ConcurrentDictionary<string, CachedZipFile> _zipFileCache = new ConcurrentDictionary<string, CachedZipFile>();
-        private DateTime _lastCacheScan = DateTime.MinValue;
+        private DateTime _nextCacheScan = DateTime.MinValue;
         private readonly IDataProvider _dataProvider;
 
         /// <summary>
@@ -71,10 +72,11 @@ namespace QuantConnect.Lean.Engine.DataFeeds
             {
                 Stream stream = null;
 
-                // scan the cache once every 3 seconds
-                if (_lastCacheScan == DateTime.MinValue || _lastCacheScan < DateTime.UtcNow.AddSeconds(-3))
+                // scan the cache once every CacheSeconds
+                var utcNow = DateTime.UtcNow;
+                if (_nextCacheScan < utcNow)
                 {
-                    CleanCache();
+                    CleanCache(utcNow);
                 }
 
                 try
@@ -82,29 +84,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                     CachedZipFile existingEntry;
                     if (!_zipFileCache.TryGetValue(filename, out existingEntry))
                     {
-                        var dataStream = _dataProvider.Fetch(filename);
-
-                        if (dataStream != null)
-                        {
-                            try
-                            {
-                                var newItem = new CachedZipFile(dataStream, filename);
-                                stream = CreateStream(newItem, entryName, filename);
-
-                                if (!_zipFileCache.TryAdd(filename, newItem))
-                                {
-                                    newItem.Dispose();
-                                }
-                            }
-                            catch (Exception exception)
-                            {
-                                if (exception is ZipException || exception is ZlibException)
-                                {
-                                    Log.Error("ZipDataCacheProvider.Fetch(): Corrupt zip file/entry: " + filename + "#" + entryName + " Error: " + exception);
-                                }
-                                else throw;
-                            }
-                        }
+                        stream = CacheAndCreateStream(filename, entryName, utcNow);
                     }
                     else
                     {
@@ -112,7 +92,17 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                         {
                             lock (existingEntry)
                             {
-                                stream = CreateStream(existingEntry, entryName, filename);
+                                if (existingEntry.Disposed)
+                                {
+                                    // bad luck, thread race condition
+                                    // it was disposed and removed after we got it
+                                    // so lets create it again and add it
+                                    stream = CacheAndCreateStream(filename, entryName, utcNow);
+                                }
+                                else
+                                {
+                                    stream = CreateStream(existingEntry, entryName, filename);
+                                }
                             }
                         }
                         catch (Exception exception)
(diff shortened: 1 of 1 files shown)


Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -170,26 +160,80 @@ public void Dispose()
...
         /// </summary>
-        private void CleanCache()
+        private void CleanCache(DateTime utcNow)
         {
-            var clearCacheIfOlderThan = DateTime.UtcNow.AddSeconds(-CacheSeconds);
-
-            // clean all items that that are older than CacheSeconds than the current date
-            foreach (var zip in _zipFileCache)
+            // we just want one thread cleaning the cache at the same time
+            if (Monitor.TryEnter(_zipFileCache))
```
mchandschuh: Nice usage of `Monitor.TryEntry` here -- no point in everyone waiting, let the others skip 👍

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -71,48 +72,37 @@ public Stream Fetch(string key)
...
                         try
                         {
                             lock (existingEntry)
                             {
-                                stream = CreateStream(existingEntry, entryName, filename);
+                                if (existingEntry.Disposed)
+                                {
+                                    // bad luck, thread race condition
+                                    // it was disposed and removed after we got it
+                                    // so lets create it again and add it
```
mchandschuh: thank you for the comment describing this case.

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -170,26 +160,80 @@ public void Dispose()
...
             }
+        }
+
+        private Stream CacheAndCreateStream(string filename, string entryName, DateTime utcNow)
+        {
+            Stream stream = null;
+            var dataStream = _dataProvider.Fetch(filename);
 
-            _lastCacheScan = DateTime.UtcNow;
+            if (dataStream != null)
```
mchandschuh: Consider inverting this `if` to reduce nesting:
``` csharp
if (dataStream == null)
{
    return null;
}
```

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -170,26 +160,80 @@ public void Dispose()
...
+            {
+                try
+                {
+                    var newItem = new CachedZipFile(dataStream, filename, utcNow);
+
+                    // here we don't need to lock over the cache item
+                    // because it was still not added in the cache
+                    stream = CreateStream(newItem, entryName, filename);
+
+                    if (!_zipFileCache.TryAdd(filename, newItem))
```
mchandschuh: `ConcurrentDictionary<TKey, TValue>` provides an `AddOrUpdate` method that may be better suited here. Another pattern I've seen in the past is to use `ConcurrentDictionary<TKey, Lazy<TValue>>` which provides thread-safety on object creation by virtue of the `Lazy<T>`. What we have here _will_ work, but we'll also be creating and disposing an entry randomly when we don't need to. Using the `Lazy<T>` will prevent this case from happening.

When using the `ConcurrentDictionary<TKey, Lazy<TValue>>` you would use
```csharp
_zipFileCache.GetOrAdd(filename, fname => new Lazy<Stream>(() => /** lazy stream creation**/)
```

and we'll be guaranteed that the returned `Lazy<Stream>` is the same one used by all thread. We'll also be guaranteed that only a _single_ stream will be created. [Here's](https://blogs.endjin.com/2015/10/using-lazy-and-concurrentdictionary-to-ensure-a-thread-safe-run-once-lazy-loaded-collection/) an article I found describing the pattern. Also of interest, is the [Remarks](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netframework-4.8) section of the `ConcurrentDictionary` MSDN entry for `GetOrAdd` describing the locking methodology used which guarantees that the returned instance is the same for all thread _but does not_ guarantee that only a single delegate is executed, hence the usage of `Lazy<T>` to ensure we only execute the factory delegate once.

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -170,26 +160,80 @@ public void Dispose()
...
+                        newItem.Dispose();
+                    }
+                }
+                catch (Exception exception)
+                {
+                    if (exception is ZipException || exception is ZlibException)
+                    {
+                        Log.Error("ZipDataCacheProvider.Fetch(): Corrupt zip file/entry: " + filename + "#" + entryName + " Error: " + exception);
+                    }
+                    else throw;
```
mchandschuh: nit - bracing style -- in general, this repository always uses braces.

Engine/DataFeeds/ZipDataCacheProvider.cs
```diff
@@ -290,12 +334,18 @@ public class CachedZipFile : IDisposable
         /// </summary>
         public string Key { get; private set; }
 
+        /// <summary>
+        /// Returns if this cached zip file is disposed
+        /// </summary>
+        public bool Disposed { get; private set; }
```
mchandschuh: Any concerns about this backing field _not_ being marked as `volatile`? Perhaps we need a `Thread.MemoryBarrier()` each time we set or retrieve the value to prevent CPU re-ordering of the instructions? Here's a great [stackoverflow](https://stackoverflow.com/a/3556877/1582922) answer discussing the lock-release semantics of .net framework as opposed to the ECMA spec itself (which uses a weaker memory model, and IIRC, mono follows the weaker model leading to bugs). The concern is the 'freshness' of the value, in other words, if ThreadA sets `Disposed=true`, then how long before ThreadB sees that `true` value since it may have cached the value from main memory. The  memory barrier (or also `volatile`) will force the CPU to retrieve the latest value from main memory (if memory serves :) )

PR #3370, opened 2019-07-05
Add new machine learning demonstration algorithms

QilongChan:
Add new machine learning demonstration algorithms for Pytorch, Keras, TensorFlow, Scikit-Learn, and NLTK.

#### Checklist:

### DIFF
--- a/Algorithm.Python/KerasNeuralNetworkAlgorithm.py
+++ b/Algorithm.Python/KerasNeuralNetworkAlgorithm.py
@@ -89,11 +89,11 @@ class KerasNeuralNetworkAlgorithm(QCAlgorithm):
             self.sell_prices[symbol.Value] = y_pred_final - np.std(y_data)
         
     def Trade(self):
-        for i in self.Portfolio.Values:
+        for holding in self.Portfolio.Values:
             # liquidate
-            if self.CurrentSlice[i.Symbol.Value].Open < self.sell_prices[i.Symbol.Value] and i.Invested:
+            if self.CurrentSlice[holding.Symbol.Value].Open < self.sell_prices[holding.Symbol.Value] and holding.Invested:
                 self.Liquidate(i.Symbol)
             
             # buy
-            if self.CurrentSlice[i.Symbol.Value].Open > self.buy_prices[i.Symbol.Value] and not i.Invested:
-                self.SetHoldings(i.Symbol, 1 / len(self.symbols))
+            if self.CurrentSlice[holding.Symbol.Value].Open > self.buy_prices[holding.Symbol.Value] and not holding.Invested:
+                self.SetHoldings(holding.Symbol, 1 / len(self.symbols))
--- a/Algorithm.Python/KerasNeuralNetworkAlgorithm.py
+++ b/Algorithm.Python/KerasNeuralNetworkAlgorithm.py
@@ -38,10 +38,10 @@ class KerasNeuralNetworkAlgorithm(QCAlgorithm):
         self.lookback = 30 # day of lookback for historical data
         
         self.Schedule.On(self.DateRules.Every(DayOfWeek.Monday), self.TimeRules.AfterMarketOpen("SPY", 28), self.NetTrain) # train Neural Network
-        self.Schedule.On(self.DateRules.Every(DayOfWeek.Monday), self.TimeRules.AfterMarketOpen("SPY", 30), self.Trade) # trading
+        self.Schedule.On(self.DateRules.Every(DayOfWeek.Monday), self.TimeRules.AfterMarketOpen("SPY", 30), self.Trade) 
         
     def NetTrain(self):
-        # get daily historical data
+        # Daily historical data is used to train the machine learning model 
         history = self.History(self.symbols, self.lookback + 1, Resolution.Daily)
         
         # dicts that store prices for training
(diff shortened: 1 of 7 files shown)


Algorithm.Python/NLTKSentimentTradingAlgorithm.py
```diff
@@ -0,0 +1,73 @@
...
+                self.Liquidate(i.Symbol)
+            
+            # buy
+            if positive_word in words and not i.Invested:
+                self.SetHoldings(i.Symbol, 1 / len(self.symbols))
+        
+    def get_text(self):
+        # import custom data
+        # Note: dl must be 1, or it will not download automatically
+        url = 'https://www.dropbox.com/s/8nhbizxq3lgpced/EconomicNews.csv?dl=1'
```
jaredbroad: Please update to: `https://www.dropbox.com/s/7xgvkypg6uxp6xl/EconomicNews.csv?dl=1`

Algorithm.Python/TensorFlowNeuralNetworkAlgorithm.py
```diff
@@ -0,0 +1,112 @@
...
+
+from System import *
+from QuantConnect import *
+from QuantConnect.Algorithm import *
+
+
+import numpy as np
+import tensorflow as tf
+
+class TensorFlowNeuralNetworkAlgorithm(QCAlgorithm):
```
jaredbroad: 1) Please add comments to each algorithm class about what it is doing. 
2) Please aggressively comment on ALL parts of the algorithm. These algorithms are _documentation_ so important to comment them well.

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
```diff
@@ -0,0 +1,97 @@
...
+
+class KerasNeuralNetworkAlgorithm(QCAlgorithm):
+
+    def Initialize(self):
+        self.SetStartDate(2013, 10, 7)  # Set Start Date
+        self.SetEndDate(2013, 10, 8) # Set End Date
+        
+        self.SetCash(100000)  # Set Strategy Cash
+        spy = self.AddEquity("SPY", Resolution.Minute)
+        self.symbols = [spy.Symbol]
```
AlexCatarino: Since we are using only want symbol, SPY, why is this list defined?
I would suggest:
```python
self.symbol = self.AddEquity("SPY", Resolution.Minute).Symbol
```
and replace every `"SPY"` occurence for `self.symbol`.
Also, it would replace all loops and dictionaries below.

On the other hand, another security could be added:
```python
qqq = self.AddEquity("QQQ", Resolution.Minute)
self.symbols = [spy.Symbol qqq.Symbol]
```

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
```diff
@@ -0,0 +1,97 @@
...
+        
+        self.lookback = 30
+        
+        self.Schedule.On(self.DateRules.Every(DayOfWeek.Monday), self.TimeRules.AfterMarketOpen("SPY", 28), Action(self.NetTrain))
+        self.Schedule.On(self.DateRules.Every(DayOfWeek.Monday), self.TimeRules.AfterMarketOpen("SPY", 30), Action(self.Trade))
```
AlexCatarino: We don't need to use `Action()`.

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+    def OnData(self, data):
+        self.data = data
```
AlexCatarino: We should remove these lines and use `self.CurrentSlice` instead.

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
```diff
@@ -0,0 +1,97 @@
...
+                    cost = model.train_on_batch(x_data, y_data)
+
+            y_pred_final = model.predict(y_data)[0][-1]
+            
+            # Follow the trend
+            self.buy_prices[symbol.Value] = y_pred_final + np.std(y_data)
+            self.sell_prices[symbol.Value] = y_pred_final - np.std(y_data)
+        
+    def Trade(self):
+        for i in self.Portfolio.Values:
```
AlexCatarino: Style comment: instead of using `i`, we should use a more meaningful name. `Portolio.Values` is a list of `SecurityHolding` and in other examples, we use `holding` for short:
```python
for holding in self.Portfolio.Values:
    symbol = holding.Symbol
    if self.CurrentSlice[symbol].Open < self.sell_prices[symbol.Value] and holding.Invested:
        self.Liquidate(symbol)
```

Algorithm.Python/NLTKSentimentTradingAlgorithm.py
```diff
@@ -0,0 +1,73 @@
...
+        spy = self.AddEquity("SPY", Resolution.Minute)
+        self.text = self.get_text()
+        self.symbols = [spy.Symbol]
+        
+        # for what extra models needed to download, please use code nltk.download()
+        nltk.download('punkt')
+        self.Schedule.On(self.DateRules.EveryDay("SPY"), self.TimeRules.AfterMarketOpen("SPY", 30), Action(self.Trade))
+
+    def OnData(self, data):
+        pass
```
AlexCatarino: We can remove `OnData` if it just `pass`.

Algorithm.Python/NLTKSentimentTradingAlgorithm.py
```diff
@@ -0,0 +1,73 @@
...
+    
+    def Trade(self):
+        current_time = str(self.Time.year) + '-' + str(self.Time.month) + '-' + str(self.Time.day)
```
AlexCatarino: Tip: use f-string 
```python
current_time = f'{self.Time.year}-{self.Time.month}-{self.Time.day}'
```

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
mchandschuh: provided the rest of the algorithm can handle a list, then I think it's nice to use list notation as it makes it easier for users to extend for multiple symbols.

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
mchandschuh: Also, `i` is typically reserved for an array/list index in a loop and not for an item of a collection, such as `holding`

Algorithm.Python/KerasNeuralNetworkAlgorithm.py
QilongChan: Thanks for all your comments. My idea is that using a list is much easier for users to further this strategy when they need to include more trading symbols.

Issue #3021, opened 2019-03-22
Notify.Email fails silently if message is null


Doggie52:
#### Expected Behavior
Sending a `Notify.Email` with `message = null` either works (i.e. sends an empty email with just the subject) or logs an error.

#### Actual Behavior
Sending a `Notify.Email` with `message = null` fails silently. No mention of this in the code or in the documentation.

#### Potential Solution
It can be useful to send an empty email for when the message you want to communicate fits nicely in the subject. If we don't want to allow empty emails, at least it should fail loudly.

#### Reproducing the Problem
I have not reproduced this locally as I haven't been running a local instance with `liveMode = true`.

On the terminal:
`Notify.Email( "<email here>", "Subject test", null );`

Observe no error message.

I don't know what happens if subject is `null` but message is non-null.

#### System Information
QC Terminal

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

Doggie52:
Closed in #3374 !

PR #3374, opened 2019-07-08
Provide unit tests for NotificationManager, NotitificationEmail and default message/subject

mchandschuh:
#### Description

Without access to the `IMessagingHandler` implementation used in the production environment, the next best solution is to provide a default value for `message` and `subject` to ensure it's never null. The likely culprit is a null reference exception that is being swallowed on a different thread where the notification messages are being processed.

This change adds unit tests for `NotificationManager` as well as provides some basic refactoring to match style guidelines, stream-line `Allow` checks, as well as removing a race condition in the rate limiting checks (perhaps should be updated to `RateGate`, but felt it was outside the scope of this PR). Additionally, the `_resetTime` was changed to use `DateTime.UtcNow` to avoid potential issues with local time zone settings and the inevitable time discontinuity that arises from things like daylight savings adjustments.

In addition to providing default values for `NotificationEmail.Subject` and `NotificationEmail.Message`, validation has been added to ensure a valid email address has been provided via a new `Validate` class with a `Validate.EmailAddress` method. The reasoning behind a new class here is to provide a location for further string based validations (such as phone number, web address, etc) and to provide a singular location for potentially complicated regular expressions.

Another change was made to the `QuantConnect.sln.DotSettings` file to include `QUANTCONNECT` in the custom dictionary to avoid R# typo warnings in the license headers.

#### Related Issue

See #3021 - Notify.Email fails silently if message is null

#### Motivation and Context

The `Notify.Email` method should not fail when given a null message or subject as these are mandatory fields when sending an email.

No. This change updates the behavior to the expected behavior of not failing.

Changes extensively tested via unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Notifications/Notification.cs
+++ b/Common/Notifications/Notification.cs
@@ -1,17 +1,20 @@
 ﻿/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- * 
- * Licensed under the Apache License, Version 2.0 (the "License"); 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
 */
+
+using System;
+using QuantConnect.Util;
 
 namespace QuantConnect.Notifications
 {
@@ -90,6 +93,8 @@ namespace QuantConnect.Notifications
     /// </summary>
     public class NotificationEmail : Notification
     {
+        private const string DefaultMessageAndSubject = "QuantConnect LEAN Email Notification";
+
         /// <summary>
         /// Send to address:
         /// </summary>
@@ -119,10 +124,15 @@ namespace QuantConnect.Notifications
         /// <param name="data">Data to attach to the email</param>
         public NotificationEmail(string address, string subject, string message, string data)
         {
-            Message = message;
-            Data = data;
-            Subject = subject;
+            if (!Validate.EmailAddress(address))
+            {
+                throw new ArgumentException($"Invalid email address: {address}");
+            }
+
             Address = address;
+            Data = data ?? string.Empty;
+            Message = message ?? DefaultMessageAndSubject;
+            Subject = subject ?? DefaultMessageAndSubject;
         }
     }
 }
--- a/Common/Notifications/Notification.cs
+++ b/Common/Notifications/Notification.cs
@@ -93,8 +93,6 @@ namespace QuantConnect.Notifications
     /// </summary>
     public class NotificationEmail : Notification
     {
-        private const string DefaultMessageAndSubject = "QuantConnect LEAN Email Notification";
-
         /// <summary>
         /// Send to address:
         /// </summary>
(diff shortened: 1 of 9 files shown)


Tests/Common/Notifications/NotificationEmailTests.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
```
Martin-Molinero: New file missing header

Common/Notifications/Notification.cs
```diff
@@ -119,10 +124,15 @@ public class NotificationEmail : Notification
...
-            Subject = subject;
+            if (!Validate.EmailAddress(address))
+            {
+                throw new ArgumentException($"Invalid email address: {address}");
+            }
+
             Address = address;
+            Data = data ?? string.Empty;
+            Message = message ?? DefaultMessageAndSubject;
+            Subject = subject ?? DefaultMessageAndSubject;
```
Martin-Molinero: Maybe it's worth documenting this new behavior in the methods xml docs? If null will use `DefaultMessageAndSubject/string.Empty`

Common/Util/Validate.cs
```diff
@@ -0,0 +1,106 @@
...
+                return null;
+            }
+
+            return emailAddress;
+        }
+
+        /// <summary>
+        /// Provides static storage of compiled regular expressions to preclude parsing on each invocation
+        /// </summary>
+        public static class RegularExpression
```
Martin-Molinero: nit: wondering if it's worth for this guy to be public and not private, thoughts?

mchandschuh: I had gone back and forth on this one and figured exposing it doesn't do any harm. The added benefit to exposing it is there can be capture groups in the regular expressions that can be useful depending on what you're doing. We would need to expose methods for fetching the capture groups for each regex. This behavior isn't currently used, but was the reasoning. I'm happy to move it to private as well.

Common/Notifications/Notification.cs
mchandschuh: Agreed. Notify docs should be updated as well as this constructor's documentation.

Tests/Common/Notifications/NotificationEmailTests.cs
mchandschuh: will fix.

mchandschuh:
> 
> `Notify.Email` fails with a system log at https://github.com/QuantConnect/Lean/blob/master/Engine/Results/LiveTradingResultHandler.cs#L1120 . Wondering if these failures should reach the users through a log/message in the cloud console, maybe through https://github.com/QuantConnect/Lean/blob/master/Engine/Results/LiveTradingResultHandler.cs#L450 ? Thinking about the other type of notification failures and other potential failures with the email type.

Great point. We should likely be exposing these to the user via the algorithm's logging facilities (or via result handler). Are there any concerns regarding leakage of details of code outside of this repository. For example, in production, we run a custom messaging handler in the cloud, so maybe that implementation needs to be updated to catch errors and raise them in a user-friendly way (perhaps there's a dedicated error type that could hold a user-friendly message)

Common/Util/Validate.cs
Martin-Molinero: Agree, there is no one right choice here, personally, I'd keep it private until required but it's good as it too 👍

Martin-Molinero:
> Great point. We should likely be exposing these to the user via the algorithm's logging facilities (or via result handler). Are there any concerns regarding leakage of details of code outside of this repository. For example, in production, we run a custom messaging handler in the cloud, so maybe that implementation needs to be updated to catch errors and raise them in a user-friendly way (perhaps there's a dedicated error type that could hold a user-friendly message)

Good point! Yeah, agree, it's more of a cloud error handling responsability 👍, so not for this PR. The cloud messaging handler is actually handling errors for custom notifications, should just be expanding it to cover all type of notifications.

jaredbroad:
Lots of nice validation goodies in this PR but would appreciate if we could pull them to QCAlgorithm.Notify.Email() as input validation steps instead of trying to solve deep in LEAN? A throw on a bad email address would easily reach user if inside API layer/Notify.Email() method. 

Recommend forcing good inputs instead of supporting nulls. Force the 1% of people who want null subjects to use string.Empty / set their own defaults.

mchandschuh:
@jaredbroad RE: validation in Notify.Email - - isn't that where it is? This PR adds validation to the notify email constructor which is invoked via the Notification Manager which in turn is directly exposed via IAlgorithm [here](https://github.com/QuantConnect/Lean/blob/0093501af4a2defff5fed0173cbc73187419991b/Common/Interfaces/IAlgorithm.cs#L128)

PR #3375, opened 2019-07-08
Prevent IB cash sync error during weekends

### LINKED ISSUE #3395: Cash Sync Maximum Retries Reached
#### Expected Behavior
The maximum number of attempts for brokerage cash sync should only be reached in when there is a connection or brokerage API issue.

#### Actual Behavior
The maximum number of attempts for brokerage cash sync has been reached when it should not be attempted.

#### Potential Solution
Review conditions for attempting retries.

#### Checklist

StefanoRaggi:
#### Description
During weekends we cannot perform cash sync with IB because algorithms are disconnected. To solve this problem the new `IBrokerage.CanPerformCashSynchronization` property has been added, allowing any brokerage implementation to decide if cash sync can be performed successfully by the `BrokerageTransactionHandler` at the current time.

#### Related Issue
Closes #3395 

#### Motivation and Context
Occasional IB live algorithm runtime error during weekends.

No.

Live algorithm deployed to IB and verified that cash sync is not performed during the weekend.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
+++ b/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
@@ -629,6 +629,22 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
                     return false;
                 }
 
+                // check if the brokerage is able to perform cash sync at this time
+                try
+                {
+                    if (!_brokerage.CanPerformCashSynchronization)
+                    {
+                        Log.Trace("BrokerageTransactionHandler.PerformCashSync(): Reentrant call, cash sync not performed");
+                        _syncedLiveBrokerageCashToday = true;
+                        _lastSyncTimeTicks = CurrentTimeUtc.Ticks;
+                        return true;
+                    }
+                }
+                catch (Exception exception)
+                {
+                    Log.Error(exception, "Error in CanPerformCashSynchronization:");
+                }
+
                 Log.Trace("BrokerageTransactionHandler.PerformCashSync(): Sync cash balance");
 
                 var balances = new List<CashAmount>();
@@ -685,6 +701,7 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
                         _algorithm.Portfolio.CashBook[cash.Symbol].SetAmount(0);
                     }
                 }
+
                 _syncedLiveBrokerageCashToday = true;
                 _lastSyncTimeTicks = CurrentTimeUtc.Ticks;
             }
--- a/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
+++ b/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
@@ -634,7 +634,7 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
                 {
                     if (!_brokerage.CanPerformCashSynchronization)
                     {
-                        Log.Trace("BrokerageTransactionHandler.PerformCashSync(): Reentrant call, cash sync not performed");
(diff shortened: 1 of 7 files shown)


Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -629,6 +629,22 @@ private bool PerformCashSync()
...
                 }
 
+                // check if the brokerage is able to perform cash sync at this time
+                try
+                {
+                    if (!_brokerage.CanPerformCashSynchronization)
+                    {
+                        Log.Trace("BrokerageTransactionHandler.PerformCashSync(): Brokerage cannot perform cash sync at this time");
+                        _syncedLiveBrokerageCashToday = true;
+                        _lastSyncTimeTicks = CurrentTimeUtc.Ticks;
```
jaredbroad: This will cause a full day delay on cash sync and the equity curve will spike unreliably. If its repeatably being called at the same time each day (and its always closed) then it will actually never sync. 

Should rethink this and design a more robust "international" solution.

jaredbroad: Perhaps we need a map of IB-server locations to server reset times and we can move the IsWithinResetTimes() logic to be based on the server location.

StefanoRaggi: Why do you think it should delay/spike?
Cash sync is currently scheduled to be at `7:45 AM NY` every day, the only times `CanPerformCashSynchronization` would return `false` would be on Saturday and Sunday mornings (when cash sync cannot be performed anyway).

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -629,6 +629,22 @@ private bool PerformCashSync()
                     return false;
                 }
 
+                // check if the brokerage is able to perform cash sync at this time
+                try
+                {
+                    if (!_brokerage.CanPerformCashSynchronization)
+                    {
+                        Log.Trace("BrokerageTransactionHandler.PerformCashSync(): Brokerage cannot perform cash sync at this time");
+                        _syncedLiveBrokerageCashToday = true;
```
mchandschuh: This seems misleading. [Here](https://github.com/QuantConnect/Lean/pull/3375/files#diff-70048ea847fdd91ea15231ea598ecfd1R635) we say we can't perform the cash sync but then [here](https://github.com/QuantConnect/Lean/pull/3375/files#diff-70048ea847fdd91ea15231ea598ecfd1R638) we say we just performed it today. I'm almost thinking it's better to constantly recheck every _n_ minutes or so until we can perform the cash sync instead of giving up for an entire day.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
mchandschuh: RE: _rethink_ -- I know this is currently only an IB-centric issue, but it does seem as though the peculiarities of the brokerage implementations are leaking into the transaction handler. Perhaps more generally would be to have brokers perform their own cash syncs (via raising events) -- or maybe a much smaller and less risky change, the brokerage implementation could tell the transaction handler at what time the next sync is. The default implementation can just always return the next `7:45AM NY` -- but then internationally sensitive brokerages, such as IB, can return a time that properly observes the scheduled server outages.

StefanoRaggi:
Closing for now as a new solution is in the works.

PR #3376, opened 2019-07-09
ReflectionTypeLoadException exception handling improvement.

IlshatGaripov:
#### Description
A) I ran into the problem recently when loading specific dll with an algorithm to the compute node with lean engine installed. It appeared that one of the algorithms in assembly depended on different dll - that was not installed on the node and this caused ReflectionTypeLoadException to be thrown when call assembly.GetTypes() inside Loader.GetExtendedTypeNames()

More than that assemblyTypes = e.Types collection when exception is thrown may contain null values and then when passed to the next linq query  NullReferenceException will be thrown unexpectedly (from t in assemblyTypes where t.IsClass... ) To protect from this we need to exclude null values

Similiar problem discussion is here:
https://stackoverflow.com/questions/7889228/how-to-prevent-reflectiontypeloadexception-when-calling-assembly-gettypes

B) Additionally, R# suggests that this block of code can be removed because always false:
if (assembly == null)
                {
                    ...
                }
and this is probably true as I could read from documentation both Load and LoadFrom will either return Assembly if they manage to find and load it or will throw one of possible exceptions . 

#### Motivation and Context
Make it less unexpected errors in lean engine, make the error handling easier and more verbose

#### Types of changes

#### Checklist:

### DIFF
--- a/AlgorithmFactory/Loader.cs
+++ b/AlgorithmFactory/Loader.cs
@@ -313,7 +313,7 @@ namespace QuantConnect.AlgorithmFactory
                     // See https://stackoverflow.com/questions/7889228/how-to-prevent-reflectiontypeloadexception-when-calling-assembly-gettypes
                     assemblyTypes = e.Types.Where(t => t != null).ToArray();
 
-                    var countTypesNotLoaded = e.Types.Count(t => t == null);
+                    var countTypesNotLoaded = e.LoaderExceptions.Length;
                     Log.Error($"Loader.GetExtendedTypeNames(): Unable to load {countTypesNotLoaded} of the requested types, " +
                               "see below for more details on what causes an issue:");
 
--- a/AlgorithmFactory/Loader.cs
+++ b/AlgorithmFactory/Loader.cs
@@ -309,7 +309,13 @@ namespace QuantConnect.AlgorithmFactory
                 }
                 catch (ReflectionTypeLoadException e)
                 {
+                    // We may want to exclude possible null values
+                    // See https://stackoverflow.com/questions/7889228/how-to-prevent-reflectiontypeloadexception-when-calling-assembly-gettypes
                     assemblyTypes = e.Types.Where(t => t != null).ToArray();
+
+                    var countTypesNotLoaded = e.Types.Count(t => t == null);
+                    Log.Error($"Loader.GetExtendedTypeNames(): Unable to load {countTypesNotLoaded} of the requested types, " +
+                              "see below for more details on what causes an issue:");
 
                     foreach (Exception inner in e.LoaderExceptions)
                     {
--- a/AlgorithmFactory/Loader.cs
+++ b/AlgorithmFactory/Loader.cs
@@ -241,12 +241,6 @@ namespace QuantConnect.AlgorithmFactory
                     var assemblyBytes = File.ReadAllBytes(assemblyPath);
                     assembly = Assembly.Load(assemblyBytes, debugInformationBytes);
                 }
-                if (assembly == null)
-                {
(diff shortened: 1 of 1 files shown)


AlgorithmFactory/Loader.cs
```diff
@@ -241,12 +241,6 @@ private bool TryCreateILAlgorithm(string assemblyPath, out IAlgorithm algorithmI
                     var assemblyBytes = File.ReadAllBytes(assemblyPath);
                     assembly = Assembly.Load(assemblyBytes, debugInformationBytes);
                 }
-                if (assembly == null)
```
mchandschuh: Gotta love R#! `Assembly.Load` and `Assembly.LoadFrom` guarantee a non-null result or an exception 👍

AlgorithmFactory/Loader.cs
```diff
@@ -315,7 +309,12 @@ public static List<string> GetExtendedTypeNames(Assembly assembly)
                 }
                 catch (ReflectionTypeLoadException e)
                 {
-                    assemblyTypes = e.Types;
+                    assemblyTypes = e.Types.Where(t => t != null).ToArray();
+
+                    foreach (Exception inner in e.LoaderExceptions)
+                    {
+                        Log.Error($"Loader.GetExtendedTypeNames(): {inner.Message}");
```
mchandschuh: Do we also want to log `e.Message` here as well?

AlgorithmFactory/Loader.cs
```diff
@@ -315,7 +309,12 @@ public static List<string> GetExtendedTypeNames(Assembly assembly)
                 }
                 catch (ReflectionTypeLoadException e)
                 {
-                    assemblyTypes = e.Types;
+                    assemblyTypes = e.Types.Where(t => t != null).ToArray();
```
mchandschuh: Very interesting! That's certainly unexpected behavior, I wonder why it behaves that way. I suppose the null entries are the ones where the type didn't load. Perhaps a count of the null entries for the log would also be beneficial? I wonder if `e.LoaderExceptions` has an item for null type, i.e, for each type load failure.

Also, thank you for providing extra context via the linked stack overflow. I think it's good practice to include such a reference either in the source code or in the commit message as well for posterity.

AlgorithmFactory/Loader.cs
IlshatGaripov: This message always says: **Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.'**  I will include similiar message to the logs along with count of the null entries .

AlgorithmFactory/Loader.cs
IlshatGaripov: It seems indeed that null values in collection do match the types that could not be loaded. And for each such type there is an exception object in LoaderExceptions  collection. Thanks for a review. I made the minor edits according to comments received. commit with those changes follows  ->

IlshatGaripov: I use this to count the number of items could not be loaded:
**var countTypesNotLoaded = e.Types.Count(t => t == null);**
Wondering should this be more efficient (faster):
**var countTypesNotLoaded = e.LoaderExceptions.Length;**
if there are exactly as many LoaderExceptions as the types that could not be loaded ?

IlshatGaripov: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.reflectiontypeloadexception.loaderexceptions?view=netframework-4.8 says about ReflectionTypeLoadException.LoaderExceptions Property:  The null values in the Types array of this instance line up with the exceptions in this array. So I think it's better use LoaderExceptions.Length.

PR #3336, opened 2019-06-21
Fix LiveEquityDataSynchronizingEnumerator for input data after the current time

StefanoRaggi:
#### Description
Previously the enumerator would get stuck and stop emitting data if one of the underlying enumerators returned a data point with the time greater than the current time. The existing unit test would only emit the first data point for the two underlying streams.

The enumerator has been updated to support data points in the future and the unit test has been extended to assert both data point counts and values.

#### Related Issue
Closes #3335 

#### Motivation and Context
`LiveEquityDataSynchronizingEnumerator` not emitting data in some cases.

#### Requires Documentation Change
No.

#### How Has This Been Tested?
- The existing unit test has been expanded to cover the failing case.
- An equity live algorithm has also been tested in the cloud

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #3381, opened 2019-07-10
Fix aggregation with aligned ticks in Trade/QuoteBarBuilderEnumerator

StefanoRaggi:
#### Description
This PR fixes an issue when processing ticks with a timestamp aligned with the bar period and only one tick per bar period. In this case incorrect bars would be emitted, by grouping every two consecutive ticks in the same bar.
This edge-case scenario is very unlikely to happen in live trading, but quite common in unit tests.

> Note to reviewers: This PR includes the changes in #3336 (since an upcoming PR with new LiveTradingDataFeed tests requires both fixes)

#### Related Issue
n/a

#### Motivation and Context
Incorrect bar aggregation in unit tests.

No.

New unit test included.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/Enumerators/QuoteBarBuilderEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/QuoteBarBuilderEnumerator.cs
@@ -84,6 +84,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
                 var utcNow = _timeProvider.GetUtcNow();
                 var currentLocalTime = utcNow.ConvertFromUtc(_timeZone);
                 var barStartTime = currentLocalTime.RoundDown(_barSize);
+
+                if (barStartTime == currentLocalTime)
+                {
+                    // adjust start time if tick aligned to bar period
+                    barStartTime = currentLocalTime.Subtract(_barSize);
+                }
+
                 working = new QuoteBar();
                 working.Update(data.Value, bidPrice, askPrice, qty, bidSize, askSize);
                 working.Period = _barSize;
--- a/Engine/DataFeeds/Enumerators/TradeBarBuilderEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/TradeBarBuilderEnumerator.cs
@@ -79,6 +79,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
                 var utcNow = _timeProvider.GetUtcNow();
                 var currentLocalTime = utcNow.ConvertFromUtc(_timeZone);
                 var barStartTime = currentLocalTime.RoundDown(_barSize);
+
+                if (barStartTime == currentLocalTime)
+                {
+                    // adjust start time if tick aligned to bar period
+                    barStartTime = currentLocalTime.Subtract(_barSize);
+                }
+
                 working = new TradeBar(barStartTime, data.Symbol, marketPrice, marketPrice, marketPrice, marketPrice, qty, _barSize);
                 _queue.Enqueue(working);
 
--- a/Engine/DataFeeds/Enumerators/LiveEquityDataSynchronizingEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/LiveEquityDataSynchronizingEnumerator.cs
@@ -32,8 +32,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
         private readonly DateTimeZone _exchangeTimeZone;
         private readonly IEnumerator<BaseData> _auxDataEnumerator;
         private readonly IEnumerator<BaseData> _tradeBarAggregator;
-        private bool _auxDataEnumeratorAdvanced;
-        private bool _tradeBarEnumeratorAdvanced;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LiveEquityDataSynchronizingEnumerator"/> class
@@ -48,13 +46,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
             _exchangeTimeZone = exchangeTimeZone;
             _auxDataEnumerator = auxDataEnumerator;
             _tradeBarAggregator = tradeBarAggregator;
-
-            // prime enumerators
-            _auxDataEnumerator.MoveNext();
-            tradeBarAggregator.MoveNext();
-
-            _auxDataEnumeratorAdvanced = true;
-            _tradeBarEnumeratorAdvanced = true;
         }
 
         /// <summary>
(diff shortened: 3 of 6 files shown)


Engine/DataFeeds/Enumerators/QuoteBarBuilderEnumerator.cs
```diff
@@ -84,6 +84,13 @@ public void ProcessData(BaseData data)
                 var utcNow = _timeProvider.GetUtcNow();
                 var currentLocalTime = utcNow.ConvertFromUtc(_timeZone);
                 var barStartTime = currentLocalTime.RoundDown(_barSize);
+
+                if (barStartTime == currentLocalTime)
+                {
+                    // adjust start time if tick aligned to bar period
+                    barStartTime = currentLocalTime.Subtract(_barSize);
```
Martin-Molinero: Won't `barStartTime = currentLocalTime.Subtract(_barSize);` cause us to emit _right away_ (when queried), since `EndTime = barStartTime + _barSize = currentLocalTime - _barSize + _barSize = currentLocalTime`, doesn't seem right from the perspective of the `BuilderEnumerator`, since it won't repect bar size?

1 Hour bar size example of the issue:
- Initial tick (1) at 10:00 AM. Quote bar starts at 10:00 AM and ends at 11:00 AM
- Second tick (2) at 11:00 AM
- Race condition between removing the `working` quote bar (1) L124 and adding a new one or not L81, this is the issue I understand we are trying to fix, right @StefanoRaggi ?

StefanoRaggi: The issue found when writing the new tests in #3384 was that bars were emitted once every two bars (i.e. bar ending at 11:00 emitted, bar ending at 12:00 not emitted, bar ending at 13:00 emitted, etc.) and each bar emitted was including two ticks instead of one. For this reason, I tried to repeat the issue with a new unit test for this enumerator -- `AggregatesHourlyTicksIntoHourlyBars`, included in this PR.

Not sure if it's actually a race condition since `liveMode` is set to `false` in this test so the timer in the `RealTimeScheduleEventService` is disabled and `MoveNext` is only called by the test.

Martin-Molinero: Modified the `AggregatesHourlyTicksIntoHourlyBars` it passes in `master`: `.MoveNext()` before adding the new data point so we get the finished working bar and the new data point generates a new bar

```
       [Test]
        public void AggregatesHourlyTicksIntoHourlyBars()
        {
            var timeProvider = new ManualTimeProvider(TimeZones.NewYork);
            var enumerator = new TradeBarBuilderEnumerator(Time.OneHour, TimeZones.NewYork, timeProvider, false);

            // noon new york time
            var currentTime = new DateTime(2015, 10, 08, 12, 0, 0);
            timeProvider.SetCurrentTime(currentTime);

            var price = 0m;

            for (var i = 0; i < 5; i++)
            {
                // advance one hour
                currentTime = currentTime.AddHours(1);
                timeProvider.SetCurrentTime(currentTime);

                // we advanced time, a new bar should be generated
                Assert.IsTrue(enumerator.MoveNext());

                // the first loop no trade bar was generated yet, end time not reached
                if (i > 0)
                {
                    Assert.IsNotNull(enumerator.Current);

                    var bar = (TradeBar)enumerator.Current;
                    Assert.AreEqual(currentTime.AddHours(-1), bar.Time);
                    Assert.AreEqual(currentTime, bar.EndTime);
                    Assert.AreEqual(price, bar.Open);
                    Assert.AreEqual(price, bar.High);
                    Assert.AreEqual(price, bar.Low);
                    Assert.AreEqual(price, bar.Close);
                }

                // add a tick, will generate a new bar
                price++;
                enumerator.ProcessData(new Tick(currentTime, Symbols.SPY, price, price, price));
            }
```

StefanoRaggi:
@Martin-Molinero Agreed, I should close this PR and perhaps add this unit test to PR #3384. 
I will also rebase #3384 to exclude the changes in this PR.

Engine/DataFeeds/Enumerators/TradeBarBuilderEnumerator.cs
```diff
@@ -79,6 +79,13 @@ public void ProcessData(BaseData data)
                 var utcNow = _timeProvider.GetUtcNow();
                 var currentLocalTime = utcNow.ConvertFromUtc(_timeZone);
                 var barStartTime = currentLocalTime.RoundDown(_barSize);
+
+                if (barStartTime == currentLocalTime)
```
mchandschuh: Is this correct? I think the builder enumerators construct bars on the interval `[start, end)` which implies that a tick that falls on an interval boundary belongs to the following interval and not the preceding interval. I think this is confirmed by the usage of `<=` [here](https://github.com/QuantConnect/Lean/pull/3381/files#diff-966cb3974d100f934938a28c9b66e9eaR124) and [here](https://github.com/QuantConnect/Lean/pull/3381/files#diff-7005b62745f9c26d610567b6c5c5b62bR119). Also, if you look at the implementation of the consolidators, they too would appear to treat a data point on an interval boundary as belonging to the following bar and not the preceding bar.

Also, I may just be reading this wrong, and if that's the case, let's add some more comments here to better describe why we're doing this and why it's the right thing to do.

Maybe more importantly for me (and the code base in general), is to confirm and document the rules around aggregation for posterity (and perhaps in the source as well).

mchandschuh:
Just thinking out loud...

I wonder if [this](https://github.com/QuantConnect/Lean/pull/3381/files#diff-b74c343bb026b3cf19b6014c361e135aR120) kind of frontier logic shouldn't be encapsulated via the `FrontierAwareEnumerator` or similar, so that the builder only need to do the null check.

I know this is existing code, so no need to change as part of this PR, but I think it's worth it to think about how we construct these enumerator stacks. The original design was to provide a decorator pattern such that each enumerator type doesn't need to concern itself with what comes before it, and I fear that invoking frontier checks directly in here might be mixing concerns.

That being said, the argument can certainly be made that the `LiveEquityDataSynchronizingEnumerator` is inherently a frontier check/sync between price and aux data, but then again, if that's _all_ this guy does, I'm wondering why we can't wrap the price/aux enumerators in a `FrontierAwareEnumerator` and then the `LiveEquityDataSynchronizingEnumerator` would simplify into your standard `SynchronizingEnumerator` (methinks). Again, no need to change, but I think there's value in not only questioning the structure of the enumerator stacks, but also providing documentation on why the enumerator stack is the way it is.

Also, it may end up being that the aforementioned suggestion is just over complicating what is a rather simple piece of code... but then again, it is apparently complicated enough to require changes :)

Engine/DataFeeds/Enumerators/TradeBarBuilderEnumerator.cs
StefanoRaggi: @mchandschuh You're right, the aggregation interval is open on the right, so this change is incorrect, as @Martin-Molinero also pointed out. This needs to be taken into account when writing unit tests with ticks on the interval boundary -- like `AggregatesHourlyTicksIntoHourlyBars`.

StefanoRaggi:
> I wonder if [this](https://github.com/QuantConnect/Lean/pull/3381/files#diff-b74c343bb026b3cf19b6014c361e135aR120) kind of frontier logic shouldn't be encapsulated via the `FrontierAwareEnumerator` or similar, so that the builder only need to do the null check.

Unfortunately, this PR was submitted including changes in #3336, which has now been merged independently. 
I _think_ the reason for having the frontier check in `LiveEquityDataSynchronizingEnumerator` was to handle receiving aux data earlier than expected (i.e. splits/dividends announced for a future date) although I verified our backend does not currently emit aux data ahead of time.

StefanoRaggi:
Closing as not-a-bug

Issue #997, opened 2017-06-29
[Crypto] Add Orderbook Support to LEAN


jaredbroad:
Build a full order book using quote ticks, trades and make it available on the Crypto security type.

ajaye2:
Are there any updates on when this will be completed?

jaredbroad:
Closing this as its not top of priority right now -- and if when implemented would be done for the whole stack not just crypto.

jaredbroad:
Reopening to gauge community interest - would be implemented for the entire stack when data is available. 

With plan focused on aggregate order book level streamed not the full book.

Martin-Molinero:
Potential solution plan
```
TickType.Quote => Best bid/ask price & qtty
	- bool TopOfBook { get; set; } = true;
Security.OrderBook => cache.OrderBook
	- lean engine does the tracking
```

ikamanu:
Yes please.  🙏

multidis:
Would be helpful +1

code-gif:
Yes please

PR #3388, opened 2019-07-10
Unify brokerage orderbook class

### LINKED ISSUE #3387: Unify OrderBook class
#### Expected Behavior

As we want to support more brokerages (#2003 , #2004 ) it's necessary to unify and reuse OrderBook.cs class.

#### Actual Behavior

Currently all crypto brokerages has its own orderbook implementation that are identical.

#### Potential Solution

Create an IOrderBook interface and maybe DefaultOrderBook class to cover basic needs.

#### Checklist

Adalyat:
#### Description

Create IOrderBook and DefaultOrderBook class in order to unify and reuse orderbook stuff for all brokerages.

#### Related Issue

Closes #3387 

#### Motivation and Context

It's required to reduce number of identical OrderBook classes we have right now

Unit tests; live trading

#### Types of changes

#### Checklist:


Brokerages/IOrderBook.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Linq;
+
+namespace QuantConnect.Brokerages
+{
+    /// <summary>
+    /// Represents an order book interface for a security.
+    /// </summary>
+    /// <typeparam name="K">Price level identifier</typeparam>
+    /// <typeparam name="V">Size at the price level</typeparam>
+    public interface IOrderBook<K, V>
```
StefanoRaggi: Awesome, the generic arguments are required to add support the Bitmex order book in the future 👍 

I think this interface should be renamed to `IOrderBookUpdater<K, V>` since it only has methods for updating the order book data. 
The non-generic `IOrderBook` interface can be added later on when addressing issue #997.

Brokerages/DefaultOrderBook.cs
```diff
@@ -17,19 +17,19 @@
...
     /// </summary>
-    public class OrderBook
+    public class DefaultOrderBook : IOrderBook<decimal, decimal>
     {
         private readonly object _locker = new object();
         private readonly Symbol _symbol;
-        private readonly SortedDictionary<decimal, decimal> _bids = new SortedDictionary<decimal, decimal>();
-        private readonly SortedDictionary<decimal, decimal> _asks = new SortedDictionary<decimal, decimal>();
+        protected readonly SortedDictionary<decimal, decimal> _bids = new SortedDictionary<decimal, decimal>();
+        protected readonly SortedDictionary<decimal, decimal> _asks = new SortedDictionary<decimal, decimal>();
```
StefanoRaggi: To follow the coding style used in our repo, only private class members should start with an underscore.
Please rename `_bids` to `Bids` and `_asks` to `Asks`

Brokerages/DefaultOrderBook.cs
```diff
@@ -153,21 +161,33 @@ public void RemoveBidRow(decimal price)
...
+        /// Common price level removal method
+        /// </summary>
+        /// <param name="priceLevel"></param>
+        public void RemovePriceLevel(decimal priceLevel)
+        {
+            if (Asks.ContainsKey(priceLevel))
+                RemoveAskRow(priceLevel);
+            else if (Bids.ContainsKey(priceLevel))
+                RemoveBidRow(priceLevel);
+        }
```
StefanoRaggi: Upon further review, I think we should `lock` the two `ContainsKey` calls, so access to the two dictionaries is synchronized everywhere.

mchandschuh:
RE: commit messages

Try to provide a descriptive subject line for each commit. Putting `[Requested Changes]` forces someone later on to have to dig into the full message. Commit messages provide a history of the code base and are invaluable when performing analysis and trying to understand how/why a piece of code evolved the way it did. Also, in general, commit messages should answer _why_ and not _what_ -- the diff will always provide the _what_changed, but the _why_ is often lost and turns out to be the more important piece of information months or years down the line when someone else comes in and tries to understand the code.

Brokerages/DefaultOrderBook.cs
```diff
@@ -57,10 +65,10 @@ public class OrderBook
         public decimal BestAskSize { get; private set; }
```
mchandschuh: Are there any concerns regarding read consistency of the best bid/ask properties? For example, it is currently possible for one thread to be reading the bid/ask while another thread is updating them, leading to a potentially inconsistent read of the data (for example, an old bid and a new ask).

Brokerages/DefaultOrderBook.cs
```diff
@@ -153,21 +161,36 @@ public void RemoveBidRow(decimal price)
...
+        /// <summary>
+        /// Common price level removal method
+        /// </summary>
+        /// <param name="priceLevel"></param>
+        public void RemovePriceLevel(decimal priceLevel)
+        {
+            lock (_locker)
+            {
+                if (Asks.ContainsKey(priceLevel))
+                    RemoveAskRow(priceLevel);
```
mchandschuh: The general style of the repository is to use braces, even for single line branches.

Brokerages/IOrderBookUpdater.cs
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Brokerages
+{
+    /// <summary>
+    /// Represents an order book interface for a security.
```
mchandschuh: This documentation seems to be inconsistent with the  behavior described in the interface. The interface provides the ability to _update_ and to be alerted about updates (via events), but _does not_ provide any read access to the actual order book data.

That being said, is there any particular reason why the bids and asks aren't exposed, for example, we _could_ add two methods, `GetBids` and `GetAsks` as `IEnumerable<KeyValuePair<K, V>>`

Also, it's OK if this abstraction is solely intended for providing updates, we'll just want to update the documentation here to match the defined behavior of the interface.

Brokerages/DefaultOrderBook.cs
```diff
@@ -153,21 +161,36 @@ public void RemoveBidRow(decimal price)
...
             }
 
             if (price == BestAskPrice)
             {
-                lock(_locker)
+                lock (_locker)
                 {
-                    BestAskPrice = _asks.Keys.FirstOrDefault();
-                    BestAskSize = BestAskPrice > 0 ? _asks[BestAskPrice] : 0;
+                    BestAskPrice = Asks.Keys.FirstOrDefault();
```
mchandschuh: I think you should be able to do `var kvp = Asks.FirstOrDefault()` and then use that to assign `BestAskPrice` and `BestAskSize` -- this precludes use from needing to access the dictionary twice. Same comment for the bids.

Brokerages/DefaultOrderBook.cs
Adalyat: I've been investigating this problem and found that we might want to use standard practice to avoid race condition and use (`lock(_locker)`) within get/set methods. I think it's good enough as our Properties are of value type. 
```
private int _myField;
private object _locker= new object();

public int MyProperty
{
    get
    {
        lock (_locker)
        {
            return _myField;
        }
    }
    set
    {
        lock (_locker)
        {
            _myField = value;
        }
    }
}
```

Any thoughts?

Brokerages/DefaultOrderBook.cs
Adalyat: Fixed

Brokerages/IOrderBookUpdater.cs
Adalyat: The main concern of this decision was #997. We might want to implement `IOrderBook ` to Lean core and make it more specific (as `SortedDictionary<decimal, decimal>` ) in order to use orderbook within algorithms. Having `GetBids` and `GetAsks` as generic methods is good enough for using them within brokerages and/or brokerage-specific algorithms, but wouldn't like to introduce this limitation and keep all algorithms compatible with all brokerages.

Brokerages/DefaultOrderBook.cs
StefanoRaggi: I agree that read access to these properties should be synchronized, maybe it would be better to have only public getters with `lock` and the update methods can just update the backing fields (also with `lock`):

``` C#
private decimal _bestBidPrice;

public decimal BestBidPrice 
{ 
    get 
    {
        lock (_locker)
        {
            return _bestBidPrice;
        }
    }
}
```

Brokerages/DefaultOrderBook.cs
```diff
@@ -89,9 +97,9 @@ public void Clear()
...
         public void UpdateBidRow(decimal price, decimal size)
         {
-            lock(_locker)
+            lock (_locker)
             {
-                _bids[price] = size;
+                Bids[price] = size;
             }
 
             if (BestBidPrice == 0 || price >= BestBidPrice)
```
StefanoRaggi: Since we should synchronize access to all `Bestxxx` properties, we should also move this `if` block inside the `lock` block above -- same for `UpdateAskRow`

Issue #3401, opened 2019-07-16
Improve TradingEconomics Downloaders


gsalaz98:
#### Expected Behavior

We are able to:
* Use parallel processing to speed up writing files to disk
* Write files to appropriate directories to clean up data directory
* Not write files to the final destination unless the download and compression complete
* Have retry logic in case an HTTP request fails
* Have Improved logging
* Have better existing file detection and skipping logic
* Write date aware filenames for data
* Ensure data consistency in live trading and backtesting by only writing data with a present `actual` field
#### Actual Behavior

Expected behavior is implemented
#### Potential Solution

Implement the Expected Behavior
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [ ] I have provided detailed steps to reproduce the issue

PR #3402, opened 2019-07-16
Improves TradingEconomics Downloaders

gsalaz98:
#### Description

* Add rategate to all requests so that we don't run over rate limits
* Improve logging across all TradingEconomics Downloaders
* Add retries to HTTP requests due to occassional failure
* Added skipping of duplicates
* Added skipping of fields with no actual value
* Added temp file writing, temp file moving
* Write to proper output directory
* Update data directory structure
* Update GetSource in BaseData implementation for TradingEconomics
* Parallelize data processing and writing to disk

#### Related Issue

#3401 
#### Motivation and Context

Alternative data efforts

Maybe - Directory structure has changed, but functionally everything else is the same

Ran TradingEconomics algorithm locally and tested downloaders locally.

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
+++ b/ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
@@ -14,11 +14,14 @@
 */
 
 using Newtonsoft.Json;
+using QuantConnect.Data.Auxiliary;
 using QuantConnect.Data.Custom.TradingEconomics;
 using QuantConnect.Logging;
+using QuantConnect.Util;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
@@ -33,12 +36,16 @@ namespace QuantConnect.ToolBox.TradingEconomicsDataDownloader
         private readonly string _destinationFolder;
         private readonly DateTime _fromDate;
         private readonly DateTime _toDate;
+        private readonly RateGate _requestGate;
 
         public TradingEconomicsCalendarDownloader(string destinationFolder)
         {
             _fromDate = new DateTime(2000, 10, 01);
             _toDate = DateTime.Now;
             _destinationFolder = destinationFolder;
+            // Rate limits on Trading Economics is one request per second
+            _requestGate = new RateGate(1, TimeSpan.FromSeconds(1));
+
             Directory.CreateDirectory(_destinationFolder);
         }
 
@@ -48,8 +55,23 @@ namespace QuantConnect.ToolBox.TradingEconomicsDataDownloader
         /// <returns>True if process all downloads successfully</returns>
         public override bool Run()
         {
+            Log.Trace("TradingEconomicsCalendarDownloader.Run(): Begin downloading calendar data");
+
+            // Create the destination directory so that we don't error out in case there's no data
+            Directory.CreateDirectory(Path.Combine(_destinationFolder, "calendar"));
+
             var stopwatch = Stopwatch.StartNew();
             var data = new List<TradingEconomicsCalendar>();
+            var availableFiles = Directory.GetFiles(Path.Combine(_destinationFolder, "calendar"), "*.zip", SearchOption.AllDirectories)
+                .Where(
+                    x =>
+                    {
+                        DateTime _;
+                        return DateTime.TryParseExact(Path.GetFileName(x).Substring(0, 8), "yyyyMMdd", CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out _);
+                    }
+                )
+                .Select(x => DateTime.ParseExact(Path.GetFileName(x).Substring(0, 8), "yyyyMMdd", CultureInfo.InvariantCulture))
+                .ToHashSet();
 
             var startUtc = _fromDate;
             while (startUtc < _toDate)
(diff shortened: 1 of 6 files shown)


ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
```diff
@@ -48,52 +55,126 @@ public TradingEconomicsCalendarDownloader(string destinationFolder)
...
         {
+            Log.Trace("TradingEconomicsCalendarDownloader.Run(): Begin downloading calendar data");
+
+            // Create the destination directory so that we don't error out in case there's no data
+            Directory.CreateDirectory(Path.Combine(_destinationFolder, "calendar"));
+
             var stopwatch = Stopwatch.StartNew();
             var data = new List<TradingEconomicsCalendar>();
+            var availableFiles = Directory.GetFiles(Path.Combine(_destinationFolder, "calendar"), "*.zip", SearchOption.AllDirectories)
+                .Where(
```
Martin-Molinero: I think this is parsing the `date` twice? Maybe could `Select` first, something like:
```suggestion
                var availableFiles = Directory.GetFiles(Path.Combine(_destinationFolder, "calendar"), "*.zip", SearchOption.AllDirectories)
                .Select(files =>
                    {
                        try
                        {

                            return DateTime.ParseExact(
                                    Path.GetFileName(files).Substring(0, 8),
                                    "yyyyMMdd",
                                    CultureInfo.InvariantCulture);
                        }
                        catch (Exception)
                        {
                            return DateTime.MaxValue;
                        }
                    }
                )
                .ToHashSet(time => time != DateTime.MaxValue);
```

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsIndicatorDownloader.cs
```diff
@@ -67,44 +94,94 @@ public override bool Run()
...
 
-                foreach (var kvp in data.GroupBy(GetFileName))
-                {
-                    var path = Path.Combine(_destinationFolder, kvp.Key);
-                    var zipPath = path.Replace(".json", ".zip");
+                // Return status code. We default to `true` so that we can identify if an error occured during the loop
+                var status = true;
 
-                    try
+                Parallel.ForEach(data.GroupBy(GetTicker),
```
Martin-Molinero: Not sure if possible, but there seems to be some logical code duplication in this method and `GetTicker` between the different `TradingEconomicsYYYYYYDownloaders` maybe there could be a base downloader with some of this logic?

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsDataDownloader.cs
```diff
@@ -48,31 +49,36 @@ protected TradingEconomicsDataDownloader()
...
-                    response.EnsureSuccessStatusCode();
+                        response.EnsureSuccessStatusCode();
 
-                    return await response.Content.ReadAsStringAsync();
+                        return await response.Content.ReadAsStringAsync();
+                    }
+                }
+                catch (HttpRequestException e)
+                {
+                    Log.Trace($"TradingEconomicsDataDownloader.HttpRequester(): HTTP Request failed with message: {e.Message} - (attempt {retries} / 5). Retrying...");
```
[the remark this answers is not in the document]
gsalaz98: From my testing, the request would succeed if it was retried immediately after it previously failed. I'll consider adding sleep between retries 👍

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
```diff
@@ -48,52 +55,126 @@ public TradingEconomicsCalendarDownloader(string destinationFolder)
...
             {
                 try
                 {
                     var endUtc = startUtc.AddMonths(1).AddDays(-1);
+
+                    Log.Trace($"TradingEconomicsCalendarDownloader.Run(): Collecting calendar data from {startUtc:yyyy-MM-dd} to {endUtc:yyyy-MM-dd}");
+
+                    if (availableFiles.Contains(endUtc))
+                    {
+                        Log.Trace($"TradingEconomicsCalendarDownloader.Run(): Skipping data because it already exists for month: {startUtc:MMMM}");
```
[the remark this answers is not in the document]
gsalaz98: Too bad it doesn't work for files before 2013 in practice :(

gsalaz98:
@Jay-Jay-D In order to maintain directory structure for Github, we must have those files (i.e. if those files are missing, they won't show up on here).

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
```diff
@@ -48,52 +55,129 @@ public TradingEconomicsCalendarDownloader(string destinationFolder)
...
+                        continue;
+                    }
+
+                    _requestGate.WaitToProceed(TimeSpan.FromSeconds(1));
+
                     var content = Get(startUtc, endUtc).Result;
                     var collection = JsonConvert.DeserializeObject<List<TradingEconomicsCalendar>>(content);
 
-                    data.AddRange(collection);
+                    // Only write data that contains the "actual" field so that we get the final
```
Jay-Jay-D: Can you please clarify this comment?

What the "actual" field means and why it getting the "final piece of unchanging data" can affect the backward consistency?

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsEarningsDownloader.cs
```diff
@@ -108,10 +151,32 @@ public override Task<string> Get(DateTime startUtc, DateTime endUtc)
...
         }
 
-        private string GetFileName(TradingEconomicsEarnings tradingEconomicsEarnings)
+        /// <summary>
+        /// Gets the ticker using map files. If the ticker is empty, we can't resolve a map file, or we can't
+        /// resolve a ticker within a map file, we return null
+        /// </summary>
+        /// <param name="tradingEconomicsEarnings">TE Earnings data</param>
+        /// <returns>Mapped ticker or null</returns>
+        private string GetMappedSymbol(TradingEconomicsEarnings tradingEconomicsEarnings)
```
Jay-Jay-D: Lovely!

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsIndicatorDownloader.cs
```diff
@@ -67,44 +98,94 @@ public override bool Run()
...
                     {
-                        Log.Error(e, $"TradingEconomicsIndicatorsDownloader(): Error creating {path}");
-                        return false;
+                        // Create the destination directory, otherwise we risk having it fail when we move
+                        // the temp file to its final destination
+                        Directory.CreateDirectory(Path.Combine(_destinationFolder, kvp.Key));
+
+                        foreach (var indicatorByDate in kvp.GroupBy(x => x.LastUpdate))
+                        {
+                            var date = indicatorByDate.Key.ToString("yyyyMMdd");
```
Jay-Jay-D: In cases like this is preferable to use `indicatorByDate.Key.ToString(DateFormat.EightCharacter)`.

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
gsalaz98: For future reference:

TE goes back and updates their dataset whenever they get data for things like Unemployment, Productivity, etc. However, they do not have any historical snapshots for how data was before they went back and updated the data. Because of this, we decided that the best way forward to maintain consistency with the data in live trading and backtesting was to only include published data instead of data for an event that hasn't occurred yet.

Issue #3404, opened 2019-07-19
Improve Estimize Downloaders


gsalaz98:
#### Expected Behavior

Improved logging experience, file writing behavior, and skipping of duplicates to save time spent processing
#### Actual Behavior

None of the above exist

#### Potential Solution

Implement the Expected Behavior
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [ ] I have provided detailed steps to reproduce the issue

PR #3405, opened 2019-07-19
Improves Estimize Downloaders

gsalaz98:
#### Description

* Add completion tracking in Estimize downloaders (excluding consensus)
* Add better logging to estimize downloaders
* Speed up estimize consensus downloader by checking for duplicates
* Move duplicated code to single method
#### Related Issue

#3404 
#### Motivation and Context

More support in LEAN for alternative data sources

No

Tested locally on Windows 10
#### Types of changes

#### Checklist:


gsalaz98:
This is not yet ready for merging or review - I'm currently implementing mapping for tickers and discovered another issue similar to the one fixed in the TradingEconomics downloader:

https://github.com/QuantConnect/Lean/blob/8e256e202a68588aa4d9552b79a09b4fe0a0bda0/ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs#L100-L112

Will have a fix out soon once I've finished verifying that there's no issues with the data with the new transformations we will apply

gsalaz98:
Ready for review 👍

Common/Data/Custom/Estimize/EstimizeEstimate.cs
```diff
@@ -98,6 +99,33 @@ public class EstimizeEstimate : BaseData
...
+        public EstimizeEstimate(string csvLine)
+        {
+            // CreatedAt[0], Id[1], AnalystId[2], UserName[3], FiscalYear[4], FiscalQuarter[5], Eps[6], Revenue[7], Flagged[8]"
+            var csv = csvLine.Split(',');
+
+            CreatedAt = DateTime.ParseExact(csv[0], "yyyyMMdd HH:mm:ss", CultureInfo.InvariantCulture);
+            Id = csv[1];
+            AnalystId = csv[2];
+            UserName = csv[3];
+            FiscalYear = Convert.ToInt32(csv[4], CultureInfo.InvariantCulture);
```
jaredbroad: Nice use of invariant

gsalaz98:
RateGate threw an error while downloading the data. Perhaps a(n) issue(s) should be opened as it pertains to this user's comment in: http://www.jackleitch.net/2010/10/better-rate-limiting-with-dot-net/

```
Jim says	

November 20, 2013 at 7:36 pm

Jack, I’ve noticed that the implementation seems suffer from two flaws:
First: The implementation of the ExitTimerCallback checks the queue during a while() loop, but then again at a subsequent if(). This is unnecessary, but also this if()’s TryPeek() output may be past-due, resulting in a negative number in timeUntilNextCheck.

Second: Environment.TickCount rolls over every 24.9 days, which can also mess with constant-up service deployments.
```

I've added extra logic to prevent RateGate failures from tanking the download process should it fail

PR #3431, opened 2019-07-25
SPY default benchmark security

Martin-Molinero:
#### Description

**Commit 1**
- Setting SPY as the default security benchmark
- The security benchmark subscription will be added at `UniverseSelection` as an internal subscription. Using its own dedicated `Security` instance which doesn't live in the `algorithms.Securities` collection.
- Reducing algorithms exposure to internal subscriptions
- `TimeSliceFactory` will prioritize higher resolution bars, when same
symbol is present twice (for non-internal subscriptions)
- Adding regression test `CustomUniverseWithBenchmarkRegressionAlgorithm`

**Commit 2**
- Adjust algorithms statistics

Found inspiration from PR https://github.com/QuantConnect/Lean/pull/2403
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/2381
Closes https://github.com/QuantConnect/Lean/issues/2260
Closes https://github.com/QuantConnect/Lean/issues/2153
#### Motivation and Context

Fix incorrect behavior when setting a security benchmark

N/A

Unit and regression tests
#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.cs
```diff
@@ -544,31 +544,16 @@ public virtual void PostInitialize()
...
-                    var startingCapital = Portfolio.TotalPortfolioValue;
-                    Benchmark = new FuncBenchmark(dt =>
-                    {
-                        var years = (dt - start).TotalDays / 365.25;
-                        return startingCapital * (decimal) Math.Exp(0.02 * years);
-                    });
+                    _benchmarkSymbol = QuantConnect.Symbol.Create("SPY", SecurityType.Equity, Market.USA);
                 }
+
+                var security = Securities.CreateSecurity(_benchmarkSymbol,
```
mchandschuh: Do we want to check the `Securities` collection to see if it already exists?

Algorithm/QCAlgorithm.cs
```diff
@@ -1094,6 +1095,10 @@ public void SetBenchmark(Symbol symbol)
         /// <param name="benchmark">The benchmark producing function</param>
         public void SetBenchmark(Func<DateTime, decimal> benchmark)
         {
+            if (_locked)
```
mchandschuh: Nice additions here w/ `_locked` 👍

Common/Data/SubscriptionManager.cs
```diff
@@ -39,7 +39,8 @@ public class SubscriptionManager
         /// <summary>
         ///     Returns an IEnumerable of Subscriptions
         /// </summary>
-        public IEnumerable<SubscriptionDataConfig> Subscriptions => _subscriptionManager.SubscriptionManagerSubscriptions;
+        /// <remarks>Will not return internal subscriptions</remarks>
+        public IEnumerable<SubscriptionDataConfig> Subscriptions => _subscriptionManager.SubscriptionManagerSubscriptions.Where(config => !config.IsInternalFeed);
```
mchandschuh: Any concerns here regarding internal currency conversion feeds?

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -336,9 +338,46 @@ public SecurityChanges ApplyUniverseSelection(Universe universe, DateTime dateTi
...
-        public bool AddPendingCurrencyDataFeeds(DateTime utcStart)
+        public bool AddPendingInternalDataFeeds(DateTime utcStart)
         {
             var added = false;
+            if (!_initializedSecurityBenchmark)
+            {
+                _initializedSecurityBenchmark = true;
+
+                var securityBenchmark = _algorithm.Benchmark as SecurityBenchmark;
+                if (securityBenchmark != null)
```
mchandschuh: I wonder if there's a better, more encapsulated way to perform this. This, upon first glance, appears to be leaking details  of the security benchmark into the universe selection. Let's chat about the reasoning behind placing this logic here and perhaps some other potential options for performing this initializing behavior.

Algorithm/QCAlgorithm.cs
Martin-Molinero: My intention was to avoid using the same security instance used by the algorithm, this allows us to avoid multiple issues e.g. `Security.Leverage` being wrong, price updates https://github.com/QuantConnect/Lean/blob/master/Engine/AlgorithmManager.cs#L338

Algorithm/QCAlgorithm.cs
Martin-Molinero: Thanks :)

Common/Data/SubscriptionManager.cs
Martin-Molinero: Not really, the idea is that the algorithm does not need to know or have access directly to any of these subscriptions - note that the backend/data feed stack do has direct access to the entire subscription collection.
And it's not used when we `EnsureCurrencyDataFeed()` https://github.com/QuantConnect/Lean/blob/master/Common/Securities/Cash.cs#L168

Engine/DataFeeds/UniverseSelection.cs
Martin-Molinero: Agree, this is the sticky point of this PR 😅 . Part of the reason I place it here is that this is where we are detecting/adding the internal currency subscriptions - But agree, I believe both additions could be handled differently - I guess what we are missing is an owner/responsible for the internal subscriptions altogether

Algorithm/QCAlgorithm.cs
mchandschuh: OK, I see the motive, I guess I'm just surprised that this works. The subscription in the data feed would need to be linked to _this_ security object, right? In other words, how does this `security` instance receive pricing updates?

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -336,9 +338,46 @@ public SecurityChanges ApplyUniverseSelection(Universe universe, DateTime dateTi
...
+
+                    var dataConfig = _algorithm.SubscriptionManager.SubscriptionDataConfigService.Add(
+                        securityBenchmark.Security.Symbol,
+                        _algorithm.LiveMode ? Resolution.Minute : Resolution.Daily,
+                        isInternalFeed: true,
+                        fillForward: false).First();
+
+                    if (dataConfig != null)
+                    {
+                        added |= _dataManager.AddSubscription(new SubscriptionRequest(
```
mchandschuh: I'm feeling better about this after additional thinking. The 'benchmark' is certainly an algorithm focused concern, so handling it explicitly in the algorithm layer is actually kind of nice 👍

PR #3433, opened 2019-07-25
Fixes Volatility Discontinuities in Security.VolatilityModel

### LINKED ISSUE #2127: Volatility Models Incorrect on Price Discontinuities
#### Expected Behavior
When using raw pricing, volatility models should still be accurate after a split

#### Actual Behavior
Volatility models inaccurate after a split as they read the price jump as a discontinuity. 

#### Potential Solution
 - Bind volatility models to split events to be re-calculated on splits; 
 - Or force vol models to always use adjusted pricing. 

#### Reproducing the Problem
https://www.quantconnect.com/terminal/processCache/?request=embedded_backtest_b248a7d30a544c3527008eac184c9bd0.html
```csharp
{
    {
    	Equity aapl;
            SetStartDate(2014, 03, 01);
            SetEndDate(2014, 08, 30);
            SetCash(100000);
            aapl = AddEquity("AAPL", Resolution.Daily);
            aapl.SetDataNormalizationMode(DataNormalizationMode.Raw);
            aapl.SetVolatilityModel( new StandardDeviationOfReturnsVolatilityModel( 7 ) );
        }

        public override void OnData(Slice data) {
            if (!Portfolio.Invested) {
                SetHoldings("AAPL", 1);
                Debug("Purchased AAPL");
            }
            
            if (data.Splits.Keys.Count > 0) {
            	Debug("SPLIT Slice >> " + data.Splits["AAPL"].ToString());
            }
        }
        
        public override void OnEndOfDay() {
            Plot("Vol", "AAPL", aapl.VolatilityModel.Volatility);
        }
    }
}
```

#### System Information
All

#### Checklist

AlexCatarino:
#### Description
Uses `Split` and `Dividend` information to update past input to the `IVolatilityModel` objects.

When there is a split and/or dividend event, the last price factor is saved in the volatility model and used to update the past price(s) to put them in line with post-event prices.

#### Related Issue
Closes #2127

#### Motivation and Context
Security volatility is an important value for trading and option pricing. Discontinuities in their values due to split and dividend events signs an artificial change in volatility.

Unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/Volatility/BaseVolatilityModel.cs
+++ b/Common/Securities/Volatility/BaseVolatilityModel.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Data;
+using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
 
 namespace QuantConnect.Securities.Volatility
@@ -34,7 +35,12 @@ namespace QuantConnect.Securities.Volatility
         /// <summary>
         /// Gets the volatility of the security as a percentage
         /// </summary>
-        public virtual decimal Volatility { get; }
+        public virtual decimal Volatility { get; } = 0m;
+
+        /// <summary>
+        /// Latest price factor to be applied
+        /// </summary>
+        public decimal? LastFactor { get; private set; }
 
         /// <summary>
         /// Sets the <see cref="ISubscriptionDataConfigProvider"/> instance to use.
@@ -54,6 +60,56 @@ namespace QuantConnect.Securities.Volatility
         /// <param name="data">The new data used to update the model</param>
         public virtual void Update(Security security, BaseData data)
         {
+        }
+
+        /// <summary>
+        /// Applies a dividend to the portfolio
+        /// </summary>
+        /// <param name="dividend">The dividend to be applied</param>
+        /// <param name="liveMode">True if live mode, false for backtest</param>
+        /// <param name="mode">The <see cref="DataNormalizationMode"/> for this security</param>
+        public virtual void ApplyDividend(Dividend dividend, bool liveMode, DataNormalizationMode mode)
+        {
+            // only apply splits in live or raw data mode
+            if (!liveMode && !(mode == DataNormalizationMode.Raw || mode == DataNormalizationMode.SplitAdjusted))
+            {
+                return;
+            }
+
+            var factor = 1 - dividend.Distribution / dividend.ReferencePrice;
+            SetLastFactor(factor);
+        }
+
+        /// <summary>
+        /// Applies a split to the model
+        /// </summary>
+        /// <param name="split">The split to be applied</param>
+        /// <param name="liveMode">True if live mode, false for backtest</param>
+        /// <param name="mode">The <see cref="DataNormalizationMode"/> for this security</param>
+        public virtual void ApplySplit(Split split, bool liveMode, DataNormalizationMode mode)
+        {
+            // only apply splits in live or raw data mode
+            if (!liveMode && mode != DataNormalizationMode.Raw)
+            {
+                return;
+            }
+
+            SetLastFactor(split.SplitFactor);
+        }
+
+        /// <summary>
+        /// Sets the value of the latest price factor
+        /// </summary>
+        /// <param name="factor">Latest price factor to be applied</param>
+        public void SetLastFactor(decimal? factor)
+        {
+            if (!LastFactor.HasValue || !factor.HasValue)
+            {
+                LastFactor = factor;
+                return;
+            }
+
+            LastFactor *= factor;
         }
 
         /// <summary>
--- a/Common/Securities/Volatility/BaseVolatilityModel.cs
+++ b/Common/Securities/Volatility/BaseVolatilityModel.cs
@@ -27,6 +27,8 @@ namespace QuantConnect.Securities.Volatility
     /// </summary>
     public class BaseVolatilityModel : IVolatilityModel
     {
+        private decimal? _lastFactor;
+
         /// <summary>
         /// Provides access to registered <see cref="SubscriptionDataConfig"/>
         /// </summary>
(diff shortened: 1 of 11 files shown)


Common/Securities/Volatility/BaseVolatilityModel.cs
```diff
@@ -56,6 +62,56 @@ public virtual void Update(Security security, BaseData data)
...
+            }
+
+            SetLastFactor(split.SplitFactor);
+        }
+
+        /// <summary>
+        /// Sets the value of the latest price factor
+        /// </summary>
+        /// <param name="factor">Latest price factor to be applied</param>
+        public void SetLastFactor(decimal? factor)
```
mchandschuh: This should likely be placed inside the property's setter to enforce encapsulation of the property's data.

Tests/Common/Securities/OptionPriceModelTests.cs
```diff
@@ -332,6 +332,14 @@ public IEnumerable<HistoryRequest> GetHistoryRequirements(Security security, Dat
             public void Update(Security security, BaseData data)
             {
             }
+
+            public void ApplyDividend(Dividend dividend, bool liveMode, DataNormalizationMode mode)
```
mchandschuh: If these were required for compilation, that implies this change will also break existing user algorithms IF they define their own `IVolatilityModel` -- perhaps we should also have a base class `VolatilityModel` if we don't already.

Common/Securities/Volatility/BaseVolatilityModel.cs
mchandschuh: Also, consider making `LastFactor` private/protected -- I don't think consumers should require/be coupled to this data

Tests/Common/Securities/OptionPriceModelTests.cs
AlexCatarino: We do have `BaseVolatilityModel`. I will simplify that class by inheriting from it.

Engine/AlgorithmManager.cs
```diff
@@ -478,7 +478,7 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                     Log.Debug($"AlgorithmManager.Run(): {algorithm.Time}: Applying Dividend: {dividend}");
 
                     Security security = null;
-                    if (_liveMode && algorithm.Securities.TryGetValue(dividend.Symbol, out security))
+                    if (_liveMode & algorithm.Securities.TryGetValue(dividend.Symbol, out security))
```
jaredbroad: I think you'll need this `&&`

Engine/AlgorithmManager.cs
```diff
@@ -514,7 +517,7 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                         Log.Debug($"AlgorithmManager.Run(): {algorithm.Time}: Applying Split for {split.Symbol}");
 
                         Security security = null;
-                        if (_liveMode && algorithm.Securities.TryGetValue(split.Symbol, out security))
+                        if (_liveMode & algorithm.Securities.TryGetValue(split.Symbol, out security))
```
jaredbroad: Same here. `&` is bitwise comparison which might work in this particular case but shouldn't be relied on.

Tests/Common/Securities/RelativeStandardDeviationVolatilityModelTests.cs
```diff
@@ -178,5 +168,99 @@ public void GetHistoryRequirementsWorksForTwoDifferentSubscriptions()
...
+            var expected = prices.StandardDeviation().SafeDecimalCast() / Math.Abs(prices.Mean().SafeDecimalCast());
+            Assert.AreEqual(expected, model.Volatility);
+        }
+
+        [Test]
+        public void HandlesPriceDiscontinuityDueToSplitAndDividend()
+        {
+            const int periods = 3;
+            var periodSpan = Time.OneMinute;
+            var model = new RelativeStandardDeviationVolatilityModel(periodSpan, periods);
```
jaredbroad: Please add unit tests for every type of Indicator Volatility Model as well

Common/Securities/Volatility/IndicatorVolatilityModel.cs
```diff
@@ -80,6 +78,26 @@ public void Update(Security security, BaseData data)
             }
         }
 
+        /// <summary>
+        /// Applies a dividend to the portfolio
+        /// </summary>
+        /// <param name="dividend">The dividend to be applied</param>
+        /// <param name="liveMode">True if live mode, false for backtest</param>
+        /// <param name="mode">The <see cref="DataNormalizationMode"/> for this security</param>
+        public void ApplyDividend(Dividend dividend, bool liveMode, DataNormalizationMode mode)
```
jaredbroad: This is relying on the owner of the indicator to handle the split-dividend events. If the stock splits in live trading it will still suffer a discontinuity (indicators don't currently automatically re-adjust post-split). 

Given we don't have indicators automatically re-initializing shouldn't we handle it for them here?

Tests/Common/Securities/RelativeStandardDeviationVolatilityModelTests.cs
AlexCatarino: Tests for `IndicatorVolatilityModel` added.

Common/Securities/Volatility/IndicatorVolatilityModel.cs
AlexCatarino: In order to handle it, we will need to implement a "refresh" logic to each indicator similar to what I have done to `RelativeStandardDeviationVolatilityModel` and `StandardDeviationOfReturnsVolatilityModel`. 
Alternatively, we could ensure that indicators always receive adjusted data (https://github.com/QuantConnect/Lean/issues/383) at least in this case.

Engine/AlgorithmManager.cs
mchandschuh: When using `&` with booleans, the net result will be the same, but the more important aspect of `&&` is it's short-circuiting functionality, so if `_liveMode` is `false`, then `algorithm.Securities.TryGetValue` is never invoked

AlexCatarino: In the last commit, I flipped the order:
```csharp
if (algorithm.Securities.TryGetValue(split.Symbol, out security && _liveMode)
```
The changed were made to precisely get `algorithm.Securities.TryGetValue` invoked.

jaredbroad:
Thank you @AlexCatarino. Closing as needs a rethink to ideally handle all systems which lean on adjusted data in a centralized way. Otherwise, we'll repeat this issue over and over with multiple different consumers.

Issue #3394, opened 2019-07-12
Enable live Kelly Criterion calculations akin to current Fitness Function


simonsonjack:
#### Expected Behavior

Implement a method for calculating [the Kelly Criterion outlined in our research notebook](https://www.quantconnect.com/forum/discussion/6194/insight-scoring-metric/p1) on a continuous basis in backtesting and for live trading/Alphas. This functionality would be the same as the current Fitness Function but instead, follow the following calculation method.

-Insights are scored based on their predicted direction and the actual change of the security at the end of the Insight duration. That is, the scoring function is 

`Direction x Security Price Change`, where Direction is -1 or 1, for down and up respectively

(NOTE: We will _ignore_ Flat insights. They will not be scored and need to be removed from the population).

-The Kelly Criterion estimate is taken using the population mean and dividing it by the population mean squared plus the population standard deviation squared. 

    KC Estimate = (mean) / ( mean^2 + stdDev^2)

-The Kelly criterion estimate standard deviation is calculated:
   
    KC std dev = square root( ( ( 1/stdDev^2) + (2*mean^2/ stdDev^4) ) / n - 1)

where n = population size.

-Finally, the p-value of the estimate is calculated

    p-value = 1 - Z(KC estimate/KC std dev)

where `Z()` is the cumulative distribution function of the Standard Normal Distribution (i.e., in python this translates to `1 - sps.norm.cdf( kellyCriterion / kellyStandardDeviation )`).

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3439, opened 2019-07-26
Add Kelly Criterion values

Martin-Molinero:
#### Description

- New `KellyCriterionManager` will be used by the
`StatisticsInsightManagerExtension` to calculate and update the
`AlphaRuntimeStatistics` with the new Kelly Criterion values, on a daily
basis, using the last year worth of samples
- Adding unit tests

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3394
#### Motivation and Context

- Add new insight scoring metric, see https://www.quantconnect.com/forum/discussion/6194/insight-scoring-metric/p1

N/A

Unit and regression tests, performance tests, cloud backtests
#### Types of changes

#### Checklist:


Engine/Alphas/StatisticsInsightManagerExtension.cs
```diff
@@ -99,16 +104,20 @@ public void OnInsightClosed(InsightAnalysisContext context)
...
             // value of the entering the trade in the account currency
-            var enterValue = volume * context.InitialValues.Price * context.InitialValues.QuoteCurrencyConversionRate;
+            var enterValue = context.InitialValues.Price * context.InitialValues.QuoteCurrencyConversionRate;
 
             // value of exiting the trade in the account currency
-            var exitValue = volume * context.CurrentValues.Price * context.CurrentValues.QuoteCurrencyConversionRate;
+            var exitValue = context.CurrentValues.Price * context.CurrentValues.QuoteCurrencyConversionRate;
 
             // total value delta between enter and exit values
             var insightValue = (int)context.Insight.Direction * (exitValue - enterValue);
```
simonsonjack: we want to measure the magnitude change rather than total value delta for kelly criterion, so the value that needs to be added in `_kellyCriterionManager.AddNewValue(insightValue, context.Insight.GeneratedTimeUtc);` should be
```
(int)context.Insight.Direction * ((exitValue / enterValue) - 1);
```

simonsonjack: This is the calculation in python, where all three are columns from a pandas DF, Reference Value Final is the stock price at Insight close, and Reference Value is the stock price at Insight creation.
```scores = df['Direction'].values * ((df['Reference Value Final'].values / df['Reference Value'].values) - 1)```

Comment (simonsonjack) [inline Engine/Alphas/StatisticsInsightManagerExtension.cs]: We want to measure the magnitude change rather than total value delta for kelly criterion, so the value that needs to be added in `_kellyCriterionManager.AddNewValue(insightValue, context.Insight.GeneratedTimeUtc);` should be
```
(int)context.Insight.Direction * ((exitValue / enterValue) - 1);
```

This is the calculation in python, where all three are columns from a pandas DF, Reference Value Final is the stock price at Insight close, and Reference Value is the stock price at Insight creation.
```scores = df['Direction'].values * ((df['Reference Value Final'].values / df['Reference Value'].values) - 1)```

Comment (Martin-Molinero) [reply Engine/Alphas/StatisticsInsightManagerExtension.cs]: Good catch, PR updated 👍

File: Tests/Common/Statistics/KellyCriterionManagerTests.cs [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+                kellyCriterionManager.AddNewValue(newValue, DateTime.UtcNow);
+                kellyCriterionManager.UpdateScores();
+            }
+
+            var estimate = kellyCriterionManager.KellyCriterionEstimate;
+            var probabilityValue = kellyCriterionManager.KellyCriterionProbabilityValue;
+
+            Console.WriteLine($"Estimate {estimate} - ProbabilityValue {probabilityValue}");
+            Assert.AreEqual(extremeCase == 1
+                ? 0.315789473684211m : -0.315789473684211m, kellyCriterionManager.KellyCriterionEstimate);
```
simonsonjack: I'm running the same calculations in Python and getting different numbers. For extreme large numbers I'm seeing an estimate of 0.000333333333 and with mean 2000.0 and variance 2000000. Same for small numbers case but on a different scale

Engine/Alphas/StatisticsInsightManagerExtension.cs
```diff
@@ -99,16 +104,26 @@ public void OnInsightClosed(InsightAnalysisContext context)
...
 
-            context.Insight.EstimatedValue = insightValue;
-            Statistics.TotalAccumulatedEstimatedAlphaValue += insightValue;
+            var insightValueFactoredByTradableVolume = insightValue * volume;
+
+            context.Insight.EstimatedValue = insightValueFactoredByTradableVolume;
+            Statistics.TotalAccumulatedEstimatedAlphaValue += insightValueFactoredByTradableVolume;
+
+            // testing showed sometimes enter value is 0, example is ScheduledUniverseSelectionModelRegressionAlgorithm
```
mchandschuh: any investigation into why we sometimes get a zero here? what the underlying cause may be?

Martin-Molinero: The initial price is zero, I'll make a separate GH issue to investigate further 👍

Engine/Alphas/DefaultAlphaHandler.cs
```diff
@@ -40,17 +40,20 @@ public class DefaultAlphaHandler : IAlphaHandler
...
         private ISecurityValuesProvider _securityValuesProvider;
-        private CancellationTokenSource _cancellationTokenSource;
         private FitnessScoreManager _fitnessScore;
         private DateTime _lastFitnessScoreCalculation;
         private readonly object _lock = new object();
 
+        /// <summary>
+        /// The cancellation token that will be cancelled when requested to exit
+        /// </summary>
+        protected CancellationTokenSource CancellationTokenSource { get; set; }
```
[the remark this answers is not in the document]
Martin-Molinero: Good idea, I'll address 👍

Issue #3045, opened 2019-04-01
Internationalisation: Use culture invariant anywhere there is a conversion


erikbengtson:
#### Expected Behavior
Lean should be able to run on non English based computers

#### Actual Behavior
Parsing prices is failing because it takes machine culture
#### Potential Solution
Use CultureInfo.InvariantCulture everywhere there is a conversion such as Convert.ToDecimal
e.g. in class OandaRestApiV20.OnPricingDataReceived change code to:
Convert.ToDecimal(data.Bids.Last().Price, CultureInfo.InvariantCulture)
Convert.ToDecimal(data.Asks.Last().Price, CultureInfo.InvariantCulture)

#### Reproducing the Problem
Change machine to other language/location

#### System Information
macOS french

#### Checklist

- [X] I have completely filled out this template
- [X] I have confirmed that this issue exists on the current `master` branch
- [X] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [X] I have provided detailed steps to reproduce the issue

jaredbroad:
Merged in QuantConnect#3045

PR #3442, opened 2019-07-29
Enforce usage of IFormatProvider/CultureInfo.InvariantCulture

mchandschuh:
#### Summary

In order to support internationalization and to ensure LEAN's consistency despite local machine settings, FxCopAnalyzers were added to flag violations of [CA1304](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1304-specify-cultureinfo?view=vs-2017) and [CA1305](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1305-specify-iformatprovider?view=vs-2017) as build errors. This will ensure that we're always parsing and  to- string-ing values while using the invariant culture. A new class, `StringExtensions` was added to hold methods aimed at providing patterns to use to enforce usage of `CultureInfo.InvariantCulture`

##### [Microsoft.CodeAnalysis.FxCopAnalyzers](https://github.com/dotnet/roslyn-analyzers)
Roslyn analyzers were added to each project file and a new `QuantConnect.ruleset` file has also been added that includes a reference to _every_ known rule at time of writing. As we update the rules for the repository, we can go into the `QuantConnect.ruleset` file and switch the `Action="None"` to `Action="Error"` to strictly enforce the rule.

---

#  :warning:

**I've tagged many of you guys for review, _mostly_ as a point of knowledge dissemination to ensure we're all on the same page regarding these changes and the preferred patterns moving forward. Also, we'll likely want to make similar changes elsewhere.**

---

#### Description

CA1304 (Use `IFormatProvider`) and CA1305 (Use `CultureInfo`) were turned on and violations will cause a build failure. Existing violations were updated to use methods located in the `StringExtensions`, which provides extension methods to make using culture-aware methods easier.

#### Related Issue

See #3045

#### Motivation and Context

LEAN is used worldwide, and as such, it should be able to run with the same results despite the local machine's settings. Updating everything to use `CultureInfo.InvariantCulture` ensures consistent runtime behavior.

No, but contributors should be aware of these changes and aim to use the new extension methods where applicable instead of in-lining the `CultureInfo.InvariantCulture` usage.

Unit tests have been added for `StringExtensions` and existing test coverage was deemed adequate for covering the other changes.

#### Types of changes

#### Checklist:


Configuration/Config.cs
```diff
@@ -230,7 +231,13 @@ public static T GetValue<T>(string key, T defaultValue = default(T))
             var token = GetToken(Settings.Value, key);
             if (token == null)
             {
-                Log.Trace(string.Format("Config.GetValue(): {0} - Using default value: {1}", key, defaultValue));
+                var defaultValueString = defaultValue is IConvertible
+                    ? ((IConvertible) defaultValue).ToString(CultureInfo.InvariantCulture)
+                    : defaultValue is IFormattable
```
jaredbroad: `IFormattable` 🤓

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -969,7 +969,7 @@ private static void Truncate(LiveResult result, DateTime start, DateTime stop)
 
         private string CreateKey(string suffix, string dateFormat = "yyyy-MM-dd")
         {
-            return $"{_job.DeployId}-{DateTime.UtcNow.ToString(dateFormat)}_{suffix}.json";
+            return $"{_job.DeployId}-{DateTime.UtcNow.ToStringInvariant(dateFormat)}_{suffix}.json";
```
jaredbroad: Should sanity check this doesn't change

Compression/ZipStreamWriter.cs
```diff
@@ -151,7 +151,7 @@ private void CopyTempFile(int attempts, bool throwOnFailure)
 
             if (throwOnFailure)
             {
-                throw new Exception(string.Format("Unable to save file: {0} after {1} attempts.", _filename, attempts));
+                throw new Exception($"Unable to save file: {_filename} after {attempts} attempts.");
```
Martin-Molinero: Wondering whats the criteria to call `.ToStringInvariant()` on some `int` like https://github.com/QuantConnect/Lean/pull/3442/files#diff-7b732fdae7d43c1eccb84cabe6af4e21R143 `order.Id.ToStringInvariant();` but here on `attempts` not?

Queues/JobQueue.cs
```diff
@@ -120,7 +120,7 @@ public AlgorithmNodePacket NextJob(out string location)
                 }
                 catch (Exception err)
                 {
-                    Log.Error(err, string.Format("Error resolving BrokerageData for live job for brokerage {0}:", liveJob.Brokerage));
+                    Log.Error(err, $"Error resolving BrokerageData for live job for brokerage {liveJob.Brokerage}:");
```
Martin-Molinero: Extra `:` at the end

ToolBox/IQFeed/IQFeedDataQueueHandler.cs
```diff
@@ -283,7 +283,7 @@ private bool CanSubscribe(Symbol symbol)
             var market = symbol.ID.Market;
             var securityType = symbol.ID.SecurityType;
 
-            if (symbol.Value.ToLower().IndexOf("universe") != -1) return false;
+            if (symbol.Value.ToLowerInvariant().IndexOfInvariant("universe") != -1) return false;
```
Martin-Molinero: Maybe this could directly call `IndexOfInvariant(ignoreCase: true), skipping `ToLowerInvariant()`

ToolBox/IEX/IEXDataQueueHandler.cs
```diff
@@ -360,7 +360,9 @@ private IEnumerable<Slice> ProcessHistoryRequests(Data.HistoryRequest request)
                 yield break;
             }
 
-            Log.Trace(string.Format("IEXDataQueueHandler.ProcessHistoryRequests(): Submitting request: {0}-{1}: {2} {3}->{4}", request.Symbol.SecurityType, ticker, request.Resolution, start, end));
+            Log.Trace("IEXDataQueueHandler.ProcessHistoryRequests(): Submitting request: " +
+                $"{request.Symbol.SecurityType}-{ticker}: {request.Resolution} {start.ToStringInvariant()}->{end.ToStringInvariant()}"
```
Martin-Molinero: Resolution is missing the `.ToStringInvariant`?

ToolBox/QuantQuoteConverter/QuantQuoteConverterProgram.cs
```diff
@@ -71,9 +71,9 @@ public static void QuantQuoteConverter(string destinationDirectory, string sourc
                 {
                     var symbol = GetSymbol(file);
                     var fileContents = File.ReadAllText(file);
-                    var data = new Dictionary<string, string> { { string.Format("{0}_{1}_Trade_Second.csv", date.ToString("yyyyMMdd"), symbol), fileContents } };
+                    var data = new Dictionary<string, string> { { $"{date.ToStringInvariant("yyyyMMdd")}_{symbol}_Trade_Second.csv", fileContents } };
 
-                    var fileDestination = string.Format("{0}/equity/{1}/{2}/{3}_trade.zip",  destinationDirectory, resolution, symbol, date.ToString("yyyyMMdd"));
+                    var fileDestination = $"{destinationDirectory}/equity/{resolution}/{symbol}/{date.ToStringInvariant("yyyyMMdd")}_trade.zip";
```
Martin-Molinero: Resolution is missing the .ToStringInvariant?

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -2041,9 +2040,9 @@ public IDataConsolidator ResolveConsolidator(Symbol symbol, Resolution? resoluti
             // data we won't be able to do anything good, we'll call it second, but it would really just be minute!
             if (timeSpan < subscription.Resolution.ToTimeSpan())
             {
-                throw new ArgumentException(string.Format("Unable to create {0} {1} consolidator because {0} is registered for {2} data. " +
-                                                          "Consolidators require higher resolution data to produce lower resolution data.",
-                    symbol, resolution.Value, subscription.Resolution)
+                throw new ArgumentException($"Unable to create {symbol} {resolution.Value} consolidator because {symbol} " +
+                    $"is registered for {subscription.Resolution} data. Consolidators require higher resolution data to " +
```
Martin-Molinero: Resolution is missing the .ToStringInvariant?

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -2070,9 +2069,8 @@ public IDataConsolidator ResolveConsolidator(Symbol symbol, TimeSpan? timeSpan)
             // data we won't be able to do anything good, we'll call it second, but it would really just be minute!
             if (timeSpan.Value < subscription.Resolution.ToTimeSpan())
             {
-                throw new ArgumentException(string.Format("Unable to create {0} consolidator because {0} is registered for {1} data. " +
-                                                          "Consolidators require higher resolution data to produce lower resolution data.",
-                    symbol, subscription.Resolution)
+                throw new ArgumentException($"Unable to create {symbol} consolidator because {symbol} is registered for " +
+                    $"{subscription.Resolution} data. Consolidators require higher resolution data to produce lower resolution data."
```
Martin-Molinero: Resolution is missing the .ToStringInvariant?

Common/Util/LeanData.cs
```diff
@@ -437,75 +448,63 @@ public static string GenerateZipEntryName(Symbol symbol, DateTime date, Resoluti
                 case SecurityType.Crypto:
                     if (resolution == Resolution.Tick && symbol.SecurityType == SecurityType.Equity)
                     {
-                        return string.Format("{0}_{1}_{2}_{3}.csv",
-                            formattedDate,
-                            symbol.Value.ToLower(),
-                            tickType,
-                            resolution
-                        );
+                        return $"{formattedDate}_{symbol.Value.ToLowerInvariant()}_{tickType}_{resolution}.csv";
```
Martin-Molinero: Resolution is missing the `.ToStringInvariant` and `tickType`?

ToolBox/FxcmDownloader/FxcmDataDownloader.cs
```diff
@@ -140,7 +140,7 @@ public IEnumerable<BaseData> Get(Symbol symbol, Resolution resolution, DateTime
             // initialize session
             RequestTradingSessionStatus();
 
-            Console.WriteLine("Downloading {0} data from {1} to {2}...", resolution, startUtc.ToString("yyyyMMdd HH:mm:ss"), endUtc.ToString("yyyyMMdd HH:mm:ss"));
+            Console.WriteLine($"Downloading {resolution} data from {startUtc.ToStringInvariant("yyyyMMdd HH:mm:ss")} to {endUtc.ToStringInvariant("yyyyMMdd HH:mm:ss")}...");
```
Martin-Molinero: Resolution is missing the .ToStringInvariant?

ToolBox/IQFeed/IQFeedDataQueueHandler.cs
```diff
@@ -604,7 +602,9 @@ public IEnumerable<Slice> ProcessHistoryRequests(HistoryRequest request)
                 end = null;
             }
 
-            Log.Trace(string.Format("HistoryPort.ProcessHistoryJob(): Submitting request: {0}-{1}: {2} {3}->{4}", request.Symbol.SecurityType, ticker, request.Resolution, start, end ?? DateTime.UtcNow.AddMinutes(-1)));
+            Log.Trace($"HistoryPort.ProcessHistoryJob(): Submitting request: {request.Symbol.SecurityType}-{ticker}: " +
+                $"{request.Resolution} {start}->{(end ?? DateTime.UtcNow.AddMinutes(-1)).ToStringInvariant()}"
```
Martin-Molinero: Resolution is missing the .ToStringInvariant?

Tests/Common/Util/LeanDataPathComponentsTests.cs
```diff
@@ -48,7 +49,7 @@ public TestCaseData[] GetTestCases()
                 from resolution in resolutions
                 from tickType in tickTypes
                 let date = resolution == Resolution.Hour || resolution == Resolution.Daily ? DateTime.MinValue : referenceDate
-                let name = string.Format("{0}-{1}-{2}-{3}", securityType.ToLower(), market.ToLower(), resolution.ToLower(), tickType.ToLower())
+                let name = $"{securityType}-{market}-{resolution}-{tickType}".ToLowerInvariant()
```
Martin-Molinero: Missing the individual `.ToStringInvariant`?

Brokerages/Alpaca/AlpacaBrokerage.DataQueueHandler.cs
```diff
@@ -104,7 +104,7 @@ private static bool CanSubscribe(Symbol symbol)
             if (symbol.ID.SecurityType != SecurityType.Equity)
                 return false;
 
-            return symbol.Value.ToLower().IndexOf("universe", StringComparison.Ordinal) == -1;
+            return symbol.Value.ToLowerInvariant().IndexOf("universe", StringComparison.Ordinal) == -1;
```
Martin-Molinero: Same comment about removing the `ToLowerInvariant` and directly calling `IndexOf` with ignore-case set to `true`
- Same for InteractiveBrokersBrokerage.cs:2455 https://github.com/QuantConnect/Lean/pull/3442/files#diff-6384ef48924c357fa33193a7b54c1a40R2455

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2129,7 +2129,7 @@ private Holding CreateHolding(IB.UpdatePortfolioEventArgs e)
             var currencySymbol = Currencies.GetCurrencySymbol(e.Contract.Currency);
             var symbol = MapSymbol(e.Contract);
 
-            var multiplier = Convert.ToDecimal(e.Contract.Multiplier);
+            var multiplier = e.Contract.Multiplier.ConvertInvariant<decimal>();
```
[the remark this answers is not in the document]
mchandschuh: The invariant culture is a non-region specific english culture -- this includes using `.` for a decimal point and `,` for thousands separators, as well as `$` sign (IIRC) for currency formatting, so in basic cases where a culture is not specified, invariant is the most likely. We might want to test to see if IB uses different serialization based on the server's home region.

Engine/Results/LiveTradingResultHandler.cs
mchandschuh: Confirmed via unit test: `Formattable_ToStringInvariant_UsesProvided_FormatParameter`

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -501,7 +501,7 @@ public FilteredIdentity FilteredIdentity(Symbol symbol, Resolution resolution, P
         /// <returns>A new FilteredIdentity indicator for the specified symbol and selector</returns>
         public FilteredIdentity FilteredIdentity(Symbol symbol, TimeSpan resolution, PyObject selector = null, PyObject filter = null, string fieldName = null)
         {
-            var name = string.Format("{0}({1}_{2})", symbol, fieldName ?? "close", resolution);
+            var name = $"{symbol}({fieldName ?? "close"}_{resolution.ToStringInvariant(null)})";
```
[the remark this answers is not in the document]
mchandschuh: The `IFormattable.ToStringInvariant(string format)` extension method is what is being invoked here since `TimeSpan` does not implement `IConvertible` -- so we're essentially calling `IFormattable.ToString(format: null, formatProvider: CultureInfo.InvariantCulture)`. We can't have a default value for `format = null` in the extension method since it would cause an ambiguity, also, specifying a null `format` should be have the same as _not_ specifying a format, so this maintains expected invariants.

Algorithm/QCAlgorithm.cs
```diff
@@ -1546,7 +1546,7 @@ public Future AddFuture(string ticker, Resolution resolution = Resolution.Minute
             {
                 if (!BrokerageModel.DefaultMarkets.TryGetValue(SecurityType.Future, out market))
                 {
-                    throw new Exception("No default market set for security type: " + SecurityType.Future);
+                    throw new KeyNotFoundException($"No default market set for security type: {SecurityType.Future}");
```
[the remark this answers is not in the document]
mchandschuh: Yea, figured I was reviewing almost every file, might as well make some of these changes as I see them 👍

Brokerages/GDAX/GDAXBrokerage.Utility.cs
```diff
@@ -115,7 +115,7 @@ private static string ConvertOrderType(Orders.OrderType orderType)
         {
             if (orderType == Orders.OrderType.Limit || orderType == Orders.OrderType.Market)
             {
-                return orderType.ToString().ToLower();
+                return orderType.ToStringInvariant();
```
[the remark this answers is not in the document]
mchandschuh: Amazing catch.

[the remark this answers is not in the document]
mchandschuh: Updated to use an extension method, `ToLower()` for enums that does `ToString().ToLowerInvariant()`

Common/Securities/MarketHoursSegment.cs
```diff
@@ -150,7 +150,7 @@ public bool Overlaps(TimeSpan start, TimeSpan end)
         /// </returns>
         public override string ToString()
         {
-            return string.Format("{0}: {1}-{2}", State, Start, End);
+            return $"{State}: {Start.ToStringInvariant(null)}-{End.ToStringInvariant(null)}";
```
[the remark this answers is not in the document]
mchandschuh: See same response for `TimeSpan` not implementing `IConvertible` and only implementing `IFormattable`

Common/StringExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// <summary>
+    /// Provides extension methods for properly parsing and serializing values while properly using
+    /// an IFormatProvider/CultureInfo when applicable
+    /// </summary>
+    public static class StringExtensions
+    {
+        /// <summary>
+        /// Parses the specified string as <see cref="long"/> using <see cref="CultureInfo.InvariantCulture"/>
+        /// </summary>
+        public static T ConvertInvariant<T>(this object convertible)
```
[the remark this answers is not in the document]
mchandschuh: Yes, but then we require consumers to have an `IConvertible` which isn't required by the `Convert` interface and would require updating types at the call sites (probably via static casting to `IConvertible`) -- this maintains existing behavior.

Brokerages/Alpaca/AlpacaBrokerage.DataQueueHandler.cs
mchandschuh: Using `StringExtensions.IndexOfInvariant(string substring, bool ignoreCase)`

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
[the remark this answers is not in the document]
mchandschuh: The implementation of `Convert.ToDecimal(string)`:
``` csharp
public static Decimal ToDecimal(string value)
{
  if (value == null)
    return Decimal.Zero;
  return Decimal.Parse(value, (IFormatProvider) CultureInfo.CurrentCulture);
}
```

We're essentially changing this to just not use the current culture of the local machine.

Common/StringExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+using System;
+using System.Globalization;
+
+namespace QuantConnect
+{
+    /// <summary>
+    /// Provides extension methods for properly parsing and serializing values while properly using
+    /// an IFormatProvider/CultureInfo when applicable
+    /// </summary>
+    public static class StringExtensions
```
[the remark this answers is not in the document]
mchandschuh: Moved `Parse*` functions to a new static class, `Parse` to enable syntax such as:
``` csharp
var myDecimal = Parse.Decimal(myString);
```
and it will do the required thing. After looking at the usage of extensions for parsing, I felt it unnecessarily polluted string's method-space (intellisense), and also from a style perspective, using extension methods for parsing just seemed incorrect.

Algorithm/QCAlgorithm.Python.cs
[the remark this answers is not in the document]
mchandschuh: So the reason this pattern was chosen was due to .net 4.5.2 not having `FormattableString`, which is what string interpolation (`$"my-{interpolated}-string"`) compiles into. Sadly, in .net 4.5.2's FxCop won't create warnings/errors for `IFormattable`/`CultureInfo` not being provided in interpolated strings, so we'll need to be vigilant in our code reviews. In an effort to combat this, I found a nuget package that adds `FormattableString` to enable the pattern preferred by the dotnet core/rosyln teams for handling invariant cultured strings:
``` csharp
using static System.FormattableString

// ... somewhere inside a function
var myCulturedString = Invariant($"my-{interpolated}-string")
```
and this will perform the formatting of each argument using `CultureInfo.InvariantCulture`.

In an effort to keep all of our formatting going through a common code path, I've added `StringExtensions.Invariant` which currently simply pipes directly through to `FormattableString.Invariant`, so it enables the exact same syntax, but felt it beneficial to have all of the cases go through a single LEAN method, so if we need to change/modify the behavior, it should be as easy as updating this single method definition.

Since changing the pattern to support the `using static` (can't be extension method), many of these cases with a `null` format string have been removed as those cases are no longer using the `StringExtensions.ToStringInvariant(string)` method, but instead using `FormattableString.Invariant(FormattableString)`

Common/Util/StringExtensions.cs
```diff
@@ -0,0 +1,213 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Globalization;
+
+namespace QuantConnect
```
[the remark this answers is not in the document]
mchandschuh: Moved to the root of  `Common`

Common/QuantConnect.csproj
```diff
@@ -717,6 +729,7 @@
     <Compile Include="Util\SecurityIdentifierJsonConverter.cs" />
     <Compile Include="Util\SeriesJsonConverter.cs" />
     <Compile Include="Util\StreamReaderEnumerable.cs" />
+    <Compile Include="Util\StringExtensions.cs" />
```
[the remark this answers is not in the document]
mchandschuh: yup, great spot -- originally placed under `Common\Util` and then copied to `Common` and looks like that got left behind. This has been fixed 👍

Compression/ZipStreamWriter.cs
mchandschuh: The initial criteria I used was anything that raised the CA1304/CA1305 errors. _Sadly_, .net 4.5.2 doesn't raise these warnings when inside an interpolated string, which is what led to so many of these being missed. My understanding is that this is because `FormattableString` isn't natively available in .net 4.5.2, but is available in 4.6+ and dotnet core, so whenever we upgrade we should get warnings for them at that time.

_In general_, any type that implements `IConvertible` or `IFormattable` should be using `CultureInfo.InvariantCulture`, so this includes all `int`s. In order to get _all_ cases, I've begun the tedious process of going through each file and inspecting each string interpolation. If we're interpolating non-string values into the string, I'm adding `using static QuantConnect.StringExtensions` and using the `Invariant` method which is implemented via `FormattableString.Invariant` which applies the invariant culture to each argument in the interpolated string. This process has been completed for `Common`.

Algorithm/QCAlgorithm.Indicators.cs
mchandschuh: I've reviewed this entire file and updated all relevant cases to use the `using static` pattern. A nice approach (methinks) was creating an overload for `CreateIndicatorName` that accepts a `FormattableString` (the interpolated string) and then performs the `Invariant(fs)` to minimize required edits and the noise that would have been added if we did `Invariant(fs)` in each invocation to `CreateIndicatorName`

Tests/Common/Util/LeanDataPathComponentsTests.cs
mchandschuh: Updated to use `using static` -- `Invariant(fs)` pattern

mchandschuh:
Rebased to account for recent merges to master.

Algorithm.CSharp/PortfolioOptimizationNumericsAlgorithm.cs
```diff
@@ -180,7 +180,7 @@ public void SetWeight(double value)
 
             public override string ToString()
             {
-                return string.Format("{0}: {1,10:P2}\t{2,10:P2}\t{3,10:P2}", Symbol.Value, Weight, Return, Risk);
+                return $"{Symbol.Value}: {Weight,10:P2}\t{Return,10:P2}\t{Risk,10:P2}";
```
StefanoRaggi: Should use `ToStringInvariant` here?

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -570,9 +570,9 @@ private IEnumerable<OrderTicket> GenerateOrders(OptionStrategy strategy, int str
 
                     if (optionSeq.Count() != 1)
                     {
-                        var error = string.Format("Couldn't find the option contract in algorithm securities list. Underlying: {0}, option {1}, strike {2}, expiration: {3}",
-                                strategy.Underlying.ToString(), optionLeg.Right.ToString(), optionLeg.Strike.ToString(), optionLeg.Expiration.ToString());
-                        throw new InvalidOperationException(error);
+                        throw new InvalidOperationException("Couldn't find the option contract in algorithm securities list. " +
+                            $"Underlying: {strategy.Underlying}, option {optionLeg.Right.ToString()}, strike {optionLeg.Strike.ToStringInvariant()}, " +
```
StefanoRaggi: `optionLeg.Right.ToString()` -- is this intentional or should be `ToStringInvariant()`?

Related question: 
I see `ToStringInvariant()` is used everywhere also for `enum` values, is this really required?
Wouldn't these be all the same?
1. `{optionLeg.Right.ToString()}`
2. `{optionLeg.Right.ToStringInvariant()}`
3. `{optionLeg.Right}`

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -667,7 +667,11 @@ private OrderResponse PreOrderChecksImpl(SubmitOrderRequest request)
 
             if (Math.Abs(request.Quantity) < security.SymbolProperties.LotSize)
             {
-                return OrderResponse.Error(request, OrderResponseErrorCode.OrderQuantityLessThanLoteSize, $"Unable to {request.OrderRequestType.ToString().ToLower()} order with id {request.OrderId} which quantity ({Math.Abs(request.Quantity)}) is less than lot size ({security.SymbolProperties.LotSize}).");
+                return OrderResponse.Error(request, OrderResponseErrorCode.OrderQuantityLessThanLoteSize,
+                    $"Unable to {request.OrderRequestType.ToLower()} order with id {request.OrderId} which " +
```
StefanoRaggi: `ToLower()` -> `ToLowerInvariant()` ?

Common/Data/Auxiliary/LocalDiskMapFileProvider.cs
```diff
@@ -39,19 +39,21 @@ public class LocalDiskMapFileProvider : IMapFileProvider
...
+            var mapFileDirectory = Path.Combine(Globals.CacheDataFolder, "equity", market.ToLowerInvariant(), "map_files");
             if (!Directory.Exists(mapFileDirectory))
             {
                 // only write this message once per application instance
                 if (Interlocked.CompareExchange(ref _wroteTraceStatement, 1, 0) == 0)
                 {
-                    Log.Error("LocalDiskMapFileProvider.GetMapFileResolver({0}): The specified directory does not exist: {1}", market, mapFileDirectory);
+                    Log.Error($"LocalDiskMapFileProvider.GetMapFileResolver({market}): " +
+                        $"The specified directory does not exist: {mapFileDirectory}"
+                    );
```
StefanoRaggi: Love to see these `string.Format` placeholder numbers disappear! 😃

Common/Packets/SecurityTypesPacket .cs
```diff
@@ -1,11 +1,11 @@
 ﻿/*
```
StefanoRaggi: Just noticed an extra space in the file name -- might want to fix, since it's such a small PR 😄

Common/Securities/AccountEvent.cs
```diff
@@ -50,7 +52,7 @@ public AccountEvent(string currencySymbol, decimal cashBalance)
         /// <filterpriority>2</filterpriority>
         public override string ToString()
         {
-            return string.Format("Account {0} Balance: {1}", CurrencySymbol, CashBalance.ToString("0.00"));
+            return Invariant($"Account {CurrencySymbol} Balance: {CashBalance:0.00}");
```
StefanoRaggi: Love the `Invariant()` extension method usage with interpolated strings ❤️

Configuration/Config.cs
```diff
@@ -230,7 +231,13 @@ public static T GetValue<T>(string key, T defaultValue = default(T))
...
+                        ? ((IFormattable) defaultValue).ToString(null, CultureInfo.InvariantCulture)
+                        : defaultValue.ToString();
```
StefanoRaggi: Do we need `defaultValue.ToString(CultureInfo.InvariantCulture)` here?

ToolBox/IQFeed/IQ/IQLookupHistorySymbolClient.cs
```diff
@@ -284,13 +282,16 @@ public int RequestTickData(string symbol, int days, bool oldToNew, Time timeStar
...
 
-            var reqString = string.Format("HTT,{0},{1},{2},{3},{4},{5},{6},{7},{8}\r\n", symbol, start.ToString("yyyyMMdd HHmmss"),
-                end.HasValue ? end.Value.ToString("yyyyMMdd HHmmss") : "", _histMaxDataPoints.ToString("0000000"),
-                timeStartInDay == null ? "" : timeStartInDay.IQFeedFormat, timeEndInDay == null ? "" : timeEndInDay.IQFeedFormat, oldToNew ? "1" : "0", reqNo, _histDataPointsPerSend.ToString("0000000"));
+            var reqString = $"HTT,{symbol},{start.ToStringInvariant("yyyyMMdd HHmmss")}," +
+                $"{(end.HasValue ? end.Value.ToStringInvariant("yyyyMMdd HHmmss") : "")}," +
+                $"{_histMaxDataPoints.ToStringInvariant("0000000")}," +
+                $"{(timeStartInDay == null ? "" : timeStartInDay.IQFeedFormat)}," +
+                $"{(timeEndInDay == null ? "" : timeEndInDay.IQFeedFormat)},{(oldToNew ? "1" : "0")}," +
+                $"{reqNo},{_histDataPointsPerSend.ToStringInvariant("0000000")}\r\n";
```
StefanoRaggi: Not a required change for now, but I think using the `Invariant()` extension method instead of multiple `ToStringInvariant()` calls seems to make these long interpolated strings shorter and more readable. 
Thoughts?

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
```diff
@@ -136,8 +135,8 @@ public override bool Run()
 
                     foreach (var calendarDataByDate in kvp.GroupBy(x => x.LastUpdate.Date))
                     {
-                        var date = calendarDataByDate.Key.ToString("yyyyMMdd");
-                        var tempPath = Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid()}.json");
+                        var date = calendarDataByDate.Key.ToStringInvariant("yyyyMMdd");
+                        var tempPath = Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid().ToStringInvariant(null)}.json");
```
StefanoRaggi: Any thoughts about making the `format` argument of `ToStringInvariant()` optional, defaulting to `null`?

mchandschuh:
I think this is good to go. Is there anything else required here?

Martin-Molinero:
Nice work! 👍

Issue #3417, opened 2019-07-23
CreatedEquityIsNotAddedToSymbolCache test randomly failing in Travis


StefanoRaggi:
#### Expected Behavior
Unit tests should always pass in Travis build.

#### Actual Behavior
The `CreatedEquityIsNotAddedToSymbolCache` unit test is randomly failing in Travis build:

```
Errors and Failures:
1) Test Failure : QuantConnect.Tests.Engine.DataFeeds.UniverseSelectionTests.CreatedEquityIsNotAddedToSymbolCache
     Expected: <SPY R735QTJ8XC9X>
  But was:  <GOOCV VP83T1ZUHROL>

at QuantConnect.Tests.Engine.DataFeeds.UniverseSelectionTests.CreatedEquityIsNotAddedToSymbolCache () [0x000f9] in <c35d11bd8aed4b4485339ed9e3c1e007>:0
```

#### Potential Solution
n/a

#### Reproducing the Problem
n/a

#### System Information
n/a

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

Issue #3410, opened 2019-07-20
Lagging History in Universe Selection function


AlexCatarino:
#### Expected Behavior
The last data from a historical data request matches the price from coarse fundamental

#### Actual Behavior
One data point lag in the historical request:
```
2019-07-02 00:00:00 
            A RPTMYV3VC57P  AA WF7IHVI76I5H  YHOO R735QTJ8XC9X  AAL VM9RIYHM8ACL  RNTA SB2PBIC0SCX1
2019-07-02           75.81            22.92              70.55             32.88              62.88
2019-07-02 00:00:00 
symbol      A RPTMYV3VC57P  AA WF7IHVI76I5H  AAL VM9RIYHM8ACL  RNTA SB2PBIC0SCX1  YHOO R735QTJ8XC9X
time                                                                                               
2019-06-29       74.476001            23.39              32.6              61.38              69.41
2019-07-03 00:00:00 
            A RPTMYV3VC57P  AA WF7IHVI76I5H  YHOO R735QTJ8XC9X  AAL VM9RIYHM8ACL  RNTA SB2PBIC0SCX1
2019-07-03           75.61             23.3               70.7             32.18              63.34
2019-07-03 00:00:00 
symbol      A RPTMYV3VC57P  AA WF7IHVI76I5H  AAL VM9RIYHM8ACL  RNTA SB2PBIC0SCX1  YHOO R735QTJ8XC9X
time                                                                                               
2019-07-02           75.81            22.92             32.88              62.88              70.55
```
#### Reproducing the Problem
```python
import pandas as pd

class PriceVolumeBreakoutFrameworkAlgorithm(QCAlgorithm):

    def Initialize(self):

        self.UniverseSettings.Resolution = Resolution.Minute

        self.SetStartDate(2019, 7, 1)
        self.SetCash(1000000)
        
        self.AddUniverse(self.SelectCoarse)


    def SelectCoarse(self, coarse):

        filtered = [x for x in coarse if x.HasFundamentalData and x.Price > 20][:5]
        symbols = [x.Symbol for x in filtered]

        fDict = dict()
        for x in filtered:
            fDict[x.Symbol] = x.AdjustedPrice

        self.Log(pd.DataFrame(data = fDict, index = [filtered[0].EndTime]))
        self.Log(self.History(symbols,1, Resolution.Daily).close.unstack(level=0))

        return symbols
```
#### Checklist
- [x] I have completely filled out this template
[...]

Issue #3409, opened 2019-07-19
Adjust start time for Universe selection


simonsonjack:
#### Expected Behavior

Universe selection should happen at the beginning of a backtest period, prior to the first data for the Start Date, even when used in conjunction when manually adding a security, i.e. `AddEquity()`.

#### Actual Behavior

Universe selection happens the day _after_ the start date and the data for the securities selected by this method doesn't get fed into the algorithm until after universe selection. However, using AddEquity or a similar manual data subscription the relevant data through OnData() _on_ the algorithm start date, so a day prior to universe selected data.

#### Reproducing the Problem

Simple test algorithm
`class MultidimensionalTransdimensionalContainmentField(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2019, 1, 18)  # Set Start Date
        self.SetEndDate(2019, 1, 23)
        self.SetCash(100000)  # Set Strategy Cash
        self.AddEquity("SPY", Resolution.Minute)
        self.UniverseSettings.Resolution = Resolution.Minute
        self.AddUniverse(self.CoarseSelectionFunction, self.FineSelectionFunction)

    # sort the data by daily dollar volume and take the top 'NumberOfSymbols'
    def CoarseSelectionFunction(self, coarse):
        self.Log(f'Coarse Universe Selection')
        # sort descending by daily dollar volume
        sortedByDollarVolume = sorted(coarse, key=lambda x: x.DollarVolume, reverse=True)

        # return the symbol objects of the top entries from our sorted collection
        return [ x.Symbol for x in sortedByDollarVolume[:5] ]

    # sort the data by P/E ratio and take the top 'NumberOfSymbolsFine'
    def FineSelectionFunction(self, fine):
        self.Log(f'Fine Universe Selection')
        # sort descending by P/E ratio
        sortedByPeRatio = sorted(fine, key=lambda x: x.ValuationRatios.PERatio, reverse=True)

        # take the top entries from our sorted collection
        return [ x.Symbol for x in sortedByPeRatio[:2] ]

    def OnData(self, data):
        self.Log(f'OnData, Active Securities: {[key.Value for key in self.ActiveSecurities.Keys]}')`

Backtest logs are attached.
[731d725da82ac4548c3c35d491a2cc4e-log (1).txt](https://github.com/QuantConnect/Lean/files/3412550/731d725da82ac4548c3c35d491a2cc4e-log.1.txt)

#### Checklist

- [x] I have completely filled out this template
[...]

PR #3418, opened 2019-07-23
Universe selection time fixes

Martin-Molinero:
#### Description

- `Algorithm.Time` will be advanced each time the Frontier time provider
is updated. This will allow universe selection data time and
`algorithm.Time` to be aligned.
- Updating Regression algorithms that were using `algorithm.Time` in the
selection method. Will also change statistics for algorithms making a history request 
during universe selection.
- Coarse selection will start from the algorithms start date (not in the
next day)
- Adding regression algorithms
- Remove unused test clases

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3410
Closes https://github.com/QuantConnect/Lean/issues/3409
Closes https://github.com/QuantConnect/Lean/issues/3417
#### Motivation and Context

- Coarse selection started a day after algorithm start date
- Coarse selection algorithm time was lagging a day behind
#### Requires Documentation Change

N/A
#### How Has This Been Tested?

Unit and regression tests
#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #3452, opened 2019-07-30
Universe selection time fixes - v2

Martin-Molinero:
#### Description

New version of previous PR https://github.com/QuantConnect/Lean/pull/3418

New changes:
- `SubscriptionSynchronizer` will emit a `TimeSlice.TimePulse` before
performing any universe selection on each time loop. This will advance
`Algorithm.Time` which will allow universe selection data time and
`Algorithm.Time` to be aligned [partial sync between datafeed and the algorithm time (previous PR was complete sync)]

Old changes - same as previous PR
- Updating Regression algorithms that were using `algorithm.Time` in the
selection method.
- Coarse selection will start from the algorithms start date (not in the
next day)
- Adding regression algorithm

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3410
Closes https://github.com/QuantConnect/Lean/issues/3409
Closes https://github.com/QuantConnect/Lean/issues/3417
#### Motivation and Context

- Coarse selection started a day after algorithm start date
- Coarse selection algorithm time was lagging a day behind

N/A

Unit and regression tests
#### Types of changes

#### Checklist:


Algorithm.CSharp/CoarseSelectionTimeRegressionAlgorithm.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Data;
+using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Test algorithm that reproduces GH issues 3410 and 3409.
```
mchandschuh: Love the GH issue references!

Algorithm.CSharp/CoarseSelectionTimeRegressionAlgorithm.cs
```diff
@@ -0,0 +1,124 @@
...
+            _historyCoarseSpyPrice = History(_spy, 1).First().Close;
+
+            return top.Select(x => x.Symbol);
+        }
+
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
```
mchandschuh: Do we want to add a check to confirm the date our universe selection was applied by checking for a non-SPY symbol and comparing w/ an expected date?

Algorithm.CSharp/CustomUniverseSelectionRegressionAlgorithm.cs
```diff
@@ -0,0 +1,107 @@
...
+                Market.USA,
+                UniverseSettings,
+                time => new[] { "SPY" });
+        }
+
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
```
mchandschuh: Similar comment here regarding checking the date.

Algorithm.CSharp/SectorExposureRiskFrameworkAlgorithm.cs
```diff
@@ -38,7 +38,7 @@ public override void Initialize()
             // Set requested data resolution
             UniverseSettings.Resolution = Resolution.Daily;
 
-            SetStartDate(2014, 03, 24);
+            SetStartDate(2014, 03, 25);
```
mchandschuh: Is this change due to only certain data being available locally?

Engine/AlgorithmManager.cs
```diff
@@ -286,6 +286,12 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                 //Set the algorithm and real time handler's time
                 algorithm.SetDateTime(time);
 
+                // the time pulse are just to advance algorithm time, lets shortcut the loop here
+                if (timeSlice.IsTimePulse)
```
mchandschuh: Cool concept!

Engine/DataFeeds/Enumerators/Factories/BaseDataCollectionSubscriptionEnumeratorFactory.cs
```diff
@@ -53,24 +42,33 @@ public IEnumerator<BaseData> CreateEnumerator(SubscriptionRequest request, IData
...
+                // we want the first selection to happen on the start time
+                // so we need the previous tradable day time, since coarse
+                // files are for each tradable date but emitted with next day time
+                var previousTradableDay = Time.GetStartTimeForTradeBars(
+                    request.Security.Exchange.Hours,
+                    request.StartTimeLocal,
+                    Time.OneDay,
+                    1,
+                    false);
+                var tradableDays = new[] { previousTradableDay }.Concat(request.TradableDays);
```
mchandschuh: Wondering if this previous day logic should be handled via the caller instead of inside here. What's the desired behavior when adding a universe _after_ the algorithm has initialized? (not sure if that's a supported case).

Algorithm.CSharp/CoarseSelectionTimeRegressionAlgorithm.cs
Martin-Molinero: This is handled by the data/active security count, without this fix it throws on the first call, so I think we are covered 👍

Algorithm.CSharp/CustomUniverseSelectionRegressionAlgorithm.cs
Martin-Molinero: Similar response - Actually this regression test was just added to verify/assert the behavior of a custom universe selection (not using coarse) was the same as the coarse selection - This regression test passes in master.

Algorithm.CSharp/SectorExposureRiskFrameworkAlgorithm.cs
Martin-Molinero: Exactly 👍

Engine/AlgorithmManager.cs
Martin-Molinero: @jaredbroad idea 😃

Engine/DataFeeds/Enumerators/Factories/BaseDataCollectionSubscriptionEnumeratorFactory.cs
Martin-Molinero: Thought about this one for a while too, I think since this is a special behavior just for Coarse, better to handle it internally, normalizing the caller code.
> What's the desired behavior when adding a universe after the algorithm has initialized? (not sure if that's a supported case).

I'd be inclined to say its not a behavior we support at the moment - the objective is for coarse selection to happen right away 👍

Engine/DataFeeds/Enumerators/Factories/BaseDataCollectionSubscriptionEnumeratorFactory.cs
```diff
@@ -53,24 +42,33 @@ public IEnumerator<BaseData> CreateEnumerator(SubscriptionRequest request, IData
...
+
+                // Behaves in the same way as in live trading
                 // (i.e. only emit coarse data on dates following a trading day)
                 // The shifting of dates is needed to ensure we never emit coarse data on the same date,
                 // because it would enable look-ahead bias.
 
                 // shift all tradeable dates forward one day
-                foreach (var date in tradableDays.Select(x => x.AddDays(1)))
+                foreach (var date in tradableDays)
```
jaredbroad: Just leaving note for posterity; not sure changes are needed other than the comment:
> shift all tradeable dates forward one day
>to:  shift all date of emitting the file forward one day to model emitting coarse midnight the next day.

Also just want to sanity-check need to confirm this won't change the dates the live enumerator stack processes.

-------
The original code, although messy was on purpose to model _time_ as in live mode. In live _time_ is always end time. Universe data from Monday is deployed on Tuesday at midnight, and as the BaseDataCollection has no period it deploys instantly. This for loop was intended to model this live behavior:

1) `tradeableDays.AddDays(1)` = D+1 at midnight end time set time, emitting on a Saturday for Friday's data; Tuesday for Monday. (like the real daily clock)
2) `GetSource()` for `date.AddDays(-1)` = Specifically fetch the file named Monday at Tuesday midnight,
3) `ForSource(Days-1)` = Set the time of the datapoint to Monday in the coarseFundamental.Reader( date )
4) yield return the datapoint on Tuesday midnight, the real end time of the bar.

Perhaps could think of this (and update comment) to be:
1) Enumerate time start of Monday midnight (including since before start of backtest).
2) Get Source()/factor file for Monday before backtest starts.
3) Emit for backtesting 00:00 Tuesday morning?

The new loop _time_ does not model live trading IMHO; but it's fine if it addresses this bug.

Martin-Molinero: > Also just want to sanity-check need to confirm this won't change the dates the live enumerator stack processes.

Correct 👍  , he is only used in backtesting https://github.com/QuantConnect/Lean/pull/3452/files#diff-312a6d73563b5bf4e87823d1637c8425R31 
Good catch, I'll update the comment.

potifreire: EPIC!!

PR #3454, opened 2019-07-31
Fix order price rounding in BrokerageTransactionHandler and EquityPriceVariationModel

### LINKED ISSUE #3453: Incorrect limit/stop price rounding in live mode when crossing the $1 equity price
#### Expected Behavior
Order prices should be rounded properly taking into account the order price instead of the current market price.

#### Actual Behavior
The `EquityPriceVariationModel` is using `security.Price` to determine the minimum price variation, causing an incorrect price rounding when the current price and the order price are on opposite sides of the $1 price mark.

#### Potential Solution
Add a price argument to `IPriceVariationModel.GetMinimumPriceVariation`

#### Reproducing the Problem
- For a symbol currently trading below $1, submit a limit sell order at a price above $1
- IB will reject the order with this message:
```
110 - The price does not conform to the minimum price variation for this contract
```

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- Added `referencePrice` argument to `IPriceVariationModel`
- Updated `EquityPriceVariationModel` and `BrokerageTransactionHandler` to use limit/stop order prices instead of current market prices

#### Related Issue
Closes #3453 

#### Motivation and Context
IB error when submitting limit orders around $1 prices:
```
110 - The price does not conform to the minimum price variation for this contract
```

No.

New unit tests included.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
+++ b/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
@@ -1116,7 +1116,7 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
         /// This procedure is needed to meet brokerage precision requirements.
         /// </remarks>
         /// </summary>
-        private void RoundOrderPrices(Order order, Security security)
+        public void RoundOrderPrices(Order order, Security security)
         {
             // Do not need to round market orders
             if (order.Type == OrderType.Market ||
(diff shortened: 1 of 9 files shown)


Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -1116,7 +1116,7 @@ public decimal RoundOffOrder(Order order, Security security)
         /// This procedure is needed to meet brokerage precision requirements.
         /// </remarks>
         /// </summary>
-        private void RoundOrderPrices(Order order, Security security)
+        public void RoundOrderPrices(Order order, Security security)
```
Martin-Molinero: If setting to `public` just for testing, maybe its worth setting it as protected, inheriting `BrokerageTransactionHandler` and accessing the method through a proxy

StefanoRaggi:
> Wondering if we should/can take care of backwards compatibility for the interface changes?

Not sure if/how many users have implemented a custom `PriceVariationModel`: 
the existing interface method does not allow price rounding to be performed properly anyway because of the missing order information.

Martin-Molinero:
> Not sure if/how many users have implemented a custom `PriceVariationModel`:
> the existing interface method does not allow price rounding to be performed properly anyway because of the missing order information.

Right 👍 . I'm thinking since we are already chaning the interface, maybe for future proof we could add an `GetMinimumPriceVariationParameters` or alike? Like we did in the fill/fee/buyingPower models. And I think it could be a good opportunity to limit the parameters to just what is required, not the entire security instace, the new price reference, the symbol properties, etc?

StefanoRaggi:
> `GetMinimumPriceVariationParameters` 
Good idea! 👍 Will update the method to take a parameters class

PR #3484, opened 2019-08-06
Fix custom sid collisions (BUG #3332)

### LINKED ISSUE #3332: Adding Multiple Custom Data Sources Results in Symbol Collision
#### Expected Behavior

Custom data sources should have unique symbols separated by type of custom data
#### Actual Behavior

The two symbols generated would collide with each other
#### Potential Solution

Add an identifier to the `Symbol.ID.Symbol` value containing the name of the custom data type
#### Reproducing the Problem

```C#
AddData<T>("TWX", Resolution.Daily).Symbol == AddData<K>("TWX", Resolution.Daily).Symbol

// returns true
```

#### System Information

N/A
#### Checklist

mchandschuh:
#### Description
Aims to make custom data symbols unique by appending the custom data type's name. Consider the following:
``` csharp
AddData<Bitcoin>("BTC");
```
This will create a `SecurityIdentifier` with a `Symbol` property equal to `"BTC.Bitcoin"` which permits:
``` csharp
// the 'alias' follows the pattern <ticker>-<type-name> where ticker is capitalized
var symbol = SecurityCache.GetSymbol("BTC.Bitcoin");

// as a backwards compatibility shim, the following will work
// provided there's no other BTC custom symbols
symbol = SecurityCache.GetSymbol("BTC");
```
Custom (Base) data universe selectors were updated to pass the custom universe data type to maintain uniqueness via `Symbol.Create` now accepting a `Type` parameter that is **only** used for `SecurityType.Base` symbols and is ignored for all other security types.

#### Commit Descriptions

_Append type to custom data symbol_

This is being done in an effort to prevent symbol collisions within the custom data (SecurityType.Base) namespace. The custom data type's name, is used for disambiguation. As written, this change will break several user algorithms that still rely on using the implicit string -> Symbol lift. Providing this type information is optional an currently only being used by AddData<T> methods. Other consumers of SecurityType.Base symbols arn't at risk for collision, such as the UserDefinedUniverse, ScheduledUniverse and others that are LEAN controlled. In order to maintain backwards compatibility, the SymbolCache was updated to do a hard search when the requested ticker was not found, looking for the prefix ('ticker.').

_Enforce consistency in AddData<T> C#/Python_

Updated the `AddData<T>` implementations to use the same code path.

---

### :warning: Backwards Compatibility Issue :warning:

Notice `HistoryAlgorithm.py` was updated. This was done to permit the regression tests to pass. I did a little debugging and it seems that the implicit string conversion to `Symbol` is not being invoked when indexing into a dictionary, such as `Securities`. What _appears_ to be happening, is the `Symbol` object's `ToString()` method is invoked, which yields `SymbolCache.GetTicker(this)` and that is used as the string indexer; when accessed, it's a simple string lookup. This appears to unearth an old standing bug where the implicit string conversion is not being used at all, but the _appearance_ of it being used given the indexing via `Symbol.ToString() => SecurityCache.GetTicker(this)`. In fact, it should cover all existing cases completely, but won't cover the new case where we perform a secondary search of a cache miss (searching for keys starting with `"BTC."` when looking up `"BTC"`). This should be rather easily implemented (if desired), by overriding the `__getitem__` and `__setitem__` methods on these particular objects to do a type check, and if string, perform the `SymbolCache.TryGetValue` result if true, and if false, with the raw string. 

---

#### Related Issue

Fixes #3332 

#### Motivation and Context

To remove the potential for symbol collisions in `SecurityType.Base`

We should update the documentation to describe the new behavior with custom data symbols. This also has implications when creating a custom data universe.

Unit tests added to cover new behavior.
Existing regression tests run (minimal edits for python, breaking change via string/symbol indexing).

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Python/HistoryAlgorithm.py
+++ b/Algorithm.Python/HistoryAlgorithm.py
@@ -44,7 +44,7 @@ class HistoryAlgorithm(QCAlgorithm):
         self.AddEquity("SPY", Resolution.Daily)
         self.AddData(QuandlFuture,"CHRIS/CME_SP1", Resolution.Daily)
         # specifying the exchange will allow the history methods that accept a number of bars to return to work properly
-        self.Securities["CHRIS/CME_SP1.QuandlFuture"].Exchange = EquityExchange()
+        self.Securities["CHRIS/CME_SP1"].Exchange = EquityExchange()
 
         # we can get history in initialize to set up indicators and such
         self.spyDailySma = SimpleMovingAverage(14)
@@ -71,17 +71,17 @@ class HistoryAlgorithm(QCAlgorithm):
             self.spyDailySma.Update(index, tradeBar["close"])
 
         # get the last calendar year's worth of quandl data at the configured resolution (daily)
-        quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1.QuandlFuture", timedelta(365))
-        self.AssertHistoryCount("History(QuandlFuture, \"CHRIS/CME_SP1.QuandlFuture\", timedelta(365))", quandlHistory, 250)
+        quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1", timedelta(365))
+        self.AssertHistoryCount("History(QuandlFuture, \"CHRIS/CME_SP1\", timedelta(365))", quandlHistory, 250)
 
         # get the last 14 bars of SPY at the configured resolution (daily)
-        quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1.QuandlFuture", 14)
-        self.AssertHistoryCount("History(QuandlFuture, \"CHRIS/CME_SP1.QuandlFuture\", 14)", quandlHistory, 14)
+        quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1", 14)
+        self.AssertHistoryCount("History(QuandlFuture, \"CHRIS/CME_SP1\", 14)", quandlHistory, 14)
 
         # we can loop over the return values from these functions and we'll get Quandl data
         # this can be used in much the same way as the tradeBarHistory above
         self.spyDailySma.Reset()
-        for index, quandl in quandlHistory.loc["CHRIS/CME_SP1.QuandlFuture"].iterrows():
+        for index, quandl in quandlHistory.loc["CHRIS/CME_SP1"].iterrows():
             self.spyDailySma.Update(index, quandl["settle"])
 
         # get the last year's worth of all configured Quandl data at the configured resolution (daily)
@@ -95,8 +95,8 @@ class HistoryAlgorithm(QCAlgorithm):
         # NOTE: using different resolutions require that they are properly implemented in your data type, since
         #  Quandl doesn't support minute data, this won't actually work, but if your custom data source has
         #  different resolutions, it would need to be implemented in the GetSource and Reader methods properly
-        #quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1.QuandlFuture", timedelta(7), Resolution.Minute)
-        #quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1.QuandlFuture", 14, Resolution.Minute)
+        #quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1", timedelta(7), Resolution.Minute)
+        #quandlHistory = self.History(QuandlFuture, "CHRIS/CME_SP1", 14, Resolution.Minute)
         #allQuandlData = self.History(QuandlFuture, timedelta(365), Resolution.Minute)
         #allQuandlData = self.History(QuandlFuture, self.Securities.Keys, 14, Resolution.Minute)
         #allQuandlData = self.History(QuandlFuture, self.Securities.Keys, timedelta(1), Resolution.Minute)
@@ -108,14 +108,14 @@ class HistoryAlgorithm(QCAlgorithm):
 
         # we can also access the return value from the multiple symbol functions to request a single
         # symbol and then loop over it
-        singleSymbolQuandl = allQuandlData.loc["CHRIS/CME_SP1.QuandlFuture"]
-        self.AssertHistoryCount("allQuandlData.loc[\"CHRIS/CME_SP1.QuandlFuture\"]", singleSymbolQuandl, 250)
+        singleSymbolQuandl = allQuandlData.loc["CHRIS/CME_SP1"]
+        self.AssertHistoryCount("allQuandlData.loc[\"CHRIS/CME_SP1\"]", singleSymbolQuandl, 250)
         for  quandl in singleSymbolQuandl:
             # do something with 'CHRIS/CME_SP1.QuandlFuture' quandl data
             pass
 
-        quandlSpyLows = allQuandlData.loc["CHRIS/CME_SP1.QuandlFuture"]["low"]
-        self.AssertHistoryCount("allQuandlData.loc[\"CHRIS/CME_SP1.QuandlFuture\"][\"low\"]", quandlSpyLows, 250)
+        quandlSpyLows = allQuandlData.loc["CHRIS/CME_SP1"]["low"]
+        self.AssertHistoryCount("allQuandlData.loc[\"CHRIS/CME_SP1\"][\"low\"]", quandlSpyLows, 250)
         for  low in quandlSpyLows:
             # do something with 'CHRIS/CME_SP1.QuandlFuture' quandl data
             pass
--- a/Algorithm.Python/HistoryAlgorithm.py
+++ b/Algorithm.Python/HistoryAlgorithm.py
@@ -34,9 +34,9 @@ from datetime import datetime, timedelta
 ### <meta name="tag" content="history" />
 ### <meta name="tag" content="warm up" />
 class HistoryAlgorithm(QCAlgorithm):
-    
+
     def Initialize(self):
-    
+
         self.SetStartDate(2013,10, 8)  #Set Start Date
         self.SetEndDate(2013,10,11)    #Set End Date
         self.SetCash(100000)           #Set Strategy Cash
@@ -44,7 +44,7 @@ class HistoryAlgorithm(QCAlgorithm):
         self.AddEquity("SPY", Resolution.Daily)
         self.AddData(QuandlFuture,"CHRIS/CME_SP1", Resolution.Daily)
         # specifying the exchange will allow the history methods that accept a number of bars to return to work properly
-        self.Securities["CHRIS/CME_SP1"].Exchange = EquityExchange()
+        self.Securities["CHRIS/CME_SP1.QuandlFuture"].Exchange = EquityExchange()
 
         # we can get history in initialize to set up indicators and such
         self.spyDailySma = SimpleMovingAverage(14)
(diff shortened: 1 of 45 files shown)


Common/SecurityIdentifier.cs
```diff
@@ -404,6 +404,40 @@ public static SecurityIdentifier GenerateEquity(DateTime date, string symbol, st
             return Generate(date, symbol, SecurityType.Equity, market);
         }
 
+        /// <summary>
+        /// Generates the <see cref="Symbol"/> property for <see cref="QuantConnect.SecurityType.Base"/> security identifiers
+        /// </summary>
+        /// <param name="dataType">The base data custom data type if namespacing is required, null otherwise</param>
+        /// <param name="symbol">The ticker symbol</param>
+        /// <returns>The value used for the security identifier's <see cref="Symbol"/></returns>
+        public static string GenerateBaseSymbol(Type dataType, string symbol)
```
Martin-Molinero: Wondering if it's worth to make this method `public` - Seems to be used just by `GenerateBase()` and testing, maybe we could test its behavior asserting on the result of `GenerateBase()`

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -298,7 +301,7 @@ public void AddUniverse(Type dataType, SecurityType securityType, string name, R
             var marketHoursDbEntry = MarketHoursDatabase.GetEntry(market, name, securityType);
             var dataTimeZone = marketHoursDbEntry.DataTimeZone;
             var exchangeTimeZone = marketHoursDbEntry.ExchangeHours.TimeZone;
-            var symbol = QuantConnect.Symbol.Create(name, securityType, market);
+            var symbol = QuantConnect.Symbol.Create(name, securityType, market, null, dataType);
```
Martin-Molinero: nit - Could avoid sending in `null` specifying `baseDataType:`

Algorithm/QCAlgorithm.cs
```diff
@@ -1810,23 +1810,7 @@ public Security AddData<T>(string ticker, Resolution resolution, bool fillDataFo
...
-                symbol,
-                resolution,
-                fillDataForward,
-                extendedMarketHours: true,
-                isCustomData: true);
-            var security = Securities.CreateSecurity(symbol, config, leverage);
-
-            AddToUserDefinedUniverse(security, new List<SubscriptionDataConfig>{ config });
-            return security;
+            return AddData(typeof(T), ticker, resolution, timeZone, fillDataForward, leverage);
```
Martin-Molinero: Nice duplicate code clean up!

Common/SecurityIdentifier.cs
```diff
@@ -404,6 +404,40 @@ public static SecurityIdentifier GenerateEquity(DateTime date, string symbol, st
...
+                GenerateBaseSymbol(dataType, symbol),
+                SecurityType.Base,
+                market,
+                forceSymbolToUpper: false
+            );
+        }
+
         /// <summary>
         /// Generates a new <see cref="SecurityIdentifier"/> for a custom security
         /// </summary>
```
Martin-Molinero: Old `GenerateBase` is not being used anymore, should we remove it? or add some new documentation for it?

Algorithm.Python/HistoryAlgorithm.py
```diff
@@ -108,16 +108,16 @@ def Initialize(self):
 
         # we can also access the return value from the multiple symbol functions to request a single
         # symbol and then loop over it
-        singleSymbolQuandl = allQuandlData.loc["CHRIS/CME_SP1"]
-        self.AssertHistoryCount("allQuandlData.loc[\"CHRIS/CME_SP1\"]", singleSymbolQuandl, 250)
+        singleSymbolQuandl = allQuandlData.loc["CHRIS/CME_SP1.QuandlFuture"]
```
Martin-Molinero: I think the only breaking change here is that `PandasData` is using `_symbol.ToString()`-> which ends up fetching the `"CHRIS/CME_SP1.QuandlFuture"` from the `SymbolCache` at https://github.com/QuantConnect/Lean/blob/master/Common/Python/PandasData.cs#L231 which breaks the `.loc["CHRIS/CME_SP1"]` -> wonder if we could add some work around to this 🤔 
Calls like `self.Securities["CHRIS/CME_SP1"]` and `self.History(QuandlFuture, "CHRIS/CME_SP1", timedelta(365))` seem to be working correctly

Common/SymbolCache.cs
```diff
@@ -115,6 +130,41 @@ public static void Clear()
...
+                    // too many matches
+                    error = new InvalidOperationException(
+                        "We located multiple potentially matching tickers. For custom data, be sure to " +
+                        "append a dot followed by the custom data type name. For example: 'BTC.Bitcoin'. " +
+                        $"Potential Matches: {string.Join(", ", match.Select(kvp => kvp.Key))}"
+                    );
+                }
+            }
+
+            return Tuple.Create(symbol != null, symbol, error);
```
Martin-Molinero: Personal opinion here 😛 : I'd be more inclined to just throw the exception and catch it if we want to ignore it internally, which will remove the need to return a `Tuple`

Common/SymbolCache.cs
```diff
@@ -115,6 +130,41 @@ public static void Clear()
...
+                }
+                else if (match.Count == 1)
+                {
+                    // exactly one match
+                    symbol = match.Single().Value;
+                }
+                else if (match.Count > 1)
+                {
+                    // too many matches
+                    error = new InvalidOperationException(
```
Martin-Molinero: What would happen if we have `"SPY"` and `"SPY.MyCustomData"`, will the user be able to use `"SPY"` from the cache? Specifically thinking if it will cause this exception to be thrown
I think it would be a good idea to add an algorithm which showcases this issue, has a collision, and what is the expected behavior

Algorithm.Python/HistoryAlgorithm.py
mchandschuh: Hmm... pandas data is using `_symbol.ToString()` -- thanks for that clue! I noticed that the `string` -> `Symbol` implicit conversion was broken for custom data as noted in the PR description. I'll investigate the referenced code to see if we can get it to attempt to resolve via the `SymbolCache`'s new heuristic, allowing resolution of old ticker-only custom symbol strings.

Common/SymbolCache.cs
mchandschuh: _Ironically_ -- it was initially written that way and later refactored following self review. A quick [google](https://www.google.com/search?q=exception+for+control+flow&rlz=1C1CHBD_enUS817US817&oq=exception+for+control+flow&aqs=chrome..69i57.3471j0j7&sourceid=chrome&ie=UTF-8) will yield many examples as to why using exceptions for control flow is an anti-pattern and instead, exceptions should be reserved for truly exceptional cases that are unable to be reasonably handled by the executing code.

Common/SymbolCache.cs
mchandschuh: Since `"SPY"` is a direct match, it will resolve properly. If there isn't already a unit test for this exact case, I'll be sure to add one.

mchandschuh: Added the following unit test:
``` csharp
[Test]
public void TryGetSymbol_FromTicker_WithConflictingSymbolWithCustomDataSuffix()
{
    var ticker = "My-Ticker";
    var symbol = Symbol.Create(ticker, SecurityType.Equity, Market.USA);
    SymbolCache.Set(ticker, symbol);

    var customTicker = $"{ticker}.CustomDataTypeName";
    var customSymbol = Symbol.Create(customTicker, SecurityType.Base, Market.USA, baseDataType: typeof(Bitcoin));
    SymbolCache.Set(customTicker, customSymbol);

    Symbol fetchedSymbol;
    Assert.IsTrue(SymbolCache.TryGetSymbol(ticker, out fetchedSymbol));
    Assert.AreEqual(symbol, fetchedSymbol);

    Symbol fetchedCustomSymbol;
    Assert.IsTrue(SymbolCache.TryGetSymbol(customTicker, out fetchedCustomSymbol));
    Assert.AreEqual(customSymbol, fetchedCustomSymbol);
}
```

Now... you bring up a _really_ great point. Consider `AddData<T>("SPY")` and `AddEquity("SPY")` -- using the `string` `"SPY"` will _always_ resolve to the equity and _never_ to the custom data type, which would require you to specify `"SPY.TypeName"`. This is another potentially breaking change, although I'm not sure how common this collision is in practice.

In the case where there is no potential for collision, say `AddData<T>("SPY")` and `AddEquity("Not-SPY")`, using `string` `"SPY"` will resolve to the only potential match, which would be `"SPY.TypeName"` since there is not entry that exactly matches `"SPY"`.

Common/SecurityIdentifier.cs
mchandschuh: I think this makes sense. This will force consumer to use the overload that requires specifying the `dataType` so the consumer is at least _forced_ to make a decision there (whether to pass `null` or some relevant `Type` instance). 👍  Great comments 👍

Algorithm/QCAlgorithm.Python.cs
mchandschuh: I can dig it 👍

Algorithm/QCAlgorithm.cs
mchandschuh: Sometimes I can't help myself! haha

Common/SecurityIdentifier.cs
mchandschuh: Removed.

mchandschuh:
I've addressed all comments _except_ the one about python and `PandasData` -- I'm investigating potential solutions now.

Martin-Molinero:
`Pandas and CustomData mapping fixes` commit:

`Performance tests HistoryRequestBenchmark - no significant performance degradation seen:`
`pr`
41.72 seconds at 22k data points per second. Processing total of 907,030 data points.
41.12 seconds at 22k data points per second. Processing total of 907,030 data points.
40.53 seconds at 22k data points per second. Processing total of 907,030 data points.
`master`
40.73 seconds at 22k data points per second. Processing total of 907,030 data points.
41.73 seconds at 22k data points per second. Processing total of 907,030 data points.
41.12 seconds at 22k data points per second. Processing total of 907,030 data points.

- Reverts changes on the `HistoryAlgorithm.py`, **keeps backwards compatibility** with the help of a wrapper to the pandas data frame instance which determines the symbol index to use.
- Changes to `RsiAlphaModel` and `BlackLittermanOptimizationPortfolioConstructionModel` way of checking that a ticker is present in the history request result - **Not backwards compatible** - Documentation was reviewed and asserted this operation is not present, although these models are linked to.

Common/SecurityIdentifier.cs
```diff
@@ -277,6 +276,10 @@ public SecurityIdentifier(string symbol, ulong properties)
             _underlying = null;
             _strikePrice = -1;
             SecurityType = (SecurityType)ExtractFromProperties(SecurityTypeOffset, SecurityTypeWidth, properties);
+            if (!Enum.IsDefined(typeof(SecurityType), SecurityType))
+            {
+                throw new ArgumentException($"The provided properties do not march with a valid {nameof(SecurityType)}", "properties");
```
mchandschuh: typo - march => match

Martin-Molinero:
Replaced by https://github.com/QuantConnect/Lean/pull/3588

Issue #3492, opened 2019-08-09
Implement Brain Data Custom Data 


gsalaz98:
#### Expected Behavior

We are able to use Brain Data custom data in LEAN
#### Actual Behavior

We are unable to use Brain Data custom data in LEAN
#### Potential Solution

Implement a data converter + required BaseData implementations
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

Martin-Molinero:
Solved at https://github.com/QuantConnect/Lean.DataSource.BrainSentiment

PR #3493, opened 2019-08-09
Implements Brain Data Custom Data Source and Adds IDataConverter Interface + Implementation

gsalaz98:
#### Description
Implements Brain Data custom data

Add SingleLineDataConverter, an implementation of IDataConverter, with pre-built methods to
allow easy creation of a converter who's data has one data point per line (CSV, TSV, etc.)

#### Related Issue

#3492 
#### Motivation and Context

Implement more custom data sources into LEAN

Reduces the amount of effort required to build a new converter while standardizing the structure of a data converter

No

Local testing of the converter + algorithms
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/BrainData/BrainDataRankingsFiveDay.cs
+++ b/Common/Data/Custom/BrainData/BrainDataRankingsFiveDay.cs
@@ -1,0 +1,95 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Globalization;
+using System.IO;
+
+namespace QuantConnect.Data.Custom.BrainData
+{
+    /// <summary>
+    /// Brain Data Machine Learning Alpha Ranking provides rankings for stocks
+    /// </summary>
+    public class BrainDataRankingsFiveDay : BaseData
+    {
+        /// <summary>
+        /// Ranking data/Machine Learning Alpha score average over period of five days
+        /// </summary>
+        public decimal RankingScore { get; set; }
+
+        /// <summary>
+        /// Empty constructor required for <see cref="Slice.Get{T}()"/>
+        /// </summary>
+        public BrainDataRankingsFiveDay()
+        {
+        }
+
+        /// <summary>
+        /// Gets the source of the data and directs LEAN where to look for it
+        /// </summary>
+        /// <param name="config">Subscription configuration</param>
+        /// <param name="date">Algorithm date</param>
+        /// <param name="isLiveMode">Is live mode</param>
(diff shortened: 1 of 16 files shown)


Common/Data/Custom/BrainData/BrainDataRankingsFiveDay.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <returns>Location of the data</returns>
+        public override SubscriptionDataSource GetSource(SubscriptionDataConfig config, DateTime date, bool isLiveMode)
+        {
+            return new SubscriptionDataSource(
+                Path.Combine(
+                    Globals.DataFolder,
+                    "alternative",
+                    "braindata",
+                    "rankings_five_day",
+                    $"{config.Symbol.Value.ToLower()}.csv"
```
Jay-Jay-D: Are you sure you want to use `Symbol.Value`?

Recall that if the symbol is correctly mapped, `Symbol.Value` is the first part of the `SecurityIdentifier`, meaning the **first** ticker for this security.

Jay-Jay-D: This behavior is observed in all BrainData `BaseData` implementations.

ToolBox/IDataConverter.cs
```diff
@@ -0,0 +1,125 @@
...
+    /// Interface defines common patterns that are involved in implementing a ToolBox converter
+    /// for custom data sources. We provide many generic implementable methods so that your converter
+    /// is capable of processing multiple data sources per instance
+    public interface IDataConverter
+    {
+        /// <summary>
+        /// Responsible for handling discovery of map files that deal with
+        /// mapping events (e.g. delistings, renames, listing after delisting, etc.)
+        /// </summary>
+        MapFileResolver MapFileResolver { get; }
```
Jay-Jay-D: Having a MapFileResolver as a definition in IDataConverter sounds like a loss of generality, it is focused in USA equity markets.

ToolBox/BrainDataConverter/BrainDataConverter.cs
```diff
@@ -0,0 +1,246 @@
...
+        public override T GetDataInstanceFromRaw<T>(string line)
+        {
+            var csv = line.ToCsv();
+
+            if (typeof(T) == typeof(BrainDataSentimentWeekly))
+            {
+                return (T)(BaseData)new BrainDataSentimentWeekly
+                {
+                    Time = DateTime.ParseExact(csv[0], "yyyy-MM-dd", CultureInfo.InvariantCulture),
+                    Symbol = Symbol.Create(csv[1], SecurityType.Base, Market.USA),
```
Jay-Jay-D: `Symbol.Create` is not always the best alternative. under the hood, it calls `GenerateEquity`
https://github.com/QuantConnect/Lean/blob/9306f262c19782636686d0334818213b130f6368/Common/SecurityIdentifier.cs#L351

Meaning that for each line it will try to load a `MapFileProvider`.

ToolBox/IDataConverter.cs
gsalaz98: I think you're right. Seeing this with a fresh mind, I think I need to reconsider the addition of this.

In general, I think this "abstraction" needs more thought, since trying to shoehorn in mapping as is is not a solid design pattern, especially considering that what you've said holds a lot of truth.

Common/Data/Custom/BrainData/BrainDataRankingsFiveDay.cs
gsalaz98: I was under the impression that `Symbol.Value` contained the most recently mapped ticker, not the first ticker. I believe the first ticker can be accessed by `Symbol.ID.Symbol`.

Jay-Jay-D: You are correct!

Sorry, I was confused.

jaredbroad:
image

Let's not create more standards. Interfaces should hide not expose implementation details of the action they're providing a pattern to achieve.

Issue #3504, opened 2019-08-13
TradingEconomicsCalendar Does Not Properly Set Symbol


gsalaz98:
#### Expected Behavior

* We are able to determine whether the slice contains data by using `slice.ContainsKey(symbol)`
* No errors occur during runtime
#### Actual Behavior

* `slice.ContainsKey(symbol)` will always fail, because the Symbol is not properly set (defaults to "")
* Sometimes we get runtime errors due to failed clone because BaseData `Clone` method fails
#### Potential Solution

Rename `Symbol` property in TradingEconomicsCalendar to `TESymbol` and implement Clone method to prevent any errors from surfacing
#### Reproducing the Problem

Run `TradingEconomicsCalendarIndicatorAlgorithm.py` - only Indicator data is outputted. It can also fail in select systems due to clone not being properly implemented
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

gsalaz98:
Closed by #3506

PR #3506, opened 2019-08-13
Fixes TradingEconomicsCalendar Empty Symbol In Algorithms and Non-Deterministic Crashes

gsalaz98:
#### Description

TradingEconomicsCalendar would previously not contain the Symbol due to the shadowing of the BaseData reserved property. We renamed `Symbol` to `TESymbol` in order to allow the Symbol property to be set with a Symbol object.

Previously, there would be non-deterministic crashing as a result of the BaseData Clone method failing. We have overridden Clone with our own implementation to prevent it from happening
#### Related Issue

#3504 
#### Motivation and Context

Stability and correctness

Maybe - `Symbol` is now `TESymbol`

Local testing + testing by Daniel (PR #3502 was created with these changes)
#### Types of changes

#### Checklist:

- [ ] I have added tests to cover my changes.  - Regression test required, although the data is not available to open-source users

### DIFF
--- a/Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
+++ b/Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
@@ -235,7 +235,7 @@ namespace QuantConnect.Data.Custom.TradingEconomics
         /// </summary>
         public override string ToString()
         {
-            var symbol = string.IsNullOrWhiteSpace(TESymbol) ? TESymbol : Ticker;
+            var symbol = string.IsNullOrWhiteSpace(TESymbol) ? Ticker : TESymbol;
             return $"{symbol} ({Country} - {Category}): {Event} : Importance.{Importance}";
         }
     }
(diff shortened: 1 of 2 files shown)


Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -150,7 +150,7 @@ public class TradingEconomicsCalendar : BaseData
         /// Unique symbol used by Trading Economics
         /// </summary>
         [JsonProperty(PropertyName = "Symbol")]
-        public string Symbol { get; set; }
+        public string TESymbol { get; set; }
```
mchandschuh: In LEAN parlance, can call this `Ticker` as it's just the `string` and not a full `Symbol` object.

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -196,12 +196,46 @@ public override BaseData Reader(SubscriptionDataConfig config, string content, D
...
+            };
+        }
+
         /// <summary>
         /// Formats a string with the Trading Economics Calendar information.
         /// </summary>
         public override string ToString()
         {
-            var symbol = string.IsNullOrWhiteSpace(Symbol) ? Symbol : Ticker;
+            var symbol = string.IsNullOrWhiteSpace(TESymbol) ? TESymbol : Ticker;
```
StefanoRaggi: The condition seems inverted here

PR #3507, opened 2019-08-14
Removes always false condition checking in BaseSetupHandler

IlshatGaripov:
#### Description
1) I noticied, This block of code in BaseSetupHandler.cs is probably always false and can be removed:

if (historyRequests.Any(x => x.Symbol == cash.ConversionRateSecurity.Symbol))
{
  continue;
}

Because if it's ever true -> then some two Cash objects have same ConversionRateSecurity - > then one of these Cash object must be a base currency? But base currency's ConversionRateSecurity is null.

2) Also, adds xml doc for HistoryRequestFactory.GetStartTimeAlgoTz()

#### Related Issue
n/a

#### Motivation and Context
small contribution to make lean code better

#### Types of changes

- [X ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

### DIFF
--- a/Engine/Setup/BaseSetupHandler.cs
+++ b/Engine/Setup/BaseSetupHandler.cs
@@ -54,12 +54,6 @@ namespace QuantConnect.Lean.Engine.Setup
             var historyRequests = new List<HistoryRequest>();
             foreach (var cash in cashToUpdate)
             {
-                // if we already added a history request for this security, skip
-                if (historyRequests.Any(x => x.Symbol == cash.ConversionRateSecurity.Symbol))
-                {
-                    continue;
-                }
-
                 var configs = algorithm
                     .SubscriptionManager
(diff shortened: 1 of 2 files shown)


Engine/Setup/BaseSetupHandler.cs
```diff
@@ -54,12 +54,6 @@ public static class BaseSetupHandler
             var historyRequests = new List<HistoryRequest>();
             foreach (var cash in cashToUpdate)
             {
-                // if we already added a history request for this security, skip
-                if (historyRequests.Any(x => x.Symbol == cash.ConversionRateSecurity.Symbol))
```
Martin-Molinero: Good catch, this shouldn't happen, if two cash share the same conversion rate security it should be the same cash symbol, which isn't a valid scenario

PR #3508, opened 2019-08-14
Reduces performance cost of new operator in Cash.cs

IlshatGaripov:
#### Description
I noticed that Cash.cs EnsureCurrencyDataFeed method creates lots of Symbol objects when fetching appropriate currency pair to construct currency conversion subscription data config. Due to performance cost of **new** operator in C#:
https://stackoverflow.com/questions/4253728/performance-cost-of-new-in-c
I think the code could be improved if we limited ourselves to creating a single Symbol after iterating over all possible currency pair string collections first

#### Related Issue
n/a

#### Motivation and Context
lean performance

#### Types of changes

- [X ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [ X] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -222,16 +222,26 @@ namespace QuantConnect.Securities
                 markets.Add(SecurityType.Cfd, markets[SecurityType.Forex]);
             }
 
-            var potentials = Currencies.CurrencyPairs.Select(fx => CreateSymbol(marketMap, fx, markets, SecurityType.Forex))
-                .Concat(Currencies.CfdCurrencyPairs.Select(cfd => CreateSymbol(marketMap, cfd, markets, SecurityType.Cfd)))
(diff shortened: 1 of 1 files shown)


Martin-Molinero:
Hi @IlshatGaripov !
Thanks for the PR, performance changes should be accompanied by a performance test, and results, which shows the improvement of the `PR` vs `master`, with details of what was tested, in what conditions, so that reviewers can arrive to the same results. 
I believe performance changes should show a measureable improvement to justify the refactor, and the associated risk. A good starting point is take an unperformant scenario/use case and find improvements points.

IlshatGaripov:
Hi! @Martin-Molinero .  Thansk for advice. I tested a block of code that calls the method with StopWatch now. Sorry haven't done this before. You are correct. Indeed, No significant improvement. All this commit must be an effect of book I read - CLR via C# - where an author constatntly considers matters from perfromance hit perspective . Well it needs a little experience to combine a theory and practise! thanks!

Issue #110, opened 2015-06-10
Make Account Currency Configurable


jaredbroad:
In CashBook class there's a constant AccountCurrency, this serves as the base currency for the portfolio. To make this a configurable property:
1. IBrokerage .GetCashBalances() should return an "account currency flag"
2. ISetupHandler would use this to set the account currency for the cashbook.

Worthy7:
Can you write this up into the documentation please

jaredbroad:
Notes from @StefanoRaggi 

```
- Move CashBook.AccountCurrency to QCAlgorithm.AccountCurrency (instance property, not static):
 - can only be set in Initialize, locked after

- Add QCAlgorithm.SetAccountCurrency:
 - should be OK to be called in Initialize, brokerage.GetCashBalance/GetOpenOrders/GetAccountHoldings are called after Initialize
 - some brokerages (e.g. GDAX) force an account currency (EUR for European users)
 - other brokerages (Oanda) allow user to choose an account currency at account creation time (USD, EUR, ...)
 - other exchanges (Binance) do not have fiat currencies (so user should be allowed to select BTC, ETH, ...)

- SecurityPortfolioModel always subtracts fees assuming in account currency
 - should be using quote currency (or the brokerage-specific currency for fees, if different from quote currency)
 - may need to refactor IFeeModel.GetOrderFee to return an object (value+currency) instead of a decimal (this might be a separate PR to be done first)

- Depending on the account currency chosen, this issue might need to be resolved first (not required for GDAX):
https://github.com/QuantConnect/Lean/issues/2060```

voidale:
Any update on this? :)

voidale:
@Martin-Molinero Awesome, good luck finishing that one, definitely will be using that when it's available.

Doggie52:
I'm assuming this would allow us to choose which currency to base our live deployment on as well? If so then this would be great news.

Correct me if I'm wrong, but currently the only "problem" with a live deployment is that the runtime statistics are reported in USD (and hence subject to <baseccy>USD fluctuations)? Otherwise it's not like orders are treated any differently if the funds are in <baseccy>, right, as that's all managed on the broker side?

StefanoRaggi:
Closed by #4806

PR #3511, opened 2019-08-15
Binance Brokerage implementation

### LINKED ISSUE #2004: [Crypto] Implement Binance Bitcoin Exchange
Seeking a complete, tested [Binance](https://www.binance.com) bitcoin exchange implementation for QuantConnect, including the IDQH implementation for streaming binance data.  

1 - IBrokerage interface implementation for Binance. All IBrokerage tests should pass.
2 - IQueueDataHandler interface implementation for Binance streaming data.
3 - If possible IDataDownloader implementation for Lean/ToolBox for Binance API.

#### Checklist

AdalyatNazirov:
Implement Binance cryptocurrency trading platform 
#### Description

Implements websocket based brokerage

#### Related Issue

Closes #2004, Relates #110 
 
#### Motivation and Context

Provides new trading platform

Requires changes in documentation for IDataDownloader as it supports new arguments values.

Unit tests covers new code

#### Types of changes

#### Checklist:


Brokerages/Binance/BinanceBrokerage.Messaging.cs
```diff
@@ -0,0 +1,317 @@
...
+using QuantConnect.Securities;
+using RestSharp;
+using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using System.Net;
+
+namespace QuantConnect.Brokerages.Binance
+{
+    public partial class BinanceBrokerage
```
mchandschuh: style - I never was a fan of the partial class pattern for brokerages. My preference would be to see multiple smaller classes instead of a single master class. The partial class pattern has been used in others, so we don't _need_ to change this as it matches existing patterns, but I do think it's worthwhile thinking about whether or not it's the best pattern. I could see what is currently a single brokerage being split into (for example) a `BinanceApiClient` to manage the request/response between lean and binance and a `BinanceBrokerage` that would use the `BinanceApiClient` as an adapter for the `IBrokerage` interface. This provides a nice separation of concerns and makes other components available for reuse.

Brokerages/Binance/BinanceBrokerage.Messaging.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// </summary>
+        private void UnlockStream()
+        {
+            Log.Trace("BinanceBrokerage.Messaging.UnlockStream(): Processing Backlog...");
+            while (_messageBuffer.Any())
+            {
+                WebSocketMessage e;
+                _messageBuffer.TryDequeue(out e);
+                OnMessageImpl(this, e);
+                OnTickerMessageImpl(this, e);
```
mchandschuh: Why have both? Why not have `OnMessageImpl` delegate to `OnTickerMessageImpl` when necessary? Seems like we end up doing some double work parsing the message twice.

Brokerages/Binance/BinanceBrokerage.Messaging.cs
```diff
@@ -0,0 +1,317 @@
...
+                }
+            }
+            catch (Exception exception)
+            {
+                OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Error, -1, $"Parsing wss message failed. Data: {e.Message} Exception: {exception}"));
+                throw;
+            }
+        }
+
+        private void OnOrderBookUpdate(OrderBookUpdateMessage ticker)
```
mchandschuh: Perhaps outside the scope of this PR, but it would be nice to have an `OrderBookCollection` or `OrderBookManager` of sorts as a self-contained component. Often times, such is the case with binance,  you can bind to their L2 data feed _without_ credentials. This could allow an algorithm trading on, for example, coinbase, to still get the binance quote books for usage within the algorithm.

Brokerages/Binance/BinanceBrokerage.Messaging.cs
```diff
@@ -0,0 +1,317 @@
...
+        private void FetchOrderBookSnapshot(Symbol symbol, BinanceOrderBook orderBook)
+        {
+            LockStream();
+
+            var endpoint = $"/api/v1/depth?symbol={_symbolMapper.GetBrokerageSymbol(symbol)}&limit=1000";
+            var request = new RestRequest(endpoint, Method.GET);
+            var response = ExecuteRestRequest(request);
+
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
```
mchandschuh: Looks like the stream doesn't get unlocked in this code path. When doing lock/unlock semantics, it's best to have it be done using a `try/finally` block to guarantee we always call `UnlockStream`. Another option is to encapsulate the locking in a single function:
``` csharp
private void WithLockedStream(Action code)
{
    try
    {
        LockStream();
        code();
    }
    finally
    {
        UnlockStream();
    }
}
```
and then consumers can call via:
``` csharp
WithLockedStream(() =>
{
    // code that requires the stream to be locked
});
```

mchandschuh: Also, it's most likely the case that we don't need to lock the stream over the _entire_ method body, but only the sections that are sensitive to the stream being locked.

Brokerages/Binance/BinanceBrokerage.Messaging.cs
```diff
@@ -0,0 +1,317 @@
...
+            }
+
+            foreach (var item in asks)
+            {
+                var price = (item[0] as string).ToDecimal();
+                var quantity = (item[1] as string).ToDecimal();
+                if (quantity == 0)
+                    orderBook.RemoveAskRow(price);
+                else
+                    orderBook.UpdateAskRow(price, quantity);
```
mchandschuh: outside the scope of this PR, but I think for encapsulation purposes it makes sense for `UpdateAskRow` to perform the `quantity == 0` check internally to guarantee its own state is consistent.

Brokerages/Binance/BinanceBrokerage.Streaming.cs
```diff
@@ -0,0 +1,86 @@
+﻿using Newtonsoft.Json.Linq;
```
mchandschuh: Please add missing license banner.

Brokerages/Binance/BinanceBrokerage.Streaming.cs
```diff
@@ -0,0 +1,86 @@
...
+            var pong = ExecuteRestRequest(ping);
+            return pong.StatusCode == HttpStatusCode.OK;
+        }
+
+        /// <summary>
+        /// Stops the session
+        /// </summary>
+        public void StopSession()
+        {
+            if (!string.IsNullOrEmpty(SessionId))
```
mchandschuh: Is the inverse an exception? In other words, if `string.IsNullOrEmpty(SessionId)` -- is that an error case such as treated in `SessionKeepAlive()` ?

Brokerages/Binance/BinanceBrokerage.Utility.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// <returns></returns>
+        private IRestResponse ExecuteRestRequest(IRestRequest request)
+        {
+            const int maxAttempts = 10;
+            var attempts = 0;
+            IRestResponse response;
+
+            do
+            {
+                if (!_restRateLimiter.WaitToProceed(TimeSpan.Zero))
```
mchandschuh: If this is the desired retry request behavior, it might make sense to extract into it's own type so other types can benefit from it.

Brokerages/Binance/BinanceBrokerage.Utility.cs
```diff
@@ -0,0 +1,177 @@
...
+        private decimal GetTickerPrice(Order order)
+        {
+            var security = _securityProvider.GetSecurity(order.Symbol);
+            var tickerPrice = order.Direction == OrderDirection.Buy ? security.AskPrice : security.BidPrice;
+            if (tickerPrice == 0)
+            {
+                var tickers = GetTickers();
+                var ticker = tickers.FirstOrDefault(t => t.Symbol == _symbolMapper.GetBrokerageSymbol(order.Symbol));
+                if (ticker == null)
+                    throw new NotSupportedException($"BinanceBrokerage: Unable to resolve currency conversion pair: {order.Symbol}");
```
mchandschuh: `KeyNotFoundException` may be better suited here.

Brokerages/Binance/BinanceBrokerageFactory.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
```
mchandschuh: Please add missing license banner.

Brokerages/Binance/BinanceOrderBook.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
```
mchandschuh: Please add missing license banner.

Brokerages/Binance/BinanceOrderBook.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Threading.Tasks;
+
+namespace QuantConnect.Brokerages.Binance
+{
+    /// <summary>
+    /// Represents a full order book for a security.
+    /// It contains prices and order sizes for each bid and ask level.
+    /// The best bid and ask prices are also kept up to date.
+    /// </summary>
+    public class BinanceOrderBook: DefaultOrderBook
```
mchandschuh: Are these intended to be completely self-contained/auto-updating or simply data structures?

@StefanoRaggi @Martin-Molinero 
I was thinking it would be nice if order books were self contained, such that they manage their own auto-update, etc. Thoughts?

Brokerages/Binance/BinanceSymbolMapper.cs
```diff
@@ -0,0 +1,268 @@
...
+                }
+                return symbols;
+            }
+        }
+
+        /// <summary>
+        /// The list of known Binance symbols.
+        /// https://api.Binance.com/v1/symbols
+        /// </summary>
+        public static readonly HashSet<string> KnownSymbolStrings = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
mchandschuh: Is there a way to resolve this information on launch, otherwise we'll always be out of date with supported symbols. Also, in some cases, symbols might not be the same on different exchanges. For example, there's multiple tokens that trade under the ticker `HOT` -- this shouldn't cause issues when running a single brokerage, but can potentially cause SID collisions.

Brokerages/Binance/Messages.cs
```diff
@@ -0,0 +1,173 @@
+﻿using Newtonsoft.Json;
```
mchandschuh: Please add missing license banners.

Common/Brokerages/BinanceBrokerageModel.cs
```diff
@@ -0,0 +1,70 @@
+﻿using QuantConnect.Orders.Fees;
```
mchandschuh: Please add missing license banner.

Common/Currencies.cs
```diff
@@ -85,7 +85,13 @@ public static class Currencies
             "EOSBTC",
             "XLMBTC",
             "ETCBTC",
-            "ZRXBTC"
+            "ZRXBTC",
+
+            //binance
+            "BTCUSDT",
+            "BCHUSDT",
+            "LTCUSDT",
```
mchandschuh: I wonder if we need this static data to be resolved on application start and/or periodically for long-running live algorithms

Common/Orders/BinanceOrderProperties.cs
```diff
@@ -0,0 +1,26 @@
+﻿using QuantConnect.Interfaces;
```
mchandschuh: Please add missing license banner.

Common/Orders/Fees/BinanceFeeModel.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// <summary>
+    /// Provides an implementation of <see cref="FeeModel"/> that models Binance order fees
+    /// </summary>
+    public class BinanceFeeModel : FeeModel
+    {
+        /// <summary>
+        /// Tier 1 maker fees
+        /// https://www.binance.com/en/fee/schedule
+        /// </summary>
+        public const decimal MakerFee = 0.001m;
```
mchandschuh: I think we should support a constructor that permits setting these values, but make tier 1 fees the default value for the default constructor.

Data/symbol-properties/symbol-properties-database.csv
```diff
@@ -716,3 +716,593 @@ bitfinex,VSYBTC,crypto,VSYBTC,BTC,1,0.00001,0.00000001
...
+binance,FTMUSDC,crypto,FTMUSDC,USDC,1,0.00001000,0.10000000
+binance,BTCBBTC,crypto,BTCBBTC,BTC,1,0.00000100,0.00100000
+binance,BCPTTUSD,crypto,BCPTTUSD,TUSD,1,0.00001000,0.10000000
+binance,BCPTPAX,crypto,BCPTPAX,PAX,1,0.00001000,0.10000000
+binance,BCPTUSDC,crypto,BCPTUSDC,USDC,1,0.00001000,0.10000000
+binance,ALGOBNB,crypto,ALGOBNB,BNB,1,0.00001000,0.10000000
+binance,ALGOBTC,crypto,ALGOBTC,BTC,1,0.00000010,0.01000000
+binance,ALGOUSDT,crypto,ALGOUSDT,USDT,1,0.00100000,0.00100000
+binance,ALGOTUSD,crypto,ALGOTUSD,TUSD,1,0.00100000,0.00100000
+binance,ALGOPAX,crypto,ALGOPAX,PAX,1,0.00100000,0.00100000
```
mchandschuh: awesome! what did you use to generate this set?

Engine/Setup/BrokerageSetupHandler.cs
```diff
@@ -216,12 +216,12 @@ public bool Setup(SetupHandlerParameters parameters)
                         //Initialise the algorithm, get the required data:
                         algorithm.Initialize();
 
-                        if (algorithm.AccountCurrency != Currencies.USD)
-                        {
-                            throw new NotImplementedException(
-                                "BrokerageSetupHandler.Setup(): calling 'QCAlgorithm.SetAccountCurrency()' " +
-                                "is only supported in backtesting for now.");
-                        }
+                        //if (algorithm.AccountCurrency != Currencies.USD)
```
mchandschuh: Is this change intentional?

Launcher/config.json
```diff
@@ -274,6 +278,20 @@
       "history-provider": "BrokerageHistoryProvider"
     },
 
+    "live-binance": {
```
mchandschuh: AWESOME! thanks for adding the `live-binance` config environment

Tests/Brokerages/Binance/BinanceBrokerageHistoryProviderTests.cs
```diff
@@ -0,0 +1,162 @@
+﻿using NodaTime;
```
mchandschuh: Please add missing license banner.

Tests/Brokerages/Binance/BinanceBrokerageTests.cs
```diff
@@ -0,0 +1,129 @@
+﻿using System;
```
mchandschuh: Please add missing license banner.

Tests/Brokerages/Binance/BinanceBrokerageTests.cs
```diff
@@ -0,0 +1,129 @@
...
+using Moq;
+using QuantConnect.Brokerages;
+using QuantConnect.Tests.Common.Securities;
+using QuantConnect.Orders;
+using QuantConnect.Logging;
+using System.Threading;
+
+namespace QuantConnect.Tests.Brokerages.Binance
+{
+    [TestFixture, Ignore("This test requires a configured and testable Oanda practice account")]
```
mchandschuh: Some copy/pasta -> `Oanda practice account`

Tests/Brokerages/Binance/BinanceFeeModelTests.cs
```diff
@@ -0,0 +1,135 @@
+﻿using NUnit.Framework;
```
mchandschuh: Please add missing license banner.

Tests/Brokerages/Binance/BinanceSymbolMapperTests.cs
```diff
@@ -0,0 +1,160 @@
+﻿using NUnit.Framework;
```
mchandschuh: Please add missing license banner.

ToolBox/BinanceDownloader/BinanceDataDownloader.cs
```diff
@@ -0,0 +1,113 @@
+﻿using NodaTime;
```
mchandschuh: Please add missing license banner.

ToolBox/BinanceDownloader/BinanceDownloaderProgram.cs
```diff
@@ -0,0 +1,68 @@
+﻿using QuantConnect.Configuration;
```
mchandschuh: Please add missing license banner.

Brokerages/Binance/BinanceSymbolMapper.cs
Adalyat: Not really. We discussed it months ago and this feature isn't implemented yet. It would be nice to have a separate class to fetch all up-to-date static data on init.

Common/Currencies.cs
Adalyat: Agreed. See my response above (https://github.com/QuantConnect/Lean/pull/3511#discussion_r316012763)

Data/symbol-properties/symbol-properties-database.csv
Adalyat: I've written console util to fetch Binance trading symbols metadata and dump them in required format. But I would like to have an opportunity to do that within Brokerage on application start (https://github.com/QuantConnect/Lean/pull/3511#discussion_r316012763)

Engine/Setup/BrokerageSetupHandler.cs
Adalyat: Yeah. But might be supplied as separate PR. It's necessary for Binance as it uses USDT as account currency. But for now I've commented it out to attract additional attention :)

ToolBox/BinanceDownloader/BinanceDownloaderProgram.cs
Adalyat: done

ToolBox/BinanceDownloader/BinanceDataDownloader.cs
Adalyat: done

Tests/Brokerages/Binance/BinanceSymbolMapperTests.cs
Adalyat: done

Tests/Brokerages/Binance/BinanceFeeModelTests.cs
Adalyat: done

Tests/Brokerages/Binance/BinanceBrokerageTests.cs
Adalyat: done

Tests/Brokerages/Binance/BinanceBrokerageHistoryProviderTests.cs
Adalyat: done

Common/Orders/BinanceOrderProperties.cs
Adalyat: done

Common/Brokerages/BinanceBrokerageModel.cs
Adalyat: done

Brokerages/Binance/Messages.cs
Adalyat: done

Brokerages/Binance/BinanceOrderBook.cs
Adalyat: done

Brokerages/Binance/BinanceBrokerageFactory.cs
Adalyat: done

Brokerages/Binance/BinanceBrokerage.Streaming.cs
Adalyat: done

Tests/Brokerages/Binance/BinanceBrokerageTests.cs
Adalyat: done

Brokerages/Binance/BinanceBrokerage.Utility.cs
Adalyat: I've checked other brokerages, and they just throw `Exception`  in that cases. I've changed to `KeyNotFoundException`

Launcher/config.json
Adalyat: You are welcome! 😄 
P.S. could you resolve this conversation if you don't expect any changes here

Brokerages/Binance/BinanceBrokerage.Streaming.cs
Adalyat: hmm.. should we throw exception if SessionId is null or empty? In my mind, there is no any point to do that as it's already empty (matches to our expectations), that's why silence in this case.

Brokerages/Binance/BinanceBrokerage.Messaging.cs
Adalyat: I reworked message handling. Should be better - no more double parsing

Common/Orders/Fees/BinanceFeeModel.cs
Adalyat: done

mchandschuh:
@Adalyat -- please respond directly to each comment. with large PRs like this it can be hard to go back and correlate all of these comments back to changes made in the code. it makes it _much_ easier to let the reviewer know that you agree with a comment/implemented as suggested/discuss options/etc

Brokerages/Binance/BinanceBrokerage.Streaming.cs
mchandschuh: I would say yes since it implies the consumer of this class _thinks_ there's an open session and is trying to stop it. It's like trying to dispose an object that was never initialized. This is not a critical issue though, more of a style/preference around temporal coupling between methods in a class.

Brokerages/Binance/BinanceSymbolMapper.cs
mchandschuh: I think the real issue is that it is **not** static data. Crypto exchanges are notorious for adding/removing symbols quite frequently. So if we're running a live algorithm we'll need this data to (potentially) produce symbol delisted events and maybe (eventually) an IPO-like event (symbol listed?) I'm happy to keep this as static data as a first revision solution, but we'll need to come up with a good pattern for managing the additions/removals from crypto exchanges.

Data/symbol-properties/symbol-properties-database.csv
mchandschuh: Very cool! It may be useful if you post your binance console app to your github and leave a link in this PR's description. We may be able to incorporate some of your console app into a `ToolBox` app.

Engine/Setup/BrokerageSetupHandler.cs
mchandschuh: Attention attracted indeed!

This requires broader support for non-USD account currencies to be properly implemented.

Adalyat: Created a new issue #3593 that covers this problem

jaredbroad: @Adalyat I think we need broader non-USD support as a separate PR as a precursor to this one. Let's connect offline to discuss the work involved for that distinct PR.

nazbrok:
does this PR support Binance futures  or do we need a specific brokerage to be able to short/long on the binance futures platform ?

Adalyat:
> does this PR support Binance futures or do we need a specific brokerage to be able to short/long on the binance futures platform ?

@nazbrok , hello. Current PR supports cash trading only. Margin trading will be implemented as part of this brokerage but in further work.

nazbrok:
> > does this PR support Binance futures or do we need a specific brokerage to be able to short/long on the binance futures platform ?
> 
> @nazbrok , hello. Current PR supports cash trading only. Margin trading will be implemented as part of this brokerage but in further work.

Great @Adalyat . I noticed you have a bitmex brokerage implementation, is it currently working ? do you plan to make a PR to integrate it in Lean official branch ?

captholley:
I have been messing with this pull request to try and run a strategy I am currently running using Margin and Futures trading. The edits I have made to this are very crude. During my dive into this pull request I am noticing some issues:

There are a lot of issues with the timestamp being out of range. This was fixed in a Binance .NET repository I have worked with a lot. I would consider adding the following to the API rest file.

`       private readonly bool autoTimestamp = true;
        private readonly TimeSpan autoTimestampRecalculationInterval = TimeSpan.FromHours(3);
        private readonly TimeSpan timestampOffset = TimeSpan.Zero;
        private double calculatedTimeOffset;
        private bool timeSynced;
        private DateTime lastTimeSync;
        public int TotalRequestsMade = 1;

        /// <summary>
        /// Timestamp in milliseconds
        /// </summary>
        /// <returns></returns>
        private long GetNonce()
        {
            var offset = autoTimestamp ? calculatedTimeOffset : 0;
            offset += timestampOffset.TotalMilliseconds;
            return ToUnixTimestamp(DateTime.UtcNow.AddMilliseconds(offset));
        }
        private static long ToUnixTimestamp(DateTime time)
        {
            return (long)(time - new DateTime(1970, 1, 1)).TotalMilliseconds;
        }

        private DateTime CheckAutoTimestamp()
        {
            if (autoTimestamp && (!timeSynced || DateTime.UtcNow - lastTimeSync > autoTimestampRecalculationInterval))
                return GetServerTime(timeSynced);

            return new DateTime(default);
        }

        /// <summary>
        /// Requests the server for the local time. This function also determines the offset between server and local time and uses this for subsequent API calls
        /// </summary>
        /// <param name="resetAutoTimestamp">Whether the response should be used for a new auto timestamp calculation</param>
        /// <returns>Server time</returns>
        public DateTime GetServerTime(bool resetAutoTimestamp = false)
        {
            var endpoint = $"/api/v3/time";
            var request = new RestRequest(endpoint, Method.GET);
            request.AddHeader(KeyHeader, ApiKey);

            if (!autoTimestamp)
            {
                var response = ExecuteRestRequest(request);
                var result = JsonConvert.DeserializeObject<Messages.BinanceCheckTime>(response.Content);
                if (!response.IsSuccessful)
                    return new DateTime(default);

                return result.ServerTime;
            }
            else
            {
                var localTime = DateTime.UtcNow;
                var response = ExecuteRestRequest(request);
                if (!response.IsSuccessful)
                    return new DateTime(default);

                var result = JsonConvert.DeserializeObject<Messages.BinanceCheckTime>(response.Content);

                if (timeSynced && !resetAutoTimestamp)
                    return result.ServerTime;

                if (TotalRequestsMade == 1)
                {
                    // If this was the first request make another one to calculate the offset since the first one can be slower
                    localTime = DateTime.UtcNow;
                    response = ExecuteRestRequest(request);
                    if (!response.IsSuccessful)
                        return new DateTime(default);
                    result = JsonConvert.DeserializeObject<Messages.BinanceCheckTime>(response.Content);
                    TotalRequestsMade++;
                }

                // Calculate time offset between local and server
                var offset = (result.ServerTime - localTime).TotalMilliseconds;
                if (offset >= 0 && offset < 500)
                {
                    // Small offset, probably mainly due to ping. Don't adjust time
                    calculatedTimeOffset = 0;
                    timeSynced = true;
                    lastTimeSync = DateTime.UtcNow;
                    Log.Trace("Brokerage.GetServerTime(): " + new BrokerageMessageEvent(BrokerageMessageType.Information, "0",
                        $"Time offset between 0 and 500ms ({offset}ms), no adjustment needed"));
                    return result.ServerTime;
                }

                calculatedTimeOffset = (result.ServerTime - localTime).TotalMilliseconds;
                timeSynced = true;
                lastTimeSync = DateTime.UtcNow;
                Log.Trace("Brokerage.GetServerTime(): " + new BrokerageMessageEvent(BrokerageMessageType.Information, "0",
                    $"Time offset set to {calculatedTimeOffset}ms"));
                return result.ServerTime;
            }
        }
`

Then call the checkautotimestamp() method prior to all private calls (orders, balances, etc.)

In addition to this, the Execution message event handler is not working properly for sells. Binance returns the executedQty as a positive no matter what. This needs to be negative for Sell orders or else the SecurityManager does not keep track of quantities properly. I changed the fillQuantity under OnFillOrder with the following and it works now:
`var fillQuantity = data.Direction == OrderDirection.Sell ? -data.LastExecutedQuantity : data.LastExecutedQuantity;`

joao-neves95:
I don't know if it was already mentioned, but this is a project worth mentioning: [Binance.Net](https://github.com/JKorf/Binance.Net).
It could facilitate a faster full Binance integration.

captholley:
So is the main issue with this the fact we are unable to change the account currency? Would it not be possible to deploy an instance of the binance futures brokerage trading seeing as this is done almost exclusively in USDT? I understand this is not exactly USD but it behaves similarly enough to where it could work, no? I mean, I have a working instance on my own desktop I created...

StefanoRaggi:
Replaced by #4688

gelinger777:
@captholley  could you make it work with Margin trading in Binance?

captholley:
yeah, its not too difficult to do. You just have to change the api base url to the margin one (check out binance api docs) and implement some workarounds within the QC framework (set the leverage, alter the order types to pay back margin, etc.). The hard part is getting it to work between the spot/margin account. Binance has these structured as separate accounts and QC is not well suited for multi-accounts/brokerages. Nothing I have made is robust enough to warrant a pull request unfortunately

gelinger777:
@captholley  how can I PM you? I want to tell you smth important?

an-xioneer:
@AdalyatNazirov Do you have any guesses when we can use the Binance integration in the QuantConnect terminal? Right now, the terminal only supports Bitfinex for crypto trading.

AdalyatNazirov:
@an-xioneer have no idea. Maybe @StefanoRaggi or @Martin-Molinero can shade light on roadmap and ETA?

Martin-Molinero:
Hey! Sorry for the late response. It will be available shortly in the cloud, we are finishing some stability testing and improvements https://github.com/QuantConnect/Lean/pull/5773

PR #3514, opened 2019-08-16
Engine code clean up

Martin-Molinero:
#### Description

- Remove `ISetupHandler` from the `IResultHandler`
- Expanding `BaseResultsHandler` to reduce code duplication in the Live and Backtesting ResultHandler
- Removing unused `DesktopResultHandler`
- Normalizing members initialization in type constructors

#### Related Issue

N/A
#### Motivation and Context

Code clean up

N/A

Unit and regression tests
#### Types of changes

#### Checklist:


Engine/AlgorithmManager.cs
```diff
@@ -95,37 +85,33 @@ public TimeSpan CurrentTimeStepElapsed
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="AlgorithmManager"/> class
         /// </summary>
         /// <param name="liveMode">True if we're running in live mode, false for backtest mode</param>
         public AlgorithmManager(bool liveMode)
         {
+            _lock = new object();
+            var timeLoopMaximum
+                = TimeSpan.FromMinutes(Config.GetDouble("algorithm-manager-time-loop-maximum", 20));
```
mchandschuh: Perhaps outside the scope of this PR, but I often find it nice to remove `Config` access from types. In this case, the object instantiating the `AlgorithmManager` could simply pass in the value of  `algorithm-manager-time-loop-maximum`.

I think this is fine as is, simply sharing some of my opinionated styling :)

Engine/Engine.cs
```diff
@@ -48,38 +47,32 @@ namespace QuantConnect.Lean.Engine
...
+        /// <param name="liveMode">True when running in live mode, false otherwise</param>
         public Engine(LeanEngineSystemHandlers systemHandlers, LeanEngineAlgorithmHandlers algorithmHandlers, bool liveMode)
         {
             _liveMode = liveMode;
-            _systemHandlers = systemHandlers;
-            _algorithmHandlers = algorithmHandlers;
+            SystemHandlers = systemHandlers;
+            AlgorithmHandlers = algorithmHandlers;
+            // this is slow, so we will do it on demand
+            _exceptionInterpreter = new Lazy<StackExceptionInterpreter>(() =>
```
mchandschuh: This is definitely not a bad way of doing it. Another thought is firing off a task to force initialization of the lazy on a threadpool thread so it can be initialized by the time it's used. Now, this is likely not important, since often times we're invoking the interpreter _after_ a fatal failure, so performance becomes less of a concern is that situation (failing/about to fail).

Engine/Engine.cs
```diff
@@ -158,25 +151,25 @@ public void Run(AlgorithmNodePacket job, AlgorithmManager manager, string assemb
...
                         algorithm,
                         job,
-                        _algorithmHandlers.Results,
-                        _algorithmHandlers.MapFileProvider,
-                        _algorithmHandlers.FactorFileProvider,
-                        _algorithmHandlers.DataProvider,
+                        AlgorithmHandlers.Results,
+                        AlgorithmHandlers.MapFileProvider,
+                        AlgorithmHandlers.FactorFileProvider,
+                        AlgorithmHandlers.DataProvider,
```
mchandschuh: I like <3 -- Looks much cleaner with the PascalCasing

Engine/Results/BacktestingResultHandler.cs
```diff
@@ -42,81 +41,75 @@ public class BacktestingResultHandler : BaseResultsHandler, IResultHandler
...
-        /// Default initializer for
+        /// Creates a new instance
         /// </summary>
         public BacktestingResultHandler()
         {
+            Messages = new ConcurrentQueue<Packet>();
+            Charts = new ConcurrentDictionary<string, Chart>();
+            IsActive = true;
+            ResamplePeriod = TimeSpan.FromMinutes(4);
+            NotificationPeriod = TimeSpan.FromSeconds(2);
```
mchandschuh: I like! Move all initialization to the same place, makes it easier to grock initial state 👍

Tests/Brokerages/Paper/PaperBrokerageTests.cs
```diff
@@ -117,8 +116,8 @@ public void AppliesDividendsOnce()
             var brokerage = new PaperBrokerage(algorithm, job);
 
             // initialize results and transactions
-            results.Initialize(job, new EventMessagingHandler(), new Api.Api(), new BrokerageSetupHandler(), transactions);
-            results.SetAlgorithm(algorithm);
+            results.Initialize(job, new EventMessagingHandler(), new Api.Api(), transactions);
+            results.SetAlgorithm(algorithm, algorithm.Portfolio.TotalPortfolioValue);
```
mchandschuh: I like this, much better than exposing the entire `ISetupHandler` just to grab initial TPV 👍

PR #3521, opened 2019-08-20
Add support for custom data mapping v2

Martin-Molinero:
#### Description

**Changes on top of PR https://github.com/QuantConnect/Lean/pull/3338** - Thank you @gsalaz98
 
 Slightly different approach in the sense that each custom data type will define whether or not it should use map files:
- Moving `UsesMapFiles` to each custom data type (instead of a 3rd static entity)
- Updating regression test with sample custom data using map files,
which can run locally
- Adding unit tests for the `SubscriptionDataReaderHistoryProvider`,
checking it maps equities and options correctly (bug in `master` - good catch @gsalaz98!)
- Adding unit test which asserts `Algorithm.AddData()` throws if provided an invalid data type (bug in `master`)

**Note:** With the help of @gsalaz98 it was noted that `FillForward` enumerators will ignore new mappings while filling forward, since it will continue to use the `previous` emitted bar (with the old Symbol value) 

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3296
#### Motivation and Context

Allows using map files for custom data.

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/Quandl.cs
+++ b/Common/Data/Custom/Quandl.cs
@@ -142,15 +142,6 @@ namespace QuantConnect.Data.Custom
         }
 
         /// <summary>
-        /// Indicates if there is support for map files
-        /// </summary>
-        /// <returns>True indicates map files should be used</returns>
-        public override bool UsesMapFiles()
-        {
-            return true;
-        }
-
-        /// <summary>
         /// Set the auth code for the quandl set to the QuantConnect auth code.
         /// </summary>
         /// <param name="authCode"></param>
--- a/Common/Data/Custom/Quandl.cs
+++ b/Common/Data/Custom/Quandl.cs
@@ -142,6 +142,15 @@ namespace QuantConnect.Data.Custom
         }
 
         /// <summary>
+        /// Indicates if there is support for map files
+        /// </summary>
+        /// <returns>True indicates map files should be used</returns>
+        public override bool UsesMapFiles()
+        {
+            return true;
+        }
+
+        /// <summary>
         /// Set the auth code for the quandl set to the QuantConnect auth code.
         /// </summary>
         /// <param name="authCode"></param>
--- a/Common/Data/SubscriptionDataConfig.cs
+++ b/Common/Data/SubscriptionDataConfig.cs
@@ -19,6 +19,9 @@ using System.ComponentModel;
 using System.Runtime.CompilerServices;
 using NodaTime;
 using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Custom;
+using QuantConnect.Data.Custom.SEC;
+using QuantConnect.Data.Custom.Tiingo;
 using QuantConnect.Securities;
 using QuantConnect.Util;
 
@@ -114,6 +117,24 @@ namespace QuantConnect.Data
         }
 
         /// <summary>
+        /// Some custom data sources should support symbol mapping because the custom data
+        /// is directly related to symbols/tickers. For example, SEC data and sentiment data on US equities.
+        /// </summary>
+        public static HashSet<Type> MapFileTypes = new HashSet<Type>
+        {
+            typeof(TiingoDailyData),
+            typeof(Quandl),
+            typeof(SECReport8K),
+            typeof(SECReport10K),
+            typeof(SECReport10Q),
+        };
+        
+        /// <summary>
+        /// Indicates if there is support for map files.
+        /// </summary>
+        public bool UsesMapFiles { get; }
+
+        /// <summary>
         /// Gets the market / scope of the symbol
         /// </summary>
         public readonly string Market;
@@ -189,6 +210,7 @@ namespace QuantConnect.Data
             IsFilteredSubscription = isFilteredSubscription;
             Consolidators = new HashSet<IDataConsolidator>();
             DataNormalizationMode = dataNormalizationMode;
+            UsesMapFiles = SecurityType == SecurityType.Equity || SecurityType == SecurityType.Option || MapFileTypes.Contains(Type);
 
             TickType = tickType ?? LeanData.GetCommonTickTypeForCommonDataTypes(objectType, SecurityType);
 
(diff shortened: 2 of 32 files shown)


Common/Data/Custom/Quandl.cs
```diff
@@ -141,6 +141,15 @@ public override SubscriptionDataSource GetSource(SubscriptionDataConfig config,
             return new SubscriptionDataSource(source, SubscriptionTransportMedium.RemoteFile);
         }
 
+        /// <summary>
+        /// Indicates if there is support for map files
+        /// </summary>
+        /// <returns>True indicates map files should be used</returns>
+        public override bool UsesMapFiles()
+        {
+            return true;
```
mchandschuh: yes! use that polymorphism!

Common/Data/SubscriptionDataConfig.cs
```diff
@@ -116,24 +113,6 @@ public string MappedSymbol
             }
         }
 
-        /// <summary>
-        /// Some custom data sources should support symbol mapping because the custom data
-        /// is directly related to symbols/tickers. For example, SEC data and sentiment data on US equities.
-        /// </summary>
-        public static HashSet<Type> MapFileTypes = new HashSet<Type>
```
mchandschuh: 👍

Common/Extensions.cs
```diff
@@ -40,6 +42,36 @@ namespace QuantConnect
...
+        public static IBaseData GetBaseDataInstance(this Type type)
+        {
+            var objectActivator = ObjectActivator.GetActivator(type);
+            if (objectActivator == null)
+            {
+                throw new ArgumentException($"Data type \'{type.Name}\' missing parameterless constructor " +
+                    $"E.g. public {type.Name}() {{ }}");
+            }
+
+            var instance = objectActivator.Invoke(new object[] { type });
```
mchandschuh: If `objectActivator` is pointing towards a parameterless constructor, then why are we passing in the `type`?

Common/Extensions.cs
```diff
@@ -40,6 +42,36 @@ namespace QuantConnect
...
+            if(instance == null)
+            {
+                // shouldn't happen but just in case...
+                throw new ArgumentException($"Failed to create instance of type \'{type.Name}\'");
+            }
+
+            var result = instance as IBaseData;
+            if (result == null)
```
mchandschuh: This type check can be performed statically via `typeof(IBaseData).IsAssignableFrom(type)` before invoking the `objectActivator` to save a few cycles and fail early.

Common/Python/PythonData.cs
```diff
@@ -42,6 +43,13 @@ public PythonData()
         public PythonData(PyObject pythonData)
         {
             _pythonData = pythonData;
+            using (Py.GIL())
+            {
+                if (pythonData.HasAttr("UsesMapFiles"))
+                {
+                    _usesMapFiles = _pythonData.UsesMapFiles();
```
mchandschuh: very cool -- I assume we're only supporting `true`/`false` constants here. Just wondering if it _should_ be a method that gets passed `self` so it could check it's symbol or similar?

Common/Extensions.cs
Martin-Molinero: Good question 😄, truth is I just copied existing implementation, didn't change anything here. Don't think its worth changing in _this_ PR though.

- Tested the following and it fails for python custom data, so maybe it's being used by PythonNet somehow:
```
        public static Func<object> GetParameterlessActivator(Type dataType)
        {
            lock (_lock)
            {
                // if we already have it, just use it
                Func<object> factory;
                if (_parameterlessActivatorsByType.TryGetValue(dataType, out factory))
                {
                    return factory;
                }
                
                var ctor = dataType.GetConstructor(new Type[] { });

                //User has forgotten to include a parameterless constructor:
                if (ctor == null) return null;

                var newExp = Expression.New(ctor);
                var lambda = Expression.Lambda(typeof(Func<object>), newExp);
                factory = (Func<object>)lambda.Compile();

                // save it for later
                _parameterlessActivatorsByType.Add(dataType, factory);

                return factory;
            }
        }
```
PS: the original implementation is from 5 years ago Mike 😄

Common/Extensions.cs
Martin-Molinero: Good point - We expect 'instance' to be `IBaseData` in most cases so I went for 'as' versus 'IsAssignableFrom' since it is actually slightly cheaper - Check the following performance test:
```
        [Test]
        public void IsAssignableFromVsAsPerformanceTest()
        {
            Stopwatch stopwatch = new Stopwatch();
            object value = "This is a test";
            var type = value.GetType();
            stopwatch.Start();

            var result = IsAssignableFromTest(type);
            stopwatch.Stop();
            Console.WriteLine("IsAssignableFrom: {0} ms elapsed", stopwatch.ElapsedMilliseconds);
            Assert.IsTrue(result);

            stopwatch.Restart();
            result = AsTest(value);
            stopwatch.Stop();
            Assert.IsTrue(result);
            Console.WriteLine("AsOperator: {0} ms elapsed", stopwatch.ElapsedMilliseconds);
        }
        private bool IsAssignableFromTest(Type type, int loopCount = 1000000)
        {
            var isConvertible = false;
            for (int attempt = 0; attempt < loopCount; attempt++)
            {
                isConvertible |= typeof(IConvertible).IsAssignableFrom(type);
            }
            return isConvertible;
        }
        private bool AsTest(object value, int loopCount = 1000000)
        {
            var isConvertible = false;
            for (int attempt = 0; attempt < loopCount; attempt++)
            {
                isConvertible |= value as string != null;
            }
            return isConvertible;
        }
```

PS: I'll add a comment

Common/Python/PythonData.cs
Martin-Molinero: Exactly - It will be passed `self`, check out the custom data defined at `CustomDataUsingMapFileRegressionAlgorithm.py`

PR #3530, opened 2019-08-22
**

### LINKED ISSUE #3528: Lean cannot evaluate CashBook Total Value at start on "live-bitfinex"
#### Expected Behavior
```
2019-08-22T12:57:51.6908448Z TRACE:: BaseSetupHandler.SetupCurrencyConversions():
Symbol      Quantity    Conversion = Value in USD
USD: $          15.64 @       1.00 = $15.64
BTC: ฿           0.00 @   10040.00 = $8.25
LTC: Ł           3.02 @     74.091 = $224.01
ETH: Ξ           0.00 @     188.33 = $0.00
-------------------------------------------------
CashBook Total Value:                $247.89
```

#### Actual Behavior
```
2019-08-22T14:47:02.4470452Z TRACE:: BaseSetupHandler.SetupCurrencyConversions():
Symbol      Quantity    Conversion = Value in USD
USD: $          15.64 @       1.00 = $15.64
BTC: ฿           0.00 @       0.00 = $0.00
LTC: Ł           3.02 @       0.00 = $0.00
ETH: Ξ           0.00 @       0.00 = $0
-------------------------------------------------
CashBook Total Value:                $15.64
```

#### Potential Solution
will be submitted as separate PR

#### Checklist

- [X ] I have completely filled out this template
- [ X] I have confirmed that this issue exists on the current `master` branch
- [ X] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [ X] I have provided detailed steps to reproduce the issue

IlshatGaripov:
#### Description
The described issue occurs because historical queries that come from BaseSetupHandler.SetupCurrencyConversions() to BitfinexBrokerage.GetHistory() are of 1-resolution-period long when it come to while loop inside GetHistory() it does not get executed. For correct operation while-loop should be changed to do-while-loop. 

Moreover, now when the method works in such a way that allows processing requests with a length of 1 bar, such a scenario is possible when an empty string comes from the exchange to our request. This happens for example when the range between historical request's start and end dates is one minute and the period does not contain active trades. So to protect from possible errors to be thrown here:
```
startMsec = candles.Last().Timestamp + resolutionInMsec;
```
we may want to check if candles received from an exchenge are not empty.

Also, minor improvements related to local variable namings and included string variables to hold api keys in BitfinexBrokerageTests for convenient testing.

#### Related Issue
Closes #3528 

Have tested in real-time. results are enclosed in related issue descriptino

#### Types of changes

#### Checklist:


Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -284,31 +284,51 @@ public override IEnumerable<BaseData> GetHistory(Data.HistoryRequest request)
                 yield break;
             }
 
+            if (request.StartTimeUtc >= request.EndTimeUtc)
+            {
+                throw new ArgumentException("BitfinexBrokerage.GetHistory(): Indvalid history request parameters, " +
```
StefanoRaggi: Thanks for adding the missing check 👍 
Just a typo in the message text: `Indvalid`

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -284,31 +284,51 @@ public override IEnumerable<BaseData> GetHistory(Data.HistoryRequest request)
...
+                if (candles.Any())
+                {
+                    startMsec = candles.Last().Timestamp + resolutionInMsec;
+                }
+                else
+                {
+                    Log.Error($"BitfinexBrokerage.GetHistory(): Exchange returned no data for {symbol}" +
+                              $"on history request from {request.StartTimeUtc:s} to {request.EndTimeUtc:s}");
+                    break;
+                }
```
StefanoRaggi: Good catch!
Depending on the requested range an empty list can be a valid response, as noted in your comment.
If we want to log this condition, I think this message should not be logged as an error, maybe better use `Log.Trace` instead?

Tests/Brokerages/Bitfinex/BitfinexBrokerageTests.cs
```diff
@@ -52,8 +57,8 @@ protected override IBrokerage CreateBrokerage(IOrderProvider orderProvider, ISec
             return new BitfinexBrokerage(
                     Config.Get("bitfinex-url", "wss://api.bitfinex.com/ws"),
                     Config.Get("bitfinex-rest", "https://api.bitfinex.com"),
-                    Config.Get("bitfinex-api-key"),
-                    Config.Get("bitfinex-api-secret"),
+                    PublicKey,
+                    PrivateKey,
```
StefanoRaggi: Why have the credentials as `protected` members (hardcoded in the test class) instead of reading them from the `config.json` file in the tests `bin/Debug` folder?
I believe all other brokerage tests are reading credentials from config

IlshatGaripov:
Hello @StefanoRaggi . thank you for review. I have pushed commit above to address the requested changes. Additionally, with a closer look at BitfinexBrokerageHistoryProviderTests I could discover all test cases were inspired by how tests are done in:
https://github.com/QuantConnect/Lean/blob/master/Tests/Brokerages/GDAX/GDAXBrokerageHistoryProviderTests.cs
where the things are mostly checked for empty / no empty result returned from the method. So the same checking might be added I thought. as System.ArgumentException for which cases are checked is not happening always - just in case of invalid symbol supplied. THanks!

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -284,31 +284,53 @@ public override IEnumerable<BaseData> GetHistory(Data.HistoryRequest request)
...
+            long resolutionInMsec = (long)request.Resolution.ToTimeSpan().TotalMilliseconds;
             string symbol = _symbolMapper.GetBrokerageSymbol(request.Symbol);
-            long startMTS = (long)Time.DateTimeToUnixTimeStamp(request.StartTimeUtc) * 1000;
-            long endMTS = (long)Time.DateTimeToUnixTimeStamp(request.EndTimeUtc) * 1000;
+            long startMsec = (long)Time.DateTimeToUnixTimeStamp(request.StartTimeUtc) * 1000;
+            long endMsec = (long)Time.DateTimeToUnixTimeStamp(request.EndTimeUtc) * 1000;
             string endpoint = $"v2/candles/trade:{resolution}:t{symbol}/hist?limit=1000&sort=1";
+            var period = request.Resolution.ToTimeSpan();
 
-            while ((endMTS - startMTS) > resolutionInMS)
```
mchandschuh: Would simply updating this to `>=` also permit this case to work?

Brokerages/Bitfinex/BitfinexBrokerage.cs
mchandschuh: Since this would be responding to a user's request for history, I think it makes sense to raise a `BrokerageMessageEvent` so the algorithm can handle it.

Brokerages/Bitfinex/BitfinexBrokerage.cs
IlshatGaripov: @mchandschuh ```>=``` would also work just fine. 

I now remember that when I was first editing a file I was also substracting a millisecond from end time :
```long endMsec = (long)Time.DateTimeToUnixTimeStamp(request.EndTimeUtc) * 1000 - 1; ```
and therefore do-while was then necessary. 

A reason for this is that Bitfinex makes trade bar's open time a timestamp. And if requesting data for the interval where the beginning and end are rounded down - as Lean I believe is doing all over the place - it will give us also return bar that starts at end time. Example. Lets say we requst data from 13:05:00 to 13:07:00 --> this will return three bars: 
13:05:00 to 13:06:00
13:06:00 to 13:07:00
13:07:00 to 13:08:00

I thought it does no harm an extra bar but if to understand the end time as an upper cap - no data is to be received after - we might want to make a correction for this and subtract a millisecond... ?

mchandschuh: RE: _example_:

I think that's a bug if we're returning data ahead of the end time. LEAN should _always_ be using end times for determining time synchronization issues, such as determining what bars fall within a requested time span. An even more egregious example would be when using daily data. If we request from 2019.09.17 to 2019.09.18, I would expect to get a single daily bar for the date of the 17th and I would **not** expect to also receive the bar for the 18th.

RE: _subtract a millisecond_:
Individual implementations can do what they need to do in order to fulfill their responsibilities. Subtracting a millisecond _may_ work, but another way to look at it is to subtract the resolution for bitfinex. If bitfinex uses start times than we can project the user's request to reflect that. For example, LEAN _should_ be using (start, end] semantics, meaning, we include all day from start to end _including_ the start and _excluding_ the end. If bitfinex's api looks at it as (start, end-resolution], then perhaps we just modify the end time of the request to shift it to a bar start time. I'm always a little hesitant to subtract a millisecond here or a tick there as it appears to be more of a bandaid than a projection of data to fit the api's conventions.

IlshatGaripov: @mchandschuh agree that it's reasonble to drop the last bar. please kindly review the commit below that serves to solve an issue. things have been changed here:

1) Additional warning was added to ensure end time is a full round of history request resolution. Idea is that bitfinex API may provide only historical bars and to avoid any ambiguity the user must be precise about timing borders. ?
2) Additional Logic to drop the last candle that start at end TimeKeeper. Solution that would modify end time by substracting the resolution amount of milliseconds will not work for the history request of 1-bar-long, where the start and time time will result in to be the same - exchange will return an error.
3) & 4) To test the things work as expected Bitfinex History provider tests have been little modified. To support the rounding agreement. And to provide the valid test case of 1 bar long historical request:
```new TestCaseData(Symbol, Resolution.Minute, Time.OneMinute, false, false),```
to ensuse this does return the data

Issue #3024, opened 2019-03-24
Enable calling Alpha Framework on schedule basis, and not only on tick data


erikbengtson:
#### Expected Behavior
1.End of day event should be called and risk management model triggered. Rational: avoid low liquidity periods, and high spread costs, avoid overnight interest costs
2.RiskManagementModel should be called on endofday
#### Actual Behavior
1.standard OnEndOfDay Event is scheduled on a fixed time market close - 10 minutes
2.RiskManagementModel is called on data resolution, but not triggered on scheduled events

#### Potential Solution
1.Allow configuration of scheduled periods.
2.add a new OnScheduledEvent On risk management model class. We should be able to schedule events and this method should be called

#### Checklist
- [X] I have completely filled out this template
- [X] I have confirmed that this issue exists on the current `master` branch
- [X] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

This is a feature request, since alpha framework methods cannot be run on schedule

erikbengtson:
workaround for point 2. Add to the Risk Model implementation this condition, and set target there

`                    
var marketClose = security.Exchange.Hours.GetNextMarketClose(algorithm.Time, security.IsExtendedMarketHours);

// define the time of day we want the event to fire before marketclose
var eventTime = marketClose.Subtract(ScheduledEvent.SecurityEndOfDayDelta);

// convert the event time into UTC
var eventUtcTime = eventTime.ConvertToUtc(security.Exchange.TimeZone);

if (algorithm.Time > eventUtcTime)
{
//set target here
}
`

michaelsrew:
https://www.quantconnect.com/docs/algorithm-reference/scheduled-events


BeforeMarketClose(symbol, min) | self.TimeRules.BeforeMarketClose("SPY", 10) 					Trigger an event a few minutes before close for a specific symbol.

does not help?

erikbengtson:
Yes it certainly helps for item 1.

However, the alpha framework, risk management model, is not aware of scheduled events.

The issue is that risk management model, is called on every tick and not on scheduled event.

As workaround, I keep a dictionary of events by symbol on the Algorithm class that is updated on schedule, and on risk management model, on tick basis, I check if the flag exists for that symbol.

jaredbroad:
[...]

PR #3534, opened 2019-08-23
Feature 3533 improve extensibility framework fillmodel

### LINKED ISSUE #3533: Improve extensibility for algorithm framework and fill model
#### Expected Behavior

Certain methods in QCAlgorithm and FillModel classes are useful for fine-tuning behavior in sub-classes:

FillModel.GetPricesCheckingPythonWrapper
FillModel.IsExchangeOpen
QCAlgorithm.Framework.OnInsightsGenerated
QCAlgorithm.Framework.InitializeInsightFields
QCAlgorithm.Framework.ProcessInsights

#### Actual Behavior

Methods are private.

#### Potential Solution

Make methods accessible in sub-classes.

#### Checklist

olegsheyner:
Make some methods accessible to sub-classes

#### Related Issue

Closes #3533 

#### Motivation and Context

Improve ability to fine-tune fill models and algorithm framework flow in sub-classes.

Regression tests pass.

#### Types of changes

#### Checklist:


Martin-Molinero:
Hi @olegsheyner,
Could you please provide a detailed use case/requirement for each of these changes, so we assert, evaluate and document them - The one I can understand is `FillModel.IsExchangeOpen`.
`FillModel.GetPricesCheckingPythonWrapper` was added for C# base methods used by a python custom `FillModel`. `Algorithm.ProcessInsights()` is the core framework pipeline structure. `Algorithm.OnInsightsGenerated()` is also part of the insight pipeline. I believe `Algorithm.InitializeInsightFields()` is private to guarantee the prices and times of an insight are set right.
Thanks!

olegsheyner:
Hi @Martin-Molinero,

Regarding `FillModel': looking to tweak certain aspects of `FillModel`, while leaving the rest of the behavior unchanged.  As a specific example, base implementation of `MarketOnOpenFill` references `GetPricesCheckingPythonWrapper` - would be ideal for the subtype implementation to continue getting prices in the same manner.

Regarding `Algorithm`, the use case is as follows.  External events, not tied to or easily represented as time series data, generate insights.  It is useful to initiate insight generation/processing cycle as triggered by outside events, but the standard framework algorithm runs only universe selection and skips insight generation if the data slice has no data.

The use case could be satisfied by extracting the insight generation/processing cycle code at the end of OnFrameworkData into a separate method and making this method available in sub-classes:

```
            var insightsEnumerable = Alpha.Update(this, slice);
            // for performance only call 'ToArray' if not empty enumerable (which is static)
            var insights = insightsEnumerable == Enumerable.Empty<Insight>()
                ? new Insight[] { } : insightsEnumerable.ToArray();

            // only fire insights generated event if we actually have insights
            if (insights.Length != 0)
            {
                OnInsightsGenerated(insights.Select(InitializeInsightFields));
            }

            ProcessInsights(insights);
```

Martin-Molinero:
Hi @olegsheyner,

`FillModel`: If the subtype is implemented in C# it does not need access to `GetPricesCheckingPythonWrapper` since it's a work around just for Python based derivates. If the derivate is a Python implementation it should be able to directly call the base `GetPrices()` (not the `GetPricesCheckingPythonWrapper`). The use case for this method is a Python derivate implementing `GetPrices()` called by a base C# implementation, this unit test https://github.com/QuantConnect/Lean/blob/master/Tests/Common/Orders/Fills/BackwardsCompatibilityFillModelsTests.cs#L447. Let me know if it makes sense.

`Algorithm`: Interesting, I think that I understand, events like `ScheduledEvent` feature? This seems like a new missing feature of Lean, similar to the one requested at https://github.com/QuantConnect/Lean/issues/3024. Although I'm not sure about exposing the framework pipeline, and letting the user derivate algorithm implement it them selfs instead.

mchandschuh:
Would the `EmitInsights` methods work for your use case? Here's an [example algorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/ConvertToFrameworkAlgorithm.cs#L76) that uses it. The `EmitInsights` function allows your algorithm to push insights through the framework infrastructure as if it was generated directly from an alpha model.

olegsheyner:
@mchandschuh, thank you - `EmitInsights` method works for my use case, no further modifications necessary to `Algorithm`

@Martin-Molinero: understood re: `FillModel.GetPricesCheckingPythonWrapper`.  So as a summary, `FillModel.IsExchangeOpen` is the only one needed.  I will amend PR accordingly.

PR #3544, opened 2019-08-27
Replace WebSocketSharp library with System.Net.WebSockets

StefanoRaggi:
#### Description
- The `WebSocketSharp` library reference has been removed and replaced with the usage of the standard `System.Net.WebSocket` classes in the LEAN `WebSocketWrapper` class.
- The `IWebSocket` interface now inherits from `IDisposable`.
- The GDAX Brokerage unit tests, which require a real account and for this reason are excluded from Travis build, have been updated and fixed -- they were failing on current master.

#### Related Issue
n/a

#### Motivation and Context
- Increase the stability of the crypto live data feeds

No.

- Local testing with crypto brokerages
- Brokerage unit tests
- QC back-end software

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/WebSocketWrapper.cs
+++ b/Brokerages/WebSocketWrapper.cs
@@ -14,19 +14,25 @@
 */
 
 using System;
-using QuantConnect.Configuration;
+using System.IO;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 using QuantConnect.Logging;
-using WebSocketSharp;
+using QuantConnect.Util;
 
 namespace QuantConnect.Brokerages
 {
     /// <summary>
-    /// Wrapper for WebSocketSharp to enhance testability
+    /// Wrapper for ClientWebSocket to enhance testability
     /// </summary>
     public class WebSocketWrapper : IWebSocket
     {
-        private WebSocket _wrapped;
+        private ClientWebSocket _wrapped;
         private string _url;
+        private readonly CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+        private ArraySegment<byte> _receiveBuffer = new ArraySegment<byte>(new byte[1024 * 8]);
 
         /// <summary>
         /// Wraps constructor
@@ -40,21 +46,7 @@ namespace QuantConnect.Brokerages
             }
 
             _url = url;
-            _wrapped = new WebSocket(url)
-            {
-                Log =
-                {
-                    Level = Config.GetBool("websocket-log-trace") ? LogLevel.Trace : LogLevel.Error,
-
-                    // The stack frame number of 3 was derived from the usage of the Logger class in the WebSocketSharp library
-                    Output = (data, file) => { Log.Trace($"{WhoCalledMe.GetMethodName(3)}(): {data.Message}", true); }
-                }
-            };
-
-            _wrapped.OnOpen += (sender, args) => OnOpen();
-            _wrapped.OnMessage += (sender, args) => OnMessage(new WebSocketMessage(args.Data));
-            _wrapped.OnError += (sender, args) => OnError(new WebSocketError(args.Message, args.Exception));
-            _wrapped.OnClose += (sender, args) => OnClose(new WebSocketCloseData(args.Code, args.Reason, args.WasClean));
+            _wrapped = new ClientWebSocket();
         }
 
         /// <summary>
@@ -63,7 +55,23 @@ namespace QuantConnect.Brokerages
         /// <param name="data"></param>
         public void Send(string data)
         {
-            _wrapped.Send(data);
+            try
+            {
+                if (!IsOpen)
+                {
+                    Log.Trace($"WebSocketWrapper.Send(): Connection not open (IsOpen:{IsOpen}, State:{_wrapped.State}): attempting connection", true);
+
+                    Connect();
+                }
+
+                var sendBuffer = new ArraySegment<byte>(Encoding.UTF8.GetBytes(data));
+
+                _wrapped.SendAsync(sendBuffer, WebSocketMessageType.Text, true, _cancellationTokenSource.Token).SynchronouslyAwaitTask();
+            }
+            catch (Exception exception)
+            {
+                OnError(new WebSocketError(exception.Message, exception));
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+                    {
+                        stream.Seek(0, SeekOrigin.Begin);
+                        using (var reader = new StreamReader(stream, Encoding.UTF8))
+                        {
+                            OnMessage(new WebSocketMessage(reader.ReadToEnd()));
+                        }
+                    }
+                    else if (webSocketReceiveResult.MessageType == WebSocketMessageType.Close)
+                    {
+                        if (webSocketReceiveResult.CloseStatus != null)
```
Martin-Molinero: Maybe we would want to send the `OnClose` event even if we don't have access to the `CloseStatus`?

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+
+                    if (webSocketReceiveResult.MessageType == WebSocketMessageType.Text)
+                    {
+                        stream.Seek(0, SeekOrigin.Begin);
+                        using (var reader = new StreamReader(stream, Encoding.UTF8))
+                        {
+                            OnMessage(new WebSocketMessage(reader.ReadToEnd()));
+                        }
+                    }
+                    else if (webSocketReceiveResult.MessageType == WebSocketMessageType.Close)
```
Martin-Molinero: Is it worth logging if we receive another message type, since we won't be handling it?

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+            catch (Exception exception)
             {
-                _wrapped.Connect();
+                OnError(new WebSocketError(exception.Message, exception));
+            }
+        }
+
+        private void ResetAndConnect()
+        {
+            _wrapped = new ClientWebSocket();
```
Martin-Molinero: Maybe we should be calling `DisposeSafely` on the old `_wrapped` 🤔  ?

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+
+            try
+            {
+                _wrapped.ConnectAsync(new Uri(_url), _cancellationTokenSource.Token).SynchronouslyAwaitTask();
+
+                if (IsOpen)
+                {
+                    OnOpen();
+
+                    new TaskFactory().StartNew(
```
Martin-Molinero: I think there is some room for a race condition around this, eg multiple concurrent calls to `Connect` triggering multiple tasks - We might need some lock and blocking of calling threads.
Documentation at https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket.receiveasync?view=netframework-4.8 states `Exactly one send and one receive is supported on each ClientWebSocket object in parallel.`

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+                if (!IsOpen)
+                {
+                    Log.Trace($"WebSocketWrapper.Send(): Connection not open (IsOpen:{IsOpen}, State:{_wrapped.State}): attempting connection", true);
+
+                    Connect();
+                }
+
+                var sendBuffer = new ArraySegment<byte>(Encoding.UTF8.GetBytes(data));
+
+                _wrapped.SendAsync(sendBuffer, WebSocketMessageType.Text, true, _cancellationTokenSource.Token).SynchronouslyAwaitTask();
```
Martin-Molinero: I think we might need some locks around `Send()`, documentation at https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocket.sendasync?view=netframework-4.8 states `Exactly one send and one receive is supported on each ClientWebSocket object in parallel.` ?

Brokerages/WebSocketWrapper.cs
```diff
@@ -82,18 +167,27 @@ public void Connect()
         /// </summary>
         public void Close()
         {
-            _wrapped.Close();
+            try
+            {
+                _wrapped.CloseAsync(WebSocketCloseStatus.NormalClosure, string.Empty, _cancellationTokenSource.Token);
```
Martin-Molinero: After closure could we continue to operate after a `Connect()`, is it a valid scenario? If so, `_cancellationTokenSource` should be reset?

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+                            try
+                            {
+                                await StartReceiving();
+                            }
+                            catch (Exception exception)
+                            {
+                                OnError(new WebSocketError(exception.Message, exception));
+                            }
+
+                            if (_wrapped.State == WebSocketState.Aborted)
```
Martin-Molinero: `StartReceiving()->ReceiveAsync()` could throw if for some reason we are not connected, in this case, we will stop receiving until someone calls `Send()` or `Open()`, should we try to reconnect ourselves instead?

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+            }
         }
 
         /// <summary>
         /// Wraps Connect method
         /// </summary>
         public void Connect()
         {
-            if (!IsOpen)
+            if (IsOpen)
```
mchandschuh: Likely need locks around these `IsOpen` checks

Brokerages/WebSocketWrapper.cs
mchandschuh: or even worse, the server can forcefully close the socket w/out sending a close message and you won't know until you call `ReceiveAsync()`

mchandschuh: _also_ -- reco using `.ConfigureAwait(false)` since some of these components may be used by applications with a message/event loop (anything w/ a UI)

Brokerages/WebSocketWrapper.cs
```diff
@@ -63,17 +55,110 @@ public void Initialize(string url)
...
+                                OnError(new WebSocketError(exception.Message, exception));
+                            }
+
+                            if (_wrapped.State == WebSocketState.Aborted)
+                            {
+                                // need to recreate ClientWebSocket instance, cannot be reused
+
+                                Log.Trace($"WebSocketWrapper.Connect(): Connection aborted (IsOpen:{IsOpen}, State:{_wrapped.State}): creating new ClientWebSocket instance", true);
+
+                                await new TaskFactory().StartNew(ResetAndConnect, _cancellationTokenSource.Token);
```
mchandschuh: why did we choose to use `new TaskFactory()` instead of `TaskFactory.StartNew` or simply `Task.Run` ?

>https://devblogs.microsoft.com/pfxteam/task-run-vs-task-factory-startnew/

Even stranger though, `ResetAndConnect` is a synchronous method but we're wrapping it in a task _and_ awaiting it.... why?

Brokerages/WebSocketWrapper.cs
mchandschuh: and synchronization/locks required here

Brokerages/WebSocketWrapper.cs
```diff
@@ -82,18 +167,27 @@ public void Connect()
...
+                OnError(new WebSocketError(exception.Message, exception));
+            }
         }
 
         /// <summary>
-        /// Wraps IsAlive
+        /// Returns true if the WebSocketState is Open
         /// </summary>
-        public bool IsOpen => _wrapped.IsAlive;
+        public bool IsOpen => _wrapped.State == WebSocketState.Open;
```
mchandschuh: `_wrapped == null` before `Initialize` is called, perhaps use the null navigation operation:  `?.`

Brokerages/WebSocketWrapper.cs
mchandschuh: The other type would be [binary](https://msdn.microsoft.com/en-us/vstudio/system.net.websockets.websocketmessagetype(v=vs.90)) and I agree we should minimally log something but honestly, I think we should probably throw. If the server is trying to send us data and we're not able to read it, that means we're now in an inconsistent state. I think that's a fatal error.

StefanoRaggi:
Closing in favor of #3644

PR #3555, opened 2019-09-03
Extends functionality of ToolBox to set the program input arguments 

IlshatGaripov:
Extends functionality of ToolBox to set the program input arguments via config.json file.

#### Description
This commit allows the toolbox users to set all program input arguments via config.json file, as well as through the command line as usually. The new feature is fully compatible with the previous functionality and does only extends the existing model, with super minor changes to any existing code. 

#### Related Issue
n/a

#### Motivation and Context
I was thinking to start working on new downloader, I thought the ability to set parameters through the config could facilitate the work 

yes 
https://github.com/QuantConnect/Lean/blob/master/ToolBox/README.md

I ran one example, the config parameters from this experiment you can see remained in config file.  It worked well. feedback and advise from experienced users on kind of tests I can supply additionally to cover new functionality would be very appreciated. 

#### Types of changes

#### Checklist:

- [X ] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -1514,5 +1514,13 @@ namespace QuantConnect
                 return s;
             }
         }
+
+        /// <summary>Indicates whether the specified string array is null or has a length of zero.</summary>
+        /// <param name="array">The array to test.</param>
+        /// <returns>true if the array parameter is null or has a length of zero, false otherwise.</returns>
+        public static bool IsNullOrEmpty(this string[] array)
+        {
+            return (array == null || array.Length == 0);
+        }
     }
 }
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -1514,13 +1514,5 @@ namespace QuantConnect
                 return s;
             }
         }
-
-        /// <summary>Indicates whether the specified string array is null or has a length of zero.</summary>
-        /// <param name="array">The array to test.</param>
-        /// <returns>true if the array parameter is null or has a length of zero, false otherwise.</returns>
-        public static bool IsNullOrEmpty(this string[] array)
-        {
-            return (array == null || array.Length == 0);
-        }
     }
 }
--- a/ToolBox/Program.cs
+++ b/ToolBox/Program.cs
@@ -17,8 +17,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Reflection;
-using Newtonsoft.Json;
 using QuantConnect.Configuration;
 using QuantConnect.ToolBox.AlgoSeekFuturesConverter;
 using QuantConnect.ToolBox.AlgoSeekOptionsConverter;
@@ -58,10 +56,10 @@ namespace QuantConnect.ToolBox
         {
             var optionsObject = ToolboxArgumentParser.ParseArguments(args);
 
-            // if no console args supplied check if config.json has working info
+            // if no console args supplied check configuration file for working info
             if (optionsObject.Count == 0)
             {
-                optionsObject = LoadOptionsObjectFromConfigurationFile("config.json");
+                optionsObject = ApplicationParser.ParseConfigurationFile("config.json");
             }
 
             var targetApp = GetParameterOrExit(optionsObject, "app").ToLower();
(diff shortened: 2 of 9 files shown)


ToolBox/Program.cs
```diff
@@ -55,12 +57,15 @@ public class Program
         public static void Main(string[] args)
         {
             var optionsObject = ToolboxArgumentParser.ParseArguments(args);
+
+            // if no console args supplied check if config.json has working info
             if (optionsObject.Count == 0)
             {
-                PrintMessageAndExit();
+                optionsObject = LoadOptionsObjectFromConfigurationFile("config.json");
```
mchandschuh: We'll want to make sure we update the documentation to be explicit about this behavior. Upon first reading of the PR I was thinking that `config.json` would be used to resolve any values not specified, which would permit a mixed-mode, some via cmd line and some via config. It looks like the way it's set up is all or nothing.

ToolBox/Program.cs
```diff
@@ -302,5 +307,71 @@ private static string GetParameterOrDefault(IReadOnlyDictionary<string, object>
 
             return value.ToString();
         }
+
+        /// <summary>
+        /// Loads values from JSON text file
+        /// </summary>
+        private static Dictionary<string, object> LoadOptionsObjectFromConfigurationFile(string path)
```
mchandschuh: Perhaps this logic makes more sense being hidden within the `ParseArguments` method instead of living in `main` ?

ToolBox/Program.cs
```diff
@@ -302,5 +307,71 @@ private static string GetParameterOrDefault(IReadOnlyDictionary<string, object>
...
+        {
+            var optionsObject = new Dictionary<string, object>();
+            try
+            {
+                // Read the text from json and fill that in a specifal object that
+                // will hold the values corresponding to json file contents
+                var toolBoxOptionsConfig = JsonConvert.DeserializeObject<ToolBoxOptionsConfiguration>(File.ReadAllText(path));
```
mchandschuh: Why not use `JObject` here? If the goal is to simply get a `Dictionary<string, object>` then the `ToolBoxOptionsConfiguration` class isn't needed. I think the following reduces this entire method to a couple lines:

``` csharp
var contents = File.ReadAllText(path);
var jObject = JObject.Parse(contents);
var optionsObject = jObject.ToObject<Dictionary<string, object>>();
```

ToolBox/Program.cs
```diff
@@ -302,5 +307,71 @@ private static string GetParameterOrDefault(IReadOnlyDictionary<string, object>
...
+                    {
+                        // if we reached here we are most probably working with 'tickers' array
+                        var valueAsStringArray = (string[])prop.GetValue(toolBoxOptionsConfig);
+                        if (valueAsStringArray.IsNullOrEmpty())
+                            continue;
+
+                        var key = GetThePropertyNameFromJsonAttribute(prop);
+
+                        // I fully copy here a piece of code from ApplicationParser.cs that has to do with
+                        // case CommandOptionType.MultipleValue - for the full compatibility!
```
mchandschuh: The fact that we're _copying_ code from `ApplicationParser` is another reason why this logic belong _inside_ `ApplicationParser`... that and it's parsing options for the application :)

ToolBox/config.json
```diff
@@ -0,0 +1,122 @@
...
+
+  //REQUIRED ALL downloaders (except QBDL, SECDL)
+  //Example: --tickers=SPY,AAPL,etc"
+  "tickers": [
+    "ETHUSD",
+    "BTCUSD"
+
+  ],
+
+  //REQUIRED ALL downloaders (except QBDL, CDL, SECDL) and IVolatilityEquityConverter, QuantQuoteConverter
```
mchandschuh: Awesome! Thanks for adding a template `config.json` file.

ToolBox/QuantConnect.ToolBox.csproj
```diff
@@ -368,6 +369,9 @@
     <None Include="CoarseUniverseGenerator\config.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="config.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
mchandschuh: Generally you want this set to `PreserveNewest` so if you have a `config.json` in your `bin/Debug` directory that you've been editing, a new build won't overwrite it w/out warning

Common/Extensions.cs
```diff
@@ -1514,5 +1514,13 @@ public static string RemoveFromEnd(this string s, string ending)
                 return s;
             }
         }
+
+        /// <summary>Indicates whether the specified string array is null or has a length of zero.</summary>
+        /// <param name="array">The array to test.</param>
+        /// <returns>true if the array parameter is null or has a length of zero, false otherwise.</returns>
+        public static bool IsNullOrEmpty(this string[] array)
```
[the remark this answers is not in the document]
IlshatGaripov: @mchandschuh thanks for another great feedback. I ended up removing extension method and build the logic into the method directly. As I moved the parsing logic into the ApplicationParser which resides in Configuration projects from where I believe I can have no access for the extension methods from Common project.

ToolBox/Program.cs
IlshatGaripov: That's correct, for those who prefer to do all the settings through config file - they can use only config. and launch from visual studio . Passing no console argument . For those who choose to use a console things will work as before.

ToolBox/Program.cs
IlshatGaripov: has been moved to ApplicationParser

ToolBox/Program.cs
IlshatGaripov: I tried your suggestion - thanks ! - it appears that no matter what values are specified in config file Json.NET will deserialize it into a Dictionary where all the properties from file will be contained as distionary key elements. At the same time, the program expects that dictionary contains keys-values only for those properties that have representative information - value is neither null nor empty. Supporting the own parser does let to achieve this. Consider this line in Program.cs:
https://github.com/QuantConnect/Lean/blob/master/ToolBox/Program.cs#L72
```
var toDate = optionsObject.ContainsKey("to-date")
                    ? DateTime.ParseExact(optionsObject["to-date"].ToString(), "yyyyMMdd-HH:mm:ss", CultureInfo.InvariantCulture)
                    : DateTime.UtcNow;
```
If we used the regular deserializer our dictionary would contain "to-date" and this line would break down the exectuion.

ToolBox/QuantConnect.ToolBox.csproj
IlshatGaripov: Thank you, I am changing this with separate commit - as well as changing the name of toolbox config file to 'toolboxoptions.json' (isn't it very complicated name?) As some downloaders may have used Lean settings config.json file and keep it in bin/Debug:
https://github.com/QuantConnect/Lean/blob/master/ToolBox/YahooDownloader/YahooDownloaderProgram.cs#L44
to avoid the conflict in namings.

ToolBox/Program.cs
mchandschuh: Great points 👍

PR #3567, opened 2019-09-09
cLOSE IT

### LINKED ISSUE #3435: Factor Files unit test failing
#### Expected Behavior

- No unit test fails on master
#### Actual Behavior

- `FactorFiles_CanBeGenerated_Accurately` unit test is failing on `master` at https://github.com/QuantConnect/Lean/blob/master/Tests/Common/Util/FactorFileGeneratorTests.cs#L70
#### Potential Solution

- fix unit test
#### Reproducing the Problem

- Run `FactorFiles_CanBeGenerated_Accurately` unit test
#### System Information

N/A
#### Checklist

IlshatGaripov:


Tests/ToolBox/FactorFileGeneratorTests.cs
```diff
@@ -33,7 +33,7 @@ public class FactorFileGeneratorTests
         private const string PermTick = "AAPL";
```
gsalaz98: (2 lines above) - Tests are still under the "TravisExclude" category.

mchandschuh:
@IlshatGaripov -- do you have a case where the generator was producing _incorrect_ factors? Is the implication here that the generator has **always** been generating the wrong factor files?

Is the test wrong or is the implementation wrong? Have we tested these changes against some set of historical data and compared it to _known_ splits/dividends and possibly producing an adjusted time series from raw data and matching that to yahoo's adjusted time series data. Yahoo is nice in that it's historical data provides the raw and adjusted closing prices each day as well as the splits and dividends.

Tests/ToolBox/FactorFileGeneratorTests.cs
```diff
@@ -33,7 +33,7 @@ public class FactorFileGeneratorTests
...
         private const string Market = "usa";
         readonly Symbol _symbol = new Symbol(SecurityIdentifier.GenerateEquity(PermTick, Market), PermTick);
-        private readonly string _dataPath = LeanData.GenerateZipFilePath(Config.Get("data-folder"),
+        private readonly string _dataPath = LeanData.GenerateZipFilePath(Config.Get("data-folder", "../../../Data"),
```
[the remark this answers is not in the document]
mchandschuh: I would go even further to say it's above a nit -- it's a duplication of a definition :)

IlshatGaripov:
@mchandschuh I think that FactorFiles_CanBeGenerated_Accurately() test does exactly that. Here is the full logic:

  We have the inner Lean algorithm that transforms factor files records during backtest in splits and dividents, which are fed into the algorithm via OnData. We consider factor file correct if when transformed by Lean algorithm it gives correct splits/dividend output for any requested historical span. There can be many files that, when transformed, give the correct result. But their values should differ only very slightly, I think this could be mathematically formulated correctly and proved, but we take it for granted. In experiment above I tested existing file by the algorithm to prove the results of transformation - splits/dividends - are correct. That is, we can assume the file itself correct and use it as benchmark.
 
What does the test do?. It does exactly what you have conveyed. It takes  raw data from YahooDownloader (I verified it's correct) then makes factor files out of them using the generator, stores resulting FactorFile in memory. Then successively iterates through time and compares the entries in temporary file against the entries of test file that we consider a benchmark. And makes sure that difference between each and every entry does not exceed the ```var tolerance = 0.00001m;```

In this way, if test does work fine, not fail, means that the generator produces a factor file whose values are only slightly different from values of a file that gives correct split/dividend values during backtest execution. Thus, generator does produce valid files.

Common/Globals.cs
```diff
@@ -39,7 +39,7 @@ static Globals()
         /// </summary>
         public static void Reset ()
         {
-            DataFolder = Config.Get("data-folder", Config.Get("data-directory", @"../../../Data/"));
+            DataFolder = Config.Get("data-folder", @"../../../Data/");
```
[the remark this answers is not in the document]
IlshatGaripov: @gsalaz98 reverting and will upload as separate comit in a minute. thanks! 
@mchandschuh agree with a statement of a issue per PR sorry has been reverted . I think the Dividend algo's behavious worth of separate discussion - will open a separate thread. thank you!

ToolBox/FactorFileGenerator.cs
```diff
@@ -226,11 +226,16 @@ private FactorFileRow CalculateNextDividendFactor(BaseData dividend, FactorFileR
...
+            var splitFactorMultiplier = Math.Round(1 / previousFactorFileRow.SplitFactor);
+
+            // adjust the dividend for both price and split factors (!)
+            var priceFactor = previousFactorFileRow.PriceFactor - splitFactorMultiplier * dividend.Value *
+                              previousFactorFileRow.PriceFactor / previousClosingPrice.Close;
 
             return new FactorFileRow(
                 previousClosingPrice.Time,
-                priceFactor.RoundToSignificantDigits(7),
+                Math.Round(priceFactor, 7),
```
mchandschuh: Curious as to the reasoning behind this change?

Tests/Common/Util/FactorFileGeneratorTests.cs
```diff
@@ -80,6 +79,16 @@ public void FactorFiles_CanBeGenerated_Accurately()
 
             var originalFactorFileInstance = FactorFile.Read(PermTick, Market);
 
+            // start date is the day of the first dividend/split entry in the table
+            var start = originalFactorFileInstance.SortedFactorFileData.Keys.Skip(1).First();
```
mchandschuh: This is a great change. Instead of picking some arbitrary date in the past (unix epoch), we pick the first date as provided in the reference files. This is OK since the test already depends on the reference files and we wouldn't want to produce factor before or after (_especially_ after) what we are able to compare against.

ToolBox/FactorFileGenerator.cs
```diff
@@ -257,7 +262,7 @@ private FactorFileRow CalculateNextSplitFactor(BaseData split, FactorFileRow pre
             return new FactorFileRow(
                     previousClosingPrice.Time,
                     previousFactorFileRow.PriceFactor,
-                    (previousFactorFileRow.SplitFactor * split.Value).RoundToSignificantDigits(6),
+                    Math.Round(previousFactorFileRow.SplitFactor / split.Value, 6),
```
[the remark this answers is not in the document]
IlshatGaripov: @mchandschuh thank you for reasonable comments, I do make changes in accordance with all suggestions. Please see a commit below. 

RE: _Since we're in here fixing this split factor bug, we might as well update the FactorFileRow.ToCsv to properly serialize factors in the precision provided by our reference files_

I am fixing the ToCsv() method. Do we need any rounding here at all ? I believe all decimals here come in already rounded to the needed extent by FactorFileGenerator?

ToolBox/FactorFileGenerator.cs
```diff
@@ -226,11 +226,16 @@ private FactorFileRow CalculateNextDividendFactor(BaseData dividend, FactorFileR
 
             TradeBar previousClosingPrice = FindPreviousTradableDayClosingPrice(eventDayData.Time);
 
-            var priceFactor = previousFactorFileRow.PriceFactor - (dividend.Value / ((previousClosingPrice.Close) * previousFactorFileRow.SplitFactor));
+            // round split factor back to natural number
+            var splitFactorMultiplier = Math.Round(1 / previousFactorFileRow.SplitFactor);
```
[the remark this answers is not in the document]
IlshatGaripov: this has been reverted. I thought that splits sould only be happening in the direction of multiple shares increase? how would otherwise to convert the prime number of shares when uncommon splits occur? Suppose one holds 7 shares how they are being converted on 2 to 3 split? but I am sure that there must be a mechanism to do it..

ToolBox/FactorFileGenerator.cs
IlshatGaripov: I was under impression that reference files were generated using Math.Round + FCL method is less error prone. I do realize now - thanks for detecting it - this was ```FactorFileRow.ToCsv()``` that did the roundings. Using RoundToSignificantDigits does make sense here indeed; in particular in roundings of SplitFactors that can have one or few zeros between the decimal point and first significant number ```0.0357143```

ToolBox/FactorFileGenerator.cs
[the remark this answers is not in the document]
mchandschuh: Great question. In the exotic cases, partial-share shareholders are typically paid out in cash at the previous day's closing market rate.

ToolBox/FactorFileGenerator.cs
mchandschuh: That was my thinking as well, so we don't lose precision as we go back further in time on securities that have experiences many large splits

Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -296,7 +296,10 @@ public static FactorFileRow Parse(string line)
         public string ToCsv(string source = null)
         {
             source = source == null ? "" : $",{source}";
-            return $"{Date.ToString(DateFormat.EightCharacter)},{Math.Round(PriceFactor, 6).Normalize()},{Math.Round(SplitFactor, 7).Normalize()},{Math.Round(ReferencePrice, 2).Normalize()}{source}";
+            return $"{Date.ToString(DateFormat.EightCharacter)}," +
```
mchandschuh: has this branch been rebased on `master`? There are new functions in `master` for properly `ToString`-ing values using the invariant culture. This `Date.ToString` can be updated to `Date.ToStringInvariant`.

IlshatGaripov: @mchandschuh I will be fixing this very shortly. :+1: Thanks!

Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -306,10 +306,9 @@ public string ToCsv(string source = null)
         {
             source = source == null ? "" : $",{source}";
             return $"{Date.ToStringInvariant(DateFormat.EightCharacter)}," +
-                   Invariant($"{Math.Round(PriceFactor, 6).Normalize()},") +
-                   Invariant($"{Math.Round(SplitFactor, 7).Normalize()},") +
-                   Invariant($"{Math.Round(ReferencePrice, 2).Normalize()}") +
-                   $"{source}";
+                   $"{PriceFactor.Normalize()}," +
```
mchandschuh: I think we wanted to keep the `Invariant` here -- this ensures that when we generate the string it uses the invariant culture instead of the machine's current culture setting.

IlshatGaripov:
@mchandschuh There were few things discovered:

First of all - the formula again - please compare :
https://github.com/QuantConnect/Lean/blob/master/Common/Data/Auxiliary/FactorFileRow.cs#L182
with the one in FactorFileGenerator.cs from this PR we evaluated to be correct.. 
The same applies to splits:
https://github.com/QuantConnect/Lean/blob/master/Common/Data/Auxiliary/FactorFileRow.cs#L226
And no roundings here !

Also, please give a look at different but related test:
https://github.com/QuantConnect/Lean/blob/master/Tests/Common/Data/Auxiliary/FactorFileTests.cs#L198
Does not fail but gives the following output (last three lines):
```
expected: 2012-08-08: 0.1271 0.1429 actual: 2012-08-08: 0.1271 0.1429  diff: 0.0000%
expected: 2005-02-25: 0.0635 0.0714 actual: 2005-02-25: 0.0635 0.0714  diff: 0.0000%
expected: 2000-06-20: 0.0318 0.0357 actual: 2000-06-20: 0.0318 0.0357  diff: 0.0000%
```
where did the earliest enty go?
```
19980102,0.8893653,0.0357143,16.25
```

Altogether it seems to me that FactorFile.Apply( ) method along with FactorFileRow's Apply methods for split/divident it uses inside may work not as they should be. And we have two options - to fix them. Or what seems to be more elegant to me to use upgraded FactorFIleGenerator with a constructor overload that takes a list of dailyDataForEquity and called it from FactorFile Apply()

FactorFIleGenerator needs this daily data to make the rows and he loads them from folder, but he only uses the tradeBars's Time and Value properties and only on the days when need to construct the factor rows and we have those values in Apply(). I am publishing a sketch of this concept and would appreciate a review. Though I could not test this entirely as FactorFileGenerator has to be return to QC.Common to enable this to run. Thanks!

IlshatGaripov:
@mchandschuh The question is does the proposed solution look fine? And would that be ok to move 
FactorFileGenerator and LeanParser (that generator uses inside ) from QuantConnect.Toolbox -> QuantConnect.Common.cproj to finilize the sketch and test the solution ?

IlshatGaripov:
@I am closing this issue because @jaredbroad is starter

IlshatGaripov:
you woudl ask what does it mean... it does nothing!

IlshatGaripov:
he is just smarter!!!

IlshatGaripov:
he is just the smartest in this community!

IlshatGaripov:
@jaredbroad thank you! terribly drunkg I am sorry. thanks!

IlshatGaripov:
@jaredbroad I don't why I not intelligent in comon dicusssion... please excue me... I do aplologize..
is that ehough?

IlshatGaripov:
it's a very bad trait to consider yourself unlike others -

IlshatGaripov:
Pllease clocse it

IlshatGaripov:
FOR REAL! TALANTED PEOPLE WHEN THEY AR AROUND YOU AFFORD TO PRONOUNCE NONSE.!

IlshatGaripov:
i AM OFFIDICALLT LEACINF PORJECT

IlshatGaripov:
Desripte @jaredbroad has so mucha good people around him I am not going to work further on his prpjjects excuse me. sorry because it's 5 pm and my girls friend is coming home at th evening there are things I have to consider. sorry good luck everybody.. jaresd is talante

IlshatGaripov:
it's an idiot thinkgs you did jared

Issue #3556, opened 2019-09-03
CoinApi symbol list duplicated symbol breaks converter


Jay-Jay-D:
#### Expected Behavior

Being able to process CoinApi symbol list 

#### Actual Behavior

There are cases where duplicated symbols may occur, as they explain in their [docs](https://docs.coinapi.io/#list-all-symbols)
> In the unlikely event when the "symbol_id" for more than one market is the same. We will append the additional term (prefixed with the "_") at the end of the duplicated identifiers to differentiate them.

#### Potential Solution

Filter the brokerage symbols before adding them into a dictionary as key.

#### Reproducing the Problem

Run `CoinApiSymbolMapper` with cases like the ones below:
![image](https://user-images.githubusercontent.com/10123175/64192524-6651c180-ce6a-11e9-9f62-f79ade9d2941.png)

#### System Information

Linux 

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3570, opened 2019-09-11
Fix error when CoinApi symbol list has duplicates

Jay-Jay-D:
#### Description

Handle cases where the symbol list has duplicated.

#### Related Issue

#3556 

#### Motivation and Context

If there are duplicates, they break the crypto data processing.

We assumed CoinAPi symbol list has no duplicates, but they those cases are expected, as we can read in their docs https://docs.coinapi.io/#list-all-symbols

>  In the unlikely event when the "symbol_id" for more than one market is the same. We will append the additional term (prefixed with the "_") at the end of the duplicated identifiers to differentiate them.

After asking CoinApi support,  they inform us that symbols with the extra suffix are for edge cases.
> There could be corner-cases for other data types like FUTURES or OPTIONS, e.g., exchange lists same option but one with American and other with European execution-style

No.

Processed ten days of Bitfinex data.

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/CoinApi/CoinApiSymbolMapper.cs
+++ b/ToolBox/CoinApi/CoinApiSymbolMapper.cs
@@ -110,13 +110,13 @@ namespace QuantConnect.ToolBox.CoinApi
             var parts = brokerageSymbol.Split('_');
             if (parts.Length != 4 || parts[1] != "SPOT")
             {
-                Log.Error($"CoinApiSymbolMapper.GetLeanSymbol(): Unsupported SymbolId: {brokerageSymbol}");
+                throw new Exception($"CoinApiSymbolMapper.GetLeanSymbol(): Unsupported SymbolId: {brokerageSymbol}");
             }
 
             string symbolMarket;
             if (!MapExchangeIdsToMarkets.TryGetValue(parts[0], out symbolMarket))
             {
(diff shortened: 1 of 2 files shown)


ToolBox/CoinApi/CoinApiSymbolMapper.cs
```diff
@@ -166,7 +166,7 @@ private void LoadSymbolMap(string[] exchangeIds)
             var result = JsonConvert.DeserializeObject<List<CoinApiSymbol>>(json);
 
             _symbolMap = result
-                .Where(x => x.SymbolType == "SPOT")
+                .Where(x => x.SymbolType == "SPOT" && x.SymbolId.Split('_').Length == 4)
```
mchandschuh: This might benefit from a code comment

ToolBox/CoinApiDataConverter/CoinApiDataConverter.cs
```diff
@@ -83,6 +83,7 @@ public bool Run()
             var success = true;
 
             var fileToProcess = _rawDataFolder.EnumerateFiles("*.gz")
+                .Where(f => f.Name.Split('_').Length == 4)
```
mchandschuh: here as well -- _although_ -- the commit message is **terrific** so the code comment isn't mandatory 👍

mchandschuh:
Thanks @Jay-Jay-D -- great stuff!

PR #3571, opened 2019-09-12
Algorithm Portfolio is Updated Before New Order Event is Triggered

AlexCatarino:
#### Description
In `BrokerageTransactionHandler.HandleOrderEvent`, when a fill order event is handled, the `IAlgorithm.Portfolio` will be updated before the NewOrderEvent event handler is called.

Due to this change, subscribers to `IOrderEventProvider.NewOrderEvent` will have a `IAlgorithm.Portfolio` that reflects the latest fill.

- Updates OrderBasedInsightGenerator to Reflect Changes in BrokerageTransactionHandler
  - Logic in `OrderBasedInsightGenerator.GenerateInsightFromFill` was changed to assumed that `SecurityHolding` reflects  the order event.

#### Related Issue
Required for #3539

#### How Has This Been Tested?
Unit tests.

#### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #3573, opened 2019-09-13
Implements AssetsUnderManagementCapacityManager

### LINKED ISSUE #3539: Implement AUM Capacity Estimate Calculation
#### Expected Behavior
Calculate an estimate of the capacity of the strategy backtest/live trade. This will provide funds with an understanding of the maximum dollar allocation they can make to the strategy to maximize its return potential without moving the market.

#### Actual Behavior
No estimate available

#### Potential Solution
Capacity will be calculated using the volume capacity per symbol per trade each day. At the end of the day, the symbol capacity will be calculated as the minimum of all individual trade capacities for the specific symbol. Then, the portfolio capacity for the day will be the sum of all individual symbol capacities. The final AUM Capacity will be displayed as the average of all of the daily capacities calculated).

Please see the excel sheet for a detailed calculation of this in various circumstances.
[Alpha Capacity Calculations.xlsx](https://github.com/QuantConnect/Lean/files/3552416/Alpha.Capacity.Calculations.xlsx)

#### Checklist

AlexCatarino:
#### Description
Calculate an estimate of the capacity of the strategy backtest/live trade. This will provide funds with an understanding of the maximum dollar allocation they can make to the strategy to maximize its return potential without moving the market.

- Implements `AssetsUnderManagementCapacityManager`
- Adds an instance of `AssetsUnderManagementCapacityManager` in the result handlers to include AUM Capacity to algorithm statistics
- Adds "AUM Capacity" entry to `ExpectedStatistics` of regression algorithms with a considerable amount of orders.

Performance test over 10 executions of RegressionAlgorithm (5433 trades):
`master` :: Mean: `7.71` Std: `0.16`
`new` :: Mean: `7.853` Std: `0.22`
Speed decrease of 1.85%.

#### Related Issue
Closes #3539
This pull request was created on top of #3571

#### Motivation and Context
Include statistics that can be used to improve algorithms. 

No.

New unit tests and regression algorithms.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Statistics/AssetsUnderManagementCapacityManager.cs
+++ b/Common/Statistics/AssetsUnderManagementCapacityManager.cs
@@ -1,0 +1,238 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
+using QuantConnect.Indicators;
+using QuantConnect.Logging;
+using QuantConnect.Util;
+
+namespace QuantConnect.Statistics
+{
+    /// <summary>
+    /// Class in charge of calculating the Assets Under Management (AUM) Capacity values.
+    /// Will use the sample values of the last year.
+    /// </summary>
+    /// <remarks>See https://www.quantconnect.com/forum/discussion/6194/insight-scoring-metric/p1 </remarks>
+    public class AssetsUnderManagementCapacityManager
+    {
+        private readonly object _lock = new object();
+
+        private readonly decimal _maximumTradeableVolume;
+        private readonly SecurityPortfolioManager _portfolio;
+        private readonly ISubscriptionDataConfigProvider _subscriptionDataConfigProvider;
+
+        private readonly Dictionary<Symbol, decimal> _maximumSymbolCapacity;
+        private readonly RollingWindow<decimal> _historicalPortfolioCapacity;
+        private DateTime _previousInputTime;
+
+        /// <summary>
+        /// Assets Under Management (AUM) Capacity
+        /// </summary>
+        public decimal AumCapacity =>
+            _historicalPortfolioCapacity.Count == 0 ? 0 : _historicalPortfolioCapacity.Average();
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AssetsUnderManagementCapacityManager"/> class
+        /// </summary>
+        /// <param name="portfolio">The algorithm's portfolio</param>
+        /// <param name="subscriptionDataConfigProvider">Provides access to registered <see cref="SubscriptionDataConfig"/></param>
+        /// <param name="orderEventProvider">Provides access to the order events</param>
+        public AssetsUnderManagementCapacityManager(
+            SecurityPortfolioManager portfolio,
+            ISubscriptionDataConfigProvider subscriptionDataConfigProvider,
+            IOrderEventProvider orderEventProvider)
+        {
+            _maximumTradeableVolume = PortfolioStatistics.GetMaximumTradeableVolume();
+            _portfolio = portfolio;
+            _subscriptionDataConfigProvider = subscriptionDataConfigProvider;
+
+            _maximumSymbolCapacity = new Dictionary<Symbol, decimal>();
+            _historicalPortfolioCapacity = new RollingWindow<decimal>(30);
+            _previousInputTime = DateTime.MinValue;
+
+            orderEventProvider.NewOrderEvent += (sender, orderEvent) =>
+            {
+                var symbol = orderEvent.Symbol;
+
+                var configs = _subscriptionDataConfigProvider.GetSubscriptionDataConfigs(symbol);
+                if (configs.IsNullOrEmpty())
+                {
+                    Log.Error($"AssetsUnderManagementCapacityManager: Could not find {symbol} in Portfolio");
+                }
+
+                try
+                {
+                    var consolidator = default(IDataConsolidator);
+                    var security = _portfolio.Securities[symbol];
+                    BaseData lastData = security.Cache.GetData<TradeBar>();
+
+                    // For high resolution data, create a consolidator that will compute
+                    // the AUM Capacity after a 5-minute period bar is closed
+                    foreach (var config in configs.Where(x => x.Resolution < Resolution.Hour))
+                    {
+                        if (config.Type.IsAssignableFrom(typeof(TradeBar)))
+                        {
+                            consolidator = new TradeBarConsolidator(TimeSpan.FromMinutes(5));
+                        }
+
+                        if (config.Type.IsAssignableFrom(typeof(Tick)) &&
+                            config.TickType == TickType.Trade)
+                        {
+                            consolidator = new TickConsolidator(TimeSpan.FromMinutes(5));
+                            lastData = security.Cache.GetData<Tick>();
+                        }
+
+                        if (consolidator == null)
+                        {
+                            continue;
+                        }
+
+                        // Warm up the consolidator to mark the begging of the period
+                        if (lastData != null)
+                        {
+                            consolidator.Update(lastData);
+                        }
+
+                        config.Consolidators.Add(consolidator);
+                        consolidator.DataConsolidated += (s, data) =>  Update(s, data, orderEvent.OrderId);
+                        return;
+                    }
+
+                    // For low resolution, use the last data available
+                    if (lastData != null)
+                    {
+                        Update(sender, lastData, orderEvent.OrderId);
+                    }
+                }
+                catch (Exception e)
+                {
+                    Log.Error($"AssetsUnderManagementCapacityManager: {e}");
+                }
+            };
+        }
+
+        /// <summary>
+        /// Updates the AUM Capacity with the latest data available after a order fill
+        /// </summary>
+        /// <param name="sender">For high resolution, sender object represents the data consolidator</param>
+        /// <param name="data">Last price trade bar available or consolidated</param>
+        /// <param name="orderId">Order fill ID</param>
+        /// <remarks>
+        /// The whole method body is locked, since order events can be triggered at any time in live mode
+        /// </remarks>
+        private void Update(object sender, IBaseData data, int orderId)
+        {
+            lock (_lock)
+            {
+                try
+                {
+                    var symbol = data.Symbol;
+                    var security = _portfolio.Securities[symbol];
+
+                    var holdingsTurnover = 1m;
+                    var holdingsValue = security.Holdings.AbsoluteHoldingsValue;
+
+                    if (holdingsValue != 0)
+                    {
+                        var order = _portfolio.Transactions.GetOrderById(orderId);
+                        var orderValue = Math.Abs(order.GetValue(security));
+                        holdingsTurnover = orderValue / holdingsValue;
+
+                        if (holdingsTurnover == 0)
+                        {
+                            RemoveConsolidator(sender, symbol);
+                            return;
+                        }
+                    }
+
+                    var totalMarketVolume = GetTotalMarketVolume(data);
+                    var totalTradeVolumeCapacity = _maximumTradeableVolume * totalMarketVolume / holdingsTurnover;
+
+                    // If it is a new day, we discard the previous day data
+                    var utcDate = data.EndTime.ConvertToUtc(security.Exchange.TimeZone).Date;
+                    if (_previousInputTime != utcDate)
+                    {
+                        _previousInputTime = utcDate;
+                        _maximumSymbolCapacity.Clear();
+                        _historicalPortfolioCapacity.Add(0m);
+                    }
+
+                    // Updates AUM Capacity if there is new information or lower total trade volume capacity for the security
+                    decimal capacity;
+                    if (!_maximumSymbolCapacity.TryGetValue(symbol, out capacity) ||
+                        totalTradeVolumeCapacity < capacity)
+                    {
+                        _maximumSymbolCapacity[symbol] = totalTradeVolumeCapacity;
+                        _historicalPortfolioCapacity[0] = _maximumSymbolCapacity.Sum(x => x.Value);
+                    }
+
+                    RemoveConsolidator(sender, symbol);
+                }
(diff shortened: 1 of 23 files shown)


Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                    {
+                        _maximumSymbolCapacity[symbol] = totalTradeVolumeCapacity;
+                        _historicalPortfolioCapacity[0] = _maximumSymbolCapacity.Sum(x => x.Value);
+                    }
+
+                    RemoveConsolidator(sender, symbol);
+                }
+                catch (Exception e)
+                {
+                    Log.Error($"AssetsUnderManagementCapacityManager.Update: {e}");
```
Martin-Molinero: There seems to be a chance of an exception being thrown and the consolidator not being removed, maybe we can move the `RemoveConsolidator(sender, symbol);` for it to be the first line in the `try/catch` block?

Engine/Results/BaseResultsHandler.cs
```diff
@@ -210,7 +215,7 @@ protected decimal GetNetReturn()
                     var trades = Algorithm.TradeBuilder.ClosedTrades;
 
                     statisticsResults = StatisticsBuilder.Generate(trades, profitLoss, equity, performance, benchmark,
-                        StartingPortfolioValue, Algorithm.Portfolio.TotalFees, totalTransactions);
+                        StartingPortfolioValue, Algorithm.Portfolio.TotalFees, totalTransactions, AssetsUnderManagementCapacityManager.AumCapacity);
```
Martin-Molinero: I think there is a small chance that `AssetsUnderManagementCapacityManager` is null if there is some error during init, we might add a check and use `0` if that's the case

Tests/Common/Statistics/AssetsUnderManagementCapacityManagerTests.cs
```diff
@@ -0,0 +1,304 @@
...
+using Moq;
+using QuantConnect.Brokerages.Backtesting;
+using QuantConnect.Data;
+using QuantConnect.Lean.Engine.TransactionHandlers;
+using QuantConnect.Tests.Engine;
+
+namespace QuantConnect.Tests.Common.Statistics
+{
+    [TestFixture]
+    public class AssetsUnderManagementCapacityManagerTests
```
Martin-Molinero: Nice tests! 👍 
I think one test case that is missing, and would be nice to have, is for the same date, the same symbol, two trades and assert it uses the minimum trade capacity of both trades?

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                            consolidator.Update(lastData);
+                        }
+
+                        config.Consolidators.Add(consolidator);
+                        consolidator.DataConsolidated += (s, data) =>  Update(s, data, orderEvent.OrderId);
+                        return;
+                    }
+
+                    // For low resolution, use the last data available
+                    if (lastData != null)
```
mchandschuh: Any concern is duplicating this data point in the consolidator? For example, having second SPY and daily SPY -- would we end up with the daily SPY config's consolidator updated each second? Should this be set up to limit the consolidators to one per unique symbol?

Common/Statistics/AssetsUnderManagementCapacityManager.cs
mchandschuh: Also, please pass the error object as it's own argument to `Log.Error` to ensure the full exception object makes it down to the configured logging framework. For QC, we end up doing something similar to what you have here, but I know the community has made adapters for various logging framework, such as log4net, which greatly benefit from having access to the raw exception object.

Another also, the catch all exception handling makes me uneasy. Is this necessary? Why not just let it throw an exception? Is there potential for a non-fatal error being raised?

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// Gets the total market volume which is a fraction of the current trade bar volume
+        /// </summary>
+        /// <param name="data">Last price trade bar available or consolidated</param>
+        private static decimal GetTotalMarketVolume(IBaseData data)
+        {
+            var tradeBar = (TradeBar) data;
+
+            var factor = tradeBar.Period == Time.OneDay
+                ? .025m
+                : .050m;
```
mchandschuh: A comment here describing _why_ we're using `0.025m` and `0.050m` -- in fact, I'm not sure I understand what these factors mean.

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                {
+                    Log.Error($"AssetsUnderManagementCapacityManager: {e}");
+                }
+            };
+        }
+
+        /// <summary>
+        /// Updates the AUM Capacity with the latest data available after a order fill
+        /// </summary>
+        /// <param name="sender">For high resolution, sender object represents the data consolidator</param>
```
mchandschuh: Is this comment correct? This [line](https://github.com/QuantConnect/Lean/pull/3573/files#diff-6b9a41074e2cd8dc4fdbf36e8b798a2fR124) appears to be passing the `sender` object provided via `IOrderEventProvider.NewOrderEvent` which is likely the transaction handler, correct?

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// <param name="data">Last price trade bar available or consolidated</param>
+        /// <param name="orderId">Order fill ID</param>
+        /// <remarks>
+        /// The whole method body is locked, since order events can be triggered at any time in live mode
+        /// </remarks>
+        private void Update(object sender, IBaseData data, int orderId)
```
mchandschuh: Why pass `object sender` here? Since this method is not _directly_ used as an event handler, we shouldn't be using `object sender` but instead should be using something more strongly typed. Given the invocations above, it appears this is sometimes the transaction handler. It may make sense to define two separate methods, one for high resolution and one for low resolution since it looks like they'll be receiving different `sender` types at run time.

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            return factor * tradeBar.Volume;
+        }
+
+        /// <summary>
+        /// Remove consolidator for a given symbol
+        /// </summary>
+        /// <param name="sender">Consolidator to be removed</param>
+        /// <param name="symbol">Symbol associated with the consolidator</param>
+        private void RemoveConsolidator(object sender, Symbol symbol)
```
mchandschuh: Since this method isn't directly used as an event handler we should remove the `object sender` and replace it with a more accurate type (perhaps `IDataConsolidator`). Also, if I'm reading this correctly, not sure if we'll ever remove consolidators for hour/daily resolutions.

Common/Statistics/PortfolioStatistics.cs
```diff
@@ -28,6 +28,7 @@ namespace QuantConnect.Statistics
     public class PortfolioStatistics
     {
         private const decimal RiskFreeRate = 0;
+        private const decimal MaximumTradeableVolume = 0.025m;
```
mchandschuh: `MaximumTradeableVolume`? or is this more like a `MaximumTradeableVolumePercent` ? This concerns me since it looks like a _magic_ number and there's no description as to why it's magic and as to how the value was determined. I've been reading these factors as 2.5%

Common/Statistics/PortfolioStatistics.cs
```diff
@@ -235,6 +236,14 @@ public static decimal GetRiskFreeRate()
             return RiskFreeRate;
         }
 
+        /// <summary>
+        /// Gets the current defined maximum tradeable volume
+        /// </summary>
+        public static decimal GetMaximumTradeableVolume()
+        {
+            return MaximumTradeableVolume;
```
mchandschuh: If we're going to make the value a constant _and then_ provide a static function to return the constant, why not just make the constant directly public? I could see if this value was settable and/or overridable via a subclass or similar, but as written, it seems like an extra level of indirection for no benefit

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -998,20 +998,20 @@ private void HandleOrderEvent(OrderEvent fill)
...
-
                         _algorithm.Portfolio.ProcessFill(fill);
                         _algorithm.TradeBuilder.ProcessFill(
                             fill,
                             securityConversionRate,
                             feeInAccountCurrency,
                             multiplier);
+
+                        // to be called after updating the Portfolio
+                        NewOrderEvent?.Invoke(this, fill);
```
mchandschuh: I'm wondering if there's any concern about silently breaking existing algorithms by changing the temporal behavior as to when this event is raised. I say this will silently break algorithms since it likely won't raise an error but rather would end in the algorithm tracking state incorrectly thereby silently corrupting the algorithm's state. _If_ this is a concern, then perhaps a _new_ event would be warranted, something like `NewOrderEventProcessed` (comments on event naming below in `<nit>`).

``` xml
<nit severity='low' type='naming-conventions' scope='events'>
```
unrelated to this PR -- `NewOrderEvent` is a **really really** _**really**_ bad event name. [Event names](https://docs.microsoft.com/en-us/dynamicsax-2012/developer/naming-conventions-delegates-and-event-handlers) should be verbs as either present tense (*ing, such as `OrderSubmitting`) and denote that the event was raised _right before_, so in this case, `OrderSubmitting` implies the event handler is invoked _before_ the order is actually submitted. Following completion of submitting the order, another event would be called `OrderSubmitted`. These past tense verbs imply that the described event has been completed.

`NewOrderEvent` is ambiguous. Is this a new order event was _initialized_ -- or was it _received_, or was it _processed_ or was it _something else_ ? It appears it is intended as not just a new order event, but actually a new _fill_ or _partial fill_ event. I think a better name for the event _as-is_ on `master` would be `OrderFillProcessing`  and this new event would be `OrderFillProcessed`

Making things even more ambiguous, the documentation for `NewOrderEvent` only states:
``` csharp
/// <summary>
/// Event fired when there is a new <see cref="QuantConnect.Orders.OrderEvent"/>
/// </summary>
/// <remarks>Will be called before the <see cref="SecurityPortfolioManager"/></remarks>
```
which doesn't even mention that the event is only raised for fill/partial fill `OrderEvent` instances.
``` xml
</nit>
```

AlexCatarino: We (@Martin-Molinero and I) have discussed this. `NewOrderEvent` is a recent feature and it was only consumed by `OrderBasedInsightGenerator`, so we choose to refactor `OrderBasedInsightGenerator` instead of creating a new event (ref. #3571).

Common/Statistics/AssetsUnderManagementCapacityManager.cs
AlexCatarino: This factor is the maximum percentage of market volume that is tradeable. These values are ad-hoc defined by research.

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                        _maximumSymbolCapacity.Clear();
+                        _historicalPortfolioCapacity.Add(0m);
+                    }
+
+                    // Updates AUM Capacity if there is new information or lower total trade volume capacity for the security
+                    decimal capacity;
+                    if (!_maximumSymbolCapacity.TryGetValue(symbol, out capacity) ||
+                        totalTradeVolumeCapacity < capacity)
+                    {
+                        _maximumSymbolCapacity[symbol] = totalTradeVolumeCapacity;
```
jaredbroad: I think the units are currently in shares not dollars. Need to multiply the `GetTotalMarketVolume` by `asset price x contract multiplier` somewhere to get AUM capacity.

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+                {
+                    var symbol = data.Symbol;
+                    var security = _portfolio.Securities[symbol];
+
+                    var holdingsTurnover = 1m;
+                    var holdingsValue = security.Holdings.AbsoluteHoldingsValue;
+
+                    if (holdingsValue != 0)
+                    {
+                        var order = _portfolio.Transactions.GetOrderById(orderId);
```
jaredbroad: Add comments please

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,238 @@
...
+
+                        if (holdingsTurnover == 0)
+                        {
+                            RemoveConsolidator(sender, symbol);
+                            return;
+                        }
+                    }
+
+                    var totalMarketVolume = GetTotalMarketVolume(data);
+                    var totalTradeVolumeCapacity = _maximumTradeableVolume * totalMarketVolume / holdingsTurnover;
```
jaredbroad: not sure if intentional but you're multiplying by 2.5% twice,

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,245 @@
...
+                return;
+            }
+
+            var consolidator = default(IDataConsolidator);
+            var hasHighResolution = false;
+            var security = _portfolio.Securities[symbol];
+            BaseData lastData = security.Cache.GetData<TradeBar>();
+
+            // For high resolution data, create a consolidator that will compute
+            // the AUM Capacity after a 5-minute period bar is closed
```
mchandschuh: maybe a note here saying that we're only adding the consolidator for the first config -- also, should there be preference of trades over quotes or vice versa? Right now we're just depending on whatever order the provider returns them in which (may) cause non-determinism later (perhaps)?

Common/Statistics/AssetsUnderManagementCapacityManager.cs
```diff
@@ -0,0 +1,245 @@
...
+
+            return factor * tradeBar.Volume * tradeBar.Price;
+        }
+
+        /// <summary>
+        /// Remove consolidator for a given symbol
+        /// </summary>
+        /// <param name="consolidator">Consolidator to be removed</param>
+        /// <param name="symbol">Symbol associated with the consolidator</param>
+        private void RemoveConsolidator(IDataConsolidator consolidator, Symbol symbol)
```
mchandschuh: This might make a nice extension method -- I can see this being applicable in many places. I like it!

Engine/AlgorithmManager.cs
```diff
@@ -545,7 +545,8 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                         var timeKeeper = algorithm.TimeKeeper;
                         foreach (var update in timeSlice.ConsolidatorUpdateData)
                         {
-                            var consolidators = update.Target.Consolidators;
+                            // Creates a list from the hashset to enable item removal in the IDataConsolidator.Update callback
+                            var consolidators = update.Target.Consolidators.ToList();
```
mchandschuh: I feel like this _shouldn't_ be required... and if a callback is causing modification of a list that raises an exception _somewhere_ than I suspect we're doing something else wrong and this may just be a bandaid.

Engine/TransactionHandlers/BrokerageTransactionHandler.cs
mchandschuh: Thanks for the reference -- not just for me, but for posterity 😄

jaredbroad:
Closing as inaccurate and we should implement a pluggable solution `ICapacityEstimateModel` and study existing papers about the solutions they've created.

PR #3575, opened 2019-09-16
Implements PsychSignal Consolidator

### LINKED ISSUE #3574: Implement PsychSignal Consolidator
#### Expected Behavior

Because PsychSignal data and other custom data sources can sometimes provide more utility when they are resampled/consolidated, being able to consolidate the data might provide further insights or utility.
#### Actual Behavior

No such consolidator exists
#### Potential Solution

Implement the consolidator
#### Reproducing the Problem

Attempt to resample/consolidate PsychSignal data in C#
#### System Information

N/A
#### Checklist

gsalaz98:
#### Description

Provides new `PsychSignalConsolidator` and `PsychSignalConsolidated` classes for usage in consolidating PsychSignal data.

Typo fix in `PsychSignalSentimentData` documentation

#### Related Issue

Closes #3574 
#### Motivation and Context

Previously, resampling data was only easily achievable by using pandas' `DataFrame.resample` method. However in C#, resampling/consolidating data poses a challenge. 

Reaches parity with Python

No

New unit tests included for 3 different Resolutions
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Consolidators/PsychSignalConsolidator.cs
+++ b/Common/Data/Consolidators/PsychSignalConsolidator.cs
@@ -70,14 +70,6 @@ namespace QuantConnect.Data.Consolidators
         /// <param name="data">The new data</param>
         protected override void AggregateBar(ref PsychSignalConsolidated workingBar, PsychSignalSentimentData data)
         {
-            var bullIntensity = data.BullIntensity;
-            var bearIntensity = data.BearIntensity;
-            var bullMinusBear = data.BullMinusBear;
-            var bullScoredMessages = data.BullScoredMessages;
-            var bearScoredMessages = data.BearScoredMessages;
-            var bullBearMessageRatio = data.BullBearMessageRatio;
-            var totalScoredMessages = data.TotalScoredMessages;
-
             if (workingBar == null)
             {
                 workingBar = new PsychSignalConsolidated
@@ -96,12 +88,12 @@ namespace QuantConnect.Data.Consolidators
                 return;
             }
 
-            UpdateBar(workingBar.BullIntensity, bullIntensity);
-            UpdateBar(workingBar.BearIntensity, bearIntensity);
-            UpdateBar(workingBar.BullMinusBear, bullMinusBear);
+            UpdateBar(workingBar.BullIntensity, data.BullIntensity);
+            UpdateBar(workingBar.BearIntensity, data.BearIntensity);
+            UpdateBar(workingBar.BullMinusBear, data.BullMinusBear);
             workingBar.BullScoredMessages += data.BullScoredMessages;
             workingBar.BearScoredMessages += data.BearScoredMessages;
-            UpdateBar(workingBar.BullBearMessageRatio, bullBearMessageRatio);
+            UpdateBar(workingBar.BullBearMessageRatio, data.BullBearMessageRatio);
             workingBar.TotalScoredMessages += data.TotalScoredMessages;
             workingBar.EndTime = GetRoundedBarTime(data.EndTime);
         }
--- a/Common/Data/Consolidators/PsychSignalConsolidator.cs
+++ b/Common/Data/Consolidators/PsychSignalConsolidator.cs
@@ -1,0 +1,122 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using QuantConnect.Data.Custom.PsychSignal;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Data.Consolidators
+{
+    /// <summary>
+    /// Consolidates PsychSignal data to the time period specified
+    /// </summary>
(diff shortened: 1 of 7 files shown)


Common/Data/Consolidators/PsychSignalConsolidator.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="data">The new data</param>
+        protected override void AggregateBar(ref PsychSignalConsolidated workingBar, PsychSignalSentimentData data)
+        {
+            var bullIntensity = data.BullIntensity;
+            var bearIntensity = data.BearIntensity;
+            var bullMinusBear = data.BullMinusBear;
+            var bullScoredMessages = data.BullScoredMessages;
+            var bearScoredMessages = data.BearScoredMessages;
+            var bullBearMessageRatio = data.BullBearMessageRatio;
+            var totalScoredMessages = data.TotalScoredMessages;
```
mchandschuh: why define these here and then not use them when initializing `workingBar`?

I think this function will read better by simply in-lining all of these variables

Common/Data/Custom/PsychSignal/PsychSignalConsolidated.cs
```diff
@@ -0,0 +1,86 @@
...
+            return new PsychSignalConsolidated
+            {
+                Time = Time,
+                Symbol = Symbol,
+                BullIntensity = BullIntensity,
+                BearIntensity = BearIntensity,
+                BullMinusBear = BullMinusBear,
+                BullScoredMessages = BullScoredMessages,
+                BearScoredMessages = BearScoredMessages,
+                BullBearMessageRatio = BullBearMessageRatio,
```
mchandschuh: I think we should be cloning all of the `Bar` objects as well. If we _don't_  deep clone these then a change to the old  object would cause the change to also be seen in the cloned object.

gsalaz98: Great catch! Thankfully the solution was fairly simple :)

Tests/Common/Data/PsychSignalConsolidatorTests.cs
```diff
@@ -0,0 +1,530 @@
...
+    public class PsychSignalConsolidatorTests
+    {
+        [Test]
+        public void ConsolidatesMockPsychSignalDataAccurately_Daily()
+        {
+            // Note: This data is not real, and was created for testing purposes only
+            var consolidator = new PsychSignalConsolidator(TimeSpan.FromDays(1));
+            consolidator.DataConsolidated += (_, data) =>
+            {
+                Assert.AreEqual(new DateTime(2019, 1, 1), data.Time);
```
Martin-Molinero: Nice tests! 👍 
I think we could add an assertion on whether the callback `DataConsolidated `  was called or not at the end of each unit test
nit - might be more precise to move this file to ` Tests/Common/Data/Custom`

Common/Data/Custom/PsychSignal/PsychSignalConsolidated.cs
```diff
@@ -0,0 +1,86 @@
...
+        public int BearScoredMessages { get; set; }
+
+        /// <summary>
+        /// OHLC Bull/Bear message ratio.
+        /// </summary>
+        public Bar BullBearMessageRatio { get; set; }
+
+        /// <summary>
+        /// Aggregate total of messages scanned.
+        /// </remarks>
```
Martin-Molinero: typo `</remarks>` -> `</summary>`

Common/Data/Consolidators/PsychSignalConsolidator.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            UpdateBar(workingBar.BullIntensity, data.BullIntensity);
+            UpdateBar(workingBar.BearIntensity, data.BearIntensity);
+            UpdateBar(workingBar.BullMinusBear, data.BullMinusBear);
+            workingBar.BullScoredMessages += data.BullScoredMessages;
+            workingBar.BearScoredMessages += data.BearScoredMessages;
+            UpdateBar(workingBar.BullBearMessageRatio, data.BullBearMessageRatio);
+            workingBar.TotalScoredMessages += data.TotalScoredMessages;
+            workingBar.EndTime = GetRoundedBarTime(data.EndTime);
```
Martin-Molinero: just checking: are `PsychSignalSentimentData` members always set in each data point? peaked on some of them and saw a bunch of `0` - if not, thinking if those cases should be skipped/ignored when consolidating? Can follow up offline

Common/Data/Consolidators/PsychSignalConsolidator.cs
```diff
@@ -0,0 +1,114 @@
...
+using System;
+using QuantConnect.Data.Custom.PsychSignal;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Data.Consolidators
+{
+    /// <summary>
+    /// Consolidates PsychSignal data to the time period specified
+    /// </summary>
+    public class PsychSignalConsolidator : PeriodCountConsolidatorBase<PsychSignalSentimentData, PsychSignalConsolidated>
```
mchandschuh: Love seeing this infrastructure work we put in years ago paying dividends :D

Tests/Common/Data/PsychSignalConsolidatorTests.cs
mchandschuh: RE: moving files to match namespaces -- completely agreed 👍

Common/Data/Consolidators/PsychSignalConsolidator.cs
gsalaz98: PsychSignal data can contain a lot of `0` values, but will always contain at least one value per row of data. This means that the consolidator will always be doing something, and not be consolidating empty data.

Martin-Molinero: Hm given it can contain a lot of 0 values, I'm trying to think of the use case here, user perspective, consider first we get `data.BullIntensity` `1` and afterward we get a bunch of `0s` -> what does the consolidated data, as a `Bar` instance, mean? What's the value in it? 🤔

Martin-Molinero: Data provider specified that 0 values could be a neutral score (information) or a no score/message case (not information, should be skipped), and that we should `differentiate them checking the bull_scored_messages and bear_scored_messages fields to see if the count is "0".`

jaredbroad:
Closing for now but let's revisit later.

PR #3580, opened 2019-09-17
Extend AddData Unit Tests For Underlying Symbol Linking

gsalaz98:
### ⚠️ This is a draft PR meant to be merged with changes not yet implemented in LEAN master. Do not merge ⚠️

This PR adds unit tests and regression algorithms that test for custom data linking to other asset classes via the `Underlying` property in the Symbol object. What this means is that we'd be able to create a link between custom data and traditional asset classes and access the traditional asset Symbol via the `Underlying` property, as well as allowing the addition of mapped Symbols in AddData by adding via a Symbol.

Regression algorithms tests for mapping behavior being applied to the `Underlying` Symbol.

Please note that this PR does not add any new functionality, only unit tests for future changes and should be failing for the meanwhile.
#### Related Issue

N/A
#### Motivation and Context

Currently, we have no reliable way of creating a link between custom data and traditional asset classes. This makes placing orders and tracking the traditional asset's Symbol very difficult and can result in placing orders for the wrong Symbol. 

We don't have a way to add a currently mapped symbol with AddData. This has implications during universe selection, where the Symbols are already mapped. AddData currently makes the assumption that the ticker passed to it is the present day ticker, which makes it impossible to subscribe to the correct ticker while the traditional asset Symbol has been mapped.

Yes. In particular, for mapped data sources, we'd have to address the fact that it can create a link to the Symbol passed through to the custom data source.

New unit tests. Subject to change
#### Types of changes

#### Checklist:

#### Additional Comments:
Will squash or rebase on request


Algorithm.Python/CustomDataUnderlyingOptionSymbolMappingRegressionAlgorithm.py
```diff
@@ -0,0 +1,71 @@
...
+### <meta name="tag" content="custom data" />
+### <meta name="tag" content="regression test" />
+### <meta name="tag" content="rename event" />
+### <meta name="tag" content="map" />
+### <meta name="tag" content="mapping" />
+### <meta name="tag" content="map files" />
+class CustomDataUnderlyingOptionSymbolMappingRegressionAlgorithm(QCAlgorithm):
+    def Initialize(self):
+        self.SetStartDate(2013, 6, 28)
+        self.SetEndDate(2013, 7, 02)
```
Martin-Molinero: I believe the `0` character should be removed

Algorithm.CSharp/CustomDataUnderlyingSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,190 @@
...
+ *
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Data;
+using QuantConnect.Data.Custom.SEC;
+using QuantConnect.Data.Fundamental;
```
Martin-Molinero: nit - `using QuantConnect.Data.Fundamental;`  can be removed

Algorithm.CSharp/CustomDataUnderlyingOptionSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,147 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect;
```
Martin-Molinero: nit - `using QuantConnect;` can be removed

Algorithm.CSharp/CustomDataUnderlyingOptionSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            if (data.SymbolChangedEvents.Any())
+            {
+                if (data.SymbolChangedEvents.ContainsKey(_customDataOptionSymbol) && data.SymbolChangedEvents.ContainsKey(_optionSymbol))
+                {
+                    var expectedUnderlying = "?FOXA";
+                    var underlying = data.SymbolChangedEvents.Keys.Where(x => x.SecurityType == SecurityType.Base && x == _customDataOptionSymbol).Single().Underlying;
+                    var symbol = data.SymbolChangedEvents.Keys.Where(x => x.SecurityType == SecurityType.Equity && x == _optionSymbol).Single();
+
+                    if (SubscriptionManager.Subscriptions.Where(x => (x.SecurityType == SecurityType.Base || x.SecurityType == SecurityType.Option || x.SecurityType == SecurityType.Equity) && x.MappedSymbol == expectedUnderlying).Count() != 3)
```
Martin-Molinero: nit - Both `Count` and `Single` can accept a `predicate` so the `Where` statement can be removed - Idem for the other csharp algo

Algorithm.CSharp/CustomDataUnderlyingOptionSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,147 @@
...
+                    if (underlying.Underlying != symbol.Underlying)
+                    {
+                        throw new Exception($"Custom data underlying->(2) does match option underlying (equity symbol). Expected {symbol.Underlying.Value} got {underlying.Underlying.Value}");
+                    }
+                    if (underlying.Underlying.Value != expectedUnderlying)
+                    {
+                        throw new Exception($"Custom data symbol value does not match expected value. Expected {expectedUnderlying}, found {underlying.Underlying.Value}");
+                    }
+
+                    SetHoldings(underlying.Underlying, 0.5);
```
Martin-Molinero: Nice 👍

Algorithm.CSharp/CustomDataUnderlyingSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,190 @@
...
+        public override void Initialize()
+        {
+            SetStartDate(2014, 3, 1);
+            SetEndDate(2014, 4, 9);
+            SetCash(100000);
+
+            UniverseSettings.Resolution = Resolution.Daily;
+
+            // Reset the symbol cache to test for ticker adding of custom data
+            SymbolCache.Clear();
```
Martin-Molinero: This shouldn't be required because the `AlgorithmRunner.RunLocalBacktest()` is already clearing the cache

Algorithm.CSharp/CustomDataUnderlyingSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,190 @@
...
+        public override void OnSecuritiesChanged(SecurityChanges changes)
+        {
+            foreach (var added in changes.AddedSecurities.Where(x => x.Symbol.SecurityType == SecurityType.Equity))
+            {
+                // It is in fact "GOOGL" we're catching here, and we're adding it as "GOOG" with the ticker,
+                // which will resolve to GOOCV in the past if we use the ticker and not the symbol
+                if (added.Symbol.Value == "GOOG")
+                {
+                    _badEquitySymbol = added.Symbol;
+                    _badCustomDataSymbol = AddData<SECReport10K>("GOOG").Symbol;
```
Martin-Molinero: What does `bad` mean in `_badCustomDataSymbol/_badEquitySymbol`, I think it could use a more precise name and some comments

Algorithm.CSharp/CustomDataUnderlyingSymbolMappingRegressionAlgorithm.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+        }
+
+        public override void OnSecuritiesChanged(SecurityChanges changes)
+        {
+            foreach (var added in changes.AddedSecurities.Where(x => x.Symbol.SecurityType == SecurityType.Equity))
+            {
+                // It is in fact "GOOGL" we're catching here, and we're adding it as "GOOG" with the ticker,
+                // which will resolve to GOOCV in the past if we use the ticker and not the symbol
+                if (added.Symbol.Value == "GOOG")
```
Martin-Molinero: Do we need this `if` statement, why? If possible, think it would be nice if this regression algorithm showed cased the final user use case

Algorithm.CSharp/CustomDataUnderlyingOptionSymbolMappingRegressionAlgorithm.cs
mchandschuh: fyi - ReSharper helps to automate these kinds of cleanups

Algorithm.CSharp/CustomDataUnderlyingSymbolMappingRegressionAlgorithm.cs
mchandschuh: further -- I don't think we ever want to use `SymbolCache.Clear()` explicitly in our example/regression algorithms as it may lead to users actually using it, which would 99.999% of the time be incorrect.

Tests/Algorithm/AlgorithmAddDataTests.cs
```diff
@@ -171,6 +174,412 @@ public void OnEndOfTimeStepDoesNotThrowWhenSeedsSameUnderlyingForTwoSecurities()
             Assert.AreEqual(data.Price, 2);
         }
 
+        //[TestCase("EURUSD", typeof(PsychSignalSentimentData), SecurityType.Cfd, false, false)]
+        //[TestCase("BTCUSD", typeof(PsychSignalSentimentData), SecurityType.Crypto, false, false)]
+        //[TestCase("CL", typeof(PsychSignalSentimentData), SecurityType.Future, false, false)]
+        //[TestCase("EURUSD", typeof(PsychSignalSentimentData), SecurityType.Forex, false, false)]
```
mchandschuh: Any reason for keeping these commented out test cases?

Tests/Algorithm/AlgorithmAddDataTests.cs
```diff
@@ -171,6 +174,412 @@ public void OnEndOfTimeStepDoesNotThrowWhenSeedsSameUnderlyingForTwoSecurities()
...
+                    asset = qcAlgorithm.AddFuture(ticker, Resolution.Daily);
+                    break;
+                default:
+                    throw new Exception($"SecurityType {securityType} is not valid for this test");
+            }
+
+            // Dummy here is meant to try to corrupt the SymbolCache. Ideally, SymbolCache should return non-custom data types with higher priority
+            // in case we want to add two custom data types, but still have them associated with the equity from the cache if we're using it.
+            // This covers the case where two idential data subscriptions are created.
+            var dummy = qcAlgorithm.AddData(customDataType, asset.Symbol, Resolution.Daily, qcAlgorithm.SubscriptionManager.Subscriptions.Where(x => x.SecurityType == securityType).Single().DataTimeZone);
```
mchandschuh: `Single` can accept a predicate, allowing you to replace the `Where` with `Single`

Martin-Molinero:
Contained by https://github.com/QuantConnect/Lean/pull/3588

PR #3587, opened 2019-09-19
Reverts "Enforce usage of IFormatProvider/CultureInfo.InvariantCulture #3442"

AlexCatarino:
Reverts #3442


jaredbroad:
Seems to be smooth now, can close

PR #3588, opened 2019-09-20
Mapped Alt Data Support w/SID Collision Avoidance

jaredbroad:
### Description
Support for usage of multiple alternative data sources which have the same underlying equity asset. The SID collision avoidance uses the alt-data-type to introduce variance in the ticker and the Symbol.Underlying allows the alt-data ticker to be mapped when the underlying ticker is mapped in corporate events.
Replaces #3484

#### Motivation and Context
Complete alternative data support with corporate actions

Yes; AddData should take a symbol not ticker whenever possible. 

Unit, Regression and Production

#### Types of changes

#### Checklist:


jaredbroad:
Replaces PR #3484

gsalaz98:
BUG:

`20190919 17:35:04.069 Trace:: Debug: Warning: The TWX equity security was set the raw price normalization mode to work with options.`

Will occur when adding with underlying Symbol, most likely due to checking if the symbol has underlying without checking the `SecurityType` of the base and underlying Symbols

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -52,56 +52,130 @@ public void SetPandasConverter()
...
+                var symbol = new Symbol(
+                    SecurityIdentifier.GenerateBase(dataType, ticker, Market.USA, dataType.GetBaseDataInstance().RequiresMapping()),
+                    ticker);
+                return AddDataImpl(dataType, symbol, resolution, timeZone, fillDataForward, leverage);
+            }
+            // If we need a mappable ticker and we can't find one in the SymbolCache, throw
+            Symbol underlying;
+            if (!SymbolCache.TryGetSymbol(ticker, out underlying))
+            {
+                throw new InvalidOperationException("Requires mapping but passed ticker which is not in the cache");
```
Martin-Molinero: I think we can further improve this error message, maybe something like: `The custom data type {dataType.Name} requires mapping but the provided ticker is not in the cache. Please add this custom data type using the related underlying Symbol instance or perform this call after the underlying Security has been added using AddSecurity() methods. An example use case can be found at {nameof(RegressionAlgorithm)}`. We want the users to understand and be able to fix the exception by themselves if possible.

Algorithm/QCAlgorithm.Python.cs
```diff
@@ -52,56 +52,130 @@ public void SetPandasConverter()
...
+
+        /// <summary>
+        /// AddData a new user defined data source, requiring only the minimum config options.
+        /// The data is added with a default time zone of NewYork (Eastern Daylight Savings Time)
+        /// </summary>
+        /// <param name="type">Data source type</param>
+        /// <param name="underlying">The underlying symbol for the custom data</param>
+        /// <param name="resolution">Resolution of the data</param>
+        /// <returns>The new <see cref="Security"/></returns>
+        public Security AddData(PyObject type, Symbol underlying, Resolution resolution = Resolution.Minute, object _ = null, object __ = null, object ___ = null)
```
Martin-Molinero: I think we need to add some documentation on when to call and what's the different use cases, for each different `AddData()` specifically the `Symbol vs string` implementations. Maybe a `<remarks>` for these new methods

Algorithm.CSharp/CustomDataAddDataCoarseSelectionRegressionAlgorithm.cs
```diff
@@ -0,0 +1,117 @@
...
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Data;
+using QuantConnect.Data.Custom.SEC;
+using QuantConnect.Data.Custom.USTreasury;
+using QuantConnect.Data.Fundamental;
```
Martin-Molinero: nit - `using QuantConnect.Data.Custom.USTreasury; using QuantConnect.Data.Fundamental;` are not required. As @mchandschuh pointed out this can be detected easily by ReSharper

Algorithm.CSharp/CustomDataAddDataCoarseSelectionRegressionAlgorithm.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            UniverseSettings.Resolution = Resolution.Daily;
+
+            AddUniverseSelection(new CoarseFundamentalUniverseSelectionModel(CoarseSelector));
+        }
+
+        public IEnumerable<Symbol> CoarseSelector(IEnumerable<CoarseFundamental> coarse)
+        {
+            var symbols = coarse.Where(x => x.HasFundamentalData && x.DollarVolume > 500000000)
+                .Select(x => x.Symbol);
```
Martin-Molinero: I think it might be better to return a fixed set of Symbols for which we have data and perform some trade. As is `OnData()` is not being called, so nothing is really being asserted here
Also I think there is value on adding another regression test almost identical except it addes the custom assets in the `QCAlgorithm.OnSecuritiesChanged` callback, so we can assert and document behavior, users will most probably use both methods

Common/Symbol.cs
```diff
@@ -454,6 +483,7 @@ public bool Equals(Symbol other)
             SecurityIdentifier sid;
             if (SecurityIdentifier.TryParse(ticker, out sid))
             {
+                Logging.Log.Trace($"Lifted ticker {ticker} - {sid.Symbol}");
```
Martin-Molinero: Is this a log we want for production? Thinking it might get verbose - nit: also it's not following current Lean logging conventions, missing scope, maybe `Symbol.ImplicitStringToSymbol()`?

Tests/Algorithm/AlgorithmAddDataTests.cs
```diff
@@ -171,6 +174,310 @@ public void OnEndOfTimeStepDoesNotThrowWhenSeedsSameUnderlyingForTwoSecurities()
...
+        [TestCase("BTCUSD", typeof(PsychSignalSentimentData), SecurityType.Crypto, false, true)]
+        [TestCase("CL", typeof(PsychSignalSentimentData), SecurityType.Future, false, true)]
+        [TestCase("EURUSD", typeof(PsychSignalSentimentData), SecurityType.Forex, false, true)]
+        [TestCase("AAPL", typeof(PsychSignalSentimentData), SecurityType.Equity, true, true)]
+        [TestCase("EURUSD", typeof(TradingEconomicsCalendar), SecurityType.Cfd, false, false)]
+        [TestCase("BTCUSD", typeof(TradingEconomicsCalendar), SecurityType.Crypto, false, false)]
+        [TestCase("CL", typeof(TradingEconomicsCalendar), SecurityType.Future, false, false)]
+        [TestCase("AAPL", typeof(TradingEconomicsCalendar), SecurityType.Equity, true, false)]
+        [TestCase("EURUSD", typeof(TradingEconomicsCalendar), SecurityType.Forex, false, false)]
+        public void AddDataSecuritySymbolWithUnderlying(string ticker, Type customDataType, SecurityType securityType, bool securityShouldBeMapped, bool customDataShouldBeMapped)
```
Martin-Molinero: For these new unit tests maybe `customDataShouldBeMapped` parameter could be retrieved directly from the custom data type since that's why Lean is going to be doing using `dataType.GetBaseDataInstance().RequiresMapping()`

Common/Symbol.cs
```diff
@@ -85,6 +86,34 @@ public static Symbol Create(string ticker, SecurityType securityType, string mar
...
 
+        /// <summary>
+        /// Creates a new Symbol for custom data. This method allows for the creation of a new Base Symbol
+        /// using the first ticker and the first traded date from the provided underlying Symbol. This avoids
+        /// the issue for mappable types, where the ticker is remapped supposing the provided ticker value is from today.
+        /// See method <see cref="SecurityIdentifier.GetFirstTickerAndDate(Interfaces.IMapFileProvider, string, string)"/>
+        /// The provided symbol is also set to <see cref="Symbol.Underlying"/> so that it can be accessed using the custom data Symbol.
+        /// This is useful for associating custom data Symbols to other asset classes so that it is possible to filter using custom data
+        /// and place trades on the underlying asset based on the filtered custom data.
+        /// </summary>
```
Martin-Molinero: Nice summary! 👍

Algorithm.CSharp/CustomDataUsingMapFileRegressionAlgorithm.cs
```diff
@@ -46,7 +46,8 @@ public override void Initialize()
             SetStartDate(2013, 06, 27);
             SetEndDate(2013, 07, 02);
 
-            _symbol = AddData<CustomDataUsingMapping>("FOXA").Symbol;
+            var foxa = QuantConnect.Symbol.Create("FOXA", SecurityType.Equity, Market.USA);
+            _symbol = AddData<CustomDataUsingMapping>(foxa).Symbol;
```
mchandschuh: this seems wrong -- we create a `Symbol` object and then we pass it to a method that accepts a string and then pull the generated symbol.... why define `foxa` at all? Does the string `"FOXA"` not work?

Algorithm.CSharp/DropboxBaseDataUniverseSelectionAlgorithm.cs
```diff
@@ -30,8 +30,10 @@ namespace QuantConnect.Algorithm.CSharp
     /// <meta name="tag" content="using data" />
     /// <meta name="tag" content="universes" />
     /// <meta name="tag" content="custom universes" />
-    public class DropboxBaseDataUniverseSelectionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    public class DropboxBaseDataUniverseSelectionAlgorithm : QCAlgorithm//, IRegressionAlgorithmDefinition
     {
+        // Regression algorithm disabled due to dropbox file missing
```
mchandschuh: I got a response -- they said my account has been deleted along with all of the files :(

Algorithm.CSharp/EstimizeDataAlgorithm.cs
```diff
@@ -34,9 +34,9 @@ public override void Initialize()
             SetStartDate(2017, 1, 1);
             SetEndDate(2017, 12, 31);
 
-            AddData<EstimizeRelease>("AAPL.R");
-            AddData<EstimizeEstimate>("AAPL.E");
-            AddData<EstimizeConsensus>("AAPL.C");
+            AddData<EstimizeRelease>("AAPL");
+            AddData<EstimizeEstimate>("AAPL");
```
mchandschuh: this looks  a wee-bit nicer 👍

Algorithm.Framework/Portfolio/BlackLittermanOptimizationPortfolioConstructionModel.py
```diff
@@ -192,7 +192,7 @@ def OnSecuritiesChanged(self, algorithm, changes):
             if not history.empty:
                 ticker = SymbolCache.GetTicker(symbol)
 
-                if ticker not in history.index.levels[0]:
+                if ticker  not in history.index.levels[0]:
```
mchandschuh: Glad this is working now -- looks like there's a double space following `ticker`

Algorithm/QCAlgorithm.Python.cs
mchandschuh: yea... and what's the deal with the ignored parameters? This is confusing to me, so definitely agree we need some `<remarks>` describing the reasoning behind this

Algorithm/QCAlgorithm.Python.cs
mchandschuh: Although, you won't be able to use `nameof(RegressionAlgorithm)` so a string literal will suffice (since its in the `Algorithm.CSharp` which isn't referenced via `Algorithm`)

Common/Data/Auxiliary/MappingExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        public static MapFile ResolveMapFile(this MapFileResolver mapFileResolver,
+            Symbol symbol,
+            Type dataType)
+        {
+            // Load the symbol and date to complete the mapFile checks in one statement
+            var symbolID = symbol.HasUnderlying ? symbol.Underlying.ID.Symbol : symbol.ID.Symbol;
+            var date = symbol.HasUnderlying ? symbol.Underlying.ID.Date : symbol.ID.Date;
+
+            return mapFileResolver.ResolveMapFile(
+                symbol.SecurityType == SecurityType.Base ? symbolID.RemoveFromEnd($".{dataType.Name}") : symbolID,
```
mchandschuh: style nit -- in general, I wouldn't embed ternary expressions especially when they get this long, just makes reading the code harder -- more legible to extract it into a variable and simply pass it into the function.

Common/Data/Custom/Estimize/EstimizeConsensus.cs
```diff
@@ -191,6 +183,15 @@ public override string ToString()
...
                    Invariant($"by {Source}");
         }
+
+        /// <summary>
+        /// Indicates if there is support for mapping
+        /// </summary>
+        /// <returns>True indicates mapping should be used</returns>
+        public override bool RequiresMapping()
+        {
+            return true;
```
mchandschuh: I do like this -- types can declare that they require mapping themselves -- right where it belongs 👍

Common/Python/PandasData.cs
```diff
@@ -60,6 +61,174 @@ public PandasData(object data)
...
+                    // this python Remapper class will work as a proxy and adjust the
+                    // input to its methods using the provided 'mapper' callable object
+                    _remapperFactory = PythonEngine.ModuleFromString("remapper",
+                        @"import wrapt
+import pandas
+from clr import AddReference
+AddReference(""QuantConnect.Common"")
+from QuantConnect import *
+
+class Remapper(wrapt.ObjectProxy):
```
mchandschuh: Glad this solution worked out!!

Common/Symbol.cs
mchandschuh: Good spot -- I'm gussing not.

gsalaz98:
New commit includes two new regression algorithms (CustomDataAddDataOnSecuritiesChangedRegressionAlgorithm) that throws the following exception when ran:

```
20190923 14:24:33.309 ERROR:: <>c__DisplayClass9_2.<Run>b__9():  System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at QuantConnect.Algorithm.QCAlgorithm.OnEndOfTimeStep() in C:\Users\Gerardo\Projects\QuantConnect\Lean\Algorithm\QCAlgorithm.Universe.cs:line 150
   at QuantConnect.Lean.Engine.AlgorithmManager.Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer synchronizer, ITransactionHandler transactions, IResultHandler results, IRealTimeHandler realtime, ILeanManager leanManager, IAlphaHandler alphas, CancellationToken token) in C:\Users\Gerardo\Projects\QuantConnect\Lean\Engine\AlgorithmManager.cs:line 695
   at QuantConnect.Lean.Engine.Engine.<>c__DisplayClass9_2.<Run>b__9() in C:\Users\Gerardo\Projects\QuantConnect\Lean\Engine\Engine.cs:line 328

20190923 14:24:33.394 ERROR:: Engine.HandleAlgorithmError(): Breaking out of parent try catch: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at QuantConnect.Algorithm.QCAlgorithm.OnEndOfTimeStep() in C:\Users\Gerardo\Projects\QuantConnect\Lean\Algorithm\QCAlgorithm.Universe.cs:line 150
   at QuantConnect.Lean.Engine.AlgorithmManager.Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer synchronizer, ITransactionHandler transactions, IResultHandler results, IRealTimeHandler realtime, ILeanManager leanManager, IAlphaHandler alphas, CancellationToken token) in C:\Users\Gerardo\Projects\QuantConnect\Lean\Engine\AlgorithmManager.cs:line 695
   at QuantConnect.Lean.Engine.Engine.<>c__DisplayClass9_2.<Run>b__9() in C:\Users\Gerardo\Projects\QuantConnect\Lean\Engine\Engine.cs:line 328
```

Issue #3603, opened 2019-09-24
Use thread-safe collection to manage consolidators


mchandschuh:
#### Expected Behavior

Exceptions shouldn't be thrown due to removing a consolidator in a consolidator's update callback event. This issue was raised here: https://github.com/QuantConnect/Lean/pull/3573#discussion_r327811944

#### Actual Behavior

Exceptions thrown when removing a consolidator via it's `Updated` event handler.

#### Potential Solution

Use a thread-safe and set-based collection

#### Reproducing the Problem

See PR review comment link above.

#### System Information

n/a

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3605, opened 2019-09-24
Use thread safe consolidator collection in SubscriptionDataConfig

mchandschuh:
#### Description

* Adds `ConcurrentSet<T>`
* Uses `ConcurrentSet<T>` for `SubscriptionDataConfig.Consolidators`

#### Related Issue

See #3603

#### Motivation and Context

An exception is thrown when removing a consolidator from inside it's `Updated` event handler.

No.

Unit tests added for `ConcurrentSet<T>` and to ensure we can remove consolidators while enumerating the consolidator collection.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Util/ConcurrentSet.cs
+++ b/Common/Util/ConcurrentSet.cs
@@ -1,0 +1,307 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Util
+{
+    /// <summary>
+    /// Provides a thread-safe set collection that mimics the behavior of <see cref="HashSet{T}"/>
+    /// </summary>
+    /// <typeparam name="T">The item type</typeparam>
+    public class ConcurrentSet<T> : ISet<T>
+    {
+        private readonly ConcurrentDictionary<T, T> _set = new ConcurrentDictionary<T, T>();
+
+        /// <summary>Gets the number of elements contained in the <see cref="T:System.Collections.Generic.ICollection`1" />.</summary>
+        /// <returns>The number of elements contained in the <see cref="T:System.Collections.Generic.ICollection`1" />.</returns>
+        public int Count => _set.Count;
+
+        /// <summary>Gets a value indicating whether the <see cref="T:System.Collections.Generic.ICollection`1" /> is read-only.</summary>
+        /// <returns>true if the <see cref="T:System.Collections.Generic.ICollection`1" /> is read-only; otherwise, false.</returns>
+        public bool IsReadOnly => false;
+
+        /// <summary>Adds an item to the <see cref="T:System.Collections.Generic.ICollection`1" />.</summary>
+        /// <param name="item">The object to add to the <see cref="T:System.Collections.Generic.ICollection`1" />.</param>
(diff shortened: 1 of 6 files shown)


Tests/Common/Securities/SubscriptionDataConfigTests.cs
```diff
@@ -40,5 +41,19 @@ public void UsedAsDictionaryKey()
...
+        public void CanRemoveConsolidatorWhileEnumeratingList()
+        {
+            var config = new SubscriptionDataConfig(typeof(TradeBar), Symbols.SPY, Resolution.Minute, TimeZones.NewYork, TimeZones.NewYork, false, false, false, false, TickType.Trade, false);
+            var consolidator1 = new TradeBarConsolidator(1);
+            var consolidator2 = new TradeBarConsolidator(2);
+            config.Consolidators.Add(consolidator1);
+            config.Consolidators.Add(consolidator2);
+            foreach (var consolidator in config.Consolidators)
+            {
+                Assert.DoesNotThrow(() => config.Consolidators.Remove(consolidator));
```
Martin-Molinero: I think the `throw` would happen in the `foreach` loop 🤔

Common/Util/ConcurrentSet.cs
```diff
@@ -0,0 +1,227 @@
...
+
+namespace QuantConnect.Util
+{
+    /// <summary>
+    /// Provides a thread-safe set collection that mimics the behavior of <see cref="HashSet{T}"/>
+    /// </summary>
+    /// <typeparam name="T">The item type</typeparam>
+    public class ConcurrentSet<T> : ISet<T>
+    {
+        private readonly ConcurrentDictionary<T, T> _set = new ConcurrentDictionary<T, T>();
```
Martin-Molinero: Think it might be cheaper memory-wise to just make the value a `byte`?

Common/Util/ConcurrentSet.cs
```diff
@@ -0,0 +1,227 @@
...
+
+        public int Count => _set.Count;
+        public bool IsReadOnly => false;
+
+        public IEnumerator<T> GetEnumerator()
```
Martin-Molinero: nit - Some basic `<summary>` documentation for these methods would be great

Common/Util/ConcurrentSet.cs
mchandschuh: I'm not sure that it's simple as there's struct alignment issues to consider, for example, when instantiating the `KeyValuePair<T, byte>` instances for enumeration. Also, if using `byte` we'll be subject to boxing/unboxing semantics since it's a value type (also when creating `KeyValuePair<T, byte>` instances during enumeration). The internal `ConcurrentDictionary<TKey, TValue>.Node` type _is_ a class, so IIRC, it isn't subject to the same alignment concerns as KVP, but we often enumerate this type which instantiates KVP instances, so it's worth noting.

Further, the potential difference in memory required, assuming the `byte` actually gets stored as a single byte vs a 4-byte dword (8-byte on when x64, obj reference), would be a maximum of 3 (7) bytes per entry, so even at 1M entries, we're still under 3MB (7MB) of memory saved.  Either way, I think 3 (or 7) bytes per entry is infinitesimally smaller than the actual object being stored making the savings practically negligible. Now, if it _does_ store it internally as a `byte`, then we're guaranteed to get the performance hit of boxing/unboxing the byte -- which I think is the bigger thing to balance vs a few KBs (in reality algos don't have more than a few hundred, maybe a thousand consolidators).

I would prefer to leave this as-is without evidence supporting it as an issue.

Common/Util/ConcurrentSet.cs
mchandschuh: agreed -- will copy docs from `ISet<T>` interface documentation.

Tests/Common/Securities/SubscriptionDataConfigTests.cs
mchandschuh: Correct -- before the change this test fails immediately when it executes line 55. Following this change no exception is thrown. Maybe I'm misunderstanding your comment? :)

Martin-Molinero: My point was that I think it throws on line 53, not 55? don't think that `config.Consolidators.Remove(consolidator)` will throw - nit type of comment 😄

Common/Util/ConcurrentSet.cs
Martin-Molinero: Yeah, good points there will never be that amount of consolidators where this matters, but it might add an overhead 👍

Tests/Common/Securities/SubscriptionDataConfigTests.cs
mchandschuh: ohh, it throws on ln55 w/out this fix due to modifying the collection via `Consolidators.Remove(consolidator)`

Issue #3618, opened 2019-09-26
Add sparse data flag to BaseData


mchandschuh:
#### Expected Behavior

LEAN doesn't spam messages of missing data on sparse data sets.

#### Actual Behavior

LEAN spams messages of missing data on sparse data sets.

#### Potential Solution

Custom data types such as `PsychSignal` can declare themselves as having a sparse data set. We'll initially be using this flag to prevent spamming log messages when data is not found. Consider collapsing this flag with the `RequiresMapping` flag into a `[Flags]` enum and exposing it as a single  property on `BaseData`, such as `BaseData.Options` or similar.

#### Reproducing the Problem

Run any derivative alternative data, such as `PsychSignal`

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

mchandschuh:
NOTE: Ignoring collapsing `RequiresMapping` and `IsSparseData` flags into an enum to reduce risk of rolling this out. We can look at cleaning that up next week.

mchandschuh:
Closed in 6ca47e010dd6507fdbb9b88c7b68f784721da1d4

PR #3623, opened 2019-09-26
Feature 3618 add sparse data flag

mchandschuh:
#### Description

Adds a new `BaseData.IsSparseData()` flag method. We're currently only using this flag to control whether or not to emit file not found errors. By default, all custom data is set to be sparse data and can be overridden where appropriate.

NOTE: I've left this as a method `IsSparseData()` but think it would be beneficial to look into collapsing these _options_ into a `[Flags]` enum. Also, leaving as a method over concerns about python being able to easily override properties.

#### Related Issue

See #3618

#### Motivation and Context

Prevents log spam which causes slow backtests.

No.

Existing unit tests cover these changes fairly well and it's hard to directly test actions taken against `Log`. I manually ran algorithms with invalid symbols to observe generated error messages. I confirmed that for default data (equity/etc) we get the error messages as desired but for custom data we do not get the error messages.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -141,11 +141,12 @@ namespace QuantConnect.Algorithm
         /// <returns>The new <see cref="Security"/></returns>
         public Security AddData(Type dataType, string ticker, Resolution resolution, DateTimeZone timeZone, bool fillDataForward = false, decimal leverage = 1.0m)
         {
+            // NOTE: Invoking methods on BaseData w/out setting the symbol may provide unexpected behavior
             var baseInstance = dataType.GetBaseDataInstance();
             if (!baseInstance.RequiresMapping())
             {
                 var symbol = new Symbol(
-                    SecurityIdentifier.GenerateBase(dataType, ticker, Market.USA, dataType.GetBaseDataInstance().RequiresMapping()),
+                    SecurityIdentifier.GenerateBase(dataType, ticker, Market.USA, baseInstance.RequiresMapping()),
                     ticker);
                 return AddDataImpl(dataType, symbol, resolution, timeZone, fillDataForward, leverage);
             }
--- a/Common/Data/BaseData.cs
+++ b/Common/Data/BaseData.cs
@@ -150,6 +150,16 @@ namespace QuantConnect.Data
         }
 
         /// <summary>
+        /// Indicates that the data set is expected to be sparse
+        /// </summary>
+        /// <returns>True if the data set represented by this type is expected to be sparse</returns>
+        public virtual bool IsSparseData()
+        {
+            // by default, we'll assume all custom data is sparse data
+            return Symbol.SecurityType == SecurityType.Base;
+        }
+
+        /// <summary>
         /// Updates this base data with a new trade
         /// </summary>
         /// <param name="lastTrade">The price of the last trade</param>
--- a/Common/Data/BaseData.cs
+++ b/Common/Data/BaseData.cs
@@ -27,51 +27,24 @@ namespace QuantConnect.Data
     /// </summary>
     public abstract class BaseData : IBaseData
     {
-        private MarketDataType _dataType = MarketDataType.Base;
-        private DateTime _time;
-        private Symbol _symbol = Symbol.Empty;
         private decimal _value;
-        private bool _isFillForward;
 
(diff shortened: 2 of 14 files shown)


Algorithm/QCAlgorithm.Python.cs
```diff
@@ -141,11 +141,12 @@ public Security AddData(PyObject type, Symbol underlying, Resolution resolution,
         /// <returns>The new <see cref="Security"/></returns>
         public Security AddData(Type dataType, string ticker, Resolution resolution, DateTimeZone timeZone, bool fillDataForward = false, decimal leverage = 1.0m)
         {
+            // NOTE: Invoking methods on BaseData w/out setting the symbol may provide unexpected behavior
```
mchandschuh: I didn't want to change behavior here, but left a note since some implementations of `RequiresMapping()` require the symbol object to be set.

Common/Data/BaseData.cs
```diff
@@ -27,76 +27,39 @@ namespace QuantConnect.Data
...
-            get
-            {
-                return _dataType;
-            }
-            set
-            {
-                _dataType = value;
-            }
-        }
+        public MarketDataType DataType { get; set; } = MarketDataType.Base;
```
mchandschuh: ReSharper clean up of auto-properties.

Common/Data/BaseData.cs
```diff
@@ -192,6 +149,16 @@ public virtual bool RequiresMapping()
...
         }
 
+        /// <summary>
+        /// Indicates that the data set is expected to be sparse
+        /// </summary>
+        /// <returns>True if the data set represented by this type is expected to be sparse</returns>
+        public virtual bool IsSparseData()
+        {
+            // by default, we'll assume all custom data is sparse data
+            return Symbol.SecurityType == SecurityType.Base;
```
mchandschuh: By default, we'll say all custom data is sparse.

Common/Data/SubscriptionDataConfigExtensions.cs
```diff
@@ -116,10 +116,18 @@ public static class SubscriptionDataConfigExtensions
...
             // we create an instance of the data type, if it is a custom type
-            // it can override RequiresMapping else it will use security type
+            // it can override RequiresMapping else it will use security type\
+            return config.GetBaseDataInstance().RequiresMapping();
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BaseData"/> type defined in <paramref name="config"/> with the symbol properly set
+        /// </summary>
+        public static BaseData GetBaseDataInstance(this SubscriptionDataConfig config)
```
mchandschuh: As part of this PR, I noticed some cases where we're creating the base data instance but not setting the symbol property. Since methods like `RequiresMapping()` and `IsSparseData()` can depend on symbol information (specifically security type), I went through and updated the easy cases to use this new method and the appropriate `BaseData` method directly. This also had the side effect of removing a few unnecessary `Activator` calls for newing up base data instances we didn't need.

Engine/DataFeeds/SubscriptionDataReader.cs
```diff
@@ -448,16 +448,14 @@ private ISubscriptionDataSourceReader CreateSubscriptionFactory(SubscriptionData
 
         private void AttachEventHandlers(ISubscriptionDataSourceReader dataSourceReader, SubscriptionDataSource source)
         {
+            // NOTE: There seems to be some overlap in InvalidSource and CreateStreamReaderError
+            //       this may be worthy of further investigation and potential consolidation of events.
```
mchandschuh: Originally had the logging here but leaving this note to posterity. I suspect `InvalidSource` and `CreateStreamReaderError` may be able to be combined and properly abstracted vs needing to cast to a `TextSubscriptionDataSourceReader` in order to access the events.

Common/Data/SubscriptionDataConfigExtensions.cs
Martin-Molinero: Nice! 👍

Algorithm/QCAlgorithm.Python.cs
Martin-Molinero: Good catch - maybe its worth adding this as a `remarks` for the `BaseData` methods `RequiresMapping/IsSparseData`

PR #3631, opened 2019-09-30
Add DynamicSecurityData and cache derivative custom data in underlying

### LINKED ISSUE #3620: Cache custom derivative data into the underlying security object and provide accessor
#### Expected Behavior

Easy to access derivative data from the underlying security object.

#### Actual Behavior

Users currently need to manage their own maps of underlying symbols to custom data symbols in order to resolve the custom data available for a particular underlying. For example, we should easily be able to access `SPY.PsychSignal` from the `SPY` security object.

#### Potential Solution

Push derivative data into the underlying security cache to make it available. This makes it available via security cache's `Get<T>()` method via `security.Cache.Get<PsychSignal>()`. We can add another `Get<T>()` to `Security` ([dementer](https://en.wikipedia.org/wiki/Law_of_Demeter) would be proud) and we can also implement a `dynamic` harness via `Security : IDynamicMetaObjectProvider` using the custom data type name as the accessor, for example, the following should work: `Securities["SPY"].PsychSignal` => `List<PsychSignal>` -- sadly it must be a list due to 1-n relationships.

**NOTE:** For now we're just applying this to custom data but we should also investigate doing this for _all_ securities w/ underlying (options/etc).

#### Reproducing the Problem

Try implementing an alt data algorithm and you'll feel the burn.

#### Checklist

mchandschuh:
NOTE: Many changes are due to updating `Security`'s constructor, as well as `DataManager` and `SecurityService` to accept the new `IRegisteredSecurityDataTypesProvider`. The `DynamicSecurityData` implementation uses this to determine whether or not we can return a default value of an empty list (of the correct type) in case we haven't received data yet. This is as opposed to throwing a `KeyNotFoundException` immediately. See `DynamicSecurityData.GetProperty(name)` for implementation details.

#### Description

Adds a new system of accessing data related to a particular security. This logic is encapsulated in the dynamic implementation of `DynamicSecurityData` and is accessible via the `security.Data` property. `DynamicSecurityData` provides a type-centric view of the security's cache which provides access to data by type. There's two separate access patterns here. The first is type-safe via C# generics:
``` csharp
Security security = Securities["GOOGL"];

// returns a single item of type T
TradeBar mostRecentTradeBar = security.Data.Get<TradeBar>();

// returns a list of type T, useful for data types where multiple data points
// may be available at a single time step, such as ticks
List<Tick> mostRecentTicks = security.Data.GetAll<Tick>();

// you can equally access derivative custom data, such as from AddData<SECReport8K>("GOOGL")
SECReport8K sec8k = security.Get<SECReport8K>();
```
We can also access the list version (`GetAll<T>`) via the usage of dynamics. This is also the recommended access pattern for python:
``` csharp
var security = Securities["GOOGL"];

// access dynamically by type name: security.Data.TypeName
List<TradeBar> mostRecentTradeBars = security.Data.TradeBar;
List<Tick> mostRecentTicks = security.Data.Tick;

// derivative custom data is also now accessible via the underlying security
// for example, if using AddData<SECReport8K>("GOOGL")
// the dynamic accessors will ALWAYS return a list, and if no data available,
// it will return an empty list provided it's a valid and registered type name
List<SECReport8K> sec8k = security.Data.SECReport8K;
```

In order to support returning the default value of an empty list in the case where we haven't received data yet, a new abstraction was added: `IRegisteredSecurityDataTypesProvider`. There's a single instance of the registered type provider per algorithm and it is owned by the `ISecurityService` since he's in charge of creating security objects. When creating securities, the data type from the configuration is added to a set to track all unique data types for the algorithm. When the algorithm attempts to access `security.Data.InvalidPropertyName`, we'll first check to see if `InvalidPropertyName` is one of the registered type names, and if so, we'll return a new empty list, and if not, we'll throw a `KeyNotFoundException`.

In addition to the security data accessors detailed above, this PR also includes several enhancements to the local regression test/debug system. Currently, we write log and order files to `./passed` and `./regression`. The passed folder is for (you guessed it!) algorithms that pass and regression contains the most recent run. This allows you to do a directory diff between `./passed` and `./regression`. This is an incredibly useful aid when trying to determine the root cause of a regression test failure. While working this issue I found that not enough data was being put into the `orders.log` files. This PR changes it to a `details.log` and saves off way more data, including daily snapshots of holdings quantity/price/value, cashbook statements, alpha runtime stats as they change and even logging of every piece of data (exception option/future contracts) that comes through in each slice object.

This makes identifying the underlying cause of a regression failure as easy as popping open a diff window :)

#### Related Issue

Closes #3620

#### Motivation and Context

Previously, users would need to manage their own mapping of derivative custom data (such as SEC report, tiingo, psychsignal, etc) to the security they were describing (underlying). This removes all of that and makes all of a security's data immediately accessible directly from the underlying security itself.

Yes!

Unit tests added for `DynamicSecurityData` type and a test added to `SecurityTests` to ensure the `SecurityCache`'s event handler is wired up properly.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Market/Tick.cs
+++ b/Common/Data/Market/Tick.cs
@@ -429,10 +429,31 @@ namespace QuantConnect.Data.Market
             return new Tick(this);
         }
 
+        /// <summary>
+        /// Formats a string with the symbol and value.
+        /// </summary>
+        /// <returns>string - a string formatted as SPY: 167.753</returns>
+        public override string ToString()
+        {
+            switch (TickType)
+            {
+                case TickType.Trade:
+                    return $"{Symbol}: Price: {Price} Quantity: {Quantity}";
+
+                case TickType.Quote:
+                    return $"{Symbol}: Bid: {BidSize}@{BidPrice} Ask: {AskSize}@{AskPrice}";
+
+                case TickType.OpenInterest:
+                    return $"{Symbol}: OpenInterest: {Value}";
+
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+        }
+
         private static decimal GetScaleFactor(SecurityType securityType)
         {
             return securityType == SecurityType.Equity || securityType == SecurityType.Option ? 10000m : 1;
         }
-
-    } // End Tick Class:
+    }
 }
--- a/Common/Securities/DynamicSecurityData.cs
+++ b/Common/Securities/DynamicSecurityData.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Dynamic;
 using System.Linq;
@@ -32,7 +33,17 @@ namespace QuantConnect.Securities
         private static readonly MethodInfo SetPropertyMethodInfo = typeof(DynamicSecurityData).GetMethod("SetProperty");
         private static readonly MethodInfo GetPropertyMethodInfo = typeof(DynamicSecurityData).GetMethod("GetProperty");
 
+        private readonly IRegisteredSecurityDataTypesProvider _registeredTypes;
         private readonly Dictionary<string, object> _storage = new Dictionary<string, object>();
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DynamicSecurityData"/> class
+        /// </summary>
+        /// <param name="registeredTypes">Provides all the registered data types for the algorithm</param>
+        public DynamicSecurityData(IRegisteredSecurityDataTypesProvider registeredTypes)
+        {
+            _registeredTypes = registeredTypes;
+        }
 
         /// <summary>Returns the <see cref="T:System.Dynamic.DynamicMetaObject" /> responsible for binding operations performed on this object.</summary>
         /// <returns>The <see cref="T:System.Dynamic.DynamicMetaObject" /> to bind this object.</returns>
@@ -72,7 +83,23 @@ namespace QuantConnect.Securities
         public void StoreData<T>(Type dataType, IReadOnlyList<T> data)
         {
             // this would, for example, be 'Bitcoin' or 'TradeBar'
-            SetProperty(dataType.Name, data);
+            if (typeof(T) == dataType)
+            {
+                SetProperty(dataType.Name, data);
+                return;
+            }
+
+            // common case where it's a List<BaseData> but dataType == TradeBar
+            // create a List<TradeBar> so that when accessed via GetAll<T> or .TradeBar
+            // the types line up as expected
+            var list = (IList) Activator.CreateInstance(typeof(List<>).MakeGenericType(dataType));
+            foreach (var datum in data)
+            {
+                // if the element type and dataType aren't in alignment we'll get an invalid cast exception here
+                list.Add(datum);
+            }
+
+            SetProperty(dataType.Name, list);
         }
 
         /// <summary>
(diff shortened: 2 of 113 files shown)


Common/Data/Market/Tick.cs
```diff
@@ -429,10 +429,31 @@ public override BaseData Clone()
...
+            }
+        }
+
         private static decimal GetScaleFactor(SecurityType securityType)
         {
             return securityType == SecurityType.Equity || securityType == SecurityType.Option ? 10000m : 1;
         }
-
-    } // End Tick Class:
+    }
```
mchandschuh: hehe @jaredbroad -- I'm surprised there's still some of these `// End XXX Class:` comments, I thought I got them all years ago :P

Common/Securities/DynamicSecurityData.cs
```diff
@@ -0,0 +1,182 @@
...
+            if (_storage.TryGetValue(name, out value))
+            {
+                return value;
+            }
+
+            // check to see if the requested name matches one of the algorithm registered data types and if
+            // so, we'll return a new empty list. this precludes us from always needing to check HasData<T>
+            foreach (var type in _registeredTypes.GetRegisteredDataTypes())
+            {
+                if (type.Name == name)
```
mchandschuh: Here's where we enforce default value semantics. Consider the first time step and assume we haven't received an `TradeBar` yet for a particular security, but we've requested _at least_ one `TradeBar` subscription already. If the algo does `security.Data.TradeBar` it will catch at this condition and we'll use activator to create a new empty `List<TradeBar>` even though there's technically nothing here. We figured this default value behavior was better than throwing `KeyNotFoundException` -- thanks @jaredbroad for thinking of this case 👍

Common/Securities/IRegisteredSecurityDataTypesProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Provides the set of base data types registered in the algorithm
+    /// </summary>
+    public interface IRegisteredSecurityDataTypesProvider
```
mchandschuh: This guy is like a glorified `HashSet<Type>`, but I wanted to encapsulate the intent behind some explicit type names. We call `RegisterType` via `ISubscriptionDataConfigService.Add` (which is implemented via `DataManager.Add`). All subscription configs come through this point, so we take `config.Type` and register it with this provider. This gives us a _whitelist_ of types that the `DynamicSecurityData` implementation uses for determining whether or not to perform the empty list default value behavior.

We currently aren't called `UnregisterType`, but it felt wrong/incomplete to not include it.

Common/Securities/DynamicSecurityData.cs
```diff
@@ -0,0 +1,182 @@
...
+    /// Provides access to a security's data via it's type. This implementation supports dynamic access
+    /// by type name.
+    /// </summary>
+    public class DynamicSecurityData : IDynamicMetaObjectProvider
+    {
+        private static readonly MethodInfo SetPropertyMethodInfo = typeof(DynamicSecurityData).GetMethod("SetProperty");
+        private static readonly MethodInfo GetPropertyMethodInfo = typeof(DynamicSecurityData).GetMethod("GetProperty");
+
+        private readonly IRegisteredSecurityDataTypesProvider _registeredTypes;
+        private readonly Dictionary<string, object> _storage = new Dictionary<string, object>();
```
Martin-Molinero: `SecurityCache` is concurrent maybe this collection should be too?

Common/Securities/Security.cs
```diff
@@ -559,6 +586,24 @@ public void SetMarketPrice(BaseData data)
...
+        public void Update(IEnumerable<BaseData> data)
+        {
+            foreach (var grp in data.GroupBy(d => d.GetType()))
+            {
+                // use the last of each type to set market price
+                SetMarketPrice(grp.Last());
+
+                // maintaining regression requires us to NOT cache FF data
+                var nonFillForward = grp.Where(d => !d.IsFillForward).ToList();
+                Cache.StoreData(nonFillForward.ToList());
```
Martin-Molinero: nit - Double `ToList()`

PR #3688, opened 2019-10-08
Adds Message Warning For Event Frequency with Incompatible Resolution

### LINKED ISSUE #3687: Improves Warning Message for Scheduled Events
#### Expected Behavior
Notify users that Scheduled Events needs data with a proper resolution to be triggered at desired moments.

#### Actual Behavior
No message. 

#### Potential Solution
In `TimeRules` class, verify whether the time rule interval and the security highest resolution are compatible. If not, log a message.

#### Reproducing the Problem
The following algorithm should warm (throw an exception?) the user that the data resolution (Daily) is incompatible with intraday scheduled event (need minute resolution or higher in this case)

```csharp
    public class ScheduledEventsAlgorithm : QCAlgorithm {
    	    SetStartDate(2013, 10, 7);
    	    SetEndDate(2013, 10, 11);
            AddEquity("SPY", Resolution.Daily);

            Schedule.On(DateRules.EveryDay("SPY"), TimeRules.AfterMarketOpen("SPY", 10), () => {
                Debug($"IsWarmingUp? {IsWarmingUp}");
            });
        }
    }
}
```
#### Checklist

AlexCatarino:
#### Description
After the interval for a new `ITimeRules` is defined, we check whether the data resolution is compatible. If not, we log a message.

#### Related Issue
Closes #3687 

#### Motivation and Context
Some users create schedule events for times that will not be triggered by actual data time. For example, after market is open with daily resolution. We add a new message to warn them that the data resolution must greater than or equal to the frequency with which the scheduled event occurs.

Yes, we could improve the section in [Scheduled Events](https://www.quantconnect.com/docs/algorithm-reference/scheduled-events) with more details.

Varying the resolution of `ScheduledEventsAlgorithm`, we can see the new messages.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Scheduling/TimeRules.cs
+++ b/Common/Scheduling/TimeRules.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using NodaTime;
+using QuantConnect.Data;
 using QuantConnect.Securities;
 using static QuantConnect.StringExtensions;
 
@@ -108,6 +109,9 @@ namespace QuantConnect.Scheduling
         public ITimeRule At(TimeSpan timeOfDay, DateTimeZone timeZone)
         {
             var name = string.Join(",", timeOfDay.TotalHours.ToStringInvariant("0.##"));
+
+            ValidateInterval(name, timeOfDay);
+
             Func<IEnumerable<DateTime>, IEnumerable<DateTime>> applicator = dates =>
                 from date in dates
                 let localEventTime = date + timeOfDay
@@ -128,7 +132,11 @@ namespace QuantConnect.Scheduling
             {
                 throw new ArgumentException("TimeRules.Every(): time span interval can not be zero or less");
             }
+
             var name = Invariant($"Every {interval.TotalMinutes:0.##} min");
+
+            ValidateInterval(name, interval);
+
             Func<IEnumerable<DateTime>, IEnumerable<DateTime>> applicator = dates => EveryIntervalIterator(dates, interval);
             return new FuncTimeRule(name, applicator);
         }
@@ -149,6 +157,9 @@ namespace QuantConnect.Scheduling
             var name = Invariant($"{symbol}: {minutesAfterOpen:0.##} min after {type}");
 
             var timeAfterOpen = TimeSpan.FromMinutes(minutesAfterOpen);
+
+            ValidateInterval(name, timeAfterOpen, security);
+
             Func<IEnumerable<DateTime>, IEnumerable<DateTime>> applicator = dates =>
                 from date in dates
                 where security.Exchange.DateIsOpen(date)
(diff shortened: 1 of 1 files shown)


Common/Scheduling/TimeRules.cs
```diff
@@ -215,5 +229,29 @@ private static IEnumerable<DateTime> EveryIntervalIterator(IEnumerable<DateTime>
...
+        /// <param name="security">The security with the data subscriptions. If not defined, will use all the available securities</param>
+        public void ValidateInterval(string name, TimeSpan interval, Security security = null)
+        {
+            var resolution = (security?.Subscriptions ?? _securities.SelectMany(x => x.Value.Subscriptions))
+                .GetHighestResolution();
+
+            if (resolution == Resolution.Daily && interval != TimeSpan.Zero ||
+                resolution == Resolution.Hour && interval.TotalHours - interval.Hours > 0)
+            {
+                Logging.Log.Trace(
```
jaredbroad: Please make this reach the user, not just the logs.

Common/Scheduling/TimeRules.cs
```diff
@@ -215,5 +229,29 @@ private static IEnumerable<DateTime> EveryIntervalIterator(IEnumerable<DateTime>
...
+
+        /// <summary>
+        /// Check whether the data resolution is compatible with a interval defined by a <see cref="ITimeRule"/>. 
+        /// </summary>
+        /// <param name="name"><see cref="ITimeRule"/> name</param>
+        /// <param name="interval">The interval defined by a <see cref="ITimeRule"/></param>
+        /// <param name="security">The security with the data subscriptions. If not defined, will use all the available securities</param>
+        public void ValidateInterval(string name, TimeSpan interval, Security security = null)
+        {
+            var resolution = (security?.Subscriptions ?? _securities.SelectMany(x => x.Value.Subscriptions))
```
jaredbroad: Please update to include the universe resolution as well. Will this work with universe only algorithms?

Not sure "GetHighestResolution()" is the right answer here either. If I have a 1-minute level subscription to schedule events but using the scheduled events to trigger trades on daily data, its the same issue.

mchandschuh: **TL;DR;** I think we should be checking both, the `SecurityManager` and then the `SubscriptionManager`.

While this particular method shouldn't be called for universe-only algorithms, you raise a good point. I think this should using the `SubscriptionManager`. IIRC, the `SecurityManager` is guaranteed to be updated with subscriptions added to a security _after_ that security has been added, but the `SubscriptionManager` has access to the `DataManager` via its `IAlgorithmSubscriptionManager` member which is capable of providing a direct snapshot of whatever subscriptions are in the data feed since the data feeds us the `DataManager` for subscription storage.... which... since it's the data feed's storage, I think we'll need to use both, since the subscriptions aren't guaranteed to be added _directly_ after adding a security.

Another concern here is that this is an `O(n)` solution. We could prevent the `O(n)` search by maintaining the value in an appropriate place (like the `DataManager`). In fact, the highest resolution should always be the FF resolution, excluding tick.

Common/Scheduling/TimeRules.cs
mchandschuh: Interesting! As time goes on, the number of components that need access to logging functionality increases. The running solution has been to throw `IAlgorithm` references around. Perhaps it's time to extract the `IAlgorithm`'s logging functions out into a separate interface and have `IAlgorithm` implement that new interface. In this way, we can avoid injecting the fattest abstract in LEAN (`IAlgorithm`) into one of the smallest components.

jaredbroad:
Closing for now but let's rethink and implement completely.

Issue #3689, opened 2019-10-08
Allow Option BidPrice or AskPrice to have zero value


StefanoRaggi:
#### Expected Behavior
If an option contract does not have a Bid price but only has an Ask price, `security.BidPrice` should have a zero value.

#### Actual Behavior
If an option contract does not have a Bid price but only has an Ask price, `security.BidPrice` is equal to `security.Price` which is also equal to `security.AskPrice`.

#### Potential Solution
Override the `BidPrice` and `AskPrice` properties in the `Option` class and return `Cache.BidPrice` or `Cache.AskPrice`

#### Reproducing the Problem
Add any far OTM option contract to the algorithm:
``` C#
var symbol = QuantConnect.Symbol.CreateOption("HGV", Market.USA, OptionStyle.American, OptionRight.Put, 21m, new DateTime(2019, 12, 20));
_security = AddOptionContract(symbol, Resolution.Minute);
```
In `OnData` log `security.BidPrice` and `security.AskPrice`, notice that they both have the same non-zero value.

#### System Information
n/a

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

StefanoRaggi:
The merged fixes above did not completely solve the original problem, which is consistently repeatable with the following algorithm (deployed to IB):

``` C#
    public class BasicTemplateAlgorithm : QCAlgorithm
    {
        public override void Initialize()
        {
            SetStartDate(2013, 10, 07);
            SetEndDate(2013, 10, 11);
            SetCash(100000);

            var option = AddOption("BHP", Resolution.Minute);
            option.SetFilter(x => x.Where(y => y.Value == "BHP   200117C00070000"));
        }

        public override void OnData(Slice data)
        {
            foreach (var chain in data.OptionChains.Values)
            {
                foreach (var contract in chain.Contracts.Values)
                {
                    var security = Securities[contract.Symbol];

                    Log($"{Time} {contract.Symbol.Value} " +
                        $"BidPrice:{contract.BidPrice} AskPrice:{contract.AskPrice} BidSize:{contract.BidSize} AskSize:{contract.AskSize} " +
                        $"Security.BidPrice:{security.BidPrice} Security.AskPrice:{security.AskPrice}");
                }
            }
        }
    }
[...]

PR #3699, opened 2019-10-09
Fix tick emit logic bugs in IB DataQueueHandler

StefanoRaggi:
#### Description
1. **Only emit quote ticks when both the price and size messages have been received**
Market data for trades and quotes are received by the IB API callbacks as separate messages, the price is always received before the size. Previously we were emitting ticks on both price and size messages.

2. **Do not emit half-quotes (if both quotes available)**
Even in very liquid instruments, we were previously sending quote ticks with e.g. only bid price and size but zero ask price and size. For consistency with other `IDataQueueHandler` implementations we now always emit complete quote ticks (for cases where there is no bid price available such as far OTM options, we'll emit the half-quote with zero for bid price and size).

3. **The tick Quantity field is now set only for Trade ticks**
Previously it was also being set for Quote ticks.

4. **Removed the possibility of emitting ticks with BidPrice >= AskPrice**
Previously this could happen.

#### Related Issue
n/a

#### Motivation and Context
Consistency and correctness of emitted ticks.

n/a

- A live algorithm running locally with extensive logging of both IB messages received, ticks emitted and ticks/bars in `OnData` slices (for all security types, especially futures and options where this code is used in the cloud). 
- Manual inspection of the above logs has shown expected behavior.

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -2481,62 +2481,83 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
 
         private void HandleTickPrice(object sender, IB.TickPriceEventArgs e)
         {
+            // tickPrice events are always followed by tickSize events,
+            // so we save off the bid/ask/last prices and only emit ticks in the tickSize event handler.
+
             Symbol symbol;
 
-            if (!_subscribedTickets.TryGetValue(e.TickerId, out symbol)) return;
+            if (!_subscribedTickets.TryGetValue(e.TickerId, out symbol))
+            {
+                return;
+            }
 
             var price = Convert.ToDecimal(e.Price);
 
-            var tick = new Tick();
-            // in the event of a symbol change this will break since we'll be assigning the
-            // new symbol to the permtick which won't be known by the algorithm
-            tick.Symbol = symbol;
-            tick.Time = GetRealTimeTickTime(symbol);
-            var securityType = symbol.ID.SecurityType;
-            tick.Value = price;
-
-            if (e.Price <= 0 &&
-                securityType != SecurityType.Future &&
-                securityType != SecurityType.Option)
-                return;
-
-            int quantity;
+            Tick tick;
             switch (e.Field)
             {
                 case IBApi.TickType.BID:
                 case IBApi.TickType.DELAYED_BID:
 
-                    tick.TickType = TickType.Quote;
-                    tick.BidPrice = price;
-                    _lastBidSizes.TryGetValue(symbol, out quantity);
-                    tick.Quantity = quantity;
-                    _lastBidPrices[symbol] = price;
+                    if (!_lastQuoteTicks.TryGetValue(symbol, out tick))
+                    {
+                        tick = new Tick
+                        {
+                            // in the event of a symbol change this will break since we'll be assigning the
+                            // new symbol to the permtick which won't be known by the algorithm
+                            Symbol = symbol,
+                            TickType = TickType.Quote
+                        };
+                        _lastQuoteTicks.TryAdd(symbol, tick);
+                    }
+
+                    // negative price (-1) means no price available, normalize to zero
+                    tick.BidPrice = price < 0 ? 0 : price;
                     break;
 
                 case IBApi.TickType.ASK:
                 case IBApi.TickType.DELAYED_ASK:
 
-                    tick.TickType = TickType.Quote;
-                    tick.AskPrice = price;
-                    _lastAskSizes.TryGetValue(symbol, out quantity);
-                    tick.Quantity = quantity;
-                    _lastAskPrices[symbol] = price;
+                    if (!_lastQuoteTicks.TryGetValue(symbol, out tick))
+                    {
+                        tick = new Tick
+                        {
+                            // in the event of a symbol change this will break since we'll be assigning the
+                            // new symbol to the permtick which won't be known by the algorithm
+                            Symbol = symbol,
+                            TickType = TickType.Quote
+                        };
+
+                        _lastQuoteTicks.TryAdd(symbol, tick);
+                    }
+
+                    // negative price (-1) means no price available, normalize to zero
+                    tick.AskPrice = price < 0 ? 0 : price;
                     break;
 
                 case IBApi.TickType.LAST:
                 case IBApi.TickType.DELAYED_LAST:
 
-                    tick.TickType = TickType.Trade;
+                    if (!_lastTradeTicks.TryGetValue(symbol, out tick))
+                    {
+                        tick = new Tick
+                        {
+                            // in the event of a symbol change this will break since we'll be assigning the
+                            // new symbol to the permtick which won't be known by the algorithm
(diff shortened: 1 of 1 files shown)


mchandschuh:
Regarding not emitting half quotes, is this desired? I'm recalling the raw data from when I was working on options and remember seeing many instances of having only a bid or ask update at a given time step. It seems like by preventing _half_ in all cases we may be filtering out potentially important data, **especially** in highly illiquid securities.

StefanoRaggi:
> Regarding not emitting half quotes, is this desired? I'm recalling the raw data from when I was working on options and remember seeing many instances of having only a bid or ask update at a given time step. It seems like by preventing _half_ in all cases we may be filtering out potentially important data, **especially** in highly illiquid securities.

I probably wasn't clear enough in the PR description for that point: 
the actual bug being fixed is that when **both sides of the quote are available** only one half of it was being emitted at any time (one tick emitted for the bid with **zero ask** plus another tick emitted for the ask with **zero bid**).
In the case of an illiquid instrument with no bid or ask, the half quote **will** be emitted -- this case was tested with the example code in issue #3689.
In fact, when the very first tick for the subscription is received the half tick will always be emitted anyway (since we don't know if one of the sides is empty).

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2557,77 +2578,134 @@ private void HandleTickSize(object sender, IB.TickSizeEventArgs e)
         {
             Symbol symbol;
 
-            if (!_subscribedTickets.TryGetValue(e.TickerId, out symbol)) return;
+            if (!_subscribedTickets.TryGetValue(e.TickerId, out symbol))
```
Martin-Molinero: An idea: Maybe `_subscribedTickets` could hold the `Symbol` and all `Tick` types in a new class, would help us avoid the extra `ConcurrentDictionaries`, their access and the `if(!) add` logics, because since we are not removing them, after the first tick the symbol will already be in the dictionaries ?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2557,77 +2578,134 @@ private void HandleTickSize(object sender, IB.TickSizeEventArgs e)
...
-                    tick.BidSize = tick.Quantity;
+                    if (tick.BidPrice == 0)
+                    {
+                        // no bid price, do not emit tick
+                        return;
+                    }
+
+                    if (tick.BidPrice > 0 && tick.AskPrice > 0 && tick.BidPrice >= tick.AskPrice)
+                    {
+                        // new bid price jumped above previous ask price, wait for new ask price
```
Martin-Molinero: Another idea 🤔  would it be worth emitting half a bid quote here, setting the ask values to 0?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2557,77 +2578,134 @@ private void HandleTickSize(object sender, IB.TickSizeEventArgs e)
...
-            var tick = new Tick();
-            // in the event of a symbol change this will break since we'll be assigning the
-            // new symbol to the permtick which won't be known by the algorithm
-            tick.Symbol = symbol;
             var securityType = symbol.ID.SecurityType;
-            tick.Quantity = AdjustQuantity(securityType, e.Size);
-            tick.Time = GetRealTimeTickTime(symbol);
-
-            if (tick.Quantity == 0) return;
+            var quantity = AdjustQuantity(securityType, e.Size);
```
Martin-Molinero: I think `quantity` could also be reported as `-1` in some cases, should we add a check for it? `"...When IBApi::EWrapper::tickPrice and IBApi::EWrapper::tickSize are reported as -1, this indicates that there is no data currently available...."`

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2481,62 +2481,83 @@ private DateTime GetRealTimeTickTime(Symbol symbol)
...
+                            // in the event of a symbol change this will break since we'll be assigning the
+                            // new symbol to the permtick which won't be known by the algorithm
+                            Symbol = symbol,
+                            TickType = TickType.Quote
+                        };
+                        _lastQuoteTicks.TryAdd(symbol, tick);
+                    }
+
+                    // negative price (-1) means no price available, normalize to zero
+                    tick.BidPrice = price < 0 ? 0 : price;
```
Martin-Molinero: Maybe we could normalize the price outside the switch so it's shared? (doesn't it also apply for `IBApi.TickType.LAST`?

StefanoRaggi: I've never seen a negative price for `IBApi.TickType.LAST` but will move the normalization up to remove duplication

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Good idea, will do 👍

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Do you have a link for that quote?

Looking at the latest IB API docs for `tickPrice` here:
https://interactivebrokers.github.io/tws-api/interfaceIBApi_1_1EWrapper.html#ae851ec3a1e0fa2d0964c7779b0c89718
```
Market data tick price callback. Handles all price related ticks. 
Every tickPrice callback is followed by a tickSize. 
A tickPrice value of -1 or 0 followed by a tickSize of 0 indicates there is no data for this field currently available, whereas a tickPrice with a positive tickSize indicates an active quote of 0 (typically for a combo contract).
```
This was also confirmed by my logs during testing, where I have never seen a negative `tickSize` returned.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Hmm... point 2 of this PR was to actually prevent us from emitting incomplete/half ticks when possible 😃 

In these normal bid/ask price changes, i.e. when both bid and ask quotes move up or down **and** are both non-zero, we will always receive both half-quotes in sequence.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
Martin-Molinero: Right, link @ https://interactivebrokers.github.io/tws-api/md_receive.html#market_data_return 3rd paragraph, I think its the case where there is no data available at all vs the field, I guess adding the check for -1 won't cause any harm?

StefanoRaggi: Ok, I'll add the normalization for negative quantity as well.
The Trade tick with non-positive quantity will be discarded anyway by `Tick.IsValid()`.

Issue #3696, opened 2019-10-08
DynamicSecurityData Performance


Martin-Molinero:
#### Expected Behavior

- `DynamicSecurityData` adds the least amount of performance overhead
#### Actual Behavior

- `DynamicSecurityData` was recently merged on PR https://github.com/QuantConnect/Lean/pull/3631 and it added a mensurable performance overhead (see https://www.quantconnect.com/performance) . Sampling for `BasicTemplateBenchmark`:
![image](https://user-images.githubusercontent.com/18473240/66441039-8fd2af80-ea0b-11e9-98bb-2b47eb3acfa3.png)

#### Potential Solution

- Refactor
#### Reproducing the Problem

Run Lean performance benchmarks
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3714, opened 2019-10-14
Performance improvements v2

Martin-Molinero:
#### Description

- `PortfolioTargetCollection` avoid calling `Count` on
ConcurrentDictionary directly -> has to take all locks
- `SecurityChanges` change Union for Concat since constructor will call
HashSet
- Make `DynamicSecurityData` hold lazy data objects
- `RegisteredSecurityDataTypesProvider` avoid looping over all
registered types, adding `TryGetType`
- `Security.Update()` will not call group by on data since this data is
already grouped by type. Adding `ContainsFillForwardData` will allows to
be lazy and not re loop through the data unless necessary
- `DefaultAlphaHandler` will use the `static`
`Enumerable.Empty<Insight>` instance when possible
- `SubscriptionSynchronizer` will be lazy to construct the
`universeData` dictionary which is not used in most of the times. Will
use `Count` vs `Any` -> `Count` is known by the dictionary
- For python algorithms `JobQueue` will respect `AlgorithmLocation`, was
using unexisting `"algorithm-path-python"`

Commit2
- Stop running threads in unit tests (try to fix travis hangs)

Performance tests `BasicTemplateBenchmark C#`:
`master`
30.69 seconds at 54k data points per second. Processing total of 1,664,679 data points.
30.69 seconds at 54k data points per second. Processing total of 1,664,679 data points.
30.30 seconds at 55k data points per second. Processing total of 1,664,679 data points.
`pr` 
27.89 seconds at 60k data points per second. Processing total of 1,664,679 data points.
27.49 seconds at 61k data points per second. Processing total of 1,664,679 data points.
27.89 seconds at 60k data points per second. Processing total of 1,664,679 data points.

#### Related Issue

Related to https://github.com/QuantConnect/Lean/issues/3696
#### Motivation and Context

Performance 

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/Security.cs
+++ b/Common/Securities/Security.cs
@@ -591,26 +591,25 @@ namespace QuantConnect.Securities
         /// on the data provided. Data is also stored into the security's data cache
         /// </summary>
         /// <param name="data">The security update data</param>
-        public void Update(IEnumerable<BaseData> data)
+        /// <param name="dataType">The data type</param>
+        /// <param name="containsFillForwardData">Flag indicating whether
+        /// <paramref name="data"/> contains any fill forward bar or not</param>
+        public void Update(IReadOnlyList<BaseData> data, Type dataType, bool? containsFillForwardData = null)
         {
-            BaseData last = null;
-            foreach (var grp in data.GroupBy(d => d.GetType()))
-            {
-                var nonFillForward = new List<BaseData>();
-                foreach (var baseData in grp)
-                {
-                    last = baseData;
-                    if (!baseData.IsFillForward)
-                    {
-                        nonFillForward.Add(baseData);
-                    }
-                }
-                // use the last of each type to set market price
-                SetMarketPrice(last);
+            // use the last of each type to set market price
+            SetMarketPrice(data[data.Count - 1]);
 
-                // maintaining regression requires us to NOT cache FF data
-                Cache.StoreData(nonFillForward);
+            var nonFillForwardData = data;
+            // maintaining regression requires us to NOT cache FF data
+            if (!containsFillForwardData.HasValue || containsFillForwardData.Value)
+            {
+                nonFillForwardData = data.Where(baseData => !baseData.IsFillForward).ToList();
+                if (nonFillForwardData.Count == 0)
+                {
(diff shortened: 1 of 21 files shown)


Common/Securities/Security.cs
```diff
@@ -591,26 +591,25 @@ public void SetMarketPrice(BaseData data)
...
+            // use the last of each type to set market price
+            SetMarketPrice(data[data.Count - 1]);
+
+            var nonFillForwardData = data;
+            // maintaining regression requires us to NOT cache FF data
+            if (!containsFillForwardData.HasValue || containsFillForwardData.Value)
             {
-                var nonFillForward = new List<BaseData>();
-                foreach (var baseData in grp)
+                nonFillForwardData = data.Where(baseData => !baseData.IsFillForward).ToList();
```
mchandschuh: Surprised you want for the `Where` instead of `for`. Creating the list explicitly you could even set the size of it since we already know the max size, not sure if it would have an impact though.

Common/Securities/SecurityCache.cs
```diff
@@ -207,21 +207,16 @@ public void AddData(BaseData data)
         /// Stores the specified data list in the cache WITHOUT updating any of the cache properties, such as Price
         /// </summary>
         /// <param name="data">The collection of data to store in this cache</param>
-        public void StoreData(IReadOnlyList<BaseData> data)
+        /// <param name="dataType">The data type</param>
+        public void StoreData(IReadOnlyList<BaseData> data, Type dataType)
         {
-            if (data.Count == 0)
```
mchandschuh: interesting... why remove this?

Common/Securities/Security.cs
Martin-Molinero: Yeah 👍  actually saw no difference after the `containsFillForwardData` flag was added, it really did the trick

Common/Securities/SecurityCache.cs
Martin-Molinero: Reviewed the consumers and it wasn't a case that could happen today, so I figured I'd remove it

PR #3718, opened 2019-10-14
Implements ConfidenceWeightedPortfolioConstructionModel

### LINKED ISSUE #3672: Create ConfidenceWeightedPortfolioConstructionModel
#### Expected Behavior
Confidence weighted portfolio construction model, similar to the "[Weighting](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Framework/Portfolio/InsightWeightingPortfolioConstructionModel.cs)" portfolio construction model. 

#### Actual Behavior
None exists.

#### Potential Solution
`ConfidenceWeightedPortfolioConstructionModel : InsightWeightingPortfolioConstructionModel` with the weighting variable substituted for Confidence. 
  
#### Checklist

AlexCatarino:
#### Description
- Adding new `ConfidenceWeightedPortfolioConstructionModel` (C# / Py) that will
generate percent `Targets` based on the latest active `Insight` `Confidence` per
`Symbol`.
   - Will ignore `Insights` that have no `Confidence`.(unit tested)
   - If the sum of all the last active `Insight` per `Symbol` is bigger than 1, it
will factor down each target percent holdings proportionally so the sum is 1. (unit tested)
   - Adding unit tests
   - Adding a new regression test framework algorithm (C#/Py)

**Note**: `ConfidenceWeightedPortfolioConstructionModel` inherits from the `InsightWeightingPortfolioConstructionModel`. Protect method `GetValue` was implemented in `IWPCM` to enable the choice of `Insight` member.

#### Related Issue
Closes #3672 

#### Motivation and Context
The confidence of insight is a good parameter to determine how much capital a portfolio should allocate in security. 

Yes, we should list this built-in model in the Portfolio Construction Model section:
https://www.quantconnect.com/docs/algorithm-framework/portfolio-construction

Unit tests and regression algorithm.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
+++ b/Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
@@ -21,21 +21,21 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
 {
     /// <summary>
     /// Provides an implementation of <see cref="IPortfolioConstructionModel"/> that generates percent targets based on the
-    /// <see cref="Insight.Weight"/>. The target percent holdings of each Symbol is given by the <see cref="Insight.Weight"/>
+    /// <see cref="Insight.Confidence"/>. The target percent holdings of each Symbol is given by the <see cref="Insight.Confidence"/>
     /// from the last active <see cref="Insight"/> for that symbol.
     /// For insights of direction <see cref="InsightDirection.Up"/>, long targets are returned and for insights of direction
     /// <see cref="InsightDirection.Down"/>, short targets are returned.
     /// If the sum of all the last active <see cref="Insight"/> per symbol is bigger than 1, it will factor down each target
     /// percent holdings proportionally so the sum is 1.
-    /// It will ignore <see cref="Insight"/> that have no <see cref="Insight.Weight"/> value.
+    /// It will ignore <see cref="Insight"/> that have no <see cref="Insight.Confidence"/> value.
     /// </summary>
-    public class InsightWeightingPortfolioConstructionModel : EqualWeightingPortfolioConstructionModel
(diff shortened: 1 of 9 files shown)


Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System.Collections.Generic;
+using System.Linq;
```
Martin-Molinero: nit - these using statements can be removed

Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            return insight.Confidence.HasValue;
+        }
+
+        /// <summary>
+        /// Method that will determine which member will be used to compute the weights and gets its value
+        /// </summary>
+        /// <param name="insight">The insight to create a target for</param>
+        /// <returns>The value of the selected insight member</returns>
+        protected override double GetValue(Insight insight) => insight.Confidence ?? 0;
```
Martin-Molinero: Good idea! 👍 
nit - no real change required - based on the implementation of `ShouldCreateTargetForInsight` we shouldn't _have_ to check if it has value or not, maybe we should even throw if we are called and it doesn't have value cause it means something isn't going as we expected 🤔

AlexCatarino: I added the check because ReSharper was asking for. :-)

PR #3743, opened 2019-10-17
Adds support for extending algo time loop for long-running scheduled events

### LINKED ISSUE #3319: Custom Timeouts for Training Models

mchandschuh:
#### Description
Adds support for long-running scheduled events. When a scheduled event is detected to have exceed one full second of execution time, we'll start requesting 'minutes' from a leaky bucket implementation managed by the `AlgorithmManager` via a proxy that also controls the delegate provided to the `Isolator`. When a scheduled event calls `AlgorithmTimeLimitManager.RequestAdditionalTime(int minutes)`, it increases the maximum amount of time the algorithm is allowed to spend in a single time loop.

 We restrict each algorithm time loop to a pre-determined amount of time.
Exceeding this limit will cause the algorithm to immediately terminate.
This quickly becomes an issue when considering users running trainable
models that have a long initialization period that exceeds the time loop
maximum.

This change provides a mechanism through which a long-running scheduled
event is permitted to keep running and is permitted to avoid the time loop
permitted by requesting additional time. Requests for additional time are
limited according to a leaky bucket implementation whose parameters are
set via the job's controls structure. The fundamental time unit for the
algorithm is a single minute.

Here's how it works. If a scheduled event takes longer than one full wall
clock second then a request is made to the leaky bucket for one more minute.
If the scheduled event continues to take more time, it will continue to
request additional minutes. Each requested minute will prevent the algorithm's
time loop check from terminating the algorithm. When the bucket is empty and
no more minutes are available to be requested, a TimeoutException is thrown
causing a cascade that ends in the algorithm's termination and status being
flipped to RuntimeError.

Additionally, this applies equally to ALL scheduled events. While some helpers
were added with the naming of Train and TrainNow to the ScheduleManager, these
methods don't do anything special and the infrastructure doesn't otherwise
flag them as different, so this feature becomes part of the core Scheduled
Event feature set.

Further, the live scheduled events were not touched and are still pending
further discussion regarding the value added by enforcing a time restriction
when simulation time and wall clock time are equivalent.

#### Related Issue

Fixes #3319

#### Motivation and Context

The aim of this change is to provide a controlled mechanism for users to run ML training that can potentially take a fairly large amount of time (up to 2 hours) without running afoul of the algorithm manager's time loop limit restrictions. We also seek to limit the total amount of time an algorithm can do this through a leaky bucket implementation whose parameters are set via the job packet.

Yes. We should add the new training methods to the API documentation along with a description of behavior and the default time limits.

Unit tests added for new components and additional tests added to check certain aspects of the integrated implementation. Also, two regression tests are added to validate throwing and refill behavior when running full stack.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
@@ -38,7 +38,7 @@ namespace QuantConnect.Algorithm.CSharp
             // this should cause the algorithm to fail
             // the regression test sets the time limit to 30 seconds and there's one extra
             // minute in the bucket, so a two minute sleep should result in RuntimeError
-            Schedule.TrainingNow(() => Thread.Sleep(TimeSpan.FromMinutes(2)));
+            Schedule.TrainingNow(() => Thread.Sleep(TimeSpan.FromMinutes(2.5)));
         }
 
         public bool CanRunLocally => true;
--- a/Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
@@ -41,7 +41,7 @@ namespace QuantConnect.Algorithm.CSharp
             Schedule.TrainingNow(() => Thread.Sleep(TimeSpan.FromMinutes(2.5)));
         }
 
-        public bool CanRunLocally => true;
+        public bool CanRunLocally => false;
         public Language[] Languages => new[] {Language.CSharp};
         public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>();
     }
--- a/Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
@@ -1,0 +1,48 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This regression algorithm is expected to fail and verifies that a training event
+    /// created in Initialize will get run AND it will cause the algorithm to fail if it
+    /// exceeds the "algorithm-manager-time-loop-maximum" config value, which the regression
+    /// test sets to 0.5 minutes.
+    /// </summary>
+    public class TrainingInitializeRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 7);
+            SetEndDate(2013, 10, 11);
+
+            AddEquity("SPY", Resolution.Daily);
+
+            // this should cause the algorithm to fail
+            // the regression test sets the time limit to 30 seconds and there's one extra
+            // minute in the bucket, so a two minute sleep should result in RuntimeError
+            Schedule.TrainingNow(() => Thread.Sleep(TimeSpan.FromMinutes(2)));
+        }
+
+        public bool CanRunLocally => true;
+        public Language[] Languages => new[] {Language.CSharp};
+        public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>();
+    }
+}
--- a/Algorithm.CSharp/TrainingScheduledRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/TrainingScheduledRegressionAlgorithm.cs
@@ -46,7 +46,7 @@ namespace QuantConnect.Algorithm.CSharp
             });
         }
 
-        public bool CanRunLocally => true;
+        public bool CanRunLocally => false;
         public Language[] Languages => new[] {Language.CSharp};
         public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>();
     }
--- a/Algorithm.CSharp/TrainingScheduledRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/TrainingScheduledRegressionAlgorithm.cs
@@ -1,0 +1,53 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This regression algorithm is NOT expected to fail and verifies that a periodic training
+    /// event created in OnData will fire and will consume 'minutes' from the leaky bucket
+    /// </summary>
+    public class TrainingScheduledRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 7);
+            SetEndDate(2013, 10, 11);
+
+            AddEquity("SPY", Resolution.Daily);
+
+            // DateRules.Tomorrow combined with TimeRules.Noon enforces that this event schedule will
+            // have exactly one time, which will fire between the first data point and the next day at
+            // midnight. So after the first data point, it will run this event and sleep long enough to
+            // exceed the static max algorithm time loop time and begin to consume from the leaky bucket
+            // the regression test sets the "algorithm-manager-time-loop-maximum" value to 30 seconds
+            Schedule.Training(DateRules.Tomorrow, TimeRules.Midnight, () =>
+            {
+                // this will consume the single 'minute' available in the leaky bucket
+                // and the regression test will confirm that the leaky bucket is empty
+                Thread.Sleep(TimeSpan.FromMinutes(1));
+            });
+        }
+
+        public bool CanRunLocally => true;
+        public Language[] Languages => new[] {Language.CSharp};
+        public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>();
+    }
+}
--- a/Algorithm.Python/TrainingExampleAlgorithm.py
+++ b/Algorithm.Python/TrainingExampleAlgorithm.py
@@ -1,0 +1,52 @@
+﻿# QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+# Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from clr import AddReference
+AddReference("System")
+AddReference("QuantConnect.Algorithm")
+AddReference("QuantConnect.Common")
+
+from System import *
+from QuantConnect import *
+from QuantConnect.Algorithm import *
+from time import sleep
+
+### <summary>
+### Example algorithm showing how to use QCAlgorithm.Train method
+### </summary>
+### <meta name="tag" content="using quantconnect" />
+### <meta name="tag" content="training" />
+class TrainingExampleAlgorithm(QCAlgorithm):
+    '''Example algorithm showing how to use QCAlgorithm.Train method'''
+
+    def Initialize(self):
+
+        self.SetStartDate(2013, 10, 7)
+        self.SetEndDate(2013, 10, 14)
+
+        self.AddEquity("SPY", Resolution.Daily)
+
+        # Set TrainingMethod to be executed immediately
+        self.Train(self.TrainingMethod)
+
+        # Set TrainingMethod to be executed at 8:00 am every Sunday
+        self.Train(self.DateRules.Every(DayOfWeek.Sunday), self.TimeRules.At(8 , 0), self.TrainingMethod)
+
+    def TrainingMethod(self):
+
+        self.Log(f'Start training at {self.Time}')
+        # Use the historical data to train the machine learning model
+        history = self.History(["SPY"], 200, Resolution.Daily)
+
+        # ML code:
+        pass
--- a/Common/IsolatorLimitResult.cs
+++ b/Common/IsolatorLimitResult.cs
@@ -44,8 +44,8 @@ namespace QuantConnect
         /// <param name="errorMessage">The error message or an empty string if no error on the current time step</param>
         public IsolatorLimitResult(TimeSpan currentTimeStepElapsed, string errorMessage)
         {
-            CurrentTimeStepElapsed = currentTimeStepElapsed;
             ErrorMessage = errorMessage;
+            CurrentTimeStepElapsed = currentTimeStepElapsed;
         }
     }
 }
--- a/Common/IsolatorLimitResultProvider.cs
+++ b/Common/IsolatorLimitResultProvider.cs
@@ -17,12 +17,12 @@ using System;
 using System.Threading;
 using System.Threading.Tasks;
 using QuantConnect.Logging;
+using QuantConnect.Scheduling;
 
 namespace QuantConnect
 {
     /// <summary>
-    /// Provides access to the <see cref="NullIsolatorLimitResultProvider"/> and can be a place for future
-    /// extension methods of <see cref="IIsolatorLimitResultProvider"/>
+    /// Provides access to the <see cref="NullIsolatorLimitResultProvider"/> and extension methods supporting <see cref="ScheduledEvent"/>
     /// </summary>
     public static class IsolatorLimitResultProvider
     {
@@ -35,17 +35,15 @@ namespace QuantConnect
         public static readonly IIsolatorLimitResultProvider Null = new NullIsolatorLimitResultProvider();
 
         /// <summary>
-        /// Executes the provided code block and while the code block is running, continually consume from
-        /// the limit result provided one token each minute. This function allows the code to run for the
-        /// first full second without requesting additional time from the provider. Following that, every
-        /// minute an additional one minute will be requested from the provider.
+        /// Convenience method for invoking a scheduled event's Scan method inside the <see cref="IsolatorLimitResultProvider"/>
         /// </summary>
-        public static void ConsumeWhileExecuting(
+        public static void Consume(
             this IIsolatorLimitResultProvider isolatorLimitProvider,
-            Action code
+            ScheduledEvent scheduledEvent,
+            DateTime scanTimeUtc
             )
         {
-            isolatorLimitProvider.ConsumeWhileExecuting(string.Empty, code);
+            isolatorLimitProvider.Consume(scheduledEvent.Name, () => scheduledEvent.Scan(scanTimeUtc));
         }
 
         /// <summary>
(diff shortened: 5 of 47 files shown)


Common/IsolatorLimitResultProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+                    // on the first iteration and every minute following, request additional time
+                    Log.Trace($"IsolatorLimitResultProvider.ConsumeForScheduledEvent({name}): Requesting additional time. Elapsed minutes: {count / 60}");
+                    if (!isolatorLimitProvider.TryRequestAdditionalTime(minutes: 1))
+                    {
+                        error = new TimeoutException("The scheduled event exceeded the available amount of time for processing. " +
+                                                    $"Elapsed: {count/60.0:0.0} minutes. Scheduled Event: {name}");
+                        break;
+                    }
+
+                    count++;
```
Martin-Molinero: I think this `count++;` should be before the first `continue`, 🤔 else it will never continue increasing?

Engine/RealTime/BacktestingRealTimeHandler.cs
```diff
@@ -123,13 +136,19 @@ public void ScanPastEvents(DateTime time)
         {
             foreach (var scheduledEvent in GetScheduledEventsSortedByTime())
             {
+                if (scheduledEvent.NextEventUtcTime > time)
+                {
+                    // since they're sorted, we don't need to go through every single one
+                    break;
```
Martin-Molinero: Isn't this already covered by the `while` below? The `_sortingScheduledEventsRequired = true;` might go in the while loop

Common/IsolatorLimitResultProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// Convenience method for invoking a scheduled event's Scan method inside the <see cref="IsolatorLimitResultProvider"/>
+        /// </summary>
+        public static void Consume(
+            this IIsolatorLimitResultProvider isolatorLimitProvider,
+            ScheduledEvent scheduledEvent,
+            DateTime scanTimeUtc
+            )
+        {
+            // perform initial filtering to prevent starting a task when not necessary
+            if (scheduledEvent.NextEventUtcTime > scanTimeUtc)
```
Martin-Molinero: Isn't the caller already checking this?

Common/Util/RateLimit/BusyWaitSleepStrategy.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace QuantConnect.Util.RateLimit
+{
+    /// <summary>
+    /// Provides a CPU intensive means of waiting for more tokens to be available in <see cref="ITokenBucket"/>.
+    /// This strategy is only viable when the requested number of tokens is expected to become available in an
+    /// extremely short period of time. This implementation aims to keep the current thread executing to prevent
+    /// potential content switches arising from a thread yielding or sleeping strategy.
+    /// </summary>
+    public class BusyWaitSleepStrategy : ISleepStrategy
```
Martin-Molinero: If this class is only used in tests it might make sense to move it there

Common/Util/RateLimit/LeakyBucket.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// This constructor initializes the bucket using the <see cref="ThreadSleepStrategy.Sleep"/> with a 1 millisecond
+        /// sleep to prevent being CPU intensive and uses the <see cref="FixedIntervalRefillStrategy"/> to refill bucket
+        /// tokens according to the <paramref name="refillAmount"/> and <paramref name="refillInterval"/> parameters.
+        /// </summary>
+        /// <param name="capacity">The maximum number of tokens this bucket can hold</param>
+        /// <param name="refillAmount">The number of tokens to add to the bucket each <paramref name="refillInterval"/></param>
+        /// <param name="refillInterval">The interval which after passing more tokens are added to the bucket</param>
+        public LeakyBucket(long capacity, long refillAmount, TimeSpan refillInterval)
+            : this(capacity, ThreadSleepStrategy.Sleeping(1),
+                new FixedIntervalRefillStrategy(RealTimeProvider.Instance, refillAmount, refillInterval)
```
Martin-Molinero: Not sure how will `RealTimeProvider.Instance` work for backtesting, shouldn't it be related to the algorithms Time instead?

Common/Util/RateLimit/LeakyBucket.cs
```diff
@@ -0,0 +1,168 @@
...
+                    // determine if the requested timeout has elapsed
+                    var currentTime = _timeProvider.GetUtcNow();
+                    var elapsedMilliseconds = (currentTime - startTime).TotalMilliseconds;
+                    if (elapsedMilliseconds > timeout)
+                    {
+                        throw new TimeoutException("The operation timed out while waiting for the rate limit to be lifted.");
+                    }
+                }
+
+                _sleep.Sleep();
```
Martin-Molinero: Not sure whats the use case for this `Consume` implementation, it will block the calling thread until the time tokens are available

Common/Util/RateLimit/TokenBucket.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Gets an <see cref="ITokenBucket"/> that always permits consumption
+        /// </summary>
+        public static ITokenBucket Null = new NullTokenBucket();
+
+        /// <summary>
+        /// Provides an overload of <see cref="ITokenBucket.Consume"/> that accepts a <see cref="TimeSpan"/> timeout
+        /// </summary>
+        public static void Consume(this ITokenBucket bucket, long tokens, TimeSpan timeout)
+        {
+            bucket.Consume(tokens, (long) timeout.TotalMilliseconds);
```
Martin-Molinero: Idem question about whats the use case for this implementation

Common/Util/RateLimit/ThreadSleepStrategy.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace QuantConnect.Util.RateLimit
+{
+    /// <summary>
+    /// Provides a CPU non-intensive means of waiting for more tokens to be available in <see cref="ITokenBucket"/>.
+    /// This strategy should be the most commonly used as it either sleeps or yields the currently executing thread,
+    /// allowing for other threads to execute while the current thread is blocked and waiting for new tokens to become
+    /// available in the bucket for consumption.
+    /// </summary>
+    public class ThreadSleepStrategy : ISleepStrategy
```
Martin-Molinero: If we are not going to be using the blocking `Consume` implementation I'm not sure I see the value on these implementations/interfaces, I'd be more inclined to simplify if possible

Algorithm.Python/TrainingExampleAlgorithm.py
```diff
@@ -0,0 +1,52 @@
...
+from QuantConnect import *
+from QuantConnect.Algorithm import *
+from time import sleep
+
+### <summary>
+### Example algorithm showing how to use QCAlgorithm.Train method
+### </summary>
+### <meta name="tag" content="using quantconnect" />
+### <meta name="tag" content="training" />
+class TrainingExampleAlgorithm(QCAlgorithm):
```
Martin-Molinero: Nice algorithm 👍
I think we could use _another_ algorithm as the C# versions that _tests_ the feature in python too

Common/Util/RateLimit/LeakyBucket.cs
```diff
@@ -0,0 +1,168 @@
...
+                // the number of tokens to add, the max of which is the difference between capacity and currently available
+                var deltaTokens = Math.Min(_capacity - _available, refilled);
+
+                // update the available number of units with the new tokens
+                _available += deltaTokens;
+
+                if (tokens > _available)
+                {
+                    // we don't have enough tokens yet
+                    Logging.Log.Trace($"LeakyBucket.TryConsume({tokens}): Failed to consumed tokens. Available: {_available}");
```
Martin-Molinero: Should it be `tokens >= _available`, can we consume all available? Maybe this log would be better as an error log

Common/IsolatorLimitResultProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+                    return;
+                }
+
+                var next = timeProvider.GetUtcNow().AddMinutes(1);
+                while (Interlocked.Read(ref finished) == 0L)
+                {
+                    if (timeProvider.GetUtcNow() >= next)
+                    {
+                        // each minute request additional time from the isolator
+                        next = next.AddMinutes(1);
```
Martin-Molinero: We might want to request time _before_ the minute/time we have is over, to avoid any race conditions with the monitoring thread

Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            SetStartDate(2013, 10, 7);
+            SetEndDate(2013, 10, 11);
+
+            AddEquity("SPY", Resolution.Daily);
+
+            // this should cause the algorithm to fail
+            // the regression test sets the time limit to 30 seconds and there's one extra
+            // minute in the bucket, so a two minute sleep should result in RuntimeError
+            Schedule.TrainingNow(() => Thread.Sleep(TimeSpan.FromMinutes(2.5)));
```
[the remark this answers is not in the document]
mchandschuh: 3 minutes? The regression test suit takes 30+ on my machine... hmm...  It technically needs at least two full minutes since the leaky bucket quantizes time by the minute. So even if the algorithm time loop limit is set to one tick, we'll still wait a minute before deducting time from the leaky bucket and then another full minute before we check again and it fails. The goal of this test is to have it pull one from the bucket and the next one fail.

What we can do is simply disable these from running, either by setting `CanRunLocally = false` or by removing the `IRegressionAlgorithmDefinition` implementation.

Algorithm.CSharp/TrainingScheduledRegressionAlgorithm.cs
```diff
@@ -0,0 +1,53 @@
...
+            // DateRules.Tomorrow combined with TimeRules.Noon enforces that this event schedule will
+            // have exactly one time, which will fire between the first data point and the next day at
+            // midnight. So after the first data point, it will run this event and sleep long enough to
+            // exceed the static max algorithm time loop time and begin to consume from the leaky bucket
+            // the regression test sets the "algorithm-manager-time-loop-maximum" value to 30 seconds
+            Schedule.Training(DateRules.Tomorrow, TimeRules.Midnight, () =>
+            {
+                // this will consume the single 'minute' available in the leaky bucket
+                // and the regression test will confirm that the leaky bucket is empty
+                Thread.Sleep(TimeSpan.FromMinutes(1));
```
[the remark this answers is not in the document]
mchandschuh: This regression algorithm aims to verify that the leaky bucket is emptied, as mentioned in the comment. I think adding a flag obscures the point of this test and would be just another test of the well established scheduled events feature.

I'm _really_ glad you brought this one up though, it made me double check `RegressionTests.cs` and noticed that I had changed the name of this test (used to be `TrainingOnDataRegressionAlgorithm`). So I've updated the regression test to reference the correct algorithm name. 👍

Common/IsolatorLimitResult.cs
```diff
@@ -44,8 +44,8 @@ public class IsolatorLimitResult
         /// <param name="errorMessage">The error message or an empty string if no error on the current time step</param>
         public IsolatorLimitResult(TimeSpan currentTimeStepElapsed, string errorMessage)
         {
-            CurrentTimeStepElapsed = currentTimeStepElapsed;
             ErrorMessage = errorMessage;
+            CurrentTimeStepElapsed = currentTimeStepElapsed;
```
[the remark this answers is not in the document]
mchandschuh: yea, I know, I'm kind of anal about it 😛 

I just spend _soooo_ much time reading code, I normally don't notice doing it anymore until I'm doing my self review lol

Common/IsolatorLimitResultProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This method exists to support scheduled events, and as such, intercepts any errors raised via the
+        /// provided code and wraps them in a <see cref="ScheduledEventException"/>. If in the future this is
+        /// usable elsewhere, consider refactoring to handle the errors in a different fashion.
+        /// </remarks>
+        public static void Consume(
+            this IIsolatorLimitResultProvider isolatorLimitProvider,
+            ITimeProvider timeProvider,
+            string name,
```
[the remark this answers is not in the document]
mchandschuh: Great eye! There used to be some logging in here but I ended up putting `Log.Debug` statements directly in the `AlgorithmTimeLimitManager`

Engine/AlgorithmTimeLimitManager.cs
```diff
@@ -0,0 +1,194 @@
...
+        public void StartNewTimeStep()
+        {
+            if (_stopped)
+            {
+                throw new InvalidOperationException("The AlgorithmTimeLimitManager may not be stopped and restarted.");
+            }
+
+            // maintains existing implementation behavior to reset the time to min value and then
+            // when the isolator pings IsWithinLimit, invocation of CurrentTimeStepElapsed will cause
+            // it to update to the current time. IIRC, this was done to avoid a potential race
```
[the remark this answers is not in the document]
mchandschuh: Awesome!! Thanks for the explanation -- I'll add that in there too

Common/Util/RateLimit/LeakyBucket.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(tokens),
+                    "Number of tokens to consume must be positive"
+                );
+            }
+
+            if (tokens > _capacity)
+            {
+                throw new ArgumentOutOfRangeException(nameof(tokens),
+                    "Number of tokens to consume must be less than or equal to the capacity"
```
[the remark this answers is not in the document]
mchandschuh: True, but an exception like that would typically arise from a misconfigured system. That being said, I agree that a consumer should be able to determine the limits of valid inputs without knowing how the type was constructed. I'll add `ITokenBucket.Capacity` so the abstraction is self contained. Great comment! 👍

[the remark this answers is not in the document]
mchandschuh: RE: specific stack and potential for error: luckily we're using a literal `1` for the `minutes` parameters (which becomes `tokens`).

Engine/AlgorithmTimeLimitManager.cs
```diff
@@ -0,0 +1,194 @@
...
+        /// <summary>
+        /// Attempts to requests additional time to continue executing the current time step.
+        /// At time of writing, this is intended to be used to provide training scheduled events
+        /// additional time to allow complex training models time to execute while also preventing
+        /// abuse by enforcing certain control parameters set via the job packet.
+        ///
+        /// Each time this method is invoked, this time limit manager will increase the allowable
+        /// execution time by the specified number of whole minutes
+        /// </summary>
+        public bool TryRequestAdditionalTime(int minutes)
```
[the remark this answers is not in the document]
mchandschuh: This is a `TryXXX` pattern. By convention, `TryXXX` methods don't throw exceptions and shouldn't corrupt the state of the object. If the consumer wants an error thrown if additional time can't be requested, then they need to call the `RequestAdditionalTime(int)` method.

Engine/RealTime/BacktestingRealTimeHandler.cs
```diff
@@ -17,6 +17,8 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
```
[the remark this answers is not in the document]
mchandschuh: hmm, 🤔 looks like I didn't do a good final review, I'm usually a stickler for dangling usings 😛

Engine/RealTime/BacktestingRealTimeHandler.cs
```diff
@@ -111,7 +115,16 @@ public void SetTime(DateTime time)
             // poke each event to see if it has fired, be sure to invoke these in time order
             foreach (var scheduledEvent in GetScheduledEventsSortedByTime())
             {
-                scheduledEvent.Scan(time);
+                if (scheduledEvent.NextEventUtcTime > time)
```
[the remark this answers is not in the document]
mchandschuh: This line is saying if the event's next time is in the future, then we don't need to call `Scan` on it. Further, we can stop checking the rest of the list since they're all sorted.

The line in `ScheduledEvent.Scan(DateTime)` is saying, if this event is at the present or in the past, then it should fire. Interestingly,  the two lines are _exactly_ opposite in their conditioning, such that, if a time passes one condition, it is guaranteed to _never_ pass the other condition.

Engine/RealTime/BacktestingRealTimeHandler.cs
mchandschuh: This statement tells us to break from the `foreach` loop, so if you have 1M scheduled events and they're _all_ in the future, we only check the first one. This is safe because they're already chronologically sorted. 

Regarding the sorted flag, I do believe you're correct. If the current event's time equals the provided `time`, then we won't fire any events but we would end up sorting again. Great eye.

Common/IsolatorLimitResultProvider.cs
mchandschuh: There's more than one caller and in `BacktestingRealTimeResultHandler` there's two difference guard clauses around the invocation. They actually have different intent as well. Also, in general I prefer to not make decisions in a method based on it's consumers. Now, the inverse isn't true -- so if there's a change that _should_ be made here (I'm not saying there is), then I would want to change the consumer of this function and not this function. The purposes of this guard clause is to prevent spinning up the task in the function in invokes just so the scheduled event can say 'its not time yet'.

Common/Util/RateLimit/BusyWaitSleepStrategy.cs
mchandschuh: It's actually part of the ported code from the referenced github repository. That being said, I think it rounds out the `ISleepStrategy` space and to be honest, I _really_ like the idea of an `ISleepStrategy`. Think about all the hard coded `Thread.Sleep( _n_ )` everywhere. I'm happy to move it to tests if there are strong opinions here, but in my opinion, it doesn't negatively impact anything and  rounds out the `ISleepStrategy` concept by showing some possible other implementations.

Common/Util/RateLimit/LeakyBucket.cs
mchandschuh: In this case, the resources we're rate limiting is CPU time itself, hence using the real time provider.

Common/Util/RateLimit/LeakyBucket.cs
mchandschuh: Leaky bucket is a rate limiting algorithm. So this `Consume` method is _arguably_ the reason the entire class exists.

Common/Util/RateLimit/TokenBucket.cs
mchandschuh: This provides a `TimeSpan` overload, merely for convenience.

Common/Util/RateLimit/ThreadSleepStrategy.cs
mchandschuh: In general, I think it's _very_ beneficial to design concepts and solutions in isolation from the current task. My general design philosophy works by first identifying the problem(s) that need to be solved. In this case, my design called for a leaky bucket implementation, so the first major commit in this PR is indeed, a full featured leaky bucket algorithm. After designing the component _at the core_ of this entire PR, I then proceeded to implement the specifics of the problem at hand.

Over the years I've learned that there is certainly balance between _just enough_ and *way* too much. In the case of these implementations, they're ported from the referenced github library, so they don't cost us anything, but provide solid abstractions and implementations to act as fundamental building blocks for future solutions. While I don't think everything should be componentized, I do find that development continues to get easier and easier if you follow this kind of approach due to the quality of the building blocks available to you.

Common/Util/RateLimit/LeakyBucket.cs
mchandschuh: We can consume all of the tokens but this condition is checking the opposite. It's asking if there's not enough tokens (requested more than available). I don't view this as an error condition since it's in a `TryXXX` method and the consumer has enough information at its disposal to perform the same logging, if desired.

Common/IsolatorLimitResultProvider.cs
mchandschuh: It sounds like you're referring to potential [nyquist] effects. I'm not sure which monitoring thread you're referring to as there's only a single thread checking times here. The maximum error in this check is based on the sleep interval of this loop, which is currently set to 5ms. So every 5ms we check if a full minute has passed, and if so, we request an extra minute. 

After writing the above, I think there's some misunderstanding here. As soon as the monitoring task starts (_very_ close to when the scheduled event starts), we start checking how long it's been running, every 5ms. When it's been running for a minimum of 1 full minute, we go and request an additional minute. The desired behavior here being that the first minute was free, but the _next_ one, you'll have to pay for. If there are minutes available in the bucket, then we increment to check the following minute, and so on and so forth. When the bucket runs out of minutes _and then_ you request another minute, the `_failed` flag will be flipped and the isolator will terminate the algorithm.

mchandschuh:
@Martin-Molinero -- thanks for the _epic_ review man 👍  While I folding the changes to `ITokenBucket` into it's original commit to keep that component code all in one spot, I addressed the other reviews in a separate commit to keep verification easier for you. Check it out here: 9d7c0a030

Algorithm.CSharp/TrainingScheduledRegressionAlgorithm.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// This regression algorithm is NOT expected to fail and verifies that a periodic training
+    /// event created in OnData will fire and will consume 'minutes' from the leaky bucket
+    /// </summary>
+    public class TrainingScheduledRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 7);
+            SetStartDate(2013, 10, 11);
```
AlexCatarino: This looks like a typo as it should be `SetEndDate(2013, 10, 11);`

mchandschuh: huh, how bout that! will fix, ty!

Algorithm.CSharp/TrainingInitializeRegressionAlgorithm.cs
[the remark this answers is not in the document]
Martin-Molinero: `takes 30+ on my machine` that's a lot! 😮 - `Note that today 155 algos take ~3 minutes to run` I wasn't counting the entire test suite but just the regression algorithms, I think setting `CanRunLocally = false` make sense

Common/IsolatorLimitResultProvider.cs
```diff
@@ -0,0 +1,110 @@
...
+                var next = timeProvider.GetUtcNow().AddMinutes(1);
+                while (Interlocked.Read(ref finished) == 0L)
+                {
+                    if (timeProvider.GetUtcNow() >= next)
+                    {
+                        // each minute request additional time from the isolator
+                        next = next.AddMinutes(1);
+
+                        // this will throw and notify the isolator that we've exceed the limits
+                        isolatorLimitProvider.RequestAdditionalTime(minutes: 1);
```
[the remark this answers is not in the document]
Martin-Molinero: Right, and since we are covered by the initial 10 min window there shouldn't be any issues 👍

Common/Util/RateLimit/ThreadSleepStrategy.cs
Martin-Molinero: Wondering which is the `referenced github library`?

mchandschuh: I left a note in the source [here](https://github.com/QuantConnect/Lean/pull/3743/files#diff-dc9ad5005cc9b09a90ed99a877a77eb2R26). It's based on this [`TokenBucket`](https://github.com/mxplusb/TokenBucket) dotnet core repository.

Issue #3739, opened 2019-10-16
SecurityCache performance


Martin-Molinero:
#### Expected Behavior

- `SecurityCache` is as performant as possible
#### Actual Behavior

- Currently we are caching twice the data by type, once just the last data point and later the entire data list `AlgorithmManager` -> `Security.Update()` https://github.com/QuantConnect/Lean/blob/master/Common/Securities/Security.cs#L609
-> `SecurityCache.StoreData()` https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityCache.cs#L225 `Security.SetMarketPrice()` -> SecurityCache.AddData() https://github.com/QuantConnect/Lean/blob/master/Common/Securities/SecurityCache.cs#L131

- Currently we storing custom data twice, once in the custom security and another in the underlying security (if any) https://github.com/QuantConnect/Lean/blob/master/Engine/AlgorithmManager.cs#L328
#### Potential Solution

- Refactor
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #3746, opened 2019-10-17
Performance improvements v3

Martin-Molinero:
#### Description

~PR on top of https://github.com/QuantConnect/Lean/pull/3714 performance improvements v2~
- `DynamicSecurityData` will be a view into the `SecurityCache` instance
- Refactors for `Security` and `SecurityCache` to avoid storing twice
the same data points in the data type cache
- Adding `SecurityCacheProvider` this class allows for two different
`Security` to share the same data type cache through different instance
of `SecurityCache`. This is used to directly access custom data types
through their underlying in a performant manner

Performance tests `BasicTemplateBenchmark C#`:
 ~`pr v2  https://github.com/QuantConnect/Lean/pull/3714`~ `master`
27.89 seconds at 60k data points per second. Processing total of 1,664,679 data points.
27.49 seconds at 61k data points per second. Processing total of 1,664,679 data points.
27.89 seconds at 60k data points per second. Processing total of 1,664,679 data points.
`this pr`
23.88 seconds at 70k data points per second. Processing total of 1,664,679 data points.
23.68 seconds at 70k data points per second. Processing total of 1,664,679 data points.
24.09 seconds at 69k data points per second. Processing total of 1,664,679 data points.

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3739
#### Motivation and Context

Need for speed

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Algorithm/Framework/Alphas/Analysis/InsightManager.cs
+++ b/Common/Algorithm/Framework/Alphas/Analysis/InsightManager.cs
@@ -211,7 +211,9 @@ namespace QuantConnect.Algorithm.Framework.Alphas.Analysis
         /// </summary>
         private void UpdateScores(ReadOnlySecurityValuesCollection securityValuesCollection)
         {
-            var removals = new List<InsightAnalysisContext>();
+            // for performance be lazy to initialize collection
+            List<InsightAnalysisContext> removals = null;
+
             foreach (var context in _openInsightContexts)
             {
                 // was this insight period closed before we update the times?
@@ -255,6 +257,10 @@ namespace QuantConnect.Algorithm.Framework.Alphas.Analysis
                     var id = context.Insight.Id;
                     _closedInsightContexts[id] = context;
 
+                    if (removals == null)
+                    {
+                        removals = new List<InsightAnalysisContext>();
+                    }
                     removals.Add(context);
                 }
 
@@ -262,7 +268,10 @@ namespace QuantConnect.Algorithm.Framework.Alphas.Analysis
                 _updatedInsightContexts.Add(context);
             }
 
-            _openInsightContexts.RemoveWhere(removals.Contains);
+            if (removals != null)
+            {
+                _openInsightContexts.RemoveWhere(removals.Contains);
+            }
         }
 
         /// <summary>Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.</summary>
--- a/Common/Securities/RegisteredSecurityDataTypesProvider.cs
+++ b/Common/Securities/RegisteredSecurityDataTypesProvider.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace QuantConnect.Securities
 {
@@ -68,7 +69,19 @@ namespace QuantConnect.Securities
         /// </summary>
         public bool TryGetType(string name, out Type type)
         {
-            return _types.TryGetValue(name, out type);
+            if (!_types.TryGetValue(name, out type))
+            {
+                // lets try a case insensitive search
+                var kvp = _types.FirstOrDefault(pair => pair.Key.ToLowerInvariant() == name.ToLowerInvariant());
+                if (kvp.Equals(default(KeyValuePair<string, Type>)))
+                {
+                    return false;
+                }
+
+                type = kvp.Value;
+                return true;
+            }
+            return true;
         }
     }
 }
--- a/Common/Securities/RegisteredSecurityDataTypesProvider.cs
+++ b/Common/Securities/RegisteredSecurityDataTypesProvider.cs
@@ -16,7 +16,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace QuantConnect.Securities
 {
@@ -31,7 +30,7 @@ namespace QuantConnect.Securities
         /// </summary>
         public static readonly IRegisteredSecurityDataTypesProvider Null = new RegisteredSecurityDataTypesProvider();
 
-        private readonly Dictionary<string, Type> _types = new Dictionary<string, Type>();
+        private readonly Dictionary<string, Type> _types = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Registers the specified type w/ the provider
@@ -69,19 +68,7 @@ namespace QuantConnect.Securities
         /// </summary>
         public bool TryGetType(string name, out Type type)
         {
-            if (!_types.TryGetValue(name, out type))
-            {
-                // lets try a case insensitive search
-                var kvp = _types.FirstOrDefault(pair => pair.Key.ToLowerInvariant() == name.ToLowerInvariant());
-                if (kvp.Equals(default(KeyValuePair<string, Type>)))
-                {
-                    return false;
-                }
-
-                type = kvp.Value;
-                return true;
-            }
-            return true;
+            return _types.TryGetValue(name, out type);
         }
     }
 }
--- a/Common/Securities/SecurityCache.cs
+++ b/Common/Securities/SecurityCache.cs
@@ -93,15 +93,6 @@ namespace QuantConnect.Securities
         public long OpenInterest { get; private set; }
 
         /// <summary>
-        /// True if the type cache being used is shared among different <see cref="SecurityCache"/> instances
-        /// </summary>
-        /// <remarks>This is useful for custom data securities which also have an underlying security,
-        /// will allow both securities to access the same data by type</remarks>
-        /// <remarks>This flag is particularly useful for performance since it helps determine
-        /// at runtime whether two securities should start sharing their type cache</remarks>
-        public bool IsTypeCacheShared { get; private set; }
-
-        /// <summary>
         /// Add a list of market data points to the local security cache for the current market price.
         /// </summary>
         /// <remarks>Internally uses <see cref="AddData"/> using the last data point of the provided list
@@ -335,9 +326,6 @@ namespace QuantConnect.Securities
                 sourceToShare._dataByType.TryAdd(kvp.Key, kvp.Value);
             }
             targetToModify._dataByType = sourceToShare._dataByType;
-
-            sourceToShare.IsTypeCacheShared = true;
-            targetToModify.IsTypeCacheShared = true;
         }
     }
 }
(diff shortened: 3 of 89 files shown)


Common/Algorithm/Framework/Alphas/Analysis/InsightManager.cs
```diff
@@ -255,14 +257,21 @@ private void UpdateScores(ReadOnlySecurityValuesCollection securityValuesCollect
...
                 }
 
                 // mark the context as having been updated
                 _updatedInsightContexts.Add(context);
             }
 
-            _openInsightContexts.RemoveWhere(removals.Contains);
+            if (removals != null)
+            {
+                _openInsightContexts.RemoveWhere(removals.Contains);
```
mchandschuh: For large numbers of insights, we might benefit from using a `HashSet<T>` here for `removals` to provide the `O(1)` `Contains` call.

Common/Securities/RegisteredSecurityDataTypesProvider.cs
```diff
@@ -30,7 +30,7 @@ public class RegisteredSecurityDataTypesProvider : IRegisteredSecurityDataTypesP
         /// </summary>
         public static readonly IRegisteredSecurityDataTypesProvider Null = new RegisteredSecurityDataTypesProvider();
 
-        private readonly Dictionary<string, Type> _types = new Dictionary<string, Type>();
+        private readonly Dictionary<string, Type> _types = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
mchandschuh: 👍

Common/Securities/SecurityCache.cs
```diff
@@ -97,6 +92,29 @@ public class SecurityCache
...
+        /// <summary>
+        /// Add a list of market data points to the local security cache for the current market price.
+        /// </summary>
+        /// <remarks>Internally uses <see cref="AddData"/> using the last data point of the provided list
+        /// and it stores by type the non fill forward points using <see cref="StoreData"/></remarks>
+        public void AddDataList(IReadOnlyList<BaseData> data, Type dataType, bool? containsFillForwardData = null)
+        {
+            var nonFillForwardData = data;
+            // maintaining regression requires us to NOT cache FF data
+            if (!containsFillForwardData.HasValue || containsFillForwardData.Value)
```
mchandschuh: maybe more readable:
``` csharp
if (containsFillForwardData != false)
{
}
```

Common/Securities/SecurityCache.cs
```diff
@@ -97,6 +92,29 @@ public class SecurityCache
...
+            {
+                nonFillForwardData = data.Where(baseData => !baseData.IsFillForward).ToList();
```
mchandschuh: I'd be curious how often we come down here -- if it's really often, might want to break this out into a `for` and manually create the list using the total size to seed it. This will prevent a potential extra list allocation when there's more than four items in the list (four is the default list capacity IIRC) and there's a surprising amount of overhead on LINQ methods, enumerators, function delegate invocation, stack popping up and down, etc...

Common/Securities/SecurityCacheProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+    public class SecurityCacheProvider
+    {
+        private readonly ConcurrentDictionary<SecurityIdentifier, ConcurrentBag<Symbol>> _relatedSymbols;
+        private readonly IDictionary<Symbol, Security> _securityProvider;
+
+        /// <summary>
+        /// Creates a new instance
+        /// </summary>
+        /// <param name="securityProvider">The security provider to use</param>
+        public SecurityCacheProvider(IDictionary<Symbol, Security> securityProvider)
```
mchandschuh: Why not use `ISecurityProvider` here?

Common/Securities/SecurityCacheProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        /// <summary>
+        /// Will return the <see cref="SecurityCache"/> instance to use for a give Symbol.
+        /// If the provided Symbol is a custom type which has an underlying we will try to use the
+        /// underlying SecurityCache type cache, if the underlying is not present we will keep track
+        /// of the custom Symbol in case it is added later.
+        /// </summary>
+        /// <returns>The cache instance to use</returns>
+        public SecurityCache GetSecurityCache(Symbol symbol)
```
mchandschuh: Very cool, love seeing this all in one place -- makes it super easy to grok. The intent is very clear here. Great work 👍

Common/Securities/SecurityCacheProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <summary>
+    /// A helper class that will provide <see cref="SecurityCache"/> instances
+    /// </summary>
+    /// <remarks>The value of this class and its logic is performance.
+    /// This class allows for two different <see cref="Security"/> to share the same
+    /// data type cache through different instances of <see cref="SecurityCache"/>.
+    /// This is used to directly access custom data types through their underlying</remarks>
+    public class SecurityCacheProvider
+    {
+        private readonly ConcurrentDictionary<SecurityIdentifier, ConcurrentBag<Symbol>> _relatedSymbols;
```
mchandschuh: not sure if there's a difference between using a `Symbol` vs `SecurityIdentifier` as a dictionary key -- I think `Symbol` delegates it's `GetHashCode` and `Equals` methods to `SecurityIdentifier`.

Common/Securities/SecurityCacheProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    // we didn't find the underlying, lets keep track of the underlying symbol which might get added in the future.
+                    // else when it is added, we would have to go through existing Securities and find any which use it as underlying
+                    if (_relatedSymbols.ContainsKey(symbol.Underlying.ID))
+                    {
+                        _relatedSymbols[symbol.Underlying.ID].Add(symbol);
+                    }
+                    else
+                    {
+                        _relatedSymbols[symbol.Underlying.ID] = new ConcurrentBag<Symbol> { symbol };
```
mchandschuh: is there any chance of this line running on multiple threads? If so, we should prefer to use
``` csharp
_relatedSymbols.AddOrUpdate(symbol.Underlying.ID,
    id => new ConcurrentBag<Symbol>{symbol},
    (id, bag) => { bag.Add(symbol); return bag; }
);
````

File: Common/Securities/SecurityCacheProvider.cs [hunk 9]
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                }
+            }
+            else if(_relatedSymbols.ContainsKey(symbol.ID))
```
mchandschuh: Why perform a `ContainsKey` check to immediately afterwards perform a `TryRemove` -- shouldn't we just perform the `TryRemove` first, which is essentially an atomic `ContainsKey` + `Remove` when true. The issue is, between `ContainsKey` returning true and invoking `TryRemove`, it may have already been removed -- which might not be an issue. Another thought in favor of collapsing these into a single `TryRemove` call is why perform the extra `ContainsKey` if it's not required?

mchandschuh: to keep your current bracing structure, you can actually just change this `else if (....)` to an `else` and it should behave the same, just without the extra check a potential race between these checks.

Engine/AlgorithmManager.cs
```diff
@@ -294,11 +289,8 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                     foreach (var security in timeSlice.SecurityChanges.AddedSecurities)
                     {
                         security.IsTradable = true;
-                        if (!algorithm.Securities.ContainsKey(security.Symbol))
-                        {
-                            // add the new security
-                            algorithm.Securities.Add(security);
-                        }
+                        // uses TryAdd, so don't need to worry about duplicates here
```
mchandschuh: I dig the comments -- _super_ useful when making changes for performance which can sometimes make code look _wrong_ -- nice!

Engine/DataFeeds/Enumerators/SynchronizingEnumerator.cs
```diff
@@ -200,7 +199,7 @@ private static void ListInsert(SynchronizedEnumerator[] list, int index, Synchro
         private static IEnumerator<BaseData> GetBruteForceMethod(IEnumerator<BaseData>[] enumerators)
         {
             var ticks = DateTime.MaxValue.Ticks;
-            var collection = new ConcurrentDictionary<IEnumerator<BaseData>, int>();
+            var collection = new Dictionary<IEnumerator<BaseData>, int>();
```
mchandschuh: Why not just use a `HashSet<T>`-- it's faster than a dictionary and it doesn't look like you're they the key/value -- just the key here.

Common/Securities/SecurityCache.cs
Martin-Molinero: Good idea - will do 👍

Common/Securities/SecurityCacheProvider.cs
Martin-Molinero: True, I'll use `Symbol` for consistency with the rest of Lean 👍

Common/Securities/SecurityCacheProvider.cs
Martin-Molinero: Another good idea 👍

Common/Securities/SecurityCacheProvider.cs
Martin-Molinero: Will do 👍 - it's even cleaner

Common/Securities/SecurityCacheProvider.cs
Martin-Molinero: 👍 will do

Engine/DataFeeds/Enumerators/SynchronizingEnumerator.cs
Martin-Molinero: Good idea 👍

Common/Securities/SecurityCacheProvider.cs
```diff
@@ -63,26 +62,22 @@ public SecurityCache GetSecurityCache(Symbol symbol)
...
-                if (_relatedSymbols.TryRemove(symbol.ID, out customSymbols))
+                if (_relatedSymbols.TryRemove(symbol, out customSymbols))
                 {
+                    // if we are here it means we added a symbol which is an underlying of some existing custom symbols
                     foreach (var customSymbol in customSymbols)
                     {
-                        Security customSecurity;
-                        if (_securityProvider.TryGetValue(customSymbol, out customSecurity))
+                        var customSecurity = _securityProvider.GetSecurity(customSymbol);
+                        if (customSecurity != null)
```
mchandschuh: no need to change -- I think at some point I added a `TryGetValue` extension method to maintain the beauty of the try pattern.

PR #3748, opened 2019-10-18
Fixes failing CBOE History request by removing duplicate property

### LINKED ISSUE #3745: PandasData Throws Duplicate Keys Exception with CBOE Class
#### Expected Behavior
No duplicate keys exception in `PandasData` for QC Custom data classes.

#### Actual Behavior
`PandasData` throws duplicate keys exception with `CBOE` class.

#### Potential Solution
Removed [CBOE.cs#L27:L30](https://github.com/QuantConnect/Lean/blob/master/Common/Data/Custom/CBOE/CBOE.cs#L27:L30)

#### Note
Removing those lines solved this issue, but `PandasData.ToPandasDataFrame` thowns another exception:
```Python.Runtime.PythonException: 'ValueError : cannot handle a non-unique multi-index!'```

#### Reproducing the Problem
Add this statement to `CachedAlternativeDataAlgorithm.py`
```python
df = self.History([self.cboeVix], 50, Resolution.Daily)
```
#### Checklist

AlexCatarino:
#### Description
When custom data classes inherited from market data classes such as `TradeBar`, it created duplicate entries. Therefore, we need to exclude the common properties in the private field `PandasData._members`.

#### Related Issue
Closes #3745 

#### Motivation and Context
Bug fix.

Unit tests and running `CachedAlternativeDataAlgorithm.py` with
```python
df = self.History([self.cboeVix], 50, Resolution.Daily)
```
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Python/PandasData.cs
+++ b/Common/Python/PandasData.cs
@@ -245,8 +245,9 @@ class Remapper(wrapt.ObjectProxy):
                 }
             }
 
-            var type = data.GetType() as Type;
-            _isCustomData = type.Namespace != "QuantConnect.Data.Market";
+            var type = data.GetType();
+            var dataMarketNamespace = typeof(Bar).Namespace;
+            _isCustomData = type.Namespace != dataMarketNamespace;
             _members = Enumerable.Empty<MemberInfo>();
             _symbol = (data as IBaseData)?.Symbol;
 
@@ -276,7 +277,14 @@ class Remapper(wrapt.ObjectProxy):
                         throw new ArgumentException($"PandasData.ctor(): More than one \'{duplicateKey}\' member was found in \'{type.FullName}\' class.");
                     }
 
-                    keys = members.Select(x => x.Name.ToLowerInvariant()).Except(_baseDataProperties).Concat(new[] { "value" });
+                    // If the custom data derives from a Market Data (e.g. Tick, TradeBar, QuoteBar), exclude its keys
+                    var baseType = type.BaseType;
+                    var inheritedProperties = baseType != null && baseType.Namespace == dataMarketNamespace
+                        ? baseType.GetProperties().ToHashSet(x => x.Name.ToLowerInvariant())
+                        : _baseDataProperties;
+
+                    keys = members.Select(x => x.Name.ToLowerInvariant()).Except(inheritedProperties).Concat(new[] {"value"});
+
                     _members = members.Where(x => keys.Contains(x.Name.ToLowerInvariant()));
                 }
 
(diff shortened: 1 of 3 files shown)


Common/Python/PandasData.cs
```diff
@@ -276,7 +277,14 @@ class Remapper(wrapt.ObjectProxy):
...
                     }
 
-                    keys = members.Select(x => x.Name.ToLowerInvariant()).Except(_baseDataProperties).Concat(new[] { "value" });
+                    // If the custom data derives from a Market Data (e.g. Tick, TradeBar, QuoteBar), exclude its keys
+                    var baseType = type.BaseType;
+                    var inheritedProperties = baseType != null && baseType.Namespace == dataMarketNamespace
+                        ? baseType.GetProperties().ToHashSet(x => x.Name.ToLowerInvariant())
+                        : _baseDataProperties;
+
+                    keys = members.Select(x => x.Name.ToLowerInvariant()).Except(inheritedProperties).Concat(new[] {"value"});
```
Martin-Molinero: `members` is being re enumerated multiple times we probably want to convert it to a list?

If the issue is duplicated `keys` why not make it a set directly? 🤔 After a quick test I think its `columns` that ends up holding duplicates in `master` and causes the exception?

AlexCatarino: What caused the exception was the fact that `_members` had `TradeBar` members. In this block [PandasData.cs#L296:L302](https://github.com/QuantConnect/Lean/blob/master/Common/Python/PandasData.cs#L296:L302), we were updating "open", "high", "low", "close" and "volume". After that, in [PandasData.cs#L319](https://github.com/QuantConnect/Lean/blob/master/Common/Python/PandasData.cs#L319), `tradeBar` would not be `null` since CBOE inherits from `TradeBar` and we would update "open", "high", "low", "close" and "volume" again.
So the idea is to remove the `TradeBar` properties from `_members`.

Martin-Molinero: Thanks for the explanation. The current solution wouldn't cover another level on inheritance, for example, a custom type inheriting from CBOE (it's a good unit test to add) since we are relying on the base type and its namespace.
Since this is just for `QuoteBar/TradeBar/Tick` https://github.com/QuantConnect/Lean/blob/master/Common/Python/PandasData.cs#L319 maybe the solution could be aware of it too - Simple one could be casting to `QuoteBar/TradeBar/Tick` (as done later) and checking if not null and removing those members 🤔 

While reviewing I realized that we are enumerating all the data, not sure the reason https://github.com/QuantConnect/Lean/pull/3748/files#diff-92c2966a841a7cd2c63b5a1ae6c9bffaR239-R246

PR #3750, opened 2019-10-19
Feature fred custom data

Jay-Jay-D:
#### Description

This PR implements a consumer for the [Federal Reserve Economic Data](https://fred.stlouisfed.org/) (FRED). For this PR approximately 500 new daily data series were added. 

#### Related Issue

N/A

#### Motivation and Context

Because more data is better.

Add new series to the documentation.

The helper classes implemented contains a decent amount of information for all series and their sources.

Added new test

#### Types of changes

#### Checklist:


Common/Data/Custom/Fred/Fred.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// <returns>New instance of USEnergy</returns>
+        public override BaseData Reader(SubscriptionDataConfig config, string line, DateTime date, bool isLiveMode)
+        {
+            var csv = line.Split(',');
+
+            return new Fred
+            {
+                Symbol = config.Symbol,
+                Value = Parse.Decimal(csv[1]),
+                Time = Parse.DateTimeExact(csv[0], "yyyyMMdd")
```
gsalaz98: We need to be careful here with the `Time` property since the time that is reported may be during the day, and not at `Time` @ 00:00:00. Perhaps it releases at `Time` @ 17:00:00? 

Either way, depending on the time they report the data, we might have to add a delay of a whole day to the time if we can't pin down the time the data is released.

Common/Data/Custom/Fred/FredApi.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <param name="isLiveMode">if set to <c>true</c> [is live mode].</param>
+        /// <returns></returns>
+        public override BaseData Reader(SubscriptionDataConfig config, string content, DateTime date, bool isLiveMode)
+        {
+            var series = JsonConvert.DeserializeObject<FredApi>(content).Observations;
+            var objectList = new List<FredApi>();
+            foreach (var observation in series)
+            {
+                decimal value;
+                if (decimal.TryParse(observation.Value, NumberStyles.Any, CultureInfo.InvariantCulture, out value))
```
gsalaz98: I'm unsure if it was available when you first started working on this, but I've added support for `Parse.TryParse(decimal, NumberStyles, out decimal value)` in the `Parse` class 😄

Common/Data/Custom/Fred/FredApi.cs
```diff
@@ -0,0 +1,149 @@
...
+                {
+                    objectList.Add(new FredApi
+                    {
+                        Symbol = config.Symbol,
+                        Time = observation.Date - TimeSpan.FromDays(1),
```
gsalaz98: Hmmm, maybe it's forward adjusted so that no lookahead bias is introduced (same as comment above regarding the `Time` property) 🤔

Common/Data/Custom/Fred/Wilshire.cs
```diff
@@ -0,0 +1,351 @@
...
+        public static class Wilshire
+        {
+            ///<summary>
+            /// Wilshire US Small-Cap Value Price Index (in Index)
+            /// Retrieved from FRED, Federal Reserve Bank of St. Louis; https://fred.stlouisfed.org/series/WILLSMLCAPVALPR
+            /// </summary>
+            /// <remarks>
+            /// The price indexes are price returns, which do not reinvest dividends. Copyright, 2016, Wilshire Associates Incorporated. Reprinted with permission. More information about the various indexes from Wilshire Associates can be found at http://www.wilshire.com/Indexes/.
+            /// </remarks>
+            public static string WilshireUSSmallCapValuePrice => "WILLSMLCAPVALPR";
```
gsalaz98: Redundant `Wilshire` in all variable names even though the class is named `Wilshire`. We'd be accessing it via `FredSeries.Wilshire.Wilshire...`

Common/Data/Custom/Fred/CBOE.cs
```diff
@@ -0,0 +1,213 @@
...
+
+namespace QuantConnect.Data.Custom.Fred
+{
+    public static partial class FredSeries
+    {
+        public static class CBOE
+        {
+            ///<summary>
+            /// CBOE Equity VIX on Google (in Index)
+            /// Retrieved from FRED, Federal Reserve Bank of St. Louis; https://fred.stlouisfed.org/series/VXGOGCLS
```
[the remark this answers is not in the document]
Jay-Jay-D: FRED actively asks for citations of all data they distribute.

Also, those pages have extra information that can be useful for our users, e.g. tags and related series.

Common/Data/Custom/Fred/Fred.cs
Jay-Jay-D: There are no specifications about the time of the day each series is reported.

Common/Data/Custom/Fred/FredApi.cs
gsalaz98: As per discussion in Slack (posting here to explain my comment since it's a little ambiguous right now):

So, because `new DateTime(yyyy, MM, dd)` sets the time to midnight by default if no Time component is passed, consider the following:

* Data from FRED, let's say VIX, is reported as `2019-09-29`, but what it actually refers to is `2019-09-29 17:00:00` (but they leave the time component out because it's daily data)

* We parse the date as `2019-09-29` because we have no time component (`new DateTime(2019, 9, 29);`)

* The `BaseData` instance will now have `2019-09-29 00:00:00` when the data was actually reported at `2019-09-29 17:00:00`

* The data point will be emitted at `2019-09-29 00:00:00` in backtests, basically getting us the value from the end of the day before it was even generated.
What I meant in that comment was that maybe they forward-adjust their times (think `Math.Ceil` but for time) so that if we parse it as `2019-09-30`, it will be for `2019-09-30 00:00:00` although the data point might actually be at `2019-09-29 17:00:00`

jaredbroad: The LEAN solution to this is to set a period for the data point and leave the Time property as the start time of the bar; the EndTime is automatically set to Time+Period.

Common/Data/Custom/Fred/CBOE.cs
```diff
@@ -0,0 +1,213 @@
...
+            /// CBOE Equity VIX on Amazon (in Index)
+            /// Retrieved from FRED, Federal Reserve Bank of St. Louis; https://fred.stlouisfed.org/series/VXAZNCLS
+            /// </summary>
+            /// <remarks>
+            /// Copyright, 2016, Chicago Board Options Exchange, Inc. Reprinted with permission.
+            /// </remarks>
+            public static string VIXOnAmazon => "VXAZNCLS";
+
+            ///<summary>
+            /// CBOE S&P 100 Volatility Index: VXO (in Index)
```
jaredbroad: Please check the comments for `&` and replace with the HTML entities version; 
https://stackoverflow.com/questions/19794929/vs2010-comments-with-an-ampersand-get-marked-as-warnings

Engine/DataFeeds/SubscriptionDataReader.cs
```diff
@@ -208,6 +209,16 @@ public void Initialize()
                 }
             }
 
+            // If Fred data, set the access token in data factory
+            var fred = _dataFactory as FredApi;
```
jaredbroad: One day we'll add an interface for this and the world will breathe a sigh of relief from setting all the auth codes generically. 🦄 🌈 🌞

Common/Data/Custom/Fred/CBOE.cs
```diff
@@ -0,0 +1,213 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+namespace QuantConnect.Data.Custom.Fred
+{
+    public static partial class FredSeries
```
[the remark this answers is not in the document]
Jay-Jay-D: In this case, we have ~500 series divides into 8 categories, and many more can be added.

I consider it makes sense to have a separated helper class. On the other hand, maybe the helper class' name can be improved.

Common/Data/Custom/Fred/FredApi.cs
Jay-Jay-D: Great insight @gsalaz98!

Common/Data/Custom/Fred/Wilshire.cs
Jay-Jay-D: I left `Wilshire` in this case because many of the properties start with a number, e.g.  `Willshire2500GrowthPrice` even when it is redundant, I consider it is better than `Willshire.TwoThousandFiveHundredGrowthPrice`.

Common/Data/Custom/Fred/Fred.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// Clones the data
+        /// </summary>
+        /// <returns>A clone of the object</returns>
+        public override BaseData Clone()
+        {
+            return new Fred()
+            {
+                Symbol = Symbol,
+                Value = Value,
+                Time = Time
```
jaredbroad: +Period setting,

Common/Data/Custom/Fred/Fred.cs
jaredbroad: for daily:
EndTime=Time.Add(TimeSpan.FromDays(1));

Common/Data/Custom/Fred/ICEBofAML.cs
```diff
@@ -0,0 +1,2191 @@
...
+            /// The index data referenced herein is the property of ICE Data Indices, LLC, its affiliates, ("ICE") and/or its Third Party Suppliers and has been licensed for use by the Federal Reserve Bank of St. Louis. ICE, its affiliates and Third Party Suppliers accept no liability in connection with its use.
+            /// Copyright, 2017, ICE Benchmark Administration. Reprinted with permission.
+            /// </remarks>
+            public static string AsiaEmergingMarketsCorporatePlusSubIndexTotalReturnIndexValue => "BAMLEMRACRPIASIATRIV";
+
+            ///<summary>
+            /// ICE BofAML Asia US Emerging Markets Liquid Corporate Plus Sub-Index Total Return Index Value (in Index)
+            /// Retrieved from FRED, Federal Reserve Bank of St. Louis; https://fred.stlouisfed.org/series/BAMLEMALLCRPIASIAUSTRIV
```
jaredbroad: Please move references to first line of remarks

PR #3751, opened 2019-10-19
Added SimpleInsightPortfolioConstructionModel to the Framework.

wardjm:
#### Description

This commit aims to add a portfolio model that can be used for the competition. The model works as follows:
For insights of direction InsightDirection.Up, long targets are returned and
    for insights of direction InsightDirection.Down, short targets are returned.
    No rebalancing shall be done, as a new insight or the age of the insight shall determine whether to exit
    the positions.
    Rules:
        1. On Up insight, increase position size by percent
        2. On Down insight, decrease position size by percent
        3. On Flat insight, move by percent towards 0
        4. On expired insight, perform a Flat insight'''

#### Motivation and Context

There currently is no portfolio model that can be used to test alphas directly related to insight production. They all involve rebalancing or equal weighting. This portfolio will initiate a new trade with every new insight. So 20 Up insights will generate 20 longs. They will be sold when the insight expires. They will be sold when a Flat insight is generated for the symbol. This is the most simple and direct approach for testing alphas. Of course a real portfolio would be more sophisticated and smarter about risk management.

Only user guides that have yet to be written.

This code is self-contained and must be imported to be used. The model was test on various time frames and universes directly on quantconnect.

#### Types of changes

#### Checklist:

- [X ] My code follows the code style of this project.
- [X ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
I do not see tests for other portfolio models.
- [X ] All new and existing tests passed.
- [X ] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`


jaredbroad:
Interesting idea thanks for the PR @wardjm; the core of your issue with EWPCM we're working on addressing in other issues (PR #3762, PR #3758, Issue #3763), but this is a new concept. Rather than "Simple" let's call it "Incremental" or "Accumulative" to be more specific. We'll take this root PR and extend it to add the C# implementation. I'll leave it open until we have the bandwidth to do that build-out.

wardjm:
Thanks! I saw those PRs go through. I think those will be really beneficial though a slightly different style than I'm proposing with this. I noticed after some poking around and viewing those other PRs that the tests were done in C#. I know nothing about C# but could try to work on something myself (although I'm pretty swamped at the moment myself). I'm starting to get more familiar with the whole layout of LEAN, and I don't want to put more work on you guys for something I'm interested in seeing done.

I was thinking of Simple because of the two follow-on classes that can inherit from this one, but Accumulative is more descriptive. To document here, the follow-on that I had in mind when writing this...
1. Something like WeightedAccumulativeInsightPortfolioConstructionModel, which could use the magnitude and confidence to take positions off after a profit target is reached.
2. Something like LongOnlyWeightedAccumulativeInsightPortfolioConstructionModel which could inherit from Weighted... but allow Down insights that truncate at flat positions, so you never take a short. I definitely don't see this getting done before the competition ends though, so I'm not sure there's as much of a use for it in general. Maybe people are interested still.

wardjm:
Hey, just commenting on progress. I've added some tests, and realized there were some extra files that needed to be updated and such. I'll give the C# a shot since I can use the tests to verify. I should be able to issue a new more complete pull request soon.

wardjm:
Sorry I think I messed up the PR. I'm going to submit a new one that is clean and reference this one.

jaredbroad:
Nice work @wardjm! Superceeded by #3863

Issue #3757, opened 2019-10-22
Adds More Refresh Options to EqualWeightingPortfolioContructionModel


AlexCatarino:
#### Expected Behavior
EWPCM can refresh the portfolio targets with different periods.

#### Actual Behavior
EWPCM can only refresh the portfolio targets with the periods from Resolution enum.

#### Potential Solution
Add constructor overload that accepts TimeSpan.

#### Reproducing the Problem
N/A

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #3758, opened 2019-10-22
Extends EWPCM Rebalancing Options

AlexCatarino:
#### Description
Adds constructor overloads to `EqualWeightingPortfolioConstructionModel` (`EWPCM`) to allow different rebalancing definitions.

It is possible to define rebalancing period with `Resolution`, `TimeSpan` (`timedelta` for Python) or a `Func<DateTime, DateTime>` (`lambda x: x+timedelta(y)` for Python). The last option lets the model use `Expiry` helper class with the members such as `EndOfWeek` and `EndOfMonth`.

#### Related Issue
Closes https://github.com/QuantConnect/Lean/issues/3757

#### Motivation and Context
The current overload limits the rebalancing periods to only the `Resolution` enum.

Yes. Show how `EWPCM` can be used with other overloads.

Unit tests. Added unit test for the new overloads.
Since the overload that accepts the `Resolution` object is converted into `Func<DateTime, DateTime>` (`lambda x: x+timedelta(y)` for Python), the new overloads are tested by consequence. 

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
+++ b/Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
@@ -33,7 +33,7 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
         /// <summary>
         /// Initialize a new instance of <see cref="InsightWeightingPortfolioConstructionModel"/>
         /// </summary>
-        /// <param name="rebalancingFunc">Rebalancing function</param>
+        /// <param name="rebalancingFunc">For a given algorithm UTC DateTime returns the next expected rebalance time</param>
         public ConfidenceWeightedPortfolioConstructionModel(Func<DateTime, DateTime> rebalancingFunc)
             : base(rebalancingFunc)
         {
--- a/Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
+++ b/Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
@@ -13,6 +13,7 @@
  * limitations under the License.
 */
 
+using System;
 using QuantConnect.Algorithm.Framework.Alphas;
 
 namespace QuantConnect.Algorithm.Framework.Portfolio
@@ -29,6 +30,24 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
     /// </summary>
     public class ConfidenceWeightedPortfolioConstructionModel : InsightWeightingPortfolioConstructionModel
     {
+        /// <summary>
+        /// Initialize a new instance of <see cref="InsightWeightingPortfolioConstructionModel"/>
+        /// </summary>
+        /// <param name="rebalancingFunc">Rebalancing function</param>
+        public ConfidenceWeightedPortfolioConstructionModel(Func<DateTime, DateTime> rebalancingFunc)
+            : base(rebalancingFunc)
+        {
+        }
+
+        /// <summary>
+        /// Initialize a new instance of <see cref="InsightWeightingPortfolioConstructionModel"/>
+        /// </summary>
+        /// <param name="timeSpan">Rebalancing frequency</param>
+        public ConfidenceWeightedPortfolioConstructionModel(TimeSpan timeSpan)
+            : base(timeSpan)
+        {
+        }
+
         /// <summary>
         /// Initialize a new instance of <see cref="ConfidenceWeightedPortfolioConstructionModel"/>
         /// </summary>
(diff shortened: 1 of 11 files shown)


Tests/Algorithm/Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModelTests.cs
```diff
@@ -354,6 +354,15 @@ public void GeneratesZeroTargetForZeroInsightConfidence(Language language)
             AssertTargets(actualTargets, new[] {new PortfolioTarget(Symbols.SPY, 0)});
         }
 
+        [Test]
+        [TestCase(Language.CSharp)]
+        [TestCase(Language.Python)]
+        public void DoesNotThrowWithAlternativeOverloads(Language language)
+        {
+            Assert.DoesNotThrow(() => SetPortfolioConstruction(language, _algorithm, TimeSpan.FromDays(1)));
+            Assert.DoesNotThrow(() => SetPortfolioConstruction(language, _algorithm, Expiry.EndOfWeek));
```
Martin-Molinero: Why not add the `Resolution` overload, just to be on the safe side. The unit test shouldn't really be aware of the underlying implementation, and not test because of it

Tests/Algorithm/Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModelTests.cs
```diff
@@ -401,6 +410,36 @@ private void SetPortfolioConstruction(Language language, QCAlgorithm algorithm)
...
+                {
+                    var name = nameof(ConfidenceWeightedPortfolioConstructionModel);
+                    var instance = Py.Import(name).GetAttr(name).Invoke(span.ToPython());
+                    var model = new PortfolioConstructionModelPythonWrapper(instance);
+                    algorithm.SetPortfolioConstruction(model);
+                }
+            }
+        }
+
+        private void SetPortfolioConstruction(Language language, QCAlgorithm algorithm, Func<DateTime, DateTime> func)
```
Martin-Molinero: Could make the last argument `dynamic argument` and use `((object)argument).ToPython()` to avoid the method duplication

Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.py
```diff
@@ -29,15 +29,22 @@ class EqualWeightingPortfolioConstructionModel(PortfolioConstructionModel):
...
         self.removedSymbols = []
         self.nextExpiryTime = UTCMIN
         self.rebalancingTime = UTCMIN
-        self.rebalancingPeriod = Extensions.ToTimeSpan(resolution)
+        self.rebalancingFunc = rebalancingParam
+
+        # If the argument is an instance if Resolution or Timedelta
+        # Redefine self.rebalancingFunc
+        if isinstance(rebalancingParam, int):
+            self.rebalancingFunc = lambda dt: dt + Extensions.ToTimeSpan(rebalancingParam)
```
Martin-Molinero: Maybe we could move `Extensions.ToTimeSpan(rebalancingParam)` outside of the lambda so we don't run it each time

Algorithm.Framework/Portfolio/ConfidenceWeightedPortfolioConstructionModel.cs
```diff
@@ -29,6 +30,24 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
     /// </summary>
     public class ConfidenceWeightedPortfolioConstructionModel : InsightWeightingPortfolioConstructionModel
     {
+        /// <summary>
+        /// Initialize a new instance of <see cref="InsightWeightingPortfolioConstructionModel"/>
+        /// </summary>
+        /// <param name="rebalancingFunc">Rebalancing function</param>
```
Martin-Molinero: I think we need a bit more documentation on this overload, specifically what the function is expected to do, eg: 'for a given algorithm UTC DateTime return the next expected rebalance time'

Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.cs
```diff
@@ -29,19 +29,37 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
     /// </summary>
     public class EqualWeightingPortfolioConstructionModel : PortfolioConstructionModel
     {
+        private readonly Func<DateTime, DateTime> _rebalancingFunc;
```
Martin-Molinero: Nice solution! 👍

jaredbroad:
@AlexCatarino please add an example using this new feature, perhaps even if its commented out in one of the existing examples.

PR #3779, opened 2019-10-29
Implements Benzinga Custom Data

### LINKED ISSUE #3775: Implement Benzinga Custom Data
#### Expected Behavior

We can use Benzinga data and process it
#### Actual Behavior

We can't use Benzinga data or process it
#### Potential Solution

Implement a processor and BaseData implementation
#### Reproducing the Problem

Try to use Benzinga data.
#### System Information

N/A
#### Checklist

gsalaz98:
#### Description

Adds Benzinga custom data

* Adds ToolBox processing application
* Adds BaseData implementation and example algorithms
#### Related Issue

Closes #3775
#### Motivation and Context

Implement additional alternative data sources for users to use

Requires new documentation

Local testing
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
+++ b/Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
@@ -1,0 +1,97 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Data;
+using QuantConnect.Data.Custom.Benzinga;
+using QuantConnect.Data.UniverseSelection;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    public class BenzingaNewsAlgorithm : QCAlgorithm
+    {
+        // Predefine a dictionary of words with scores to scan for in the description
+        // of the news article
+        private readonly Dictionary<string, double> _words = new Dictionary<string, double>()
+        {
+            {"bad", -0.5}, {"good", 0.5},
+            {"negative", -0.5}, {"great", 0.5},
+            {"growth", 0.5}, {"fail", -0.5},
+            {"failed", -0.5}, {"success", 0.5},
+            {"nailed", 0.5}, {"beat", 0.5},
+            {"missed", -0.5}, {"slipped", -0.5},
+            {"outperforming", 0.5}, {"underperforming", -0.5},
+            {"outperform", 0.5}, {"underperform", -0.5}
+        };
+
+        /// <summary>
+        /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
+        /// </summary>
+        public override void Initialize()
+        {
+            SetStartDate(2018, 10, 12);
+            SetEndDate(2018, 11, 25);
+            SetCash(100000);
+
+            UniverseSettings.Resolution = Resolution.Daily;
+            AddUniverseSelection(new CoarseFundamentalUniverseSelectionModel(CoarseSelector));
+        }
+
+        public IEnumerable<Symbol> CoarseSelector(IEnumerable<CoarseFundamental> coarse)
+        {
+            // Add Benzinga news data from the filtered coarse selection
+            var symbols = coarse.Where(x => x.HasFundamentalData && x.DollarVolume > 50000000)
+                .Select(x => x.Symbol)
+                .Take(10);
+
+            foreach (var symbol in symbols)
+            {
+                AddData<BenzingaNews>(symbol);
+            }
+
+            return symbols;
+        }
+
+        public override void OnData(Slice data)
+        {
+            // Get all Benzinga data and loop over it
+            foreach (var article in data.Get<BenzingaNews>().Values)
+            {
+                // Get the list of matching words we have sentiment definitions for
+                var intersection = article.Contents.ToLowerInvariant()
+                    .Split(' ')
+                    .Intersect(_words.Keys);
+
+                // Get the sentiment score
+                var sentimentScore = intersection.Select(x => _words[x]).Sum();
+
+                // Set holdings equal to 1/10th of the sentiment score we get
+                SetHoldings(article.Symbol.Underlying, sentimentScore / 10);
+            }
+        }
+
+        public override void OnSecuritiesChanged(SecurityChanges changes)
+        {
+            foreach (var r in changes.RemovedSecurities.Where(x => x.Symbol.SecurityType == SecurityType.Equity))
+            {
+                // If removed from the universe, liquidate and remove the custom data from the algorithm
+                Liquidate(r.Symbol);
+                RemoveSecurity(QuantConnect.Symbol.CreateBase(typeof(BenzingaNews), r.Symbol, Market.USA));
+            }
+        }
+    }
+}
--- a/Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
+++ b/Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
@@ -13,6 +13,7 @@
  * limitations under the License.
 */
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Algorithm.Framework.Selection;
@@ -20,7 +21,7 @@ using QuantConnect.Data;
 using QuantConnect.Data.Custom.Benzinga;
 using QuantConnect.Data.UniverseSelection;
 
-namespace QuantConnect.Algorithm.CSharp
+namespace QuantConnect.Algorithm.CSharp.AltData
 {
     public class BenzingaNewsAlgorithm : QCAlgorithm
     {
@@ -42,7 +43,8 @@ namespace QuantConnect.Algorithm.CSharp
             // Add Benzinga news data from the filtered coarse selection
             var symbols = coarse.Where(x => x.HasFundamentalData && x.DollarVolume > 50000000)
                 .Select(x => x.Symbol)
-                .Take(10);
+                .Take(10)
+                .ToList();
 
             foreach (var symbol in symbols)
             {
@@ -64,7 +66,7 @@ namespace QuantConnect.Algorithm.CSharp
                 var selectedSymbol = article.Symbols.SingleOrDefault(x => x.Symbol == article.Symbol.Underlying);
                 if (selectedSymbol == null)
                 {
-                    continue;
+                    throw new Exception($"Could not find current Symbol {article.Symbol.Underlying} even though it should exist");
                 }
 
                 // Sometimes sentiment is not included with the article by Benzinga.
--- a/Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
+++ b/Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
@@ -24,20 +24,6 @@ namespace QuantConnect.Algorithm.CSharp
 {
     public class BenzingaNewsAlgorithm : QCAlgorithm
     {
-        // Predefine a dictionary of words with scores to scan for in the description
-        // of the news article
-        private readonly Dictionary<string, double> _words = new Dictionary<string, double>()
-        {
-            {"bad", -0.5}, {"good", 0.5},
-            {"negative", -0.5}, {"great", 0.5},
-            {"growth", 0.5}, {"fail", -0.5},
-            {"failed", -0.5}, {"success", 0.5},
-            {"nailed", 0.5}, {"beat", 0.5},
-            {"missed", -0.5}, {"slipped", -0.5},
-            {"outperforming", 0.5}, {"underperforming", -0.5},
-            {"outperform", 0.5}, {"underperform", -0.5}
-        };
-
         /// <summary>
         /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
         /// </summary>
@@ -71,22 +57,32 @@ namespace QuantConnect.Algorithm.CSharp
             // Get all Benzinga data and loop over it
             foreach (var article in data.Get<BenzingaNews>().Values)
             {
-                // Get the list of matching words we have sentiment definitions for
-                var intersection = article.Contents.ToLowerInvariant()
-                    .Split(' ')
-                    .Intersect(_words.Keys);
+                // Select the same Symbol we're getting a data point for
+                // from the articles list so that we can get the sentiment of the article.
+                // We use the underlying Symbol because the Symbols included in the `Symbols` property
+                // are equity Symbols.
+                var selectedSymbol = article.Symbols.SingleOrDefault(x => x.Symbol == article.Symbol.Underlying);
+                if (selectedSymbol == null)
+                {
+                    continue;
+                }
 
-                // Get the sentiment score
-                var sentimentScore = intersection.Select(x => _words[x]).Sum();
+                // Sometimes sentiment is not included with the article by Benzinga.
+                // We have to check for null values before using it.
+                var sentimentScore = selectedSymbol.Sentiment;
+                if (sentimentScore == null)
+                {
+                    continue;
+                }
 
                 // Set holdings equal to 1/10th of the sentiment score we get
-                SetHoldings(article.Symbol.Underlying, sentimentScore / 10);
+                SetHoldings(article.Symbol.Underlying, sentimentScore.Value / 10);
             }
         }
 
         public override void OnSecuritiesChanged(SecurityChanges changes)
         {
-            foreach (var r in changes.RemovedSecurities.Where(x => x.Symbol.SecurityType == SecurityType.Equity))
+            foreach (var r in changes.RemovedSecurities)
             {
                 // If removed from the universe, liquidate and remove the custom data from the algorithm
                 Liquidate(r.Symbol);
--- a/Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
+++ b/Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
@@ -1,0 +1,75 @@
+﻿# QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+# Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from clr import AddReference
+AddReference("System")
+AddReference("QuantConnect.Algorithm")
+AddReference("QuantConnect.Algorithm.Framework")
+AddReference("QuantConnect.Common")
+
+from System import *
+from QuantConnect import *
+from QuantConnect.Algorithm import *
+from QuantConnect.Algorithm.Framework.Selection import *
+from QuantConnect.Data.Custom.Benzinga import *
+from QuantConnect.Data.UniverseSelection import *
+
+class BenzingaNewsAlgorithm(QCAlgorithm):
+
+    def Initialize(self):
+        self.words = {
+            "bad": -0.5, "good": 0.5,
+            "negative": -0.5, "great": 0.5,
+            "growth": 0.5, "fail": -0.5,
+            "failed": -0.5, "success": 0.5,
+            "nailed": 0.5, "beat": 0.5,
+            "missed": -0.5, "slipped": -0.5,
+            "outperforming": 0.5, "underperforming": -0.5,
+            "outperform": 0.5, "underperform": -0.5
+        }
+
+        self.SetStartDate(2018, 10, 12)
+        self.SetEndDate(2018, 11, 25)
+        self.SetCash(100000)
+
+        self.UniverseSettings.Resolution = Resolution.Daily
+        self.AddUniverseSelection(CoarseFundamentalUniverseSelectionModel(self.CoarseSelector))
+
+    def CoarseSelector(self, coarse):
+        # Add Benzinga news data from the filtered coarse selection
+        symbols = [i.Symbol for i in coarse if i.HasFundamentalData and i.DollarVolume > 50000000][:10]
+
+        for symbol in symbols:
+            self.AddData(BenzingaNews, symbol)
+
+        return symbols
+
+    def OnData(self, data):
+        for article in data.Get(BenzingaNews).Values:
+            # Split the article into words in all lowercase
+            articleWords = article.Contents.lower().split(" ")
+
+            # Get the list of matching words we have sentiment definitions for
+            intersection = set(self.words.keys()).intersection(articleWords)
+
+            # Get the sentiment score
+            sentimentScore = sum([self.words[i] for i in intersection])
+
+            # Set holdings equal to 1/10th of the sentiment score we get
+            self.SetHoldings(article.Symbol.Underlying, sentimentScore / 10.0)
+
+    def OnSecuritiesChanged(self, changes):
+        for r in [i for i in changes.RemovedSecurities if i.Symbol.SecurityType == SecurityType.Equity]:
+            # If removed from the universe, liquidate and remove the custom data from the algorithm
+            self.Liquidate(r.Symbol)
+            self.RemoveSecurity(Symbol.CreateBase(BenzingaNews, r.Symbol, Market.USA))
--- a/Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
+++ b/Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
@@ -60,7 +60,7 @@ class BenzingaNewsAlgorithm(QCAlgorithm):
                     break
 
             if selectedSymbol is None:
-                continue
+                raise Exception(f"Could not find current Symbol {article.Symbol.Underlying} even though it should exist")
 
             # Sometimes sentiment is not included with the article by Benzinga.
             # We have to check for null values before using it.
--- a/Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
+++ b/Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
@@ -27,17 +27,6 @@ from QuantConnect.Data.UniverseSelection import *
 class BenzingaNewsAlgorithm(QCAlgorithm):
 
     def Initialize(self):
-        self.words = {
-            "bad": -0.5, "good": 0.5,
-            "negative": -0.5, "great": 0.5,
-            "growth": 0.5, "fail": -0.5,
-            "failed": -0.5, "success": 0.5,
-            "nailed": 0.5, "beat": 0.5,
-            "missed": -0.5, "slipped": -0.5,
-            "outperforming": 0.5, "underperforming": -0.5,
-            "outperform": 0.5, "underperform": -0.5
-        }
-
         self.SetStartDate(2018, 10, 12)
         self.SetEndDate(2018, 11, 25)
         self.SetCash(100000)
@@ -55,21 +44,35 @@ class BenzingaNewsAlgorithm(QCAlgorithm):
         return symbols
 
     def OnData(self, data):
+        # Get all Benzinga data and loop over it
         for article in data.Get(BenzingaNews).Values:
-            # Split the article into words in all lowercase
-            articleWords = article.Contents.lower().split(" ")
+            selectedSymbol = None
 
-            # Get the list of matching words we have sentiment definitions for
-            intersection = set(self.words.keys()).intersection(articleWords)
+            # Use loop instead of list comprehension for clarity purposes
 
-            # Get the sentiment score
-            sentimentScore = sum([self.words[i] for i in intersection])
+            # Select the same Symbol we're getting a data point for
+            # from the articles list so that we can get the sentiment of the article
+            # We use the underlying Symbol because the Symbols included in the `Symbols` property
+            # are equity Symbols.
+            for x in article.Symbols:
+                if x.Symbol == article.Symbol.Underlying:
+                    selectedSymbol = x
+                    break
+
+            if selectedSymbol is None:
+                continue
+
+            # Sometimes sentiment is not included with the article by Benzinga.
+            # We have to check for null values before using it.
+            sentimentScore = selectedSymbol = selectedSymbol.Sentiment
+            if sentimentScore is None:
+                continue
 
             # Set holdings equal to 1/10th of the sentiment score we get
             self.SetHoldings(article.Symbol.Underlying, sentimentScore / 10.0)
 
     def OnSecuritiesChanged(self, changes):
-        for r in [i for i in changes.RemovedSecurities if i.Symbol.SecurityType == SecurityType.Equity]:
+        for r in changes.RemovedSecurities:
             # If removed from the universe, liquidate and remove the custom data from the algorithm
             self.Liquidate(r.Symbol)
             self.RemoveSecurity(Symbol.CreateBase(BenzingaNews, r.Symbol, Market.USA))
--- a/Common/Data/Custom/Benzinga/BenzingaMetadata.cs
+++ b/Common/Data/Custom/Benzinga/BenzingaMetadata.cs
@@ -1,0 +1,43 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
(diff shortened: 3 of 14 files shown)


ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+                    if (articleExists)
+                    {
+                        Log.Error($"BenzingaDataConverter.WriteToFile(): Restoring backup article file: {backupArticleFile} to original location: {finalArticleFile}");
+                        File.Move(backupArticleFile, finalArticleFile);
+                    }
+                }
+            }
+
+            // Now compress all of the data we have for a given day
+            var finalDataDirectory = new DirectoryInfo(Path.Combine(_destinationDirectory.FullName, "content", date.ToStringInvariant(DateFormat.EightCharacter)));
```
Martin-Molinero: Isn't this `finalDirectory`?

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+
+        /// <summary>
+        /// Process the data to BenzingaNews instances
+        /// </summary>
+        /// <param name="parentSourceDirectory">Source directory to read files from</param>
+        /// <param name="date"></param>
+        /// <returns></returns>
+        private IEnumerable<BenzingaNews> Process(DirectoryInfo dateDirectory)
+        {
+            foreach (var publication in dateDirectory.GetFiles("*.xml", SearchOption.TopDirectoryOnly))
```
Martin-Molinero: Should they be ordered in some way?

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+            return true;
+        }
+
+        /// <summary>
+        /// Process the data to BenzingaNews instances
+        /// </summary>
+        /// <param name="parentSourceDirectory">Source directory to read files from</param>
+        /// <param name="date"></param>
+        /// <returns></returns>
+        private IEnumerable<BenzingaNews> Process(DirectoryInfo dateDirectory)
```
Martin-Molinero: nit - Documentation is outdated

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+                    if (ticker["#text"] == null)
+                    {
+                        continue;
+                    }
+
+                    instance.Symbols.Add(new BenzingaSymbolData()
+                    {
+                        Exchange = exchange,
+                        Sentiment = sentiment,
+                        Symbol = new Symbol(SecurityIdentifier.GenerateEquity(ticker["#text"].ToString().Trim(), Market.USA, mapSymbol: false), ticker["#text"].ToString().Trim())
```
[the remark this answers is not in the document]
gsalaz98: I considered mapping for the tickers (and my first implementation did map them), but I discovered that the tickers in the articles are essentially already "mapped" because the news article reports on the tickers as of that date. I'm willing to be convinced that my approach is wrong here, but would like to discuss a bit if possible 👍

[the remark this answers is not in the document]
Martin-Molinero: Even if already mapped, was there a reason not to map them here too?
I was thinking this is something that we want because it sets the `SID` date value correctly and not to `DefaultDate`, and we want this because it ends up being part of the SID properties and used for equality. What do you think?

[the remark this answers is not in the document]
gsalaz98: I think you're right. What if a user wants to place a trade on the basket of stocks mentioned in an article? With the SID incorrectly set, it might mess up the order in some unknown way. The risks outweigh the benefits of not mapping here, so I'll update it 👍

ToolBox/Benzinga/BenzingaDataConverter.cs
gsalaz98: The end result is the same regardless of ordering as we generate references to each article individually

Common/Data/Custom/Benzinga/BenzingaNews.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// Categories that the article belongs to
+        /// </summary>
+        [JsonProperty("category"), JsonConverter(typeof(SingleValueListConverter<BenzingaCategory>))]
+        public List<BenzingaCategory> Category { get; set; }
+
+        /// <summary>
+        /// Unique ID assigned to the article by Benzinga
+        /// </summary>
+        [JsonProperty("bz:id")]
+        public string Id { get; set; }
```
jaredbroad: Lets make it a habit to fill in the <example> xml too for all of these so we can review/document properly. Is this a string id?

Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
```diff
@@ -0,0 +1,75 @@
...
+    def CoarseSelector(self, coarse):
+        # Add Benzinga news data from the filtered coarse selection
+        symbols = [i.Symbol for i in coarse if i.HasFundamentalData and i.DollarVolume > 50000000][:10]
+
+        for symbol in symbols:
+            self.AddData(BenzingaNews, symbol)
+
+        return symbols
+
+    def OnData(self, data):
```
jaredbroad: Can you please update the examples to use the BZ Symbol Sentiment fields?

Algorithm.Python/AltData/BenzingaNewsAlgorithm.py
```diff
@@ -0,0 +1,75 @@
...
+            intersection = set(self.words.keys()).intersection(articleWords)
+
+            # Get the sentiment score
+            sentimentScore = sum([self.words[i] for i in intersection])
+
+            # Set holdings equal to 1/10th of the sentiment score we get
+            self.SetHoldings(article.Symbol.Underlying, sentimentScore / 10.0)
+
+    def OnSecuritiesChanged(self, changes):
+        for r in [i for i in changes.RemovedSecurities if i.Symbol.SecurityType == SecurityType.Equity]:
```
jaredbroad: I don't "think" this is required anymore ` if i.Symbol.SecurityType == SecurityType.Equity` as we don't pass custom data here. If you see it -> its a bug somewhere =)

Common/Data/Custom/Benzinga/BenzingaNews.cs
gsalaz98: Going through all the fields (as a quick self review):

* `Title: string`
* `Contents: string`
* `PublicationDate: DateTime`
* `Author: string`
* `Category: List<BenzingaCategory>`
    - `Resource: string` - can be links with ticker included in them. Decided against using enum here
    - `Description: string` - same as Resource
* `Id: string` - Only numerical representations appear in the raw data. Will update to int
* `RevisionId: string` - Only numerical representations appear in the raw data. Will update to int
* `Symbols: List<BenzingaSymbolData>`
    - `Sentiment: decimal?` - Not all fields contain sentiment alongside with them. Maybe I should default to 0 if sentiment is missing? But at the same time, a sentiment of 0 means something if it's been deliberately included...
    - `Exchange: string` - Thought about making this an enum, but might be vendor specific since we have some exchanges like "ETF" in the raw data
    - `Symbol: Symbol`
* `RevisionDate: DateTime`
* `Metadata: BenzingaMetadata`
    - `IsPro: bool`
    - `FirstRun: bool`
    - `Kind: string` - Always appears as "source" in the raw data, but I believe having it as an enum value isn't right since we'd have an enum with only one value, and it will break if another value shows up during processing

Just a quick synopsis of the fields present and some additional notes containing some design considerations.

Algorithm.CSharp/AltData/BenzingaNewsAlgorithm.cs
```diff
@@ -0,0 +1,93 @@
...
+            foreach (var article in data.Get<BenzingaNews>().Values)
+            {
+                // Select the same Symbol we're getting a data point for
+                // from the articles list so that we can get the sentiment of the article.
+                // We use the underlying Symbol because the Symbols included in the `Symbols` property
+                // are equity Symbols.
+                var selectedSymbol = article.Symbols.SingleOrDefault(x => x.Symbol == article.Symbol.Underlying);
+                if (selectedSymbol == null)
+                {
+                    continue;
```
Martin-Molinero: Is this possible? if this `article` instance is for the `Symbol.Underlying` why wouldn't it be present? seems like a bug in the data converter/source?

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+                    {
+                        Log.Trace($"BenzingaDataConverter.WriteToFile(): Ticker {ticker.Symbol.Value} is invalid in Windows. Skipping");
+                        continue;
+                    }
+
+                    var tickerPath = new DirectoryInfo(Path.Combine(_destinationDirectory.FullName, ticker.Symbol.Value.ToLowerInvariant()));
+                    tickerPath.Create();
+
+                    var referenceFilePath = Path.Combine(tickerPath.FullName, $"{date:yyyMMdd}.csv");
+                    var tickerReferences = new HashSet<string>();
```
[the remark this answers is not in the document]
Martin-Molinero: Whops, sorry, I see you are using `OrderBy` now 👍 - let's just assert that Benzinga does not add old news with newer ids? Note this _did_ happen for `Tiingo`

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+        private IEnumerable<BenzingaNews> Process(DirectoryInfo dateDirectory)
+        {
+            foreach (var publication in dateDirectory.GetFiles("*.xml", SearchOption.TopDirectoryOnly))
+            {
+                var document = new XmlDocument();
+                document.LoadXml(File.ReadAllText(publication.FullName));
+
+                var item = JsonConvert.DeserializeObject<JObject>(JsonConvert.SerializeXmlNode(document))["rss"]["channel"]["item"];
+                // Benzinga complies with RSS standards and uses RFC 822 for DateTime
+                var publicationDate = Parse.DateTime(item["pubDate"].ToString());
```
Martin-Molinero: nit - `publicationDate` is never used

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                var instance = JsonConvert.DeserializeObject<BenzingaNews>(item.ToString());
+
+                // Strip all HTML tags from the article, then convert HTML entities to their string representation
+                // e.g. "<html><p>Apple&#39;s Earnings</p></html>" would become "Apple's Earnings"
+                instance.Contents = WebUtility.HtmlDecode(Regex.Replace(instance.Contents, @"<[^>]*>", " "));
+                instance.Symbols = new List<BenzingaSymbolData>();
+                instance.Metadata = new BenzingaMetadata()
+                {
+                    IsPro = item["bz:type"]["@pro"].ToString() == "1",
+                    FirstRun = item["bz:type"]["@pro"].ToString() == "1",
```
Martin-Molinero: Seems `FirstRun` is being set same as `IsPro`?

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+
+            foreach (var article in news)
+            {
+                var createdReference = false;
+
+                // For each ticker, add a reference to the article we're going to store under the "content" folder
+                foreach (var ticker in article.Symbols)
+                {
+                    // Invalid character in path
+                    if (ticker.Symbol.Value.Contains(":"))
```
Martin-Molinero: I think this invalid symbol is going to be stored with the article? Maybe we should filter it before so that it doesn't end up being shipped and consumed
Probably should also apply for the exchange check bellow?

Common/Data/Custom/Benzinga/BenzingaNews.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// </summary>
+        [JsonProperty("bz:revisionid")]
+        public int RevisionId { get; set; }
+
+        /// <summary>
+        /// Symbols that this news article applies to
+        /// </summary>
+        /// <remarks>
+        /// Initialize this outside of the JSON deserialization process since we're unable to
+        /// choose how we want the tickers referenced in the articles as (mapped or unmapped)
```
Martin-Molinero: I think this comment is not valid anymore?

Common/Data/Custom/Benzinga/BenzingaNews.cs
```diff
@@ -0,0 +1,226 @@
...
+        /// <summary>
+        /// Date that the article was revised on
+        /// </summary>
+        [JsonProperty("bz:revisiondate")]
+        public DateTime RevisionDate { get; set; }
+
+        /// <summary>
+        /// Metadata associated with the article
+        /// </summary>
+        /// <remarks>Initialize separately, same as <see cref="Symbols" />
```
Martin-Molinero: nit - I believe `remarks` is not being closed

Common/Data/Custom/Benzinga/BenzingaNews.cs
```diff
@@ -0,0 +1,226 @@
...
+            };
+        }
+
+        /// <summary>
+        /// Converts the instance to string
+        /// </summary>
+        /// <returns>Article title and contents</returns>
+        public override string ToString()
+        {
+            return $"{Symbol} - Article title: {Title}\nArticle contents:\n{Contents}";
```
Martin-Molinero: Good idea 👍 - Might be worth adding the `Time`

Common/Data/Custom/Benzinga/BenzingaMetadata.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// Is the article a first run
+        /// </summary>
+        [JsonProperty("@firstrun")]
+        public bool FirstRun { get; set; }
+
+        /// <summary>
+        /// Type of article
+        /// </summary>
+        [JsonProperty("#text")]
+        public string Kind { get; set; }
```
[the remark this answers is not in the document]
mchandschuh: RE: brittle -- I'd argue it makes it _less_ brittle -- failing fast is a *good* property, otherwise it's silently failing and you don't even know it! Good call @Martin-Molinero 👍

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                }
+                else
+                {
+                    foreach (var tickerData in JArray.Parse(item["bz:ticker"].ToString()))
+                    {
+                        tickers.Add(tickerData);
+                    }
+                }
+
+                var instance = JsonConvert.DeserializeObject<BenzingaNews>(item.ToString());
```
mchandschuh: seeing the `ToString()`'s everywhere is kind of unexpected. Per the [documentation](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_ToString.htm), the `ToString()` method on `JToken` and it's descendants is to:
> Returns the indented JSON for this token.

Whereas the `Value<T>(propertyName)` function ([docs](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_Value__1.htm)):
> Gets the `JToken` with the specified key converted to the specified type.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+
+                // Strip all HTML tags from the article, then convert HTML entities to their string representation
+                // e.g. "<html><p>Apple&#39;s Earnings</p></html>" would become "Apple's Earnings"
```
mchandschuh: I love seeing examples from the raw data -- you'll thank yourself later 😄

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                instance.Contents = WebUtility.HtmlDecode(Regex.Replace(instance.Contents, @"<[^>]*>", " "));
+                instance.Symbols = new List<BenzingaSymbolData>();
+                instance.Metadata = new BenzingaMetadata()
```
mchandschuh: style nit -- trailing empty parenthesis `()` when using object initialization syntax

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                }
+
+                // Check for only a single ticker before we try to iterate
+                if (item["bz:ticker"].Type == JTokenType.Object)
+                {
+                    tickers.Add(item["bz:ticker"]);
+                }
+                else
+                {
+                    foreach (var tickerData in JArray.Parse(item["bz:ticker"].ToString()))
```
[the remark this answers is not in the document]
mchandschuh: _also_ -- why not convert these into objects right away? You define all the required `[JsonProperty]` attributes -- you can do something like... `token.ToObject<BenzingaNews>()` ([docs](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_ToObject__1.htm))

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                {
+                    IsPro = item["bz:type"]["@pro"].ToString() == "1",
+                    FirstRun = item["bz:type"]["@pro"].ToString() == "1",
+                    Kind = item["bz:type"]["#text"].ToString()
+                };
+
+                foreach (var ticker in tickers)
+                {
+                    var sentiment = ticker["@sentiment"] == null ? (decimal?)null : Parse.Decimal(ticker["@sentiment"].ToString());
+                    var exchange = ticker["@exchange"] == null ? string.Empty : ticker["@exchange"].ToString();
```
mchandschuh: If these were converted into objects above as suggested -- much of this code simplifies.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+
+                    if (ticker["#text"] == null)
+                    {
+                        continue;
```
mchandschuh: it's good style to place guard clauses at the very top of the scope. In this case, there's no value in parsing `sentiment` and `exchange` just to get to a bad `#text` value to skip the entire `ticker` -- but again -- I think most of the code in this loop collapses -- I recommend _briefly_ reviewing those docs, reading this code it looks like you're going against the grain of the library.

I'm gonna share something else here totally unrelated only because it's one of the coolest features of the library `JToken` side -- I actually went and found an [SO question](https://stackoverflow.com/a/29782523/1582922) I could answer using the feature so I would never lose it!! Basically you can provide a string expression, such as "symbols[*].ticker" which would work on an object like:
``` js
{
  symbols: [
   { ticker: 'ticker-1' },
   { ticker: 'ticker-2' },
   { ticker: 'ticker-3' },
   { not-ticker: 'ticker-1' }
  ]
}
```

and it would happily ignore the `not-ticker` -- pretty nice feature when working using the JToken library. _That_ being said, in general, the JToken side of the house is typically used when you don't have the contracts coded up (custom class w/ `[JsonProperty]` attributes.

ToolBox/Benzinga/BenzingaDataConverter.cs
mchandschuh: Is this common? Do we want to provide a mapping -- map to disk and map back when reading in? Just wondering what and how much data we're losing by performing this filtering step -- _also_ -- perhaps we should do an `OS.IsWindows` check here since *nix systems don't give two shakes about colons in paths.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                    // Invalid character in path
+                    if (ticker.Symbol.Value.Contains(":"))
+                    {
+                        Log.Trace($"BenzingaDataConverter.WriteToFile(): Ticker {ticker.Symbol.Value} contains invalid character ':'. Skipping");
+                        continue;
+                    }
+                    // ETFs have a special exchange tag associated with them.
+                    // Sometimes, we have an empty exchange for some tickers. Let's try and map them first before giving up on them.
+                    // Example: SPOT (Spotify) on 2018-10-01 has a missing exchange even though it has already IPO'd
+                    if (ticker.Exchange != "NASDAQ" && ticker.Exchange != "NYSE" && ticker.Exchange != "ETF" && !string.IsNullOrEmpty(ticker.Exchange))
```
mchandschuh: recommend defining something like:

``` csharp
private static readonly IReadOnlyCollection<string> ValidExchanges = new HashSet<string>
{
    "NASDAQ", "NYSE", "ETF"
}
```

do we need this not empty predicate? it kinda threw me... it _looks_ like you're saying the exchange must be one of these three valid exchanges (`NASDAQ`/`NYSE`/`ETF`) -- which means who cares if it's null or empty -- it's not one of the pre-ordained exchanges, so we toss it anyway.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                    {
+                        Log.Trace($"BenzingaDataConverter.WriteToFile(): Ticker {ticker.Symbol.Value} is not in NYSE, NASDAQ, or is not an ETF. Skipping");
+                        continue;
+                    }
+                    // We can't write these tickers in Windows, so we'll skip them
+                    if (_isWindows && _forbiddenTickers.Contains(ticker.Symbol.Value))
```
mchandschuh: totally style and feel free to ignore this, but my preference would be to use `OS.IsWindows` directly

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                    {
+                        Log.Trace($"BenzingaDataConverter.WriteToFile(): Ticker {ticker.Symbol.Value} is invalid in Windows. Skipping");
```
mchandschuh: Thinking about mapping again -- I guess we don't care _too_ much since we do the data processing in *nix ? Also, thinking about breaking this functionality out into a component, could get it for free

ToolBox/Benzinga/BenzingaDataConverter.cs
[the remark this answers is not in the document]
mchandschuh: you can also do:

``` csharp
var tickerPath = _destinationDirectory.CreateSubDirectory(ticker.Symbol.Value.ToLowerInvariant());
```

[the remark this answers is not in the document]
mchandschuh: RE: _new_ articles with _old_ ids -- why can't people just use something like `require('uuid/v4')` and be done with it (.net `Guid.NewGuid()` -- we (as in humanity) solved this problem decades ago -- they even do uuid's that have a time component -- I think v1 and v3 maybe and one of those is sortable -- anyway -- more just me blabbing

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                    var tickerPath = new DirectoryInfo(Path.Combine(_destinationDirectory.FullName, ticker.Symbol.Value.ToLowerInvariant()));
+                    tickerPath.Create();
+
+                    var referenceFilePath = Path.Combine(tickerPath.FullName, $"{date:yyyMMdd}.csv");
+                    var tickerReferences = new HashSet<string>();
+
+                    if (File.Exists(referenceFilePath))
+                    {
+                        tickerReferences = File.ReadAllLines(referenceFilePath)
+                            .ToHashSet();
```
mchandschuh: style nit -- I would throw the `ToHashSet()` on the same line

not a nit -- **why** are we using a `HashSet<T>` here at all? hash sets, or really sets in general, are great for two reasons, one is guaranteeing uniqueness and for `O(1)`  `Contains`/`Exists` calls. Do we expect the file at `referenceFilePath` to contain duplicates?

I suppose even if we _don't_ expect dupes, I'm sure it's nice being able to guarantee that no dupes will be written. it's nice to define _invariants_ like that -- helps you reason about code and more importantly, and downstream applications that might pick up this application's output -- such as algorithms 😄

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                        Log.Trace($"BenzingaDataConverter.WriteToFile(): Moving existing reference file to backup location: {existingReferenceFileBackupPath}");
+                        File.Move(referenceFilePath, existingReferenceFileBackupPath);
+                    }
+
+                    try
+                    {
+                        Log.Trace($"BenzingaDataConverter.WriteToFile(): Writing reference to {referenceFilePath}");
+                        File.Move(referenceFileTemp, referenceFilePath);
+                    }
+                    catch (Exception e)
```
mchandschuh: _super_ nit -- typically in this codebase we use `error` when handling exceptions.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+                        {
+                            Log.Error($"BenzingaDataConverter.WriteToFile(): Moving backup file: {existingReferenceFileBackupPath} back to original location: {referenceFilePath}");
+                            File.Move(existingReferenceFileBackupPath, referenceFilePath);
+                        }
+
+                        continue;
+                    }
+
+                    // Clean up after ourselves so that we don't accidentally run out of storage space
+                    File.Delete(existingReferenceFileBackupPath);
```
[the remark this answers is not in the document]
mchandschuh: yay! components!!

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+                    {
+                        Log.Error($"BenzingaDataConverter.WriteToFile(): Restoring backup article file: {backupArticleFile} to original location: {finalArticleFile}");
+                        File.Move(backupArticleFile, finalArticleFile);
+                    }
+                }
+            }
+
+            // Now compress all of the data we have for a given day
+            var compressedFinal = new FileInfo(Path.Combine(_destinationDirectory.FullName, "content", $"{date:yyyyMMdd}.zip"));
+            var compressedFinalBackup = new FileInfo(Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid()}.zip"));
```
mchandschuh: I like seeing you work w/ `FileInfo` -- I used to always just use strings, but keeping things as a `FileInfo` can be very powerful, far superior to holding the value as a string. There's actual an _evil_ in programming called _primitive obsession_ -- it's where you're afraid to create types and instead just pass simple structs everywhere.

Here's a refactoring guru article on [primitive obsession](https://refactoring.guru/smells/primitive-obsession) for anyone following along at home that is interested... well... even if you're not interested the link ain't goin' anywhere 😃

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+
+            if (compressedFinal.Exists)
```
mchandschuh: not applicable here, but an interesting and potentially headache-inducing side effect of using `FileInfo` is that they only load their data once -- the first time you access anything derived from the file's headers/options. You can update an instance of `FileInfo` manually by calling the `Refresh()` function. Here's an [SO thread](https://github.com/dotnet/corefx/issues/42447) discussing it along with the [method documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.refresh?view=netframework-4.8). You'll note this is pointing at `FileSystemInfo` -- which is the base class of `FileInfo` and he's the one that defines the `Refresh` function.

EDIT: You're totally all over this -- didn't see a usage of `FileSystemInfo.Refresh()` until hitting enter on this one. Leaving here for posterity.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+            }
+
+            try
+            {
+                Log.Trace($"BenzingaDataConverter.WriteToFile(): Compressing {finalDirectory.Name} to {compressedFinal.FullName}");
+                Compression.ZipDirectory(finalDirectory.FullName, compressedFinal.FullName, includeRootInZip: false);
+            }
+            catch (Exception e)
+            {
+                compressedFinal.Refresh();
```
mchandschuh: oh snap!!! well done -- you can tell I'm doing this line by line. Many people don't know about the `Refresh()` function and it's behavior -- I'm going to leave the comments about it for posterity.

ToolBox/Benzinga/BenzingaProgram.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.IO;
+
+namespace QuantConnect.ToolBox.Benzinga
+{
+    public static class BenzingaProgram
+    {
+        public static void BenzingaDataConverter(string source, string destination, string day)
+        {
+            var date = Parse.DateTimeExact(day, "yyyyMMdd");
```
mchandschuh: 👍

Common/Data/Custom/Benzinga/BenzingaMetadata.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// <summary>
+    /// Article metadata
+    /// </summary>
+    public class BenzingaMetadata
+    {
+        /// <summary>
+        /// Is the article for pro subscribers only
+        /// </summary>
+        [JsonProperty("@pro")]
+        public bool IsPro { get; set; }
```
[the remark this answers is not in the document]
gsalaz98: It's all about how Json.NET deals with serializing XML to JSON. It adds a `#` prefix to the inner element. An example is `<a href="example">1234</a>` would show up as `#a` for `1234` and `@href`. 

I decided to go this way instead of using an auto-generated class from `xsd.exe` since I think it's a cleaner solution and bloats the codebase less (and makes it more understandable/maintainable too in my opinion)

Common/Data/Custom/Benzinga/BenzingaSymbolData.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// Exchange the symbol belongs to
+        /// </summary>
+        [JsonProperty("@exchange")]
+        public string Exchange { get; set; }
+
+        /// <summary>
+        /// Symbol
+        /// </summary>
+        [JsonProperty("#text")]
+        public Symbol Symbol { get; set; }
```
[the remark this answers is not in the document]
gsalaz98: I store the Symbol with the appropriate SID in the final news file. That way we already use the existing `SymbolJsonConverter` when deserializing the data from disk.

Note: we calculate the SID in the processing stage.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,346 @@
...
+        /// <param name="sourceDirectory">Directory to read data from. This should be the Benzinga data folder (e.g. /Data/alternative/benzinga)</param>
+        /// <param name="destinationDirectory">Directory to write data to. This should be the final Benzinga data folder (e.g. /Data/alternative/benzinga)</param>
+        public BenzingaDataConverter(DirectoryInfo sourceDirectory, DirectoryInfo destinationDirectory)
+        {
+            _sourceDirectory = sourceDirectory;
+            _destinationDirectory = destinationDirectory;
+            _destinationDirectory.Create();
+            _isWindows = OS.IsWindows;
+
+            _mapFileProvider = Composer.Instance.GetExportedValueByTypeName<IMapFileProvider>(Config.Get("map-file-provider", "LocalDiskMapFileProvider"));
```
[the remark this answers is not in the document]
gsalaz98: Yes - this is for our backend operations and the need for flexibility there

ToolBox/Benzinga/BenzingaDataConverter.cs
gsalaz98: Just a note for you 😄 : I avoided using `OS.IsWindows` since there is close to absolutely no possibility that we will have the OS switch on us while the program is running. In addition to that, since the property would perform a [few allocations per call](https://referencesource.microsoft.com/#mscorlib/system/environment.cs,7d2f1469d916fc63) (albeit probably in a low heap generation), it wouldn't matter. But what I'm concerned about is possible syscalls being made trying to identify the platform causing the processing to slow down since we're leaving userspace, especially in a hotter path.

Agreed with you that OS.IsWindows is much cleaner, but those were the design considerations I made when I chose to cache the `OS.IsWindows` call.

ToolBox/Benzinga/BenzingaDataConverter.cs
gsalaz98: With regards to stripping the colon character ":" (since GitHub is glitching out and won't let me reply to your review) - The colon character exclusively appears for international markets, and because we currently don't support other markets, I chose to strip them out. 

And yes, I'd like to split this into another class and refactor, but that's for another PR

ToolBox/Benzinga/BenzingaDataConverter.cs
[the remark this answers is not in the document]
gsalaz98: w/ regards to your last comment: I need to pre-process the values before I can include it in the instance I'm constructing.

ToolBox/Benzinga/BenzingaDataConverter.cs
gsalaz98: The not empty predicate is there to attempt a recovery of the ticker in question if the exchange tag was for some reason omitted. 

`// Example: SPOT (Spotify) on 2018-10-01 has a missing exchange even though it has already IPO'd`

It's a little ugly, so I'll clarify what the purpose of the `string.IsNullOrEmpty` by creating a new parameter for this called `attemptRecovery` 👍

gsalaz98:
This PR will be temporarily delayed while I work on implementing support for data sourced from Benzinga's API. Reviews will be addressed in that commit as well as added support for the API feed.

ToolBox/Benzinga/BenzingaDataConverter.cs
```diff
@@ -0,0 +1,321 @@
...
+using System.Xml;
+
+namespace QuantConnect.ToolBox.Benzinga
+{
+    public class BenzingaDataConverter
+    {
+        private readonly DirectoryInfo _sourceDirectory;
+        private readonly DirectoryInfo _destinationDirectory;
+        private readonly bool _isWindows;
+        private readonly HashSet<string> _forbiddenTickers = new HashSet<string>()
```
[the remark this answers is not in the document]
Martin-Molinero: Created a new issue to track this https://github.com/QuantConnect/Lean/issues/3849 👍

jaredbroad:
Closing until completed.

PR #3786, opened 2019-10-30
Add streaming custom data support to LiveTradingDataFeed

### LINKED ISSUE #3785: Add streaming custom data support to LiveTradingDataFeed
#### Expected Behavior
`LiveTradingDataFeed` supports streaming custom data types

#### Actual Behavior
`LiveTradingDataFeed` does not support streaming custom data types

#### Potential Solution
n/a

#### Reproducing the Problem
n/a

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- Added `SubscriptionTransportMedium.Streaming`
- Set `SubscriptionTransportMedium.Streaming` in live mode for the following classes:
  - `Tick`, `TradeBar`, `QuoteBar`, `OpenInterest`, `TiingoNews`, `PsychSignalSentiment`
- Updated `LiveTradingDataFeed` to handle streaming of live custom data

#### Related Issue
Closes #3785 

#### Motivation and Context
Enable streaming of custom data in the cloud.

No,

- New unit tests
- Live algorithm receiving PsychSignal custom data

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/Tiingo/TiingoNews.cs
+++ b/Common/Data/Custom/Tiingo/TiingoNews.cs
@@ -102,6 +102,12 @@ namespace QuantConnect.Data.Custom.Tiingo
         /// <returns>The <see cref="SubscriptionDataSource"/> instance to use</returns>
         public override SubscriptionDataSource GetSourceForAnIndex(SubscriptionDataConfig config, DateTime date, string index, bool isLiveMode)
         {
+            if (isLiveMode)
+            {
+                // this data type is streamed in live mode
+                return new SubscriptionDataSource(string.Empty, SubscriptionTransportMedium.Streaming);
+            }
+
             var source = Path.Combine(
                 Globals.DataFolder,
                 "alternative",
--- a/Common/Data/Custom/Tiingo/TiingoNews.cs
+++ b/Common/Data/Custom/Tiingo/TiingoNews.cs
@@ -132,12 +132,8 @@ namespace QuantConnect.Data.Custom.Tiingo
         {
             if (isLiveMode)
             {
-                var tiingoTicker = TiingoSymbolMapper.GetTiingoTicker(config.Symbol);
-                var url = Invariant($"https://api.tiingo.com/tiingo/news?tickers={tiingoTicker}&startDate={date:yyyy-MM-dd}&token={Tiingo.AuthCode}&sortBy=crawlDate");
-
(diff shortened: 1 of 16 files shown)


Common/Data/Custom/Tiingo/TiingoNews.cs
```diff
@@ -102,6 +102,12 @@ public class TiingoNews : IndexedBaseData
         /// <returns>The <see cref="SubscriptionDataSource"/> instance to use</returns>
         public override SubscriptionDataSource GetSourceForAnIndex(SubscriptionDataConfig config, DateTime date, string index, bool isLiveMode)
         {
+            if (isLiveMode)
```
Martin-Molinero: `GetSource` will be called first and based on it `GetSourceForAnIndex` _might_ be called (for example for backtesting -> `FileFormat.Index` ) - So I think we want this `if` there?

StefanoRaggi: Oops! the change was actually meant to be in `GetSource` :)

Engine/DataFeeds/LiveTradingDataFeed.cs
```diff
@@ -324,50 +329,56 @@ protected Subscription CreateDataSubscription(SubscriptionRequest request)
...
-                                tickEnumerator.Enqueue(data);
-                                subscription.OnNewDataAvailable();
-                                if (tick.TickType != TickType.OpenInterest)
+                        case nameof(Tick):
+                        default:
+                            // tick or streaming custom data subscriptions can pass right through
+                            var tickEnumerator = new EnqueueableEnumerator<BaseData>();
+
+                            _exchange.AddDataHandler(
+                                request.Configuration.Symbol,
```
Martin-Molinero: One thing to note is that for custom subscriptions, this `Symbol` instance might have an `underlying` which is **used for equality** (https://github.com/QuantConnect/Lean/blob/master/Common/SecurityIdentifier.cs#L754) and the incoming data might not have the underlying set, so we need to assert the `underlying` is set in the data producer or remove it here  🤔

StefanoRaggi: I think the data producer should be responsible for generating the correct symbol identifiers and setting the underlying symbol if required. If the symbol in the incoming data doesn't have the underlying set, the data handler will not be called and this is expected behavior, right?

PR #3800, opened 2019-11-04
Adds EndTime to PsychSignalSentiment

### LINKED ISSUE #3798: PsychSignalSentiment does not set time Period and live offset delay
#### Expected Behavior

We model reality as close as possible
#### Actual Behavior

PsychSignal currently sets `Time` as `EndTime` and no `Period` is set.
Additionally, there is no fixed live offset delay (i.e. delay from when time changes to XX:YY and until the data is available to get from the API)
#### Potential Solution

Introduced a fixed offset for PsychSignal and correctly model Time/EndTime
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

gsalaz98:
#### Description

Previously, we would only set the `Time` property, which `EndTime` would set for us. However, we had no frame of reference as to when the data was began being aggregated. 

In addition to this, we had no mechanism to model the delay from the API when retrieving data from it. I have set this delay to zero for the meanwhile.

#### Related Issue

Closes #3798 
#### Motivation and Context

Live trading reconciliation

No

Local regression tests and cloud testing
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
+++ b/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
@@ -47,7 +47,7 @@ namespace QuantConnect.Data.Custom.PsychSignal
         /// Time it takes for data to become available on the PsychSignal API after
         /// the time advances to the next minute
         /// </summary>
-        public static TimeSpan FixedLiveOffset { get; set; } = TimeSpan.FromSeconds(0);
+        public static TimeSpan FixedLiveOffset { get; set; } = TimeSpan.FromSeconds(15);
 
         /// <summary>
         /// Bullish intensity as reported by psychsignal
--- a/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
+++ b/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
@@ -34,8 +34,8 @@ namespace QuantConnect.Data.Custom.PsychSignal
         /// the time it should be emitted in backtesting/live trading
         /// </summary>
         public override DateTime EndTime {
-            get { return Time + Period + FixedLiveOffset; }
-            set { Time = value - Period; }
+            get { return Time.Add(Period).Add(FixedLiveOffset); }
+            set { Time = value.Subtract(Period); }
         }
 
         /// <summary>
@@ -173,7 +173,7 @@ namespace QuantConnect.Data.Custom.PsychSignal
         {
             return new PsychSignalSentiment
             {
-                EndTime = EndTime,
+                Time = Time,
                 Symbol = Symbol,
                 BullIntensity = BullIntensity,
                 BearIntensity = BearIntensity,
--- a/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
+++ b/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
@@ -33,9 +33,10 @@ namespace QuantConnect.Data.Custom.PsychSignal
         /// The EndTime of the bar represents the time the data ended and
         /// the time it should be emitted in backtesting/live trading
         /// </summary>
-        public override DateTime EndTime {
-            get { return Time.Add(Period).Add(FixedLiveOffset); }
-            set { Time = value.Subtract(Period); }
+        public override DateTime EndTime
+        {
+            get { return Time + Period + FixedLiveOffset; }
+            set { Time = value - Period; }
         }
 
         /// <summary>
@@ -47,7 +48,7 @@ namespace QuantConnect.Data.Custom.PsychSignal
         /// Time it takes for data to become available on the PsychSignal API after
         /// the time advances to the next minute
         /// </summary>
-        public static TimeSpan FixedLiveOffset { get; set; } = TimeSpan.FromSeconds(15);
+        private static readonly TimeSpan FixedLiveOffset = TimeSpan.FromSeconds(15);
 
         /// <summary>
         /// Bullish intensity as reported by psychsignal
--- a/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
+++ b/Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
@@ -30,6 +30,26 @@ namespace QuantConnect.Data.Custom.PsychSignal
     public class PsychSignalSentiment : BaseData
     {
         /// <summary>
+        /// The EndTime of the bar represents the time the data ended and
+        /// the time it should be emitted in backtesting/live trading
+        /// </summary>
+        public override DateTime EndTime {
+            get { return Time + Period + FixedLiveOffset; }
+            set { Time = value - Period; }
+        }
+
+        /// <summary>
+        /// Time from the start of the bar until the end of the bar per each data point
+        /// </summary>
+        public TimeSpan Period { get; set; } = TimeSpan.FromMinutes(1);
+
+        /// <summary>
+        /// Time it takes for data to become available on the PsychSignal API after
+        /// the time advances to the next minute
+        /// </summary>
+        public static TimeSpan FixedLiveOffset { get; set; } = TimeSpan.FromSeconds(0);
+
+        /// <summary>
         /// Bullish intensity as reported by psychsignal
         /// </summary>
         public decimal BullIntensity { get; set; }
@@ -127,7 +147,7 @@ namespace QuantConnect.Data.Custom.PsychSignal
 
                 return new PsychSignalSentiment
                 {
-                    Time = timestamp,
+                    EndTime = timestamp,
                     Symbol = config.Symbol,
                     BullIntensity = bullIntensity,
                     BearIntensity = bearIntensity,
@@ -153,7 +173,7 @@ namespace QuantConnect.Data.Custom.PsychSignal
         {
             return new PsychSignalSentiment
             {
-                Time = Time,
+                EndTime = EndTime,
                 Symbol = Symbol,
                 BullIntensity = BullIntensity,
                 BearIntensity = BearIntensity,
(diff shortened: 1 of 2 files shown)


Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
```diff
@@ -29,6 +29,26 @@ namespace QuantConnect.Data.Custom.PsychSignal
...
+        /// <summary>
+        /// Time from the start of the bar until the end of the bar per each data point
+        /// </summary>
+        public TimeSpan Period { get; set; } = TimeSpan.FromMinutes(1);
+
+        /// <summary>
+        /// Time it takes for data to become available on the PsychSignal API after
+        /// the time advances to the next minute
+        /// </summary>
+        public static TimeSpan FixedLiveOffset { get; set; } = TimeSpan.FromSeconds(0);
```
Jay-Jay-D: ¿Can you please explain this property?

jaredbroad: In live trading the bars appear 15s after the minute; we should make backtesting model this as closely as possible. We could set it to 15s now so backtesting accurately modeled...

Martin-Molinero:
- Lean is using `EndTime` so there shouldn't be an issue since the end times are always going forward
- Since the offset is less than the data resolution ( Minute right? ) it might be rounded down, see https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/SubscriptionData.cs#L65 where we will round down before emitting - I think this could potentially happen in other places too.
 - The overlapping could confuse users.

Could we use a 1 minute offset instead? Would be pessimistic but would avoid these issues

gsalaz98:
> 
> 
>     * Lean is using `EndTime` so there shouldn't be an issue since the end times are always going forward
> 
>     * Since the offset is less than the data resolution ( Minute right? ) it might be rounded down, see https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/SubscriptionData.cs#L65 where we will round down before emitting - I think this could potentially happen in other places too.
> 
>     * The overlapping could confuse users.
> 
> 
> Could we use a 1 minute offset instead? Would be pessimistic but would avoid these issues

Solely based on the fact that we will have time rounded down on us is enough to set a 1 minute offset. However, we should try to also aim to model live trading as accurately as possible. I'd like to discuss a solution with you if possible since there's a lot to consider here 👍

Martin-Molinero:
- @gsalaz98 reviewed the stack and it seems that the only sticky point would be consolidators not being triggered https://github.com/QuantConnect/Lean/blob/master/Engine/AlgorithmManager.cs#L1277 which do not exist today.
- My round down comment isn't valid because its applied on the data.Time and the offset is in the data.EndTime - see https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/SubscriptionData.cs#L65 closely
- Regarding history requests specifying bar counts as periods https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.History.cs#L199, since the underlying implementation uses resolution it might cause it to not include some data point, since end time might be after, but note that PsychSignalSentiment is flagged as sparse data too, so the behavior of this api call wouldn't be that deterministic either way

jaredbroad:
Any feedback on the Travis errors?

```Errors and Failures:
1) Test Failure : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes(AAPL.TiingoNews R735QTJ8XC9W,Hour,1,0,0,0,0,48,False)
     Expected: 48
  But was:  47

at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes (QuantConnect.Symbol symbol, QuantConnect.Resolution resolution, System.Int32 days, System.Int32 expectedTicksReceived, System.Int32 expectedTradeBarsReceived, System.Int32 expectedQuoteBarsReceived, System.Int32 expectedAuxPointsReceived, System.Int32 expectedCustomPointsReceived, System.Boolean shouldThrowException) [0x00ab0] in <81539c702ded4bbc88110d66c72d5b06>:0

2) Test Failure : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes(AAPL.TiingoNews R735QTJ8XC9W,Minute,1,0,0,0,0,120,False)
     Expected: 120
  But was:  119

at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes (QuantConnect.Symbol symbol, QuantConnect.Resolution resolution, System.Int32 days, System.Int32 expectedTicksReceived, System.Int32 expectedTradeBarsReceived, System.Int32 expectedQuoteBarsReceived, System.Int32 expectedAuxPointsReceived, System.Int32 expectedCustomPointsReceived, System.Boolean shouldThrowException) [0x00ab0] in <81539c702ded4bbc88110d66c72d5b06>:0

3) Test Failure : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes(AAPL.TiingoNews R735QTJ8XC9W,Tick,1,0,0,0,0,24,False)
     Expected: 24
  But was:  23

at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes (QuantConnect.Symbol symbol, QuantConnect.Resolution resolution, System.Int32 days, System.Int32 expectedTicksReceived, System.Int32 expectedTradeBarsReceived, System.Int32 expectedQuoteBarsReceived, System.Int32 expectedAuxPointsReceived, System.Int32 expectedCustomPointsReceived, System.Boolean shouldThrowException) [0x00ab0] in <81539c702ded4bbc88110d66c72d5b06>:0
```

Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
```diff
@@ -29,6 +29,26 @@ namespace QuantConnect.Data.Custom.PsychSignal
     /// </summary>
     public class PsychSignalSentiment : BaseData
     {
+        /// <summary>
+        /// The EndTime of the bar represents the time the data ended and
+        /// the time it should be emitted in backtesting/live trading
+        /// </summary>
+        public override DateTime EndTime {
+            get { return Time.Add(Period).Add(FixedLiveOffset); }
+            set { Time = value.Subtract(Period); }
```
mchandschuh: This property syntax doesn't seem to follow the rest of the code base which would place the opening brace on it's own line. What you have here is reminiscent of the K&R bracing style common in c/c++ and javascript

Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
```diff
@@ -29,6 +29,26 @@ namespace QuantConnect.Data.Custom.PsychSignal
...
+        }
+
+        /// <summary>
+        /// Time from the start of the bar until the end of the bar per each data point
+        /// </summary>
+        public TimeSpan Period { get; set; } = TimeSpan.FromMinutes(1);
```
mchandschuh: @jaredbroad  -- I still think we should have done `BaseData.Period` :P

Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
```diff
@@ -29,6 +29,26 @@ namespace QuantConnect.Data.Custom.PsychSignal
...
+
+        /// <summary>
+        /// Time it takes for data to become available on the PsychSignal API after
+        /// the time advances to the next minute
+        /// </summary>
+        public static TimeSpan FixedLiveOffset { get; set; } = TimeSpan.FromSeconds(15);
```
mchandschuh: Does this need to be public? This sounds like a particular implementation detail that the rest of the system should be none-the-wiser of. In fact, we would typically handle offsets like this simply via `Reader`. That being said, I do like having it explicit and the accompanying documentation. 👍  -- I think this should _minimally_ be `readonly` -- but ideally it would:

``` csharp
private static readonly TimeSpan FixedLiveOffset = TimeSpan.FromSeconds(15);
```

Also, the comments about the synchronizer's round-down behavior is tough. The issue is that the round down interval is the fill-forward interval which is dependent on _all_ subscriptions and not just this one. An interesting hack that could be done is we could provide this type a dedicated entry in `market-hours-database.json` with a time zone offset that accounts for the 15 seconds :P -- totally untested idea -- but methinks it would work and avoid coupling w/ other subscription resolution's in the algorithm.

Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
```diff
@@ -122,7 +142,7 @@ public override BaseData Reader(SubscriptionDataConfig config, string line, Date
 
                 return new PsychSignalSentiment
                 {
-                    Time = timestamp,
+                    EndTime = timestamp,
```
mchandschuh: A note for posterity... in LEAN, the `EndTime` is the *only* time that matters. We don't (or at least _shouldn't_) use the `Time` property for anything in the engine.

gsalaz98:
After some thought and inspection, I'm pretty sure that the test is failing because of my changes 😢 

Probably with the time advances, we can't squeeze in an additional data point per day, causing the number of points to drop from 24 to 23, 48 to 47, and 120 to 119. 

This is backed up by the fact that the `FuncDataQueueHandler` uses `PsychSignalSentiment` as the custom data to emit, and sets the `EndTime` parameter, causing it to be advanced 15s into the future:

https://github.com/QuantConnect/Lean/blob/cf1b6c340aae834058827db4c3c9a6cbbf98ad9b/Tests/Engine/DataFeeds/LiveTradingDataFeedTests.cs#L1319-L1345

@StefanoRaggi - Could you please review the test to see if my changes are the source of the test failing?

StefanoRaggi:
@gsalaz98 Given that the `EndTime` is being shifted 15 seconds forward, I believe the expected count should be one less, as reported by the failing test.
BTW, the tests were updated recently to throw for `PsychSignal` (as unsupported streaming data), in fact the `PsychSignalSentiment` tests are passing -- the `TiingoNews` tests are failing because the data type enqueued is `PsychSignalSentiment` in both cases 😃 
I think the test should be updated to enqueue the correct data type.

Common/Data/Custom/PsychSignal/PsychSignalSentiment.cs
Martin-Molinero: I believe the fill-forward interval should be provided directly from the subscription configuration https://github.com/QuantConnect/Lean/blob/master/Engine/DataFeeds/SubscriptionCollection.cs#L153 so it shouldn't be affected by this interval being directly added to the end time

PR #3812, opened 2019-11-08
Fix IB Forex zero conversion rates

### LINKED ISSUE #3811: Conversion rates not available for some Forex currencies with IB
#### Expected Behavior
Conversion rates for valid Forex pairs should always be available.

#### Actual Behavior
Conversion rates for a few currencies such as `SGD` and `HUF` cannot be set when deploying to the IB brokerage.

#### Potential Solution
The default **"market"** for IB Forex symbols is currently set to `Market.FXCM`.
As already mentioned in #3731, Oanda supports many more FX pairs than FXCM, so the solution could be to update the IB brokerage and brokerage model to use `Market.Oanda` instead of `Market.FXCM`.

#### Reproducing the Problem
- Deploy any algorithm to an IB account with some cash in `SGD`.
- At cash sync time -- or simply calling this method:
``` C#
Portfolio.CashBook.ConvertToAccountCurrency(100, "SGD")
```
the algorithm will be terminated with the following error:
```
System.ArgumentException: The conversion rate for SGD is not available.
```
This happens because there is no `USDSGD` symbol available with the FXCM market and the conversion rate for `SGD` is zero.

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- Updated `InteractiveBrokersBrokerage` and `InteractiveBrokersBrokerageModel` to use `Market.Oanda` instead of `Market.FXCM` when generating Forex symbols.
- Fixed a bug where the default brokerage model was being cached in the  `CurrencySubscriptionDataConfigManager` instead of the actual brokerage model set in the algorithm.

#### Related Issue
Closes #3811 

#### Motivation and Context
- Algorithm runtime error at cash sync time with algorithm deployed to IB with `SGD` or `HUF` cash in the account.

No.

- Converted some cash to SGD in the IB account
- Verified that the conversion rate is zero and the exception is thrown with current `master` code
- Verified that the conversion rate is non-zero and no exception is thrown using the PR code

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -2173,7 +2173,7 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
         {
             var securityType = ConvertSecurityType(contract);
             var ibSymbol = securityType == SecurityType.Forex ? contract.Symbol + contract.Currency : contract.Symbol;
-            var market = securityType == SecurityType.Forex ? Market.FXCM : Market.USA;
+            var market = securityType == SecurityType.Forex ? Market.Oanda : Market.USA;
 
             if (securityType == SecurityType.Future)
             {
@@ -2479,7 +2479,7 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
 
             return
                 (securityType == SecurityType.Equity && market == Market.USA) ||
-                (securityType == SecurityType.Forex && market == Market.FXCM) ||
+                (securityType == SecurityType.Forex && market == Market.Oanda) ||
                 (securityType == SecurityType.Option && market == Market.USA) ||
                 (securityType == SecurityType.Future);
         }
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -2173,7 +2173,8 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
         {
             var securityType = ConvertSecurityType(contract);
             var ibSymbol = securityType == SecurityType.Forex ? contract.Symbol + contract.Currency : contract.Symbol;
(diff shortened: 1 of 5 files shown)


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2479,7 +2479,7 @@ private bool CanSubscribe(Symbol symbol)
 
             return
                 (securityType == SecurityType.Equity && market == Market.USA) ||
-                (securityType == SecurityType.Forex && market == Market.FXCM) ||
+                (securityType == SecurityType.Forex && market == Market.Oanda) ||
                 (securityType == SecurityType.Option && market == Market.USA) ||
                 (securityType == SecurityType.Future);
```
Martin-Molinero: Maybe to avoid the duplication, could we use the `InteractiveBrokersBrokerageModel.DefaultMarketMap` directly for these checks and the one above (L2176)?

Tests/QuantConnect.Tests.csproj
```diff
@@ -756,7 +756,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
     <Content Include="TestData\FredVixData.json">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
Martin-Molinero: Nice one 👍

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Updated `MapSymbol` to reuse the `DefaultMarketMap`, but I would leave `Subscribe` as is for a few reasons:
- CFD trading with IB is not officially supported/tested
- With futures we are currently allowing all markets (will probably have to get back to this when we'll start addressing international markets)
- `Subscribe` is only called when deploying with the IB feed (but not with the QC feed)

PR #3801, opened 2019-11-04
Added constructors taking PyObject

prasadsom:
Added constructors taking `PyObject`

#### Description
Added overloaded constructor with `PyObject` as parameter. `PeriodCountConsolidatorBase` constructor convert `PyObject` to delegate .

#### Related Issue
#3780 
#### Motivation and Context
Though there is workaround available, suggestion is not to rely on `System` package. So overloaded C# class constructors

#### Requires Documentation Change
NA

#### How Has This Been Tested?
Using snippet in #3780 

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #3816, opened 2019-11-09
Adds Constructor Overload to PeriodCountConsolidatorBase that Accepts PyObject

### LINKED ISSUE #3780: Adds Constructor Overload to PeriodCountConsolidatorBase that Accepts PyObject
#### Expected Behavior
Be able to use the equivalent of `Func<DateTime, CalendarInfo> func` in python.

#### Actual Behavior
Can only use the `Expiry` helper class.

#### Potential Solution
Implement constructor overload that accepts a `PyObject` that represent a `Func<DateTime, CalendarInfo> func`

#### Reproducing the Problem
```python
class ConsolildatorFuncPy(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2014, 5, 1)
        self.SetEndDate(2014, 5, 15)
        
        self.AddForex("EURUSD")
        
        x = QuoteBarConsolidator(self.Func)
        x.DataConsolidated += self.OnDataConsolidated
        self.SubscriptionManager.AddConsolidator("EURUSD", x)

    def Func(self, dt):
        period = timedelta(days=1)
        start_time = datetime(dt.year, dt.month, dt.day, 17)
        if start_time > dt:
            start_time = start_time - period
        return CalendarInfo(start_time, period)

    def OnDataConsolidated(self, sender, bar):
        self.Debug(bar.EndTime)
```
#### Checklist

AlexCatarino:
#### Description
Adds constructor overload to `PeriodCountConsolidatorBase` (and inheritaded classed) that accepts `PyObject`.

#### Related Issue
Closes #3780 #3801 

Thank you, @prasadsom for #3801. Its commits are part of this pull request. I just added unit tests for all consolidators.

#### Motivation and Context
Be able to use the equivalent of `Func<DateTime, CalendarInfo>` func in Python.

Yes. In the consolidator's docs, there is no reference to `Func<DateTime, CalendarInfo>` contructor overload.

Extends existing unit tests to new `PyObject` overload.

#### Types of changes

#### Checklist:


Common/Data/Consolidators/TradeBarConsolidatorBase.cs
```diff
@@ -66,11 +67,17 @@ protected TradeBarConsolidatorBase(Func<DateTime, CalendarInfo> func)
...
+        protected TradeBarConsolidatorBase(PyObject pyfuncobj)
+            : base(pyfuncobj)
         {
-            get { return (TradeBar) WorkingData; }
         }
+
+        /// <summary>
+        /// Gets a copy of the current 'workingBar'.
+        /// </summary>
+        public TradeBar WorkingBar => (TradeBar) WorkingData;
```
mchandschuh: Why add this? It hides the base's property. _Also_ -- typically, if you're going to hide a base class property you should be explicit by using the `new` keyword:

``` csharp
public new TradeBar WorkingBar => (TradeBar)WorkingData;
```

I'm not a fan of this change. If the goal is to provide strongly typed data, then why not do it using generics? `PeriodCountConsolidatorBase<T, TradeBar>` exposes the correct types, so this definition could be pushed all the up (or down depending on how you look at it) the type heirarchy so that _everyone_ gets this behavior instead of just this class _and also_ we won't need to hide properties which can get yucky since behavior becomes dependent on the static (compile-time) type of an object. I've seen some pretty gnarly bugs from such semantics.

Feel free to reach out to discuss further.

mchandschuh:
Are both of these intended to be merged?
> https://github.com/QuantConnect/Lean/pull/3801

AlexCatarino:
@mchandschuh , this PR is meant to **replace** #3801. 
#3801 will be closed when this is merged.

Common/Data/Consolidators/TradeBarConsolidatorBase.cs
jaredbroad: @AlexCatarino ?

Martin-Molinero: The `WorkingBar` was already present, I believe he just refactored into an expression-bodied property, right @AlexCatarino ?

Tests/Common/Data/CalendarConsolidatorsTests.cs
```diff
@@ -26,8 +27,47 @@ namespace QuantConnect.Tests.Common.Data
     [TestFixture]
     public class CalendarConsolidatorsTests
     {
+        private Dictionary<Language, dynamic> _weeklyFuncDictionary;
+        private Dictionary<Language, dynamic> _monthlyFuncDictionary;
```
Martin-Molinero: Nice tests! 👍

Common/Data/Consolidators/TradeBarConsolidatorBase.cs
AlexCatarino: Right, @Martin-Molinero.

PR #3833, opened 2019-11-13
Corrected unit test

prasadsom:
Fixed Factor files unit test
#### Description 
There are two issues here. After downloading data from yahoo. Split factor is assigned as whole number instead of fraction. ie. 7 instead of 1/7
Second issue, in my opinion aapl.csv price factor numbers are slightly different from what they should be.
That's my assumption based on calculation of price factor in code. 
https://github.com/QuantConnect/Lean/blob/dfbbf0aba76a8023f83f84a2cbfd770879f2bd62/ToolBox/FactorFileGenerator.cs#L229

for date 20180208 according to calculation, which I think is correct ,
```
previousFactorFileRow.PriceFactor = 0.9961585
dividend.Value = 0.63
previousClosingPrice.Close = 155.16
previousFactorFileRow.SplitFactor = 1

```
with these values `priceFactor = 0.9920982` . Over 29 rows, difference between what's in file and calculated value falls outside tolerance. And hence test will fail. That's why I modified aapl.csv file
#### Related Issue

#3435 

NA

Passed unit test
#### Types of changes

#### Checklist:


ToolBox/YahooDownloader/YahooDataDownloader.cs
```diff
@@ -105,7 +105,7 @@ public Queue<BaseData> DownloadSplitAndDividendData(Symbol symbol, DateTime star
                         parsed.Add(new Split
                         {
                             Time = Parse.DateTimeExact(values[0].Replace("-", string.Empty), DateFormat.EightCharacter),
-                            Value = ParseAmount(values[1])
+                            Value = 1/ParseAmount(values[1])
```
prasadsom: In present form it's assigning whole value. eg. for 1 to 7 split. It's assigning 7 instead of 1/7

prasadsom:
> A possible solution is copying the test factor files into the TestData folder and redirect the test to that file.
In that , case would it be better to do comparison in different way in unit test?  May be by increasing tolerance for price factor comparison. In this case, two places after decimal(0.01) ?

Jay-Jay-D:
One of the characteristics of a good unit test is isolation, it shouldn't have a dependency on any outside factor.

If you check `AAPL` factor file history, you can see it changes. Therefore, if the sample data changes or is updated, it'll break the test.

That's why I insist on adding the testing factor file into the test data folder.

prasadsom:
Thanks @Jay-Jay-D , @Martin-Molinero I changed location of inverting split value. And modified calculation of pricefactor as mentioned in PR #3567 . This will fix unit test without changing factor file

PR #3837, opened 2019-11-14
Refactor/Clean-Up Composer and Improve LEAN Startup Performance

### LINKED ISSUE #3836: Clean up Composer, be mindful of performance and maintaining existing behavior
Another thing I'm wondering is  that `Composer.Reset()` was changed to **not** perform a reset. This seems misleading. I haven't changed that as my goal was to not change any behavior, but I'm thinking we should minimally make `Reset()` actually perform a reset again and then (if required) add a separate method that simply clears the `_exportedValues` cache -- but I suspect the reason `Reset()` was changed was to improve test speeds, in which case, perhaps we just don't call `Reset()` -- or again, if required, we could add another method that does what we it it in an appropriately named method.

Relevant commits:
> `e4fc00efbf876e39280344187017d707c325ca56`
> `9865fe63e0da90c85ba26ecd4fb0067b7bbab825`

#### Expected Behavior

Code should generally be kept clean and readable.

#### Actual Behavior

Over time, as is with all things, the code in `Composer.cs` has gotten a little unwieldy, making it hard to understand exactly what it's doing without really digging into it.

#### Potential Solution

Perform a refactor but be extremely mindful of performance as most of the recent changes in `Composer.cs` are aimed at improving LEAN startup times and unit test times.

#### Reproducing the Problem

Run LEAN. For timing, I used the following code in `Program.cs`'s `Main()` function:
<details><summary>Code snippet to reproduce timings used for this PR</summary>
``` csharp
static void Main(string[] args)
{
    var start = DateTime.UtcNow;

    //Initialize:
    var mode = "RELEASE";
    #if DEBUG
        mode = "DEBUG";
    #endif

    if (OS.IsWindows)
    {
        Console.OutputEncoding = System.Text.Encoding.UTF8;
    }

    // expect first argument to be config file name
    if (args.Length > 0)
    {
        Config.MergeCommandLineArgumentsWithConfiguration(LeanArgumentParser.ParseArguments(args));
    }

    var environment = Config.Get("environment");
    var liveMode = Config.GetBool("live-mode");
    Log.DebuggingEnabled = Config.GetBool("debug-mode");
    Log.LogHandler = Composer.Instance.GetExportedValueByTypeName<ILogHandler>(Config.Get("log-handler", "CompositeLogHandler"));

    //Name thread for the profiler:
    Thread.CurrentThread.Name = "Algorithm Analysis Thread";
    Log.Trace("Engine.Main(): LEAN ALGORITHMIC TRADING ENGINE v" + Globals.Version + " Mode: " + mode + " (" + (Environment.Is64BitProcess ? "64" : "32") + "bit)");
    Log.Trace("Engine.Main(): Started " + DateTime.Now.ToShortTimeString());

    //Import external libraries specific to physical server location (cloud/local)
    LeanEngineSystemHandlers leanEngineSystemHandlers;
    try
    {
        leanEngineSystemHandlers = LeanEngineSystemHandlers.FromConfiguration(Composer.Instance);
    }
    catch (CompositionException compositionException)
    {
        Log.Error("Engine.Main(): Failed to load library: " + compositionException);
        throw;
    }

    //Setup packeting, queue and controls system: These don't do much locally.
    leanEngineSystemHandlers.Initialize();

    //-> Pull job from QuantConnect job queue, or, pull local build:
    string assemblyPath;
    var job = leanEngineSystemHandlers.JobQueue.NextJob(out assemblyPath);

    if (job == null)
    {
        throw new Exception("Engine.Main(): Job was null.");
    }

    LeanEngineAlgorithmHandlers leanEngineAlgorithmHandlers;
    try
    {
        leanEngineAlgorithmHandlers = LeanEngineAlgorithmHandlers.FromConfiguration(Composer.Instance);
    }
    catch (CompositionException compositionException)
    {
        Log.Error("Engine.Main(): Failed to load library: " + compositionException);
        throw;
    }

    var end = DateTime.UtcNow;
    var elapsed = end - start;
    Log.Error($"COMPOSER INITIALIZATION/RESOLUTION TIME: {elapsed.TotalMilliseconds:0.00}ms");
```
</details>

#### Checklist

mchandschuh:
#### Description

This is mostly a clean up/refactoring of the existing logic and behavior.
The only real behavior change is not waiting for the CompositionContainer
to be fully resolved before finishing the init procedures. 90% of the call
we make to the Composer are QC types and therefore handled by the special
case code that explicitly loads QC assemblies and caches their types.

Another behavior change was made to really trim down the number of types
that were being cached from the QC assemblies. We were caching ALL public
types that weren't abstract/interface/enum, but since it's a proxy for MEF
we should really have the [InheritedExport] filter added. Adding this cut
down the number of QC types were caching by over 100x. Another perf gain
was in how we store the cached items. The existing code had the cache in
a list with over 10k types in it and then would search that list doing
an IsAssignableFrom and MatchesTypeName. That's a lot of items to hit.
This commit changes the storage to be a dictionarty keyed by the contract
type. We use the type that declared the [InheritedExport] as well as the
implementation type to support requested exact types, but I'm not sure
that this was required but at first glance it seemed to be supported by
the existing implementation.

This change cuts down LEAN's startup time (defined as the time it takes
to go from the top of Main to just following the initialization/resolution
of the algorithm handlers. This section of code's performance is 100%
dominated by the Composer's initialization and resolution functions. Avg
times on 'master' were around 350ms for a hot start and over 1.5s for a
cold start. Cold start meaning first run after fresh build. This change
shows an average hot start time of about 210ms and a cold start time of
about 350ms

Closes #3836

I've also included a `Composer.HardReset()` method that resets all state within the composer, including reloading assemblies from disk.

Surprisingly, using:
``` csharp
public static readonly Composer Instance = new Composer();
``` 
instead of a `Lazy<Composer>` or even _just_ a computed property, saved over ~150ms!! I never would have expected that to provide such a larger performance improvement.

#### Related Issue

See #3836 

#### Motivation and Context

While I'm not a fan of the `Composer` these days and would love to replace it... probably with a simple type name lookup to remove MEF... I still hold some nostalgia towards it, so it felt good to clean it up and make it look a little prettier. I initially went to look at it because it was taking ~15 seconds to start LEAN, but that was for a completely unrelated reason.

No.

Tested using the existing unit tests as well as thorough performance testing, including profiling as well as timing LEAN startup. See the issue for the timing snippet used.

#### Types of changes

- [x] Refactor w/ added bonus of performance improvement 😛 

#### Checklist:

### DIFF
--- a/Common/Util/Composer.cs
+++ b/Common/Util/Composer.cs
@@ -19,11 +19,11 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using System.ComponentModel.Composition.Hosting;
-using System.ComponentModel.Composition.Primitives;
 using System.ComponentModel.Composition.ReflectionModel;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using QuantConnect.Configuration;
 using QuantConnect.Logging;
@@ -35,13 +35,18 @@ namespace QuantConnect.Util
     /// </summary>
     public class Composer
     {
-        private static string PluginDirectory;
-        private static readonly Lazy<Composer> LazyComposer = new Lazy<Composer>(
-            () =>
-            {
-                PluginDirectory = Config.Get("plugin-directory");
-                return new Composer();
-            });
+        // this is purposefully a computed property to ensure setting 'plugin-directory' config will actually
+        // be used. the only requirement is that the value is set before accessing Composer.Instance
+        private static string PluginDirectory => Config.Get("plugin-directory");
+
+        // grab assemblies from current executing directory if not defined by 'composer-dll-directory' configuration key
+        private static readonly string PrimaryDllLookupDirectory = new DirectoryInfo(
+            Config.Get("composer-dll-directory", AppDomain.CurrentDomain.BaseDirectory)
+        ).FullName;
+
+        private static readonly bool LoadFromPluginDirectory = !string.IsNullOrWhiteSpace(PluginDirectory)
+            && new DirectoryInfo(PluginDirectory).FullName != PrimaryDllLookupDirectory
+            && Directory.Exists(PluginDirectory);
 
         /// <summary>
         /// Gets the singleton instance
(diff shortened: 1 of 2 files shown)


mchandschuh:
Taking this down to focus on the WFO PR. AFAIK, there wasn't a real issue, I just noticed some things. For posterity -- there was an error in the compiler with these changes:
> System.IndexOutOfRangeException: Index was outside the bounds of the array.
Also, there seems to be an issue with running the entire test suit w/ it hanging on launch -- likely a side-effect of an error being thrown from within a static initializer (all static field initializers are essentially executed via static constructors)

PR #3838, opened 2019-11-14
Fix IndicatorBase.Equal implementation 

### LINKED ISSUE #3835: Equal Indicators must be equal
#### Expected Behavior

Equal Indicators must be equal

#### Actual Behavior

Equal Indicators are not equal. Not even equal to itself.

#### Potential Solution

Fix bug in `Equal` override method for `BaseIndicator`.

https://github.com/QuantConnect/Lean/blob/b9fcddb60394c76586426a131fa6fa09d6073390/Indicators/IndicatorBase.cs#L177

#### Reproducing the Problem

```cs
[Test]
public void EqualIndicatorsAreEqual()
{
    var indicator1 = new SimpleMovingAverage(5);
    var indicator2 = new SimpleMovingAverage(5);
    Assert.True(indicator1.Equals(indicator2));
}

[Test]
public void IndicatorMustbeEqualToItself()
{
    var indicator1 = new SimpleMovingAverage(5);
    Assert.True(indicator1.Equals(indicator1));
}
```

#### System Information

Local Win7

#### Checklist

Jay-Jay-D:
#### Description

This PR fixes an issue with the overridden method `IndicatorBase.Equals`

#### Related Issue

Closes #3835

#### Motivation and Context

Objects must be equal to itself, at least :)

No.

Added unit test.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/IndicatorBase.cs
+++ b/Indicators/IndicatorBase.cs
@@ -192,7 +192,7 @@ namespace QuantConnect.Indicators
                 var cur = type.IsGenericType ? type.GetGenericTypeDefinition() : type;
                 if (typeof(IndicatorBase<>) == cur)
                 {
-                    return true;
+                    return ReferenceEquals(this, obj);
                 }
                 type = type.BaseType;
             }
--- a/Indicators/IndicatorBase.cs
+++ b/Indicators/IndicatorBase.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Diagnostics;
 using QuantConnect.Data;
+using QuantConnect.Data.Market;
 using QuantConnect.Logging;
 
 namespace QuantConnect.Indicators
@@ -184,7 +185,13 @@ namespace QuantConnect.Indicators
             // solely relying on reference semantics (think hashset/dictionary impls)
 
             if (ReferenceEquals(obj, null)) return false;
-            if (obj.GetType().IsSubclassOf(typeof (IndicatorBase<>))) return ReferenceEquals(this, obj);
+            var type = obj.GetType();   
+            if (type.IsSubclassOf(typeof(IndicatorBase<IndicatorDataPoint>))||
+                type.IsSubclassOf(typeof(IndicatorBase<IBaseDataBar>)) ||
+                type.IsSubclassOf(typeof(IndicatorBase<TradeBar>)))
+            {
+                return ReferenceEquals(this, obj);
+            }
 
             try
             {
--- a/Indicators/IndicatorBase.cs
+++ b/Indicators/IndicatorBase.cs
@@ -185,14 +185,18 @@ namespace QuantConnect.Indicators
             // solely relying on reference semantics (think hashset/dictionary impls)
 
             if (ReferenceEquals(obj, null)) return false;
-            var type = obj.GetType();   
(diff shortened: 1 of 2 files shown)


Indicators/IndicatorBase.cs
```diff
@@ -184,7 +185,13 @@ public override bool Equals(object obj)
             // solely relying on reference semantics (think hashset/dictionary impls)
 
             if (ReferenceEquals(obj, null)) return false;
-            if (obj.GetType().IsSubclassOf(typeof (IndicatorBase<>))) return ReferenceEquals(this, obj);
+            var type = obj.GetType();   
+            if (type.IsSubclassOf(typeof(IndicatorBase<IndicatorDataPoint>))||
+                type.IsSubclassOf(typeof(IndicatorBase<IBaseDataBar>)) ||
+                type.IsSubclassOf(typeof(IndicatorBase<TradeBar>)))
```
Martin-Molinero: Nice!
Wonder if it would be better to check for the generic interface, to support any `IndicatorBase<T>`, maybe something like https://stackoverflow.com/a/18233467?

Tests/Indicators/IndicatorTests.cs
```diff
@@ -155,6 +156,34 @@ public void EqualsMethodShouldNotThrowExceptions()
             Assert.IsFalse(res);
         }
 
+        [Test]
+        public void IndicatorMustBeEqualToItself()
```
Martin-Molinero: Awesome unit test! 👍

Indicators/IndicatorBase.cs
Jay-Jay-D: TBH, I'm not sure how to implement it for the generic `T`, that was the cause of the issue. 

Also, I'm not sue if it worth, this set of generics was taken from the `CommonIndicatorTests`, that is used for *all* indicators.
https://github.com/QuantConnect/Lean/blob/b8b89ab0b60e05dae10c6072e7f9ea1313b378de/Tests/Indicators/CommonIndicatorTests.cs#L49

On top of that, the test added covers more than 40 indicators.

Martin-Molinero: Actually something like this https://stackoverflow.com/a/457708 is what we would need, sorry for sending the wrong link, `IndicatorBase<T>` is a class, not an interface 🤦‍♂. I did a quick test and it seems to work correctly

Jay-Jay-D: Done!

Indicators/IndicatorBase.cs
```diff
@@ -184,8 +185,18 @@ public override bool Equals(object obj)
...
             if (ReferenceEquals(obj, null)) return false;
-            if (obj.GetType().IsSubclassOf(typeof (IndicatorBase<>))) return ReferenceEquals(this, obj);
+            var type = obj.GetType();
 
+            while (type != null && type != typeof(object))
+            {
+                var cur = type.IsGenericType ? type.GetGenericTypeDefinition() : type;
+                if (typeof(IndicatorBase<>) == cur)
+                {
+                    return true;
```
Martin-Molinero: Shouldn't it compare references here?

Martin-Molinero: I think we need a small unit tests which tests it returns false correctly

PR #3855, opened 2019-11-15
IB brokerage updates

StefanoRaggi:
#### Description
- Removed four IB error codes from the WarningCodes list since those are only informational messages from IB (not warnings or errors).
  - **2104** - _Market data farm connection is OK:usfarm_
  - **2106** - _A historical data farm is connected._
  - **2107** - _HMDS data farm connection is inactive but should be available upon demand.ushmds_
  - **2108** - _A market data farm connection has become inactive but should be available upon demand._
- Replaced the `BrokerageMessageEvent` with `Log.Trace` for rate-limited IB API calls.
- Fixed a potential IB duplicate order id error (**ErrorCode: 103**).
According to the IB API documentation, when calling `reqAllopenOrders` we need to ensure that future `orderId`s will be greater than the ones returned in the call:
https://interactivebrokers.github.io/tws-api/order_submission.html

#### Related Issue
n/a

#### Motivation and Context
- Improve IB messaging and removing unnecessary error logging
- Fix potential duplicate order id error

No.

Local and Cloud testing with IB deployments.

#### Types of changes

#### Checklist:


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -2222,11 +2246,11 @@ private static decimal RoundPrice(decimal input, decimal minTick)
         /// <returns>The new IB ID</returns>
         private int GetNextBrokerageOrderId()
         {
-            // spin until we get a next valid id, this should only execute if we create a new instance
-            // and immediately try to place an order
-            while (_nextValidId == 0) { Thread.Yield(); }
-
-            return Interlocked.Increment(ref _nextValidId);
+            lock (_nextValidIdLocker)
```
Martin-Molinero: Could there be a race condition between the call to `reqAllOpenOrders` and setting the new `_nextValidId` and another thread coming through `GetNextBrokerageOrderId()`, in live trading we use a separate thread for processing the orders. Maybe when we call `reqAllOpenOrders` we should already have the lock to `_nextValidIdLocker` so that we can update the id before any new usage

StefanoRaggi: I think the race condition between `reqAllOpenOrders` and `GetNextBrokerageOrderId` should never happen because `IBrokerage.GetOpenOrders` is only called once at startup in `BrokerageSetupHandler.Setup`

PR #3863, opened 2019-11-20
Added AccumulativeInsightPortfolioConstructionModel.

wardjm:
#### Description

See PR#3751

This commit aims to add a portfolio model that can be used for the competition. The model works as follows:
For insights of direction InsightDirection.Up, long targets are returned and
for insights of direction InsightDirection.Down, short targets are returned.
No rebalancing shall be done, as a new insight or the age of the insight shall determine whether to exit
the positions.
Rules:
1. On Up insight, increase position size by percent
2. On Down insight, decrease position size by percent
3. On Flat insight, move by percent towards 0
4. On expired insight, perform a Flat insight

#### Related Issue

See PR#3751

#### Motivation and Context

There currently is no portfolio model that can be used to test alphas directly related to insight production. They all involve rebalancing or equal weighting. This portfolio will initiate a new trade with every new insight. So 20 Up insights will generate 20 longs. They will be sold when the insight expires. They will be sold when a Flat insight is generated for the symbol. This is the most simple and direct approach for testing alphas. Of course a real portfolio would be more sophisticated and smarter about risk management.

I tried to keep all the inline documentation updated, but the site documentation could include how to use this model.

I tested on backtest algorithms using QuantConnect online while designing it. Several key components are added to the CI testing framework that cover proper long/short/flat scenarios. More complicated tests could of course be added later.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
+++ b/Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
@@ -1,0 +1,244 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Algorithm.Framework.Alphas;
+using QuantConnect.Data.UniverseSelection;
+
+namespace QuantConnect.Algorithm.Framework.Portfolio
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="IPortfolioConstructionModel"/> that allocates percent of account
+    /// to each insight, defaulting to 3%.
+    /// For insights of direction <see cref="InsightDirection.Up"/>, long targets are returned and
+    /// for insights of direction <see cref="InsightDirection.Down"/>, short targets are returned.
+    /// No rebalancing shall be done, as a new insight or the age of the insight shall determine whether to exit
+    /// the positions.
+    /// Rules:
+    ///    1. On Up insight, increase position size by percent
+    ///    2. On Down insight, decrease position size by percent
+    ///    3. On Flat insight, move by percent towards 0
+    ///    4. On expired insight, perform a Flat insight'''
+    /// </summary>
+    public class AccumulativeInsightPortfolioConstructionModel : PortfolioConstructionModel
+    {
+        private DateTime _rebalancingTime;
+        private List<Symbol> _removedSymbols = new List<Symbol>();
+        private readonly InsightCollection _insightCollection = new InsightCollection();
+        private DateTime? _nextExpiryTime;
+        private double _percent;
+        private Dictionary<Insight, int> usedInsight = new Dictionary<Insight, int>();
+        private Dictionary<Insight, int> expiredList = new Dictionary<Insight, int>();
+        private Dictionary<Symbol, double> positionSizes = new Dictionary<Symbol, double>();
+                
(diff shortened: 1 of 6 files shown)


Algorithm.Python/AccumulativeInsightPortfolioConstructionModel.py
```diff
@@ -0,0 +1,59 @@
...
+from QuantConnect.Algorithm.Framework.Portfolio import *
+from QuantConnect.Algorithm.Framework.Risk import *
+from QuantConnect.Algorithm.Framework.Selection import *
+from datetime import timedelta
+
+### <summary>
+### Test algorithm using 'AccumulativeInsightPortfolioConstructionModel.py' and 'ConstantAlphaModel'
+### generating a constant 'Insight' 
+### </summary>
+class ConfidenceWeightedFrameworkAlgorithm(QCAlgorithm):
```
Martin-Molinero: Thanks for adding an algorithm!
I think it would be nice for both file and algorithm name to be aligned, maybe something like `AccumulativeInsightPortfolioRegressionAlgorithm`.
Could we add the C# equivalent, I see it was present in the previous PR https://github.com/QuantConnect/Lean/pull/3751/files#diff-de9ca8289336617ab4798e31de371a72R30. 
Note that the implementation of the algorithm should be adjusted to assert on the behavior of this new portfolio construction model, current implementation was designed to test weighted based portfolio construction models. Same as for the expected statistics https://github.com/QuantConnect/Lean/pull/3751/files#diff-de9ca8289336617ab4798e31de371a72R77
*It seems this algorithm file is pending to be added to the project

Tests/Algorithm/Framework/Portfolio/AccumulativeInsightPortfolioConstructionModelTests.cs
```diff
@@ -0,0 +1,329 @@
...
+            var actualTargets = _algorithm.PortfolioConstruction.CreateTargets(_algorithm, insights).ToList();
+            Assert.AreEqual(2, actualTargets.Count);
+
+            AssertTargets(expectedTargets, actualTargets);
+        }
+
+        [Test]
+        [TestCase(Language.Python)]
+        [TestCase(Language.CSharp)]
+        public void GeneratesTargetsForInsightsWithNoConfidence(Language language)
```
Martin-Molinero: Some of these unit tests were designed for the weighted based portfolio construction models and won't really test this new one. It seems they should be replaced with a few specific unit tests, asserting on the accumulative behavior

Martin-Molinero:
Closing for now, pending requested changes

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,244 @@
...
+                return result;
+        }
+                
+        /// <summary>
+        /// Create portfolio targets from the specified insights
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance</param>
+        /// <param name="insights">The insights to create portfolio targets from</param>
+        /// <returns>An enumerable of portfolio targets to be sent to the execution model</returns>
+        public override IEnumerable<IPortfolioTarget> CreateTargets(QCAlgorithm algorithm, Insight[] insights)
```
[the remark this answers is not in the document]
wardjm: Sorry for the delayed response! I was trying to deliberately break away from this model since this is fundamentally different. All the behaviour comes from emitting insights and never from portfolio balancing. So I don't think there will ever be duplication of code and inheriting might even be misleading.

Tests/Algorithm/Framework/Portfolio/AccumulativeInsightPortfolioConstructionModelTests.cs
wardjm: I went through and made sure they were all updated and applied to this portfolio construction model (and deleted the ones that don't). The confidence one should behave correctly here. It is twofold: a) it does test a new condition, b) I'm looking forward to inheriting from this parent model to create a smarter AccumulativeInsight model that can use confidence. In that model, a confidence of 0 should not generate a target. So we should see two behaviours uniquely tested where we ensure updating something in the future doesn't break the tests.

Algorithm.Python/AccumulativeInsightPortfolioConstructionModel.py
wardjm: Sorry, this was one aspect that I was still unsure of, so it was put as a placeholder until we could get proper regression tests. I can look into it now.

wardjm: Is there a way to run these tests locally before pushing the commits up? I'm on Linux...

wardjm:
What's the right way to get the changes in here? I pushed them to my branch and thought they would automatically populate here.

PR #3873, opened 2019-11-24
Ensure that PeriodCountConsolidatorBase classes only consolidate a single symbol

### LINKED ISSUE #3872: Consolidators should only consolidate for a single symbol
#### Expected Behavior
Consolidators should not allow consolidation for multiple symbols.  Seeing more than one symbol in a consolidator is most likely a bug in an algorithm.

#### Actual Behavior
Consolidators allow tick/quote/tradebar data with multiple symbols.

#### Potential Solution
The symbol in the data being consolidated could be checked against the first symbol seen by the consolidator.

I have a change that I will send shortly.

#### Reproducing the Problem
I have a change that I will send shortly.

#### System Information
macOS (but not really relevant)

#### Checklist

RohanTalip:
#### Description
Ensures that consolidators derived from the PeriodCountConsolidatorBase class should only consolidate for a single symbol.

#### Related Issue

Closes #3872.

#### Motivation and Context

Without this change it is possible to incorrectly use a consolidator with multiple symbols, which can end up with incorrect data in the consolidator.

Probably not.  The exception should be self-explanatory.

Unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Consolidators/PeriodCountConsolidatorBase.cs
+++ b/Common/Data/Consolidators/PeriodCountConsolidatorBase.cs
@@ -114,6 +114,7 @@ namespace QuantConnect.Data.Consolidators
         /// For example, if time span is 1 minute, we have [10:00, 10:01): so data at 10:01 is not 
         /// included in the bar starting at 10:00.
         /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when multiple symbols are being consolidated.</exception>
         /// <param name="data">The new data for the consolidator</param>
         public override void Update(T data)
         {
@@ -123,7 +124,7 @@ namespace QuantConnect.Data.Consolidators
             }
             else if (symbol != data.Symbol)
             {
-                throw new Exception($"Previous consolidated symbol ({symbol}) is not the same as in the current data ({data.Symbol})");
+                throw new InvalidOperationException($"Previous consolidated symbol ({symbol}) is not the same as in the current data ({data.Symbol})");
             }
 
             if (!ShouldProcess(data))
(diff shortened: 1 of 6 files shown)


Common/Data/Consolidators/PeriodCountConsolidatorBase.cs
```diff
@@ -115,6 +117,15 @@ protected PeriodCountConsolidatorBase(Func<DateTime, CalendarInfo> func)
         /// <param name="data">The new data for the consolidator</param>
         public override void Update(T data)
         {
+            if (symbol == null)
+            {
+                symbol = data.Symbol;
+            }
+            else if (symbol != data.Symbol)
+            {
+                throw new Exception($"Previous consolidated symbol ({symbol}) is not the same as in the current data ({data.Symbol})");
```
Martin-Molinero: Maybe it would be more precise if instead of using `Exception` we used `InvalidOperationException`, what do you think?

Tests/Common/Data/TradeBarConsolidatorTests.cs
```diff
@@ -194,6 +194,41 @@ public void ConsolidatesOHLCV()
             Assert.AreEqual(300L, consolidated.Volume);
         }
 
+        [Test]
+        public void DoesNotConsolidateDifferentSymbols()
```
Martin-Molinero: Awesome unit tests 👍

Common/Data/Consolidators/PeriodCountConsolidatorBase.cs
RohanTalip: Good suggestion.  Done.

Tests/Common/Data/TradeBarConsolidatorTests.cs
RohanTalip: Thanks 😄

jaredbroad:
Thank you @RohanTalip!

PR #3898, opened 2019-12-08
Added initial CashAmount property to BacktestNodePacket

### LINKED ISSUE #3895: Add optional initial cash to BacktestNodePacket
#### Expected Behavior

- `BacktestNodePacket` allows specifying an optional initial `CashAmount` for an algorithm, that will override the algorithm set start cash
#### Actual Behavior

- To set a custom initial CashAmount requires modifying the algorithm
#### Potential Solution

- Implement
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

prasadsom:
#### Description
- Added optional `CashAmount` property to `BacktestNodePacket` class.
- Adding unit tests

#### Related Issue
Closes #3895 
#### Motivation and Context
By adding this property, able to set initial cash amount for algorithms without modifying alorithms

NA
Added unit test, regression tests, cloud backtest

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Packets/BacktestNodePacket.cs
+++ b/Common/Packets/BacktestNodePacket.cs
@@ -87,6 +87,11 @@ namespace QuantConnect.Packets
         public bool IsDebugging => Breakpoints.Any();
 
         /// <summary>
+        /// Optional initial cash amount if set
+        /// </summary>
+        [JsonProperty(PropertyName = "decCashAmount")]
+        public decimal? CashAmount;
+        /// <summary>
         /// Default constructor for JSON
         /// </summary>
         public BacktestNodePacket()
@@ -112,6 +117,7 @@ namespace QuantConnect.Packets
             ProjectId = projectId;
             UserPlan = userPlan;
             Name = name;
+            CashAmount = startingCapital;
             Language = Language.CSharp;
             Controls = new Controls
             {
--- a/Common/Packets/BacktestNodePacket.cs
+++ b/Common/Packets/BacktestNodePacket.cs
@@ -19,6 +19,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using Newtonsoft.Json;
+using QuantConnect.Securities;
 
 namespace QuantConnect.Packets
 {
@@ -89,8 +90,8 @@ namespace QuantConnect.Packets
         /// <summary>
         /// Optional initial cash amount if set
         /// </summary>
-        [JsonProperty(PropertyName = "decCashAmount")]
-        public decimal? CashAmount;
+        public CashAmount? CashAmount;
+
         /// <summary>
         /// Default constructor for JSON
         /// </summary>
@@ -109,7 +110,15 @@ namespace QuantConnect.Packets
         /// Initialize the backtest task packet.
         /// </summary>
         public BacktestNodePacket(int userId, int projectId, string sessionId, byte[] algorithmData, decimal startingCapital, string name, UserPlan userPlan = UserPlan.Free) 
-            : base (PacketType.BacktestNode)
+            : this (userId, projectId, sessionId, algorithmData, name, userPlan, new CashAmount(startingCapital, Currencies.USD))
+        {
+        }
+
+        /// <summary>
+        /// Initialize the backtest task packet.
+        /// </summary>
+        public BacktestNodePacket(int userId, int projectId, string sessionId, byte[] algorithmData, string name, UserPlan userPlan = UserPlan.Free, CashAmount? startingCapital = null)
+            : base(PacketType.BacktestNode)
         {
             UserId = userId;
             Algorithm = algorithmData;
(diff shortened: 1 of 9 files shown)


Common/Packets/BacktestNodePacket.cs
```diff
@@ -112,6 +117,7 @@ public BacktestNodePacket(int userId, int projectId, string sessionId, byte[] al
             ProjectId = projectId;
             UserPlan = userPlan;
             Name = name;
+            CashAmount = startingCapital;
```
prasadsom: startingCapital parameter was already passed in constructor, but was not used. Now assigned to new property created.

Common/Packets/BacktestNodePacket.cs
```diff
@@ -86,6 +86,11 @@ public class BacktestNodePacket : AlgorithmNodePacket
         /// </summary>
         public bool IsDebugging => Breakpoints.Any();
 
+        /// <summary>
+        /// Optional initial cash amount if set
+        /// </summary>
+        [JsonProperty(PropertyName = "decCashAmount")]
```
prasadsom: Was trying to find name convention for assigning Json property. Could not find one. Correction welcome

Martin-Molinero: I think for improved future proof, we should use a `List<CashAmount>`, each `CashAmount` instance holds a value and a currency. Regarding the JSON property name, it seems to me it should be `aCashAmounts`

Common/Packets/BacktestNodePacket.cs
Martin-Molinero: We might keep this constructor definition which defaults to `USD` currency and calls a new one which receives a `List<CashAmount>`.

Engine/Setup/BacktestingSetupHandler.cs
```diff
@@ -190,6 +190,12 @@ public bool Setup(SetupHandlerParameters parameters)
                         algorithm.SetEndDate(job.PeriodFinish.Value);
                     }
 
+                    //set initial cash, if present in the job
+                    if (job.CashAmount.HasValue)
```
Martin-Molinero: I believe we need to update `QuantConnect.Queues.JobQueue.NextJob() Line 150` not to send in a cash amount else it will always use it instead of the algorithm cash amount. This can be see when running the regression tests.

We should probably loop through the entire `algorithm.Portfolio.CashBook` and set those amounts to 0, similar to https://github.com/QuantConnect/Lean/blob/master/Engine/Setup/BrokerageSetupHandler.cs#L229 . `algorithm.SetCash()` will default to the account currency and might leave other currencies set.

Tests/Common/Packets/BacktestNodePacketTests.cs
```diff
@@ -102,5 +103,16 @@ public void RoundTripWithJobDates()
...
+        [Test]
+        public void InitialCashAmount()
+        {
+            var job = new BacktestNodePacket(1, 2, "3", null, 9m, $"{nameof(BacktestNodePacketTests)}.Pepe");
+            Assert.AreEqual(job.CashAmount.Value,9m);
+
+            job.CashAmount = 10m;
+            var serialized = JsonConvert.SerializeObject(job);
+            var job2 = JsonConvert.DeserializeObject<BacktestNodePacket>(serialized);
+            Assert.AreEqual(job.CashAmount, job2.CashAmount);
```
Martin-Molinero: I think we also need a specific unit test that asserts the initial cash amount is respected, similar to `JobDatesAreRespected`. 
Also lets please add another unit tests that asserts that other currencies set by the algorithm are also set to 0, not only the account currency, the use case pointed in the previous comment

Common/Packets/BacktestNodePacket.cs
mchandschuh: If the aim is to support multiple currencies, I would lean towards a dictionary instead of a list to enforce uniqueness.

Engine/Setup/BacktestingSetupHandler.cs
mchandschuh: Speaking of regression test, does it make sense to define a regression test for this feature? There's a lot of moving parts between job/setup and algorithm execution.

Tests/Common/Packets/BacktestNodePacketTests.cs
```diff
@@ -102,5 +103,125 @@ public void RoundTripWithJobDates()
...
+                },
+                Language.CSharp,
+                AlgorithmStatus.Completed);
+
+            AlgorithmRunner.RunLocalBacktest(parameter.Algorithm,
+                parameter.Statistics,
+                parameter.AlphaStatistics,
+                parameter.Language,
+                parameter.ExpectedFinalStatus,
+                initialCash: 1000000); // 1M vs 100K that is set in BasicTemplateDailyAlgorithm (10x)
```
jaredbroad: 🎉

prasadsom:
@Martin-Molinero thanks for finishing it. Apologies,  was off coding during holiday season, and couldn't reply.

Martin-Molinero:
> 
> 
> @Martin-Molinero thanks for finishing it. Apologies, was off coding during holiday season, and couldn't reply.

Hey @prasadsom ! no worries, thanks for giving it a shot 👍

PR #3903, opened 2019-12-10
Refactors Python Version of the API

### LINKED ISSUE #3896: Review Python Version of API
#### Expected Behavior
Python version of API should be able to make all requests successfully.

#### Actual Behavior
Some methods use params keyword instead of data to make requests. It leads to some issues.

#### Potential Solution
Review and correct.

#### Checklist

AlexCatarino:
#### Description
- Refactor to simplify `quantconnect.api.Api` class. Adds `Execute` method that accepts a bolean to distinguish a `POST` from a `GET` request.

In the previous version, some methods were using the `params` argument instead of `data` in the `POST` case which threw an exception for big json objects.

- Adds option to save logs and backtest report to disk.
- Adds `Result` class to optionally convert the json objects into `pandas.DataFrame` when getting backtest or live results

- Subversion bump

#### Related Issue
Closes #3896 

#### Motivation and Context
Fix bug. Improve.

No.

Unit tests and manually with real cases.

#### Types of changes

#### Checklist:


PythonToolbox/quantconnect/api.py
```diff
@@ -11,448 +11,497 @@
...
         Args:
             projectId(int): The project to which the file should be added.
             name(str): The name of the new file.
             content(str): The content of the new file.
         Returns:
-            ProjectResponse that contains information regarding the project.
-        """
-        request = Request('POST', "files/create",
-            params = 
+            Disctionary that includes information about the newly created file
```
jaredbroad: typo

PythonToolbox/tests/test_api.py
```diff
@@ -11,126 +11,169 @@
...
 import os
 
 class TestApi(unittest.TestCase):
 
     def setUp(self):
         # Please add your credentials. Found at https://www.quantconnect.com/account 
-        self.userId = ""
-        self.testToken = ""
+        self.userId = 0
+        self.testToken = ''
```
Martin-Molinero: Maybe should it try to load them from the config?

PythonToolbox/quantconnect/api.py
```diff
@@ -11,448 +11,497 @@
...
-    def __init__(self, userId, token):
-        self.api_connection = ApiConnection(userId, token)
-
-    def connected(self):
-        """Check if Api is successfully connected with correct credentials"""
-        return self.api_connection.connected()
-
-    def read_project(self, projectId):
-        """Get details about a single project.
+    '''QuantConnect.com Interaction Via API.
```
Martin-Molinero: Wondering why we need a python implementation of the API, couldn't we use PythonNet to use existing API implementation?

PythonToolbox/tests/test_api.py
AlexCatarino: Since we are not adding the config to the package, we are not using it. It would have to be copied because of its relative position.

PythonToolbox/quantconnect/api.py
AlexCatarino: We want to provide a 100% Python implementation to avoid pythonnet compatibility issues with the user python version.

PR #3906, opened 2019-12-10
Bug 3905 factor file.apply method removes the factor file first row

### LINKED ISSUE #3905: FactorFile.Apply method removes the factor file's first row
#### Expected Behavior

The `FactorFile.Apply` method should be able to add or update new events, without modifying the factor file structure.

#### Actual Behavior

When new events are added with the `Apply` method, the first row of the factor file is removed.

#### Potential Solution

Fix the logic when parsing `FactorFileRow` into events.

#### Reproducing the Problem

Add the following asserts:
```cs
Assert.True(actual.First().Date == new DateTime(1998, 01, 02),
                $"Factor file first row changed from 1998-01-02 to {actual.First().Date:yyyy-MM-dd} after applying new event");
```
to the [apply method tests](https://github.com/QuantConnect/Lean/blob/master/Tests/Common/Data/Auxiliary/FactorFileTests.cs#L198).

#### System Information

Windows 7

#### Checklist

Jay-Jay-D:
#### Description

This PR fixes the `FactorFile.Apply` method. Previously, it removed the first row of the `FactorFile` object after applying a new event.

Also the Apply method now can handle duplicated events.

Finally, This PR adds support for exponential notation in QuantQuote's factor files.

#### Related Issue

Closes #3905 

#### Motivation and Context

First row in factor file is important because the factor files are read backwards.

No.

Add new assertion to existing tests and added new tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Auxiliary/FactorFile.cs
+++ b/Common/Data/Auxiliary/FactorFile.cs
@@ -318,13 +318,7 @@ namespace QuantConnect.Data.Auxiliary
             var splitsAndDividends = GetSplitsAndDividends(data[0].Symbol, exchangeHours);
 
             var combinedData = splitsAndDividends.Concat(data)
-                .DistinctBy(e =>
-                    {
-                        var asSplit = e as Split;
-                        var eventType = asSplit == null ? "Dividend" : "Split";
-                        return $"{e.Symbol.ID}{eventType}{e.Time.ToStringInvariant(DateFormat.EightCharacter)}{e.Value.SmartRounding().Normalize()}";
-                    }
-                )
+                .DistinctBy(e => $"{e.GetType().Name}{e.Time.ToStringInvariant(DateFormat.EightCharacter)}")
                 .OrderByDescending(d => d.Time.Date);
 
             foreach (var datum in combinedData)
--- a/Common/Data/Auxiliary/FactorFile.cs
+++ b/Common/Data/Auxiliary/FactorFile.cs
@@ -1,17 +1,17 @@
-﻿﻿/*
- * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
- * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- *
+﻿/*
+* QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+* Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


Tests/Common/Data/Auxiliary/FactorFileTests.cs
```diff
@@ -227,8 +325,8 @@ public void AppliesSplitAndDividendAtSameTime()
             var factorFile = new FactorFile("AAPL", expected.SortedFactorFileData.Where(kvp => kvp.Value.PriceFactor >= .8m).Select(kvp => kvp.Value));
             var actual = factorFile.Apply(new List<BaseData>
             {
-                new Split(Symbols.SPY, reference.AddDays(-364), 100m, 1 / 2m, SplitType.SplitOccurred),
-                new Dividend(Symbols.SPY, reference.AddDays(-364), 12.5m, 100m)
+                new Split(Symbols.AAPL, reference.AddDays(-364), 100m, 1 / 2m, SplitType.SplitOccurred),
+                new Dividend(Symbols.AAPL, reference.AddDays(-364), 12.5m, 100m)
```
jaredbroad: 👍  Nice catch.

Common/Data/Auxiliary/FactorFile.cs
```diff
@@ -315,7 +315,16 @@ public FactorFile Apply(List<BaseData> data, SecurityExchangeHours exchangeHours
...
 
-            var combinedData = GetSplitsAndDividends(data[0].Symbol, exchangeHours).Concat(data)
+            var splitsAndDividends = GetSplitsAndDividends(data[0].Symbol, exchangeHours);
+
+            var combinedData = splitsAndDividends.Concat(data)
+                .DistinctBy(e =>
+                    {
+                        var asSplit = e as Split;
+                        var eventType = asSplit == null ? "Dividend" : "Split";
+                        return $"{e.Symbol.ID}{eventType}{e.Time.ToStringInvariant(DateFormat.EightCharacter)}{e.Value.SmartRounding().Normalize()}";
```
mchandschuh: Do we need the symbol id in this expression? Isn't everything the same symbol? We basically just want to remove duplicates, so I think something like `${e.GetType().Name}{e.Time.ToStringInvariant(DateFormat.EightCharacter)}"` should suffice

Common/Data/Auxiliary/FactorFileRow.cs
```diff
@@ -113,29 +113,20 @@ public static IEnumerable<FactorFileRow> Read(string permtick, string market, ou
...
-            var hasInfEntry = false;
             factorFileMinimumDate = null;
 
             var rows = new List<FactorFileRow>();
 
             // parse factor file lines
             foreach (var line in lines)
             {
-                if (line.Contains("inf"))
+                if (line.Contains("inf") || line.Contains("e+"))
```
mchandschuh: Hmm, does the `e+` actually mean it's inf? I suppose by convention we don't serialize using scientific notation, but perhaps more correctly we could define inf as `"inf"` or anything larger than, for example, `1e6`. If we are going to say that _all_ scientific notation equates to inf, then we'll want to document this behavior somewhere since it's unexpected. For example, a value such as `1.23e+2` would be flagged unexpectedly as an inf value

Tests/Common/Data/Auxiliary/FactorFileTests.cs
```diff
@@ -295,6 +393,9 @@ public void EmptyFactorFileReturnsEmptyListForSplitsAndDividends(string contents
             Assert.IsEmpty(factorFile.GetSplitsAndDividends(Symbols.SPY, SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork)));
         }
 
+      
+
```
mchandschuh: The style of the project uses a single blank line between functions.

Common/Data/Auxiliary/FactorFileRow.cs
Jay-Jay-D: Exponential notation is treated as `inf ` is because of the loss of precision. In all cases, the represented number is a round up to a few (and sometimes wrong) decimals, E.g., 1.6e+6 when the correct factor is  1562500. I'll add that as comment.

Worth notice that there are only 6 factor files with exponential notation, 10 cases with `inf`, and 2 cases where both cases are present.

Issue #1040, opened 2017-07-28
Virtual file system or object storage to fetch variables and state between live deployments


Maly-Lemire:
Would it be possible to provide a virtual file system with about 1mb available for free?
You could charge for more space.

I could do a pull request that provides a wrapper around File class or a simple Dictionary for saving variables with size and available space check. I would need a server side directory with proper access for saving the data. 

Thanks

jaredbroad:
Thanks @Maly-Lemire --  to understand the use case:
 - Are you storing data between backtests? (e.g. training data, train once, use it multiple times)
 - Or looking to literally model a file system?

"Temporary store" is a popular request and would be a well used feature. To make it compatible with our web platform need a design like this: A IStorageProvider which with Initialize, Get and Set methods for a key-value pair. Initialize could take the api class. 

It should be MEF'd in during the start up and set to the algorithm in in the SetupHandler via IAlgorithm. For LEAN you could implement a "FileSystemStorageProvider" which saves to a local directory.

4real:
Key value storage, preferably possible to share via backtest and live instances, and even edit externally, would be great. I'm using Amazon S3 for it now but I can't say I recommend it for the average user given the amount of time necessary to set it up.

Use cases:
- Saving/restoring algo state over restarts
- Warming up algorithm with backtest to present day for computation too heavy for Initialize
- Another way to give parameters to algorithm if some keys can be made accessible externally via GUI or REST

In fact, building this on S3 would be very viable.

jaredbroad:
- Would also be useful if this object was available as a stream. (For Accord library neural net loading)

Maly-Lemire:
Sorry about the lack of reply but I've managed to use warm up for what I needed to do. Since I won't need this feature, I may never do it. Sorry about that.

DMTSource:
Would be great to be able to work with pickled(python) files either taken from local, or created during a backtest. For example: You are working with DEAP, Tensorflow, or SKLearn to perform machine learning of some kind. In these cases a checkpoint or pickle file is created to persist the model in order to continue training, or to be used for inference live. It would be great to be able to create such a file and or upload it, and then read/write to it via an algo going forward.

jaredbroad:
👍 PR might look like this:

IObjectStorage {
    object Get(string key)
    void Put(string key, object value)
}

QCAlgorithm {
     IObjectStorage Storage;
}

//Usage like:
Algorithm.Storage.Get()
Algorithm.Storage.Put()

[...]

PR #3910, opened 2019-12-11
Fix Api connection null ref and UnzipData encoding

### LINKED ISSUE #3909: Fix null reference exception, support encoding in Compression.UnzipData
These two commits are pulled out of #1040 as they're a quick and easy review and should be merged regardless of that issue's final result.

#### Expected Behavior

* Null reference exceptions shouldn't be thrown when checking `result.Success`. This was happening when receiving an empty string as a response.
* Support specifying an `Encoding` in `Compression.UnzipData` which calls int `string.GetBytes(Encoding)`. This issue here was the `GetBytes` function defaults to `Encoding.ASCII` which is _almost_ never a good idea, unless you're absolutely positive the character set will be restricted to 7 bits.

#### Actual Behavior

Null refs abound and `GetBytes` assumes single byte characters, failed on unicode characters, which are valid keys of a dictionar (string is C# are UTF16, double wide)

#### Potential Solution

Check response for null and allow specifying `Encoding`

#### Reproducing the Problem

Visual inspection.

#### Checklist

mchandschuh:
#### Description

* Support encoding in Compression.UnzipData
* Prevent null ref and log unsuccessful api responses

#### Related Issue

Fixes #3909 

#### Motivation and Context

These changes were made as part of working on #1040 but are valuable in isolation, particularly the encoding change in compression.

No.

Unit test added for compression change and manually tested `ApiConnection` changes.

#### Types of changes

#### Checklist:

### DIFF
--- a/Api/ApiConnection.cs
+++ b/Api/ApiConnection.cs
@@ -109,11 +109,16 @@ namespace QuantConnect.Api
                     return false;
                 }
 
+                if (!restsharpResponse.IsSuccessful)
+                {
+                    Log.Error($"ApiConnect.TryRequest(): Content: {restsharpResponse.Content}");
+                }
+
                 responseContent = restsharpResponse.Content;
                 result = JsonConvert.DeserializeObject<T>(responseContent);
-                if (!result.Success)
+
+                if (result == null || !result.Success)
                 {
-                    //result;
                     return false;
                 }
             }
--- a/Compression/Compression.cs
+++ b/Compression/Compression.cs
@@ -192,8 +192,9 @@ namespace QuantConnect
         /// Uncompress zip data byte array into a dictionary string array of filename-contents.
         /// </summary>
         /// <param name="zipData">Byte data array of zip compressed information</param>
+        /// <param name="encoding">Specifies the encoding used to read the bytes. If not specified, defaults to ASCII</param>
         /// <returns>Uncompressed dictionary string-sting of files in the zip</returns>
-        public static Dictionary<string, string> UnzipData(byte[] zipData)
+        public static Dictionary<string, string> UnzipData(byte[] zipData, Encoding encoding = null)
         {
             // Initialize:
             var data = new Dictionary<string, string>();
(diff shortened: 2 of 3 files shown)


Api/ApiConnection.cs
```diff
@@ -109,11 +109,16 @@ public bool TryRequest<T>(RestRequest request, out T result)
                     return false;
                 }
 
+                if (!restsharpResponse.IsSuccessful)
+                {
+                    Log.Error($"ApiConnect.TryRequest(): Content: {restsharpResponse.Content}");
```
mchandschuh: I'm not sure if we want this here, but it was pretty useful for debugging. How often are we expecting unsuccessful messages, and is there any reason we _shouldn't_ log those responses?

Api/ApiConnection.cs
```diff
@@ -109,11 +109,16 @@ public bool TryRequest<T>(RestRequest request, out T result)
...
+                }
+
                 responseContent = restsharpResponse.Content;
                 result = JsonConvert.DeserializeObject<T>(responseContent);
-                if (!result.Success)
+
+                if (result == null || !result.Success)
```
StefanoRaggi: Nice catch! 
I was seeing quite a few null refs on these API calls in syslogs 👍

Api/ApiConnection.cs
StefanoRaggi: This error condition shouldn't be too frequent since exceptions should already be caught on L107,
agree with logging the error here.

Compression/Compression.cs
```diff
@@ -192,8 +192,9 @@ public static bool ZipCreateAppendData(string path, string entry, string data)
         /// Uncompress zip data byte array into a dictionary string array of filename-contents.
         /// </summary>
         /// <param name="zipData">Byte data array of zip compressed information</param>
+        /// <param name="encoding">Specifies the encoding used to read the bytes. If not specified, defaults to ASCII</param>
         /// <returns>Uncompressed dictionary string-sting of files in the zip</returns>
-        public static Dictionary<string, string> UnzipData(byte[] zipData)
+        public static Dictionary<string, string> UnzipData(byte[] zipData, Encoding encoding = null)
         {
```
StefanoRaggi: Nice!

PR #3911, opened 2019-12-11
Add IObjectStore/LocalObjectStore/QCAlgorithm.ObjectStore

### LINKED ISSUE #1040: Virtual file system or object storage to fetch variables and state between live deployments

mchandschuh:
#### Description

Defines a persistent object store accessible to algorithms. The implementation provided with this commit uses the local file system to enable persistence between algorithm runs. This information is intended to be project specific. On the local file system, individual folders are created for each algorithm's type name and inside the algorithm-specific folder is where the persisted data will be stored.

At run time, the objects persisted on disk are loading into RAM as `byte[]` to enable fast access to the data. Data is periodically flushed from RAM to disk to ensure the persistence it kept current in the case of the unfortunate app crash. Additionally, when the algorithm finishes or terminates, the object store's contents are flushed to disk.

NOTE: The first two commits in this PR are handled via #3910 and should be rebased away following merge.

#### Related Issue

Fixes #1040

#### Motivation and Context

This is a long desired featured by the community. A major use case is saving a particular compute-heavy model to the object store so it doesn't need to be recomputed on each and every backtest. In the same spirit, using a backtest to initialize some algorithm state before switching it over to live mode.

Yes, this feature should be added to the documentation

Tested locally with unit tests and manual execution as well as tested in the cloud.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/ObjectStoreExampleAlgorithm.cs
+++ b/Algorithm.CSharp/ObjectStoreExampleAlgorithm.cs
@@ -1,0 +1,133 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Data;
+using QuantConnect.Indicators;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+using QuantConnect.Storage;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This algorithm showcases some features of the <see cref="IObjectStore"/> feature.
+    /// One use case is to make consecutive backtests run faster by caching the results of
+    /// potentially time consuming operations. In this example, we save the results of a
+    /// history call. This pattern can be equally applied to a machine learning model being
+    /// trained and then saving the model weights in the object store.
+    /// </summary>
+    public class ObjectStoreExampleAlgorithm : QCAlgorithm
+    {
+        private const string SPY_Close_ObjectStore_Key = "spy_close";
+        private Security SPY;
+        private Identity SPY_Close;
+        private ExponentialMovingAverage SPY_Close_EMA10;
+        private ExponentialMovingAverage SPY_Close_EMA50;
+
+        // track last year of close and EMA10/EMA50
+        public readonly RollingWindow<IndicatorDataPoint> SPY_Close_History = new RollingWindow<IndicatorDataPoint>(252);
+        public readonly RollingWindow<IndicatorDataPoint> SPY_Close_EMA10_History = new RollingWindow<IndicatorDataPoint>(252);
+        public readonly RollingWindow<IndicatorDataPoint> SPY_Close_EMA50_History = new RollingWindow<IndicatorDataPoint>(252);
+
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 07);
+            SetEndDate(2013, 10, 11);
+
+            SPY = AddEquity("SPY", Resolution.Minute);
+
+            // define indicators on SPY daily closing prices
+            SPY_Close = Identity(SPY.Symbol, Resolution.Daily);
+            SPY_Close_EMA10 = SPY_Close.EMA(10);
+            SPY_Close_EMA50 = SPY_Close.EMA(50);
+
+            // each time an indicator is updated, push the value into our history rolling windows
+            SPY_Close.Updated += (sender, args) =>
+            {
+                // each time we receive new closing price data, push our window to the object store
+                SPY_Close_History.Add(args);
+            };
+
+            SPY_Close_EMA10.Updated += (sender, args) => SPY_Close_EMA10_History.Add(args);
+            SPY_Close_EMA50.Updated += (sender, args) => SPY_Close_EMA50_History.Add(args);
+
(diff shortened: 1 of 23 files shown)


Common/Packets/Controls.cs
```diff
@@ -96,6 +96,21 @@ public class Controls
...
+        [JsonProperty(PropertyName = "storageLimitMB")]
+        public int StorageLimitMB;
+
+        /// <summary>
+        /// Limits the number of files to be held under the <see cref="IObjectStore"/>
+        /// </summary>
+        [JsonProperty(PropertyName = "storageFileCountMB")]
+        public int StorageFileCount;
+
+        [JsonProperty(PropertyName = "persistenceIntervalSeconds")]
```
jaredbroad: Please add a comment;
"Delay between persisting data to the store"

Engine/Engine.cs
```diff
@@ -128,6 +128,9 @@ public void Run(AlgorithmNodePacket job, AlgorithmManager manager, string assemb
                     // initialize the alphas handler with the algorithm instance
                     AlgorithmHandlers.Alphas.Initialize(job, algorithm, SystemHandlers.Notify, SystemHandlers.Api);
 
+                    // initialize the object store
+                    AlgorithmHandlers.ObjectStore.Initialize(algorithm.Name, job.UserId, job.ProjectId, job.UserToken, job.Controls);
```
jaredbroad: It might be nice to include messaging abilities to report issues. It could be passed in here. `SystemHandlers.Notify`

Common/Packets/Controls.cs
mchandschuh: What are your thoughts on `IObjectStore` sporting some events? This enables inverting the dependencies so we're not injecting dependencies into the `IObjectStore`, instead, `QCAlgorithm` would bind handlers to the `IObjectStore.SaveFailed` event (or similar, perhaps a more generic `IObjectStore.ErrorRaised` event would be better).

Wiring up dependencies using events keeps components loosely coupled and also precludes us from having to add the _next_ dependency (whatever that may be). All that being said, if you prefer to push the dependencies into the `IObjectStore`, I'm happy to oblige :)

Engine/Engine.cs
mchandschuh: Also, are we still OK with errors being _thrown_ and then notifying the user about issues only during persistence since it's outside of their call stack or have the notifications for _all_ object store errors?

Engine/Storage/LocalObjectStore.cs
```diff
@@ -0,0 +1,335 @@
...
+        /// <summary>
+        /// Deletes the object data for the specified key
+        /// </summary>
+        /// <param name="key">The object key</param>
+        /// <returns>True if the delete operation was successful</returns>
+        public bool Delete(string key)
+        {
+            if (key == null)
+            {
+                throw new ArgumentNullException(nameof(key));
```
Martin-Molinero: I believe the dictionary will already throw `ArgumentNullException` if the key is null, not sure we need these extra checks? idem for the other methods

Common/Storage/ObjectStore.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Xml.Serialization;
+using Newtonsoft.Json;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Storage
+{
+    /// <summary>
+    /// Helper class for easier access to <see cref="IObjectStore"/> methods
+    /// </summary>
+    public static class ObjectStore
```
Martin-Molinero: Nice 👍

Algorithm.CSharp/ObjectStoreExampleAlgorithm.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// This algorithm showcases some features of the <see cref="IObjectStore"/> feature.
+    /// One use case is to make consecutive backtests run faster by caching the results of
+    /// potentially time consuming operations. In this example, we save the results of a
+    /// history call. This pattern can be equally applied to a machine learning model being
+    /// trained and then saving the model weights in the object store.
+    /// </summary>
+    public class ObjectStoreExampleAlgorithm : QCAlgorithm
```
Martin-Molinero: It would be nice to have the `py` version, sometimes pythonnet gives surprises 😅

Common/Packets/Controls.cs
```diff
@@ -112,6 +132,9 @@ public Controls()
             BacktestingMaxInsights = 10000;
             MaximumDataPointsPerChartSeries = 4000;
             SecondTimeOut = 300;
+            StorageLimitMB = 5;
+            StorageFileCount = 100;
+            PersistenceIntervalSeconds = 5;
```
Martin-Molinero: Maybe `PersistenceIntervalSeconds` could be a bit bigger. It seems that we are going to be storing a significant amount of data

jaredbroad: If dirty; for local object store only; it would store worst-case 5mb every 5 seconds to local disk.

Engine/Storage/LocalObjectStore.cs
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+            // read from RAM
+            var contents = ReadBytes(key);
+
+            // write byte array to storage directory
+            var path = Path.Combine(AlgorithmStorageRoot, $"{key.ToMD5()}.dat");
+            File.WriteAllBytes(path, contents);
```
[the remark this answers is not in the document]
jaredbroad: It's for some AI libraries who need their models to be loaded from physical files ( don't permit loading AI models via data ). This allows us to get a temp file to pass into Tensorflow for loading.

Engine/Storage/LocalObjectStore.cs
```diff
@@ -0,0 +1,335 @@
...
+                _persistenceTimer.Change(Timeout.Infinite, Timeout.Infinite);
+
+                Persist();
+
+                _persistenceTimer?.DisposeSafely();
+
+                // if the object store was not used, delete the empty storage directory created in Initialize
+                if (!Directory.EnumerateFileSystemEntries(AlgorithmStorageRoot).Any())
+                {
+                    Directory.Delete(AlgorithmStorageRoot);
```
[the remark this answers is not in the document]
jaredbroad: Suggest we should add checks to make it only store to root and not allow subdirectories

Common/Packets/Controls.cs
Martin-Molinero: oh right, I forgot about the if dirty and the 5MB limit 👍

Issue #3877, opened 2019-11-26
AlgorithmManager.ProcessSplitSymbols(): No subscriptions found for XXX, IsTradable: False, Active: True


AlexCatarino:
#### Expected Behavior
Avoid exception in `AlgorithmManager.ProcessSplitSymbols()`.

#### Actual Behavior
Exception is thrown when Universe Selection removes the subscription that user-defined security should keep.

#### Potential Solution
Do not remove a subscription if the security is active in the user-defines universe.

#### Reproducing the Problem
```csharp
namespace QuantConnect.Algorithm.CSharp {
    public class BasicTemplateAlgorithm : QCAlgorithm {
        
        private IEnumerable<Symbol> _symbols = new List<Symbol>();
        
        public override void Initialize() {
            SetStartDate(2005, 1, 1);
            SetEndDate(2005, 4, 1);
            SetCash(50000);

            UniverseSettings.Resolution = Resolution.Daily;
            AddUniverse(
                coarse => coarse
                    .Where(x => x.HasFundamentalData && x.Price > 5)
                    .OrderByDescending(x => x.DollarVolume)
                    .Take(100)
                    .Select(x => x.Symbol),

                fine => {
                    _symbols = fine.Select(x => x.Symbol);
                    return _symbols;
                });

            var spy = AddEquity("SPY", UniverseSettings.Resolution).Symbol;

            Schedule.On(DateRules.MonthStart(spy), TimeRules.At(10, 0), () =>
            {
                var weight = 0.99m / 10;
                foreach(var symbol in _symbols)
                {
                    if (Securities[symbol].IsTradable)
                    { 
                        AddEquity(symbol, UniverseSettings.Resolution);
                    }
                    SetHoldings(symbol, weight);
                }
            });
        }
    }
}
```
@Martin-Molinero, here is a [clone](https://www.quantconnect.com/terminal/#open/3460363).
#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #3915, opened 2019-12-12
Fix ProcessSplitSymbols exception

Martin-Molinero:
#### Description

- `UserDefinedUniverse` will no longer be removed as a data subscription.
- When `algorithm.AddData()` is called a universe selection data point
will be added to the `UserDefinedUniverse` subscription to trigger
selection and add the requested data.
- `DataManager` will make sure an active subscriptions
`SubscriptionDataConfig` will be present in the configuration collection
- Adding unit and regression tests

**Notes**:
- Forex regression algorithms now get an extra data point because added subscriptions start from the **current** algorithm data point, so **we will re-emit at that Time** with the new subscription data. (adding some logic we could potentially drop the data from this package)
- `RemoveSecurity()` call will be processed in the next loop.
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/3877
#### Motivation and Context

Fix exception, clean up

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/DataManager.cs
+++ b/Engine/DataFeeds/DataManager.cs
@@ -39,7 +39,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         private readonly MarketHoursDatabase _marketHoursDatabase;
         private readonly ITimeKeeper _timeKeeper;
         private readonly bool _liveMode;
-        private readonly IAlgorithm _algorithm;
         private readonly IRegisteredSecurityDataTypesProvider _registeredTypesProvider;
 
         /// There is no ConcurrentHashSet collection in .NET,
@@ -78,7 +77,6 @@ namespace QuantConnect.Lean.Engine.DataFeeds
             _marketHoursDatabase = marketHoursDatabase;
             _liveMode = liveMode;
             _registeredTypesProvider = registeredTypesProvider;
-            _algorithm = algorithm;
 
             // wire ourselves up to receive notifications when universes are added/removed
             algorithm.UniverseManager.CollectionChanged += (sender, args) =>
@@ -170,6 +168,10 @@ namespace QuantConnect.Lean.Engine.DataFeeds
         /// <returns>True if the subscription was created and added successfully, false otherwise</returns>
         public bool AddSubscription(SubscriptionRequest request)
         {
+            // guarantee the configuration is present in our config collection
+            // this is related to GH issue 3877: where we added a configuration which we also removed
+            _subscriptionManagerSubscriptions.TryAdd(request.Configuration, request.Configuration);
+
             Subscription subscription;
             if (DataFeedSubscriptions.TryGetValue(request.Configuration, out subscription))
             {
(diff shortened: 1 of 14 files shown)


Engine/DataFeeds/DataManager.cs
```diff
@@ -170,6 +168,10 @@ public void RemoveAllSubscriptions()
         /// <returns>True if the subscription was created and added successfully, false otherwise</returns>
         public bool AddSubscription(SubscriptionRequest request)
         {
+            // guarantee the configuration is present in our config collection
+            // this is related to GH issue 3877: where we added a configuration which we also removed
+            _subscriptionManagerSubscriptions.TryAdd(request.Configuration, request.Configuration);
```
mchandschuh: Interesting. This change looks good and make complete sense, interesting case there removing/re-adding subscriptions. I think previously we've thrown around ideas of reference counting the subscriptions.

Engine/DataFeeds/Enumerators/Factories/TimeTriggeredUniverseSubscriptionEnumeratorFactory.cs
```diff
@@ -52,8 +59,82 @@ public TimeTriggeredUniverseSubscriptionEnumeratorFactory(ITimeTriggeredUniverse
...
+                    var collection = new BaseDataCollection(time, symbol);
+
+                    ((InjectionEnumerator) enumerator).InjectDataPoint(collection);
+                };
+            }
+
+            return enumerator;
+        }
+
+        private class InjectionEnumerator : IEnumerator<BaseData>
```
mchandschuh: I've grown more comfortable with this guy's implementation over the last couple days. Nice work 👍

Engine/DataFeeds/SubscriptionFrontierTimeProvider.cs
```diff
@@ -57,12 +57,20 @@ private void UpdateCurrentTime()
             long earlyBirdTicks = MaxDateTimeTicks;
             foreach (var subscription in _subscriptionManager.DataFeedSubscriptions)
             {
-                if (subscription.Current == null
+                // this if should just be 'subscription.Current == null' but its affected by GH issue 3914
+                if (// this is a data subscription we just added
+                    // lets move it next to find the initial emit time
+                    subscription.Current == null
```
mchandschuh: per the `IEnumerator<T>` spec, `IEnumerator<T>.Current` is not defined _before_ a call to `MoveNext()` has happened and the framework implementations will throw an `InvalidOperationException`. I think our convention has been to initialize `Current` to null for data enumerators, so I think this is safe for our application, but wanted to make note of the broken contract here. This (and a few other instances) make me think LEAN would benefit from defining it's own `IEnumerator<BaseData>` type. The goal would be to support the same style of composition (decorators/russian dolls), but singling out data enumerators as a first-class citizen in LEAN's object model.

PR #3940, opened 2019-12-20
Feature accumulative model

wardjm:
See PR #3863

### DIFF
--- a/Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
+++ b/Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
@@ -36,7 +36,7 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
     /// </summary>
     public class AccumulativeInsightPortfolioConstructionModel : PortfolioConstructionModel
     {
-        private List<Symbol> _removedSymbols = new List<Symbol>();
+        private Array<Symbol> _removedSymbols = new Array<Symbol>();
         private readonly InsightCollection _insightCollection = new InsightCollection();
         private DateTime? _nextExpiryTime;
         private readonly double _percent;
@@ -47,30 +47,31 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
         /// <summary>
         /// Initialize a new instance of <see cref="AccumulativeInsightPortfolioConstructionModel"/>
         /// </summary>
-        /// <param name="percent">The amount of portfolio to allocate to a single insight</param>
+        /// <param name="percent">The percentage amount of the portfolio value to allocate 
+        /// to a single insight. The value of percent should be in the range [0,1]. 
+        /// The default value is 0.03.</param>
         public AccumulativeInsightPortfolioConstructionModel(double percent = 0.03)
         {
             _percent = Math.Abs(percent);
         }
-
        
         /// <summary>
         /// Method that will determine if the portfolio construction model should create a
         /// target for this insight
         /// </summary>
         /// <param name="insight">The insight to create a target for</param>
-        /// <returns>True if the portfolio should create a target for the insight</returns>
-        public virtual bool ShouldCreateTargetForInsight(Insight insight)
+        /// <returns>Determines if the portfolio construction model should create a target for this insight</returns>
+        protected virtual bool ShouldCreateTargetForInsight(Insight insight)
         {
             return true;
         }
 
         /// <summary>
-        /// Will determine the target percent for each insight
+        /// Determines the target percent for each insight
         /// </summary>
         /// <param name="activeInsights">The active insights to generate a target for</param>
         /// <returns>A target percent for each insight</returns>
-        public Dictionary<Insight, double> DetermineTargetPercent(ICollection<Insight> activeInsights)
+        private Dictionary<Insight, double> DetermineTargetPercent(IEnumerable<Insight> activeInsights)
         {
             var result = new Dictionary<Insight, double>();
 
@@ -78,40 +79,43 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
             {
                 if (usedInsight.ContainsKey(insight))
                 {
-                   continue;
+                    continue;
                 }
+
                 usedInsight[insight] = 1;
-                if (positionSizes.ContainsKey(insight.Symbol))
+
+                double targetPercent;
+                if (positionSizes.TryGetValue(insight.Symbol, out targetPercent))
                 {
-                   positionSizes[insight.Symbol] += _percent * (int)insight.Direction;
+                    targetPercent += _percent * (int)insight.Direction;
                 }
                 else
                 {
-                     positionSizes[insight.Symbol] = _percent * (int)insight.Direction;
+                    targetPercent = _percent * (int)insight.Direction;
                 }
 
-                if (insight.Direction == 0)
+                if (insight.Direction == InsightDirection.Flat)
                 {
                    // We received a Flat
 
                    // if adding or subtracting will push past 0, then make it 0
-                   if (Math.Abs(positionSizes[insight.Symbol]) < _percent)
+                   if (Math.Abs(targetPercent) < _percent)
                    {
-                      positionSizes[insight.Symbol] = 0;
+                        targetPercent = 0;
                    }
 
                    // otherwise, we flatten by percent
-                   if (positionSizes[insight.Symbol] > 0)
+                   if (targetPercent > 0)
                    {
-                        positionSizes[insight.Symbol] -= _percent;
+                        targetPercent -= _percent;
                    }
-                   if (positionSizes[insight.Symbol] < 0)
+                   if (targetPercent < 0)
                    {
-                      positionSizes[insight.Symbol] += _percent;
+                        targetPercent += _percent;
                    }
                 }
-            
-                result[insight] = positionSizes[insight.Symbol];
+                positionSizes[insight.Symbol] = targetPercent;
+                result[insight] = targetPercent;
             }
             return result;
         }
@@ -122,7 +126,7 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
         /// </summary>
         /// <param name ="activeInsights">The insights that are expiring and must be updated
         /// <returns>A target percent for each insight</returns>
-        public Dictionary<Insight, double> UpdateExpiredInsights(ICollection<Insight> activeInsights)
+        private Dictionary<Insight, double> UpdateExpiredInsights(IEnumberable<Insight> activeInsights)
         {
             var result = new Dictionary<Insight, double>();
         
@@ -132,17 +136,20 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
                 {
                     continue;
                 }
+                
                 expiredList[insight] = 1;
 
+                double targetPercent = positionSizes[insight.Symbol];
                 // if an expiring insight pushes it past 0, then flatten to 0
-                if ( (Math.Abs(positionSizes[insight.Symbol]) < _percent) && (insight.Direction != 0) )
+                if ( (Math.Abs(targetPercent) < _percent) && (insight.Direction != InsightDirection.Flat) )
                 {
-                    positionSizes[insight.Symbol] = 0;
+                    targetPercent = 0;
                 }
                 else
                 {
-                    positionSizes[insight.Symbol] -= _percent * (int)insight.Direction;
+                    targetPercent -= _percent * (int)insight.Direction;
                 }
+                positionSizes[insight.Symbol] = targetPercent;
                 result[insight] = positionSizes[insight.Symbol];
             }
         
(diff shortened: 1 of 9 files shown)


Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// <summary>
+        /// Initialize a new instance of <see cref="AccumulativeInsightPortfolioConstructionModel"/>
+        /// </summary>
+        /// <param name="percent">The amount of portfolio to allocate to a single insight</param>
+        public AccumulativeInsightPortfolioConstructionModel(double percent = 0.03)
+        {
+            _percent = Math.Abs(percent);
+        }
+
+
```
mchandschuh: nit - extra line break

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+        private DateTime? _nextExpiryTime;
+        private readonly double _percent;
+        private readonly Dictionary<Insight, int> usedInsight = new Dictionary<Insight, int>();
+        private readonly Dictionary<Insight, int> expiredList = new Dictionary<Insight, int>();
+        private readonly Dictionary<Symbol, double> positionSizes = new Dictionary<Symbol, double>();
+                
+        /// <summary>
+        /// Initialize a new instance of <see cref="AccumulativeInsightPortfolioConstructionModel"/>
+        /// </summary>
+        /// <param name="percent">The amount of portfolio to allocate to a single insight</param>
```
mchandschuh: Doc should include the range. Since it's called  percent, users may think its in the range of [0, 100], but it appears to be within the range of [0, 1]

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+        public AccumulativeInsightPortfolioConstructionModel(double percent = 0.03)
+        {
+            _percent = Math.Abs(percent);
+        }
+
+       
+        /// <summary>
+        /// Method that will determine if the portfolio construction model should create a
```
mchandschuh: nit - I like to phrase doc in  the current vs future tense:
``` csharp
/// <summary>
/// Determines if the portfolio construction model should create a target for this insight
/// </summary>

File: Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs [hunk 4]
```diff
@@ -0,0 +1,243 @@
...
+        /// </summary>
+        /// <param name="insight">The insight to create a target for</param>
+        /// <returns>True if the portfolio should create a target for the insight</returns>
+        public virtual bool ShouldCreateTargetForInsight(Insight insight)
+        {
+            return true;
+        }
+
+        /// <summary>
+        /// Will determine the target percent for each insight
```
mchandschuh: as another example:

``` xml
/// <summary>
/// Determines the target percent for each insight
/// </summary>
``

File: Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs [hunk 5]
```diff
@@ -0,0 +1,243 @@
...
+        /// </summary>
+        /// <param name="activeInsights">The active insights to generate a target for</param>
+        /// <returns>A target percent for each insight</returns>
+        public Dictionary<Insight, double> DetermineTargetPercent(ICollection<Insight> activeInsights)
```
mchandschuh: Should make this parameter `IEnumerable<Insight>`. In general, you want to select the most general and least restrictive interface possible. In this case, the only action performed on `activeInsights` is enumeration, which is best communicated via this `IEnumerable<T>` interface.

mchandschuh: also -- does this method need to be public? it looks like an internal implementation detail.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+       
+        /// <summary>
+        /// Method that will determine if the portfolio construction model should create a
+        /// target for this insight
+        /// </summary>
+        /// <param name="insight">The insight to create a target for</param>
+        /// <returns>True if the portfolio should create a target for the insight</returns>
+        public virtual bool ShouldCreateTargetForInsight(Insight insight)
```
mchandschuh: Should this method be public? Since this class is a model, the only methods called publicly on it will be the ones defined on the interface. This might be better suited as a `protected` method. This visibility modifier will still permit derived classes to invoke it in support of the `virtual` modifier.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+            foreach (var insight in activeInsights)
+            {
+                if (usedInsight.ContainsKey(insight))
+                {
+                   continue;
+                }
+                usedInsight[insight] = 1;
+                if (positionSizes.ContainsKey(insight.Symbol))
+                {
+                   positionSizes[insight.Symbol] += _percent * (int)insight.Direction;
```
mchandschuh: _never_ use a dictionary like this :)

You want to pull the value out into a variable and work with the variable. Using a dictionary in this way causes many many many invocations to `Symbol.GetHashCode()` and `Symbol.Equals()` while also constantly updating the internal state of the dictionary via the `this[index]` getter/setter.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+            var result = new Dictionary<Insight, double>();
+
+            foreach (var insight in activeInsights)
+            {
+                if (usedInsight.ContainsKey(insight))
+                {
+                   continue;
+                }
+                usedInsight[insight] = 1;
+                if (positionSizes.ContainsKey(insight.Symbol))
```
mchandschuh: this can be converted into:
``` csharp
double targetPercent;
if (positionSizes.TryGetValue(insight.Symbol, out targetPercent)
{
    // code that acts on 'targetPercent'
}
else
{
    targetPercent = _percent * (int)insight.Direction;
}
```

And then, all the way at the bottom of the loop:
``` csharp
positionSizes[insight.Symbol] = targetPercent;
```

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+                {
+                   positionSizes[insight.Symbol] += _percent * (int)insight.Direction;
+                }
+                else
+                {
+                     positionSizes[insight.Symbol] = _percent * (int)insight.Direction;
+                }
+
+                if (insight.Direction == 0)
```
mchandschuh: I believe `insight.Direction` is of type `InsightDirection` which is an enum. For readability purposes, please use the enumerated values as they much better capture the intent. Using numeric literals is akin to using magic numbers.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+            return result;
+        }
+
+        /// <summary>
+        /// Will determine the target percent for each expired insight (since expired insights
+        /// will remove the amount accumulated in portfolio)
+        /// </summary>
+        /// <param name ="activeInsights">The insights that are expiring and must be updated
+        /// <returns>A target percent for each insight</returns>
+        public Dictionary<Insight, double> UpdateExpiredInsights(ICollection<Insight> activeInsights)
```
mchandschuh: Does this need to be public? Also, `activeInsights` can be an `IEnumerable<Insight>` instead.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+                
+        /// <summary>
+        /// Create portfolio targets from the specified insights
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance</param>
+        /// <param name="insights">The insights to create portfolio targets from</param>
+        /// <returns>An enumerable of portfolio targets to be sent to the execution model</returns>
+        public override IEnumerable<IPortfolioTarget> CreateTargets(QCAlgorithm algorithm, Insight[] insights)
+        {
+            var targets = new List<IPortfolioTarget>();
```
mchandschuh: This method can be converted into using the `yield` syntax. I'll post a comment at each line to help you with the conversion 👍 

Yield syntax instructs the compiler to create an enumerable generator block. At compile time, an `IEnumerator<T>` instance will be generated from your method body. Pretty cool!

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            if (algorithm.UtcTime <= _nextExpiryTime &&
+                insights.Length == 0 &&
+                _removedSymbols == null)
+            {
+                return targets;
```
mchandschuh: The `yield break` statement is equivalent to a `return` statement, except it doesn't return anything, instead it signals that the enumeration of the collection has ended.

conversion:
``` csharp
yield break;
```

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+            }
+
+            // Validate we should create a target for this insight
+            _insightCollection.AddRange(insights.Where(ShouldCreateTargetForInsight));
+
+            // Create flatten target for each security that was removed from the universe
+            if (_removedSymbols != null)
+            {
+                var universeDeselectionTargets = _removedSymbols.Select(symbol => new PortfolioTarget(symbol, 0));
+                targets.AddRange(universeDeselectionTargets);
```
mchandschuh: We can `yield` each of these targets directly.

conversion:
``` csharp
foreach (var target in _removedSymbols.Select(symbol => new PortoflioTarget(symbol,0))
{
    yield return target;
}
```

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            // Determine target percent for the given insights
+            var percents = DetermineTargetPercent(lastActiveInsights);
+
+            foreach (var insight in percents.Keys)
+            {
+                var target = PortfolioTarget.Percent(algorithm, insight.Symbol, percents[insight]);
+                if (target != null)
+                {
+                    targets.Add(target);
```
mchandschuh: This can be directly `yield`ed.

conversion:
``` csharp
yield return target;
```

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+            var expiredInsights = _insightCollection.RemoveExpiredInsights(algorithm.UtcTime);
+
+                        percents = UpdateExpiredInsights(expiredInsights);
+
+            foreach (var insight in percents.Keys)
+            {
+                var target = PortfolioTarget.Percent(algorithm, insight.Symbol, percents[insight]);
+                if (target != null)
+                {
+                    targets.Add(target);
```
mchandschuh: conversion:
``` csharp
yield return target;
```

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+                }
+                else
+                {
+                    errorSymbols.Add(insight.Symbol);
+                }
+            }
+
+            _nextExpiryTime = _insightCollection.GetNextExpiryTime();
+
+            return targets;
```
mchandschuh: conversion:
delete this line.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+                else
+                {
+                    errorSymbols.Add(insight.Symbol);
+                }
+            }
+
+            // Get expired insights and create flatten targets for each symbol
+            var expiredInsights = _insightCollection.RemoveExpiredInsights(algorithm.UtcTime);
+
+                        percents = UpdateExpiredInsights(expiredInsights);
```
mchandschuh: nit - indentation here is wrong.

Algorithm.Framework/Portfolio/AccumulativeInsightPortfolioConstructionModel.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        /// <summary>
+        /// Event fired each time the we add/remove securities from the data feed
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance that experienced the change in securities</param>
+        /// <param name="changes">The security additions and removals from the algorithm</param>
+        public override void OnSecuritiesChanged(QCAlgorithm algorithm, SecurityChanges changes)
+        {
+            // Get removed symbol and invalidate them in the insight collection
+            _removedSymbols = changes.RemovedSecurities.Select(x => x.Symbol).ToList();
```
mchandschuh: We could make `_removedSymbols` and array (`Symbol[]`) -- and then we won't need to use `ToArray()` below.I checked elsewhere in this file and it doesn't appear we're using any of the list features.

wardjm:
Thanks for the feedback! I'm completely brand new to C#, but been programming Python/C/C++ for a long time. My main question is kind of github related probably, but is there any way to get these changes reflected in this pull request, or do I have to submit a new PR after updating my files?

jaredbroad:
Hi @wardjm you could put new commits on the end of your personal `wardjm:feature-accumulative-model` branch and they'll automatically show up in this pull-request.

image

wardjm:
Sorry I'm apparently going to have to wait until I can get back to my normal dev box with mono installed so I can sort this out in private...

jaredbroad:
Great work @wardjm 🎉

PR #3955, opened 2019-12-31
Move IObjectStore Extension Methods to Interface

AlexCatarino:
#### Description
Moves IObjectStore extension methods to interface.
Creates abstract class to implement common methods.

#### Related Issue
Closes #3952

#### Motivation and Context
Pythonnet does not support extension methods, therefore using helper methods would require a different API call in C# and Python.

#### Requires Documentation Change
No.

#### How Has This Been Tested?
Existing unit tests.

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

Issue #3952, opened 2019-12-31
Extension Methods for ObjectStore Not Supported in Python


AlexCatarino:
#### Expected Behavior
Be able to use `Read`, `ReadString`, `ReadJson`, `ReadXml` and `Save` extension methods in both C# and Python algorithms.

#### Actual Behavior
Not supported. pythonnet doesn't support extension methods.

#### Potential Solution
Move those methods to the `IObjectStore` interface or create a base class that implement them.

#### Reproducing the Problem
```x = self.ObjectStore.Read("my_saved_data")```

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

Martin-Molinero:
Close by https://github.com/QuantConnect/Lean/pull/3956

PR #3959, opened 2020-01-02
Adds Python Example of ObjectStoreExampleAlgorithm

AlexCatarino:
#### Description
Adds python example of `ObjectStoreExampleAlgorithm`
Minor change in C# version to keep consistency between examples.

#### Related Issue
Ref.: #3952 #3955 #3956 

#### Motivation and Context
Keep examples in C# and Python if possible.

Yes. This PR provides an example that can be added to the docs.

Executing the new algorithm. Results match C# version.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Python/ObjectStoreExampleAlgorithm.py
+++ b/Algorithm.Python/ObjectStoreExampleAlgorithm.py
@@ -74,7 +74,7 @@ class ObjectStoreExampleAlgorithm(QCAlgorithm):
 
             # save our warm up data so next time we don't need to issue the history request
             self.ObjectStore.Save(self.SPY_Close_ObjectStore_Key,
-                '\n'.join([f'{x.EndTime},{x.Value}' for x in sorted(self.SPY_Close_History, key=lambda x: x.EndTime)]))
+                '\n'.join(reversed([f'{x.EndTime},{x.Value}' for x in self.SPY_Close_History])))
 
             # Can also use ObjectStore.SaveBytes(key, byte[])
             # and to read  ObjectStore.ReadBytes(key) => byte[]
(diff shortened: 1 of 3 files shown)


Algorithm.Python/ObjectStoreExampleAlgorithm.py
```diff
@@ -0,0 +1,102 @@
...
+            # if our object store doesn't have our data, fetch the history to initialize
+            # we're pulling the last year's worth of SPY daily trade bars to fee into our indicators
+            history = self.History(self.SPY, timedelta(365), Resolution.Daily).close.unstack(0).squeeze()
+
+            for time, close in history.iteritems():
+                self.SPY_Close.Update(time, close)
+
+            # save our warm up data so next time we don't need to issue the history request
+            self.ObjectStore.Save(self.SPY_Close_ObjectStore_Key,
+                '\n'.join([f'{x.EndTime},{x.Value}' for x in sorted(self.SPY_Close_History, key=lambda x: x.EndTime)]))
```
Martin-Molinero: A couple of doubts:
- Interesting that we have to sort the RollingWindow, shouldn't it already be sorted? Seems we are doing the same for C# though
- Wondering if we could directly store the RollingWindow data points, not creating an intermediate string item?

AlexCatarino: - The `RollingWindow` objects are sorted in reverse order. The most recent item is found in index 0.
- We may store it directly, but how would we read the storage? Thinking about that, I choose `csv` format because pandas can read it. We need to come up with an example using a machine learning model.

Martin-Molinero: - Rolling window: maybe we should use `reverse` for C# and Py instead, else its kind of implying the data isn't sorted at all
- Seems we need a specific issue (or we could use this one) to add python Tests for this and fix/add new overloads if required

jaredbroad:
Let's make a more realistic example. Perhaps some expensive training where you want to save the state of the ml model, or save that an event has happened. Or save the count of the number of deployments etc.. something which the `History` API cannot do.

PR #3979, opened 2020-01-08
Fixes incorrect beta calculation

### LINKED ISSUE #3927: Beta Calculation Incorrect
The Beta Calculation in the strategy overview always shows low beta for every strategy even buy and hold the Benchmark. 

I think the issue stems from the way the overview is lining up the strategy and benchmark performance data. It seems the first entry of the benchmark data is duplicated meaning everything else is off by a day when comparing.

gsalaz98:
#### Description

Fixes incorrect beta calculation

* Reworks Sampling portion of AlgorithmManager loop
* Fixes issue with miscalculation of various portfolio statistics
* Fixes data alignment issue in statistics calculation

We now truncate the first few values of the "Daily Performance"
and benchmark performance series in order to align the series
correctly and attain accurate statistics at the cost of reducing
our sample size by 1 or 2 in the StatisticsBuilder class.

The shape of the data has also changed due to these changes.
Changes to the shape are:
  * The zero at the end of the "Daily Performance" series is gone
  * "Daily Performance" series now begins with zero (may be a bug, would like to discuss below).

We also now sample at the end of the AlgorithmManager loop to ensure
that the sampling events are fired even if we have no remaining points
left. This means that sampling after the Stream(...) loop is no longer
needed (if the algorithm successfully ran to the end), though we do
still force sample at the end just in case.

And finally, we've removed the Sample...(...) methods from the
AlgorithmManager and are now sampling inside the
`results.ProcessSynchronousEvents()` call.

#### Related Issue

Closes #3927 
#### Motivation and Context

Ensure statistics are calculated correctly

Users of the `BacktestResultPacket`, `LiveResultPacket`, `BacktestResult`, `LiveResult`  classes should be aware of this change since existing behavior has changed.

New unit test, manual verification of statistics. Tested on QC cloud, potential plotting issue.

#### Types of changes

#### Checklist:


Common/Statistics/StatisticsBuilder.cs
```diff
@@ -280,49 +302,41 @@ private static List<double> CreateBenchmarkDifferences(SortedDictionary<DateTime
...
-                listPerformance.Add(0);
-                Log.Trace("StatisticsBuilder.EnsureSameLength(): Padded Performance");
-            }
-            while (listPerformance.Count > listBenchmark.Count)
-            {
-                listBenchmark.Add(0);
-                Log.Trace("StatisticsBuilder.EnsureSameLength(): Padded Benchmark");
-            }
+            var resampledBenchmark = benchmark.GroupBy(x => x.Key.Date)
+                .Select(kvp => kvp.Last())
```
Martin-Molinero: Wondering if we can trust the result of the GroupBy operation to be sorted?

Engine/Results/BacktestingResultHandler.cs
```diff
@@ -772,14 +775,43 @@ public virtual void ProcessSynchronousEvents(bool forceProcess = false)
...
+                // which is used to save on bandwidth costs when uploading interim algorithm results to the cloud.
+                SampleEquity(time, Math.Round(currentPortfolioValue, 4));
+
+                // The benchmark will have an invalid first value since we don't
+                // calculate the percentage change between open and close for the first point.
+                //
+                // The performance series will contain the cumulative percent gain from market open to close
+                // for the first data point. This means that the first point between benchmark and performance
+                // are incompatible with each other, and must be removed for accurate calculations.
+                var portfolioValuePercentChange = _previousDailyPortfolioValue == 0 ? 0 : Math.Round((currentPortfolioValue - _previousDailyPortfolioValue) * 100 / _previousDailyPortfolioValue, 10);
```
Martin-Molinero: I think we are missing to tweak the logic to sample the saved portfolio value vs the current, else it seems it will use the current value, which is in the next date already?

Engine/Results/BacktestingResultHandler.cs
```diff
@@ -772,14 +775,43 @@ public virtual void ProcessSynchronousEvents(bool forceProcess = false)
...
             var time = Algorithm.UtcTime;
+            var shouldSample = _previousProcessTime.Date != time.Date;
+            var currentPortfolioValue = Algorithm.Portfolio.TotalPortfolioValue;
+
+            if (shouldSample)
+            {
+                // Sample all of the metrics daily once we're certain that all feeds
+                // and prices have been setup properly. In the past, we sampled at the beginning of the AlgorithmManager loop,
+                // which caused issues with the alignment of data in the statistics calculation phase.
+                SampleBenchmark(time.Date, Algorithm.Benchmark.Evaluate(time).SmartRounding());
```
Martin-Molinero: I think we could make these `Sample` methods protected?

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1070,6 +1073,26 @@ protected virtual string CreateSafeChartName(string chartName)
...
+            if (time.Date != _previousProcessTime.Date)
+            {
+                // The benchmark will have an invalid first value since we don't
+                // calculate the percentage change between open and close for the first point.
+                //
+                // The performance series will contain the cumulative percent gain from market open to close
+                // for the first data point. This means that the first point between benchmark and performance
+                // are incompatible with each other, and must be removed for accurate calculations.
+                var portfolioValuePercentChange = _previousDailyPortfolioValue == 0 ? 0 : Math.Round((currentPortfolioValue - _previousDailyPortfolioValue) * 100 / _previousDailyPortfolioValue, 10);
+                SamplePerformance(time.Date, portfolioValuePercentChange);
```
Martin-Molinero: Were we not sampling performance for live trading before, seems like a bug? 🤔 

_Maybe_ some of this logic, private fields/calls (for example _previousDailyPortfolioValue/_previousProcessTime) could be moved to the shared `BaseResultsHandler` ?

jaredbroad:
Closing pending new solution.

PR #3981, opened 2020-01-08
Implements PortfolioBias in EWPCM and IWPCM

### LINKED ISSUE #3769: Implements Portfolio Bias in Portfolio Construction Models
#### Expected Behavior
Be able to create Portfolio Construction Models with Portfolio Bias. PCM with PortfolioBias.Long will convert short signals to flat and adjust the targets proportionally.

#### Actual Behavior
The portfolio bias is a concern of the alpha models.

#### Potential Solution
Implements logic in the EWPCM to enforce portfolio bias.

#### Reproducing the Problem
N/A

#### Checklist

AlexCatarino:
#### Description
Implements `PortfolioBias` in EWPCM and IWPCM. With this new feature, these PCMs will ignore insights that do not respect the desired bias. E.g. for `PortfolioBias.Long`, on Insights with `InsightDirection.Up` will be converted into `PortfolioTarget.Quantity` greater than zero and other `InsightDirection` will result in `PortfolioTarget.Quantity` of zero.

#### Related Issue
Closes #3769 

#### Motivation and Context
Be able to create Portfolio Construction Models with Portfolio Bias.

Yes. Showcase this option.

New regression algorithm and unit tests.

#### Types of changes

#### Checklist:


Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.py
```diff
@@ -101,12 +116,17 @@ def CreateTargets(self, algorithm, insights):
         for symbol, g in groupby(activeInsights, lambda x: x.Symbol):
             lastActiveInsights.append(sorted(g, key = lambda x: x.GeneratedTimeUtc)[-1])
 
-        # Determine target percent for the given insights
-        percents = self.DetermineTargetPercent(lastActiveInsights)
+        # Determine target percent for the given insights that respect the portfolio bias
+        percents = self.DetermineTargetPercent(list(filter(self.RespectPortfolioBias, lastActiveInsights)))
```
Martin-Molinero: nit - just checking, do we need to call `list()` here?

Tests/Algorithm/Framework/Portfolio/LongOnlyInsightWeightingPortfolioConstructionModelTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Linq;
+
+namespace QuantConnect.Tests.Algorithm.Framework.Portfolio
+{
+    [TestFixture]
+    public class LongOnlyInsightWeightingPortfolioConstructionModelTests : InsightWeightingPortfolioConstructionModelTests
+    {
+        private const double _weight = 0.01;
+
+        //public override double? Weight => Algorithm.Securities.Count == 0 ? default(double) : 1d / Algorithm.Securities.Count;
```
Martin-Molinero: Love the unit tests 👍 

nit - commented line. A few of the using statements aren't required. `_weight` isn't required. Typo bellow `paramenter`. Same for `LongOnlyEqualWeightingPortfolioConstructionModelTests`. Suggest using ReSharper (or similar) to catch these nit cases before PR.

Tests/Algorithm/Framework/Portfolio/InsightWeightingPortfolioConstructionModelTests.cs
```diff
@@ -13,311 +13,54 @@
...
-            var insights = new[] { GetInsight(Symbols.SPY, InsightDirection.Up, _algorithm.UtcTime) };
-
-            foreach (var target in _algorithm.PortfolioConstruction.CreateTargets(_algorithm, insights))
+            var direction = InsightDirection.Down;
+            if (PortfolioBias != PortfolioBias.LongShort && (int)direction != (int)PortfolioBias)
             {
-                var holding = _algorithm.Portfolio[target.Symbol];
-                holding.SetHoldings(holding.Price, target.Quantity);
-                _algorithm.Portfolio.SetCash(_startingCash - holding.HoldingsValue);
+                direction = InsightDirection.Flat;
```
Martin-Molinero: Hm maybe could use `Direction.Up` when LongOnly bias, else this test won't really cover what its meant to? or actually even better directly use `Direction.Up` for all the cases.

Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.py
AlexCatarino: Yes. Python `[filter](https://docs.python.org/3/library/functions.html#filter)` returns an "iterator" that pythonnet doesn't convert into a C# List.

Tests/Algorithm/Framework/Portfolio/InsightWeightingPortfolioConstructionModelTests.cs
AlexCatarino: It is right. When the `PortfolioBias` and `InsightDirection` don't meet (`PortfolioBias.LongOnly` and `InsightDirection.Down`), the `InsightDirection` is `InsightDirection.Flat`. We are creating insights with `InsightDirection.Down` in that test.

Tests/Algorithm/Framework/Portfolio/EqualWeightingPortfolioConstructionModelTests.cs
```diff
@@ -44,7 +44,7 @@ public class EqualWeightingPortfolioConstructionModelTests
         public virtual PortfolioBias PortfolioBias => PortfolioBias.LongShort;
 
         [TestFixtureSetUp]
-        public void SetUp()
+        public virtual void SetUp()
```
Martin-Molinero: Wondering why you made this guy virtual, I think no body is overriding him?

Algorithm.Framework/Portfolio/InsightWeightingPortfolioConstructionModel.cs
```diff
@@ -54,7 +54,7 @@ public static InsightWeightingPortfolioConstructionModel LongOnly(TimeSpan timeS
         /// Creates a new instance of <see cref="InsightWeightingPortfolioConstructionModel"/> with Long portfolio bias.
         /// </summary>
         /// <param name="resolution">Rebalancing frequency</param>
-        public static InsightWeightingPortfolioConstructionModel LongOnly(Resolution resolution = Resolution.Daily)
+        public new static InsightWeightingPortfolioConstructionModel LongOnly(Resolution resolution = Resolution.Daily)
```
Martin-Molinero: Nice catch 👍

Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.cs
```diff
@@ -92,6 +124,16 @@ public virtual bool ShouldCreateTargetForInsight(Insight insight)
             return result;
         }
 
+        /// <summary>
+        /// Method that will determine if a given insight respects the portfolio bias
+        /// </summary>
+        /// <param name="insight">The insight to create a target for</param>
+        /// <returns>True if the insight respects the portfolio bias</returns>
+        public bool RespectPortfolioBias(Insight insight)
```
Martin-Molinero: Why make `RespectPortfolioBias` public and not protected?

jaredbroad:
Looking forward here we also have issue #https://github.com/QuantConnect/Lean/issues/3986.

So we have 2-3 parameters; rebalance frequency, minimum order size, and now portfolio bias.

I don't think we should use pseudo "extension methods" like this as the pattern gets messy and unpredictable. 

Perhaps we could consider a pattern of constructor args for all PCM's (like how all indicators are aligned)?

jaredbroad:
Other concepts to consider: Percent deviation from the equal weighting for PCM 

https://www.quantconnect.com/forum/discussion/7182/semi-constant-rebalanced-portfolio-scrp/p1
>absolute sum of current portfolio weights diverges from the absolute sum of given portfolio weights (at initialization) more than the pre-defined threshold e.g. 10%

MinimumOrderSize which is execution concern,
https://github.com/QuantConnect/Lean/issues/3986

Both should reduce churn in different ways.

jaredbroad:
And if we add to these models we should add to all of them simultaneously and keep the API constructors consistent between then all. Then likely it'll consolidate to the base class -- then it'll be just 3 lines of code in each consumer to reuse that base class code.

Algorithm.Framework/Portfolio/EqualWeightingPortfolioConstructionModel.cs
```diff
@@ -131,12 +173,17 @@ public override IEnumerable<IPortfolioTarget> CreateTargets(QCAlgorithm algorith
 
             var errorSymbols = new HashSet<Symbol>();
 
-            // Determine target percent for the given insights
-            var percents = DetermineTargetPercent(lastActiveInsights);
+            // Determine target percent for the given insights that respect the portfolio bias
+            var percents = DetermineTargetPercent(lastActiveInsights.Where(RespectPortfolioBias).ToList());
```
Martin-Molinero: Could we directly apply the `.Where(RespectPortfolioBias)` when creating `lastActiveInsights` instead?

Issue #2799, opened 2019-01-04
IB Brokerage occasionally fails to complete cash sync


StefanoRaggi:
#### Expected Behavior
The IB Brokerage should never fail to complete the daily cash synchronization.

#### Actual Behavior
In some cases (especially during weekends), a live algorithm deployed to IB is terminated with the following error message:
```
System.Exception: The maximum number of attempts for brokerage cash sync has been reached.
```

#### Potential Solution
The `InteractiveBrokersBrokerage.DownloadAccount()` method is currently assuming that the `updatePortfolio` IB callback is always being called after the account download, which will in turn call `Reset` on the `ManualResetEvent` instance.
According to the IB documentation [here](https://interactivebrokers.github.io/tws-api/interfaceIBApi_1_1EWrapper.html#a790ccbe25033df73996f36a79ce2ce5a) the `updatePortfolio` callback will only be called when positions change.

The solution is to add this line at the top of the `InteractiveBrokersBrokerage.DownloadAccount()` method:
```
            _accountHoldingsResetEvent.Reset();
```

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

bradhallett:
I have received this error this morning: "Runtime Error: The maximum number of attempts for brokerage cash sync has been reached.: StackTrace: System.Exception: The maximum number of attempts for brokerage cash sync has been reached." It appears that this issue is recurring. Can this be looked into further?

PR #3988, opened 2020-01-09
IB Brokerage bug fixes

StefanoRaggi:
#### Description
- Refactored some IB state variables into the new `InteractiveBrokersStateManager` class.
- Fixed a bug where one of the state variables was not being reset between IBGateway restarts.
- Recovery from a disconnect caused by the user logging into TWS remotely (and ignoring the TWS warning message) will no longer be attempted: the algorithm will be terminated immediately with the `"Existing session detected"` error message.

#### Related Issue
Ref. #2799 #3395 

#### Motivation and Context
- Occasional cash sync errors reported by a few users

No.

- Live testing in QC cloud

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
+++ b/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
@@ -3191,27 +3191,27 @@ namespace QuantConnect.Brokerages.InteractiveBrokers
         {
             if (_loginFailed)
             {
-                throw new Exception(
-                    "InteractiveBrokersBrokerage.CheckIbAutomaterErrors(): " +
-                    "Login failed. " +
-                    "Please check the validity of your login credentials.");
+                HandleIbAutomaterError("Login failed. Please check the validity of your login credentials.");
             }
 
             if (_existingSessionDetected)
             {
-                throw new Exception(
-                    "InteractiveBrokersBrokerage.CheckIbAutomaterErrors(): " +
-                    "An existing session was detected and will not be automatically disconnected. " +
-                    "Please close the existing session manually.");
+                HandleIbAutomaterError("An existing session was detected and will not be automatically disconnected. " +
+                                       "Please close the existing session manually.");
             }
 
             if (_securityDialogDetected)
             {
-                throw new Exception(
-                    "InteractiveBrokersBrokerage.CheckIbAutomaterErrors(): " +
-                    "A security dialog was detected for Second Factor/Code Card Authentication. " +
-                    "Please opt out of the Secure Login System: Manage Account > Security > Secure Login System > SLS Opt Out");
+                HandleIbAutomaterError("A security dialog was detected for Second Factor/Code Card Authentication. " +
+                                       "Please opt out of the Secure Login System: Manage Account > Security > Secure Login System > SLS Opt Out");
             }
+        }
+
+        private void HandleIbAutomaterError(string message)
+        {
+            OnMessage(new BrokerageMessageEvent(BrokerageMessageType.Error, "IBAutomater", message));
+
+            throw new Exception($"InteractiveBrokersBrokerage.CheckIbAutomaterErrors(): {message}");
         }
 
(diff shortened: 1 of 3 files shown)


Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -142,8 +140,6 @@ private enum Region { America, Europe, Asia }
         // IB requests made through the IB-API must be limited to a maximum of 50 messages/second
         private readonly RateGate _messagingRateLimiter = new RateGate(50, TimeSpan.FromSeconds(1));
 
-        // used for reconnection attempts
-        private bool _previouslyInResetTime;
         // used to limit logging
         private bool _isWithinScheduledServerResetTimesLastValue;
```
Martin-Molinero: Should we add this flag into the state manager too?

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -1353,8 +1360,11 @@ private void TryWaitForReconnect()
 
             if (!isResetTime)
             {
-                if (_previouslyInResetTime)
+                if (_stateManager.PreviouslyInResetTime)
                 {
+                    // do not restart after IBAutomater errors
+                    CheckIbAutomaterErrors();
```
Martin-Molinero: Maybe we could move this directly into the `ResetGatewayConnection()`, so we avoid having it spread and we don't have to rely on the caller to do the right thing.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
```diff
@@ -3126,14 +3136,14 @@ private void OnIbAutomaterOutputDataReceived(object sender, OutputDataReceivedEv
...
                 _ibAutomaterInitializeEvent.Set();
             }
 
-            // an existing session was detected and IBAutomater clicked the "Exit Application" button
-            else if (e.Data.Contains("Existing session detected") && !_performingRelogin)
+            // an existing session was detected
+            else if (e.Data.Contains("Existing session detected"))
             {
-                _existingSessionDetected = true;
+                _stateManager.ExistingSessionDetected = true;
```
Martin-Molinero: It seems `LoginFailed`, `ExistingSessionDetected` and `SecurityDialogDetected` all trigger termination. What do you think if we had a single error code flag? Which we set with defined enum values, with the objective of simplifying the state machine

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: I didn't add this one in initially because it was only used for logging purposes, but I agree it can be moved as well, will do.

Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs
StefanoRaggi: Hmm, the reason for having this check in the caller was to avoid throwing exceptions from the reset handler thread,
also, we have four calls to `CheckIbAutomaterErrors` where only two of them are before triggering the reset event.

Martin-Molinero: Right 👍  although the reset handler thread is already calling `CheckIbAutomaterErrors()` after restarting the ibAutomater.

StefanoRaggi: Correct, this should be addressed in the new version of `CheckIbAutomaterErrors()`, which now also sets the `algorithm.RuntimeError` via the `BrokerageMessageEvent`. The version in the current master only throws the exception which is logged in the reset handler thread, but the algorithm keeps running.

jaredbroad:
Approving on condition this will enjoy a refactor and tidy up soon...

PR #3993, opened 2020-01-13
Adding Support for Equity L1 Quote Data

### LINKED ISSUE #3589: Support Equity Quotes
#### Expected Behavior

Lean should be able to process Equity quotes. 

#### Actual Behavior

Lean can only backtest equity with trades.

#### Potential Solution

Modifications in code:
- Add `QuoteBars` for equities in `DefaultDataTypes`
https://github.com/QuantConnect/Lean/blob/b84263840b2d149604463c3d61d98bff7e697f36/Common/Data/SubscriptionManager.cs#L207

- If `QuoteBar` will be the default tick type for Equity, we should modify `GetCommonTickType`
https://github.com/QuantConnect/Lean/blob/b84263840b2d149604463c3d61d98bff7e697f36/Common/Util/LeanData.cs#L620

- Here only one subscription is added for equities, we need to add two subscriptions.
https://github.com/QuantConnect/Lean/blob/b84263840b2d149604463c3d61d98bff7e697f36/Algorithm/QCAlgorithm.Universe.cs#L454

Those modifications were included in [this branch](https://github.com/QuantConnect/Lean/tree/equity-taq-wip). It shows an expected behavior using a basic `QCAlogorithm` and a Framework algorithm. But it is a simple draft, new tests should be added, the regressions test expected values need to update. Also, we need to include new sample data.

#### Checklist

Jay-Jay-D:
#### Description

This PR implements the modifications needed in order to backtesting equities with Trades and Quotes (Quotes).

Sample data for equity was updated.

Many test were updated, new test were added.

~TODO: add low resolution data.~

#### Related Issue

Closes #3589 

#### Motivation and Context
.
Quotes ask and bid are more realistic prices to use as filling prices. 

It should update all pages in which equities are refereed as trades only.

A new regression test was added to confirm the fill prices are the expected.

Also given that the sample data for equities was updated, many test and regression test were updated/fixed.

#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1738,8 +1739,13 @@ public SubscriptionDataConfig GetSubscription(Symbol symbol, TickType? tickType)
...
                 var subscriptions = SubscriptionManager.Subscriptions.OrderBy(x => x.TickType);
 
+                if (tickType == null)
+                {
+                    tickType = LeanData.GetCommonTickType(symbol.SecurityType);
+                }
+
                 // find our subscription to this symbol
-                subscription = subscriptions.FirstOrDefault(x => x.Symbol == symbol && (tickType == null || tickType == x.TickType));
+                subscription = subscriptions.SingleOrDefault(x => x.Symbol == symbol && x.TickType == tickType);
```
Martin-Molinero: I'm not sure this should be Single? Can we guarantee there won't be other subscriptions for that symbol and tick type?
Wondering why change previous behavior if `tickType == null`?

Algorithm.CSharp/EquityTradeAndQuotesRegressionAlgorithm.cs
```diff
@@ -0,0 +1,126 @@
...
+            SetStartDate(2013, 10, 07);  //Set Start Date
+            SetEndDate(2013, 10, 11);    //Set End Date
+            SetCash(100000);             //Set Strategy Cash
+
+            
+            SetSecurityInitializer(x => x.SetDataNormalizationMode(DataNormalizationMode.Raw));
+
+            _symbol = AddEquity("IBM", Resolution.Minute).Symbol;
+
+            Schedule.On(DateRules.EveryDay(_symbol), TimeRules.AfterMarketOpen(_symbol, 0), () => { _canTrade = true; });
```
Martin-Molinero: Why is this `_canTrade` flag required?

Algorithm.CSharp/EquityTradeAndQuotesRegressionAlgorithm.cs
```diff
@@ -0,0 +1,126 @@
...
+                Log($"{Time:s} {orderEvent.Direction}");
+                var expectedFillPrice = orderEvent.Direction == OrderDirection.Buy ? Securities[_symbol].AskPrice : Securities[_symbol].BidPrice;
+                if (orderEvent.FillPrice != expectedFillPrice)
+                {
+                    throw new Exception($"Fill price is not the expected for OrderId {orderEvent.OrderId} at Algorithm Time {Time:s}." +
+                                        $"\n\tExpected fill price: {expectedFillPrice}, Actual fill price: {orderEvent.FillPrice}");
+                }
+            }
+        }
+
```
Martin-Molinero: Some more ideas for testing:
- Could also add an assert for the subscriptions in the `OnSecurityChanges` callback, make sure expected types are added correctly
- Could check that the data is actually being pumped into the algorithm at `OnData`
- Could make a 1 time history request during initialize and check the correct data is sent there too

Algorithm.CSharp/BasicTemplateCryptoAlgorithm.cs
```diff
@@ -203,25 +203,25 @@ public override void OnEndOfAlgorithm()
...
-            {"Profit-Loss Ratio", "0"},
-            {"Alpha", "-5.74"},
-            {"Beta", "1.931"},
-            {"Annual Standard Deviation", "0.254"},
-            {"Annual Variance", "0.065"},
-            {"Information Ratio", "-35.579"},
-            {"Tracking Error", "0.123"},
-            {"Treynor Ratio", "-1.494"},
-            {"Total Fees", "$85.33"}
+            {"Total Trades", "16"},
```
Martin-Molinero: We might need clarification on why the total trades count changed too?

Algorithm.Python/BasicTemplateAlgorithm.py
```diff
@@ -38,7 +38,7 @@ def Initialize(self):
         self.SetEndDate(2013,10,11)    #Set End Date
         self.SetCash(100000)           #Set Strategy Cash
         # Find more symbols here: http://quantconnect.com/data
-        self.AddEquity("SPY", Resolution.Second)
+        self.AddEquity("SPY", Resolution.Minute)
```
Martin-Molinero: Nice catch! 👍

Algorithm.CSharp/EquityTradeAndQuotesRegressionAlgorithm.cs
```diff
@@ -0,0 +1,126 @@
...
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
+        {
+            if (!Portfolio.Invested && _canTrade)
+            {
+                SetHoldings(_symbol, 1);
```
Martin-Molinero: This will test market orders, think we should test other order types too (maybe unit test)? Thinking about which price should the fill models use 🤔

Algorithm/QCAlgorithm.Indicators.cs
Jay-Jay-D: >I'm not sure this should be Single? Can we guarantee there won't be other subscriptions for that symbol and tick type?
To my best understand, the only case where we can have more than one subscription for a given symbol and tick type is if we add a symbol with multiple resolutions. 

But that case is not implemented in the actual master; there is no definition of which resolution it should return. 

Any thoughts?

> Wondering why change previous behavior if tickType == null?
After asking @AlexCatarino, he explained that in cases where the security has both tick types (crypto and soon equities),  the expected subscription should be the one with the _preferred_ tick type.

I added a test that fails in actual master for the crypto case. 
https://github.com/QuantConnect/Lean/blob/81f201e7dfa954f2081393551a8a73c8afcd0ee2/Tests/Algorithm/AlgorithmResolveConsolidatorTests.cs#L31

Algorithm.CSharp/EquityTradeAndQuotesRegressionAlgorithm.cs
Jay-Jay-D: Why is this _canTrade flag required?
The algorithm trades each 15 minutes, but if it sends an order at 16:00 it'll be an open market order an it fill price won't be the quote side **close**, but the the quote side **open**.

So to keep it simple, I don't allow it to send orders after 15:45.

`DataNormalizationMode.Raw` is just because it was easier to check prices directly from the file data :slightly_smiling_face:

Algorithm.CSharp/BasicTemplateCryptoAlgorithm.cs
Jay-Jay-D: The algorithm uses indicators to determine the trades.

In turn, previous to the changes in [Algorithm/QCAlgorithm.Indicators.cs ](https://github.com/QuantConnect/Lean/pull/3993/files/81f201e7dfa954f2081393551a8a73c8afcd0ee2#diff-95fe47a1b5da9a59bc5c9cf89a905ca0R1744), Crypto indicators were updated with trades, now the indicators are updated from quotes.

Jay-Jay-D:
Wondering why are we modifying trade.zip files shouldn't we be adding some quote.zip files instead?

Trades and quotes are generated from a single data source, different from the actual trades. That's why the trades are also updated.

Algorithm.CSharp/BasicTemplateCryptoAlgorithm.cs
jaredbroad: >now the indicators are updated from quotes.
I think crypto should also be updated with trades like equities. As we've seen quotes can vary so widely its almost random noise.  This would be reversing the default tick type right? It would also make it simpler for crypto users to create indicators?

Jay-Jay-D: > This would be reversing the default tick type right?

Exactly.

Algorithm/QCAlgorithm.Indicators.cs
Martin-Molinero: True, but using `SingleOrDefault` seems like a breaking change compared to `master`, since it will throw, I'd be inclined to respect current behavior and use `FirstOrDefault` instead.

Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
```diff
@@ -39,7 +39,7 @@ public void PerformsMarketFillBuy()
         {
             var model = new ImmediateFillModel();
             var order = new MarketOrder(Symbols.SPY, 100, Noon);
-            var config = CreateTradeBarConfig(Symbols.SPY);
+            var config = CreateQuoteBarConfig(Symbols.SPY);
```
Martin-Molinero: I don't think these unit tests are performing as expected, most of them are setting a `new TradeBar` into the security cache so there is no quote bar to use and uses the trade bar instead. Also, I think we need *new* tests for QuoteBar configurations but keeping existing tests too?

Algorithm/QCAlgorithm.Indicators.cs
Jay-Jay-D: Good point!

Common/Data/Market/Tick.cs
```diff
@@ -226,17 +227,47 @@ public Tick(SubscriptionDataConfig config, string line, DateTime date)
...
+                                Suspicious = (csv[index++] == "1");
+                            }
+                        }
+                        else if (TickType == TickType.Quote)
+                        {
+                            BidPrice = csv[index++].ToDecimal() / scaleFactor;
+                            BidSize = csv[index++].ToDecimal();
+                            AskPrice = csv[index++].ToDecimal() / scaleFactor;
+                            AskSize = csv[index++].ToDecimal();
+                            if (BidPrice == 0 || AskPrice == 0)
```
Jay-Jay-D: nit:
```cs
                                Value = BidPrice + AskPrice;
                                if (BidPrice * AskPrice == 0)
                                {
                                    Value /= 2;
                                }
```

Martin-Molinero: Nice 👍 
```cs
Value = BidPrice + AskPrice;
if (BidPrice * AskPrice != 0)
{
     Value /= 2;
}
```

PR #3996, opened 2020-01-14
Add HRW Wheat, Brent Crude, Low Sulfur Gasoil and Random Length Lumbe…

### LINKED ISSUE #3995: Add new commodity instruments to Lean
#### Expected Behavior
Be able to use the following within QuantConnect Lean:
- KC HRW Wheat (KE:CBOT)
- Random Length Lumber (LBS:CME)
- Brent (B:ICE)
- Low Sulfur Gasoil (G:ICE)

#### Actual Behavior
Futures contract chains cannot be used because no rules exist for calculating expiry dates

#### Potential Solution
Add the above contracts to Lean including the following:
- market-hours-database.json
- symbol-properties-database.csv
- QuantConnect.Securities.Futures (Futures.cs)
- FuturesExpiryFunctions.cs
- FuturesExpiryFunctionsTestData.xml
- FuturesExpiryFunctionsTests

#### Checklist

adam-may:
#### Description
Closes #3995 

#### Related Issue
#3995 

#### Motivation and Context
Addresses some missing instruments across CME, CBOT, ICE. Note that a second Wheat contract was introduced, so the naming of the current wheat contract enumeration was slightly changed (prefixed with SRW), but a backwards compatible "synonym" was put in to prevent the change from being breaking

None

I was able to subscribe to the new instruments using "AddFuture" calls within an algorithm

#### Types of changes

#### Checklist:


Common/Securities/Future/Futures.cs
```diff
@@ -39,10 +39,16 @@ public static class Grains
             public const string BlackSeaWheatFinanciallySettledPlatts = "BWF";
 
             /// <summary>
-            /// Wheat Futures
+            /// Chicago SRW Wheat Futures
             /// </summary>
             /// <returns>The symbol</returns>
-            public const string Wheat = "ZW";
+            public const string SRWWheat = "ZW";
```
adam-may: Are there any concerns with this? A second wheat contract was added, so needed to clarify the original one (it's a breaking change)

jaredbroad: Which has the most liquidity and would be the most referred to contract? I'm guessing Chicago's --> so Chicago can be the default, with the exchange specification copies being another 2 new properties:

```public const string Wheat => SRWWheat;
public const string SRWWheat = "ZW";
public const string HRWWheat = "KE";```

Comment (adam-may) [reply Common/Securities/Future/Futures.cs]: I had thought about that and wasn't 100% comfortable with making a breaking change. Will amend.

File: Data/market-hours/market-hours-database.json [hunk 2]
```diff
@@ -510,6 +510,140 @@
...
+          "state": "market"
+        },
+        {
+          "start": "08:30:00",
+          "end": "13:20:00",
+          "state": "market"
+        }
+      ],
+      "saturday": [],
+      "holidays": []
```
gsalaz98: Hey Adam, it's good to see you again

We'll be needing the holidays that the contracts you've added observe, as well as any early closes. Since each contract's trading schedule can differ drastically (even in the same commodity group), we add holidays for each contract, and some assets can have early closes (usually around thanksgiving time, or right before a holiday like Good Friday).

You can usually source these calendars from the CME or ICE websites.

### CME Instructions

For any CME based contract (CME, CBOT, NYMEX, COMEX), we can source historical holiday and early close data from
https://www.cmegroup.com/tools-information/holiday-calendar.html 

I'd recommend downloading the ZIP files listed at the bottom of the page, going back to the year the contract first started trading. 

Example: 2020 Calendar - https://www.cmegroup.com/tools-information/holiday-calendar/files/2020-holiday-calendars.zip

I would personally recommend to ignore the "compact" versions of the holiday calendar, as it usually isn't complete and does not cover specific exceptions that apply to certain contracts.

Please take note of the time zone the calendar is in. It will usually be in Central Time (UTC+6), but it could have potentially changed in the past. Also, you'll want to use the `Calendar Date` row to determine what day the hours listed apply to. 

You'll want to locate what group your contract belongs to. You can usually do this by visiting the contract's page on the CME website and looking at the URL. For example, for `KE`, this is the URL to the contract specs:

```
https://www.cmegroup.com/trading/agricultural/grain-and-oilseed/kc-wheat_contract_specifications.html
                                 ^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^
                                 (category)   (product type)
                                              ^^^^^^^^^^^^^^
                                              You'll want to use this section of the URL to determine
                                              what CME classifies the contract as. You will be able
                                              to find out if any exceptions apply to them in the 
                                              holiday calendar.
```

In the December 2020 calendar, we can see that there is an exception for `Grain and Oilseeds` under the `Grain, Oilseed & MGEX Products` category. 

Using this as an example, we see that on `Thursday, December 24` the contract stops trading at `12:05`. Since our `exchangeTimeZone` is set to America/Chicago (central time) in Market Hours Database, and the close is usually at `13:20:00`, we should add this under the `earlyCloses` property as:

```
"earlyCloses": {
        "12/24/2020": "12:05:00"
}
```

Next, looking at `Friday, December 25`, we see the message `Globex Closed` - Mark these as holidays under the `holidays` property (M/D/YYYY).

On `Sunday, December 27`, we should check if there is a time beneath the `Halt` column that applies to our contract. In this case, the entry is empty, so we can skip it. If it isn't empty, it usually signals an early close, and it must be added to the `earlyCloses` property.

On `Monday, December 28`, our `Close` time is equal to the closing time we have for "friday" in the Market Hours Database already, so we can skip it.

A few additional notes on CME calendars:

* Sometimes, there is a big arrow pointing to another row. These arrows indicate that the status it starts from extends until the field it is pointing to
* `Market Closed` status is equivalent to `Globex Closed`, and should also be counted as a holiday.
* Thanksgiving has a special field called "Early Close" that will contain early closing times before or after thanksgiving.
* Before 2011-2012, the calendars are sometimes not provided in spreadsheet format.
* If you don't find a category for the contract you are wanting to add, find the category that it applies to, and see if it contains entries for the entire category.

### ICE Instructions
You can find the holiday schedules (early closes) and holiday calendar for all ICE contracts on this page: https://www.theice.com/futures-us/notices

To source entries for the `holiday` property, you'll want to look at the `Holiday Calendar` notices. 
To source entries for the `earlyCloses` property, you'll want to look at the `Holiday Schedule` or `Holiday Trading Schedule` notices.

**Examples:**
You can find the 2020 holiday calendar here: https://www.theice.com/publicdocs/futures_us/exchange_notices/2020_Trading_Holiday_Calendar_20190716.pdf

For MLK day 2020, you can find the holiday schedule here: https://www.theice.com/publicdocs/futures_us/exchange_notices/ICE_Futures_US_2020MLKDay_Holiday_20191114.pdf

--------

Please let me know if you have any questions, and I'll do my best to answer them.

adam-may: No problems about adding holidays and early closes. I saw that a lot of the others hadn't been populated so wasn't 100% sure on what was standard. How much forward/back data do you need/expect?

I've had experience with the trading calendars and where they get added to the system when I was doing some other work (the international futures, so shouldn't be a big issue.

gsalaz98: For back data, we usually try to get as much as we can until the asset didn't exist, or the data ran out. In the case of CME, they only provide holiday calendar and schedules going back to 2009, so 2009 would be the natural stopping point for many contracts that have been trading for a while now.

For forward data, we can extrapolate for some dates like Christmas and New Years, but sometimes there isn't enough data/the schedule hasn't been established, so it's okay if it only reaches to the end of this year.

For forward data as well, in FuturesExpiryTestFunctionsTestData.xml, there are some cases, especially for Middle Eastern/Singapore/Malaysian based contracts, that there are some hidden holidays embedded into the expiry schedule. A prime example of this are contracts with countries that celebrate the holiday "Eid al-Adha" or observe U.K. Bank Holidays.

Finally, taking a look at the "Future-usa-ADB" entry in MHDB, you can see we've added additional entries designating this year's "Eid al-Adha" since additional test cases sourced from CME revealed the fact that they observed this holiday.

https://github.com/QuantConnect/Lean/blob/master/Data/market-hours/market-hours-database.json#L7331

https://www.cmegroup.com/trading/energy/crude-oil/brent-dubai-swap-futures_contract_specifications.html

Data/market-hours/market-hours-database.json
```diff
@@ -601,9 +601,330 @@
...
+        "07/03/2020",
+        "09/06/2020",
+        "11/26/2020",
+        "12/25/2020",
+        "01/01/2021"
+      ],
+      "earlyCloses": {
+        "03/28/2013": "14:00:00",
+        "07/03/2013": "12:00:00",
+        "11/27/2013": "13:20:00",
```
gsalaz98: I'm seeing this as `13:15:00` on my end; maybe we're using different data sources? I've attached the spreadsheet I'm sourcing data from in this comment:

https://www.dropbox.com/s/ujzj77vc99qdiqt/2013-thanksgiving.xls?dl=0

Common/Securities/Future/Futures.cs
```diff
@@ -758,6 +782,17 @@ public static class Indices
             public const string USDDenominatedIbovespa = "IBV";
         }
 
+        /// <summary>
+        /// Lumber and Pulp group
```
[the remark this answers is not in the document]
adam-may: Yep, true... Resharper's refactor->rename won't pick that up. My bad. ;)

Common/SymbolRepresentation.cs
```diff
@@ -145,18 +145,12 @@ public static string GenerateFutureTicker(string underlying, DateTime expiration
...
-                    month = 1;
-                    year++;
-                }
-            }
+            month += FuturesExpiryUtilityFunctions.ExpiresInPreviousMonth(underlying);
+            // Get the month back into the allowable range, allowing for a wrap 
+            month--;
+            year += month / 12;
+            month %= 12;
+            month++;
```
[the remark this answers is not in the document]
adam-may: Yep - fair call. I'll add some commentary in

Data/market-hours/market-hours-database.json
```diff
@@ -601,9 +601,330 @@
         }
       ],
       "saturday": [],
-      "holidays": []
+      "holidays": [
+        "01/01/2013",
+        "01/20/2013",
+        "01/21/2013",
```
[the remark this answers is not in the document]
adam-may: Yep

Data/market-hours/market-hours-database.json
```diff
@@ -601,9 +601,330 @@
...
+        "12/31/2019": "13:20:00",
+        "04/09/2020": "13:20:00",
+        "07/02/2020": "12:05:00",
+        "11/25/2020": "13:20:00",
+        "12/24/2020": "12:05:00",
+        "12/31/2020": "13:20:00"
+      },
+      "lateOpens": {
+        "01/02/2013": "09:30:00",
+        "01/22/2013": "19:00:00",
```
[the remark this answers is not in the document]
adam-may: Yep

Data/market-hours/market-hours-database.json
```diff
@@ -601,9 +601,330 @@
         }
       ],
       "saturday": [],
-      "holidays": []
+      "holidays": [
+        "01/01/2013",
+        "01/20/2013",
+        "01/21/2013",
+        "02/17/2013",
+        "02/18/2013",
```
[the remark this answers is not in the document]
adam-may: Yep

Data/market-hours/market-hours-database.json
```diff
@@ -601,9 +601,330 @@
...
+        "05/24/2020",
+        "07/03/2020",
+        "09/06/2020",
+        "11/26/2020",
+        "12/25/2020",
+        "01/01/2021"
+      ],
+      "earlyCloses": {
+        "03/28/2013": "14:00:00",
+        "07/03/2013": "12:00:00",
```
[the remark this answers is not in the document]
adam-may: Yep

Data/market-hours/market-hours-database.json
adam-may: Yep, using the same spreadsheet, just that some of the data was hard to read with the overarching column headings originally, but it's much clearer now I've gone through the following year's spreadsheets, which are much clearer!

adam-may:
Code review comments addressed. For the MH changes, I think I just wasn't reading that format of the spreadsheet quite correctly, but on second reading (once I had read the subsequent years' spreadsheets), everything is much clearer now (i.e. the patterns in the hours)

PR #4003, opened 2020-01-14
Fix Beta calculation for non-daily algorithms

### LINKED ISSUE #3927: Beta Calculation Incorrect

gsalaz98:
# ⚠️ Warning ⚠️ 

This commit breaks accurate calculations for algorithms that
only make use of `Daily` resolution data. Previously, because
the benchmark was added in Daily resolution in backtesting, any
algorithm that only made use of daily data would have an accurate
calculation for beta and various other statistics.

These changes serve to fix the statistics calculations of non-daily
resolution algorithms, with daily resolution to be revisited at a later
time.

Regression algorithm statistics and a few BacktestNodePacket test statistics have not been updated. They'll be updated once this PR is approved.

Also please note that a few tests should be failing. They are:

* Most regression tests
* QuantConnect.Tests.Common.Packets/BacktestNodePacketTests/ClearsOtherCashAmounts
* QuantConnect.Tests.Common.Packets/BacktestNodePacketTests/InitialCashAmountIsRespected
* QuantConnect.Tests.Common.Packets/BacktestNodePacketTests/JobDatesAreRespected
* QuantConnect.Tests.Engine.Results/BacktestingResultHandlerTests/BasicTemplateDailyAlgorithmSamplesNotMisalignedAbsolute
* QuantConnect.Tests.Engine.Results/BacktestingResultHandlerTests/ResolutionSwitchingAlgorithmSamplesNotMisalignedAbsolute
* QuantConnect.Tests.Engine.Results/BacktestingResultHandlerTests/SamplesNotMisalignedRelative("BasicTemplateDailyAlgorithm")
* QuantConnect.Tests.Engine.Results/BacktestingResultHandlerTests/SamplesNotMisalignedRelative("ResolutionSwitchingAlgorithm")

Any additional failing tests are most likely a new bug.
#### Description

 Refactors sampling strategy in AlgorithmManager loop

* Refactors AlgorithmManager loop
* Refactors StatisticsBuilder methods and strategy for series alignment
* Move sampling logic to the corresponding IResultHandler
* Changes benchmark resolution to Resolution.Hour
* Modifies IResultHandler to enable external sampling
* Adds BacktestResultHandler unit tests
* Adds ResolutionSwitchingAlgorithm to test misalignment
* Adds support to AlgorithmRunner to store algorithm IResultHandler
* Removed Sample[a-zA-Z]+ methods from IResultHandler definition
* Converted Sample[a-zA-Z]+ methods from public to protected
* Updated inheritors of BaseResultHandler to use new accessibility
modifiers
* Removes useless code in ResolutionSwitchingAlgorithm

#### Related Issue

Closes #3927 
#### Motivation and Context

Provide accurate statistics to end users. Cleans up AlgorithmManager code.

Encourage users to not use daily resolution if possible

Unit tests, new unit tests
#### Types of changes

#### Checklist:


Engine/AlgorithmManager.cs
```diff
@@ -728,10 +705,8 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
             SetStatus(AlgorithmStatus.Completed);
 
             //Take final samples:
-            results.SampleRange(algorithm.GetChartUpdates());
-            results.SampleEquity(_previousTime, Math.Round(algorithm.Portfolio.TotalPortfolioValue, 4));
-            SampleBenchmark(results, backtestMode ? _previousTime.Date : _previousTime);
-            SamplePerformance(results, _previousTime.Date, force:true);
+            results.Sample(previousTime, force: true);
```
Martin-Molinero: Why use `previousTime` here? Since we are outside the streaming loop at this stage, wouldn't both be equals? Maybe we can join `time` in the loop with this `previousTime` outside the loop, into a single global `time`?

Engine/Results/BacktestingResultHandler.cs
```diff
@@ -54,6 +55,9 @@ public class BacktestingResultHandler : BaseResultsHandler, IResultHandler
         private bool _processingFinalPacket;
         private readonly HashSet<string> _chartSeriesExceededDataPoints;
 
+        protected decimal _closingPortfolioValue;
+        protected DateTime _previousTime;
```
Martin-Molinero: We could move these into the `BaseResultsHandler`? Maybe to be more precise, could rename to `_previousUtcSampleTime`/`_dailyPortfolioValue`

nit - `using QuantConnect.Lean.Engine.DataFeeds;` not required

Engine/Results/LiveTradingResultHandler.cs
```diff
@@ -1062,6 +1066,43 @@ protected virtual string CreateSafeChartName(string chartName)
             return Uri.EscapeDataString(chartName);
         }
 
+        /// <summary>
+        /// Samples benchmark continuously, and on date change, samples performance and portfolio equity
+        /// </summary>
+        /// <param name="time">Current time step of the AlgorithmManager loop</param>
+        /// <param name="force">Forces processing of equity and performance if true</param>
+        public virtual void Sample(DateTime time, bool force = false)
```
Martin-Molinero: Seems both live and backtesting implementations are very similar, can we join them and place it into the `BaseResultsHandler`? I think the difference is that Live is sampling the `Benchmark` on each call?

Common/Statistics/StatisticsBuilder.cs
```diff
@@ -265,64 +269,45 @@ private static IEnumerable<PeriodRange> GetPeriodRanges(int periodMonths, DateTi
...
-                listPerformance.Add(0);
-                Log.Trace("StatisticsBuilder.EnsureSameLength(): Padded Performance");
-            }
-            while (listPerformance.Count > listBenchmark.Count)
-            {
-                listBenchmark.Add(0);
-                Log.Trace("StatisticsBuilder.EnsureSameLength(): Padded Benchmark");
-            }
+            // GroupBy(...) is guaranteed to preserve the order the elements are in.
+            // See http://msdn.microsoft.com/en-us/library/bb534501 for more information.
```
Martin-Molinero: Nice 👍

Common/Statistics/StatisticsBuilder.cs
```diff
@@ -93,23 +93,28 @@ public static class StatisticsBuilder
...
-            // so we add missing dates with zero value
-            var missingPerformanceDates = benchmark.Keys.Where(x => !performance.ContainsKey(x));
-            foreach (var date in missingPerformanceDates)
+            // In very rare circumstances, we might have multiple entries for a single day in backtesting.
+            // These multiple entries will all be located at the end of the `pointsBenchmark` and `pointsPerformance`
+            // collections, since we force sample at the end of the algorithm.
+            // For good measure and to put any alignment issues to rest, let's resample both collections
+            // to daily resolution just in case.
+            var benchmark = ResampleDaily(ChartPointToDictionary(pointsBenchmark, fromDate, toDate));
+            var performance = ResampleDaily(ChartPointToDictionary(pointsPerformance, fromDate, toDate));
```
Martin-Molinero: Nice 👍

Tests/Engine/Results/BacktestingResultHandlerTests.cs
```diff
@@ -0,0 +1,391 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Tests.Engine.Results
+{
+    [TestFixture]
+    public class BacktestingResultHandlerTests
+    {
+        private Dictionary<string, BacktestingResultHandler> _resultsCache = new Dictionary<string, BacktestingResultHandler>();
```
Martin-Molinero: If we are just consuming the `Charts`, I'd suggest for us to just store them instead, if we need more in the future we can add it too

Tests/AlgorithmRunner.cs
```diff
@@ -44,6 +44,12 @@ namespace QuantConnect.Tests
     /// </summary>
     public static class AlgorithmRunner
     {
+        /// <summary>
+        /// Dictionary used to store algorithm <see cref="BacktestingResultHandler"/> instances keyed by the algorithm's <see cref="Language"/>
+        /// </summary>
+        public static Dictionary<Language, Dictionary<string, BacktestingResultHandler>> AlgorithmResults = new Dictionary<Language, Dictionary<string, BacktestingResultHandler>>();
```
Martin-Molinero: I think this isn't required? Since `BacktestingResultHandlerTests` has his own cache? Don't think we can really trust a global cache for results because for example start and end dates could change between executions

Tests/Engine/Results/BacktestingResultHandlerTests.cs
```diff
@@ -0,0 +1,391 @@
...
+
+        [TestCase(nameof(BasicTemplateAlgorithm))]
+        [TestCase(nameof(BasicTemplateDailyAlgorithm))]
+        [TestCase(nameof(ResolutionSwitchingAlgorithm))]
+        public void SamplesNotMisalignedRelative(string algorithm)
+        {
+            var backtestResults = GetResults(algorithm, new DateTime(2013, 10, 7), new DateTime(2013, 10, 11));
+            var benchmarkSeries = backtestResults.Charts["Benchmark"].Series["Benchmark"];
+            var equitySeries = backtestResults.Charts["Strategy Equity"].Series["Equity"];
+            var performanceSeries = backtestResults.Charts["Strategy Equity"].Series["Daily Performance"];
```
Martin-Molinero: I think we are also missing tests that go through the code paths in `StatisticsBuilder`, where part of the changes took place? Maybe simpler unit tests, just for the `StatisticsBuilder`, where for example we can assert it throws if given "misaligned values", it resamples daily

Tests/AlgorithmRunner.cs
gsalaz98: We use it to extract the BacktestingResultHandler from the AlgorithmRunner and load it into the BacktestingResultHandlerTests. It will overwrite any existing value (if there is one). The only caching we do is localized within the result handler test fixture.

Engine/Results/LiveTradingResultHandler.cs
gsalaz98: Personally, I think it's best that each `Sample` method is implemented separately since they have their own concerns with how they're sampling. Mixing these concerns might make us reliant on specific base class behavior. 

I'll talk to you more about this, would like to know what your thoughts are 👍

gsalaz98: Update: we looked into addressing this review, but we concluded that it would require additional effort outside the scope of this PR.

jaredbroad:
Closing as tests still not passing. Please open when ready for review.

gsalaz98:
I've gone ahead and changed the existing tests so that they are passing. 

There were a few that couldn't be modified due to the nature of the test, but have been ignored with a message linking to this PR. They are:
* `SamplesNotMisalignedRelative("BasicTemplateDailyAlgorithm")`
* `SamplesNotMisalignedRelative("ResolutionSwitchingAlgorithm")`

I've also addressed Martin's review and updated regression test statistics.

Tests/Engine/Results/BacktestingResultHandlerTests.cs
```diff
@@ -27,19 +28,21 @@ namespace QuantConnect.Tests.Engine.Results
     [TestFixture]
     public class BacktestingResultHandlerTests
     {
-        private Dictionary<string, BacktestingResultHandler> _resultsCache = new Dictionary<string, BacktestingResultHandler>();
+        private object _lock = new object();
```
Martin-Molinero: Why is this lock required?

Engine/Results/BaseResultsHandler.cs
```diff
@@ -106,6 +106,16 @@ public abstract class BaseResultsHandler
...
 
+        /// <summary>
+        /// Closing portfolio value. Used to calculate daily performance.
+        /// </summary>
+        protected decimal _dailyPortfolioValue;
+
+        /// <summary>
+        /// Last time the <see cref="IResultHandler.Sample(DateTime, bool)"/> method was called in UTC
+        /// </summary>
+        protected DateTime _previousUtcSampleTime;
```
Martin-Molinero: nit - since they are now `protected` they should be renamed to `DailyPortfolioValue` and `PreviousUtcSampleTime`

Tests/Common/Statistics/StatisticsBuilderTests.cs
```diff
@@ -0,0 +1,130 @@
...
+
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using QuantConnect.Statistics;
+
+namespace QuantConnect.Tests.Common.Statistics
+{
+    [TestFixture]
+    public class StatisticsBuilderTests
```
Martin-Molinero: Nice 👍

Tests/Engine/Results/BacktestingResultHandlerTests.cs
```diff
@@ -0,0 +1,396 @@
...
+                var results = AlgorithmRunner.AlgorithmResults[Language.CSharp][algorithm];
+                AlgorithmRunner.AlgorithmResults[Language.CSharp].Remove(algorithm);
+
+                return results;
+            }
+        }
+
+        [TestCase(nameof(BasicTemplateAlgorithm))]
+        [TestCase(nameof(BasicTemplateDailyAlgorithm), Ignore=true, IgnoreReason="Failing. Please see PR for more information: https://github.com/QuantConnect/Lean/pull/4003")]
+        [TestCase(nameof(ResolutionSwitchingAlgorithm), Ignore=true, IgnoreReason="Failing. Please see PR for more information: https://github.com/QuantConnect/Lean/pull/4003")]
```
Martin-Molinero: I think these tests shouldn't be ignored?

Tests/Engine/Results/BacktestingResultHandlerTests.cs
```diff
@@ -0,0 +1,396 @@
...
+
+            Assert.AreEqual(new DateTime(2013, 10, 7), performance.FirstKey().Date);
+            Assert.AreEqual(new DateTime(2013, 10, 11), performance.LastKey().Date);
+            Assert.AreEqual(5, performance.ValueCount);
+            Assert.AreEqual(5, performance.KeyCount);
+            Assert.AreEqual(0.0, performance.GetAt(0));
+            Assert.AreEqual(0.0, performance.GetAt(1));
+            Assert.AreEqual(Math.Round(-0.01148770, 6), Math.Round(performance.GetAt(2), 6));
+            Assert.AreEqual(Math.Round(0.011161310, 6), Math.Round(performance.GetAt(3), 6));
+            // Unknown value should go here.
```
Martin-Molinero: nit - this comment seems off?

Tests/AlgorithmRunner.cs
gsalaz98: I don't think so. We still need a way to get the BacktestingResultHandler out of the AlgorithmRunner call.

Martin-Molinero: Maybe we can create a custom class which holds the `LeanEngineAlgorithmHandlers`, the `AlgorithmManager` and return that instead 🤔

Engine/DataFeeds/UniverseSelection.cs
```diff
@@ -381,7 +381,7 @@ public bool AddPendingInternalDataFeeds(DateTime utcStart)
 
                     var dataConfig = _algorithm.SubscriptionManager.SubscriptionDataConfigService.Add(
                         securityBenchmark.Security.Symbol,
-                        _algorithm.LiveMode ? Resolution.Minute : Resolution.Daily,
+                        _algorithm.LiveMode ? Resolution.Minute : Resolution.Hour,
```
jaredbroad: 127 files changed due to `+2 char; -2 char`....

jaredbroad:
Congrats @Martin-Molinero @gsalaz98

PR #4051, opened 2020-01-31
MapFileResolver returns null if asked for delisted tickers

### LINKED ISSUE #4049: MapResolver should return null if we ask for a delisted ticker
#### Expected Behavior

`MapResolver.ResolveMapFile(ticker, date)` should return `null ` if we ask for a ticker delisted before `date`

#### Actual Behavior

It returns the last `MapFile` with the ticker

#### Potential Solution

Check the `MapResolver.ResolveMapFile` date logic.

#### Reproducing the Problem

This test should pass, but it fails in current `master`.
```cs
[Test]
public void ResolvesDelistedSymbolAsNull()
{
	var mapResolver = new MapFileResolver(new List<MapFile>
	{
		new MapFile(
			"amz",
			new List<MapFileRow>
			{
				new MapFileRow(new DateTime(1998, 01, 02), "uwz"),
				new MapFileRow(new DateTime(1998, 09, 11), "uwz"),
				new MapFileRow(new DateTime(2004, 12, 31), "amz")
			}
		)
	});
	var date = new DateTime(2007, 01, 03);
	var ticker = "AMZ";
	Assert.IsNull(mapResolver.ResolveMapFile(ticker, date));
}
```

#### System Information

Windows

#### Checklist

Jay-Jay-D:
#### Description

Update dates logic in `MapFileResolver.ResolveMapFile(symbol, date)` for cases where asked `date` is posterior to delisting date.

Also, check for `null `at `SecurityIdentifier.GetFirstTickerAndDate`. 

#### Related Issue

Closes #4049 

#### Motivation and Context

If we ask for a ticker for a date posterior to its delisting, `MapResolver` shouldn't return any map at all.

No.

New test was added, all tests are green.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Auxiliary/MapFileResolver.cs
+++ b/Common/Data/Auxiliary/MapFileResolver.cs
@@ -148,6 +148,7 @@ namespace QuantConnect.Data.Auxiliary
                     {
                         // the searched date is greater than the last date in the list, return the last entry
                         indexOf = entries.Keys.Count - 1;
+                        if (indexOf == 0) return null;
                     }
                     else
                     {
--- a/Common/SecurityIdentifier.cs
+++ b/Common/SecurityIdentifier.cs
@@ -541,11 +541,13 @@ namespace QuantConnect
         {
             var resolver = mapFileProvider.Get(market);
             var mapFile = resolver.ResolveMapFile(tickerToday, mappingResolveDate ?? DateTime.Today);
-
+            
             // if we have mapping data, use the first ticker/date from there, otherwise use provided ticker and DefaultDate
-            return mapFile.Any()
-                ? Tuple.Create(mapFile.FirstTicker, mapFile.FirstDate)
-                : Tuple.Create(tickerToday, DefaultDate);
+            if (mapFile != null && mapFile.Any())
+            {
+                return Tuple.Create(mapFile.FirstTicker, mapFile.FirstDate);
+            }
+            return Tuple.Create(tickerToday, DefaultDate);
         }
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


Common/Data/Auxiliary/MapFileResolver.cs
```diff
@@ -148,6 +148,7 @@ public MapFile ResolveMapFile(string symbol, DateTime date)
                     {
                         // the searched date is greater than the last date in the list, return the last entry
                         indexOf = entries.Keys.Count - 1;
+                        if (indexOf == 0) return null;
```
Martin-Molinero: Not sure this is right 🤔 - For the added unit test, add for example `new MapFileRow(new DateTime(2005, 12, 31), "amz")` and should still return `null`, but it doesn't, right?
I think it should directly return `null` in this `if` case, see comment on line 149

Jay-Jay-D: > add for example new MapFileRow(new DateTime(2005, 12, 31), "amz") and should still return

Rows in map files with duplicated tickers only makes sense in the first row (listing date) or in the last row (deslisting date or if it is still listed, 2050-12-31).

Adding a new row duplicating `AMZ` wouldn't be a correct test case.

Martin-Molinero: Agreed to directly return `null` on slack, pending change

Jay-Jay-D:
The actual behavior, i.e. return the latest available map file even when delisted, was intended.

It was explicitly tested here:
https://github.com/QuantConnect/Lean/blob/835c99e55c2ff8ca41d694c608c56e336767d94c/Tests/Engine/DataFeeds/Auxiliary/MapFileResolverTests.cs#L134

Also, `algorithm.AddEquity("AAA", Resolution.Daily);` uses today as date for resolving the map file.

PR #4054, opened 2020-01-31
Refactors Trading Economics Calendar Events Storage

gsalaz98:
Master performance (Release):

```
Debug: Algorithm Id:(TradingEconomicsAlgorithm) completed in 5.86 seconds at 199k data points per second. Processing total of 1,165,396 data points.
Debug: Algorithm Id:(TradingEconomicsAlgorithm) completed in 5.63 seconds at 207k data points per second. Processing total of 1,165,396 data points.
Debug: Algorithm Id:(TradingEconomicsAlgorithm) completed in 5.84 seconds at 200k data points per second. Processing total of 1,165,396 data points.
```

PR performance (Release):

```
Debug: Algorithm Id:(TradingEconomicsAlgorithm) completed in 6.04 seconds at 193k data points per second. Processing total of 1,166,612 data points.
Debug: Algorithm Id:(TradingEconomicsAlgorithm) completed in 5.84 seconds at 200k data points per second. Processing total of 1,166,612 data points.
Debug: Algorithm Id:(TradingEconomicsAlgorithm) completed in 5.64 seconds at 207k data points per second. Processing total of 1,166,612 data points.
```
#### Description

* Adds support for live algorithms using TE calendar events
* Modifies TradingEconomics tickers to include country
* Fixes bug where `TECal.Clone(...)` could result in non-deterministic
results
* Fixes bugs in TradingEconomicsEventFilter
* Adds new CSV parsing methods in StreamReaderExtensions
* Moves ToolBox/TradingEconomicsCalendarTests to Common tests folder
* Updates TradingEconomicsCalendar tests to test CSV parsing
* TradingEconomicsCalendarDownloader now writes files to disk as JSON
* Removes requirement of having an "actual" field for calendar events

#### Related Issue

N/A

#### Motivation and Context

Data integrity

No

Processed all available data successfully, new unit tests, modified unit tests, local backtesting
#### Types of changes

    * Potentially breaking change for current users of TE calendar data

#### Checklist:


Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -242,6 +335,37 @@ public override string ToString()
...
+                EndTime.ToStringInvariant("yyyyMMdd HH:mm:ss"),
+                LastUpdate.ToStringInvariant("yyyyMMdd HH:mm:ss"),
+                Ticker,
+                Actual.ToStringInvariant(),
+                CalendarId,
+                Category,
+                Country,
+                (int)DataType,
+                DateSpan,
+                $"\"{EventRaw}\"",
```
Martin-Molinero: Could we skip re-adding the double quotes here? Same bellow

Tests/Common/Data/Custom/TradingEconomicsTests.cs
```diff
@@ -72,10 +75,13 @@ public void CalendarFilterAppliedCorrectly(string eventName, string expected)
         public void DeserializesProperly()
         {
             var instance = new TradingEconomicsCalendar();
+            var csvBytes = Encoding.UTF8.GetBytes(JsonConvert.DeserializeObject<List<TradingEconomicsCalendar>>(TestCalendarJson).First().ToCsv());
+            var calendarStream = new StreamReader(new MemoryStream(csvBytes));
```
Martin-Molinero: I think we should add a _new_ test for the new StreamReader Reader implementation and keep a test for the other existing reader too, maybe even assert the generate the same result

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -250,6 +374,137 @@ public override DateTimeZone DataTimeZone()
...
+        }
+
+        /// <summary>
+        /// Converts country name to currency code (ISO 4217)
+        /// </summary>
+        /// <param name="country">Country name</param>
+        /// <returns>ISO 4217 currency code</returns>
+        public static string CountryToCurrencyCode(string country)
+        {
+            var regions = CultureInfo.GetCultures(CultureTypes.SpecificCultures).Select(x => new RegionInfo(x.LCID));
```
Martin-Molinero: It would be nice to have some test for `CountryToCurrencyCode` too. We are performing a `Contains` operation which is case sensitive, should it be insensitive instead? `FirstOrDefault` could return null, `ISOCurrencySymbol` will cause null ref exception. _Maybe_ we would want to keep regions of line 505 in a static collection for performance 🤔

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
```diff
@@ -180,52 +115,61 @@ private bool ProcessData(List<TradingEconomicsCalendar> data)
             var status = true;
 
             Parallel.ForEach(
-                data.GroupBy(x => GetTicker(x.Ticker, x.Category, x.Country)),
+                data.GroupBy(x => x.Country.ToLowerInvariant()),
```
Martin-Molinero: Is `x.Country` not normalized that we requrie `.ToLowerInvariant()`? Think it should be normalized/always the same

Common/Data/Custom/TradingEconomics/TradingEconomicsEventFilter.cs
```diff
@@ -191,7 +191,7 @@ public static string FilterEvent(string eventName)
                 filteredName = SingleWordFilter(filteredName, kvp.Key, kvp.Value);
             }
 
-            return filteredName;
+            return filteredName.Trim();
```
Martin-Molinero: I think it seems we need at least two new tests for `FilterEvent`: for this line and the change bellow

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -169,33 +199,96 @@ public string Event
...
-            var source = Path.Combine(Globals.DataFolder, "alternative", "trading-economics", "calendar", symbol, Invariant($"{date:yyyyMMdd}.zip"));
-            return new SubscriptionDataSource(source, SubscriptionTransportMedium.LocalFile, FileFormat.Collection);
+            var country = config.Symbol.Value.ToLowerInvariant()
+                .Split(new[] { "//" }, StringSplitOptions.None)
+                .First();
+
+            if (isLiveMode)
+            {
+                var transportMedium = _useStreaming ? SubscriptionTransportMedium.Streaming : SubscriptionTransportMedium.RemoteFile;
+                var liveSource = $"{_baseUrl}/calendar/country/{country.Replace('-', ' ')}/{date:yyyy-MM-dd}/{date.AddDays(1):yyyy-MM-dd}?c={_apiKey}&format=json";
```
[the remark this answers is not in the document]
jaredbroad: Ridiculously they require spaces in the URL. %20 would work fine too. `Replace("-", "%20")`

Martin-Molinero:
Thanks for the performance tests! 👍  Might be a good idea to increase the backtest length so that any noise is smaller in contrast

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -32,6 +39,15 @@ public class TradingEconomicsCalendar : BaseData
...
         private string _filteredEvent;
+        private string _originalEvent;
+        private const string _baseUrl = "https://api.tradingeconomics.com";
+        private static bool _useStreaming = Config.GetBool("trading-economics-use-streaming");
+        private static string _apiKey = Config.Get("trading-economics-api-key");
+
+        /// <summary>
+        /// Delimiter used to separate country from ticker in <see cref="TradingEconomics.Calendar"/> entries
+        /// </summary>
+        public const string Delimiter = "//";
```
Jay-Jay-D: Is there a reason for choosing this delimiter?

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -169,33 +199,96 @@ public string Event
         /// <returns>Subscription Data Source.</returns>
         public override SubscriptionDataSource GetSource(SubscriptionDataConfig config, DateTime date, bool isLiveMode)
         {
-            var symbol = config.Symbol.Value.ToLowerInvariant();
-            var source = Path.Combine(Globals.DataFolder, "alternative", "trading-economics", "calendar", symbol, Invariant($"{date:yyyyMMdd}.zip"));
-            return new SubscriptionDataSource(source, SubscriptionTransportMedium.LocalFile, FileFormat.Collection);
+            var country = config.Symbol.Value.ToLowerInvariant()
+                .Split(new[] { "//" }, StringSplitOptions.None)
```
Jay-Jay-D: Here you can use the `Delimiter` field, there are other cases like this one.

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
```diff
@@ -242,6 +335,37 @@ public override string ToString()
             return Invariant($"{symbol} ({Country} - {Category}): {Event} : Importance.{Importance}");
         }
 
+        /// <summary>
+        /// Convert this instance to a CSV file (tab delimited)
+        /// </summary>
+        /// <returns>string as CSV</returns>
+        public string ToCsv()
```
Jay-Jay-D: The docstring still says tab delimited.

BTW, is there a reason why you add some entries with quotes?

Common/Util/StreamReaderExtensions.cs
```diff
@@ -76,6 +76,46 @@ public static decimal GetDecimal(this StreamReader stream, char delimiter = Defa
...
         }
 
+        /// <summary>
+        /// Reads the next string value from the CSV file
+        /// </summary>
+        /// <param name="stream">The data stream</param>
+        /// <param name="delimiter">The data delimiter character to use, default is ','</param>
+        /// <returns>The string value read from CSV</returns>
+        public static string GetNextCsv(this StreamReader stream, char delimiter = DefaultDelimiter)
+        {
```
Jay-Jay-D: It is not clear to me the reason for this extension method, it is only used here:

https://github.com/QuantConnect/Lean/blob/7fd9862eb1ab69f285e0d50e6acc623ad5df30eb/Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs#L249

Why don't simply use `streamReader.ReadLine().ToCsvData()` in `TradingEconomicsCalendar.Reader()`?

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
gsalaz98: These fields can contain the delimiter in their value (`,`), which we escape by placing the entire value around double quotes. Unfortunately, they can not be removed

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
gsalaz98: There are instances where the country name could differ in capitalization

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
gsalaz98: We needed a character that would not clash with other `Symbol` implementations and also characters that are not invalid (`|` and ` `).

Common/Data/Custom/TradingEconomics/TradingEconomicsCalendar.cs
gsalaz98: Yes, we need the quotes in order to escape values that contain the `,` delimiter as part of their values.

Common/Util/StreamReaderExtensions.cs
gsalaz98: This implementation has added benefits in being able to parse values enclosed in quotes. I do plan on using that implementation in the future for other data sources 😄

Jay-Jay-D: `ToCsvData` is able to parse values enclosed in quotes. 

Unless better criteria, I would recommend removing this method and its usage in `TradingEconomicsCalendar.Reader()`. 

Any thoughts @Martin-Molinero?

gsalaz98: There is one benefit to the new method, and that is that we completely let go of the integer indexing style of reading string values from the CSV.

So for example, if we have this data and it changes to the new format:
Old:
```
20190101 00:00:00,20190101 00:00:00,hello,world!
```

New:
```
hello,20190101 00:00:00,world!,20190101 00:00:00
```

The old way of putting it into code would be like this (using ToCsvData(...)):

```
fn Reader(...)
{
    var data = reader.ReadLine().ToCsvData();

    instance.Time = Parse.Exact(data[0], "yyyyMMdd HH:mm:ss");
    instance.EndTime = Parse.Exact(data[1], "yyyyMMdd HH:mm:ss");
    instance.Text = data[2];
    instance.Last = data[3];
}
```

But would have to be done like this after the changes:

```
fn Reader(...)
{
    var data = reader.ReadLine().ToCsvData();

    instance.Text = data[0];
    instance.Time = Parse.Exact(data[1], "yyyyMMdd HH:mm:ss");
    instance.Last = data[2];
    instance.EndTime = Parse.Exact(data[3], "yyyyMMdd HH:mm:ss");
}
```

Which can be very error prone when updating the number of the CSV index!

However, if we use the new `GetNextCsv()` method, we can rewrite the above to the following:
Old:
```
fn Reader(...)
{
    instance.Time = reader.GetDateTime("yyyyMMdd HH:mm:ss");
    instance.EndTime = reader.GetDateTime("yyyyMMdd HH:mm:ss");
    instance.Text = reader.GetNextCsv();
    instance.Last = reader.GetNextCsv();
}
```

New:
```
fn Reader(...)
{
    instance.Text = reader.GetNextCsv();
    instance.Time = reader.GetDateTime("yyyyMMdd HH:mm:ss");
    instance.Last = reader.GetNextCsv();
    instance.EndTime = reader.GetDateTime("yyyyMMdd HH:mm:ss");
}
```
All it required was just switching the lines around to the order we encountered them.

Given how much I've messed up CSV indexes in the past, I'd much rather prefer using `GetNextCsv()` 😄

Tests/Common/Util/StreamReaderExtensionsTests.cs
```diff
@@ -113,6 +113,31 @@ public void GetDecimalEmptyString()
             Assert.AreEqual(0, smartStream.GetDecimal());
         }
 
+        [Test]
+        public void GetNextCsv()
+        {
+            var stream = "1,2,3,\"hello\",\"world\"\r\n3,2,1,\"goodbye\",\"world\"\n2,3,1,\"hello again, world!\",\"goodbye again, world!\"\r\n".ToStream();
```
Martin-Molinero: Thanks for the unit test!
I think we need a few more though, it will fail if the stream does not end in `\r\n`, please test with different possible endings `,` or `"`, etc

Common/Util/StreamReaderExtensions.cs
```diff
@@ -76,6 +76,50 @@ public static decimal GetDecimal(this StreamReader stream, char delimiter = Defa
...
+                        continue;
+                    }
+                }
+                else if (current == '"' && (stream.Peek() == delimiter || stream.Peek() == '\n' || stream.Peek() == '\r' &&
+                    (stream.Peek() != '\n' || stream.Read() == '\n')))
+                {
+                    inQuotes = false;
+                    current = (char)stream.Read();
+                    continue;
+                }
```
Martin-Molinero: I think this is equivalent? 🤔 

```
while (!((!inQuotes && current == delimiter) || current == '\n' || current == '\r' && (stream.Peek() != '\n' || stream.Read() == '\n') || current == NoMoreData))
{
    if (!inQuotes && current == '"')
    {
        inQuotes = true;
        current = (char)stream.Read();
    }

    if (current == '"' && (stream.Peek() == delimiter || stream.Peek() == '\n' || stream.Peek() == '\r' &&
        (stream.Peek() != '\n' || stream.Read() == '\n')))
    {
        inQuotes = false;
        current = (char)stream.Read();
        continue;
    }

    builder.Append(current);
    current = (char)stream.Read();
}
```

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
Martin-Molinero: It would be nice if it was normalized, all upper case maybe? Same for the Symbols `public const string InterestRate = "United-States//FDTR"` -> "UNITED-STATES//FDTR" - and same for the data path 🤔

Common/Data/Custom/TradingEconomics/TradingEconomics.Calendar.cs
```diff
@@ -0,0 +1,3571 @@
...
+
+            /// <summary>
+            /// Australia
+            /// </summary>
+            public static class Australia
+            {
+                /// <summary>
+                /// Balance of Trade
+                /// </summary>
+                public const string BalanceOfTrade = "Australia//AUITGSB";
```
Martin-Molinero: nit - Why not use the `Delimiter` for these?

Tests/Common/Util/StreamReaderExtensionsTests.cs
Martin-Molinero: *Think its missing to check the `NoMoreData` case

Common/Util/StreamReaderExtensions.cs
Jay-Jay-D: As always, you have a sound reason why.

My main concerns are readability and maintainability of code that - in principle - makes a task already implemented.

Jay-Jay-D:
Respect https://github.com/QuantConnect/Lean/pull/4054#discussion_r374902507

All our data use lower cases for paths; I'd strongly suggest respecting the standard for this data source too.

Common/Util/StreamReaderExtensions.cs
Martin-Molinero: The main value of the `StreamReader` is performance. `GetNextCsv()` will parse the line and generate the final strings in the same operation, while `ReadLine()` + `ToCsvData()` will read the line, generate a new string, later parse that string again and generate the final strings. We know this makes the `StreamReader` ~50% faster. More at https://github.com/QuantConnect/Lean/pull/3857 . **But its more complex**, I think I agree with @Jay-Jay-D, if there isn't a measurable performance gain, the added complexity is not worth the trouble. The gain will be measurable from the Lean user perspective with big data sets.
@gsalaz98 regarding the indexing, maybe it could use and `int index = 0`, which gets incremented on each access `[index++]`

ToolBox/TradingEconomicsDataDownloader/TradingEconomicsCalendarDownloader.cs
Martin-Molinero: Nevermind the data path 👍  https://github.com/QuantConnect/Lean/pull/4054#issuecomment-582371867

Issue #1330, opened 2017-11-18
Specify file paths for config and output files


p9raffi:
Currently, with the exception of transaction logs, the path to any kind of output appears to be hard-coded to the current directory of QuantConnect.Lean.Launcher.exe. Even then, the transaction logs path must be specified within config.json. Additionally, the path to config.json is also fixed.

Being able to pass in various paths from the command line would make things a lot more flexible in that project management would be easier and external code could have control over where things go and what they are named. This could be used for keeping track of outputs associated with individual backtests, as well as other things.

The current list of files that it would be useful to have command-line arguments for:

Configuration File
Backtest Results
Live Results (this would be a path only)
Transactions Log
Backtest Log File

jeffneuen:
I would second this as being pretty important for those running Lean in a docker container. If I understand the current workflow properly, when you build the Docker container, the build copies Lean/Launcher/* into the container, and runs QuantConnect.Lean.Launcher.exe from inside the container. After the container is run and the backtest is completed, backtest result .json fuel is then dropped into Lean/Launcher in the container, and then you're stuck with results trapped inside the container filesystem.

It seems like if it was possible to specify the location of config.json, that would save a lot of container rebuilds on every edit of that one file. You could mount the volume when you launch the docker container.

Additionally, if if was possible to specify the location of the backtest output .json file, it would make it much easier to get the backtest results outside of the container via volume mounting on container launch.

Maybe there are already ways to do this that I'm missing.

PR #4156, opened 2020-03-04
Feature 1330 specify paths for output files

adam-may:
#### Description
Adds output directory configuration options for insights and results

#### Related Issue
Addresses #1330

#### Motivation and Context
Ability to build a read only docker container with the x-copied binaries, then a mounted directory for saving the algorithm outputs outside of the container

Added/Changed Lean Launcher command line parameters.

Unit tests written

#### Types of changes

#### Checklist:


Engine/Alphas/DefaultAlphaHandler.cs
```diff
@@ -251,9 +252,10 @@ protected virtual void StoreInsights()
                     var insights = InsightManager.AllInsights.OrderBy(insight => insight.GeneratedTimeUtc).ToList();
                     if (insights.Count > 0)
                     {
-                        var directory = Path.Combine(Directory.GetCurrentDirectory(), AlgorithmId);
-                        var path = Path.Combine(directory, "alpha-results.json");
+                        var baseDirectory = Config.Get("insights-destination-folder", Directory.GetCurrentDirectory());
```
jaredbroad: Minor but would be good to store `path` to local variable to avoid checking config 
`Config.Get("insights-destination-folder", Directory.GetCurrentDirectory());`

adam-may: Yes, fair call. I'll change it.

Engine/Results/BaseResultsHandler.cs
```diff
@@ -156,6 +157,17 @@ public abstract class BaseResultsHandler
...
+        /// <summary>
+        /// Directory location to store results
+        /// </summary>
+        protected string ResultsDestinationFolder;
+
+        /// <summary>
+        /// Root filename to store results
+        /// </summary>
+        /// <remarks>Should not include the file extension here</remarks>
+        protected string ResultsDestinationFilePrefix => Config.Get("results-destination-file-prefix", AlgorithmId);
```
Martin-Molinero: Wondering if the `AlgorithmId` could have the same effect as `ResultsDestinationFilePrefix`? Maybe we need a way of setting the `AlgorithmId` (BacktestId and DeployId at JobQueue.cs) to be different than `AlgorithmTypeName`, but maybe defaulting to it 🤔. That way we would get uniqueness across all users of `AlgorithmId`

Engine/Alphas/DefaultAlphaHandler.cs
Martin-Molinero: Wonder if we need a separate `insights-destination-folder` than `results-destination-folder`, could they be the same? Thinking of what is the use case for having them separate

adam-may: I couldn't think of one to be honest, but figured you could always default them to the same thing, but it'd be difficult to go the other way. Happy to combine them to one value.

adam-may: I've followed your question I've merged the insights directory location into results directory location. It can always be spun out later if required.

Common/Algorithm/Framework/Alphas/Analysis/IInsightManager.cs
```diff
@@ -0,0 +1,69 @@
+using System;
+using System.Collections.Generic;
+
+namespace QuantConnect.Algorithm.Framework.Alphas.Analysis
+{
+    /// <summary>
+    /// Encapsulates the storage and on-line scoring of insights.
+    /// </summary>b
+    public interface IInsightManager
```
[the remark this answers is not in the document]
adam-may: Both done

Common/Algorithm/Framework/Alphas/Analysis/IInsightManager.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="insightIds">The insights ids to be removed</param>
+        void RemoveInsights(IEnumerable<Guid> insightIds);
+
+        /// <summary>
+        /// Gets all insight analysis contexts that have been updated since this method's last invocation.
+        /// Contexts are marked as not updated during the enumeration, so in order to remove a context from
+        /// the updated set, the enumerable must be enumerated.
+        /// </summary>
+        /// <returns></returns>
+        IEnumerable<InsightAnalysisContext> GetUpdatedContexts();
```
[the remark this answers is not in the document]
adam-may: Removed both from Interface

Engine/Results/BaseResultsHandler.cs
adam-may: @Martin-Molinero - I'm not quite sure what to do with this one? Can you give me an example of what you're thinking? (Happy to make the change)

Martin-Molinero: I was thinking something like the following https://github.com/QuantConnect/Lean/commit/b49fb478c98ad3e83fee728b75bd266711573033, unit tests need an update too I believe

PR #4175, opened 2020-03-09
Adds null value skipping and handling to Result deserialization via new JsonConverter

gsalaz98:
#### Description

We now skip any encountered null values during the deserialization process via a new JsonConverter implementation.

This was done to prevent runtime crashes due to null values in the backend report generation service, as well as preventing ourselves from presenting incorrect data about an algorithm if a null value is defaulted to `0`.

#### Related Issue

N/A

#### Motivation and Context

Makes report generation more stable and fixes issues with JSON parsing by ignoring null values.

No

Local testing and backend testing
#### Types of changes

#### Checklist:


Report/NullResultValueTypeJsonConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            foreach (JProperty property in token["Charts"].Children())
+            {
+                foreach (JProperty seriesProperty in property.Value["Series"])
+                {
+                    var newValues = new List<JToken>();
+                    foreach (var entry in seriesProperty.Value["Values"])
+                    {
+                        if (entry["x"] == null || entry["x"].Value<long?>() == null ||
+                            entry["y"] == null || entry["y"].Value<decimal?>() == null)
```
Martin-Molinero: Wondering in which case could this happen? `Series` use `ChartPoint` types for which 'x' and 'y' are not nullable 🤔. Could we try fixing this on the producer side instead? It would be nice to have some unit test too

gsalaz98: ChartPoint values will never be `null` under normal circumstances when exporting via the QC Terminal. However, some backend services can potentially provide a `ChartPoint` with null value(s).

Good call on the unit test. I'll go ahead and add one for this specific case 👍

PR #4239, opened 2020-03-25
Update Statistics.cs

simonsonjack:
Updates to fix AnnualPerformance and TrackingError statistics.

#### Description

Updated Annual Performance to reflect the effect of compounding returns when annualizing a daily average return. Tracking Error changed to reflect the correct formula, which is the square root of the variance of the element-wise difference between algorithm and benchmark returns.

#### Motivation and Context

Previous calculations were not correct.

Added unit tests to the project.
#### Types of changes

#### Checklist:


Tests/Common/Statistics/AnnualPerformanceTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using NUnit.Framework;
+
+namespace QuantConnect.Tests.Common.Statistics
+{
+    [TestFixture]
+    public class AnnualPerformanceTests
+    {
+        [Test]
+        public void ZeroTradingDays()
```
jaredbroad: Please write a test against the SPY daily test data included in LEAN with realistic daily performances and realistic annual performance. 

You can reuse the same array of SPY performance data to cherry pick periods with high returns and test better annual performances. Use a [TestData] notation on the test method to pass in a range of dates and expected results.

Tests/Common/Statistics/TrackingErrorTests.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        [Test]
+        public void IdenticalPerformance()
+        {
+            var random = new Random();
+
+            var benchmarkPerformance = Enumerable.Repeat(random.NextDouble(), 252).ToList();
+            var algoPerformance = benchmarkPerformance.Select(element => element).ToList();
+
+            var result = QuantConnect.Statistics.Statistics.TrackingError(algoPerformance, benchmarkPerformance);
```
jaredbroad: Cool :+1:

Tests/Common/Statistics/TrackingErrorTests.cs
```diff
@@ -0,0 +1,69 @@
...
+                algoPerformance.Add((baseReturn * 2) + 2);
+            }
+
+            var result = QuantConnect.Statistics.Statistics.TrackingError(algoPerformance, benchmarkPerformance);
+
+            Assert.AreEqual(0.0, result);
+        }
+
+        [Test]
+        public void AllZeros()
```
jaredbroad: Please also add non 0 test with SPY and AAPL daily performance datasets.

Common/Statistics/Statistics.cs
```diff
@@ -517,7 +517,13 @@ public static double Alpha(List<double> algoPerformance, List<double> benchmarkP
         /// <returns>Value for tracking error</returns>
         public static double TrackingError(List<double> algoPerformance, List<double> benchmarkPerformance)
         {
-            return Math.Sqrt(AnnualVariance(algoPerformance) - 2 * Correlation.Pearson(algoPerformance, benchmarkPerformance) * AnnualStandardDeviation(algoPerformance) * AnnualStandardDeviation(benchmarkPerformance) + AnnualVariance(benchmarkPerformance));
+            var performanceDifference = new List<double>();
```
jaredbroad: If arrays have different length will this explode? Try to always input check and not assume inputs are valid

simonsonjack: It would. Elsewhere there seems to be an assumption that array lengths are the same, but I'll add in some error handling. Would we want to return a specific value if they aren't the same length?

jaredbroad:
This will also change all the regression test statistics - please ask for help to learn how to update the unit tests and then post the new test values.

Common/Statistics/Statistics.cs
jaredbroad: You can use your best judgment here.

simonsonjack: I'm going to have it return NaN. Pretty much all of the statistics rely on the assumption that the algo and benchmark lists are of equal length, so un-equal lists would blow up a number of statistics. We could add in error handling or I could dig deeper into how these are used in Lean to see if we handle these possible cases elsewhere.

Tests/Common/Statistics/AnnualPerformanceTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        [SetUp]
+        public void GetSPY()
+        {
+            var symbol = Symbol.Create("SPY", SecurityType.Equity, Market.USA);
+            var path = LeanData.GenerateZipFilePath(Globals.DataFolder, symbol, new DateTime(2020, 3, 1),Resolution.Daily, TickType.Trade);
+            foreach (var line in QuantConnect.Compression.ReadLines(path))
+            {
+                var elements = line.Split(',').ToList();
+                var format = "yyyyMMdd HH:mm";
+                var bar = new TradeBar(DateTime.ParseExact(elements[0], format, CultureInfo.InvariantCulture),
```
Martin-Molinero: Nice 👍 
I think it would be better to use `TradeBar.ParseEquity()` directly, it would save all the parsing and having to know how to do it too! Maybe we could push this to a static method and reuse in these different tests, let's chat offline

simonsonjack:
Making new PR from a clean branch

PR #4248, opened 2020-03-26
Update Lean Statistics

simonsonjack:
Updates to fix AnnualPerformance and TrackingError statistics.

#### Description

Updated Annual Performance to reflect the effect of compounding returns when annualizing a daily average return. Tracking Error changed to reflect the correct formula, which is the square root of the variance of the element-wise difference between algorithm and benchmark returns.

#### Motivation and Context

Previous calculations were not correct.

Added unit tests to the project.
#### Types of changes

#### Checklist:


Common/Statistics/PortfolioStatistics.cs
```diff
@@ -300,5 +297,30 @@ private static decimal GetAnnualVariance(List<double> performance, int tradingDa
...
+
+        /// <summary>
+        /// Tracking error volatility (TEV) statistic - a measure of how closely a portfolio follows the index to which it is benchmarked
+        /// </summary>
+        /// <remarks>If algo = benchmark, TEV = 0</remarks>
+        /// <param name="listPerformance">Double collection of algorithm performance values</param>
+        /// <param name="listBenchmark">Double collection of benchmark performance values</param>
+        /// <param name="tradingDaysPerYear">Trading days per year for the assets in portfolio</param>
+        /// <returns>Tracking error value</returns>
+        private static decimal GetTrackingError(List<double> listPerformance, List<double> listBenchmark, int tradingDaysPerYear = 252)
```
Martin-Molinero: I'd suggest not adding this new method an use the existing fixed implementation at `Statistics`, we can add the optional `tradingDaysPerYear` argument

Common/Statistics/PortfolioStatistics.cs
```diff
@@ -285,7 +282,7 @@ private static decimal DrawdownPercent(SortedDictionary<DateTime, decimal> equit
         /// <returns>Double annual performance percentage</returns>
         private static decimal GetAnnualPerformance(List<double> performance, int tradingDaysPerYear = 252)
         {
-            return (decimal)performance.Average() * tradingDaysPerYear;
+            return (decimal)Math.Pow((performance.Average() + 1), tradingDaysPerYear) - 1;
```
Martin-Molinero: I'd suggest removing this method and using the existing fixed implementation at Statistics

Tests/Common/Statistics/AnnualPerformanceTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Tests.Common.Statistics
+{
+    [TestFixture]
+    public class AnnualPerformanceTests
+    {
+        private List<TradeBar> _spy = new List<TradeBar>();
+
+        [SetUp]
```
Martin-Molinero: Think we could use `TestFixtureSetUp` here so that the setup happens once for all the fixture and not once per test, for teardown it would be `TestFixtureTearDown`. Same for `TrackingErrorTests`

Nice tests!

PR #4328, opened 2020-04-21
Refactor Future Market

Martin-Molinero:
#### Description

- Replacing usage of generic `Market.USA` for Future Symbols. Will now use the correct exchange market for each Symbol, if not specified it will be provided by the `SymbolPropertiesDatabase`
- Moving future data and margins to their correct folder
- Updating the `FuturesExpiryFunctions` to be keyed by `Symbol`. If provided `Symbol` isn't canonical it will create a temp canonical `Symbol` to use instead.
- Updating unit and regression tests. Regression tests OrderListHash changed because `Market` is part of the Symbol in the Order.
- Updating `AlgoSeekFuture` reader and processor
- Updating `IQFeedDataQueueHandler` and `IQFeedDataQueueUniverseProvider`
- Updating `InteractiveBrokersBrokerage` and `InteractiveBrokersFeeModel`
#### Related Issue

N/A
#### Motivation and Context

Futures are not fungible. A Future contract with the same ticker in different exchanges is another asset. Until this PR Lean was using `Market.USA` for all future exchanges, not allowing the same future ticker to exist for different exchanges.

N/A

Unit and regression tests
#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.cs
```diff
@@ -1528,7 +1534,8 @@ public Future AddFuture(string ticker, Resolution? resolution = null, string mar
         {
             if (market == null)
             {
-                if (!BrokerageModel.DefaultMarkets.TryGetValue(SecurityType.Future, out market))
+                if (!SymbolPropertiesDatabase.TryGetMarket(ticker, SecurityType.Future, out market)
+                    && !BrokerageModel.DefaultMarkets.TryGetValue(SecurityType.Future, out market))
```
gsalaz98, line 1538: Is there a reason why we chose not to override the `market` parameter if it equals `Market.USA`? 

When we go to fetch the entry from the MHDB, it will inevitably fail even if we let the user use the `Market.USA` market: https://github.com/QuantConnect/Lean/pull/4328/files#diff-0417c0a6ef4ab67693f9e40a4a597914R201-R213

Martin-Molinero: Thanks for the thorough review! 👍 
We discussed this and avoided remapping it, because it would be rare if users specified a Market currently

PR #4337, opened 2020-04-24
Enable parallel unit testing

Martin-Molinero:
#### Description

- Enable parallel unit testing, Travis build now takes ~13min!
- Adding a few unit test and micro performance improvements
#### Related Issue

N/A
#### Motivation and Context

Faster tests are better tests

N/A

Unit and regression test, many times
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/Results/BaseResultsHandler.cs
+++ b/Engine/Results/BaseResultsHandler.cs
@@ -332,7 +332,7 @@ namespace QuantConnect.Lean.Engine.Results
         /// <param name="result">The results to save</param>
         public virtual void SaveResults(string name, Result result)
         {
-            File.WriteAllText(GetResultsPath(name), JsonConvert.SerializeObject(result));
+            File.WriteAllText(GetResultsPath(name), JsonConvert.SerializeObject(result, Formatting.Indented));
         }
 
         /// <summary>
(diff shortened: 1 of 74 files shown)


Messaging/StreamingApi.cs
```diff
@@ -35,10 +33,10 @@ public static class StreamingApi
         public static readonly bool IsEnabled = Config.GetBool("send-via-api");
 
         // Client for sending asynchronous requests.
-        private static readonly RestClient Client = new RestClient("http://streaming.quantconnect.com")
+        private static readonly Lazy<RestClient> Client = new Lazy<RestClient>(() => new RestClient("http://streaming.quantconnect.com")
```
jaredbroad, line 36: I'd prefer to delete rather than update this ancient code; which since its a new larger problem suggest we just revert these changes and fix this class properly in a dedicated PR.

Engine/Results/BaseResultsHandler.cs
```diff
@@ -316,7 +314,7 @@ public virtual string SaveLogs(string id, List<LogEntry> logs)
         /// <param name="result">The results to save</param>
         public virtual void SaveResults(string name, Result result)
         {
-            File.WriteAllText(GetResultsPath(name), JsonConvert.SerializeObject(result, Formatting.Indented));
```
jaredbroad, line 319: I don't see the harm in leaving this indented for ease of use locally? Why did you choose to remove?

Martin-Molinero: So regression tests are as fast as possible, I'll revert the change here and only apply it for the `RegressionResultHandler` 👍

jaredbroad:
🎉 tests #FTW !

Issue #4284, opened 2020-04-06
Missing Methods in PandasData.Remapper


AlexCatarino:
#### Expected Behavior
PandasData.Remapper handles all Pandas.DataFrames methods.

#### Actual Behavior
Some methods are still not supported, such as `ffill` and `fillna`

#### Potential Solution
Adds all the methods from pandas.DataFrame/ pandas.Series.

#### Reproducing the Problem
```python
symbols = [self.AddEquity("SPY")]
prices =self.History(symbols, 10, Resolution.Daily).ffill().fillna(0.0)
for symbol in symbols:
    df = prices.loc[symbol]
```
yields 
```TypeError : object is not callable```
#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #4365, opened 2020-05-04
Fixes Not Wrapped pandas.Dataframe

AlexCatarino:
#### Description
`Remapper.__getitem__`  was not returns a `Remapper` object when the result was `pandas.DataFrame`. It is needed for a sequence of `.loc.` calls.
Refactors `Remapper._self_mapper` to handle tuples where the key can be found in both first and second positions.

Update unit tests that should test `Symbol` object as key, but were using `str(Symbol)`.

#### Related Issue
Related to #4284 

#### Motivation and Context
But fix 

No.

New unit tests. Tested in Research Notebook in QuantConnect Cloud.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Python/PandasData.cs
+++ b/Common/Python/PandasData.cs
@@ -82,39 +82,26 @@ class Remapper(wrapt.ObjectProxy):
     def __init__(self, wrapped):
         super(Remapper, self).__init__(wrapped)
 
-    # Our remapping method. Originally implemented in C# but some cases were not working
-    # correctly and using Py did the trick
     def _self_mapper(self, key):
-        # this is to improve user experience, they can use Symbol
-        # as key and we convert it to string for pandas
+        ''' Our remapping method.
+        Originally implemented in C# but some cases were not working correctly and using Py did the trick.
+        This is to improve user experience, they can use Symbol as key and we convert it to string for pandas
+        '''
         if isinstance(key, Symbol):
-            key = str(key.ID)
+            return str(key.ID)
+
         # this is the most normal use case
-        elif isinstance(key, str):
+        if isinstance(key, str):
             tupleResult = SymbolCache.TryGetSymbol(key, None)
             if tupleResult[0]:
                 return str(tupleResult[1].ID)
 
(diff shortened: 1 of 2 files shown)


Tests/Python/PandasConverterTests.cs
```diff
@@ -488,7 +488,7 @@ public void BackwardsCompatibilityDataFrame_get_OnPropertyUsingSymbol()
                 dynamic test = PythonEngine.ModuleFromString("testModule",
                     @"
 def Test(dataFrame, symbol):
-    data = dataFrame.lastprice.get(str(symbol))
+    data = dataFrame.lastprice.get(symbol)
```
Martin-Molinero, line 491: Maybe its better to do the following to avoid losing any test case for the new and old tests, and it's just 1 more line
```csharp
[TestCase("symbol")]
[TestCase("str(symbol)")]
public void BackwardsCompatibilityDataFrame_get_OnPropertyUsingSymbol(string symbol)
{
    using (Py.GIL())
    {
        dynamic test = PythonEngine.ModuleFromString("testModule",
            $@"
def Test(dataFrame, symbol):
data = dataFrame.lastprice.get({symbol})
if data.empty:
raise Exception('Data is empty')").GetAttr("Test");

        Assert.DoesNotThrow(() => test(GetTestDataFrame(Symbols.SPY), Symbols.SPY));
    }
}

File: Common/Python/PandasData.cs [hunk 2]
```diff
@@ -82,39 +82,26 @@ class Remapper(wrapt.ObjectProxy):
...
-                            firstTuple = (result, key[0][1])
-                        if len(key[1]) == 1:
-                            return (firstTuple,)
-                        else:
-                            return (firstTuple, key[1])
+        # If the key is a tuple, convert the items
+        if isinstance(key, tuple) and 2 >= len(key) >= 1:
+            item0 = self._self_mapper(key[0])
+            return (item0,) if len(key) == 1 else \
+                (item0, self._self_mapper(key[1]))
```
Martin-Molinero, line 103: Nice! 👍

PR #4401, opened 2020-05-11
Fix Estimize processing issues that prevented two downloaders from completing

gsalaz98:
#### Description

Prevents the `EstimizeEstimatesDataDownloader`, `EstimizeReleaseDataDownloader` types from crashing when encountering a ticker with a "defunct" indicator, but we are unable to determine the delimiter.

~Note: I think we should move this shared behavior into the abstract class both of the Downloaders inherit from to avoid code duplication. Willing to do it as part of this PR.~

Done 😃 
#### Related Issue

N/A
#### Motivation and Context

Ensures Estimize processes successfully

No

Processed data for both `EstimizeEstimatesDataDownloader` and `EstimizeReleaseDataDownloader` successfully
#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/EstimizeDataDownloader/EstimizeDataDownloader.cs
+++ b/ToolBox/EstimizeDataDownloader/EstimizeDataDownloader.cs
@@ -35,6 +35,11 @@ namespace QuantConnect.ToolBox.EstimizeDataDownloader
     {
         private readonly string _clientKey;
         private readonly int _maxRetries = 5;
+        private static readonly List<char> _defunctDelimiters = new List<char>
+        {
+            '-',
+            '_'
+        };
 
         /// <summary>
         /// Control the rate of download per unit of time.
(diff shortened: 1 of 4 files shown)


Martin-Molinero:
> Note: I think we should move this shared behavior into the abstract class both of the Downloaders inherit from to avoid code duplication. Willing to do it as part of this PR.

Sounds great 😄 and some unit tests for the parser method would be great too. Could be a public static as `NormalizeTicker()`

ToolBox/EstimizeDataDownloader/EstimizeReleaseDataDownloader.cs
```diff
@@ -78,13 +78,15 @@ public override bool Run()
...
-                    var ticker = company.Ticker;
-                    if (ticker.IndexOf("defunct", StringComparison.OrdinalIgnoreCase) > 0)
+                    var estimizeTicker = company.Ticker;
+                    string ticker;
+
+                    if (!TryNormalizeDefunctTicker(estimizeTicker, out ticker))
                     {
-                        var length = ticker.IndexOf('-');
-                        ticker = ticker.Substring(0, length).Trim();
+                        Log.Error($"EstimizeReleaseDataDownloader(): Defunct ticker {estimizeTicker} is unable to be parsed. Continuing...");
```
Martin-Molinero, line 86: missing `continue` statement

ToolBox/EstimizeDataDownloader/EstimizeDataDownloader.cs
```diff
@@ -159,6 +164,39 @@ protected void SaveContentToFile(string destinationFolder, string ticker, IEnume
...
+                {
+                    var length = ticker.IndexOf(delimChar);
+
+                    // Continue until we exhaust all delimiters
+                    if (length == -1)
+                    {
+                        continue;
+                    }
+
+                    nonDefunctTicker = ticker.Substring(0, length).Trim();
```
Martin-Molinero, line 188: Based on the tests, `Trim` isn't really required

gsalaz98: I'll update the tests to include a `TICKER -defunct` case. I've seen tickers with a space between the defunct indicator before, hence the `.Trim()`

PR #4407, opened 2020-05-12
Add a futures universe selection model that uses open interest

michael-sena:
Adds a purposed futures universe selection model, that will assist with identifying the current active contract / contracts.

#### Description
Universe selection is done by obtaining the last open interest for each contract.  The contracts are then sorted by this value, and provided back for use.

#### Related Issue
N/A

#### Motivation and Context
While the front month in a futures chain is often also the most outstanding orders, this isn't always the case. In order to select the right contract in the chain, open interest should be used to sort the contracts.

This model provides an easy way for that to be done.

N/A

Yes using a broker, and comparing the broker's expected active contract to the returned results.

#### Types of changes

#### Checklist:
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>` - *it is missing the issue, does that matter?*

Credits also to @StefanoRaggi for his assistance in this model, and the original version of this class.

### DIFF
--- a/Algorithm.CSharp/OpenInterestFuturesRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OpenInterestFuturesRegressionAlgorithm.cs
@@ -1,0 +1,139 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Futures framework algorithm that uses open interest to select the active contract.
+    /// </summary>
+    /// <meta name="tag" content="regression test" />
+    /// <meta name="tag" content="futures" />
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="filter selection" />
+    public class OpenInterestFuturesRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        private static readonly HashSet<DateTime> ExpectedExpiryDates = new HashSet<DateTime>
+        {
+            new DateTime(2013, 12, 27),
+            new DateTime(2014, 02, 26)
+        };
+
+        public override void Initialize()
+        {
+            UniverseSettings.Resolution = Resolution.Tick;
+
+            SetStartDate(2013, 10, 08);
+            SetEndDate(2013, 10, 11);
+            SetCash(10000000);
+
+            // set framework models
+            SetUniverseSelection(
+                new OpenInterestFutureUniverseSelectionModel(
+                    this,
+                    t => new[] {QuantConnect.Symbol.Create(Futures.Metals.Gold, SecurityType.Future, Market.COMEX)},
+                    null,
+                    ExpectedExpiryDates.Count
+                )
+            );
+        }
+
+        public override void OnData(Slice slice)
+        {
+            if (Transactions.OrdersCount == 0 && slice.HasData)
+            {
+                var matched = slice.Keys.Where(s => !ExpectedExpiryDates.Contains(s.ID.Date)).ToList();
+                if (matched.Count != 0)
+                {
+                    throw new Exception($"{matched.Count}/{slice.Keys.Count} were unexpected expiry date(s): " + string.Join(", ", matched.Select(x => x.ID.Date)));
+                }
+
+                foreach (var symbol in slice.Keys)
+                {
+                    MarketOrder(symbol, 1);
+                }
+            }
+            else if (Portfolio.Any(p => p.Value.Invested))
+            {
+                Liquidate();
+            }
+        }
+
+        /// <summary>
+        ///     This is used by the regression test system to indicate if the open source Lean repository has the required data to
+        ///     run this algorithm.
+        /// </summary>
+        public bool CanRunLocally { get; } = true;
+
+        /// <summary>
+        ///     This is used by the regression test system to indicate which languages this algorithm is written in.
+        /// </summary>
+        public Language[] Languages { get; } = {Language.CSharp};
+
+        /// <summary>
+        /// This is used by the regression test system to indicate what the expected statistics are from running the algorithm
+        /// </summary>
+        public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>
+        {
+            {"Total Trades", "4"},
+            {"Average Win", "0.00%"},
+            {"Average Loss", "0.00%"},
+            {"Compounding Annual Return", "0.003%"},
+            {"Drawdown", "0%"},
+            {"Expectancy", "0.351"},
+            {"Net Profit", "0.000%"},
+            {"Sharpe Ratio", "0"},
+            {"Probabilistic Sharpe Ratio", "0%"},
+            {"Loss Rate", "50%"},
+            {"Win Rate", "50%"},
+            {"Profit-Loss Ratio", "1.70"},
+            {"Alpha", "0"},
+            {"Beta", "0"},
+            {"Annual Standard Deviation", "0"},
+            {"Annual Variance", "0"},
+            {"Information Ratio", "-58.133"},
+            {"Tracking Error", "0.173"},
+            {"Treynor Ratio", "0"},
+            {"Total Fees", "$7.40"},
+            {"Fitness Score", "0.017"},
+            {"Kelly Criterion Estimate", "0"},
+            {"Kelly Criterion Probability Value", "0"},
+            {"Sortino Ratio", "79228162514264337593543950335"},
+            {"Return Over Maximum Drawdown", "79228162514264337593543950335"},
+            {"Portfolio Turnover", "0.017"},
+            {"Total Insights Generated", "0"},
+            {"Total Insights Closed", "0"},
+            {"Total Insights Analysis Completed", "0"},
+            {"Long Insight Count", "0"},
+            {"Short Insight Count", "0"},
+            {"Long/Short Ratio", "100%"},
+            {"Estimated Monthly Alpha Value", "$0"},
+            {"Total Accumulated Estimated Alpha Value", "$0"},
+            {"Mean Population Estimated Insight Value", "$0"},
+            {"Mean Population Direction", "0%"},
+            {"Mean Population Magnitude", "0%"},
+            {"Rolling Averaged Population Direction", "0%"},
+            {"Rolling Averaged Population Magnitude", "0%"},
+            {"OrderListHash", "824503313"}
+        };
+    }
+}
--- a/Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
+++ b/Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
@@ -1,0 +1,123 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NodaTime;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Framework.Selection
+{
+    /// <summary>
+    ///     Selects contracts in a futures universe, sorted by open interest.  This allows the selection to identifiy current
+    ///     active contract.
+    /// </summary>
+    public class OpenInterestFutureUniverseSelectionModel : FutureUniverseSelectionModel
+    {
+        private readonly int? _chainContractsLookupLimit;
+        private readonly IAlgorithm _algorithm;
+        private readonly int? _resultsLimit;
+        private readonly MarketHoursDatabase _marketHoursDatabase;
+
+        /// <summary>
+        ///     Creates a new instance of <see cref="OpenInterestFutureUniverseSelectionModel" />
+        /// </summary>
+        /// <param name="algorithm">Algorithm</param>
+        /// <param name="futureChainSymbolSelector">Selects symbols from the provided future chain</param>
+        /// <param name="chainContractsLookupLimit">Limit on how many contracts to query for open interest</param>
+        /// <param name="resultsLimit">Limit on how many contracts will be part of the universe</param>
+        public OpenInterestFutureUniverseSelectionModel(IAlgorithm algorithm, Func<DateTime, IEnumerable<Symbol>> futureChainSymbolSelector, int? chainContractsLookupLimit = 6,
+            int? resultsLimit = 1) : base(TimeSpan.FromDays(1), futureChainSymbolSelector)
+        {
+            _marketHoursDatabase = MarketHoursDatabase.FromDataFolder();
+            if (algorithm == null)
+            {
+                throw new ArgumentNullException(nameof(algorithm));
+            }
+
+            _algorithm = algorithm;
+            _resultsLimit = resultsLimit;
+            _chainContractsLookupLimit = chainContractsLookupLimit;
+        }
+
+        /// <summary>
+        ///     Defines the future chain universe filter
+        /// </summary>
+        protected override FutureFilterUniverse Filter(FutureFilterUniverse filter)
+        {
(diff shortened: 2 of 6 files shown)


Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using QuantConnect.Algorithm.Framework.Selection;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Farmework.Selection
```
StefanoRaggi, line 27: Small typo here

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <returns>Filtered set</returns>
+        public IEnumerable<Symbol> FilterByOpenInterest(IEnumerable<Symbol> contracts)
+        {
+            return contracts.OrderBy(x => x.ID.Date).Take(_chainContractsLookupLimit).OrderByDescending(GetOpenInterest).ThenBy(x => x.ID.Date).Take(_resultsLimit);
+        }
+
+        private decimal GetOpenInterest(Symbol symbol)
+        {
+            var current = _algorithm.Time;
+            var previousBusinessDate = GetPreviousBusinessDay(_algorithm.Time, symbol);
```
StefanoRaggi, line 80: Doesn't this assume the algorithm time zone defaulting to `TimeZones.NewYork`? 
Maybe it's better to use `algorithm.UtcTime` instead?

Tests/Algorithm/Framework/Selection/OpenInterestFutureUniverseSelectionModelTests.cs
```diff
@@ -0,0 +1,122 @@
...
+using QuantConnect.Configuration;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Tests.Algorithm.Framework.Selection
+{
+    [TestFixture]
+    public class OpenInterestFutureUniverseSelectionModelTests
```
StefanoRaggi, line 18: Nice tests! 👍

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+                            null,
+                            true,
+                            false,
+                            DataNormalizationMode.Raw,
+                            TickType.OpenInterest
+                        )
+                    },
+                    TimeZones.Utc
+                )
+                .FirstOrDefault();
```
Martin-Molinero, line 101: Maybe we should get the `lastOrDefault()` instead?

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                var message = new StringBuilder(nameof(OpenInterestFutureUniverseSelectionModel)).Append('.')
+                    .Append(MethodBase.GetCurrentMethod().Name)
+                    .Append("(): Received ")
+                    .Append(ticks.Count)
+                    .Append(" entries for ")
+                    .Append(symbol.ToString())
+                    .Append(" tick data containing open interest. Will use first. Values: ")
+                    .Append(string.Join(",", ticks));
+                _algorithm.Error(message.ToString());
```
Martin-Molinero, line 120: Not sure this is an error? The history call could return more than 1 data point using tick resolution

Tests/Algorithm/Framework/Selection/OpenInterestFutureUniverseSelectionModelTests.cs
```diff
@@ -0,0 +1,122 @@
...
+        };
+        private Mock<IHistoryProvider> _mockHistoryProvider;
+        private OpenInterestFutureUniverseSelectionModel _underTest;
+
+        [SetUp]
+        public void SetUp()
+        {
+            Config.Reset();
+            MarketHoursDatabase.Reset();
+            Config.Set("data-folder", Path.Combine(TestContext.CurrentContext.TestDirectory, "..", "..", "..", "Data"));
```
Martin-Molinero, line 43: Wondering why are these required, I gave it a quick test and it seemed to work correctly without them too?
```
            Config.Reset();
            MarketHoursDatabase.Reset();
            Config.Set("data-folder", Path.Combine(TestContext.CurrentContext.TestDirectory, "..", "..", "..", "Data"));
```

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            _algorithm = algorithm;
+            _resultsLimit = resultsLimit;
+            _chainContractsLookupLimit = chainContractsLookupLimit;
+            _exchangeHours = SecurityExchangeHours.AlwaysOpen(TimeZones.Utc);
+        }
+
+        /// <inheritdoc cref="FutureUniverseSelectionModel.Filter" />
+        protected override FutureFilterUniverse Filter(FutureFilterUniverse filter)
```
Martin-Molinero, line 62: > /// inheritdoc cref="FutureUniverseSelectionModel.Filter" 

I believe we don't use `inheritdoc` because it broke the documentation generator

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+                        )
+                    },
+                    TimeZones.Utc
+                )
+                .FirstOrDefault();
+
+            var ticks = result?.Ticks?[symbol];
+            if (ticks == null || ticks.Count == 0)
+            {
+                _algorithm.Log($"WARN: Unable to retrieve open interest for: {symbol}");
```
Martin-Molinero, line 106: Does this happen often? I think we should add "OpenInterestFutureUniverseSelectionModel" in the log message

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.Farmework.Selection
+{
+    /// <summary>
+    ///     Selects contracts in a futures universe, sorted by open interest.  This allows the selection to identifiy current
+    ///     active contract.
+    /// </summary>
+    public class OpenInterestFutureUniverseSelectionModel : FutureUniverseSelectionModel
```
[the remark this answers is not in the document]
michael-sena: What do you mean by that sorry?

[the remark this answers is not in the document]
Martin-Molinero: Like this guy `BasicTemplateFuturesFrameworkAlgorithm` https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/BasicTemplateFuturesFrameworkAlgorithm.cs, a small simple algorithm which uses open-source available data to assert the behavior of the selection model and is an example of it usage.
*No need to define a custom alpha or PortfolioConstructionModel as `BasicTemplateFuturesFrameworkAlgorithm` does, just using the new `OpenInterestFutureUniverseSelectionModel` and placing at least 1 trade (as a way of asserting the behavior)

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
jaredbroad: Because we try to make all our comments readable without navigating so it is easier to understand in a web environment like Github/QC

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: Removed

Tests/Algorithm/Framework/Selection/OpenInterestFutureUniverseSelectionModelTests.cs
michael-sena: Left-over from code in our own project. :P

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: The contracts being provided in should all be active, so not having a value doesn't sound right to me.  If it were me, I'd rather it be an error - but I figured others would probably only want a warning.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: I've added the full doc.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
RichardDale: Note that most futures exchanges report open interest significantly after that trading session has finished, so it's quite common for it to be null (unknown) until that time.

Other exchanges provide an estimate then update it a day later.

It's certainly possible to pad it forward but it would be stale versus the other data points.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: Changed to taking the last, should there be multiple values.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+            var previousBusinessDate = GetPreviousBusinessDay(_algorithm.Time, symbol);
+            var result = _algorithm.HistoryProvider.GetHistory(
+                    new[]
+                    {
+                        new HistoryRequest(
+                            previousBusinessDate,
+                            current,
+                            typeof(Tick),
+                            symbol,
+                            Resolution.Tick,
```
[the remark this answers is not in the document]
michael-sena: I've changed to using the exchange hours as suggested.  Looks good. 

As for Resolution.Minute, for Open Interest, my understanding is that it only updates daily.  So the intent is to only get the most recent value, quickly.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        /// <summary>
+        ///     Creates a new instance of <see cref="OpenInterestFutureUniverseSelectionModel" />
+        /// </summary>
+        /// <param name="algorithm">Algorithm</param>
+        /// <param name="futureChainSymbolSelector">Selects symbols from the provided future chain</param>
+        /// <param name="chainContractsLookupLimit">Limit on how many contracts to query for open interest</param>
+        /// <param name="resultsLimit">Limit on how many contracts will be part of the universe</param>
+        public OpenInterestFutureUniverseSelectionModel(IAlgorithm algorithm, Func<DateTime, IEnumerable<Symbol>> futureChainSymbolSelector, int chainContractsLookupLimit = 6,
+            int resultsLimit = 1) : base(TimeSpan.FromDays(1), futureChainSymbolSelector)
```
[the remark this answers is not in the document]
michael-sena: Added

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: Addressed, as mentioned in lower conversations.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: Fixed

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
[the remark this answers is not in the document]
michael-sena: Okay, I've added a basic regression algorithm.  It really doesn't do much other than have in the statistics the number of expected trades be 4 - which matches the limit on the symbols.

I'm not sure how beneficial it is though, because it seems like the regression suite doesn't use actual data - and therefore, the historical look-up of open interest isn't providing anything of significance.

Do you have any suggestions?

Otherwise, is it worth keeping?

michael-sena:
@Martin-Molinero - With the batch history lookups, while history look-ups are passed to the provider in bulk, I believe if it's then passed to a broker, the broker can only receive a single requests?  (Looking at: `IBrokerage.GetHistory(HistoryRequest request)`.  May be worth allowing brokers (if they're able to) to query in bulk.

michael-sena:
I believe I've addressed all comments now, and resolved conversations outstanding with me.  Awaiting your review!

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: Let me know if you're happy with the approach I've taken @RichardDale, following @Martin-Molinero's suggestions.

RichardDale: There are implications for an open interest data point update changing the Universe "after the fact" especially where you have futures rolls occurring on the basis of open interest.  

Most futures traders understand this already but it's certainly a caveat emptor if you're trading spot month crude oil futures!

Otherwise, the approach looks good.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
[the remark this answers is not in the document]
Martin-Molinero: Cool, your right! Sounds good 👍

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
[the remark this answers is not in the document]
Martin-Molinero: >  It really doesn't do much other than have in the statistics the number of expected trades

That's okay 👍 my suggestion was adding it as a simple example and as a regression test (which asserts on the stats)

> doesn't use actual data

How/on what is it trading? Open-source data folder has GC future tick data available with open interest https://github.com/QuantConnect/Lean/tree/master/Data/future/comex/tick/gc, using the GC future in those available days should work for using the `OpenInterestFutureUniverseSelectionModel` and placing some trade?

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: As far as I know, there isn't any mechanism to do data revisions in Quant Connect.  So I this aspect is a bit baked into the cake, but I'll leave it up to @Martin-Molinero to decide if this can be resolved or addressed.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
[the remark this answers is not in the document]
michael-sena: Okay, I've updated it again with an additional check.  Hopefully it's a bit more inline with the expected approach now.

It's trading gold, but it's test is to look at the contract selection.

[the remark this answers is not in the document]
Martin-Molinero: Looks great ty 👍

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
Martin-Molinero: > there isn't any mechanism to do data revisions in Quant Connect

Raw _backtesting_ data is available at https://www.quantconnect.com/data/tree
Regarding live trading, I believe it will drill down to the history provider being used. The history request is at the most 1 day backward with no fill forwarding from which we take the last available value. If desired, the backward history request window could be reduced, but as is seems good to me 👍

Algorithm.CSharp/OpenInterestFuturesRegressionAlgorithm.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// <summary>
+        ///     This is used by the regression test system to indicate if the open source Lean repository has the required data to
+        ///     run this algorithm.
+        /// </summary>
+        public bool CanRunLocally { get; } = true;
+
+        /// <summary>
+        ///     This is used by the regression test system to indicate which languages this algorithm is written in.
+        /// </summary>
+        public Language[] Languages { get; } = {Language.CSharp, Language.Python};
```
Martin-Molinero, line 90: nit - `Language.Python` should be removed, else it will try loading the unexisting python version of this algorithm.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// </summary>
+        /// <param name="contracts">Contracts to filter</param>
+        /// <returns>Filtered set</returns>
+        public IEnumerable<Symbol> FilterByOpenInterest(IReadOnlyDictionary<Symbol, SecurityExchangeHours> contracts)
+        {
+            var filtered = _chainContractsLookupLimit.HasValue ? contracts.Keys.OrderBy(x => x.ID.Date).Take(_chainContractsLookupLimit.Value) : contracts.Keys;
+            var openInterest = filtered.GroupBy(x => contracts[x]).SelectMany(g => GetOpenInterest(g.Key, g.Select(i => i))).ToDictionary(x => x.Key, x => x.Value);
+
+            decimal o;
+            filtered = filtered.OrderByDescending(s => openInterest.TryGetValue(s, out o) ? o : -1).ThenBy(x => x.ID.Date);
```
Martin-Molinero, line 79: Maybe should orderBy directly on the resulting `openInterest`? else if the symbol is not in `openInterest` it will use -1 and it might return that symbol.
What do you think about adding something like:
```
if (openInterest.Count == 0)
{
    _algorithm.Error("OpenInterestFutureUniverseSelectionModel.FilterByOpenInterest(): failed to get historical open interest, will select no symbol");
    return Enumerable.Empty<Symbol>();
}
```

Algorithm.CSharp/OpenInterestFuturesRegressionAlgorithm.cs
michael-sena: Ah okay, noted.

Algorithm.Framework/Selection/OpenInterestFutureUniverseSelectionModel.cs
michael-sena: Okay, I've adopted a form of this.  Let me know if you're okay with that, and resolve if so.

michael-sena:
Hazzah!

PR #4411, opened 2020-05-13
Add support for saving Futures/Options data in LeanDataWriter

StefanoRaggi:
#### Description
- Added new `includeExpired` flag to `IDataQueueUniverseProvider.Lookup`, to enable history requests for **expired** futures and options contracts. 
- Updated the IB brokerage implementation to handle expired contracts in history requests.
- Added a new `DownloadAndSave()` method in `LeanDataWriter` to allow downloading futures/options data from any `IBrokerage` implementation and saving in LEAN format.
- Added new method in the `IBDataDownloader` class for downloading options/futures data.

#### Motivation and Context
- History requests for expired futures/options contracts are not supported
- `LeanDataWriter` does not support writing Futures/Options data

- Tested downloading historical data from IB

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/ToolBox/IBDownloader/IBDataDownloaderTests.cs
+++ b/Tests/ToolBox/IBDownloader/IBDataDownloaderTests.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Linq;
 using NUnit.Framework;
+using QuantConnect.Logging;
 using QuantConnect.ToolBox.IBDownloader;
 
 namespace QuantConnect.Tests.ToolBox.IBDownloader
@@ -24,16 +25,20 @@ namespace QuantConnect.Tests.ToolBox.IBDownloader
     [Ignore("These tests require the IBGateway to be installed.")]
     public class IBDataDownloaderTests
     {
-        [TestCase("ES", Resolution.Minute)]
-        public void DownloadsFuturesData(string ticker, Resolution resolution)
+        [TestCase("ES", Resolution.Daily, 15)]
+        [TestCase("ES", Resolution.Hour, 15)]
+        [TestCase("ES", Resolution.Minute, 15)]
+        public void DownloadsFuturesData(string ticker, Resolution resolution, int days)
         {
+            Log.LogHandler = new ConsoleLogHandler();
+
             const SecurityType securityType = SecurityType.Future;
 
             using (var downloader = new IBDataDownloader())
             {
                 var symbols = downloader.GetChainSymbols(ticker, securityType, true).ToList();
 
-                var startDate = DateTime.UtcNow.Date.AddDays(-15);
+                var startDate = DateTime.UtcNow.Date.AddDays(-days);
                 var endDate = DateTime.UtcNow.Date;
 
                 downloader.DownloadAndSave(symbols, resolution, securityType, TickType.Trade, startDate, endDate);
--- a/Tests/ToolBox/IBDownloader/IBDataDownloaderTests.cs
+++ b/Tests/ToolBox/IBDownloader/IBDataDownloaderTests.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
 using QuantConnect.Logging;
@@ -25,13 +26,17 @@ namespace QuantConnect.Tests.ToolBox.IBDownloader
     [Ignore("These tests require the IBGateway to be installed.")]
     public class IBDataDownloaderTests
     {
+        [SetUp]
+        public void Setup()
+        {
+            Log.LogHandler = new ConsoleLogHandler();
+        }
+
         [TestCase("ES", Resolution.Daily, 15)]
         [TestCase("ES", Resolution.Hour, 15)]
         [TestCase("ES", Resolution.Minute, 15)]
         public void DownloadsFuturesData(string ticker, Resolution resolution, int days)
         {
-            Log.LogHandler = new ConsoleLogHandler();
-
             const SecurityType securityType = SecurityType.Future;
 
             using (var downloader = new IBDataDownloader())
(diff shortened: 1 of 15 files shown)


ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,229 @@ public void Write(IEnumerable<BaseData> source)
...
+            }
+
+            var dataType = LeanData.GetDataType(_resolution, _tickType);
+
+            var marketHoursDatabase = MarketHoursDatabase.FromDataFolder();
+
+            var ticker = symbols.First().ID.Symbol;
+            var market = symbols.First().ID.Market;
+
+            var canonicalSymbol = Symbol.CreateFuture(ticker, market, SecurityIdentifier.DefaultDate);
```
Martin-Molinero, line 129: Maybe at the start we should assert all the provided symbols correspond to the same Future, else this might get confusing/messy. Also we should assert the security type is a Future? what about options?
Worth adding to the xml method documentation too.

ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,229 @@ public void Write(IEnumerable<BaseData> source)
...
+
+            var historyBySymbol = new Dictionary<Symbol, List<IGrouping<DateTime, BaseData>>>();
+            var historyBySymbolDailyOrHour = new Dictionary<Symbol, List<BaseData>>();
+
+            foreach (var symbol in symbols)
+            {
+                var exchangeHours = marketHoursDatabase.GetExchangeHours(symbol.ID.Market, symbol, _securityType);
+                var dataTimeZone = marketHoursDatabase.GetDataTimeZone(symbol.ID.Market, symbol, _securityType);
```
Martin-Molinero, line 137: `dataTimeZone` & `exchangeHours` should be shared by all contract symbols under the same Future?

ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,229 @@ public void Write(IEnumerable<BaseData> source)
...
+
+                var historyRequest = new HistoryRequest(
+                    startTimeUtc,
+                    endTimeUtc,
+                    dataType,
+                    symbol,
+                    _resolution,
+                    exchangeHours,
+                    dataTimeZone,
+                    _resolution == Resolution.Tick ? (Resolution?)null : _resolution,
```
Martin-Molinero, line 147: Wonder why null if Tick, maybe worth a comment and maybe even a log to let the user know?

ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,229 @@ public void Write(IEnumerable<BaseData> source)
...
+                    false,
+                    DataNormalizationMode.Raw,
+                    _tickType
+                );
+
+                var history = brokerage.GetHistory(historyRequest)
+                    .Select(
+                        x =>
+                        {
+                            x.Time = x.Time.ConvertTo(exchangeHours.TimeZone, dataTimeZone);
```
Martin-Molinero, line 158: 👍

ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,229 @@ public void Write(IEnumerable<BaseData> source)
...
+            {
+                DownloadAndSaveDailyOrHour(symbols, canonicalSymbol, historyBySymbolDailyOrHour);
+            }
+            else
+            {
+                DownloadAndSaveMinuteOrSecondOrTick(symbols, startTimeUtc, endTimeUtc, canonicalSymbol, historyBySymbol);
+            }
+        }
+
+        private void DownloadAndSaveDailyOrHour(
```
Martin-Molinero, line 185: We already have `WriteDailyOrHour`, wondering what's the difference (worth adding some xml doc) could we reuse part of it or merge them? Same for `WriteMinuteOrSecondOrTick`

ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,229 @@ public void Write(IEnumerable<BaseData> source)
...
+                }
+            }
+
+            if (_resolution == Resolution.Daily || _resolution == Resolution.Hour)
+            {
+                DownloadAndSaveDailyOrHour(symbols, canonicalSymbol, historyBySymbolDailyOrHour);
+            }
+            else
+            {
+                DownloadAndSaveMinuteOrSecondOrTick(symbols, startTimeUtc, endTimeUtc, canonicalSymbol, historyBySymbol);
```
Martin-Molinero, line 181: nit - these methods only are saving the data? so not sure they should be called `DownloadAnd`

ToolBox/LeanDataWriter.cs
```diff
@@ -45,19 +44,18 @@ public class LeanDataWriter
         /// <param name="symbol">Symbol string</param>
         /// <param name="dataDirectory">Base data directory</param>
         /// <param name="resolution">Resolution of the desired output data</param>
-        /// <param name="dataType">Write the data to trade files</param>
-        public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory, TickType dataType = TickType.Trade)
+        /// <param name="tickType">The tick type</param>
+        public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory, TickType tickType = TickType.Trade)
```
Martin-Molinero, line 48: nit - maybe this constructor could call the new one

ToolBox/LeanDataWriter.cs
StefanoRaggi: The difference is the new methods write multiple symbols for the same root ticker in the same zip file (futures/options), the old methods write a single symbol: merging them seems a bit risky for now.

Tests/ToolBox/IBDownloader/IBDataDownloaderTests.cs
```diff
@@ -45,5 +50,40 @@ public void DownloadsFuturesData(string ticker, Resolution resolution, int days)
...
+        {
+            Assert.Throws<ArgumentException>(
+                () =>
+                {
+                    using (var downloader = new IBDataDownloader())
+                    {
+                        var startDate = DateTime.UtcNow.Date.AddDays(-1);
+                        var endDate = DateTime.UtcNow.Date;
+
+                        var symbols = new List<Symbol> { Symbols.Future_ESZ18_Dec2018, Symbols.Future_CLF19_Jan2019 };
```
Martin-Molinero, line 82: Nice tests 👍 !

ToolBox/LeanDataWriter.cs
```diff
@@ -87,6 +100,244 @@ public void Write(IEnumerable<BaseData> source)
...
+            {
+                var historyRequest = new HistoryRequest(
+                    startTimeUtc,
+                    endTimeUtc,
+                    dataType,
+                    symbol,
+                    _resolution,
+                    exchangeHours,
+                    dataTimeZone,
+                    _resolution,
```
Martin-Molinero, line 162: Don't think we want any fill forwarding at all, should we just set it to null?

ToolBox/LeanDataWriter.cs
Martin-Molinero: Would be nice to add as xml docs for the method 👍

PR #4433, opened 2020-05-19
Fixes Second, Tick Resolution history requests for Alpaca

gsalaz98:
#### Description

Fixes history requests with resolutions greater than Minute. Previously, the request would return 0 data points when deploying a live Alpaca instance (non-QC cloud).
#### Related Issue

N/A
#### Motivation and Context

Fixes bug

No

Ran algorithm from master requesting ticks w/ various History methods, and received no data.
AlpacaBrokerageHistoryProviderTests were failing, they're now passing.
#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/Alpaca/AlpacaBrokerage.Utility.cs
+++ b/Brokerages/Alpaca/AlpacaBrokerage.Utility.cs
@@ -277,26 +277,24 @@ namespace QuantConnect.Brokerages.Alpaca
         /// <returns>The list of ticks</returns>
         private IEnumerable<Tick> DownloadTradeTicks(Symbol symbol, DateTime startTimeUtc, DateTime endTimeUtc, DateTimeZone requestedTimeZone)
         {
-            var startTime = startTimeUtc;
+            // The Polygon API only accepts nanosecond level resolution for the expected epoch time.
+            // It is also an inclusive time, so we must increment this by one in order to get the
+            // expected results when paginating.
+            var previousTimestamp = (long?)(DateTimeHelper.GetUnixTimeMilliseconds(startTimeUtc) * 1000000);
 
-            while (startTime < endTimeUtc)
+            while (startTimeUtc < endTimeUtc)
             {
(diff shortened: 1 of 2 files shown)


Brokerages/Alpaca/AlpacaBrokerage.Utility.cs
```diff
@@ -277,26 +277,24 @@ private IEnumerable<TradeBar> DownloadTradeBars(Symbol symbol, DateTime startTim
...
-                var time = startTime;
-                var items = task.SynchronouslyAwaitTaskResult()
-                    .Items
-                    .Where(x => x.Timestamp >= time)
-                    .ToList();
-
-                if (!items.Any())
+                var dateUtc = startTimeUtc.Date;
+                var date = startTimeUtc.ConvertFromUtc(requestedTimeZone).Date;
+                var task = _polygonDataClient.ListHistoricalTradesAsync(new HistoricalRequest(symbol.Value, date)
```
StefanoRaggi, line 291: Wondering if Alpaca is expecting a specific time zone for the `date` argument in the `HistoricalRequest` constructor, maybe UTC? Passing the LEAN exchange time zone to the Alpaca API seems a bit strange to me. 🤔

gsalaz98: I don't know if we can find out. Prior to implementing this, I read through the Polygon docs and didn't find anything related to the expected timezone. 

However, since the market open and close are within the same UTC date in the US, this won't have any impact when querying for USA data.

Martin-Molinero: Since this is the original implementation too, let's keep it but lets please shoot an email to polygon for they to clarify what time zones are these values in: `date`, the `Timestamp`, the returned data itself. From https://polygon.io/docs/ one of the APIs return a flag indicating it's UTC, but does not specify for the rest.

Issue #4196, opened 2020-03-13
Slice dynamic accessors return types


Martin-Molinero:
#### Expected Behavior

- `Slice` dynamic accessors (slice enumeration, `slice[Symbol]`, `slice.Keys`, `slice.Count`, `slice.ContainsKey()`) are deterministic regarding the data type they return and their behavior when it's not present.
This deterministic feature allows maintaining backward compatibility when adding new data types.
#### Actual Behavior

- `Slice` dynamic accessors rely on the order of the provided data, which relies on the order of subscriptions (whether its History requests or `AlgorithManager` data loop). This was uncovered when merging https://github.com/QuantConnect/Lean/pull/3993 that added QuoteBars for equities. Since QuoteBars came later than TradeBars, slice dynamic accessors starting returning QB, breaking algorithms that relied on the data type being returned. 
A first approach to solve this was https://github.com/QuantConnect/Lean/pull/4185, which guaranteed that TB would be returned, but the issue https://github.com/QuantConnect/Lean/issues/4190 uncovered that when setting `FillForward` to false, the non-deterministic issue would still be present since it's not uncommon to have QuoteBars but no TradeBars.
#### Potential Solution

- Add new default data types per security to be used when using slice dynamic accessors:
    - `Slice.ContainsKey()` should return false if the slice does not have the default type for a given Symbol even if other types are present
    - `Slice.Count` should only take into account symbols which have their default data type present
    - `Slice[Symbol]` should only return default data types and throw `KeyNotFoundException ` if not present
    - `Slice.Enumerator` should only enumerate default data types. This will require making sure that existing clients of the enumeration are not taking advantage of the non-deterministic return type. A couple of potential solutions could be: internally to slice, keep a separated generic data bag, used for enumerating in the cases where the data type isn't required (similar to currently existing `slice._data`); replace enumeration for a more precise slice data retrieval using the existing data collections `slice.Bars/QuoteBars/Ticks/Splits/etc`
        1- See `SliceExtension.PushThrough()` enumerating provided slice. Used internally in Lean for consuming History request return slices.
         2- See `AlgorithmManager` warmup enumerating the history requests

     Due to both use cases above (1 & 2), we looked into only applying this deterministic behavior for slices which were generated for the algorithm manager data loop, not history requests, but this was discarded since History requests themselves also suffered from non-deterministic dynamic accessors.

[...]

PR #4439, opened 2020-05-20
Updates Smart Insider transactions and intentions to use different time index

gsalaz98:
#### Description

We'll now use the time at which Smart Insider added a given event to
their database. Because time indexes for Smart Insider are now
non-duplicated per symbol, the History request will properly work.

#### Related Issue

N/A
#### Motivation and Context

Accurate data

No

Reprocessed all data locally.
Tested history request locally in notebook.
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
+++ b/Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
@@ -209,9 +209,8 @@ namespace QuantConnect.Data.Custom.SmartInsider
             {
                 Symbol = config.Symbol
             };
-            // Files are made available at the earliest @ 17:00 U.K. time
-            intention.Time = intention.Time.AddHours(17).ConvertTo(TimeZones.London,config.DataTimeZone);
 
+            intention.Time = intention.Time.ConvertFromUtc(config.DataTimeZone);
             return intention;
         }
 
@@ -281,6 +280,7 @@ namespace QuantConnect.Data.Custom.SmartInsider
         public override string ToLine()
         {
             return string.Join("\t",
+                TimeProcessedUtc?.ToStringInvariant("yyyyMMdd HH:mm:ss"),
                 TransactionID,
                 JsonConvert.SerializeObject(EventType).Replace("\"", ""),
                 LastUpdate.ToStringInvariant("yyyyMMdd"),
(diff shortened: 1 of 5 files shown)


Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
```diff
@@ -209,9 +209,8 @@ public override BaseData Reader(SubscriptionDataConfig config, string line, Date
             {
                 Symbol = config.Symbol
             };
-            // Files are made available at the earliest @ 17:00 U.K. time
-            intention.Time = intention.Time.AddHours(17).ConvertTo(TimeZones.London,config.DataTimeZone);
 
+            intention.Time = intention.Time.ConvertFromUtc(config.DataTimeZone);
```
Martin-Molinero, line 213: Think these conversions (here and in https://github.com/QuantConnect/Lean/pull/4439/files#diff-fac74792ac2784b4b039c950f38c2fa2R241) could be avoided if SmartInsider specified UTC as data time zone? Default is `TimeZones.NewYork`

ToolBox/SmartInsider/SmartInsiderConverter.cs
```diff
@@ -255,7 +255,7 @@ private void WriteToFile<T>(DirectoryInfo destinationDirectory, Dictionary<strin
                 fileContents.AddRange(kvp.Value);
 
                 var tsvContents = fileContents
-                    .OrderBy(x => x.LastUpdate)
+                    .OrderBy(x => x.TimeProcessedUtc.Value)
                     .Select(x => x.ToLine())
                     .Distinct();
```
Martin-Molinero, line 260: The `.Distinct();` call here was the underlying issue since he was dropping data points, is it needed?

gsalaz98: `.Distinct()` only makes sure we don't have any duplicate lines of CSV, it doesn't affect the number of points a `History` call would get back

Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
gsalaz98: Good point, I'll update the PR 👍

Martin-Molinero:
The underlying issue is that the Slice object will only hold the last custom data point per Symbol https://github.com/QuantConnect/Lean/blob/master/Common/Data/Slice.cs#L466. Related to https://github.com/QuantConnect/Lean/issues/4196

PR #4472, opened 2020-05-28
Fixes History Request Data Time Zone

### LINKED ISSUE #4471: Futures' History Request Empty On Data Day Change
#### Expected Behavior
History request is not empty when there is data.

#### Actual Behavior
History requests for Futures between 8 pm and 12 am returns an empty list.

#### Potential Solution
N/A

#### Reproducing the Problem
```python
class BasicTemplateFuturesAlgorithm(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2013, 10, 8)
        self.SetEndDate(2013, 10, 10)
        self.AddFuture(Futures.Indices.SP500EMini).SetFilter(0, 90)

    def OnData(self, slice):
        if self.Time.hour < 21:
            return

        for chain in slice.FutureChains:
            contracts = sorted([x for x in chain.Value if x.OpenInterest > 1000],
                              key=lambda x: x.OpenInterest, reverse=True)

            if len(contracts) == 0: continue
            symbol = contracts[0].Symbol

            if self.History(symbol, 60).empty:
                self.Log(f'{self.Time} :: {symbol.Value}')
```
#### Checklist

AlexCatarino:
#### Description
The tradable days of the history request should respect the data time zone since the data source files also do.

Upgrade `BasicTemplateFuturesHistoryAlgorithm` to a regression algorithm and add a scheduled event to test history requests every hour.

#### Related Issue
Closes #4471

#### Motivation and Context
Bug fix.

No.

New regression tests.
Run algorithm from the community forum, [History Bug when using Futures Contracts](https://www.quantconnect.com/forum/discussion/8238), in QuantConnect Cloud.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/BasicTemplateFuturesHistoryAlgorithm.cs
+++ b/Algorithm.CSharp/BasicTemplateFuturesHistoryAlgorithm.cs
@@ -20,6 +20,9 @@ using QuantConnect.Data;
 using QuantConnect.Orders;
 using QuantConnect.Securities;
 using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Util;
+using QuantConnect.Interfaces;
+using System.Collections.Generic;
 
 namespace QuantConnect.Algorithm.CSharp
 {
@@ -32,7 +35,7 @@ namespace QuantConnect.Algorithm.CSharp
     /// <meta name="tag" content="history and warm up" />
     /// <meta name="tag" content="history" />
     /// <meta name="tag" content="futures" />
-    public class BasicTemplateFuturesHistoryAlgorithm : QCAlgorithm
+    public class BasicTemplateFuturesHistoryAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
     {
         // S&P 500 EMini futures
         private string [] roots = new []
@@ -54,6 +57,17 @@ namespace QuantConnect.Algorithm.CSharp
             }
 
             SetBenchmark(d => 1000000);
+
+            Schedule.On(DateRules.EveryDay(), TimeRules.Every(TimeSpan.FromHours(1)), MakeHistoryCall);
+        }
+
+        private void MakeHistoryCall()
+        {
+            var history = History(10, Resolution.Minute);
+            if (history.IsNullOrEmpty())
+            {
+                throw new Exception($"Empty history at {Time}");
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Engine/HistoricalData/SubscriptionDataReaderHistoryProvider.cs
```diff
@@ -79,6 +79,11 @@ public override IEnumerable<Slice> GetHistory(IEnumerable<HistoryRequest> reques
         /// </summary>
         protected virtual Subscription CreateSubscription(HistoryRequest request, DateTime start, DateTime end)
         {
+            // Tradable dates are defined with the data time zone to access the right source
+            var tradableDates = Time.EachTradeableDay(request.ExchangeHours,
+                start.ConvertFromUtc(request.DataTimeZone),
+                end.ConvertFromUtc(request.DataTimeZone));
+
             // data reader expects these values in local times
             start = start.ConvertFromUtc(request.ExchangeHours.TimeZone);
```
adam-may, line 88: I'd be tempted to create a new variable for start (e.g. exchangeTimeStart) rather than re-using the existing one. This is for 2 reasons - firstly, it gives context to the timezone you're expecting and secondly, it prevents errors such as if these two lines were put above the tradableDates variable creation above.

Engine/HistoricalData/SubscriptionDataReaderHistoryProvider.cs
```diff
@@ -79,6 +79,11 @@ public override IEnumerable<Slice> GetHistory(IEnumerable<HistoryRequest> reques
         /// </summary>
         protected virtual Subscription CreateSubscription(HistoryRequest request, DateTime start, DateTime end)
         {
+            // Tradable dates are defined with the data time zone to access the right source
+            var tradableDates = Time.EachTradeableDay(request.ExchangeHours,
```
Martin-Molinero, line 83: Believe we should use `EachTradeableDayInTimeZone` the same as the data subscription stack, see https://github.com/QuantConnect/Lean/blob/master/Common/Data/UniverseSelection/SubscriptionRequest.cs#L77, which checks the exchange is open in exchange tz, versus here we will be checking the exchange is open in data tz.

Algorithm.CSharp/BasicTemplateFuturesHistoryAlgorithm.cs
```diff
@@ -54,6 +57,17 @@ public override void Initialize()
...
 
             SetBenchmark(d => 1000000);
+
+            Schedule.On(DateRules.EveryDay(), TimeRules.Every(TimeSpan.FromHours(1)), MakeHistoryCall);
+        }
+
+        private void MakeHistoryCall()
+        {
+            var history = History(10, Resolution.Minute);
+            if (history.IsNullOrEmpty())
```
Martin-Molinero, line 67: To be more precise could we assert on the data quantity? Would be nice to have an `OnEndOfAlgorithm` which also asserts the final count or something alike as a safe guard if this method was never called

Engine/HistoricalData/SubscriptionDataReaderHistoryProvider.cs
AlexCatarino: Thank you, @adam-may. I used `startTimeLocal` and `endTimeLocal` to be coherent with naming across Lean where "Local" means "Exchange".

Algorithm.Python/BasicTemplateFuturesHistoryAlgorithm.py
```diff
@@ -46,32 +46,40 @@ def Initialize(self):
...
+        self.successCount = 0
+
+    def MakeHistoryCall(self):
+        history = self.History(self.Securities.keys(), 10, Resolution.Minute)
+        if len(history) < 10:
+            raise Exception(f'Empty history at {self.Time}')
+        self.successCount += 1
+
+    def OnEndOfAlgorithm(self):
+        if self.successCount < 49:
```
Martin-Molinero, line 62: Nice 👍
A nice improvement would be registering with a comment how did this 49 come to be

Issue #3461, opened 2019-07-31
Future-usa-ES Missing Holidays in MHDB


xinweimetrics:
#### Expected Behavior

On 2017/12/31, FrontMonth() should return ES contract with expiry on 2018/03/16.

#### Actual Behavior

On 2017/12/31, FrontMonth() returns ES contract with expiry on 2018/06/15.

#### Potential Solution

`holidays` is empty for this security. If `holidays` are specified, the issue will not happen as backtesting would ignore date 2017/12/31 due to market early close on New Year Eve.

Note: A related issue is that the ES data file is incomplete on 2017/12/31. If information about the ES contract with expiry on 2018/03/16 is included, the issue also would not have happened. I've reported this [missing file issue](https://www.quantconnect.com/forum/discussion/6232/es-minute-the-issue-starts-from-dec-31st-2017-and-continues-until-dec-31st-2017/p1) on the Data Explorer.

#### Reproducing the Problem

The problem is reported in this ticket: https://www.quantconnect.com/forum/discussion/6036/es-frontmonth-changes-often-early-2018/p1 and can be reproduced using the backtest the user attached.

#### System Information

Cloud.

#### Checklist

- [X] I have completely filled out this template
- [X] I have confirmed that this issue exists on the current `master` branch
- [X] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [X] I have provided detailed steps to reproduce the issue

Martin-Molinero:
Note: there are a few futures missing to be completed in the MHDB, not only ES.

PR #4486, opened 2020-06-03
Adds EmaCrossFuturesFrontMonthAlgorithm

AlexCatarino:
#### Description
This example shows how to create an EMA cross algorithm for a futures' front contract. Once the contract is added, the indicators are registered to a new consolidator and warmed up with historical data. When a contract is removed, the consolidator is removed and the indicators are reset. We don't need to liquidate it, because it's liquidated automatically since it has expired.

#### Motivation and Context
Provide an example for common user requests. 

Yes. I can be featured at [Futures' section](https://www.quantconnect.com/docs/data-library/futures).

Locally and in QuantConnect Cloud using 20190101-2020603 date range.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/EmaCrossFuturesFrontMonthAlgorithm.cs
+++ b/Algorithm.CSharp/EmaCrossFuturesFrontMonthAlgorithm.cs
@@ -56,7 +56,7 @@ namespace QuantConnect.Algorithm.CSharp
             // Add a custom chart to track the EMA cross
             var chart = new Chart("EMA Cross");
             chart.AddSeries(new Series("Fast", SeriesType.Line, 0));
-            chart.AddSeries(new Series("Slow", SeriesType.Line, 1));
+            chart.AddSeries(new Series("Slow", SeriesType.Line, 0));
             AddChart(chart);
         }
 
--- a/Algorithm.CSharp/EmaCrossFuturesFrontMonthAlgorithm.cs
+++ b/Algorithm.CSharp/EmaCrossFuturesFrontMonthAlgorithm.cs
@@ -46,6 +46,7 @@ namespace QuantConnect.Algorithm.CSharp
             var future = AddFuture(Futures.Metals.Gold);
 
             // Only consider the front month contract
+            // Update the universe once per day to improve performance
             future.SetFilter(x => x.FrontMonth().OnlyApplyFilterAtMarketOpen());
 
             // Create two exponential moving averages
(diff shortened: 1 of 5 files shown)


Algorithm.CSharp/EmaCrossFuturesFrontMonthAlgorithm.cs
```diff
@@ -0,0 +1,128 @@
...
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 08);
+            SetEndDate(2013, 10, 10);
+            SetCash(1000000);
+
+            var future = AddFuture(Futures.Metals.Gold);
+
+            // Only consider the front month contract
+            future.SetFilter(x => x.FrontMonth().OnlyApplyFilterAtMarketOpen());
```
Martin-Molinero, line 49: Might be worth adding a comment why is `OnlyApplyFilterAtMarketOpen` used? For performance I presume

Algorithm.Python/EmaCrossFuturesFrontMonthAlgorithm.py
```diff
@@ -0,0 +1,104 @@
...
+        self.consolidator = self.ResolveConsolidator(self.symbol, Resolution.Minute)
+        self.RegisterIndicator(self.symbol, self.fast, self.consolidator)
+        self.RegisterIndicator(self.symbol, self.slow, self.consolidator)
+
+        # Update the consolidator with historical data
+        # This will update the registered indicators
+        history = self.History(self.symbol, self.slow.WarmUpPeriod)
+        for index, row in history.iterrows():
+            time = index[2]
+            bar = TradeBar(time, self.symbol, row.open, row.high, row.low, row.close, 0)
```
Martin-Molinero, line 97: We could also use the WarmUpIndicator() helper method API?

Algorithm.Python/QuantConnect.Algorithm.PythonTools.pyproj
```diff
@@ -133,7 +134,6 @@
     <Compile Include="PearsonCorrelationPairsTradingAlphaModelFrameworkAlgorithm.py" />
     <Compile Include="PortfolioRebalanceOnCustomFuncRegressionAlgorithm.py" />
     <Compile Include="PortfolioRebalanceOnDateRulesRegressionAlgorithm.py" />
-    <Compile Include="PsychSignalSentimentRegressionAlgorithm.py" />
```
Martin-Molinero, line 136: 👍

Algorithm.Python/EmaCrossFuturesFrontMonthAlgorithm.py
AlexCatarino: Great idea. It's particular useful in Python to avoid pandas conversion.

AlexCatarino:
> Couple of 0 points in the charts

I have tested the algorithm with `Futures.Indices.SP500EMini` and 2019010-20200605 date range to evaluate whether it was an issue with `WarmUpIndicator()` or a data one. 

We can rule out an issue with `WarmUpIndicator()` because there is only one zero data-point in the chart that corresponds to Jan 1st, 2019. It doesn't happen on Jan 1st, 2020, because there is no security change on that day.

The zero data-point on that day is caused by the missing holiday in the MHDB that is reported in #3461.

Issue #4435, opened 2020-05-19
History request results are inconsistent across multiple methods


gsalaz98:
#### Expected Behavior

History requests all behave alike
#### Actual Behavior

History requests result in different results based on the method called.
Example:

```c#
var spyHistory1 = History<Tick>(spy, prev, now, Resolution.Tick).ToList();
var spyHistory2 = History(new[] { spy }, prev, now, Resolution.Tick).ToList();
var spyHistory3 = History(spy, prev, now, Resolution.Tick).ToList();

Log($"{spyHistory1.Count} :: {spyHistory2.Count} :: {spyHistory3.Count} {spyHistory1.Count == spyHistory2.Count}, {spyHistory1.Count == spyHistory3.Count}");
```

Output:
```
20200519 16:10:22.388 Trace:: Log: 464060 :: 518686 :: 0 False, False
```
#### Potential Solution

N/A - research further

Edit: Standardizing the execution path all History requests take might be worthwhile.
Using the `History(new[] { Symbol }, ...);` path might prove to be flexible enough to
provide the output the other History requests expect by returning `slices.SelectMany(x => x.Get<T>().Values)`

#### Reproducing the Problem

Example algorithm:
```c#
using System;
using System.Linq;
using QuantConnect.Data;
using QuantConnect.Data.Market;

namespace QuantConnect.Algorithm.CSharp
{
    public class BasicTemplateAlgorithm : QCAlgorithm
    {
        public override void Initialize()
        {
            SetStartDate(2013, 10, 8);
            SetEndDate(2013, 10, 9);
            SetCash(100000);

            var spy = AddEquity("SPY", Resolution.Tick).Symbol;
            var now = new DateTime(2013, 10, 9);
            var prev = now - TimeSpan.FromDays(2);

            var spyHistory1 = History<Tick>(spy, prev, now, Resolution.Tick).ToList();
            var spyHistory2 = History(new[] { spy }, prev, now, Resolution.Tick, fillForward: false).ToList();
            var spyHistory3 = History(spy, prev, now, Resolution.Tick).ToList();

            Log($"{spyHistory1.Count} :: {spyHistory2.Count} :: {spyHistory3.Count} {spyHistory1.Count == spyHistory2.Count}, {spyHistory1.Count == spyHistory3.Count}");
        }

        public override void OnData(Slice data)
        {
        }
    }
}
```
#### System Information

Windows 10 x64 - Using default data w/ Lean
#### Checklist

[...]

PR #4497, opened 2020-06-09
Fix Resolution.Tick TradeBar history request

Martin-Molinero:
#### Description

Based on @gsalaz98 findings exposed on issue https://github.com/QuantConnect/Lean/issues/4435:
- Throw exception if requesting `Resolution.Tick` using the TradeBar
  history request API
- Adding unit tests, asserting current behavior

#### Related Issue

Partially resolution for https://github.com/QuantConnect/Lean/issues/4435
#### Motivation and Context

Improve API user experience

N/A

Unit and regression tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -250,8 +250,8 @@ namespace QuantConnect.Algorithm
 
                 var exchange = GetExchangeHours(x);
                 var res = GetResolution(x, resolution);
-                var start = _historyRequestFactory.GetStartTimeAlgoTz(x, periods, res.Value, exchange);
-                return _historyRequestFactory.CreateHistoryRequest(config, start, Time.RoundDown(res.Value.ToTimeSpan()), exchange, res);
+                var start = _historyRequestFactory.GetStartTimeAlgoTz(x, periods, res, exchange);
+                return _historyRequestFactory.CreateHistoryRequest(config, start, Time.RoundDown(res.ToTimeSpan()), exchange, res);
             });
 
             return History(requests.Where(x => x != null)).Get<T>().Memoize();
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.History.cs
```diff
@@ -406,6 +400,12 @@ public IEnumerable<TradeBar> History(Symbol symbol, DateTime start, DateTime end
                 Error("Calling History<TradeBar> method on a Forex or CFD security will return an empty result. Please use the generic version with QuoteBar type parameter.");
             }
 
+            if (resolution.HasValue && resolution == Resolution.Tick)
```
gsalaz98, line 403: If the `resolution` value is null, will this check be bypassed if the only subscription we have for a given `Symbol` is in `Resolution.Tick`? 

We select the highest resolution available in the added Security if it exists in the following method, which is called from `CreateDateRangeHistoryRequests`:
https://github.com/QuantConnect/Lean/blob/a3c5fd98712be4e8416f08bfb548957e9aabcb4a/Algorithm/QCAlgorithm.History.cs#L707

gsalaz98:
Also, I'm not sure this PR closes the issue #4435 since `History<Tick>(...)` is still only adding the `TickType.Quote` subscription rather than both `TickType.Trade` and `TickType.Quote`, compared to the `History(Symbol[], ...)` call. I've left a comment on the original issue detailing my findings 👍

Algorithm/QCAlgorithm.History.cs
Martin-Molinero: Nice catch! Addressed 👍

Martin-Molinero:
> Also, I'm not sure this PR closes the issue #4435 since `History<Tick>(...)` is still only adding the `TickType.Quote` subscription rather than both `TickType.Trade` and `TickType.Quote`, compared to the `History(Symbol[], ...)` call. I've left a comment on the original issue detailing my findings 👍

True, the difference is that `History(Symbol[], ...)` returns a `Slice` object 🤔. If we were to add both tick types in the result, it would change existing behavior and would end up as a mix of tick types, sometimes Trade, sometimes Quote, somewhat related to issue https://github.com/QuantConnect/Lean/issues/4196.
Another case I though of while reviewing is also related to 4196:

``` CSharp
// Trades and quotes
var result = _algorithm.History(new [] { Symbols.SPY }, start.AddHours(9.8), start.AddHours(10), Resolution.Tick).ToList();
foreach (var slice in result)
{
    if (slice.Ticks.Any(pair => pair.Value.Any(tick => tick.TickType == TickType.Quote) &&
                                pair.Value.Any(tick => tick.TickType == TickType.Trade)))
    {
        // slice.Ticks has both Quote and Trade ticks, what will the following return?
        var ticks = slice.Get<Tick>();
    }
}
```
Let's keep issue #4435 open till we define the expected behavior 👍

Issue #4522, opened 2020-06-19
Previous Quote Bar Close is not equal to Current Quote bar Open when using TickQuoteBarConsolidator


AnandVishnu:
#### Expected Behavior
For quote bar previous close should equal to current bar open when using TickQuoteBarConsolidator

#### Actual Behavior

TickQuoteBarConsolidator creates a new bar every period. So open will take the first value in the consolidated period

#### Potential Solution

Store the previous tick. when ever a new working bar is created, initialise it with previous tick

        protected override void AggregateBar(ref QuoteBar workingBar, Tick data)
        {
            if (workingBar == null)
            {
                workingBar = new QuoteBar
                {
                    Symbol = data.Symbol,
                    Time = GetRoundedBarTime(data.Time),
                    Bid = null,
                    Ask = null
                };

                // sync last tick  close price with this tick open price
                if (_previousTick != null)
                {
                    workingBar.Update(0, _previousTick.BidPrice, _previousTick.AskPrice, 0, _previousTick.BidSize, _previousTick.AskSize);
                }

                if (Period.HasValue) workingBar.Period = Period.Value;
            }

            // update the bid and ask
            workingBar.Update(0, data.BidPrice, data.AskPrice, 0, data.BidSize, data.AskSize);
            if (!Period.HasValue) workingBar.EndTime = GetRoundedBarTime(data.EndTime);
            _previousTick = (Tick) data.Clone();
        }

#### Reproducing the Problem

Following unit test will fail

        public void LastCloseAndCurrentOpenPriceShouldBeSameConsolidatedOnCount()
        {
            QuoteBar quoteBar = null;
            var creator = new TickQuoteBarConsolidator(1);
            creator.DataConsolidated += (sender, args) =>
            {
                quoteBar = args;
            };

            var reference = DateTime.Today;
            var tick1 = new Tick
            {
                Symbol = Symbols.SPY,
                Time = reference,
                TickType = TickType.Quote,
                AskPrice = 25,
                BidPrice = 24,

            };

            creator.Update(tick1);

            var tick2 = new Tick
            {
                Symbol = Symbols.SPY,
                Time = reference.AddSeconds(1),
                TickType = TickType.Quote,
                AskPrice = 36,
                BidPrice = 35,
            };

            creator.Update(tick2);

            Assert.AreEqual(tick1.AskPrice, quoteBar.Ask.Open, "Ask Open not equal to Previous Close");
[...]

PR #4523, opened 2020-06-19
Quote bar previous close and current open should be same

AnandVishnu:
For quote bar previous close should equal to current bar open when using TickQuoteBarConsolidator

#### Description

Store the previous tick. when ever a new working bar is created, initialise it with previous tick

#### Related Issue

 https://github.com/QuantConnect/Lean/issues/4522

#### Motivation and Context

This change is required so that correct return series could be constructed when using Future Rolling contracts data.

#### Requires Documentation Change

No

#### How Has This Been Tested?

Yes. Added unit tests for the same

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [X] I have added tests to cover my changes. 
- [X] All new and existing tests passed.
- [X] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #4531, opened 2020-06-23
Previous QuoteBar Close is not equal to Current QuoteBar Open

Martin-Molinero:
#### Description

- On top of PR https://github.com/QuantConnect/Lean/pull/4523, thank you @AnandVishnu: Refactor to use `Consolidated` as source of the previous close. This is because using previous Tick might not be enough because ticks could only be a bid or ask update, with the other side being 0 for that tick (no update). Updating unit tests
- Applying fix for `QuoteBarConsolidator` and `QuoteBarBuilderEnumerator` as well. Adding unit tests
#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/4522
#### Motivation and Context

- Quotes are valid until a new quote arrives

N/A

Unit and regression tests
#### Types of changes

#### Checklist:


jaredbroad:
This solves the quote bar consistency between bars; I think we need something in LEAN for the fill forward of previous bars (ones built with a genuine new tick), and something here to not write the bars to disk if it's a fill-forward bar. I don't know if it can be merged without all three parts handled. 

If these two things are covered here and I missed it we should expand the comments.

PR #4541, opened 2020-06-26
Heikin-Ashi Indicator Fix for Volume Information

JonathanWheeler:
#### Expected Behavior
```csharp
Debug(_HA4Hour.CurrentBar.ToString());
```
should output:
```
BTCUSD: O: 7917.1081 H: 7957 L: 7832.03 C: 7900.44 V: 1230.1984
```

#### Actual Behavior
Outputs:
```
0: O: 7917.1081 H: 7957 L: 7832.03 C: 7900.44 V: 0
```

#### Potential Solution
Update `HeikinAshi.ComputeNextValue(IBaseDataBar input)` to:  
1. Always store `input.Symbol`.
2. When `input` is a `TradeBar` or `RenkoBar`, also store `input.Volume`. (`QuoteBar` does not have Volume information.)

#### Reproducing the Problem
```csharp
using System;
using QuantConnect.Data;
using QuantConnect.Data.Market;
using QuantConnect.Indicators;

namespace QuantConnect.Algorithm.CSharp.Skylight
{
    public class TradingAlgorithm : QCAlgorithm
    {
        private static readonly string _marketGDAX = Market.GDAX;
        private static readonly string _tickerBTCUSD = "BTCUSD";
        private static readonly Symbol _symbolBTCUSD = QuantConnect.Symbol.Create(_tickerBTCUSD, SecurityType.Crypto, _marketGDAX);

        private readonly HeikinAshi _HA4Hour = new HeikinAshi($"HA({_symbolBTCUSD}_4hr)");
        private readonly RollingWindow<TradeBar> _HA4HourRollingWindow = new RollingWindow<TradeBar>(2);

        {
            SetStartDate(new DateTime(2020, 06, 01, 00, 00, 0);
            SetEndDate(new DateTime(2020, 06, 03, 00, 00, 0));
            SetCash(100000);
            AddCrypto(_symbolBTCUSD, Resolution.Hour, _marketGDAX);
            Consolidate<TradeBar>(_symbolBTCUSD, TimeSpan.FromHours(4), Consolidated4HourTradeBarHandler);
        }

        {
        }

        private void Consolidated4HourTradeBarHandler(TradeBar consolidated)
        {
            _HA4Hour.Update(consolidated); // Symbol and Volume information is lost at this step.
            _HA4HourRollingWindow.Add(_HA4Hour.CurrentBar);

            Debug(_HA4Hour.CurrentBar.ToString()); // Symbol and Volume information is lost.
            
            if (_HA4HourRollingWindow[0].Close > _HA4HourRollingWindow[1].Close && !Portfolio.Invested)
            {
                Debug($" BUY: {_HA4HourRollingWindow[0]}"); // Symbol and Volume information is still lost.
                SetHoldings(_symbolBTCUSD, 1);
            }
            else if (_HA4HourRollingWindow[0].Close < _HA4HourRollingWindow[1].Close && Portfolio.Invested)
            {
                Debug($"SELL: {_HA4HourRollingWindow[0]}"); // Symbol and Volume information is still lost.
                SetHoldings(_symbolBTCUSD, 0);
            }
        }
    }
}
```

#### System Information
macOS 10.15.5  
JetBrains Rider 2020.1.3  
C# 6.0  
.NETFramework v4.6.2  
Mono v6.10.0.105

#### Checklist


JonathanWheeler:
N.B.: This is my first pull request on GitHub. If I made any errors or faux pas, I welcome feedback and constructive criticism!

jaredbroad:
Nice PR @JonathanWheeler =) We'll review and get back to you shortly

JonathanWheeler:
(Travis CI seems to have failed for this build for an unrelated reason; it passed in a prior build. Closing and reopening this issue to trigger a rebuild.)

Indicators/HeikinAshi.cs
```diff
@@ -112,6 +117,16 @@ protected override decimal ComputeNextValue(IBaseDataBar input)
                 Low.Update(input.Time, Math.Min(input.Low, Math.Min(Open, Close)));
             }
             
+            // IBaseDataBar subtypes TradeBar and RenkoBar have a Volume property, but QuoteBar does not.
+            var volume = 0.0m;
+            if (input is TradeBar)
+                volume = ((TradeBar) input).Volume;
+            else if (input is RenkoBar)
+                volume = ((RenkoBar) input).Volume;
```
Martin-Molinero, line 125: Nice fix! 👍 
nit - in Lean the coding style is usually to add all brackets, so for this case would be:
```
if (input is TradeBar)
{
    volume = ((TradeBar) input).Volume;
}
else if (input is RenkoBar)
{
    volume = ((RenkoBar) input).Volume;
}

```

Indicators/HeikinAshi.cs
```diff
@@ -53,10 +53,15 @@ public class HeikinAshi : BarIndicator, IIndicatorWarmUpPeriodProvider
...
+        /// <summary>
+        /// Gets the Heikin-Ashi Symbol
+        /// </summary>
+        public Symbol Symbol { get; private set; }
+
         /// <summary>
         /// Gets the Heikin-Ashi current TradeBar
         /// </summary>
-        public TradeBar CurrentBar => new TradeBar(Open.Current.Time, Symbol.Empty, Open, High, Low, Close, Volume);
+        public TradeBar CurrentBar => new TradeBar(Open.Current.Time, Symbol, Open, High, Low, Close, Volume);
```
Martin-Molinero, line 64: I think I'd be more inclined to actually remove this `CurrentBar/Symbol` property entirely, think this is code that remains from when `HeikinAshi` only supported `TradeBar` input, updated at https://github.com/QuantConnect/Lean/commit/f4ec05c13cca5e16dc0ed8d0a8b169a5f10182e6. 
If the input being used it QuoteBars/RenkoBars, this TradeBar CurrentBar doesn't match? Also I think this isn't part of the Indicator base clases or interfaces nor is there any other indicator with `public TradeBar CurrentBar`

Indicators/HeikinAshi.cs
JonathanWheeler: Thanks! I 100% support nit-picking for consistent syntax! Nothing makes me twitch more when reading code, haha. 😄 I'll revise. 🙂

Indicators/HeikinAshi.cs
JonathanWheeler: Hey Martin! Thank you for your feedback. 🙂 👍 

Based on your feedback, I removed references to ```Symbol``` and ```CurrentBar```. Since I can no longer refer to ```_HA4Hour.CurrentBar``` in my example code above, I made the following changes:
```CSharp
// Changed <TradeBar> to <HeikinAshi>:
private readonly RollingWindow<HeikinAshi> _HA4HourRollingWindow = new RollingWindow<HeikinAshi>(2);

...

// Changed _HA4Hour.CurrentBar to just _HA4Hour:
_HA4HourRollingWindow.Add(_HA4Hour);
```
However, my example algorithm suddenly broke, performing zero trades.

After some debugging, I now understand why. Although I'm still updating ```_HA4Hour``` using ```_HA4Hour.Update(consolidated);``` (which is computed and updated correctly as expected), ```_HA4Hour``` still refers to the same reference when calling ```_HA4HourRollingWindow.Add(_HA4Hour);```, thus effectively causing all ```_HA4HourRollingWindow``` values to be "updated" as well. In reality, all ```RollingWindow``` values are just referring to the same updated object.

I remember encountering a similar problem in Python using QuantConnect's web interface. Since I couldn't properly debug it, I never figured it out! However, now I understand why.

My expectation in both (Python & C#) cases was that ```RollingWindow.Add(...)``` would automatically make a copy of the object passed to it if needed. However, upon further consideration, I understand that doing so isn't so straightforward, for a variety of reasons. In any case, this isn't very intuitive.

What's the best way of resolving this? Would it be appropriate to add a copy constructor to ```HeikinAshi```, or something similar? This would allow me to refactor my code to: ```_HA4HourRollingWindow.Add(new HeikinAshi(_HA4Hour));```

Martin-Molinero: Hey! I think what you need is to directly pass the last consolidated value into the rolling window, for example:
```c#
var rollingWindow = new RollingWindow<IndicatorDataPoint>(2);
var heikinAshi = HeikinAshi("SPY");
heikinAshi.Updated += (sender, updated) => rollingWindow.Add(updated);
```
Complete example [C#](https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/RollingWindowAlgorithm.cs), [Py](https://github.com/QuantConnect/Lean/blob/master/Algorithm.Python/RollingWindowAlgorithm.py). Was this what you were looking for?

>Based on your feedback, I removed references to Symbol and CurrentBar

Sounds good 👍

JonathanWheeler: Hi Martin! Thank you for the suggestion. 🙂 

I tried this, but this won't work. I'm writing an algorithm that uses Heikin-Ashi candles as a trade signal, and more specifically on whether the Heikin-Ashi ```Open``` and ```Close``` values are expanding or contracting in size. Since ```IndicatorDataPoint.Value``` only stores ```Close```, this won't work.

(Side note: I'm using this technique on many different simultaneous consolidation periods and with long rolling windows, so I'm trying to keep my code as clean as possible. This will quickly become a mess otherwise. This is why I care so much.)

The solutions I can think of are:
1. Adding a copy constructor, as mentioned above; or
2. Creating a ```CustomHeikinAshi``` subclass as a client-side file, and placing a copy constructor there. This would require that ```protected set;``` is added to ```HeikinAshi```'s properties; or
3. Re-introduce ```CurrentBar``` and ```Symbol```. (On this point, I'd argue that Heikin-Ashi candles are unique in that they have a "is a" relationship with both TradeBar and Indicator; Heikin-Ashi candles can be used as a substitute for traditional candles on many platforms, such as TradingView.)

What do you think? I'm open to other suggestions as well.

JonathanWheeler: Upon further consideration, solutions 1 and 2 don't really resolve this problem either, because then the same "duplicate reference" problem gets pushed to ```IndicatorBase``` for ```Open```, ```High```, ```Low```, ```Close```, and ```Volume```. (I just tried it.) I think solution 3 may be the only viable option?

JonathanWheeler: Hi @Martin-Molinero! Would it make sense to make QuantConnect.Algorithm.CSharp (and .Python) reference QuantConnect.Algorithm.Indicator? This would allow indicators to be subclassed. This would allow me to override ```ComputeNextValue(...)```, as well as add ```CurrentBar()``` as an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) in my client-side code instead.

Martin-Molinero: Hey @JonathanWheeler ! Sorry for the late response.

>Thank you for the suggestion. 🙂 I tried this, but this won't work.

Would some of these options work for you?
```
var rollingWindow = new RollingWindow<Bar>(2);
heikinAshi.Updated += (sender, updated) =>
{
    // we generate a new bar with current values
    rollingWindow.Add(new Bar(heikinAshi.Open, heikinAshi.High, heikinAshi.Low, heikinAshi.Close));
};

var rollingWindowTradeBar = new RollingWindow<TradeBar>(2);
heikinAshi.Updated += (sender, updated) =>
{
    // we generate a new bar with current values
    rollingWindowTradeBar.Add(new TradeBar(heikinAshi.Open.Current.Time, "SPY", heikinAshi.Open, heikinAshi.High, heikinAshi.Low, heikinAshi.Close, heikinAshi.Volume));
};
```

JonathanWheeler: Hey @Martin-Molinero! No worries!

Thank you for the suggestions. I don't think these options will work, because I need the separate EMA values of each ```TradeBar```'s OHLC value, and then throw these over to ```HeikinAshi``` (by putting them into a temporary ```TradeBar```, and then passing this to ```HeikinAshi.Update```). It was more verbose than I would have liked, but I was able to accomplish what I needed.

I went ahead and incorporated your comments into my latest changes, removing ```CurrentBar```, reverting my ```Symbol``` addition, and only adding ```Volume``` to ```HeikinAshi```. How's it look? 🙂

JonathanWheeler: Effect of the change:

```C#
Debug(_HA4Hour.CurrentBar.ToString());
```
Previously outputs:
```C#
0: O: 7917.1081 H: 7957 L: 7832.03 C: 7900.44 V: 0
```
Now outputs:
```C#
0: O: 7917.1081 H: 7957 L: 7832.03 C: 7900.44 V: 1230.1984
```

PR #4546, opened 2020-06-29
Update fill price modeling for stop market orders in FillModel

### LINKED ISSUE #4545: ImmediateFillModel can have unrealistic fill prices for stop market orders
#### Current stop market fill price logic, for reference:
For a buy stop market order (vice versa for sell stops), fill price is the maximum of:
1. Stop price (no slippage applied)
2. Current bar's close + slippage

### Scenario 1
1. Buy stop market order is entered with stop price of 10
2. Current bar opens at 8 and closes at 12, triggering the stop order to be filled

**Expected:** Fill price = 10 + slippage (because the price opened below the stop price and presumably moved through it during the current bar's time period)

**Actual:** Fill price = 12 + slippage (this introduces 20% additional slippage, on top of the existing slippage model, that likely would not have been realized)

### Scenario 2
1. Buy stop market order is entered with stop price of 10
2. Current bar opens at 12 and closes at 8, triggering the stop order to be filled

**Expected:** Fill price = 12 + slippage (because the bar "gap" opened higher than the stop price, which would trigger the stop to be filled at the open)

**Actual:** Fill price = 10 (in reality this would have filled around 12 since that was the open price when the order triggered, so this is effectively introducing negative slippage, in addition to not applying the actual slippage model)

### Potential Solution
Change stop market fill price logic to:
1. If the current bar "passes through" the stop price, fill price = stop price +- slippage
2. If the current bar "gaps" open beyond the stop price, fill price = open price +- slippage

I believe this fill price logic models reality more closely than the existing logic. And it could reduce confusion on where slippage is coming from. When I was initially seeing strange differences between fill and stop prices I assumed it was the slippage model. So I implemented a custom slippage model to reduce it to 0. To my surprise, this had no impact on the large slippage figures I was seeing on stop orders when the price was clearly moving through the stop prices.

One potential issue with this logic is long-interval bars that have price "gaps" within them, e.g., weekly or monthly bars. In this case it's not always safe to assume the price passed through the stop price. But I think making that assumption is, on average, much more realistic than taking the close price or stop price at the end of the bar interval.

### Reproducing the Problem
See examples above.

### Checklist

bhamer:
#### Description
- Updated fill price modeling for stop market orders in FillModel.cs
- Updated relevant unit tests to highlight expected behavior

#### Related Issue
Closes #4545

#### Motivation and Context
Generate more realistic fill prices for stop market orders when backtesting.

N/A

Updated the relevant unit tests for ImmediateFillModel. I debated breaking the scenarios out into separate tests but I ended up leaving them together to keep it consistent with the rest of the order type tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
+++ b/Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
@@ -284,7 +284,7 @@ namespace QuantConnect.Tests.Common.Orders.Fills
             );
             security.SetLocalTimeKeeper(TimeKeeper.GetLocalTimeKeeper(TimeZones.NewYork));
             var slippageModel = new ConstantSlippageModel(.01m);
-            security.SetSlippageModel(new ConstantSlippageModel(.01m));
+            security.SetSlippageModel(slippageModel);
 
             // scenario 1: price doesn't trigger stop market fill
             security.SetMarketPrice(new IndicatorDataPoint(Symbols.SPY, Noon, 101m));
@@ -338,7 +338,7 @@ namespace QuantConnect.Tests.Common.Orders.Fills
             );
             security.SetLocalTimeKeeper(TimeKeeper.GetLocalTimeKeeper(TimeZones.NewYork));
             var slippageModel = new ConstantSlippageModel(.01m);
-            security.SetSlippageModel(new ConstantSlippageModel(.01m));
+            security.SetSlippageModel(slippageModel);
 
             // scenario 1: price doesn't trigger stop market fill
             security.SetMarketPrice(new IndicatorDataPoint(Symbols.SPY, Noon, 102m));
--- a/Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
+++ b/Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
@@ -21,6 +21,7 @@ using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
 using QuantConnect.Orders;
 using QuantConnect.Orders.Fills;
+using QuantConnect.Orders.Slippage;
 using QuantConnect.Securities;
 using QuantConnect.Securities.Forex;
 using QuantConnect.Tests.Common.Data;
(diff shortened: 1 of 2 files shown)


jaredbroad:
Hi @bhamer interesting PR thank you. I guess a tangible example of this would be a gap down overnight and passing the stop price of the opening bar?

bhamer:
Hi @jaredbroad, the general case is a bar that triggers the stop and where there's a meaningful difference between the open price, close price, and stop price. A gap up/down isn't necessary, but it is a common case.

So using your example of a gap down past the sell stop price, in reality the fill would occur at the open but the current fill model is effectively filling it at the close. This matters when the price has moved meaningfully between the open and close time for that particular bar.

Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
```diff
@@ -282,29 +283,41 @@ public void PerformsStopMarketFillBuy()
                 new SecurityCache()
             );
             security.SetLocalTimeKeeper(TimeKeeper.GetLocalTimeKeeper(TimeZones.NewYork));
-            security.SetMarketPrice(new IndicatorDataPoint(Symbols.SPY, Noon, 101m));
+            var slippageModel = new ConstantSlippageModel(.01m);
+            security.SetSlippageModel(new ConstantSlippageModel(.01m));
```
AlexCatarino, line 287: We don't need to create two instances of `ConstantSlippageModel`:
```csharp
security.SetSlippageModel(slippageModel);
```
Same [here](https://github.com/QuantConnect/Lean/pull/4546/files#diff-b29d6b609cdd690136095fe84443a4a5R341).

AlexCatarino:
@bhamer, should we not expect the same behavior for limit orders and stop-limit orders?

bhamer:
@AlexCatarino I think limit fills (with or without a stop trigger) can be improved using a similar approach, i.e., utilizing the fact that we know open price came before HLC prices. But I view the limit fills separately since they'll have different logic for determining the fill price than market fills, which is what this is.

It's a good question though, they don't necessarily need to be viewed separately. And you prompted a thought that it could make sense to refactor a bit and break out the fill logic (market, limit) from the trigger logic (stop, MOO, MOC).

Thoughts? If the team agrees with the approach to stop market orders I can put together a more holistic PR that includes the other order types.

Tests/Common/Orders/Fills/ImmediateFillModelTests.cs
bhamer: agree

bhamer:
Adding to the previous comment - 
I think it's more important to update stop-market fills because they're _usually_ filled at an unrealistic price, i.e., it's rare that a stop market order would fill at exactly the stop price (no slippage) or the close price plus slippage of the bar that triggered it. Whereas limit fills are _usually_ filled at the limit price.

PR #4554, opened 2020-07-01
Adds Validation for barSize Parameter of RenkoConsolidator Constructor

### LINKED ISSUE #4553: RenkoConsolidator Division By Zero
#### Expected Behavior
Do not allow setting RenkoConsolidator.BarSize to zero.

#### Actual Behavior
It's possible to set `RenkoConsolidator.BarSize`. It causes unhandled `DivideByZeroException` in `RenkoConsolidator.UpdateClassic`

#### Potential Solution
Thrown `ArgumentException` in `RenkoConsolidator` construction.

#### Reproducing the Problem
```python
renkoConsolidator = RenkoConsolidator(round(self.atr.Current.Value,4))
```
where 
`round(self.atr.Current.Value,4)` is zero.

#### Checklist

AlexCatarino:
#### Description
Adds validation for `barSize` parameter of `RenkoConsolidator` constructor since `RenkoConsolidator.BarSize` cannot be negative or zero.

#### Related Issue
Closes #4553 

#### Motivation and Context
Without this validation, available in some constructor overloads, the consolidator throws an unhandled exception `DivideByZeroException`. 

New unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Consolidators/RenkoConsolidator.cs
+++ b/Common/Data/Consolidators/RenkoConsolidator.cs
@@ -85,18 +85,9 @@ namespace QuantConnect.Data.Consolidators
         /// <param name="barSize">The constant value size of each bar</param>
         /// <param name="evenBars">When true bar open/close will be a multiple of the barSize</param>
         public RenkoConsolidator(decimal barSize, bool evenBars = true)
+            : this(barSize, x => x.Value, x => 0, evenBars)
         {
-            if (barSize < Extensions.GetDecimalEpsilon())
-            {
-                throw new ArgumentOutOfRangeException(nameof(barSize), "RenkoConsolidator bar size must be positve and greater than 1e-28");
-            }
 
-            _barSize = barSize;
-            _selector = x => x.Value;
-            _volumeSelector = x => 0;
-            _evenBars = evenBars;
-
-            Type = RenkoType.Classic;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -68,6 +68,11 @@ public RenkoConsolidator(decimal barSize, RenkoType type)
                 throw new ArgumentException($"RenkoConsolidator can only be initialized with RenkoType.Wicked. For RenkoType.Classic, please use the other constructor overloads.");
             }
 
+            if (barSize < Extensions.GetDecimalEpsilon())
+            {
+                throw new ArgumentOutOfRangeException(nameof(barSize), "RenkoConsolidator bar size must be positve and greater than 1e-28");
```
Martin-Molinero, line 73: Cool! 
Could we have the constructors call each other? With these smart constructors, I think it's a nice approach to have them call each other to avoid having to duplicate this logic so many times, which is error-prone too and actually lead to this issue

Tests/Common/Data/RenkoConsolidatorTests.cs
```diff
@@ -737,5 +737,24 @@ public void WickedTwoRisingThenThreeFallingGapRenkos()
             Assert.AreEqual(openRenko.Low, 7.9);
             Assert.AreEqual(openRenko.Close, 7.9);
         }
+
+        [Test]
+        public void WickedArgumentExceptionRenkoTypeClassicTest()
+        {
+            Assert.Throws<ArgumentException>(() => new RenkoConsolidator(2.5m, RenkoType.Classic));
```
jaredbroad, line 744: This API choice concerns me; I think it would be better to split this into a class per Renko or rethink the API for this class so the user can't accidentally cause an explosion by putting a perfectly valid looking arg into the constructor.

jaredbroad: This whole class is a candidate for deprecation and replacement with two separate classes; maybe making the old class inherit from its replacement to maintain backward compatibility.

Martin-Molinero: Agree 👍, if `RenkoType` can only have 1 value then it doesn't seem like an argument at all

Common/Data/Consolidators/RenkoConsolidator.cs
mchandschuh: 1e-28 is super small -- I'm guessing an out of range exception brought us here. In my own checks I'll usually limit it to 1e-20 to avoid divisions pushing me out of range.

jaredbroad:
Closing pending @aarjaneiro refactor

PR #4569, opened 2020-07-09
Allow LeanDataWriter to create zip files for futures data

michael-sena:
`LeanDataWriter` is able to create data files in the expected QC format for all symbols, except futures where multiple contracts exist in the file.  This is a format supported by the `LeanDataReader`, but there isn't a writer equivalent.

#### Description
The approach I took was to add a `boolean` value that controls whether a file should be appended to or not.
If the ZIP entry exists, it will be replaced in its entirity.
However, any existing entries will be maintained.

This may not be the best approach, and there are potential issues - such as retention of stale data.
I'd be happy to hear suggestions on how it can be improved / a better approach.

I've also included 2 fixes that I found while trying to use this class.
* I couldn't override the output location when using the end-to-end retrieve & write function.  It always wrote to the global data directory.
* Data wasn't grouped if the input had a time portion attached.

#### Motivation and Context
Unable to generate futures data zips without any custom code, as the file was always overridden.

N/A

I have included unit tests for the futures change.
I also have tested it locally, generating ~5 years of data across a set of futures, with each output file containing multiple contracts.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/ToolBox/LeanDataWriterTests.cs
+++ b/Tests/ToolBox/LeanDataWriterTests.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using NUnit.Framework;
 using QuantConnect.Data.Market;
+using QuantConnect.Securities;
 using QuantConnect.ToolBox;
 using QuantConnect.Util;
 
@@ -64,6 +65,28 @@ namespace QuantConnect.Tests.ToolBox
             var data = QuantConnect.Compression.Unzip(filePath);
 
             Assert.AreEqual(data.First().Value.Count(), 3);
+        }
+
+        [Test]
+        public void LeanDataWriter_CanWriteFutureWithMultipleContracts()
+        {
+            var contract1 = Symbol.CreateFuture(Futures.Indices.SP500EMini, Market.CME, new DateTime(2020, 02, 01));
+            var filePath1 = LeanData.GenerateZipFilePath(_dataDirectory, contract1, _date, Resolution.Second, TickType.Quote);
+            var leanDataWriter1 = new LeanDataWriter(Resolution.Second, contract1, _dataDirectory, TickType.Quote);
+            leanDataWriter1.Write(GetQuoteBars(contract1));
+
+            var contract2 = Symbol.CreateFuture(Futures.Indices.SP500EMini, Market.CME, new DateTime(2020, 03, 01));
+            var filePath2 = LeanData.GenerateZipFilePath(_dataDirectory, contract2, _date, Resolution.Second, TickType.Quote);
+            var leanDataWriter2 = new LeanDataWriter(Resolution.Second, contract2, _dataDirectory, TickType.Quote);
+            leanDataWriter2.Write(GetQuoteBars(contract2));
+
+            Assert.AreEqual(filePath1, filePath2);
+            Assert.IsTrue(File.Exists(filePath1));
+            Assert.IsFalse(File.Exists(filePath1 + ".tmp"));
+
+            var data = QuantConnect.Compression.Unzip(filePath1).ToDictionary(x => x.Key, x => x.Value.ToList());
+            Assert.AreEqual(2, data.Count);
+            Assert.That(data.Values, Has.All.Count.EqualTo(3));
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


ToolBox/LeanDataWriter.cs
```diff
@@ -45,13 +46,15 @@ public class LeanDataWriter
         /// <param name="dataDirectory">Base data directory</param>
         /// <param name="resolution">Resolution of the desired output data</param>
         /// <param name="tickType">The tick type</param>
-        public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory, TickType tickType = TickType.Trade)
+        /// <param name="appendToZips">Append to existing zip rather than replace them.  Entries will still be replaced entirely.</param>
+        public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory, TickType tickType = TickType.Trade, bool appendToZips = false)
```
Jay-Jay-D, line 50: The new argument `appendToZips` will be used only for Futures, right?

If so, I suggest to remove it as argument and define its value based on `_securityType`.

Tests/ToolBox/LeanDataWriterTests.cs
```diff
@@ -66,6 +67,28 @@ public void LeanDataWriter_CanWriteForex()
             Assert.AreEqual(data.First().Value.Count(), 3);
         }
 
+        [Test]
+        public void LeanDataWriter_CanWriteFutureWithMultipleContracts()
```
Jay-Jay-D, line 71: Lovely test!

ToolBox/LeanDataWriter.cs
michael-sena: @Jay-Jay-D - Yes, I wondered about that too.  My thinking is that I wanted to allow the overwriting functionality to still be possible, regardless of the security type.  But it could still be valuable to default to that for futures.

I've made a change that allows that value to be defaulted to the security type provided.  That way it would be possible to create Lean archives without appending if so desired.

I'm happy to get rid of the optional input entirely and base it solely on security type if you think that's better.

Jay-Jay-D: Yes, please remove the optional argument, it'll make it less error-prone.

michael-sena: @Jay-Jay-D - I've removed the optional argument in favour of deriving `appendToZips` from the trade type.

I also called into a common constructor so that `appendToZips` logic didn't exist in 2 places.

ToolBox/LeanDataWriter.cs
```diff
@@ -45,13 +46,14 @@ public class LeanDataWriter
         /// <param name="dataDirectory">Base data directory</param>
         /// <param name="resolution">Resolution of the desired output data</param>
         /// <param name="tickType">The tick type</param>
-        public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory, TickType tickType = TickType.Trade)
+        public LeanDataWriter(Resolution resolution, Symbol symbol, string dataDirectory, TickType tickType = TickType.Trade) : this(
```
Martin-Molinero, line 49: `this :` beautiful 👍

ToolBox/LeanDataWriter.cs
```diff
@@ -472,13 +481,21 @@ private void WriteFile(string filePath, string data, DateTime date)
...
 
-            // Write out this data string to a zip file
-            Compression.Zip(data, tempFilePath, LeanData.GenerateZipEntryName(_symbol, date, _resolution, _tickType));
-
-            // Move temp file to the final destination with the appropriate name
-            File.Move(tempFilePath, filePath);
+            if (_appendToZips)
+            {
+                var entryName = LeanData.GenerateZipEntryName(_symbol, date, _resolution, _tickType);
+                Compression.ZipCreateAppendData(filePath, entryName, data);
```
Martin-Molinero, line 487: `Compression.ZipCreateAppendData` won't update the entry if it already exists and log an error and return `false`. Believe this can happen if calling `Write(IEnumerable<BaseData> source)`, they don't seem to be removing the entry if it exists as `DownloadAndSave()` is, this seems in part to the code duplication we have here 😅  - Maybe we can add an override argument to `Compression.ZipCreateAppendData`. Thoughts @Jay-Jay-D

PR #4584, opened 2020-07-21
Bitfinex Brokerage updates

StefanoRaggi:
#### Description
- Upgrade Bitfinex API to v2 (current stable and recommended version)
- Order submit/cancel/update using WebSocket instead of REST (to prevent delays caused by rate limits)
- Removed URL config settings for WebSocket and REST API endpoints (for easier updating in the future)
- Bugfix: no longer emits trade ticks on trades snapshot message

#### Related Issue
n/a

#### Motivation and Context
- Live trading stability

No.

- Brokerage unit tests (all green)
- Live trading locally + QC cloud
- Tested streaming for all symbols in QC backend application

#### Types of changes

#### Checklist:


Brokerages/Bitfinex/BitfinexSubscriptionManager.cs
```diff
@@ -319,26 +319,45 @@ private void OnMessage(object sender, WebSocketMessage e)
...
+                            case "hb":
+                                return;
+
+                            // trade
+                            case "tu":
+                                OnUpdate(webSocket, channel.ToStringInvariant(), token[2].ToObject<string[]>());
+                                break;
+
+                            // ignore
+                            case "te":
```
Martin-Molinero, line 344: Wondering what's `te`? Maybe we should add why we ignore it in the comment

Brokerages/Bitfinex/BitfinexSubscriptionManager.cs
```diff
@@ -462,9 +485,6 @@ private void OnSnapshot(BitfinexWebSocketWrapper webSocket, string channelId, st
                     case "book":
                         ProcessOrderBookSnapshot(channel, entries);
                         return;
-                    case "trades":
-                        ProcessTradesSnapshot(channel, entries);
```
Martin-Molinero, line 466: Wondering why was this a bug?

Brokerages/Bitfinex/BitfinexBrokerage.cs
```diff
@@ -175,10 +201,10 @@ public override List<Order> GetOpenOrders()
                     continue;
                 }
 
-                order.Quantity = item.Side == "sell" ? -item.OriginalAmount : item.OriginalAmount;
-                order.BrokerId = new List<string> { item.Id };
+                order.Quantity = item.Amount;
+                order.BrokerId = new List<string> { item.Id.ToStringInvariant() };
                 order.Symbol = _symbolMapper.GetLeanSymbol(item.Symbol);
-                order.Time = Time.UnixTimeStampToDateTime(item.Timestamp);
+                order.Time = Time.UnixMillisecondTimeStampToDateTime(item.MtsCreate);
```
Martin-Molinero, line 207: We might be able to set the LastUpdateTime with `MtsUpdate` I think, maybe if different from `MtsCreate`?

Brokerages/QuantConnect.Brokerages.csproj
```diff
@@ -352,8 +352,19 @@
...
+    <Compile Include="Bitfinex\Messages\Position.cs" />
+    <Compile Include="Bitfinex\Messages\Ticker.cs" />
+    <Compile Include="Bitfinex\Messages\TradeExecutionUpdate.cs" />
+    <Compile Include="Bitfinex\Messages\Wallet.cs" />
+    <Compile Include="Bitfinex\Converters\TradeExecutionUpdateConverter.cs" />
+    <Compile Include="Bitfinex\Converters\OrderConverter.cs" />
+    <Compile Include="Bitfinex\Messages\OrderFlags.cs" />
+    <Compile Include="Bitfinex\Converters\TickerConverter.cs" />
+    <Compile Include="Bitfinex\Converters\PositionConverter.cs" />
+    <Compile Include="Bitfinex\Converters\WalletConverter.cs" />
```
Martin-Molinero, line 367: Love the new converters 👍

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
```diff
@@ -37,16 +39,21 @@ namespace QuantConnect.Brokerages.Bitfinex
...
-        private volatile bool _streamLocked;
-        private readonly RateGate _restRateLimiter = new RateGate(8, TimeSpan.FromMinutes(1));
+        private readonly RateGate _restRateLimiter = new RateGate(10, TimeSpan.FromMinutes(1));
         private readonly ConcurrentDictionary<int, decimal> _fills = new ConcurrentDictionary<int, decimal>();
         private readonly BitfinexSubscriptionManager _subscriptionManager;
         private readonly SymbolPropertiesDatabase _symbolPropertiesDatabase;
         private readonly IDataAggregator _aggregator;
 
+        // map ClientOrderId -> LEAN order
+        private readonly ConcurrentDictionary<long, Order> _orderMap = new ConcurrentDictionary<long, Order>();
```
Martin-Molinero, line 54: Seems a bit of duplication with `CachedOrderIDs`? We are not removing any order from this collection.
On `GetOpenOrders` I only see we update `CachedOrderIDs` if already present? Should we be adding them instead 🤔 same for `_orderMap`

Martin-Molinero (commented):
Looks good! Leaving some concerns and questions. Sorry for the late review 😅

Brokerages/Bitfinex/BitfinexSubscriptionManager.cs
StefanoRaggi: The `te` message is a trade execution, similar to `tu` (trade execution update):
both messages are sent for each order fill but only the `tu` message contains fields for the fees.
I'll add a comment to clarify

Brokerages/Bitfinex/BitfinexSubscriptionManager.cs
StefanoRaggi: When successfully subscribing to a symbol, the exchange sends a snapshot containing the latest 20-40 ticks (can't remember the exact number), this is historical data which should not be emitted as live ticks.

Brokerages/Bitfinex/BitfinexBrokerage.cs
StefanoRaggi: I'll check if `MtsUpdate` is also updated on each fill, in this case we should update `LastFillTime` instead

Brokerages/Bitfinex/BitfinexBrokerage.Messaging.cs
StefanoRaggi: In `_orderMap` the key is not the Lean `OrderId` but the Bitfinex `ClientOrderId` field (`cid`) which is used to match async messages. 
In the docs they mention this field is meant to be unique in the (although not enforced):
https://docs.bitfinex.com/reference#ws-auth-input-order-new

I think using the `CachedOrderIDs` dictionary, we could have duplicate id issues with existing orders, if an algorithm is started twice on the same day, maybe should use a timestamp instead?

Martin-Molinero (approved):

PR #4591, opened 2020-07-28
Feature emv indicator addition

jovuribe:
#### Description

Adds the new indicator/feature Ease of Movement Value by Richard W. Arms Jr. to lean. It attempts to "ease" the price movement of an underlier by mixing momentum and volume information into one value. The output is mainly used by traders to assess the strength of a trend or to mitigate stop run risks (aka stop loss hunting). 
#### Motivation and Context

This feature gives QC users another form of analysis that helps smooth out outliers that would otherwise occur in an SMA. 

An addition to the indicator page

This indicator still needs to be tested and code reviewed.

#### Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

#### Checklist:

- [ x ] My code follows the code style of this project.
- [ x ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).


Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// Creates a new EaseOfMovement indicator with the specified period
+        /// </summary>
+        /// <param name="name">The name of this indicator</param>
+        /// <param name="period">The period over which to perform to computation</param>
+        public EaseOfMovementValue(string name, int period)
+            : base(name)
+        {
+            WarmUpPeriod = period;
+            _maximum = new Maximum(period);
+            _minimum = new Minimum(period);
```
AlexCatarino, line 74: You could use
```csharp
_previousHighMaximum = new Delay(1).MAX(period);
_previousLowMinimum = new Delay(1).MIN(period);
```
to eliminate `PreviousHighPrice` and `PreviousLowPrice`. I don't think they should be public members.

Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,102 @@
...
+            _maximum.Update(new IndicatorDataPoint { Value = input.High });
+            _minimum.Update(new IndicatorDataPoint { Value = input.Low });
+
+            var MIDvalue = ((input.High + input.Low) / 2) - ((PreviousHighPrice + PreviousLowPrice) / 2);
+            var MIDratio = ((input.Volume / 10000) / (input.High - input.Low));
+
+            PreviousHighPrice = input.High;
+            PreviousLowPrice = input.Low;
+
+            return (MIDvalue / MIDratio);
```
AlexCatarino, line 96: We need some divide by zero error prevention in this method. If `input.High = input.Low`, an exception will be thrown at line 91 and if `input.Volume == 0`, `MIDratio` will be zero and an exception will be thrown at line 96.

Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        }
+
+
```
AlexCatarino, line 100: nit. Please don't forget to remove some empty lines. :-)

Indicators/EaseOfMovementValue.cs
jovuribe: Thanks Alex. I made all the changes you requested. I also removed the maximum indicator as it was redundant.

Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// Gets a flag indicating when this indicator is ready and fully initialized
+        /// </summary>
+        public override bool IsReady => Samples >= 2;
+
+        /// <summary>
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public int WarmUpPeriod => 2;
+
+        public override void Reset()
```
Martin-Molinero, line 40: nit - In Lean we usually place Constructor first. Missing xml docs

Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Computes the next value for this indicator from the given state.
+        /// </summary>
+        /// <param name="input">The input value to this indicator on this time step</param>
+        /// <returns>A a value for this indicator</returns>
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            if (_previousHighMaximum == null) _previousHighMaximum = input.High;
+            if (_previousLowMinimum == null) _previousLowMinimum = input.Low;
```
Martin-Molinero, line 72: `_previous*****` will never be null since thy are value types, maybe we want to use nullable decimal type `decimal?` or maybe we could just check the current Sample count to determine if we should compute, but still set `_previous*****` always. Less `if`s the better!

Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,88 @@
...
+
+            if (input.Volume == 0 || input.High == input.Low)
+            {
+                return 0;
+            }
+
+            var midValue = ((input.High + input.Low) / 2) - ((_previousHighMaximum + _previousLowMinimum) / 2);
+            var midRatio = ((input.Volume / 10000) / (input.High - input.Low));
```
Martin-Molinero, line 80: Seems `10000` should be an argument of the indicator maybe with that default value?

Indicators/EaseOfMovementValue.cs
```diff
@@ -0,0 +1,88 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    /// This indicator computes the n-period Ease of Movement Value using the following:
```
Martin-Molinero, line 20: `This indicator computes the **n-period**` seems we just look at current and previous? At https://www.daytrading.com/ease-of-movement it talks about a moving average too

Indicators/EaseOfMovementValue.cs
jovuribe: Hi Martin, thank you for the feedback. The xml doc that is missing is QuantConnect.Indicators.xml?

Indicators/EaseOfMovementValue.cs
jovuribe: Gottcha. I'll make these changes.

Indicators/EaseOfMovementValue.cs
jovuribe: Yep. That is something that can be added. It is a simple moving average of the 1-period EMV.

Tests/Indicators/EaseOfMovementValueTests.cs
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+
+        protected override Action<IndicatorBase<TradeBar>, double> Assertion
+        {
+            get { return (indicator, expected) => Assert.AreEqual(expected, (double)indicator.Current.Value, 1); }
+        }
+
+        [Test]
+        public virtual void PeriodSet()
```
Martin-Molinero, line 59: nit - don't think there is a need for this unit test to be virtual

PR #4596, opened 2020-07-29
Best efforts trying to get this running in a docker image. Added step…

cottrell:
Adding this as a discussion point/feedback point. 

If it is possible to run this stuff locally to try it out I would like to make this more obvious to new users. 

The Makefile now builds and runs a docker image but the example I tried in a notebook failed with various errors on creation of an algo.

…s to a Makefile for clarity. Docker builds and spins up but the algo I tried threw many errors. There are likely other dependencies that are not setup. Leaving this here for follow up.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -267,3 +267,6 @@ Launcher/Plugins/*
 /ApiPython/quantconnect.egg-info/*
 
 QuantConnect.Lean.sln.DotSettings*
+
+notebooks/
+.*.swp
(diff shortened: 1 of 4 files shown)


Martin-Molinero:
Hey @cottrell!
Thanks for the PR, we've recently worked on improving our research documentation and on adding run scripts that handle all the details with default values, at https://github.com/QuantConnect/Lean/pull/4622. We've also enabled allowing users to retrieve their notebooks out from the containers easily. I'll be closing this PR since I think it's been addressed for now. Please check it out and let us know if you have any issues or suggestions.

PR #4618, opened 2020-08-07
Performance Issue #3622 Log handling via a dedicated thread

### LINKED ISSUE #3622: Consolidate logging to reduce thread contention on single log file
#### Expected Behavior

Log spamming shouldn't block threads.

#### Actual Behavior

Log spamming causes lots of thread contention as threads wait for their turn to write _and flush_ to a single log file.

#### Potential Solution

Queue up log messages and have a separate thread be responsible for writing them. We should also consolidate the messages to provide for less write operations in total.

#### Reproducing the Problem

Run and algorithm lots of sparse data sets and observe the slower than expected run time due to log file contention from missing data log spam.

#### Checklist

C-SELLERS:
#### Description

The current implementation of log handling has thread contention over which thread is using the logger. The new baseloghandler class implements a queue log system that allows the rest of the engine to post log messages to a logging queue and then alert the logging thread that there is new messages that need to be processed. Then the logging thread can process each message in the queue through the implementing classes log methods. 

This PR does change the functionality of the console and file log handlers to implement the new abstract BaseLogHandler class; now the console and file log handlers contain a set of methods for writing to their respective outs (WriteDebug, WriteTrace, WriteError) which is called on by the Logger Thread.

#### Related Issue

Closes #3622 

#### Motivation and Context

- Performance improvement since it removes thread contention for logger
- Separates i/o functions from the rest of the worker and engine threads which is a safer approach.

None

Unit test included TestLoggingSpeeds(int x); these are ignored by default.
x = number of threads writing to the log. Each thread writes 1000 messages concurrently to the log. 

Comparing the current implementation vs this new one shows pretty significant differences, cuts the time to about half.

#### Types of changes

#### Checklist:

### DIFF
--- a/Logging/BaseLogHandler.cs
+++ b/Logging/BaseLogHandler.cs
@@ -15,7 +15,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Globalization;
 using System.Threading;
 
 namespace QuantConnect.Logging
@@ -26,7 +25,6 @@ namespace QuantConnect.Logging
     public abstract class BaseLogHandler : ILogHandler
     {
         private readonly ConcurrentQueue<LogEntry> _logsQueue;
-        private const string DateFormat = "yyyyMMdd HH:mm:ss";
         private readonly Thread _logThread;
         private readonly CancellationTokenSource _cancellationTokenSource;
         private readonly AutoResetEvent _messageEvent;
@@ -77,16 +75,16 @@ namespace QuantConnect.Logging
         }
 
         /// <summary>
-        /// Thread function in charge of dispersing log messages
+        /// Thread function in charge of dispersing log messages to abstract write methods
         /// </summary>
-        public void WriteThread()
+        private void WriteThread()
         {
-            try
-            {
-                while(!_cancellationTokenSource.IsCancellationRequested && _messageEvent.WaitOne())
-                {
-                    LogEntry entry;
+            LogEntry entry;
 
+            while (!_cancellationTokenSource.IsCancellationRequested && _messageEvent.WaitOne())
+            {   
+                try
+                {
                     while (_logsQueue.TryDequeue(out entry))
                     {
                         switch(entry.MessageType)
@@ -102,27 +100,28 @@ namespace QuantConnect.Logging
                                 break;
                         }
                     }
+                } catch (Exception exception)
+                {
+                    Console.WriteLine("Logging thread has failed: " + exception);
+                    Thread.Sleep(10);
                 }
-            } catch (Exception exception)
-            {
-                Console.WriteLine("Logging thread has failed: " + exception);
             }
         }
 
         /// <summary>
         /// Function to write a trace statement
         /// </summary>
-        public abstract void WriteTrace(string message);
+        protected abstract void WriteTrace(string message);
 
         /// <summary>
         /// Function to write a debug statement
         /// </summary>
-        public abstract void WriteDebug(string message);
+        protected abstract void WriteDebug(string message);
 
         /// <summary>
         /// Function to write a error statement
         /// </summary>
-        public abstract void WriteError(string message);
+        protected abstract void WriteError(string message);
 
         /// <summary>
         /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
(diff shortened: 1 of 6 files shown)


mchandschuh:
What is the behavior during fatal runtime errors? These log files are the singular source for determining why a live algorithm unexpectedly fails. If these messages are backed in a queue then it's possible they never get written to disk before the process terminates. This is potentially exacerbated by the fact that the thread is initialized as a background thread which means the CLR will exit the process with that thread still running. We may want to test add some tests to provide confidence around such failure modes.

Tests/Logging/FileLogHandlerTests.cs
```diff
@@ -64,5 +66,52 @@ public void UsesGlobalFilePath()
...
+                }
+
+                //Wait for all threads to complete
+                foreach (var thread in threads)
+                {
+                    thread.Join();
+                }
+            }
+
+            var end = DateTime.Now;
```
mchandschuh, line 111: Should always use `UtcNow` -- it sucks when you run a test and you just happen to cross over daylight savings time :)

Logging/BaseLogHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+                                break;
+                            case LogType.Debug:
+                                WriteDebug(entry.Message);
+                                break;
+                            case LogType.Error:
+                                WriteError(entry.Message);
+                                break;
+                        }
+                    }
+                } catch (Exception exception)
```
mchandschuh, line 103: catch should be on a new line

Logging/BaseLogHandler.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <filterpriority>2</filterpriority>
+        public virtual void Dispose()
+        {
+            try
+            {   
+                _cancellationTokenSource.Cancel();
+                _messageEvent.Set();
+                if (!_logThread.Join(1000))
+                {
+                    Console.WriteLine($"Timeout waiting for '{_logThread.Name}' thread to stop");
```
mchandschuh, line 138: I think in places of the code we re-route the console, so this may not be safe unless we first grab the reference in a static constructor and save it (`Console.Out`)

Logging/FileLogHandler.cs
```diff
@@ -84,15 +83,14 @@ public void Trace(string text)
...
+            base.Dispose();
+
+            if (_writer.IsValueCreated)
             {
-                if (_writer.IsValueCreated)
-                {
-                    _disposed = true;
-                    _writer.Value.Dispose();
-                }
+                _disposed = true;
```
mchandschuh, line 92: I have a sneaking suspicion that we've introduced a potential race by removing the synchronization on dispose. While it's incredibly unlikely (even impossible in this code base) for multiple threads to call dispose at the same time, it's arguably negligible on performance to leave it in. Locked objects that are not in contention are pretty cheap to acquire, especially when the acquire count is once (one app exit)

Logging/FileLogHandler.cs
```diff
@@ -20,14 +20,13 @@
 namespace QuantConnect.Logging
 {
     /// <summary>
-    /// Provides an implementation of <see cref="ILogHandler"/> that writes all log messages to a file on disk.
+    /// Provides an implementation of <see cref="BaseLogHandler"/> that writes all log messages to a file on disk.
     /// </summary>
-    public class FileLogHandler : ILogHandler
+    public class FileLogHandler : BaseLogHandler
     {
         private bool _disposed;
```
mchandschuh, line 27: This may need to be marked as `volatile` now since it's being read from a background thread but its value could be changed from a foreground thread which makes the read in `WriteMessage` potentially stale.

Another option is to change it to a `CancellationTokenSource` which manages the thread memory barriers correctly and efficiently.

Logging/BaseLogHandler.cs
Martin-Molinero: True! I think we could actually directly call `WriteError` here, `ConsoleLogHandler` and `FileLogHandler` will work correctly, file handler is disposing after the base

Martin-Molinero:
> 
> 
> What is the behavior during fatal runtime errors? These log files are the singular source for determining why a live algorithm unexpectedly fails. If these messages are backed in a queue then it's possible they never get written to disk before the process terminates. This is potentially exacerbated by the fact that the thread is initialized as a background thread which means the CLR will exit the process with that thread still running. We may want to test add some tests to provide confidence around such failure modes.
> 
> Also, when running locally it's incredibly useful that the console logs are synchronously ordered and relate to current processing events. With this change, we break this temporal coupling and the console logs are now a view of delayed processing. The value of this is left to the reader to decide.

Yeah, I've been thinking of this too, we could add a `Log` config bool flag, e.g `Log.AsynchronousEnabled`, so we can disable this feature _if_ we have some explosion so solve. The `BaseLogHandler` could check this config value on `Error/Debug/Trace` and enqueue or directly push the log.
Note that the order of the logging is still respected here and the time will match the enqueue time too.

Martin-Molinero:
We will close this for now as `WillNoFix` since the benefit for Lean logging usage does not overcome the problems associated with runtime explosions and the potential loss of those precious last logs.

PR #4621, opened 2020-08-10
Refactors SecurityCache to Support Access to Tick of Different Type

### LINKED ISSUE #4617: SecurityCache Does Not Distinguish By TickType
#### Expected Behavior
Be able to get the last `Tick` of all `TickType` deterministically.

#### Actual Behavior
We are only able to get the last `Tick` that can be of any particular `TickType`.

#### Potential Solution
Save `Tick` by `TickType`.

#### Reproducing the Problem
See #4457 

#### Checklist

AlexCatarino:
#### Description
Adds two read-only lists of `BaseData` to save the last list of `Tick` of `TickType.Trade` and `TickType.Quote`. Changes methods accordingly to get and set these lists.

#### Related Issue
Closes #4617

#### Motivation and Context
It's important to be able to access the last tick of each type since fill models take different actions on a different kind of `Tick`.

Unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/SecurityCache.cs
+++ b/Common/Securities/SecurityCache.cs
@@ -36,6 +36,8 @@ namespace QuantConnect.Securities
         private DateTime _lastQuoteBarUpdate;
         private DateTime _lastOHLCUpdate;
         private BaseData _lastData;
+        private IReadOnlyList<BaseData> _lastTickQuotes = new List<BaseData>();
+        private IReadOnlyList<BaseData> _lastTickTrades = new List<BaseData>();
         private ConcurrentDictionary<Type, IReadOnlyList<BaseData>> _dataByType = new ConcurrentDictionary<Type, IReadOnlyList<BaseData>>();
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Common/Securities/SecurityCache.cs
```diff
@@ -139,18 +141,30 @@ public void AddData(BaseData data)
...
+                        if (tick.AskSize != 0) AskSize = tick.AskSize;
+                        break;
+
+                    case TickType.OpenInterest:
+                        OpenInterest = (long)tick.Value;
+                        return;
+                }
+
+                if (tick.Value != 0) Price = tick.Value;
+                _lastData = data;
```
Martin-Molinero, line 167: I think `_lastData = data;` will be set bellow if appropriate? One thing that's different here is that before if fillFowarding we wouldn't update the `Price/AskSize/..etc`, seems we should respect that behavior

Common/Securities/SecurityCache.cs
```diff
@@ -294,6 +320,11 @@ public T GetData<T>()
         /// </summary>
         public IEnumerable<T> GetAll<T>()
         {
+            if (typeof(T) == typeof(Tick))
+            {
+                return _lastTickTrades.Concat(_lastTickQuotes).Cast<T>();
+            }
+
             IReadOnlyList<BaseData> list;
```
Martin-Molinero, line 328: Believe we still need to modify the other methods in the security cache for the behavior to be consistent, `Reset`, `TryGetValue`, `ShareTypeCacheInstance`. Think `HasData/GetData` should also use the new collections so we can avoid duplicating the Tick storage with `_dataByType`

Tests/Common/Securities/SecurityCacheTests.cs
```diff
@@ -102,6 +102,56 @@ public void UseLatestTradebarIfThereIsntAvailableQuotebar()
...
+            // Adding OpenInterest as Tick or OpenInterest should not matter
+            var openInterest = new OpenInterest(time, Symbol.Empty, 1000);
+            securityCache.StoreData(new[] { openInterest }, typeof(Tick));           // Add as Tick
+            securityCache.StoreData(new[] { openInterest }, typeof(OpenInterest));   // Add as OI
+
+            // Assert
+            Assert.True(securityCache.GetData<Tick>().Equals(trade));
+
+            Assert.True(securityCache.GetAll<Tick>().LastOrDefault(x => x.TickType == TickType.Quote).Equals(quote));
+            Assert.True(securityCache.GetAll<Tick>().LastOrDefault(x => x.TickType == TickType.Trade).Equals(trade));
```
Martin-Molinero, line 152: Nice tests 👍  I think we are just missing some more for the other methods on SecurityCache

PR #4644, opened 2020-08-22
Bug Backtesting Brokerage Clones

### LINKED ISSUE #2846: Exercised/assigned option quantity used to compute Equity average price
#### Expected Behavior
When an option contract is exercised/assigned, the equity order average price must be the strike price

#### Actual Behavior
When an option contract is exercised/assigned, the equity order average price differs from the strike price.

#### Potential Solution
Do not account the option contract quantity when computing the quantity filled.

#### Reproducing the Problem
```python

class BasicTemplateOptionsAlgorithm(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2017, 2, 1)
        self.SetEndDate(2017, 3, 1)
        self.SetCash(100000)

        option = self.AddOption("IBM")
        self.option_symbol = option.Symbol
        option.SetFilter(-10, +10, timedelta(0), timedelta(30))


    def OnData(self,slice):
        if self.Portfolio.Invested: return

        chain = slice.OptionChains.GetValue(self.option_symbol)
        if chain is None:
            return
        contracts = sorted(chain, key = lambda x: x.Strike)
        if len(contracts) == 0: return
        symbol = contracts[-1].Symbol
            self.MarketOrder(symbol, 1)
 

    def OnOrderEvent(self, orderEvent):
        self.Debug("IBM Price "+str(self.Securities["IBM"].Price))
        self.Log(str(orderEvent))
```
The order event shows $200 for the assigned IBM shares, but the recoded order average price is $198.0198019802 USD (=200/101).

#### Checklist

### LINKED ISSUE #4445: Orders Being Partially Filled By The Same Bar in Backtesting
#### Expected Behavior
Do not try to fill an order with the same data.

#### Actual Behavior
1- For a single algorithm manager loop, the backtesting brokerage is triggered immediately for each order, and in the case of Market orders he fills immediately too. Also the backtesting brokerage is triggered at the end of each loop.
2- Once the backtesting brokerage gets an Order he creates a copy. So in the case of a partial fill the orders won't be updated, for example `LastFillTime`. Note that this is the Order which gets sent to the fill models. Since `LastFillTime` isn't sent, it can't use this information to decide whether it should emit a new partial fill + `1-` produces two fills per loop unless the fill model keeps track himself of the last fill time.

#### Reproducing the Problem
```csharp
    public class CustomModelsAlgorithm : QCAlgorithm {
        private Security _security;
        private Symbol _spy;

            SetStartDate(2013, 10, 01);
            SetEndDate(2013, 10, 31);
            _security = AddEquity("SPY", Resolution.Hour);
            _spy = _security.Symbol;
            _security.SetFillModel(new CustomFillModel(this));
        }

        public void OnData(TradeBars data) {
            var openOrders = Transactions.GetOpenOrders(_spy);
            if (openOrders.Count != 0) return;

            if (Time.Day > 10 && _security.Holdings.Quantity <= 0) {
                var quantity = CalculateOrderQuantity(_spy, .5m);
                MarketOrder(_spy, quantity, asynchronous: true); // async needed for partial fill market orders
            } else if (Time.Day > 20 && _security.Holdings.Quantity >= 0) {
                var quantity = CalculateOrderQuantity(_spy, -.5m);
                MarketOrder(_spy, quantity, asynchronous: true); // async needed for partial fill market orders
            }
        }

        public class CustomFillModel : ImmediateFillModel {
            private readonly QCAlgorithm _algorithm;
            private readonly Dictionary<long, decimal> _absoluteRemainingByOrderId = new Dictionary<long, decimal>();

            public CustomFillModel(QCAlgorithm algorithm) {
                _algorithm = algorithm;
            }

            public override OrderEvent MarketFill(Security asset, MarketOrder order) {
                decimal absoluteRemaining;
                if (!_absoluteRemainingByOrderId.TryGetValue(order.Id, out absoluteRemaining)) {
                    absoluteRemaining = order.AbsoluteQuantity;
                    _absoluteRemainingByOrderId.Add(order.Id, order.AbsoluteQuantity);
                }

                var absoluteFillQuantity = (int) (Math.Min(absoluteRemaining, 100));
                fill.FillQuantity = Math.Sign(order.Quantity) * absoluteFillQuantity;

                if (absoluteRemaining == absoluteFillQuantity) {
                    fill.Status = OrderStatus.Filled;
                    _absoluteRemainingByOrderId.Remove(order.Id);
                } else {
                    absoluteRemaining -= absoluteFillQuantity;
                    _absoluteRemainingByOrderId[order.Id] = absoluteRemaining;
                    fill.Status = OrderStatus.PartiallyFilled;
                }
                _algorithm.Log("CustomFillModel: " + fill);
                return fill;
            }
        }
    }
}
```
#### Checklist

C-SELLERS:
#### Description

In the current implementation the `BacktestingBrokerage` passes around a clone of an order for processing. As a result when trying to make use of order variables such as `LastFillTime`, or `StopTriggered` that would be updated by the Backtesting Brokerage, it won't work since the Algorithms order is not updated.

In correcting this issue, I uncovered other issues related to the processing of `OptionExercised` orders. Without a clone the `OptionExercised` order takes on the value assigned by the ticket in `HandleOrderEvent()`, The initial tickets `AverageFillPrice` = 0 since the first event is removing the expired option. If I did not correct this problem after removing the `BackTestingBrokerage` clone, it would have allowed the option exercised price to be changed permanently to 0, therefore exercising a option to buy 100 shares of X at $0.

To fix this, I placed a conditional that if the `OrderType` is `OptionExercised`, the price is fixed and should not be changed.

Furthermore I had to make changes to some of our regression algorithms `OrderListsHash` to match the output. These changes have been analyzed and tested to be sure they reflects what the values should be.

Attached is a report on the Regression Algorithms that had to be changed and why.

[Regression Algorithm Changes Report.docx](https://github.com/QuantConnect/Lean/files/5119444/Regression.Algorithm.Changes.Report.docx)

#### Related Issue

Closes #4445 - With the clone removed the order will now have the `LastFillTime` updated
Closes #2846 - The `OptionExercised` order `price` will now not be skewed by the removal of the option. It is handled in the processing of the ticket.

#### Motivation and Context

Read above!

None

All regression tests have been run and are successful. The changed regression algorithms are described in the attached document above.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/BacktestingBrokerageRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/BacktestingBrokerageRegressionAlgorithm.cs
@@ -1,0 +1,263 @@
+﻿using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Interfaces;
+using QuantConnect.Orders;
+using QuantConnect.Orders.Fees;
+using QuantConnect.Orders.Fills;
+using QuantConnect.Securities;
+using QuantConnect.Securities.Option;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+
+    /// <summary>
+    /// This regression algorithm tests the order processing of the backtesting brokerage.
+    /// We open an equity position that should fill in two parts, on two different bars.
+    /// We open a long option position and let it expire so we can exercise the position.
+    /// To check the orders we use OnOrderEvent and throw exceptions if verification fails.
+    /// </summary>
+    /// <meta name="tag" content="backtesting brokerage" />
+    /// <meta name="tag" content="regression test" />
+    /// <meta name="tag" content="options" />
+    class BacktestingBrokerageRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        private Security _security;
+        public Symbol Spy;
+        private Option _option;
+        public Symbol OptionSymbol;
+        private bool _optionBought = false;
+        private bool _equityBought = false;
+
+        private decimal _optionStrikePrice;
+
+        /// <summary>
+        /// Initialize the algorithm
+        /// </summary>
+        public override void Initialize()
+        {
+            SetCash(1000000);
+            SetStartDate(2015, 12, 24);
+            SetEndDate(2015, 12, 28);
+
+            _security = AddEquity("SPY", Resolution.Hour);
+            Spy = _security.Symbol;
+
+            _security.SetFillModel(new PartialMarketFillModel(2));
+
+
+            _option = AddOption("GOOG");
+            OptionSymbol = _option.Symbol;
+
+            // Set our strike/expiry filter for this option chain
+            _option.SetFilter(u => u.IncludeWeeklys()
+                                   .Strikes(-2, +2)
+                                   .Expiration(TimeSpan.Zero, TimeSpan.FromDays(10)));
+        }
+
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice data)
+        {   
+            if (!_equityBought && data.ContainsKey(Spy)) {
+                //Buy our Equity
+                var quantity = CalculateOrderQuantity(Spy, .1m);
+                MarketOrder(Spy, quantity, asynchronous: true);
+                _equityBought = true;
+            }
+
+            if (!_optionBought)
+            {
+                // Buy our option
+                OptionChain chain;
+                if (data.OptionChains.TryGetValue(OptionSymbol, out chain))
+                {
+                    // Find the second call strike under market price expiring today
+                    var contracts = (
+                        from optionContract in chain.OrderByDescending(x => x.Strike)
+                        where optionContract.Right == OptionRight.Call
+                        where optionContract.Expiry == Time.Date
+                        where optionContract.Strike < chain.Underlying.Price
+                        select optionContract
+                        ).Take(2);
+
+                    if (contracts.Any())
+                    {
+                        var optionToBuy = contracts.FirstOrDefault();
+                        _optionStrikePrice = optionToBuy.Strike;
+                        MarketOrder(optionToBuy.Symbol, 2);
+                        _optionBought = true;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// All order events get pushed through this function
+        /// </summary>
+        /// <param name="orderEvent">OrderEvent object that contains all the information about the event</param>
+        public override void OnOrderEvent(OrderEvent orderEvent)
+        {   
+            // Get the order from our transactions
+            var order = Transactions.GetOrderById(orderEvent.OrderId);
+
+            // Based on the type verify the order
+            switch(order.Type)
+            {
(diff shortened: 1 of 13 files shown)


Brokerages/Backtesting/BacktestingBrokerage.cs
```diff
@@ -517,8 +517,7 @@ public override void Disconnect()
         /// <returns></returns>
         private void SetPendingOrder(Order order)
         {
-            // only save off clones!
-            _pending[order.Id] = order.Clone();
```
jaredbroad, line 521: Can we please add an order immutability regression test? I don't think we explicitly cover that -- something like we try 10 different ways to edit the order value but from QCAlgorithm layer it should be impossible

C-SELLERS: Added a regression algorithm for that, from what I see, it seems like I covered the only ways to try and modify the order. Since you only get the ticket in the algorithm I had to go through `Transactions`!

C-SELLERS:
Closing to rerun Travis!

Algorithm.CSharp/BacktestingBrokerageRegressionAlgorithm.cs
```diff
@@ -0,0 +1,274 @@
...
+    /// </summary>
+    /// <meta name="tag" content="backtesting brokerage" />
+    /// <meta name="tag" content="regression test" />
+    /// <meta name="tag" content="options" />
+    class BacktestingBrokerageRegressionAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        private Security _security;
+        public Symbol Spy;
+        private Option _option;
+        public Symbol OptionSymbol;
```
Martin-Molinero, line 44: nit - a couple of unrequired publics here

Algorithm.CSharp/BacktestingBrokerageRegressionAlgorithm.cs
```diff
@@ -0,0 +1,274 @@
...
+            {
+                throw new Exception("OptionExercise order price should be strike price!!");
+            }
+        }
+
+        /// <summary>
+        /// PartialMarketFillModel that allows the user to set the number of fills and restricts
+        /// the fill to only one per bar.
+        /// </summary>
+        public class PartialMarketFillModel : ImmediateFillModel
```
Martin-Molinero, line 181: nit - Let's make this class private too so it's not publicly available and adds noise into the namespace

Algorithm.CSharp/BacktestingBrokerageRegressionAlgorithm.cs
```diff
@@ -0,0 +1,274 @@
...
+        }
+
+        /// <summary>
+        /// To verify OptionExercise orders is process correctly
+        /// </summary>
+        /// <param name="order">Order object to analyze</param>
+        public void VerifyOptionExercise(Order order)
+        {
+            // If the option price isn't the same as the strike price, its incorrect
+            if (order.Price != _optionStrikePrice)
```
Martin-Molinero, line 171: Think we should assert quantities too. Let's also add an override of the `OnEndOfAlgorithm` where we assert the final status of the algorithms holdings, quantities and prices to make sure the modeling worked as expected

Martin-Molinero (approved):

Martin-Molinero (changes requested):
Requires a rebase from master, some tests we added recently for the option filtering are failing

Martin-Molinero (approved):

PR #4650, opened 2020-08-24
Bug 4031 Change data depending on configuration

### LINKED ISSUE #4031: Options added after Initialize with already existing equity
#### Expected Behavior

- Options added after Initialize with already existing equity are deterministic and work correctly.
#### Actual Behavior

- Options require raw data mode but by default, equities use adjusted data mode. When an option is added and the underlying equity has already been added to Lean, it will set the data mode to raw https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.Universe.cs#L78 `ConfigureUnderlyingSecurity()`, but the data of the underlying might of already been all loaded by the workers so this change has no effect in the data itself.

#### Potential Solution

- Existing equity subscription should be removed/discarded and a new subscription should be added using the raw data mode
#### Reproducing the Problem

See AddRemoveOptionUniverseRegressionAlgorithm
#### System Information

N/A
#### Checklist

AdalyatNazirov:
#### Description

Precalculate Raw and Adjusted data for `FileSystemDataFeed`. Choose one of them on demand depending on configuration settings.

#### Related Issue

Closes #4031

#### Motivation and Context

Options require raw data mode but by default, equities use adjusted data mode. When an option is added and the underlying equity has already been added to Lean, it will set the data mode to raw but the data of the underlying might of already been all loaded by the workers so this change has no effect in the data itself.

N/A

- Manual testing
- Unit tests
- Benchmarks. Algorithhms are run on Intel i5 6200u, 8GB RAM, SSD. Use all SPY minute resolution for both BasicTemplateBenchmark and EmptyMinute400EquityBenchmark (hardcoded data source `equity\usa\minute\spy`). Due to test requirements we also need to provide factor files for all Symbols, so I hardcoded SPY factor file too. Results are below

image

#### Types of changes

#### Checklist:


Engine/DataFeeds/SubscriptionUtils.cs
```diff
@@ -82,9 +92,21 @@ public static class SubscriptionUtils
...
                         }
 
-                        var subscriptionData = SubscriptionData.Create(subscription.Configuration, exchangeHours,
-                            subscription.OffsetProvider, enumerator.Current);
+                        var data = enumerator.Current;
+
+                        if (enablePriceScale && data?.Time.Date > lastTradableDate)
+                        {
+                            lastTradableDate = data.Time.Date;
+                            currentScale = GetScaleFactor(factorFile, config.DataNormalizationMode, data.Time.Date);
```
Martin-Molinero, line 100: There's a race condition between this `config.DataNormalizationMode` and the internal to `                         SubscriptionData subscriptionData = SubscriptionData.Create`, think we should get the mode once

Engine/DataFeeds/SubscriptionData.cs
```diff
@@ -72,6 +80,41 @@ public static SubscriptionData Create(SubscriptionDataConfig configuration, Secu
                 data.Time = data.Time.ExchangeRoundDownInTimeZone(configuration.Increment, exchangeHours, configuration.DataTimeZone, configuration.ExtendedMarketHours);
             }
 
+            if (factor.HasValue && (factor.Value != 1 || configuration.SumOfDividends != 0))
```
Martin-Molinero, line 83: Let's simplify this behavior so we only support for now going from Any data mode -> Raw and other cases throw invalid operation exception 👍

Martin-Molinero (changes requested):
Another hard issue solved! Almost there, let's just simplify behavior for now to fix the related issue 👍

Martin-Molinero (approved):

PR #4662, opened 2020-08-26
[Travis #14081] Autogenerated Python Stubs

quantconnect-server:


Algorithm.Python/stubs/QuantConnect/Algorithm/Framework/Alphas.py_0
```diff
@@ -0,0 +1,131 @@
+# encoding: utf-8
```
gsalaz98, line 1: This file and `Alphas.py_1` are duplicates of the contents of `QuantConnect/Algorithm/Framework/Alphas/__init__[0-9].py` - but don't matter for now since they're not declared as actual Python files, so the types will resolve correctly.

This is a bug in the underlying generator that'll eventually need to be fixed, maybe caused by namespace clash between QuantConnect.Algorithm and QuantConnect.Algorithm.Framework

Issue #4680, opened 2020-09-02
Portfolio.CashBook is not synchronous with my brokerage holding in Bitfinex


KwFung7:
I'm using **Bitfinex** as my algorithm brokerage. The algorithm run **once per hours** for crypto day trading.
The **Portfolio.CashBook** and its total value are always asynchronous with actual Bitfinex holding after **SetHoldings**,
so my algorithm sometimes run with outdated Portfolio value during live trading, this situation won't happen in backtest.
It will resume normal **after redeployment or one day after**.

For example, my strategy always maintein 90% of Portfolio value on ETH, I assume that someday ETH value raised, then my strategy will sell some of my ETH to maintein 90% ratio. but unfortunately in live mode, the amount in CashBook can't update on time, so the program keep sell out my ETH on brokerage until nothing left.

#### Expected Behavior
I expected that the Portfolio.Cashbook should always align with actual Bitfinex holding.

#### Actual Behavior
Synchronization seems only run once per day, it may cause the wrong value on Portfolio.CashBook
Or it may be problem on SetHolding function in specific brokerage such as Bitfinex

#### Potential Solution
#4681 
- I checked **Brokerage.cs**, the synchronization only run once per day, can it run much more frequent such as once per hour ?
- It may be problem on SetHolding function in specific brokerage such as Bitfinex

#### Reproducing the Problem
It is easy to reproduce in my account but I dont wanna to screenshot my account holding.

#### System Information
Quantconnect Live Trading with Chrome on Window 10

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #4681, opened 2020-09-02
Syncronize with brokerage once per hour instead of once per day (#4680)

KwFung7:
#### Description
I checked Brokerage.cs, the synchronization with brokerage only run once per day, can it run much more frequent such as once per hour instead of once per day ?

#### Related Issue
https://github.com/QuantConnect/Lean/issues/4680

#### Motivation and Context
I'm using Bitfinex as my algorithm brokerage. The algorithm run once per hours for crypto day trading.
The Portfolio.CashBook and its total value are always asynchronous with actual Bitfinex holding after SetHoldings,
so my algorithm sometimes run with outdated Portfolio value during live trading, this situation won't happen in backtest.
It will resume normal after redeployment or one day after.

For example, my strategy always maintein 90% of Portfolio value on ETH, I assume that someday ETH value raised, then my strategy will sell some of my ETH to maintein 90% ratio. but the amount in CashBook can't update on time, so the program keep sell out my ETH on brokerage until nothing left.

I guess no documentation need to be updated, please advice.

Don't know how to test it, I'm not familiar with this infrastructure, please advice.

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/Brokerage.cs
+++ b/Brokerages/Brokerage.cs
@@ -31,11 +31,8 @@ namespace QuantConnect.Brokerages
     /// </summary>
     public abstract class Brokerage : IBrokerage
     {
-        // 7:45 AM (New York time zone)
-        private static readonly TimeSpan LiveBrokerageCashSyncTime = new TimeSpan(7, 45, 0);
-
         private readonly object _performCashSyncReentranceGuard = new object();
-        private bool _syncedLiveBrokerageCashToday = true;
+        private bool _syncedLiveBrokerageCash = true;
         private long _lastSyncTimeTicks = DateTime.UtcNow.Ticks;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


jaredbroad:
Welcome to LEAN @KwFung7! Thank you for your contribution. I can see this being a configuration option -- but perhaps we can default to once per day. There are risks involved with synchronizing, if you are placing trades in the middle of synchronization it becomes very hard to maintain a clean state and it can result in having a larger cash balance than expected. We try and work around this by synchronizing premarket when few people trade and looking for trades and changes that occur during sync but it is still not a simple process.

For those who opt for more sync this is fine -- can you please update it to be a static configuration based variable in your code -- defaulting to 24 hours; and then we can merge this with few little edits.

KwFung7:
I agreed that its high risk code change, I will try to update it to configuration based later. Thanks.

jaredbroad:
Hey @KwFung7! Closing pending edits per review but please resubmit when you have a chance to address the review.

Issue #3707, opened 2019-10-10
FillForwardEnumerator + History infinite loop


Martin-Molinero:
#### Expected Behavior

- No infinite loop situation
#### Actual Behavior

- In some cases due to daylight savings `FillForwardEnumerator` always emits the same fill forward bar (time doesn't advance)
#### Potential Solution

`FillForwardEnumerator` could check new emit time and compare it with previous. If the previous bar was also a fillforward (true for this case), there is no need to `RoundDown()` the new emit time which causes this issue.
#### Reproducing the Problem

This issue was discovered testing https://github.com/QuantConnect/Research/issues/12#issuecomment-528204874, can be trimmed down to the following unit test:
```
        [Test]
        public void InfiniteLoop()
        {
            // sydney time advanced Sunday, 6 October 2019, 02:00:00 clocks were turned forward 1 hour to
            // Sunday, 6 October 2019, 03:00:00 local daylight time instead.
            var timeAt = new DateTime(2019, 10, 6, 10, 0, 0);
            var expected = new DateTime(2019, 10, 5, 10, 0, 0);

            var exchangeRoundedAt = timeAt.RoundDownInTimeZone(Time.OneDay, TimeZones.Sydney, TimeZones.Utc);
            Assert.AreEqual(expected, exchangeRoundedAt);
        }
```
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #4700, opened 2020-09-07
Fix duplicated history entries when contains daylight saving time change

### LINKED ISSUE #4630: Duplicate Entries From History Call
#### Expected Behavior
When requesting historical data, each row in the resulting DataFrame should be unique.

#### Actual Behavior
When requesting history with
```
qb = QuantBook()
symbols = [ Symbol.Create(x, SecurityType.Forex, Market.Oanda) 
            for x in ["EURUSD","GBPUSD","USDCAD", "AUDUSD", "NZDUSD"] ]
qb.History(symbols, datetime(2010, 11, 5), datetime(2010, 11, 8), Resolution.Daily)
```
the resulting DataFrame contains duplicate entries, as seen below:

> EURUSD 8G  2011-11-03 20:00:00    1.382260
>            2011-11-04 20:00:00    1.378930
>            2011-11-04 20:00:00    1.378930
>            2011-11-06 20:00:00    1.378800

#### Potential Solution
Unknown.

#### Reproducing the Problem
See the backtest and notebook [here](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_0b1ebfd55ae4878b16209770daf7dced.html).

#### System Information
Run in the QC web terminal. Windows 10.

#### Checklist

AdalyatNazirov:
#### Description

Correct next.EndTime value according to daylight saving time shift.

#### Related Issue

Closes #4630 

#### Motivation and Context

Duplicated history items are return on the board of daylight saving time interval

N/A

Created regression test that fetch history for EURUSD, SPY for daily resolution. No duplicated items should be returned

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
+++ b/Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
@@ -24,12 +24,8 @@ using QuantConnect.Interfaces;
 namespace QuantConnect.Algorithm.CSharp
 {
     /// <summary>
-    /// Basic template algorithm simply initializes the date range and cash. This is a skeleton
-    /// framework you can use for designing an algorithm.
+    /// Regression test algorithm simply fetch history on boarder of Daylight Saving Time shift
     /// </summary>
-    /// <meta name="tag" content="using data" />
-    /// <meta name="tag" content="using quantconnect" />
-    /// <meta name="tag" content="trading and orders" />
     public class DaylightSavingTimeHistoryRegressionTest : QCAlgorithm, IRegressionAlgorithmDefinition
     {
         private Symbol[] _symbols = new[]
@@ -45,11 +41,7 @@ namespace QuantConnect.Algorithm.CSharp
             SetStartDate(2011, 11, 10);  //Set Start Date
             SetEndDate(2011, 11, 11);    //Set End Date
             SetCash(100000);             //Set Strategy Cash
-            
-            // Find more symbols here: http://quantconnect.com/data
-            // Forex, CFD, Equities Resolutions: Tick, Second, Minute, Hour, Daily.
-            // Futures Resolution: Tick, Second, Minute
-            // Options Resolution: Minute Only.
+
             for (int i = 0; i < _symbols.Length; i++)
             {
                 var symbol = _symbols[i];
@@ -64,9 +56,6 @@ namespace QuantConnect.Algorithm.CSharp
                     throw new Exception($"Duplicated bars were issued for time {time}");
                 }
             }
-
-            // There are other assets with similar methods. See "Selecting Options" etc for more details.
-            // AddFuture, AddForex, AddCfd, AddOption
         }
 
         /// <summary>
@@ -77,7 +66,7 @@ namespace QuantConnect.Algorithm.CSharp
         /// <summary>
         /// This is used by the regression test system to indicate which languages this algorithm is written in.
         /// </summary>
-        public Language[] Languages { get; } = { Language.CSharp, Language.Python };
+        public Language[] Languages { get; } = { Language.CSharp };
 
         /// <summary>
         /// This is used by the regression test system to indicate what the expected statistics are from running the algorithm
--- a/Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
+++ b/Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
@@ -1,0 +1,129 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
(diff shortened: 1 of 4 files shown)


Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
```diff
@@ -0,0 +1,129 @@
...
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Basic template algorithm simply initializes the date range and cash. This is a skeleton
+    /// framework you can use for designing an algorithm.
+    /// </summary>
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="using quantconnect" />
+    /// <meta name="tag" content="trading and orders" />
+    public class DaylightSavingTimeHistoryRegressionTest : QCAlgorithm, IRegressionAlgorithmDefinition
```
Martin-Molinero, line 33: nit - XML docs seem outdated, would be nice to explain and reference the issue being tested

Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
```diff
@@ -0,0 +1,129 @@
...
+                    .Where(g => g.Count() > 1);
+                if (duplications.Any())
+                {
+                    var time = duplications.First().Key;
+                    throw new Exception($"Duplicated bars were issued for time {time}");
+                }
+            }
+
+            // There are other assets with similar methods. See "Selecting Options" etc for more details.
+            // AddFuture, AddForex, AddCfd, AddOption
```
Martin-Molinero, line 69: nit - let's please remove these unrelated comments, same as above

Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        /// <summary>
+        /// This is used by the regression test system to indicate if the open source Lean repository has the required data to run this algorithm.
+        /// </summary>
+        public bool CanRunLocally { get; } = true;
+
+        /// <summary>
+        /// This is used by the regression test system to indicate which languages this algorithm is written in.
+        /// </summary>
+        public Language[] Languages { get; } = { Language.CSharp, Language.Python };
```
Martin-Molinero, line 80: Believe `Language.Python` python algorithm isn't implemented, it's okay, but think we should remove the language from here, else it would cause a failing python algo

Engine/DataFeeds/Enumerators/FillForwardEnumerator.cs
```diff
@@ -288,7 +290,8 @@ protected virtual bool RequiresFillForwardData(TimeSpan fillForwardResolution, B
                 var potentialUtc = _offsetProvider.ConvertToUtc(item.ReferenceDateTime) + item.Interval;
                 var potentialInTimeZone = _offsetProvider.ConvertFromUtc(potentialUtc);
                 var potentialBarEndTime = RoundDown(potentialInTimeZone, item.Interval);
-                if (potentialBarEndTime < next.EndTime)
+                var nextEndTime = _offsetProvider.ConvertFromUtc(previousTimeUtc) + nextPreviousTimeDelta + _dataResolution;
+                if (potentialBarEndTime < nextEndTime)
```
Martin-Molinero, line 294: Changes in this method are high risk and hard to understand 😅 , I believe we need more code comments about what's happening here for example: why `_offsetProvider.ConvertFromUtc(previousTimeUtc)` and not directly `previous.Time`? Why can't we use `next.EndTime` directly? Why `previous.Time + nextPreviousTimeDelta` instead of `previous.EndTime`? Else we will forget and change it back. Think we need simple unit tests covering these `why` too.

Algorithm.CSharp/DaylightSavingTimeHistoryRegressionTest.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            // There are other assets with similar methods. See "Selecting Options" etc for more details.
+            // AddFuture, AddForex, AddCfd, AddOption
+        }
+
+        /// <summary>
+        /// This is used by the regression test system to indicate if the open source Lean repository has the required data to run this algorithm.
+        /// </summary>
+        public bool CanRunLocally { get; } = true;
```
Martin-Molinero, line 75: Nice regression test 👍  thanks for getting it to run locally

Martin-Molinero (changes requested):
Nice one cracking this hard issue! 👍 Almost there, think we need a few comments explaining the fix and some simple unit tests covering it would be great

Engine/DataFeeds/Enumerators/FillForwardEnumerator.cs
```diff
@@ -293,8 +298,7 @@ protected virtual bool RequiresFillForwardData(TimeSpan fillForwardResolution, B
                 var potentialUtc = _offsetProvider.ConvertToUtc(item.ReferenceDateTime) + item.Interval;
                 var potentialInTimeZone = _offsetProvider.ConvertFromUtc(potentialUtc);
                 var potentialBarEndTime = RoundDown(potentialInTimeZone, item.Interval);
-                // apply the same timezone to next and potential bars
-                // calculate next endTime by adding duration to the previous data time because incoming next.EndTime is swallowing one hour
+                // apply the same timezone to next and potential bars because incoming next.EndTime can swallow one hour
                 var nextEndTime = next.EndTime - daylightMovement;
                 if (potentialBarEndTime < nextEndTime)
```
Martin-Molinero, line 303: Just a note: `potentialBarEndTime` in the other hand isn't affected by the daylight savings since it's generated from UTC after fixing a similar issue that caused an infinite loop situation see https://github.com/QuantConnect/Lean/issues/3707, and before this change `next.EndTime` was affected and it caused this issue

Martin-Molinero (approved):

Engine/DataFeeds/Enumerators/FillForwardEnumerator.cs
AdalyatNazirov: exactly. potentialBarEndTime is calculated using right _offsetProvider movement and didn't swallow hour anywhere

PR #4703, opened 2020-09-09
Bug 4487 Get Fundamental for CSharp

### LINKED ISSUE #4487: Implement New Overload For QuantBook.GetFundamental that Returns C# Object
#### Expected Behavior
Be able to use [`QuantBook.GetFundamental`](https://github.com/QuantConnect/Lean/blob/master/Research/QuantBook.cs#L151) in C# Notebooks

#### Actual Behavior
Since [`QuantBook.GetFundamental`](https://github.com/QuantConnect/Lean/blob/master/Research/QuantBook.cs#L151) returns a PyObject object, it's not usable in C# Notebooks.

#### Potential Solution
Add new overload that returns a C# object.

#### Reproducing the Problem
N/A

#### Checklist

C-SELLERS:
#### Description

Adding an additional overload for using GetFundamental() in C# notebooks. This also led to a refactor of Pythons GetFundamental so they could share the same base functions. I was able to maintain the original functionality for Python DataFrames as well, so the return type for calls from Python will still be a pandas dataframe

Changes:
- C# GetFundmental() that returns IEnumerable of DataDictionaries (A dictionary for each day)
- Multiple overloads for input types (String, Strings, Symbol, Symbols) *Python too!*
- GetAllFundamentals helper function that utilizes `FineFundamentalSubscriptionEnumerator`
- Parallel processing of Data Requests
- Option for full fundamental data obj (no selector, C# only!) 
- Pythons retrieval of Fundamental data is now shared with C# counterpart, reference private function GetAllFundamentals
- New PythonUtil function `ConvertToSymbols` which attempts to convert a given python input into a list of symbols.

Performance:

-Year long request of data for `["AAPL","AIG","BAC","GOOG","IBM"]` in the cloud
Master v9303: 18.769 Seconds
This PR: 7.077

-5 Year long request of data for `["AAPL","AIG","BAC","GOOG","IBM"]` in the cloud
Master v9303: 53.204 Seconds
This PR: 23.677

#### Related Issue

closes #4487

#### Motivation and Context

There was no usable GetFundamental function in C#

None

Lots of Unit tests for all types of requests.

Also try using these notebooks with the branch to see how it operates
[FundamentalsNotebookTests.zip](https://github.com/QuantConnect/Lean/files/5221575/FundamentalsNotebookTests.zip)

#### Types of changes

#### Checklist:


Research/QuantBook.cs
```diff
@@ -45,11 +47,37 @@ public class QuantBook : QCAlgorithm
     {
         private dynamic _pandas;
         private IDataCacheProvider _dataCacheProvider;
+        static private bool _isPythonNotebook;
```
Martin-Molinero, line 50: Minor but don't think this guy needs to be static anymore

Research/QuantBook.cs
```diff
@@ -45,11 +47,37 @@ public class QuantBook : QCAlgorithm
...
+                        "   print('No IPython installed')\n" +
+                        "   def IsPythonNotebook():\n" +
+                        "       return false\n").GetAttr("IsPythonNotebook").Invoke();
+                    isPython.TryConvert(out _isPythonNotebook);
+                }
+            }
+            catch
+            {
+                //Default to false
+                _isPythonNotebook = false;
```
Martin-Molinero, line 77: This shouldn't happen so we should probably log an error message here too

Research/QuantBook.cs
```diff
@@ -144,62 +172,144 @@ public QuantBook() : base()
...
-        /// <returns></returns>
-        public PyObject GetFundamental(PyObject tickers, string selector, DateTime? start = null, DateTime? end = null)
+        /// <returns>pandas DataFrame</returns>
+        public PyObject GetFundamental(PyObject input, string selector, DateTime? start = null, DateTime? end = null)
         {
+            //Null selector is not allowed for Python DataFrame
             if (string.IsNullOrWhiteSpace(selector))
             {
-                return "Invalid selector. Cannot be None, empty or consist only of white-space characters".ToPython();
+                throw new Exception("Invalid selector. Cannot be None, empty or consist only of white-space characters");
```
Martin-Molinero, line 187: Think would be better to use `ArgumentException` here instead, same for the ones at `PyConvertToSymbols`

Research/QuantBook.cs
```diff
@@ -144,62 +172,144 @@ public QuantBook() : base()
...
+                    }
 
-                        var factory = new TextSubscriptionDataSourceReader(_dataCacheProvider, config, date, false);
-                        var source = new SubscriptionDataSource(fileName.FullName, SubscriptionTransportMedium.LocalFile);
-                        var value = factory.Read(source).Select(x => GetPropertyValue(x, selector)).First();
+                    dayDictionary.Add(kvp.Key, valueToAdd);
 
-                        list.Add(Tuple.Create(symbol, date, value));
+                    // Add the dayDictionary to the results if it isn't already entered
+                    if (!result.Contains(dayDictionary))
```
Martin-Molinero, line 243: Let's circle back I think there is some room for improvement in this loop, specifically the `result.Find/Contains`. Maybe we could send into `GetAllFundamental` a data handler so we can specify different behavior for C#/Py so it's easier to return.

Research/QuantBook.cs
```diff
@@ -656,5 +766,119 @@ private object GetPropertyValue(object baseData, string fullName)
 
             return baseData;
         }
+
+        /// <summary>
+        /// Convert Python input to a list of Symbols
+        /// </summary>
+        /// <param name="input">Object with the desired property</param>
+        /// <returns>List of Symbols</returns>
+        private IEnumerable<Symbol> PyConvertToSymbols(PyObject input)
```
Martin-Molinero, line 775: Nice method! Thinking we might wanna move it into the Common extensions/utils section so it can be reused and we can add some simple tests for it covering his behavior

Research/QuantBook.cs
```diff
@@ -656,5 +766,119 @@ private object GetPropertyValue(object baseData, string fullName)
...
+                        symbol,
+                        Resolution.Daily,
+                        TimeZones.NewYork,
+                        TimeZones.NewYork,
+                        false,
+                        false,
+                        false
+                    );
+                var security = new Security(
+                    SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork),
```
Martin-Molinero, line 857: `SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork)` don't think we should be using this guy, it will cause `request.TradableDays` to go through all the days of the week, could we actually use an existing API like `Securities.CreateSecurity`?

Research/QuantBook.cs
```diff
@@ -656,5 +766,119 @@ private object GetPropertyValue(object baseData, string fullName)
...
+            //SubscriptionRequest does not except nullable DateTimes, so set a startTime and endTime
+            var startTime = start.HasValue ? (DateTime)start : QuantConnect.Time.BeginningOfTime;
+            var endTime = end.HasValue ? (DateTime)end : QuantConnect.Time.EndOfTime;
+
+            //Collection to store our results
+            var data = new Dictionary<Symbol, List<BaseData>>();
+
+            //Build factory
+            var factory = new FineFundamentalSubscriptionEnumeratorFactory(false);
+            var fileProvider = new DefaultDataProvider();
```
Martin-Molinero, line 840: We should be using the actual data provider instance here (see `algorithmHandlers.DataProvider`), it might not be the `DefaultDataProvider`

Research/QuantBook.cs
```diff
@@ -656,5 +766,119 @@ private object GetPropertyValue(object baseData, string fullName)
...
+                    SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork),
+                    config,
+                    new Cash(Currencies.USD, 0, 1),
+                    SymbolProperties.GetDefault(Currencies.USD),
+                    ErrorCurrencyConverter.Instance,
+                    RegisteredSecurityDataTypesProvider.Null,
+                    new SecurityCache()
+                );
+                var request = new SubscriptionRequest(false, null, security, config, startTime, endTime);
+                var enumerator = factory.CreateEnumerator(request, fileProvider);
```
Martin-Molinero, line 866: Think we should wrap the `enumerator` around a `using` statement so it's properly disposed of

Tests/Research/QuantBookFundamentalTests.cs
```diff
@@ -0,0 +1,167 @@
...
+using System;
+using System.Collections.Generic;
+using QuantConnect.Data.Market;
+using QuantConnect.Research;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Tests.Research
+{
+    [TestFixture]
+    public class QuantBookFundamentalTests
```
Martin-Molinero, line 27: Sweet tests! 👍

Tests/Research/QuantBookFundamentalTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            _endDate = new DateTime(2014, 4, 1);
+
+            using (Py.GIL())
+            {
+                _module = Py.Import("Test_QuantBookFundamental");
+            }
+        }
+
+        // All 4 of the accepted types of input for fundamental data request, with varying selectors
+        private static readonly object[] FundamentalRequestTestCases =
```
Martin-Molinero, line 50: nit - code style is to place privates at the bottom. First setup methods, second all tests and finally any private helper classes/methods

Tests/Research/RegressionScripts/Test_QuantBookFundamental.py
```diff
@@ -0,0 +1,32 @@
...
+AddReference("QuantConnect.Common")
+
+from System import *
+from QuantConnect import *
+from QuantConnect.Data import *
+from QuantConnect.Research import *
+from datetime import datetime, timedelta
+import pandas as pd
+
+class FundamentalHistoryTest():
```
Martin-Molinero, line 27: Wondering if we should just join this guy with the existing `Test_QuantBookHistory.py`

Martin-Molinero (changes requested):
Nice! Almost there, some final tweaks and we should be good to go

Tests/Research/QuantBookFundamentalTests.cs
C-SELLERS: Thank you!

Research/readme.md
```diff
@@ -1,9 +1,39 @@
 ﻿QuantConnect Research Project
 =============
+Currently we have two ways to use QuantConnect research notebooks, you can either install and run locally or just use our docker image (Recommended).
 
-There is an up to date docker image available at [quantconnect/research](https://hub.docker.com/repository/docker/quantconnect/research). You can pull this image with `docker pull quantconnect/research`.
+The up to date docker image available at [quantconnect/research](https://hub.docker.com/repository/docker/quantconnect/research). You can pull this image with `docker pull quantconnect/research`.
 
 
+# Running Jupyter Locally
```
Martin-Molinero, line 8: nit - I would move running locally after running on the docker image since we prefer users using that approach

Research/QuantBook.cs
Martin-Molinero: Think this guy still has to be replaced by the new `private IDataProvider _dataProvider;`

Research/QuantBook.cs
```diff
@@ -144,62 +175,122 @@ public QuantBook() : base()
...
-                if (!PyList.IsListType(tickers))
+                var data = new PyDict();
+                foreach (var day in fundamentalData.OrderBy(x => x.Key))
                 {
-                    var tmp = new PyList();
-                    tmp.Append(tickers);
-                    tickers = tmp;
+                    var orderedValues = day.Value.OrderBy(x => x.Key.ID.ToString());
+                    var columns = orderedValues.Select(x => x.Key.ID.ToString());
+                    var values = orderedValues.Select(x => x.Value);
```
Martin-Molinero, line 206: We should double-check these two selects are not causing the OrderBy to happen twice, maybe it has a cache? 🤔

Common/Util/PythonUtil.cs
```diff
@@ -228,5 +228,57 @@ private static PyObject GetModule()
...
+                    symbol = QuantConnect.Symbol.Create(symbolString, SecurityType.Equity, Market.USA);
+                    symbolsList = new List<Symbol> { symbol };
+                }
+                else if (input.TryConvert(out symbol))
+                {
+                    symbolsList = new List<Symbol> { symbol };
+                }
+                else
+                {
+                    throw new Exception($"Cannot convert object {input.Repr()} to symbol");
```
Martin-Molinero, line 278: nit - `ArgumentException` 😛

Martin-Molinero (commented):
Minor final tweaks! 👍

Common/Util/PythonUtil.cs
C-SELLERS: Whoops!

Research/QuantBook.cs
C-SELLERS: True. I think it might very well be happening twice so I will use orderedValues as a list instead to avoid the possible multiple enumeration.

Martin-Molinero (approved):
Sweet! nice performance improvement too 🎉

PR #4728, opened 2020-09-17
Improve Error Message For Arithmetic Overflow In Decimal Cast of Double

### LINKED ISSUE #4726: Arithmetic Overflow Exception at CalculateQuantity Double Overload
#### Expected Behavior
No unhandled exceptions.

In the particular case that has exhibited this exception, there is a successful history request:
```python
df = self.History(symbol, 100000, Resolution.Minute).open[:-1]   # success or .open would raise an exception
``` 
after that, the order quantity is calculated with a target that derived from it. Since the calculations use `df.pct_change(100)`, it might have created NaN.

#### Potential Solution
Use `SafeDecimalCast()` instead of `(decimal)` cast at least in all methods at QCAlgorithm.Trading.cs.

#### Reproducing the Problem
```python
target = float('nan')   # or sys.float_info.max
qnty = self.CalculateOrderQuantity(t, target)
```
#### Checklist

AlexCatarino:
#### Description
- SafeDecimalCast Throws Exception For Non-Finite Numbers
- Fixes Arithmetic Overflow Exception in QCAlgorithm.Trading Methods
    - Replace decimal cast for `SafeDecimalCast()`.
    - If the algorithm uses a non-finite number in QCAlgorithm trading methods, it will throw with an user-frieldly exception message.
- Fixes KellyCriterionProbabilityValue Calculation

#### Related Issue
Closes #4726

#### Motivation and Context
Improves exception message.

Unit and regression tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Statistics/KellyCriterionManager.cs
+++ b/Common/Statistics/KellyCriterionManager.cs
@@ -108,7 +108,7 @@ namespace QuantConnect.Statistics
                         (1 / variance + 2 * averagePowered / variancePowered)
                         / _insightValues.Count - 1);
 
-                    KellyCriterionProbabilityValue =
+                    KellyCriterionProbabilityValue = kellyCriterionStandardDeviation.IsNaNOrZero() ? 1 :
                         1 - _normalDistribution.CumulativeDistribution(kellyCriterionEstimate / kellyCriterionStandardDeviation)
                             .SafeDecimalCast();
                 }
--- a/Algorithm/QCAlgorithm.Trading.cs
+++ b/Algorithm/QCAlgorithm.Trading.cs
@@ -54,7 +54,7 @@ namespace QuantConnect.Algorithm
         /// <seealso cref="Buy(Symbol, decimal)"/>
         public OrderTicket Buy(Symbol symbol, double quantity)
         {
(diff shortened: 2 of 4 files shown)


Common/Statistics/KellyCriterionManager.cs
```diff
@@ -108,7 +108,7 @@ public void UpdateScores()
                         (1 / variance + 2 * averagePowered / variancePowered)
                         / _insightValues.Count - 1);
 
-                    KellyCriterionProbabilityValue =
+                    KellyCriterionProbabilityValue = kellyCriterionStandardDeviation.IsNaNOrZero() ? 1 :
```
Martin-Molinero, line 111: nit - before we were returning `0` for Nan and Zero (old `SafeDecimalCast`), think we should maintain existing behavior here?

Martin-Molinero (approved):

Common/Statistics/KellyCriterionManager.cs
AlexCatarino: This change maintains existing behavior.  When `kellyCriterionStandardDeviation` is NaN or Zero, the CFD calculation is also NaN or Zero, and `KellyCriterionProbabilityValue` is 1.

PR #4742, opened 2020-09-22
Bug 4722: Prevent Repetitive Factor File Numerical Precision Warnings

### LINKED ISSUE #4722: Prevent Repetitive Factor File Warning
#### Expected Behavior
The following message is only logged once per security:
> "Data for symbol DRYS has been limited due to numerical precision issues in the factor file. The starting date has been set to 8/11/2016."

#### Actual Behavior
The message above is logged multiple times.

#### Potential Solution
Use the pattern that is in place for map files.
Maybe we should also improve the message, since "The starting date has been set to 31/12/2050." troubles users. :-D

#### Reproducing the Problem
Any universe selection model that selected DRYS for example.

#### Checklist

mchandschuh:
This change aims to remove some debug message spam observed when a security is added/removed multiple times with a factor file with INF entries forcing the 'starting date' to be in the future. We accomplish this by tracking which symbols have raised the numerical precision warning event and only passing the message through to the result handler the first time for each symbol.

#### Description

* Add DataProviderEventArgs base class for IDataProviderEvents event args

> This base class includes a Symbol property. This will empower event listeners
to make decisions based on which security (symbol) raised the event. The immediate
use case is preventing multiple numerical precision messages for the same security.
This pattern can equally be applied to other error messages that are raised each
time a security is added to a universe.

* Update ConcurrentSet.Add to use ISet<T>.Add returning bool

> It's a very common pattern to use if (set.Add(item)) which is enabled
via bool ISet<T>.Add(item) but not enabled via void ICollectiont<T>.Add(item).
This change simples changes the default Add implementation to use the ISet<T>
overload and relegates the ICollection<T>.Add implementation to be explicit.

* Prevent multiple numerical precision messages for same symbol

> If a security is continually added/removed from a universe, then the user will
see this message each time the security is added. This results in some spam.
This change simply remembers for which symbols we've notified the user about the
numerical precision issue.

#### Related Issue

Fixes #4722

#### Motivation and Context

Limit debug message spam in console

No.

Manually tested using DRYS over a period of 2016-03-01 to 2017-03-01 w/ a manual universe that adds/removes DRYS each month. This was run against current `origin/master` where the message spam was observed and again against `bug-4722-prevent-repetitive-factor-file-warning` where the message was only emitted once.

RE: Unit tests:
The `SubscriptionDataReaderSubscriptionEnumeratorFactory` class is a beast to unit test due to multiple varying dependencies. It additionally uses the `new` operator directly and even invokes a static function, `CorporateEventEnumeratorFactory.CreateEnumerators`. In order to create unit tests without needing to mock the entire world (likely making the tests incredibly fragile), this type would need to be decomposed. For one, a factory could be provided for creating the `SubscriptionDataReader`, thereby removing some of the dependencies. Additionally, the result of the `CorporateEventEnumeratorFactory.CreateEnumerators` could be composed externally to this factory type and encapsulated in yet another factory type. In theory, factory types exist as an abstraction over the `new` operator and it looks like this factory has been overloaded, 'newing up' multiple objects (in this case enumerator stacks), making unit testing very involved.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/DataProviderEvents.cs
+++ b/Common/DataProviderEvents.cs
@@ -21,9 +21,29 @@ using QuantConnect.Securities;
 namespace QuantConnect
 {
     /// <summary>
+    /// Defines a base class for <see cref="IDataProviderEvents"/>
+    /// </summary>
+    public abstract class DataProviderEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Gets the symbol being processed that generated the event
+        /// </summary>
+        public Symbol Symbol { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataProviderEventArgs"/> class
+        /// </summary>
+        /// <param name="symbol">Symbol being processed that generated the event</param>
+        protected DataProviderEventArgs(Symbol symbol)
+        {
+            Symbol = symbol;
+        }
+    }
+
+    /// <summary>
     /// Event arguments for the <see cref="IDataProviderEvents.InvalidConfigurationDetected"/> event
     /// </summary>
-    public sealed class InvalidConfigurationDetectedEventArgs : EventArgs
+    public sealed class InvalidConfigurationDetectedEventArgs : DataProviderEventArgs
     {
         /// <summary>
         /// Gets the error message
@@ -33,8 +53,10 @@ namespace QuantConnect
         /// <summary>
         /// Initializes a new instance of the <see cref="InvalidConfigurationDetectedEventArgs"/> class
         /// </summary>
+        /// <param name="symbol">Symbol being processed that generated the event</param>
         /// <param name="message">The error message</param>
-        public InvalidConfigurationDetectedEventArgs(string message)
+        public InvalidConfigurationDetectedEventArgs(Symbol symbol, string message)
+            : base(symbol)
         {
             Message = message;
         }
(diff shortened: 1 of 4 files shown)


mchandschuh:
Looks like the travis push build failed via:

```
1) Failed : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes<TiingoNews>(AAPL.TiingoNews R735QTJ8XC9W,Hour,1,0,0,0,0,48,False, 0: ¤0.00)
  Expected: 48
  But was:  47
  at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTypes[T] (QuantConnect.Symbol symbol, QuantConnect.Resolution resolution, System.Int32 days, System.Int32 expectedTicksReceived, System.Int32 expectedTradeBarsReceived, System.Int32 expectedQuoteBarsReceived, System.Int32 expectedAuxPointsReceived, System.Int32 expectedCustomPointsReceived, System.Boolean shouldThrowException, T customDataType) [0x00000] in <df16975dd96c448386d8e24e2e08ef8f>:0 
```

This appears to be a flickering test as it passed in the PR and locally but not in the push.

Martin-Molinero (approved):

Issue #4731, opened 2020-09-18
Fix Option Expiration Tags


haxdds:
#### Expected Behavior

1. When a long OTM option expires, its tag should read option expiration.

2. When a short OTM option expires, its orderevent.IsAssignment should be False.

#### Actual Behavior

1, When a long OTM option expires, the corresponding tag reds `Automatic option exercise on expiration - Adjusting(or removing) the exercised/assigned option`. 

2. When a short OTM option expires, its orderevent.IsAssignment is set to True although no assignment orders take place.

#### Potential Solution

1. Changing the tag for the options expiration orderevent.

#### Reproducing the Problem

1, See this [backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_4475aa9b5fef0a6a3f9e8e72a9ab15b3.html)

2. See this [backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_cd20e2d61054e12c8a73694a12e5200b.html)

#### System Information

Windows 10

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #4743, opened 2020-09-22
Bug 4731 Fix Option Expiration Order Tags and Order Event IsAssignment Flag

mchandschuh:
Standardizes order tags and order events related to option exercise/assignment. All orders/events related to option expiry are now prefix with `Option Expired` and individual events include the following:

OTM: OTM
ITM & long: Automatic Exercise
ITM & short: Automatic Assignment

Since the transaction handler combines the order tag and the order event message, we end up with the following:

Option Expired - OTM
Option Expired - Automatic Exercise
Option Expired - Automatic Assignment

The `OrderEvent.IsAssignment` flag was also updated to only be true in ITM assignment cases.

#### Description

* Add OrderRight.GetExerciseDirection(isShort) extension

> Returns the OrderDirection resulting from exercise/assignment of a particular
option right

* Fix option exercise/assignment order tags and order event messages

> The algorithm manager was doing work to determine whether or not the option ended
in exercise or assignment at expiration. This decision should be left for the exercise
model to decide -- from the algorithm manager's perspective, all that matters is that
the option was expired. The DefaultExerciseModel was updated to properly track whether
the option expired with automatic assignment or exercise, dependending on whether or
not we wrote or bought the option (held liability or right, respectively). Updated unit
tests to check for order event counts and order event messages for option exercise cases.

#### Related Issue

Fixes: #4731

#### Motivation and Context

Fixes inaccurate order tags and order event is assignment property

No.

Existing unit tests in `SecurityPortfolioManagerTests` expanded to test for the new `OrderEvent.Message` values as well as the `IsAssignment` flag value.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Orders/OptionExercise/DefaultExerciseModel.cs
+++ b/Common/Orders/OptionExercise/DefaultExerciseModel.cs
@@ -32,52 +32,52 @@ namespace QuantConnect.Orders.OptionExercise
         /// <param name="order">Order to update</param>
         public IEnumerable<OrderEvent> OptionExercise(Option option, OptionExerciseOrder order)
         {
+            var underlying = option.Underlying;
+
+            var isShort = order.Quantity < 0;
             var utcTime = option.LocalTime.ConvertToUtc(option.Exchange.TimeZone);
 
-            var optionQuantity = order.Quantity;
-            var assignment = order.Quantity < 0;
-            var underlying = option.Underlying;
-            var exercisePrice = order.Price;
-            var fillQuantity = option.GetExerciseQuantity(order.Quantity);
-            var exerciseQuantity =
-                    option.Symbol.ID.OptionRight == OptionRight.Call ? fillQuantity : -fillQuantity;
-            var exerciseDirection = assignment?
-                    (option.Symbol.ID.OptionRight == OptionRight.Call ? OrderDirection.Sell : OrderDirection.Buy):
-                    (option.Symbol.ID.OptionRight == OptionRight.Call ? OrderDirection.Buy : OrderDirection.Sell);
+            var inTheMoney = option.IsAutoExercised(underlying.Close);
 
-            var addUnderlyingEvent = new OrderEvent(order.Id,
-                            underlying.Symbol,
-                            utcTime,
-                            OrderStatus.Filled,
-                            exerciseDirection,
-                            exercisePrice,
-                            exerciseQuantity,
-                            OrderFee.Zero,
-                            "Option Exercise/Assignment");
+            // we're assigned only if we get exercised against, meaning we wrote the option and it was exercised by the buyer
+            var isAssignment = inTheMoney && isShort;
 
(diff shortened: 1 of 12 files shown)


Common/Orders/OptionExercise/DefaultExerciseModel.cs
```diff
@@ -32,52 +32,56 @@ public class DefaultExerciseModel : IOptionExerciseModel
...
+            if (inTheMoney && option.ExerciseSettlement == SettlementType.PhysicalDelivery)
             {
-                return new[] { optionRemoveEvent, addUnderlyingEvent };
-            }
+                var right = option.Symbol.ID.OptionRight;
 
-            return new[] { optionRemoveEvent };
-        }
+                // TODO : Why doesn't this method take into account the directionality of the quantity like other quantity properties/methods
+                var fillQuantity = option.GetExerciseQuantity(order.Quantity);
```
Martin-Molinero, line 68: > // TODO : Why doesn't this method take into account the directionality of the quantity like other quantity properties/methods

You are referring to the required `right == OptionRight.Call ? fillQuantity : -fillQuantity`? All consumers of `GetExerciseQuantity` seem to be doing the same operation except for `OptionExerciseOrder` that seems like a bug? 🤔 Agree it kinda feels it should be internal to the `GetExerciseQuantity`

Martin-Molinero (approved):
Nice 👍, think we are missing to update a couple of unit tests.

Common/Orders/OptionExercise/DefaultExerciseModel.cs
mchandschuh: my thoughts exactly -- throughout lean we use `Quantity` to mean the directional quantity and `AbsoluteQuantity` to mean the magnitude of the quantity. I'll put an additional commit in here refactoring this method to follow this convention.

mchandschuh:
I pushed the changes that fix this immediate issue. I'm continuing to investigating inconsistencies related to `OptionExerciseOrder.Quantity` -- The `TODO` in this commit is among the issues being investigated. I've also decided to _not_ include the `Option.GetExerciseQuantity` method in this commit since it's implementation needed to be logically incorrect in order to maintain existing behavior. I'll be opening a separate PR that includes this new method along with other changes aimed at improving the consistency and usage of `OptionExerciseOrder.Quantity` and explicitly removing the _positive only_ soft requirement on that value (it wasn't being obeyed in most parts of the code anyway).

Martin-Molinero:
>  I've also decided to not include the Option.GetExerciseQuantity method in this commit since it's implementation needed to be logically incorrect in order to maintain existing behavior. I'll be opening a separate PR that includes this new method along with other changes

Sounds good 👍 (issue at https://github.com/QuantConnect/Lean/issues/4766). there are a couple of python unit test which need an update for this PR `PythonOptionTests.FilterReturnsUniverseRegression` & `PythonOptionTests.FilterReturnsListRegression`

Issue #4686, opened 2020-09-03
Remove YahooDataDownloader from LEAN ToolBox


Martin-Molinero:
#### Expected Behavior

- YahooDataDownloader does not fail to download Split & dividends.
#### Actual Behavior

- YahooDataDownloader failing download split & dividend.
```

System.FormatException : Input string was not in a correct format.
   at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
   at System.Number.ParseDecimal(String value, NumberStyles options, NumberFormatInfo numfmt)
   at QuantConnect.Parse.Decimal(String value) in D:\QuantConnect\MyLean\Lean\Common\Parse.cs:line 126
   at QuantConnect.ToolBox.YahooDownloader.YahooDataDownloader.DownloadSplitAndDividendData(Symbol symbol, DateTime startUtc, DateTime endUtc) in D:\QuantConnect\MyLean\Lean\ToolBox\YahooDownloader\YahooDataDownloader.cs:line 113
   at QuantConnect.Tests.Common.Util.FactorFileGeneratorTests.FactorFiles_CanBeGenerated_Accurately() in D:\QuantConnect\MyLean\Lean\Tests\Common\Util\FactorFileGeneratorTests.cs:line 73

```
#### Potential Solution

- Review format & update YahooDataDownloader  
#### Reproducing the Problem

- See `FactorFileGeneratorTests`
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

jamesculshaw:
Hi,

I would like to do this if its still available

PR #4750, opened 2020-09-23
Bug 4686 Fix yahoodatadownloader failing download split and dividend

nazbrok:
#### Description

The decimal parsing in the method YahooDataDownloader.DownloadSplitAndDividendData was throwing and exception when the number was in the exponential format (9.8E-4)

#### Related Issue

related issue : https://github.com/QuantConnect/Lean/issues/4686

#### Motivation and Context

Was not working

No

Unit test

#### Types of changes

#### Checklist:

- [x ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).

### DIFF
--- a/ToolBox/YahooDownloader/YahooDataDownloader.cs
+++ b/ToolBox/YahooDownloader/YahooDataDownloader.cs
@@ -113,7 +113,7 @@ namespace QuantConnect.ToolBox.YahooDownloader
                         parsed.Add(new Dividend
                         {
                             Time = Parse.DateTimeExact(values[0].Replace("-", string.Empty), DateFormat.EightCharacter),
-                            Value = Parse.Decimal(values[1])
+                            Value = Parse.Decimal(values[1], System.Globalization.NumberStyles.Float)
                         });
                     }
                 }


Martin-Molinero:
Hey @nazbrok, Thanks for the fix! `FactorFiles_CanBeGenerated_Accurately` test seems to still be failing with this fix?

nazbrok:
@Martin-Molinero I don't think this is related. The fix was about the conversion issue during the parsing of Dividend values from Yahoo (https://query1.finance.yahoo.com/v7/finance/download/AAPL?period1=0&period2=253402300800&interval=1d&events=dividend&crumb=19LN9kd9jlN).

The test you are refering compare calculated PriceFactor values (indeed from the downloaded yahoo dividend data) to those stored in https://github.com/QuantConnect/Lean/blob/96da867109296a2fc44e78a72f6f908ab4fa791a/Data/equity/usa/factor_files/aapl.csv

The calculated values are different and since this part of code didn't move from ages, as far as I can see, nor the aapl.csv file I wonder if the data returned by Yahoo changed (is that possible ?).

Here is the formula used to calculate the PriceFactor

https://github.com/QuantConnect/Lean/blob/96da867109296a2fc44e78a72f6f908ab4fa791a/ToolBox/FactorFileGenerator.cs#L229-L233

In our case, the first one calculated if for the date 10/05/2018.

The data used from Yahoo are : 

```
Date | Dividends
11/05/18 | 0.1825
```
The previousClosingPrice.Close is 190.03

Since it's the first value calculated, here `previousFactorFileRow ` is initialized like this

https://github.com/QuantConnect/Lean/blob/96da867109296a2fc44e78a72f6f908ab4fa791a/ToolBox/FactorFileGenerator.cs#L76-L80

So according to the PriceFactor formula we have 

`PriceFactor = 1 - 0.1825 / 1 / 190.03 / 1 = **0.9990396**`

While in AAPL.csv we have for this date a PriceFactor of **0.9961585**

image



According to those sites the historical dividend for this date is not 0.1825 like yahoo told us but 0.73
 https://www.nasdaq.com/market-activity/stocks/aapl/dividend-history
https://www.streetinsider.com/dividend_history.php?q=AAPL

image

If I calculate the price factor using $0.73 I get 

`PriceFactor = 1 - 0.73 / 1 / 190.03 / 1 = **0.9961585**`

Which is our calculated pricefactor present in AAPL.csv


I checked on Yahoo's page if the data shown were the same as returned and it is

image
 

So in short, to me Yahoo is not reliable for dividends historical data on AAPL.
I think my PR correctly fix the initial bug but there is another issue to create for the Yahoo dividend historical data for AAPL

jaredbroad:
Great analysis @nazbrok. I 💯 agree yahoo data is not reliable but your fix seems solid for the E-issue. 👍

Martin-Molinero (approved):
Thanks! 👍

PR #4797, opened 2020-10-01
Include Order.Tag/OrderEvent.Message in their ToString, Fix default tag values

### LINKED ISSUE #4795: Order.Tag in ToString; Default Order.Tag Value; OrderEvent.IsAssignment in ToString
#### Expected Behavior

* `Order.ToString()` includes the tag.
* `LimitOrder`/`StopMarketOrder`/`StopLimitOrder` default their `Tag` properties when the `tag` constructor parameter is null or empty
* `OrderEvent.ToString()` includes `IsAssignment` when security type is option

#### Actual Behavior

* [`Order.ToString()`](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/Order.cs#L295) does not include the order tag.
* [`LimitOrder`](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/LimitOrder.cs#L62)/[`StopMarketOrder`](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/StopLimitOrder.cs#L74)/[`StopLimitOrder`](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/StopMarketOrder.cs#L62) default their `Tag` properties to include their limit/stop prices when the `tag` constructor parameter only when empty.
* [`OrderEvent.IsAssignment`](https://github.com/QuantConnect/Lean/blob/master/Common/Orders/OrderEvent.cs#L254) is never included in `OrderEvent.ToString()`

#### Potential Solution

* Add `Order.Tag` to the end of `Order.ToString()` _if_ it's not null or empty
* Update constructor checks to use `string.IsNullOrEmpty(tag)`
* Append `IsAssignment: {IsAssignment}` at the end of `OrderEvent.ToString()` when `Symbol.SecurityType == SecurityType.Option`

#### Checklist

mchandschuh:
⚠️ This PR is on top of #4796 due to regression changes in both algorithms. Both changes only impact order list hash values.

#### Description

*Include Order.Tag/OrderEvent.Message in their ToString, Fix default tag values

>There was inconsistencies in what we were checking for. The order constructors
default the tag parameter to an empty string but Order.CreateOrder checks for
a null string. Additionally, the order constructors (limit,stopmarket,stoplimit)
would check for an empty string and if so, apply a default order tag.

>This change cleans these checks up using string.IsNullOrEmpty and also removes the
check from Order.CreateOrder since we're passing the tag into the various order
constructors.

#### Related Issue

Fixes #4795

#### Motivation and Context

Ensure usage of default order tags and improve debugging experience by including them in the `ToString()` methods.

No.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -2263,5 +2263,22 @@ namespace QuantConnect
                     return isShort ? OrderDirection.Buy : OrderDirection.Sell;
             }
         }
+
+        /// <summary>
+        /// Gets the <see cref="OrderDirection"/> for the specified <paramref name="quantity"/>
+        /// </summary>
+        public static OrderDirection GetOrderDirection(decimal quantity)
+        {
+            var sign = Math.Sign(quantity);
+            switch (sign)
+            {
+                case 1:  return OrderDirection.Buy;
+                case 0:  return OrderDirection.Hold;
+                case -1: return OrderDirection.Sell;
+                default: throw new ApplicationException(
+                    $"The skies are falling and the oceans are rising! Math.Sign({quantity}) returned {sign} :/"
+                );
+            }
+        }
     }
 }
--- a/Common/Orders/Order.cs
+++ b/Common/Orders/Order.cs
@@ -292,7 +292,8 @@ namespace QuantConnect.Orders
         /// <filterpriority>2</filterpriority>
         public override string ToString()
         {
-            return Invariant($"OrderId: {Id} (BrokerId: {string.Join(",", BrokerId)}) {Status} {Type} order for {Quantity} unit{(Quantity == 1 ? "" : "s")} of {Symbol}");
+            var tag = string.IsNullOrEmpty(Tag) ? string.Empty : $": {Tag}";
+            return Invariant($"OrderId: {Id} (BrokerId: {string.Join(",", BrokerId)}) {Status} {Type} order for {Quantity} unit{(Quantity == 1 ? "" : "s")} of {Symbol}{tag}");
         }
 
         /// <summary>
@@ -433,10 +434,6 @@ namespace QuantConnect.Orders
             }
             order.Status = OrderStatus.New;
             order.Id = orderId;
-            if (tag != null)
-            {
-                order.Tag = tag;
-            }
             return order;
         }
     }
--- a/Common/Securities/Option/Option.cs
+++ b/Common/Securities/Option/Option.cs
@@ -22,8 +22,8 @@ using QuantConnect.Orders.OptionExercise;
 using Python.Runtime;
 using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
-using QuantConnect.Util;
 using System.Collections.Generic;
+using QuantConnect.Orders;
 
 namespace QuantConnect.Securities.Option
 {
@@ -226,13 +226,45 @@ namespace QuantConnect.Securities.Option
         }
 
         /// <summary>
-        /// Returns the actual number of the underlying shares that are going to change hands on exercise. For instance, after reverse split
-        /// we may have 1 option contract with multiplier of 100 with right to buy/sell only 50 shares of underlying stock.
+        /// Returns the directional quantity of underlying shares that are going to change hands on exercise/assignment of all
+        /// contracts held by this account, taking into account the contract's <see cref="Right"/> as well as the contract's current
+        /// <see cref="ContractUnitOfTrade"/>, which may have recently changed due to a split/reverse split in the underlying security.
         /// </summary>
-        /// <returns></returns>
-        public decimal GetExerciseQuantity(decimal quantity)
+        /// <remarks>
+        /// Long option positions result in exercise while short option positions result in assignment. This function uses the term
+        /// exercise loosely to refer to both situations.
+        /// </remarks>
+        public decimal GetExerciseQuantity()
         {
-            return quantity * ContractUnitOfTrade;
+            // negate Holdings.Quantity to match an equivalent order
+            return GetExerciseQuantity(-Holdings.Quantity);
+        }
+
+        /// <summary>
+        /// Returns the directional quantity of underlying shares that are going to change hands on exercise/assignment of the
+        /// specified <paramref name="exerciseOrderQuantity"/>, taking into account the contract's <see cref="Right"/> as well
+        /// as the contract's current <see cref="ContractUnitOfTrade"/>, which may have recently changed due to a split/reverse
+        /// split in the underlying security.
+        /// </summary>
+        /// <remarks>
+        /// Long option positions result in exercise while short option positions result in assignment. This function uses the term
+        /// exercise loosely to refer to both situations.
+        /// </remarks>
+        /// <paramref name="exerciseOrderQuantity">The quantity of contracts being exercised as provided by the <see cref="OptionExerciseOrder"/>.
+        /// A negative value indicates exercise (we are long and the order quantity is negative to bring us (closer) to zero.
+        /// A positive value indicates assignment (we are short and the order quantity is positive to bring us (closer) to zero.</paramref>
+        public decimal GetExerciseQuantity(decimal exerciseOrderQuantity)
+        {
+            // when exerciseOrderQuantity > 0 [ we are short ]
+            //      && right == call => we sell to contract holder  => negative
+            //      && right == put  => we buy from contract holder => positive
+
+            // when exerciseOrderQuantity < 0 [ we are long ]
+            //      && right == call => we buy from contract holder => positive
+            //      && right == put  => we sell to contract holder  => negative
+
+            var sign = Right == OptionRight.Call ? -1 : 1;
+            return sign * exerciseOrderQuantity * ContractUnitOfTrade;
         }
 
         /// <summary>
@@ -437,7 +469,7 @@ namespace QuantConnect.Securities.Option
         public void SetFilter(PyObject universeFunc)
         {
             ContractFilter = new FuncSecurityDerivativeFilter(universe =>
-            {              
+            {
                 var optionUniverse = universe as OptionFilterUniverse;
                 using (Py.GIL())
                 {
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -1032,34 +1032,36 @@ namespace QuantConnect.Lean.Engine
             {
                 // check if we are holding position
                 var security = algorithm.Securities[delistings[i].Symbol];
-                if (security.Holdings.Quantity == 0) continue;
+                if (security.Holdings.Quantity == 0)
+                {
+                    continue;
+                }
 
                 // check if the time has come for delisting
                 var delistingTime = delistings[i].Time;
                 var nextMarketOpen = security.Exchange.Hours.GetNextMarketOpen(delistingTime, false);
                 var nextMarketClose = security.Exchange.Hours.GetNextMarketClose(nextMarketOpen, false);
 
-                if (security.LocalTime < nextMarketClose) continue;
+                if (security.LocalTime < nextMarketClose)
+                {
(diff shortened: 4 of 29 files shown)


Engine/AlgorithmManager.cs
```diff
@@ -1032,34 +1032,36 @@ private static void ProcessDelistedSymbols(IAlgorithm algorithm, List<Delisting>
...
-                    request = new SubmitOrderRequest(OrderType.Market, security.Type, security.Symbol,
-                        -security.Holdings.Quantity, 0, 0, algorithm.UtcTime, "Liquidate from delisting");
+                    // tx handler will determine auto exercise/assignment
+                    tag = "Option Expired";
+                    orderType = OrderType.OptionExercise;
                 }
 
-                algorithm.Transactions.ProcessRequest(request);
+                // submit an order to liquidate on market close or exercise (for options)
+                var request = new SubmitOrderRequest(orderType, security.Type, security.Symbol,
```
Martin-Molinero, line 1060: Nice to this standarized! 👍

Engine/Results/RegressionResultHandler.cs
```diff
@@ -167,6 +188,124 @@ public override void SetAlphaRuntimeStatistics(AlphaRuntimeStatistics statistics
...
+
+        /// <summary>
+        /// Set the current runtime statistics of the algorithm.
+        /// These are banner/title statistics which show at the top of the live trading results.
+        /// </summary>
+        /// <param name="key">Runtime headline statistic name</param>
+        /// <param name="value">Runtime headline statistic value</param>
+        public override void RuntimeStatistic(string key, string value)
+        {
+            base.RuntimeStatistic(key, value);
```
Martin-Molinero, line 269: I think this `override` seems unrequired?

Engine/Results/RegressionResultHandler.cs
```diff
@@ -167,6 +188,124 @@ public override void SetAlphaRuntimeStatistics(AlphaRuntimeStatistics statistics
...
+                WriteLine($"OnSecuritiesChanged:ADD: {string.Join("|", added)}");
+            }
+
+            if (changes.RemovedSecurities.Count > 0)
+            {
+                var removed = changes.RemovedSecurities
+                    .Select(security => security.Symbol.ToString())
+                    .OrderBy(symbol => symbol);
+
+                WriteLine($"OnSecuritiesChanged:REM: {string.Join("|", removed)}");
```
Martin-Molinero, line 305: `SecurityChanges` already has a nice `ToString()` implementation maybe we could directly use that?

Engine/Results/RegressionResultHandler.cs
```diff
@@ -76,7 +85,13 @@ public override void SetAlgorithm(IAlgorithm algorithm, decimal startingPortfoli
             }
 
             _writer = new StreamWriter(LogFilePath);
```
Martin-Molinero, line 87: I think we might want to take the lock on `_sync` here to avoid any race condition leaving some log remaining in the `_preInitializeLines`?

Common/Securities/SecurityPortfolioModel.cs
```diff
@@ -38,7 +38,7 @@ public virtual void ProcessFill(SecurityPortfolioManager portfolio, Security sec
             var quoteCash = security.QuoteCurrency;
 
             //Get the required information from the vehicle this order will affect
-            var isLong = security.Holdings.IsLong;
+            var isLong = security.Holdings.IsLong;
```
Martin-Molinero, line 41: nit - typo

Tests/Common/Securities/SecurityPortfolioManagerTests.cs
```diff
@@ -2402,7 +2402,7 @@ public void InternalPartialCashSettledPutAssignmentReducesSomeCash()
             securities[Symbols.SPY].SetMarketPrice(new TradeBar { Time = securities.UtcTime, Symbol = Symbols.SPY, Close = 189 });
             securities[Symbols.SPY_P_192_Feb19_2016].Holdings.SetHoldings(1, -2);
 
-            transactions.AddOrder(new SubmitOrderRequest(OrderType.OptionExercise, SecurityType.Option, Symbols.SPY_P_192_Feb19_2016, -1, 0, 0, securities.UtcTime, ""));
+            transactions.AddOrder(new SubmitOrderRequest(OrderType.OptionExercise, SecurityType.Option, Symbols.SPY_P_192_Feb19_2016, 1, 0, 0, securities.UtcTime, ""));
```
Martin-Molinero, line 2405: Maybe it would be more explicitly if the sign was the inverse of the holdings sign? Or directly the `- security.Holdings.Quantity` like the algo manager (and `/ 2` for those kind of partial tests)

Martin-Molinero (commented):
Nice improvements 👍

Common/Orders/Order.cs
```diff
@@ -292,7 +292,8 @@ public virtual void ApplyUpdateOrderRequest(UpdateOrderRequest request)
         /// <filterpriority>2</filterpriority>
         public override string ToString()
         {
-            return Invariant($"OrderId: {Id} (BrokerId: {string.Join(",", BrokerId)}) {Status} {Type} order for {Quantity} unit{(Quantity == 1 ? "" : "s")} of {Symbol}");
+            var tag = string.IsNullOrEmpty(Tag) ? string.Empty : $": {Tag}";
+            return Invariant($"OrderId: {Id} (BrokerId: {string.Join(",", BrokerId)}) {Status} {Type} order for {Quantity} unit{(Quantity == 1 ? "" : "s")} of {Symbol}{tag}");
```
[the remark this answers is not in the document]
mchandschuh: Yep, the order sign flipped which changed them on the previous PR. I ran regression with modified to string methods to re-flip the quantities to ensure it was the only change and it was. For this PR the only changes made were to the tags, the other changes are from the previous PR #4796

Common/Securities/Option/Option.cs
```diff
@@ -226,13 +226,30 @@ public decimal GetAggregateExerciseAmount()
...
-        /// we may have 1 option contract with multiplier of 100 with right to buy/sell only 50 shares of underlying stock.
+        /// Returns the directional quantity of underlying shares that are going to change hands on exercise/assignment of the
+        /// specified <paramref name="exerciseOrderQuantity"/>, taking into account the contract's <see cref="Right"/> as well
+        /// as the contract's current <see cref="ContractUnitOfTrade"/>, which may have recently changed due to a split/reverse
+        /// split in the underlying security.
         /// </summary>
-        /// <returns></returns>
-        public decimal GetExerciseQuantity(decimal quantity)
+        /// <remarks>
+        /// Long positions result in exercise while short positions result in assignment. This function uses the term exercise
```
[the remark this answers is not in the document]
mchandschuh: I'll add [option] positions. We can compute exercise quantity for OTM positions as well.

Common/Securities/Option/Option.cs
```diff
@@ -226,13 +226,30 @@ public decimal GetAggregateExerciseAmount()
         }
 
         /// <summary>
-        /// Returns the actual number of the underlying shares that are going to change hands on exercise. For instance, after reverse split
-        /// we may have 1 option contract with multiplier of 100 with right to buy/sell only 50 shares of underlying stock.
+        /// Returns the directional quantity of underlying shares that are going to change hands on exercise/assignment of the
+        /// specified <paramref name="exerciseOrderQuantity"/>, taking into account the contract's <see cref="Right"/> as well
+        /// as the contract's current <see cref="ContractUnitOfTrade"/>, which may have recently changed due to a split/reverse
+        /// split in the underlying security.
```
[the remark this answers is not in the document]
mchandschuh: Yes, I'll review this comment and make sure the 'viewpoint' is made explicit

Common/Extensions.cs
```diff
@@ -2263,5 +2263,22 @@ public static OrderDirection GetExerciseDirection(this OptionRight right, bool i
...
+        public static OrderDirection GetOrderDirection(decimal quantity)
+        {
+            var sign = Math.Sign(quantity);
+            switch (sign)
+            {
+                case 1:  return OrderDirection.Buy;
+                case 0:  return OrderDirection.Hold;
+                case -1: return OrderDirection.Sell;
+                default: throw new ApplicationException(
+                    $"The skies are falling and the oceans are rising! Math.Sign({quantity}) returned {sign} :/"
```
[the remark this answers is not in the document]
mchandschuh: haha, I love doing things like this 😛

Common/Securities/Option/Option.cs
```diff
@@ -226,13 +226,30 @@ public decimal GetAggregateExerciseAmount()
...
+        /// A negative value indicates exercise (we are long and the order quantity is negative to bring us (closer) to zero.
+        /// A positive value indicates assignment (we are short and the order quantity is positive to bring us (closer) to zero.</paramref>
+        public decimal GetExerciseQuantity(decimal exerciseOrderQuantity)
         {
-            return quantity * ContractUnitOfTrade;
+            // when exerciseOrderQuantity > 0 [ we are short ]
+            //      && right == call => we sell to contract holder  => negative
+            //      && right == put  => we buy from contract holder => positive
+
+            // when exerciseOrderQuantity < 0 [ we are long ]
```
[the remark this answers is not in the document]
mchandschuh: the thinking was callers of this method use `order.Quantity`. The determine 'our' exercise quantity (whoops, it got removed) I used to have one that didn't take any args and it would use `-Holdings.Quantity` -- I can add that back as a convenience function. By parameterizing it, you can compute exercise quantities agnostic to the current holdings, which I think can be powerful for algorithms.

Engine/AlgorithmManager.cs
mchandschuh: yea! amazing how some branches collapse on standardization!

Engine/Results/RegressionResultHandler.cs
mchandschuh: Thanks for catching this -- I meant to actually do something here :/ -- Will update!

Engine/Results/RegressionResultHandler.cs
mchandschuh: Yea, I reviewed it, the issue is it doesn't _sort_ the members alphabetically, so between regression runs you can end up with a different order.

Engine/Results/RegressionResultHandler.cs
mchandschuh: Absolutely, will update 👍

Common/Securities/SecurityPortfolioModel.cs
mchandschuh: ugh, I hate trailing white space, danke

Tests/Common/Securities/SecurityPortfolioManagerTests.cs
mchandschuh: yes, I like this. When making these changes I knew they were _correct_ but they seemed arbitrary. Will update 👍

Common/Securities/Option/Option.cs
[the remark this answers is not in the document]
mchandschuh: Thinking more about this, _technically_ you can be assigned an OTM option as well, so long as the long side exercises the OTM -- it's just incredibly unlikely, but a more common case is having holdings large enough to move the market and a strike price close enough that the estimated slippage from executing that many underlying shares makes it better to simply exercise the contract. Consider 10k put contracts for 1M underlying shares at strike=$23 and underlying market price=$24. For something like SLV, you're likely to get a net execution of _less than_ $23 if you actually executed this as a market order vs the guaranteed execution of $23 if exercised. Additionally, if someone is considering taking the opposite position, they might not want to hammer the market which could encourage others to sell as well, perhaps leading to breaking a prevailing trend in an illiquid security.

PR #4804, opened 2020-10-01
OptionChain and OptionContract improvements

Martin-Molinero:
#### Description

- QCAlgorithm.AddUniverse will return the added Universe instance.
- Adding new AddChainedOptionUniverse will monitor a Universe changes
  and will spwan new OptionChainUniverse from it's selections. Adding
  regression test Py/C#.
- Adding new OptionContractUniverse that will own option contracts and
  their underlying symbol. Adding regression test
- Fix double notification for security changes, bug seen in updated
  UniverseSelectionRegressionAlgorithm
- Remove UniverseSelection special handling for Option and Future chains
- Fix DataManager not removing SubscriptionDataConfigs for Subscriptions
  which finished before being removed from the universe
- Refactor detection of user added Universe so that they do not get
  removed after calling the UniverseSelectionModel

#### Related Issue

N/A
#### Motivation and Context

Improve Option Lean support

Probably adding new examples

Unit and regression tests
#### Types of changes

#### Checklist:


Algorithm.CSharp/CoarseFineOptionUniverseChainRegressionAlgorithm.cs
```diff
@@ -0,0 +1,185 @@
...
+            SetEndDate(2014, 06, 06);
+
+            var selectionUniverse = AddUniverse(enumerable => new[] { Time.Date <= new DateTime(2014, 6, 5) ? _twx : _aapl },
+                enumerable => new[] { Time.Date <= new DateTime(2014, 6, 5) ? _twx : _aapl });
+
+            AddChainedEquityOptionUniverseSelectionModel(selectionUniverse, universe =>
+            {
+                return universe.IncludeWeeklys()
+                    .FrontMonth()
+                    .Contracts(universe.Take(5));
```
Martin-Molinero, line 54: Should add an assert checking the underlying has a price

Martin-Molinero: Added assert, made sure underlying has a price live/backtesting, since is already handled by the OptionChainUniverse

Common/Data/UniverseSelection/FuturesChainUniverse.cs
```diff
@@ -150,5 +149,30 @@ public override bool CanRemoveMember(DateTime utcTime, Security security)
...
+        /// </summary>
+        /// <param name="security">The security to get subscriptions for</param>
+        /// <param name="currentTimeUtc">The current time in utc. This is the frontier time of the algorithm</param>
+        /// <param name="maximumEndTimeUtc">The max end time</param>
+        /// <param name="subscriptionService">Instance which implements <see cref="ISubscriptionDataConfigService"/> interface</param>
+        /// <returns>All subscriptions required by this security</returns>
+        public override IEnumerable<SubscriptionRequest> GetSubscriptionRequests(Security security, DateTime currentTimeUtc, DateTime maximumEndTimeUtc,
+            ISubscriptionDataConfigService subscriptionService)
+        {
+            if (Future.Symbol.Underlying == security.Symbol)
```
Martin-Molinero, line 164: This isn't working because the `FutureChainUniverse` isn't selecting the underlying, note that this is also broken in master where `Future.Underlying` is always null. Should it be adding the underlying symbol as done by the `OptionChainUniverse`? We could also create/get the underlying security when creating the `OptionChainUniverse/FutureChainUniverse` and set it there, or even required it as an argument

Martin-Molinero: Refactor deferred for issue https://github.com/QuantConnect/Lean/issues/4821, suggest removal of `Security Future.Underlying`

jaredbroad (approved):

jaredbroad:
Great regression algorithms

 - Missing 2 regression algorithms for python for educational purposes.

 - Renaming to `AddChainedOptionUniverse`. We can try our best to keep it under a uniform API method until we need to split it into asset-class specific endpoints.

 - `if (!Portfolio.Invested && _option != null && Securities[_option.Symbol].Price != 0 && !_traded)` Portfolio.Invested not needed

Martin-Molinero:
Pushed fix @gsalaz98, thanks for the catch 👍

Algorithm.CSharp/CoarseFineOptionUniverseChainRegressionAlgorithm.cs
```diff
@@ -0,0 +1,189 @@
...
+            _aapl = QuantConnect.Symbol.Create("AAPL", SecurityType.Equity, Market.USA);
+            UniverseSettings.Resolution = Resolution.Minute;
+
+            SetStartDate(2014, 06, 05);
+            SetEndDate(2014, 06, 06);
+
+            var selectionUniverse = AddUniverse(enumerable => new[] { Time.Date <= new DateTime(2014, 6, 5) ? _twx : _aapl },
+                enumerable => new[] { Time.Date <= new DateTime(2014, 6, 5) ? _twx : _aapl });
+
+            AddChainedOptionUniverse(selectionUniverse, universe =>
```
jaredbroad, line 50: Lets review this name

jaredbroad (approved):

Algorithm.CSharp/CoarseFineOptionUniverseChainRegressionAlgorithm.cs
jaredbroad: `AddUniverseOptions(universe, filter)` with sugar added for short cut later -> `AddOption(symbol or universe, filter)`

Issue #4567, opened 2020-07-08
Refactor Fill Models


AlexCatarino:
#### Expected Behavior
Stop Market, Limit, and Stop Limit Orders should be triggered by trade data.

#### Actual Behavior
Stop Market, Limit, and Stop Limit Orders should be triggered by quote data.

#### Potential Solution
`FillModel.GetPrices` method should be smart and return the relevant price(s). For the referenced orders types, the trigger price should be the `High` and `Low` of the last `TradeBar`. Since the time of the `High` and `Low` and unknown, perhaps we should consider filling with the close price.

#### Reproducing the Problem
N/A

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

adam-may:
Hi @AlexCatarino , I looked into this in some depth a while back. A good deal of the markets I've looked at will trigger the stops when the price is touched by the quote (i.e. it doesn't have to trade at that price for the stop to be triggered).

You also mention Limit orders in the description, but I think that's a mistake as they don't have anything to do with quotes?

AlexCatarino:
Here is the behavior as described by Interactive Brokers:
[Limit Orders](https://www.interactivebrokers.com/en/index.php?f=593):

> Step 1 – Enter a Limit Buy Order
> XYZ stock has a current Ask price of 34.00 and you want to use a Limit order to buy 100 shares when the market price falls to 33.50.
> You create the Limit order as shown above.
> Step 2 – Order Transmitted, Market Price Begins to Fall
> As soon as you transmit your Limit order, the market price of XYZ stock begins to fall.
> Step 3 – Market Price Falls to Limit Price, Order Filled
> The market price of XYZ continues to fall until it touches your Limit Price of 33.50. Your order for 100 shares is filled.
where the market price is the market offer.

[Stop Market](https://www1.interactivebrokers.com/en/index.php?f=609)

[...]

PR #4805, opened 2020-10-01
Adds Equity Fill Model

AlexCatarino:
#### Description
- Adds a fill model tailored for equity;
- Adds new unit tests for the new fill model and fix regression tests;
- Change the default fill model in the Equity class;
- Change the brokerage models to get the fill model according to the security type.

#### Related Issue
Addresses #4567 

#### Motivation and Context
Bug fix

New unit tests and regression tests,

#### Types of changes

#### Checklist:


Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -0,0 +1,634 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Orders.Fills
+{
+    /// <summary>
+    /// Implements <see cref="IFillModel"/> for Equity
+    /// </summary>
+    public class EquityFillModel : FillModel
```
adam-may, line 29: This is very similar to a Fill Model that we built for another product type, so maybe the term "EquityFillModel" should be changed to describe how the model fills the orders?

Martin-Molinero (changes requested):
Nice first step 👍 leaving some concerns, requests 😃
*believe there are a few regression algorithms failing with current impl

Common/Orders/Fills/EquityFillModel.cs
Martin-Molinero: Hey @adam-may! Curious which product type? The current plan as seen in the `DefaultBrokerageModel` is to create a fill model per security type but they could potentially inherit from each other is it's correct

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -0,0 +1,634 @@
...
+                if (quoteBar != null)
+                {
+                    return quoteBar;
+                }
+            }
+
+            if (subscribedTypes.Contains(typeof(Tick)))
+            {
+                var quote = asset.Cache.GetAll<Tick>().LastOrDefault(x => x.TickType == TickType.Quote);
+                if (quote != null)
```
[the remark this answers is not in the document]
Martin-Molinero: hm I thought I was wrong above but for example if `QuoteBar` resolution is daily and we have tick resolution subscription most of the day we would be using old data? same for `TradeBar` 🤔, note that `master` is using `Ticks` first too

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -125,10 +126,18 @@ public override OrderEvent StopMarketFill(Security asset, StopMarketOrder order)
                     // Buy Stop: If Price Above set point, Buy:
                     if (lastTradeBar.High >= order.StopPrice)
                     {
-                        DateTime unused;
+                        // Bar opens above stop price, fill at open price
+                        if (lastTradeBar.Open >= order.StopPrice)
+                        {
+                            fill.FillPrice = lastTradeBar.Open + slip;
```
Martin-Molinero, line 132: Sweet 👍

Martin-Molinero (approved):

Martin-Molinero (commented):
`StopLossOnOrderEventRegressionAlgorithm` & `UpdateOrderRegressionAlgorithm` seem to be failing @AlexCatarino

PR #4806, opened 2020-10-02
Set Account Base Currency from Brokerage in Live Mode

StefanoRaggi:
#### Description
- Added a new property: `IBrokerage.AccountBaseCurrency`
- All brokerage implementations now set `IBrokerage.AccountBaseCurrency` after a successful connection
- In `BrokerageSetupHandler.Setup()` the USD account currency check has been removed and the algorithm account currency is set equal to the brokerage account base currency
- The `brokerage.Connect()` method is now called before `algorithm.Initialize()`
- `CashBook.ToString()` now returns the correct account currency symbol, instead of the hardcoded `$` sign

#### Related Issue
#110

#### Motivation and Context
- Enable users with non-USD accounts to use their brokerage account base currency

No.

- Tested locally and in cloud with all brokerages

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/GDAX/GDAXBrokerage.cs
+++ b/Brokerages/GDAX/GDAXBrokerage.cs
@@ -35,6 +35,8 @@ namespace QuantConnect.Brokerages.GDAX
     {
         private const int MaxDataPointsPerHistoricalRequest = 300;
 
+        private static readonly HashSet<string> FiatCurrencies = new List<string> { "EUR", "GBP", "USD" }.ToHashSet();
+
         #region IBrokerage
         /// <summary>
         /// Checks if the websocket connection is connected or in the process of connecting
@@ -184,6 +186,16 @@ namespace QuantConnect.Brokerages.GDAX
             }
 
             return success.All(a => a);
+        }
+
+        /// <summary>
+        /// Connects the client to the broker's remote servers
+        /// </summary>
+        public override void Connect()
+        {
+            base.Connect();
+
+            AccountBaseCurrency = GetAccountBaseCurrency();
         }
 
         /// <summary>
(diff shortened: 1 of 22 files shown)


Brokerages/GDAX/GDAXBrokerage.cs
```diff
@@ -35,6 +35,8 @@ public partial class GDAXBrokerage : BaseWebsocketsBrokerage
     {
         private const int MaxDataPointsPerHistoricalRequest = 300;
 
+        private static readonly HashSet<string> FiatCurrencies = new List<string> { "EUR", "GBP", "USD" }.ToHashSet();
```
Martin-Molinero, line 38: nit - could use defined `Currencies`?

Martin-Molinero (commented):
Nice! Are we missing to update `Bitfinex` too?

Brokerages/GDAX/GDAXBrokerage.cs
StefanoRaggi: These are the only three currencies allowed for deposits to Coinbase Pro:
https://help.coinbase.com/en/pro/managing-my-account

StefanoRaggi:
> Nice! Are we missing to update `Bitfinex` too?

Bitfinex was left with the default base currency of USD intentionally because there is no single (and enforced) base currency like CoinbasePro. 

They seem to accept deposits in any of these currencies: USD, EUR, GBP, JPY, CNH
https://support.bitfinex.com/hc/en-us/articles/115004405873-A-Beginner-s-Guide-to-Bitfinex#h_01EJ0VD5GP4AA9XYZKA1RD42QQ

Engine/Setup/BrokerageSetupHandler.cs
```diff
@@ -168,6 +168,40 @@ public bool Setup(SetupHandlerParameters parameters)
...
+                }
+
+                if (!brokerage.IsConnected)
+                {
+                    // if we're reporting that we're not connected, bail
+                    AddInitializationError("Unable to connect to brokerage.");
+                    return false;
+                }
+
+                Log.Trace($"BrokerageSetupHandler.Setup(): {brokerage.Name} account base currency: {brokerage.AccountBaseCurrency}");
```
jaredbroad, line 198: Please post to algorithm debug message

Brokerages/Oanda/OandaRestApiV1.cs
```diff
@@ -64,6 +64,15 @@ public OandaRestApiV1(OandaSymbolMapper symbolMapper, IOrderProvider orderProvid
         {
         }
 
+        /// <summary>
+        /// Gets the account base currency
+        /// </summary>
+        public override string GetAccountBaseCurrency()
```
jaredbroad, line 70: Does OANDA still support V1 accounts?

jaredbroad (changes requested):
- Awesome PR
 - Please also include some way to set the currency of the strategy equity curve.

Brokerages/Oanda/OandaRestApiV1.cs
StefanoRaggi: The V1 API was deprecated two years ago:
https://developer.oanda.com/rest-live/introduction/

Existing V1 accounts were migrated to V20 and users received this notice:
`
On 8 December 2017 we will be migrating your legacy trade account(s) to a new V20 account(s). This will enable OANDA to ensure that your account(s) meet the anticipated regulatory requirements at the beginning of 2018 so you can continue to trade. On this date the funds in your legacy account(s) and any open positions will be transferred to your new V20 account(s). 
`

I think we can safely remove the Oanda V1 implementation from `OandaBrokerage` in a separate PR.

Martin-Molinero (approved):

PR #4828, opened 2020-10-07
DividedEventProvider distribution computation

Martin-Molinero:
#### Description

- Update regression algorithm which was using a different reference
  price when calculating the dividend
- Adjust divided event provider to compute distribution using factor
  file reference price, if not 0. Adding unit tests
- For equities, only emit auxiliary data points for
  TradeBar configurations, not for QuoteBars, nor internal.
- NewTradableDateEventArgs will provide last raw price avoid any price normalization

#### Related Issue

N/A
#### Motivation and Context

Normalize the distribution computation by using the FactorFile reference price. Avoiding any differences due to the underlying data enumerator-> extended market hours.

N/A

Unit and regression tests
#### Types of changes

#### Checklist:


Engine/DataFeeds/Enumerators/DividendEventProvider.cs
```diff
@@ -64,9 +65,16 @@ public IEnumerable<BaseData> GetEvents(NewTradableDateEventArgs eventArgs)
...
+                    decimal close;
+                    if (_referencePrice != null && _referencePrice != 0)
+                    {
+                        close = _referencePrice.Value;
+                    }
+                    else
+                    {
+                        close = eventArgs.LastRawPrice ?? 0;
+                    }
+
```
jaredbroad, line 77: Ideally, factor files can fill forward if no ref price and remove this branch and LastRawPrice.

The split warning can't get the factor file price as it'll introduce look-ahead bias. So sticking to as is:
https://github.com/QuantConnect/Lean/pull/4828/files#diff-a41c0f7b0b53c1edfae1df63b198fbb7R98

jaredbroad (approved):

Engine/DataFeeds/Enumerators/DividendEventProvider.cs
Martin-Molinero: Split and Dividend event provider will throw an exception when there is no reference price available 👍

PR #4832, opened 2020-10-08
Test 4249 fix regression algorithms executed together

### LINKED ISSUE #4249: A few python regression algorithms fail when executed with the rest
#### Expected Behavior

- All regression algorithms pass when executed together 
#### Actual Behavior

- A few python regression algorithms fail when run along the rest:
image

#### Potential Solution

- I believe this behavior has to do with the way python net is handling imports and caching them, I suspect of partial imports (for example `from QuantConnect.Algorithm import QCAlgorithm` vs `from QuantConnect.Algorithm import *`). **Should run tests in subgroups until we can find which is the test that is causing these regression tests to fail** when run together.
- Part of the solution of this issue should be determining and documenting whats the underlying issue so that it can be avoided in the future.
#### Reproducing the Problem

Run all python regression algorithms together
#### System Information

N/A
#### Checklist

C-SELLERS:
#### Description

Resolve a few regression algorithms that fail when ran together

#### Related Issue

Closes #4249 

#### Motivation and Context

When running all regression tests they always have a few that will fail. These changes modify them just enough to allow them to all run together.

The underlying problem is rather complex and has to do with the PythonNet's import cache. The PythonEngine is static between tests and any imported libraries in one algorithm will carry on to the other. Researching the problem we have three guidelines to try and follow until the PythonNet issues can be resolved:

- Remove unnecessary imports
- Avoid partial imports, they cause problems for down the line imports for the same namespace
- If importing a namespace within a namespace, also import the intermediate namespaces. I.E. if importing `from QuantConnect.Data.Custom.SEC import *` also import  `from QuantConnect.Data.Custom import *`  and `QuantConnect.Data import *`. Again this can affect importing down the testing chain. 

One of the most important findings when researching this problem is that the `QuantConnect.Common`'s `Market` namespace can be squashed by `QuantConnect.Data.Market`'s namespace. This was a leading cause in a lot of the regressions that rely on `Common`'s `Market.USA` attribute. If the `Data.Market` has been imported in a test that runs before a test that needs `Market.USA` then there is likely to be problems. A hotfix just for the purpose of running a test algorithm is using the following code in the Py Algorithm:

```
from collections import namedtuple
Market = namedtuple('_Market', 'USA')('usa')
```

Unfortunately this isn't a solution for all tests because it seems impossible to undo the imports. I have tried using both importlib and sys.modules to remove the imports and allow a fresh state between tests to no avail.

This PythonNet module that I created during to those attempts may be useful in the future.
```  
                 module = PythonEngine.ModuleFromString(
                    "moduleTools",
                    "import sys\n" +
                    "import importlib\n" +
                    "def getModules():\n" +
                    "   return list(sys.modules.keys())\n" +
                    "def resetModules(modules):\n" +
                    "   for m in list(sys.modules.keys()):\n" +
                    "       if m not in modules:\n" +
                    "           sys.modules.pop(m)\n" +
                    "def reloadModules(modules):\n" +
                    "   importlib.invalidate_caches()\n" +
                    "   for x in list(sys.modules.keys()):\n" +
                    "       if x not in modules:\n" +
                    "           try:\n" +
                    "               importlib.reload(sys.modules(x))\n" +
                    "           except:\n" +
                    "               print(x)\n"
                 );
```

None

All regression unit test ran and pass on first go.

If you run others sets of tests with the `RegressionTests` there can be issues.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Python/ConsolidateRegressionAlgorithm.py
+++ b/Algorithm.Python/ConsolidateRegressionAlgorithm.py
@@ -22,7 +22,6 @@ from QuantConnect import *
 from QuantConnect.Algorithm import *
 from QuantConnect.Indicators import *
 from QuantConnect.Securities import *
-from QuantConnect.Data.Market import *
 from QuantConnect.Data.Consolidators import *
 from CustomDataRegressionAlgorithm import Bitcoin
 from datetime import timedelta
--- a/Algorithm.Python/CustomConsolidatorRegressionAlgorithm.py
+++ b/Algorithm.Python/CustomConsolidatorRegressionAlgorithm.py
@@ -18,7 +18,7 @@ AddReference("QuantConnect.Algorithm")
 AddReference("QuantConnect.Indicators")
 
 from QuantConnect import *
-from QuantConnect.Python import PythonConsolidator
+from QuantConnect.Python import *
 from QuantConnect.Algorithm import *
 from QuantConnect.Data import *
 from QuantConnect.Data.Market import *
--- a/Algorithm.Python/CustomConsolidatorRegressionAlgorithm.py
+++ b/Algorithm.Python/CustomConsolidatorRegressionAlgorithm.py
@@ -13,19 +13,18 @@
 
 from clr import AddReference
 AddReference("System")
+AddReference("QuantConnect.Common")
 AddReference("QuantConnect.Algorithm")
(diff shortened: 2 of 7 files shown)


Martin-Molinero (approved):
Thanks 👍 Think we should open an issue in PythonNet with the findings too

PR #4838, opened 2020-10-10
Bug 4835 api tests failing

### LINKED ISSUE #4835: API Tests Failing
#### Expected Behavior

All API tests pass

#### Actual Behavior

image

Basic API tests are failing when they shouldn't be. Here is my findings:

Endpoint problems:
- CreateBacktest() is failing on the CSharpProject & PythonProject test; No error
- AddProjectFile() is failing on the FSharpProject test <= The error say it doesn't support file extension?? It is a .fs file? Did we remove FSharp?
- ReadPrices() is failing for both 1Symbol and MultipleSymbols, API response has no helpful data on the problem is.

Test Problem:
- GetDividends and GetSplits tests fail because the AAPL data does not match the asserted values. Endpoint seem to be working, but values are slightly off? Maybe an adjustment to the data?
- CRUD_ProjectFiles_SucessFully : The apis return on our fake file doesn't match the code I give it, in fact the response code attribute is null. Could be endpoint issue, requires more studying. 

For the brokerage/data provider tests, maybe we should seperate them since they require configs for each API?

#### Potential Solution

Go through and adjust tests, maybe separate brokerage tests for simplifying test fixtures 

#### Reproducing the Problem

Run RestApiTests with UserId and Token

#### Checklist

C-SELLERS:
#### Description

API & API test fixes to reflect current web API.

#### Related Issue

Closes #4835 

#### Motivation and Context

Ref  #4835 

Address some dead API endpoints, cleanup of certain api test functions, fixes for misaligned lean to web api.

API Documentation

All QuantConnect.Tests.Api tests are 

#### Types of changes

#### Checklist:


Tests/API/ApiIBTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            var compile = _api.CreateCompile(project.Projects.First().ProjectId);
+            Assert.IsTrue(compile.Success);
+
+            // Create default algorithm settings
+            var settings = new InteractiveBrokersLiveAlgorithmSettings(user,
+                                                                       password,
+                                                                       account);
+
+            // Wait for project to compile
+            Thread.Sleep(10000);
```
Martin-Molinero, line 87: Think we could replace this (and the other cases) by a call to `_api.ReadCompile()` every 1sec, also as a way of testing that endpoint

Tests/API/ApiTradierTests.cs
```diff
@@ -0,0 +1,114 @@
...
+                Assert.IsTrue(liquidateLive.Success);
+
+                // Stop live algorithm
+                var stopLive = _api.StopLiveAlgorithm(project.Projects.First().ProjectId);
+                Assert.IsTrue(stopLive.Success);
+
+                // Delete the project
+                var deleteProject = _api.DeleteProject(project.Projects.First().ProjectId);
+                Assert.IsTrue(deleteProject.Success);
+            }
```
Martin-Molinero, line 109: Pretty cool tests! 👍 They seem **very** similar though, I think we could create a base `ApiLiveTradingTests` (or similar) and abstract the core behavior into a method there?

Martin-Molinero (changes requested):
Cool stuff! Leaving minor requests
The branch needs a rebase from `master` due to the rename QuantConnect.API to QuantConnect.Api

C-SELLERS:
Currently working through web API problems, code on Lean side is ready for review.

C-SELLERS:
All web API endpoint issues related to these tests have now been fixed, the only things left to be tested is Oanda, IB, FXCM, and Tradier specific tests. Do we have some test accounts to run these through or should I create some?

C-SELLERS:
Issues with `live/*` endpoints in `ApiLiveTradingTest`, park this PR until more information on Web API params are given and fixes for `Api.cs` are in place.

C-SELLERS:
@Martin-Molinero 

Alright fixes are in for `ApiLiveTradingTest`'s, unfortunately there is no demo Tradier account so I can't confirm that one passes but the core logic is shared and the rest all pass.

Also an issue with `LiveAlgorithmSettings_CanBeCreated_Successfully()` needs to be resolved but I created a ticket for it at #4864  Currently it will fail because it expects all supported Brokerages to have a LiveAlgorithmSettings subclass, we could remove the exception for the default case of the switch statement therefore allowing it to pass.

Brokerages/ApiPriceProvider.cs
```diff
@@ -54,19 +54,8 @@ public ApiPriceProvider(int userId, string userToken)
...
-
-            var priceData = result.Prices.FirstOrDefault(x => x.Symbol == symbol);
-            if (priceData == null)
-            {
-                throw new Exception($"No price data available for symbol: {symbol.Value}");
-            }
-
-            return priceData.Price;
+            //NOP ReadPrices endpoint has been removed
+            throw new Exception("Prices endpoint is no longer supported");
```
Martin-Molinero, line 58: nit - might be more precise to throw something like `InvalidOperationException`

Algorithm.CSharp/BasicTemplateAlgorithm.cs
```diff
@@ -119,4 +119,4 @@ public override void OnData(Slice data)
             {"OrderListHash", "491919591"}
         };
     }
-}
+}
```
Martin-Molinero, line 122: nit - might be a nice git exercise to remove this file 😃

Tests/Api/ApiLiveTradingTests.cs
```diff
@@ -0,0 +1,346 @@
...
+using NUnit.Framework;
+using QuantConnect.Api;
+using QuantConnect.Brokerages;
+using QuantConnect.Configuration;
+
+namespace QuantConnect.Tests.API
+{
+    [TestFixture, Ignore("These tests require configured and active accounts to Interactive Brokers, FXCM, Oanda, and Tradier " +
+         "as well as your Organization ID and available live nodes")]
+    class ApiLiveTradingTests
```
Martin-Molinero, line 30: Should be `public` I think? Not sure it will be found by the test framework if not

Tests/Api/ApiLiveTradingTests.cs
```diff
@@ -0,0 +1,346 @@
...
+            Assert.IsTrue(liveLogs.Success);
+            Assert.IsTrue(liveLogs.Logs.Any());
+        }
+
+        /// <summary>
+        /// Runs the algorithm with the given settings and file
+        /// </summary>
+        /// <param name="settings">Settings for Lean</param>
+        /// <param name="file">File to run</param>
+        private void RunLiveAlgorithm(BaseLiveAlgorithmSettings settings, ProjectFile file)
```
Martin-Molinero, line 286: Ah beautiful!

Martin-Molinero (approved):
Looks good! Minor final comments

> Also an issue with LiveAlgorithmSettings_CanBeCreated_Successfully() needs to be resolved but I created a ticket for it at #4864 Currently it will fail because it expects all supported Brokerages to have a LiveAlgorithmSettings subclass

Since the fixture is ignored I think it looks good as is 👍

Algorithm.CSharp/BasicTemplateAlgorithm.cs
C-SELLERS: Necessary change, removes /r/n from end of file. Important for test that uploads code and then reads the code in the cloud to compare.

Brokerages/ApiPriceProvider.cs
C-SELLERS: Gotcha!

Tests/Api/ApiLiveTradingTests.cs
C-SELLERS: Got it! It seems to work anyway but might as well clarify that it is public!

Tests/Api/ApiLiveTradingTests.cs
C-SELLERS: Thank you!

PR #4847, opened 2020-10-12
bug-#4846-Fail on restart investing after liquidation on MaximumDrawdownPercentPortfolio.py

### LINKED ISSUE #4846: MaximumDrawdownPercentPortfolio.py trailing stop bug
#### Expected Behavior

After liquidation requirements were met, on the next rebalance period the algo should restart investing

#### Actual Behavior

Didn't restart reinvestment

#### Potential Solution

reset trailing high value of portfolio

#### Reproducing the Problem

if portfolio experiences a drawdown exceed the preset level of aversion on using `self.SetRiskManagement(MaximumDrawdownPercentPortfolio())`, it stuck and won't proceed anymore.

#### Checklist

31620000:
#### Description

If trailing stop is True, the algo reached liquidation requirement will always be stuck as the 'high' value of portfolio is always higher than current value by preset drawdown percent. It will never restart investing on next rebalancing period and beyond.

#### Related Issue

Fixes #4846

#### Motivation and Context

I added a line so that the trailing high value could be reset after liquidation and the investment process won't be stop by high value always more than current value by drawdown percent.

Nil

https://travis-ci.org/github/QuantConnect/Lean/builds/735379137

#### Types of changes

#### Checklist:


Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.py
```diff
@@ -50,6 +50,7 @@ def ManageRisk(self, algorithm, targets):
 
         # Update trailing high value if in trailing mode
         if self.isTrailing and self.portfolioHigh < currentValue:
+            self.initialised = False # reset the trailing high value for restart investing on next rebalcing period
```
Martin-Molinero, line 53: Hey! Thanks for your contribution.
The effect of `self.initialised = False` seems to be resetting the `self.portfolioHigh` which we are already setting here too? Not sure _why_ this change is required and the effect it's having

>If trailing stop is True, the algo reached liquidation requirement will always be stuck as the 'high' value of portfolio is always higher than current value by preset drawdown percent. It will never restart investing on next rebalancing period and beyond.

^ From the reported issue. It should restart investing once it's out of the drawdown?
I think we should add some unit test (see MaximumDrawdownPercentPerSecurityTests) reproducing the underlying reported issue so it's clearer and we make sure the expected behavior is covered

31620000: ooops sorry I added it at a wrong line. Should be the if condition below

31620000:
I now edited to the correct version. As per enquiry, the original file won't change the self.portfolioHigh when the liquidation happened, so:
1. the currentValue is always less than high value and `if self.isTrailing and self.portfolioHigh < currentValue:` will never fulfil; and 
2. since currentValue and self.portfolioHigh won't change after liquidation, `if pnl < self.maximumDrawdownPercent` will always fulfil and never be able to invest

Algorithm.Framework/Risk/MaximumDrawdownPercentPortfolio.py
31620000: plz see comment on below and review

Martin-Molinero: Thanks, it makes sense! 👍 We should update the C# implementation too `MaximumDrawdownPercentPortfolio.cs` I'll make a GH issue for adding some tests for this

Martin-Molinero:
Hey!
I went ahead and updated the C# implementation and added a check so that we only reset the portfolio value after we actually have liquidated a collection of targets. Found this while testing with the regression test (which have been updated),  the portfolio construction model might only rebalance daily so there could be no targets until next rebalance, without the new check it wouldn't adjust the risk on the actual targets.

31620000:
Thx a lot for the effort!

Martin-Molinero (approved):

PR #4856, opened 2020-10-13
Allow account currency to be overridden by the algorithm

### LINKED ISSUE #4855: SetAccountCurrency cannot be called more than once
#### Expected Behavior
- Users should be able to override the account currency set automatically from a live brokerage (new feature merged in #4806)

#### Actual Behavior
- In live mode, after the account currency has been set from the brokerage, `algorithm.SetAccountCurrency()` does not change the account currency and only logs this warning:
```
SecurityPortfolioManager.SetAccountCurrency(): account currency has already been set to EUR. Will ignore new value USD.
```

#### Potential Solution
Remove the `_setAccountCurrencyWasCalled` check

#### Reproducing the Problem
- Deploy an algorithm to Oanda `EUR` account with `SetAccountCurrency("USD")` in `algorithm.Initialize()`

#### System Information
n/a

#### Checklist

StefanoRaggi:
#### Description
- The existing check in `SecurityPortfolioManager.SetAccountCurrency()` was preventing algorithms from selecting a different currency from the one automatically fetched from the live brokerage -- this check has been removed

#### Related Issue
Closes #4855 

#### Motivation and Context
- Unable to deploy a live algorithm to an Oanda `EUR` account and select `USD` as the algorithm currency

No.

- Added new unit tests
- Tested live algorithm with Oanda as described in #4855 

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/SecurityPortfolioManager.cs
+++ b/Common/Securities/SecurityPortfolioManager.cs
@@ -34,6 +34,7 @@ namespace QuantConnect.Securities
     public class SecurityPortfolioManager : ExtendedDictionary<SecurityHolding>, IDictionary<Symbol, SecurityHolding>, ISecurityProvider
     {
         // flips to true when the user called SetCash(), if true, SetAccountCurrency will throw
+        private bool _setAccountCurrencyWasCalled;
         private bool _setCashWasCalled;
         private bool _isTotalPortfolioValueValid;
         private decimal _totalPortfolioValue;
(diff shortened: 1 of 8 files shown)


Tests/Common/Packets/BacktestNodePacketTests.cs
```diff
@@ -250,7 +250,6 @@ internal class TestInitialCashAmountAlgorithm : BasicTemplateDailyAlgorithm
         {
             public override void Initialize()
             {
-                SetAccountCurrency("EUR");
```
Martin-Molinero, line 253: The idea here is that the backtest job packet overrides the algorithms account currency which gets ignored, believe this is why we were directly returning at the `SecurityPortfolioManager` when already set, kinda of the inverse behavior that this PR wants to achieve. Let's discuss offline but maybe the check being removed, if still required, could potentially be applied only for backtesting at the algorithm layer 🤔

Common/Securities/SecurityPortfolioManager.cs
```diff
@@ -534,20 +533,6 @@ public void SetAccountCurrency(string accountCurrency)
...
-            {
-                if (accountCurrency != CashBook.AccountCurrency)
-                {
-                    Log.Trace("SecurityPortfolioManager.SetAccountCurrency():" +
-                              $" account currency has already been set to {CashBook.AccountCurrency}." +
-                              $" Will ignore new value {accountCurrency}");
-                }
-                return;
-            }
-            _setAccountCurrencyWasCalled = true;
```
Martin-Molinero, line 549: We agreed on maintaining existing behavior, and for the cases where the brokerage does not know which is the account currency `AccountBaseCurrency` would be null, so the `BrokerageSetupHandler` won't set account current and the user algorithm can. Suggest we do this for the `PaperBrokerage : BacktestingBrokerage` so that paper user algorithms can test with other account currencies, since in any case the default account current for the algorithm is already `Currencies.USD`

Martin-Molinero (changes requested):

Martin-Molinero (approved):

PR #4879, opened 2020-10-20
Bug 4757 Synchronize Security Additions

### LINKED ISSUE #4757: Synchronization After Newly Added Securities
#### Expected Behavior
After a Security is added, the next iteration has a different date/time and synchronized data.

#### Actual Behavior
After a Security is added, the next iteration repeats the date/time and doesn't have synchronized data.

#### Potential Solution
Add new securities with start date/time greater than the current algorithm time. 

#### Reproducing the Problem
```python
class TestingAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2020, 9, 22)
        self.SetEndDate(self.StartDate)
        self.AddEquity('SPY')
        
    def OnData(self, slice):
        if self.Time.hour == 9 and self.Time.minute < 33:
            self.Log(f'Time before calling AddEquity: {self.Time} ' + ' '.join([x.Value for x in slice.Bars.Keys]))
            self.AddEquity('QQQ')
            self.Log(f'Time after calling AddEquity: {self.Time} ' + ' '.join([x.Value for x in slice.Bars.Keys]))
```
Output
```
2020-09-22 09:31:00 Time before calling AddEquity: 2020-09-22 09:31:00 SPY
2020-09-22 09:31:00 Time after calling AddEquity: 2020-09-22 09:31:00 SPY
2020-09-22 09:31:00 Time before calling AddEquity: 2020-09-22 09:31:00 QQQ
2020-09-22 09:31:00 Time after calling AddEquity: 2020-09-22 09:31:00 QQQ
2020-09-22 09:32:00 Time before calling AddEquity: 2020-09-22 09:32:00 QQQ SPY
2020-09-22 09:32:00 Time after calling AddEquity: 2020-09-22 09:32:00 QQQ SPY
```
#### Checklist

C-SELLERS:
#### Description

Change `TimeTriggeredUniverseSubscriptionEnumeratorFactory` processing on universe 's CollectionChanged to sync new subscriptions with the next timestep.

This lead to necessary changes to a few regressions because their behaviour is related to the bug.

Also one change to `BacktestResultHandler` where sampling was directly adding points to series, now it uses Series.AddPoint() which verifies the validity of the point before entering. This was a breaking issue post changes because of a duplicate keys being entered into Series value.

#### Related Issue

Closes #4757 

#### Motivation and Context

In order to synchronize data that is added outside of initialize with the next timestep I had to bump the time on the collection by 1 tick so that it doesn't generate a point for the current time thus triggering OnData for a given time more than once.

Here is an example of the issue:

```
        private bool _called = false;

        {
            SetStartDate(2014, 06, 05);  //Set Start Date

        }

        {
            var symbols = data.Bars.Keys.Select(x => x.Value).ToList();
            Log($"Add Equity Called: {_called} : {Time} : "+ string.Join(" ", symbols));
            if (!_called)
            {
                AddEquity("AAPL", Resolution.Minute);
                _called = true;
            }
            else if (Time.Equals(new DateTime(2014, 06, 5, 12, 0, 0)))
            {
                RemoveSecurity("AAPL");
            }
        }
```

##### Before Changes:
image

As you can see, AAPL is added at 9:31AM and then OnData gets triggered again for 9:31AM in the next loop. Instead we should see it not show until 9:32AM in Sync with the next SPY data.

##### After Changes:
image

Now the data is synced with SPY and we do not pass through OnData twice with the same timestamp. 

I can also confirm that removal still works as expected, I put a condition that if it was a removal action to not alter the time.

image

 

N/A

All regressions are passing and those that needed adjustments have been altered. Travis tests run locally in docker also all pass.

#### Types of changes

#### Checklist:


Algorithm.CSharp/UniverseSharingSubscriptionTradableRegressionAlgorithm.cs
```diff
@@ -111,29 +111,29 @@ public override void OnSecuritiesChanged(SecurityChanges changes)
             {"Total Trades", "1"},
             {"Average Win", "0%"},
             {"Average Loss", "0%"},
-            {"Compounding Annual Return", "75.079%"},
-            {"Drawdown", "2.200%"},
+            {"Compounding Annual Return", "69.904%"},
+            {"Drawdown", "2.000%"},
```
Martin-Molinero, line 115: Believe these stats changed because this algorithm also reproduces the underlying issue, and after the fix it enters the trade a day later.
Let's please add an explicity check in the 3 algorithms `OnData()` throwing if they get a duplicate time

Martin-Molinero (approved):
Thanks for the fix! Let's add an assert in the algorithms reproducing the issue

Algorithm.CSharp/UniverseSharingSubscriptionTradableRegressionAlgorithm.cs
C-SELLERS: Agreed, will go ahead and add those checks!

jaredbroad:
Ideally, this could be fixed closer to the source of the bug. I feel like the root bug is the start time of the subscription is being set to a value in the "past" and not "now" so it is not filtering a bar that should be filtered. By the time we have finished OnData even in backtesting some time has passed so it is already 0931.050. It might be some start time of the subscription on the config to allow this.

Fixing in a universe factory with added tick feels too far from that root cause. But for posterity it's here because there is no start time on the configuration files created when AddSecurity is called; and its start time is set by that universe Collection Change event. I hope in 2 years I hope we understand why we're adding a tick here :)

Martin-Molinero (approved):
Thanks! 👍

PR #4880, opened 2020-10-21
Local Object Store Refactor and Fixes

### LINKED ISSUE #4811: LocalObjectStore.GetFilePath Rewrites Existing File With New Name
#### Expected Behavior
No file with duplicated information is created.

#### Actual Behavior
`LocalObjectStore.GetFilePath` creates a new file with information found in files loaded in `LocalObjectStore.Initialize`

#### Potential Solution
N/A

#### Reproducing the Problem
```python
import os
qb = QuantBook()

for i in range(2):
    qb.ObjectStore.Save(f'{chr(i+97)}.txt', f'{i}'*100)

keys = [str(j).split(',')[0][1:] for _, j in enumerate(qb.ObjectStore.GetEnumerator())]
sizes = [os.path.getsize(qb.ObjectStore.GetFilePath(key)) for key in keys]
for i in keys:
    path = qb.ObjectStore.GetFilePath(i)
    print(f"{i} - {path} - {os.path.getsize(path)/1e6:.1f} MB")
    print(f'Deleted? {qb.ObjectStore.Delete(i)}')
print(f"Total size: {sum(sizes)/1e6:.1f} MB in {len(sizes)} files")
```
Run this notebook once, the comment out lines 4 and 5, and run it again. We can see that there are more files in every new cell execution.

#### Checklist

C-SELLERS:
#### Description

- Refactor LocalObjectStore to no longer use temp copies of an object in `GetFilePath`, instead use the local copy, this fixes issue #4811.
- `GetFilePath` persists data before giving the requester the path.
- Fix for control checks to save data, logic was broken as it never counted the new data being added.
- Store data using conversion of keys to Base64 to handle odd key strings in path. The only limitation in chars in the key string is having '?' in the key. In a spot divisible by 3 this will cause Base64 to output '/' which is the only breaking case.
- New tests to prove behaviour

#### Related Issue

Closes #4811 

#### Motivation and Context

`LocalObjectStore`'s `GetFilePath` currently writes a copy of a requested object to a file in the ObjectStore's storage directory and returns this copies path; this works fine until you start Lean again and the copy is loaded into the LocalObjectStore as its own object.

To resolve this I have dropped the use of temp files altogether and will only store one file for each object, now when the user calls `GetFilePath` we ensure the data is written to storage and then return a path to it . This way there are no duplicates, the file is still retained and we can ensure the data in the file is up to date when requested.

N/A

All regressions ran and passed, LocalObjectStore tests all pass, Travis tests pass in docker

Verified behaviour in Issue #4811 does not occur

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/Storage/LocalObjectStore.cs
+++ b/Engine/Storage/LocalObjectStore.cs
@@ -59,7 +59,7 @@ namespace QuantConnect.Lean.Engine.Storage
         private TimeSpan _persistenceInterval;
         private readonly string _storageRoot = Path.GetFullPath(Config.Get("object-store-root", "./storage"));
         private readonly ConcurrentDictionary<string, byte[]> _storage = new ConcurrentDictionary<string, byte[]>();
-        private string TempObjectStorage => Path.Combine(AlgorithmStorageRoot, "temp");
+        private readonly object _persistLock = new object();
 
         /// <summary>
         /// Provides access to the controls governing behavior of this instance, such as the persistence interval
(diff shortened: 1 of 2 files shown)


Martin-Molinero (approved):
Thanks for the tests! 👍

jaredbroad (changes requested):
IMHO AddTick, Create Temp folders -- are warning signals for a quick fix hiding an underlying design issue. When we see that we should reflect on what the underlying cause and motivations that got us to this point were.

LocalObjectStore should likely just _return the local file path_ which would eliminate the creation of duplicates. Cloud concerns should be addressed separately.

Engine/Storage/LocalObjectStore.cs
```diff
@@ -250,7 +265,7 @@ public bool Delete(string key)
 
                 try
                 {
-                    var path = GetFilePathForKey(key);
+                    var path = PathForKey(key);
                     if (File.Exists(path))
                     {
                         File.Delete(path);
```
Martin-Molinero, line 271: I think `PersistData()` should clear the directory before writing to disk and we can remove this delete from here?

Tests/Common/Storage/LocalObjectStoreTests.cs
```diff
@@ -305,6 +315,72 @@ public void DisposeRemovesEmptyStorageFolder()
...
+                Assert.IsTrue(qb.ObjectStore.ContainsKey("a.txt"));
+
+                // Assert the store has only a.txt
+                var store = qb.ObjectStore.GetEnumerator().AsEnumerable().ToList();
+                Assert.IsTrue(store.Count == 1);
+                Assert.IsTrue(store[0].Key == "a.txt");
+
+                // Create a copy for use using GetFilePath, check that it exists
+                var path = qb.ObjectStore.GetFilePath("a.txt");
+                Assert.IsTrue(File.Exists(path));
```
Martin-Molinero, line 357: Nice tests! We should also add the 'Delete' case: We add a key that gets persisted and later we delete it. When we create a new object store it shouldn't re load it

Engine/Storage/LocalObjectStore.cs
```diff
@@ -279,19 +294,23 @@ public bool Delete(string key)
         /// <returns>The path for the file</returns>
         public string GetFilePath(string key)
```
Martin-Molinero, line 295: Realized we need this guy to be virtual since its calling `PersistData()` now

Tests/Common/Storage/LocalObjectStoreTests.cs
```diff
@@ -305,6 +321,108 @@ public void DisposeRemovesEmptyStorageFolder()
...
+            {
+                Assert.IsTrue(_store.SaveString($"{i}", $"{i}"));
+            }
+
+            // Write 1 more; should throw
+            Assert.IsFalse(_store.SaveString("breaker", "gotem"));
+        }
+
+        [Test]
+        public void DeletedObjectIsNotReloaded()
```
Martin-Molinero, line 391: Cool! 👍

Martin-Molinero (approved):

PR #4884, opened 2020-10-25
feat(ToolBox\IQFeed): prevent unordered ticks to be processed

### LINKED ISSUE #4649: Unordered data coming from IQFeed throw exception when writing data to disk
#### Expected Behavior

Even if some faulty data come, write data to disk

#### Actual Behavior

Throw exception from WriteMinuteOrSecondOrTick method

#### Potential Solution

Manually filter unordered data before passing to LeanDataWriter

#### Reproducing the Problem

Fetching data from NYSE stocks that returns unordered ticks

#### Checklist

mathpaquette:
closes #4649

#### Description
After discussing with IQFeed support, they confirmed to me that exceptionally they might return unordered data. In order to prevent unexpected exception, I added some checks before return the data.

#### Related Issue

#4649 

#### Motivation and Context

Unexcepted exception from https://github.com/QuantConnect/Lean/blob/master/ToolBox/LeanDataWriter.cs#L364

Tested locally without impacted the normal behavior.

#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/IQFeed/IQFeedFileHistoryProvider.cs
+++ b/ToolBox/IQFeed/IQFeedFileHistoryProvider.cs
@@ -62,7 +62,7 @@ namespace QuantConnect.ToolBox.IQFeed
             var ticker = _symbolMapper.GetBrokerageSymbol(request.Symbol);
             if (string.IsNullOrEmpty(ticker))
             {
-                Log.Trace($"IQFeedHistoryProvider.ProcessHistoryRequests(): Unable to retrieve ticker from Symbol: ${request.Symbol}");
+                Log.Trace($"IQFeedFileHistoryProvider.ProcessHistoryRequests(): Unable to retrieve ticker from Symbol: ${request.Symbol}");
                 return Enumerable.Empty<BaseData>();
             }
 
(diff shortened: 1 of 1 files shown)


mathpaquette:
image

mathpaquette:
image

mathpaquette:
I've also experienced that on Interval data as well so adding the protection for all data type.

ToolBox/IQFeed/IQFeedFileHistoryProvider.cs
```diff
@@ -270,6 +270,28 @@ private static string GetHistoryRequestKey(string ticker, DateTime startDate, Da
             return $"{ticker}-{startDate}-{endDate}";
         }
 
+        /// <summary>
+        /// Prevent IQFeed from returning unordered data of all sort. This might happen exceptionally.
+        /// </summary>
+        /// <param name="data"></param>
+        /// <returns></returns>
+        private static IEnumerable<BaseData> FilterUnorderedData(IEnumerable<BaseData> data)
```
Martin-Molinero, line 278: Curious why we filter instead of ordering the data?

Martin-Molinero (commented):
Thanks for the fix 👍 
Interesting that this affects **historical** IQFeed data, it seems to me they could order it on their side.

ToolBox/IQFeed/IQFeedFileHistoryProvider.cs
mathpaquette: good question... basically its just unexpected but might happen. I dont what to impact the overperformance for 0.001%

mathpaquette:
> Thanks for the fix 👍
> Interesting that this affects **historical** IQFeed data, it seems to me they could order it on their side.

Yes I agree. Ill follow up with them. But I think we need this little patch for now.

ToolBox/IQFeed/IQFeedFileHistoryProvider.cs
Martin-Molinero: Sounds good 👍

Martin-Molinero (approved):

PR #4891, opened 2020-10-28
Expand Filtering of Contracts by type to Futures

### LINKED ISSUE #4725: Exclude Weekly Futures By Default
#### Expected Behavior
Like options, futures' contracts should not include weekly contracts that would be requested by `IncludeWeeklys()` method.

#### Actual Behavior
Weeklies are included by default and there no API method to remove them.

#### Potential Solution
Implement `IncludeWeeklys()`.

#### Reproducing the Problem
In live mode, we got `VX30U20` (for ref., `VX23U20` is the standard contract):
> 2020-09-14T13:40:50.2704570Z TRACE:: Brokerage.OnMessage(): Information - Code: 10197 - No market data during competing live session. Origin: Subscribe: VX30U20 (FUT VIX USD CFE)

which is a weekly contract.

#### Checklist

C-SELLERS:
#### Description

- Support filtering Future contracts by Standard or Weeklys
- Combine core filtering logic of Options and Futures contracts into new base class `ContractSecurityFilterUniverse<T>` to reduce code duplication
- Addition of tests to verify correct behaviour of filters

#### Related Issue

Closes #4725 

#### Motivation and Context

Just as with Options we expect Futures to only include standard contracts by default. This PR implements the functionality to filter out these contracts with the same interface as OptionFilterUniverse, (i.e. .IncludeWeeklys(), WeeklysOnly). Now by default (just like Options) Futures contracts will not include weeklys, you must use the filter functions above to allow them through the filter.

Also upon inspection it was clear that both filter setups (OptionsFilterUniverse & FutureFilterUniverse) were basically exactly the same except a couple of unique filters. So instead of expanding the duplication of code I combined their core logic into a shared base class and then left the unique functions for implementation in their respective classes. 

N/A

- Test built to cover case of filtering out weeklys, standards, and neither 
- Test for symbols without an expiry function (flags as standard by default**) 
- Ran Live Paper Trading through IB to observe the Weekly contracts for "VX" being filtered out as described in #4725 
- Will test in the cloud

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/ContractSecurityFilterUniverse.cs
+++ b/Common/Securities/ContractSecurityFilterUniverse.cs
@@ -31,9 +31,10 @@ namespace QuantConnect.Securities
     where T: ContractSecurityFilterUniverse<T>
     {
         /// <summary>
-        /// Defines listed contract types
+        /// Defines listed contract types with Flags attribute
         /// </summary>
-        public enum Type : int
+        [Flags]
+        protected enum ContractExpirationType : int
         {
             /// <summary>
             /// Standard contracts
@@ -43,10 +44,36 @@ namespace QuantConnect.Securities
             /// <summary>
             /// Non standard weekly contracts
             /// </summary>
-            Weeklys = 2
+            Weekly = 2
         }
 
-        internal IEnumerable<Symbol> _allSymbols;
+        /// <summary>
+        /// Expiration Types allowed through the filter
+        /// Standards only by default
+        /// </summary>
+        protected ContractExpirationType Type = ContractExpirationType.Standard;
+
+        /// <summary>
+        /// The underlying price data
+        /// </summary>
+        protected BaseData UnderlyingInternal;
+
+        /// <summary>
+        /// All Symbols in this filter
+        /// Marked internal for use by extensions
+        /// </summary>
+        internal IEnumerable<Symbol> AllSymbols;
+
+        /// <summary>
+        /// Mark this filter dynamic for regular reapplying
+        /// Marked internal for use by extensions
+        /// </summary>
+        internal bool IsDynamicInternal;
+
+        /// <summary>
+        /// True if the universe is dynamic and filter needs to be reapplied
+        /// </summary>
+        public bool IsDynamic => IsDynamicInternal;
 
         /// <summary>
         /// The underlying price data
@@ -56,21 +83,10 @@ namespace QuantConnect.Securities
             get
             {
                 // underlying value changes over time, so accessing it makes universe dynamic
-                _isDynamic = true;
-                return _underlying;
+                IsDynamicInternal = true;
+                return UnderlyingInternal;
             }
         }
-
-        protected BaseData _underlying;
-
-        /// <summary>
-        /// True if the universe is dynamic and filter needs to be reapplied
-        /// </summary>
-        public bool IsDynamic => _isDynamic;
-
-        //TODO: Naming Conventions
-        internal bool _isDynamic;
-        protected Type _type = Type.Standard;
 
         /// <summary>
         /// Constructs ContractSecurityFilterUniverse
@@ -84,36 +100,22 @@ namespace QuantConnect.Securities
         /// </summary>
         protected ContractSecurityFilterUniverse(IEnumerable<Symbol> allSymbols, BaseData underlying)
         {
-            _allSymbols = allSymbols;
-            _underlying = underlying;
-            _type = Type.Standard;
-            _isDynamic = false;
+            AllSymbols = allSymbols;
+            UnderlyingInternal = underlying;
+            Type = ContractExpirationType.Standard;
+            IsDynamicInternal = false;
         }
 
         /// <summary>
-        /// Includes universe of non-standard weeklys contracts (if any) into selection
+        /// Function to determine if the given symbol is a standard contract
         /// </summary>
-        /// <returns></returns>
-        public T IncludeWeeklys()
-        {
-            _type |= Type.Weeklys;
-            return (T) this;
-        }
-
-        /// <summary>
-        /// Sets universe of weeklys contracts (if any) as a selection
-        /// </summary>
-        /// <returns></returns>
-        public T WeeklysOnly()
-        {
-            _type = Type.Weeklys;
-            return (T) this;
-        }
+        /// <returns>True if standard type</returns>
+        protected abstract bool IsStandard(Symbol symbol);
 
         /// <summary>
         /// Returns universe, filtered by contract type
         /// </summary>
-        /// <returns></returns>
+        /// <returns>Universe with filter applied</returns>
         internal T ApplyTypesFilter()
         {
             // memoization map for ApplyTypesFilter()
@@ -132,15 +134,15 @@ namespace QuantConnect.Securities
                 return res;
             };
 
-            _allSymbols = _allSymbols.Where(x =>
+            AllSymbols = AllSymbols.Where(x =>
             {
-                switch (_type)
+                switch (Type)
                 {
-                    case Type.Weeklys:
+                    case ContractExpirationType.Weekly:
                         return !memoizedIsStandardType(x);
-                    case Type.Standard:
+                    case ContractExpirationType.Standard:
                         return memoizedIsStandardType(x);
-                    case Type.Standard | Type.Weeklys:
+                    case ContractExpirationType.Standard | ContractExpirationType.Weekly:
                         return true;
                     default:
                         return false;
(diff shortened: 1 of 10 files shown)


Common/Securities/ContractSecurityFilterUniverse.cs
```diff
@@ -0,0 +1,295 @@
...
+        {
+            get
+            {
+                // underlying value changes over time, so accessing it makes universe dynamic
+                _isDynamic = true;
+                return _underlying;
+            }
+        }
+
+        protected BaseData _underlying;
```
Martin-Molinero, line 64: nit - code style protected should be `Underlying` and have xml docs
We should place privates first, later protected and finally public properties/fields

Common/Securities/ContractSecurityFilterUniverse.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        /// <summary>
+        /// True if the universe is dynamic and filter needs to be reapplied
+        /// </summary>
+        public bool IsDynamic => _isDynamic;
+
+        //TODO: Naming Conventions
+        internal bool _isDynamic;
+        protected Type _type = Type.Standard;
```
Martin-Molinero, line 73: Let's address the `TODO:` 😛 . Suggest we also rename `_type`. We might add a `remarks` to the `internal` type fields noting they are internal so that the extension methods can access them

Common/Securities/Future/FutureFilterUniverse.cs
```diff
@@ -19,40 +19,16 @@
...
-            {
-                return _isDynamic;
-            }
-        }
-
-        internal bool _isDynamic;
-
         /// <summary>
         /// Constructs FutureFilterUniverse
         /// </summary>
```
Martin-Molinero, line 34: I think it would be nice if this constructor should be calling the base one and not setting those fields here?

Common/Securities/Future/FutureFilterUniverse.cs
```diff
@@ -154,54 +58,6 @@ public FutureFilterUniverse ExpirationCycle(int[] months)
             var monthHashSet = months.ToHashSet();
             return this.Where(x => monthHashSet.Contains(x.ID.Date.Month));
```
Martin-Molinero, line 59: nit - Let's please add the missing documentation for this method

Common/Securities/ContractSecurityFilterUniverse.cs
```diff
@@ -0,0 +1,295 @@
...
+            _underlying = underlying;
+            _type = Type.Standard;
+            _isDynamic = false;
+        }
+
+        /// <summary>
+        /// Includes universe of non-standard weeklys contracts (if any) into selection
+        /// </summary>
+        /// <returns></returns>
+        public T IncludeWeeklys()
```
Martin-Molinero, line 97: Let's rereview the `xml` documentation of these classes and make sure it's all covered 👍

Tests/Common/Securities/FutureFilterTests.cs
```diff
@@ -61,6 +61,93 @@ public void FiltersExpiryRange()
             Assert.AreEqual(symbols[7], filtered[4]);
         }
 
+        [Test]
+        public void FiltersOutWeeklys()
```
Martin-Molinero, line 65: Sweet tests! 💯 
nit - maybe we should add `ByDefault` to the test name

Common/Securities/Future/FutureSymbol.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        /// <summary>
+        /// Returns true if the option is a standard contract that expires 3rd Friday of the month
+        /// </summary>
+        /// <param name="symbol">Future symbol</param>
+        /// <returns></returns>
+        public static bool IsStandard(Symbol symbol)
+        {
+            var date = symbol.ID.Date;
+            var symbolToCheck = symbol.HasUnderlying ? symbol.Underlying : symbol;
```
Martin-Molinero, line 34: Wonder in which case is this required `symbol.HasUnderlying ? symbol.Underlying : symbol` wouldn't really expect there to be an underlying symbol for the future?

Common/Securities/Future/FutureSymbol.cs
```diff
@@ -0,0 +1,68 @@
...
+                var standardDate = expiryFunction(date);
+
+                // If the date on this symbol and the nearest standard date are equal then it is a standard contract
+                if (date == standardDate)
+                {
+                    return true;
+                }
+
+                // Date is non-standard, return false
+                return false;
```
Martin-Molinero, line 49: nit - could directly `return date == standardDate`
nit - maybe we could rename `date` to something like `contractExpirationDate`?

Common/Securities/Future/FutureSymbol.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+            catch
+            {
+                Log.Error($"Could not find standard date for {symbolToCheck}, will be classified as weekly (non-standard)");
+                return false;
```
Martin-Molinero, line 54: Suggest by default we return true here, else since by default we filter weeklys we would filter out all contracts?

Martin-Molinero (commented):
Nice work 👍 great consolidation
Leaving some nit/style comments

Common/Securities/Future/FutureSymbol.cs
C-SELLERS: Definitely, I wasn't to sure how we wanted that to operate but it makes sense to keep the contract in if we can't verify its non-standard

>** Do we want symbols that are found to not have an expiry function to be flagged as non-standard???

Common/Securities/ContractSecurityFilterUniverse.cs
```diff
@@ -0,0 +1,340 @@
...
+                }
+            }).ToList();
+
+            return (T) this;
+        }
+
+        /// <summary>
+        /// Refreshes this filter universe
+        /// </summary>
+        /// <param name="allSymbols">All the options contract symbols</param>
```
C-SELLERS, line 158: Fix documentation

C-SELLERS (commented):
One thing I missed and I think we are good to go

Common/Securities/Future/FutureSymbol.cs
```diff
@@ -0,0 +1,61 @@
...
+            var contractExpirationDate = symbol.ID.Date;
+
+            try
+            {
+                // Use our FutureExpiryFunctions to determine standard contracts dates.
+                var expiryFunction = FuturesExpiryFunctions.FuturesExpiryFunction(symbol);
+                var standardExpirationDate = expiryFunction(contractExpirationDate);
+
+                // Return true if the dates match
+                return contractExpirationDate == standardExpirationDate;
```
Martin-Molinero, line 42: Suggest we compare the `Date` component here, `expiryFunction` might return the hour component, but for example I believe IB won't (see https://github.com/QuantConnect/Lean/blob/master/Brokerages/InteractiveBrokers/InteractiveBrokersBrokerage.cs#L2229). Let's give this case a live test too to be on the safe side👍

Martin-Molinero (commented):
Sweet 👍 just one last minor comment

Common/Securities/Future/FutureSymbol.cs
C-SELLERS: Good idea! I did run it through a live test with IB and saw the weeklys get filtered out so can confirm it works and there was no issue with the full DateTime comparison. But using the date component is probably a safer bet!

C-SELLERS: Made the change, testing IB again to confirm behaviour!

C-SELLERS: You were correct! The time components of the DateTimes did not match up so it is for the best that we use the `Date` component.

Martin-Molinero (approved):
🚀

Tests/Engine/DataFeeds/LiveTradingDataFeedTests.cs
```diff
@@ -2023,7 +2023,7 @@ public void HandlesFutureAndOptionChainUniverse(SecurityType securityType)
             else if (securityType == SecurityType.Future)
             {
                 var future = algorithm.AddFuture(Futures.Indices.SP500EMini, Resolution.Minute);
-                future.SetFilter(x => x);
+                future.SetFilter(x => x.IncludeWeeklys());
```
jaredbroad, line 2026: Can you please show the impact on the total count of contracts from this change?

jaredbroad (approved):

Tests/Engine/DataFeeds/LiveTradingDataFeedTests.cs
C-SELLERS: Total count of contracts is the same! 2 in 2 out!

The reason this was adjusted is the contract dates for ES we created are as follows:
12/19/19
3/19/20

But the standard contracts for those months are:
12/20/19
3/20/20

See [here](https://www.barchart.com/futures/quotes/ESZ20/historical-prices) for verification. 

Without the IncludeWeeklys() function on the filter those two dates get filtered out!

PR #4894, opened 2020-10-29
QB Set Start Date Relative to Data Release

### LINKED ISSUE #4892: Research default Time
#### Expected Behavior

- Research history request of the last 6 bars returns yesterdays data
#### Actual Behavior

- Research default Time is yesterday `0:00` so it returns data of the prev day
#### Potential Solution

- If it's after 9am NY, we could set the Time to today 0:00
#### Reproducing the Problem

on the 28/10:
image

#### System Information
QC cloud
#### Checklist

C-SELLERS:
#### Description

Set QuantBook start date to appropriate day relative to data release times

#### Related Issue

Closes #4892 

#### Motivation and Context

Fixes the issue described in #4892 where when starting a QB the StartDate is set to yesterday causing history requests to not reply with the most recent history.

N/A

Local QBs to prove Algorithm time is as expected. Test was run ~5:30 10/28

Before:
image

After:
image

#### Types of changes

#### Checklist:

### DIFF
--- a/Research/QuantBook.cs
+++ b/Research/QuantBook.cs
@@ -95,8 +95,8 @@ namespace QuantConnect.Research
                     _pandas = Py.Import("pandas");
                 }
 
-                // By default, set start date to end data which is yesterday
-                SetStartDate(EndDate);
+                // Issue #4892 : Set start time relative to TimeZone
+                SetStartDate(DateTime.UtcNow.ConvertFromUtc(TimeZone));
 
                 // Sets PandasConverter
                 SetPandasConverter();
--- a/Research/QuantBook.cs
+++ b/Research/QuantBook.cs
@@ -95,8 +95,21 @@ namespace QuantConnect.Research
                     _pandas = Py.Import("pandas");
                 }
 
-                // Issue #4892 : Set start time relative to TimeZone
-                SetStartDate(DateTime.UtcNow.ConvertFromUtc(TimeZone));
+                // Issue #4892 : Set start time relative to NY time 
+                // when the data is available from the previous day
+                var newYorkTime = DateTime.UtcNow.ConvertFromUtc(TimeZones.NewYork);
+                var hourThreshold = Config.GetInt("qb-data-hour", 9);
+                
+                // If it is after our hour threshold; then we can use today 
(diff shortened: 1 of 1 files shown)


Research/QuantBook.cs
```diff
@@ -95,8 +95,21 @@ public QuantBook() : base()
...
+                var hourThreshold = Config.GetInt("qb-data-hour", 9);
+                
+                // If it is after our hour threshold; then we can use today 
+                if (newYorkTime.Hour >= hourThreshold)
+                {
+                    SetStartDate(newYorkTime);
+                }
+                else
+                {
+                    SetStartDate(newYorkTime - TimeSpan.FromDays(1));
```
Martin-Molinero, line 110: Thanks! I think it would be clearer if we kept existing behavior in this `else`, using `EndDate` -> `SetStartDate(EndDate);` but for the `if` above I think we should also set the end date to `newYorkTime`? We might want to double check what are the resulting dates being set for them

C-SELLERS: EndDate is set by base (`QCAlgorithm`) to `DateTime.Now.AddDays(-1)`; 

In the case of EndDate being -1 days from StartDate it doesn't cause any problems with History requests. I think this would only be an issue in the case of QB running as an Algorithm. 

In the `else` statement if we use EndDate it is possible for a user in UTC + 12 (with default 9AM hour threshold) to pass the first `if` and still set the date to same value. I figure this way we cover the edge case and never make requests for data that is not available yet.

C-SELLERS: I just checked setting `EndDate` to newYorkTime, but it actually has a verification check that stops anything greater than yesterday being set :/

```
            if (end > DateTime.Now.Date.AddDays(-1))
            {
                end = DateTime.Now.Date.AddDays(-1);
            }
```

So we won't be able to change it to the current day.

Martin-Molinero: Makes sense thank you, let's create a separate issue to fix the end date being off from the algorithm timezone

Martin-Molinero (approved):

jaredbroad (approved):

PR #4916, opened 2020-11-05
Support Offset in DateRules MonthStart/End WeekStart/End

### LINKED ISSUE #4006: Adds Offset to DateRules.MonthEnd, DateRules.MonthStart
#### Expected Behavior
Be able to schedule an event x days after the start of the month or before the end of the month, like it is possible to add an offset AfterMarketOpen and BeforeMarketClose.

#### Actual Behavior
It is not possible to use schedule events to run events on a given day of the month.

#### Potential Solution
Create AfterMonthStart(int days) and BeforeMonthEnd(int days) method in DateRules.

#### Checklist

### LINKED ISSUE #4664: DateRules.WeekStart for FX incorrect
#### Expected Behavior
When using DateRules.WeekStart('EURUSD'), start of trading week should be Sunday 5 PM (EST)

#### Actual Behavior
Returns Tuesday

#### Reproducing the Problem
Logs:
```
2020-01-07 00:00:00
2020-01-14 00:00:00
2020-01-21 00:00:00
2020-01-28 00:00:00
2020-02-04 00:00:00
2020-02-11 00:00:00
2020-02-18 00:00:00
2020-02-25 00:00:00
```

Example Algo:
```
class TachyonNadionContainmentField(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2020, 1, 1)
        self.SetEndDate(2020,3,1)
        self.SetCash(100)
        self.SetTimeZone("America/New_York")
        self.AddForex("EURUSD", Resolution.Minute, Market.Oanda)

        # Print time that this is called every start of week; expect Sunday 5 PM EST        
        self.Schedule.On(self.DateRules.WeekStart('EURUSD'),
                         self.TimeRules.AfterMarketOpen('EURUSD',0),
                         self.PrintTime
        )
    def OnData(self, data):
        pass
    
    def PrintTime(self):
        self.Debug(self.Time)
```

#### System Information
LEAN master v9121 on QC Cloud

#### Checklist

- [x ] I have completely filled out this template
- [ x] I have confirmed that this issue exists on the current `master` branch
- [ x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [ x] I have provided detailed steps to reproduce the issue

C-SELLERS:
#### Description

- Expand DateRules to allow offsets in DateRules MonthStart, MonthEnd, WeekStart and Weekend. 
- Unify behavior of MonthStart/MonthEnd WeekStart/WeekEnd iterators and refactor into shared core logic.
- Include dynamic scheduling of events based on Exchange tradable dates. Takes into consideration holidays for start/end of week/month and offsets on all functions.
- Refactor DateRules.cs to reduce duplicate code in DateRule creation
- Expand tests for new behavior

#### Related Issue

Closes #4006 
Close  #4664

#### Motivation and Context

There was no method of scheduling events on any given day of the month or any given day of the week. This PR expands DateRules to allow that functionality.

For Week rules:
WeekStart Offsets Allowed : 
 - No Symbol: 0 to 6 since Monday is default, 6 is Sunday
 - w/ Symbol: Limited by security standard weekly schedule, ex: Spy is 0-4 Since M-F, BTCUSD is 0-6 since Su-S
WeekEnd Offsets Allowed : 
- No Symbol: 0 to 6 since Friday is default, 6 is Saturday
- w/ Symbol:  Limited by security standard weekly schedule, ex: Spy is 0-4 Since F-M, BTCUSD is 0-6 since S-Su

For Month rules
MonthStart Offsets Allowed : 0 to 30 since 1st is default, 0 is 1st, 30 is 31st OR last day of month, whichever is less
MonthEnd Offsets Allowed : 0 to 30 since Last day is default, 0 is Last day, 30 is (Last day - 30) OR 1st, whichever is greater

Scheduling documentation and DateRule documentation should be updated to reflect this new functionality

All original DateRules tests are passing and I included a lot of new DateRules tests to cover the new behavior. 

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Scheduling/DateRules.cs
+++ b/Common/Scheduling/DateRules.cs
@@ -31,7 +31,6 @@ namespace QuantConnect.Scheduling
     {
         private readonly DateTimeZone _timeZone;
         private readonly SecurityManager _securities;
-        private readonly string _format = "+#;-#;''";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DateRules"/> helper class
@@ -88,14 +87,14 @@ namespace QuantConnect.Scheduling
         /// <summary>
         /// Specifies an event should fire on each of the specified days of week
         /// </summary>
-        /// <param name="day">The day the event shouls fire</param>
+        /// <param name="day">The day the event should fire</param>
         /// <returns>A date rule that fires on every specified day of week</returns>
         public IDateRule Every(DayOfWeek day) => Every(new[] { day });
 
         /// <summary>
         /// Specifies an event should fire on each of the specified days of week
         /// </summary>
-        /// <param name="days">The days the event shouls fire</param>
+        /// <param name="days">The days the event should fire</param>
         /// <returns>A date rule that fires on every specified day of week</returns>
         public IDateRule Every(params DayOfWeek[] days)
         {
@@ -115,7 +114,7 @@ namespace QuantConnect.Scheduling
         /// <summary>
         /// Specifies an event should fire every day the symbol is trading
         /// </summary>
-        /// <param name="symbol">The symbol whose exchange is used to determine tradeable dates</param>
+        /// <param name="symbol">The symbol whose exchange is used to determine tradable dates</param>
         /// <returns>A date rule that fires every day the specified symbol trades</returns>
         public IDateRule EveryDay(Symbol symbol)
         {
@@ -135,14 +134,14 @@ namespace QuantConnect.Scheduling
         }
 
         /// <summary>
-        /// Specifies an event should fire on the first tradeable date for the specified
+        /// Specifies an event should fire on the first tradable date for the specified
         /// symbol of each month
         /// </summary>
         /// <param name="symbol">The symbol whose exchange is used to determine the first
-        /// tradeable date of the month</param>
+        /// tradable date of the month</param>
         /// <param name="daysOffset"> The amount of days to offset the schedule by; must
         /// be between 0 and 15. Use MonthEnd for unreachable dates</param>
-        /// <returns>A date rule that fires on the first tradeable date + offset for the
+        /// <returns>A date rule that fires on the first tradable date + offset for the
         /// specified security each month</returns>
         public IDateRule MonthStart(Symbol symbol, int daysOffset = 0)
         {
@@ -152,11 +151,8 @@ namespace QuantConnect.Scheduling
                 throw new ArgumentOutOfRangeException("daysOffset", "DateRules.MonthStart() : Offset must be between 0 and 15");
             }
 
-            // Get the symbol security, define our name, and return the new DateRule
-            var security = symbol == null ? null : GetSecurity(symbol);
-            var offsetString = daysOffset.ToString(_format, CultureInfo.InvariantCulture);
-            var name = symbol == null ? $"MonthStart{offsetString}" : $"{symbol.Value}: MonthStart{offsetString}";
-            return new FuncDateRule(name, (start, end) => MonthStartIterator(security, start, end, daysOffset));
+            // Create the new DateRule and return it
+            return new FuncDateRule(GetName(symbol, "MonthStart", daysOffset), (start, end) => MonthIterator(GetSecurity(symbol), start, end, daysOffset, true));
         }
 
         /// <summary>
@@ -171,14 +167,14 @@ namespace QuantConnect.Scheduling
         }
 
         /// <summary>
-        /// Specifies an event should fire on the last tradeable date for the specified
+        /// Specifies an event should fire on the last tradable date for the specified
         /// symbol of each month
         /// </summary>
         /// <param name="symbol">The symbol whose exchange is used to determine the last
-        /// tradeable date of the month</param>
+        /// tradable date of the month</param>
         /// <param name="daysOffset"> The amount of days to offset the schedule by; must
         /// be between -15 and 0. Use MonthStart for unreachable dates</param>
-        /// <returns>A date rule that fires on the last tradeable date for the specified security each month</returns>
+        /// <returns>A date rule that fires on the last tradable date + offset for the specified security each month</returns>
         public IDateRule MonthEnd(Symbol symbol, int daysOffset = 0)
         {
             // Check that our offset is allowed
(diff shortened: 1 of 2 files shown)


Tests/Common/Scheduling/DateRulesTests.cs
```diff
@@ -243,12 +396,85 @@ public void EndOfWeekWithSymbol()
...
+            rule = rules.WeekStart(1);
+            Assert.AreEqual("WeekStart+1", rule.Name);
+
+            rule = rules.WeekStart(Symbols.SPY);
+            Assert.AreEqual("SPY: WeekStart", rule.Name);
+
+            rule = rules.WeekStart(Symbols.SPY, 3);
+            Assert.AreEqual("SPY: WeekStart+3", rule.Name);
+        }
+
```
Martin-Molinero, line 477: Nice tests! 👍 
I think it would be nice to add some Crypto tests for which the market never closes

Common/Scheduling/DateRules.cs
```diff
@@ -220,113 +260,76 @@ private Security GetSecurity(Symbol symbol)
...
             }
         }
 
-        private static IEnumerable<DateTime> WeekStartIterator(Security security, DateTime start, DateTime end)
+        private static IEnumerable<DateTime> WeekIterator(Security security, DateTime start, DateTime end, int offset, bool searchForward)
         {
-            var skippedMarketClosedDay = false;
+            var scheduledDayOfWeek = searchForward ?
+                // For WeekStart we are searching forward so day of the week is : Monday (1) + offset
+                (DayOfWeek)((1 + offset) % 7) // Math to wrap enum around for Sunday case
```
Martin-Molinero, line 311: See https://github.com/QuantConnect/Lean/issues/4664, should using Monday, or some other day, as the start of the week depend on the target security/security type? Not sure how if worth the complexity that we would have to add, might get error-prone...
We could also keep the current implementation where we define week start as Monday and week end as Friday, with this `offset` feature it will allow users to achieve whatever weekday they want. Think it's worth adding a test case for `DateRules.WeekStart('EURUSD')` too.
Thoughts @jaredbroad @AlexCatarino @C-SELLERS

Martin-Molinero (commented):
Another nice one 💯

Common/Scheduling/DateRules.cs
C-SELLERS: Yeah now that you mention it it does make sense to have it like that, it would require some refactoring but it should be possible as long as I have access to the "WeekStart" day for any given security. 🤔 

Would we still leave Monday as default "WeekStart" for non-symbol rules?

Martin-Molinero: > it would require some refactoring but it should be possible as long as I have access to the "WeekStart" day for any given security. 🤔

yeah, think we could follow something like `start of the week: the day after the last non tradable day of the week`, for always open exchanges it could also default to Monday

> Would we still leave Monday as default "WeekStart" for non-symbol rules?

sounds good 💯

C-SELLERS:
After some major refactors I was able to get the desired behavior working. Some notes for discussion and clarity: 

MonthStart/MonthEnd:
- Offset is now tradable days offset for the given security. In the case of no security it just offsets the amount of given days. This replaces the per-review behavior which just offset from the 1st or Last of the month respectively and then found a nearby tradable date if it wasn't open on the scheduled day.
- The scheduled date is determined in the MonthIterator at the beginning of each month, where it determines the date by searching **tradable days only** meaning it will skip over all closures and they won't be counted in the offset.

EX:
MonthStart(SPY, 5) 
- Say the 1st is a Holiday and is Monday and on Tuesday the 2nd the Market is open
- The 2nd is assigned as the "starting" point for this month because the 1st is a closed; 
- Now we offset 5 trading days off of the 2nd; (for simplicity sake none of these are holidays) we only count open days so:
 3rd Wednesday, 4th Thursday, 5th Friday, 8th Monday, 9th Tuesday. Now the DateRule will contain the 9th.

I believe this is the behavior we are looking for but I want to be sure because it is a bit complex in the implementation. The WeekStart/WeekEnd is currently a ~little~ different and I want to make these uniform so let me know what you think.

WeekStart/Weekend:
- Offset is now tradable days offset as well, except it determines the day of the week ahead of iteration. So instead of skipping holidays in determining it's "starting" point, it only takes into consideration the base Security schedule for any given week.
- This means that if I create a WeekStart rule for SPY with offset 2, the expected day of the week is always Wednesday, it never takes into consideration if Monday or Tuesday is a holiday. But in the case Wednesday is a holiday it will seek forward the next trading day.

EX:
WeekStart(SPY, 3)
- It determines that the first trading day in the standard schedule of the week for Spy is Monday, so Monday + 3 is Thursday.
- Every week on Thursday it will check to see if the day is a holiday, if it isn't that Thursday gets added to the DateRule, if it is, it seeks forward the next trading day to add to the DateRule.


So although these are very similar there is a difference in determining the **"starting point"** to offset from. My question is which one of these behaviours do we actually want? Once we are sure I can make it uniform and go from there.

Common/Scheduling/DateRules.cs
```diff
@@ -122,87 +123,144 @@ public IDateRule EveryDay(Symbol symbol)
...
-        /// Specifies an event should fire on the first tradeable date for the specified
-        /// symbol of each month
+        /// Specifies an event should fire on the first tradable date + offset for the specified symbol of each month
         /// </summary>
-        /// <param name="symbol">The symbol whose exchange is used to determine the first
-        /// tradeable date of the month</param>
-        /// <returns>A date rule that fires on the first tradeable date for the specified security each month</returns>
-        public IDateRule MonthStart(Symbol symbol)
+        /// <param name="symbol">The symbol whose exchange is used to determine the first tradable date of the month</param>
+        /// <param name="daysOffset"> The amount of days to offset the schedule by; must be between 0 and 30</param>
```
Martin-Molinero, line 139: nit - think would be nice to add `The amount of **tradable** days` which makes it different from the other method overload, same for the other methods

Common/Scheduling/DateRules.cs
```diff
@@ -212,6 +270,12 @@ public IDateRule WeekEnd(Symbol symbol)
         /// <returns>The security object matching the given symbol</returns>
         private Security GetSecurity(Symbol symbol)
```
Martin-Molinero, line 271: It's almost as this method should actually return the `SecurityExchangeHours` instance directly, would save us from duplicating the `var securitySchedule = security == null ? SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork) : security.Exchange.Hours;` and passing around a security instance

Tests/Common/Scheduling/DateRulesTests.cs
```diff
@@ -102,6 +103,23 @@ public void StartOfMonthNoSymbolMidMonthStart()
             Assert.AreEqual(11, count);
         }
```
Martin-Molinero, line 105: Think it would be nice to add a test for `SPY/BTCUSD/EURUSD/Fut_SPY_Feb19_2016` with no offset and assert the base start/end weekday directly

Common/Scheduling/DateRules.cs
```diff
@@ -220,116 +284,138 @@ private Security GetSecurity(Symbol symbol)
...
+
+            // Determine the weekly base day and boundary to schedule off of
+            DayOfWeek weeklyBaseDay;
+            DayOfWeek weeklyBoundaryDay;
+            if (security == null)
+            {
+                // Searching forward Monday is baseDay, with boundary being the following Sunday
+                // Searching backward Friday is baseDay, with boundary being the previous Saturday
+                weeklyBaseDay = searchForward ? DayOfWeek.Monday : DayOfWeek.Friday;
+                weeklyBoundaryDay = searchForward ? DayOfWeek.Saturday + 1 : DayOfWeek.Sunday - 1;
```
Martin-Molinero, line 396: Curious why +-1 and not directly the days Sunday/Saturday?

Common/Scheduling/DateRules.cs
```diff
@@ -122,87 +123,144 @@ public IDateRule EveryDay(Symbol symbol)
...
+            }
+
+            return new FuncDateRule(GetName(null, "WeekEnd", -daysOffset), (start, end) => WeekIterator(null, start, end, daysOffset, false));
         }
 
         /// <summary>
-        /// Specifies an event should fire on the last tradeable date for the specified
+        /// Specifies an event should fire on the last - offset tradable date for the specified
         /// symbol of each week
-        /// </summary>
```
Martin-Molinero, line 199: nit - missing `</summary>`

Common/Scheduling/DateRules.cs
```diff
@@ -122,87 +123,144 @@ public IDateRule EveryDay(Symbol symbol)
...
+            }
+
+            return new FuncDateRule(GetName(null, "WeekStart", daysOffset), (start, end) => WeekIterator(null, start, end, daysOffset, true));
         }
 
         /// <summary>
-        /// Specifies an event should fire on the first tradeable date for the specified
-        /// symbol of each week
+        /// Specifies an event should fire on the first tradable date + offset for the specified
+        /// symbol each week;
```
Martin-Molinero, line 200: nit - `;`

Martin-Molinero (approved):
Sweet implementation 👍
Looks good, leaving minor details

PR #4936, opened 2020-11-14
Fixes weeklies parsing, causing certain futures to be inaccessible in QCAlgorithm

gsalaz98:
#### Description

  The FuturesExpiryFunction expects the contract month of the Future,
  not the expiration. As a result, the contract gets filtered as a
  weekly contract, rather than as a standard due to the discrepancy
  between the expiry dates when the contract month differs from the
  expiry date's month.

  A very important fact to note is that futures can and do expire prior
  to the contract month. BZ,(brent crude financial futures) expire two
  months prior to the contract month, CL one month prior, etc.

  There has been an addition that contains a "reverse" futures expiry function
  lookup table. We use this to lookup the contract month to re-calculate
  the Future expiry.

  This PR also fixes dairy and adds extra expiry dates. Dairy can have
  an expiry *after* the contract month, so a new path was added to the
  SymbolRepresentation to ensure that these contracts are loaded
  correctly.
#### Related Issue

N/A (thanks to Aram Darakchian for pointing out this issue on Slack)
#### Motivation and Context

Fixes bug, improving stability and usability of product

No

N/A - tests will be added soon
#### Types of changes

#### Checklist:

- [ ] All new and existing tests passed. (new tests must be added, tests might not be passing)

### DIFF
--- a/Common/Securities/Future/FuturesExpiryUtilityFunctions.cs
+++ b/Common/Securities/Future/FuturesExpiryUtilityFunctions.cs
@@ -335,16 +335,17 @@ namespace QuantConnect.Securities.Future
         }
 
         /// <summary>
-        /// Returns the number of months prior to the contract month that the future expires
+        /// Gets the number of months between the contract month and the expiry date.
         /// </summary>
-        /// <param name="underlying">The future symbol</param>
-        /// <param name="expiryDate">Expiry date to use to look up contract month delta</param>
-        /// <returns>The number of months prior it expires</returns>
-        public static int ExpiresInPreviousMonth(string underlying, DateTime? expiryDate = null)
+        /// <param name="underlying">The future symbol ticker</param>
+        /// <param name="expiryDate">Expiry date to use to look up contract month delta. Only used for dairy, since we need to lookup its contract month in a pre-defined table.</param>
+        /// <returns>The number of months between the contract month and the contract expiry</returns>
+        public static int GetDeltaBetweenContractMonthAndContractExpiry(string underlying, DateTime? expiryDate = null)
         {
             int value;
             if (expiryDate != null && _dairyUnderlying.Contains(underlying))
             {
+                // Dairy can expire in the month following the contract month.
                 var dairyReportDate = expiryDate.Value.Date.AddDays(1);
                 if (_reverseDairyReportDates.ContainsKey(dairyReportDate))
                 {
--- a/Common/Securities/Future/FuturesExpiryUtilityFunctions.cs
+++ b/Common/Securities/Future/FuturesExpiryUtilityFunctions.cs
@@ -25,6 +25,19 @@ namespace QuantConnect.Securities.Future
     /// </summary>
     public static class FuturesExpiryUtilityFunctions
     {
+        private static readonly Dictionary<DateTime, DateTime> _reverseDairyReportDates = FuturesExpiryFunctions.DairyReportDates
+            .ToDictionary(kvp => kvp.Value, kvp => kvp.Key);
+
+        private static readonly HashSet<string> _dairyUnderlying = new HashSet<string>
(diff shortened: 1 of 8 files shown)


Common/Securities/Future/FuturesExpiryUtilityFunctions.cs
```diff
@@ -325,15 +338,29 @@ public static DateTime DairyLastTradeDate(DateTime time, TimeSpan? lastTradeTime
...
         {
             int value;
+            if (expiryDate != null && _dairyUnderlying.Contains(underlying))
+            {
+                var dairyReportDate = expiryDate.Value.Date.AddDays(1);
+                if (_reverseDairyReportDates.ContainsKey(dairyReportDate))
+                {
+                    var contractMonth = _reverseDairyReportDates[dairyReportDate];
+                    // Gets the distance between two months in months
+                    return ((contractMonth.Year - dairyReportDate.Year) * 12) + contractMonth.Month - dairyReportDate.Month;
```
Martin-Molinero, line 353: Cool! So due to `Dairy can have an expiry after the contract month` this could return a negative value? Maybe we should rename this function so it's more precise

> DateTime? expiryDate = null

We need this argument because dairy isn't consistent?

Common/Util/LeanData.cs
```diff
@@ -524,7 +524,10 @@ public static string GenerateZipEntryName(Symbol symbol, DateTime date, Resoluti
 
                 case SecurityType.Future:
                     var expiryDate = symbol.ID.Date;
-                    var monthsToAdd = FuturesExpiryUtilityFunctions.ExpiresInPreviousMonth(symbol.ID.Symbol); 
+                    // GH review comment:
+                    // We skip adding the expiry component to maintain backwards compatibility. @Jay-Jay-D let me know if
+                    // you'd like to fix this and reprocess futures data
+                    var monthsToAdd = FuturesExpiryUtilityFunctions.ExpiresInPreviousMonth(symbol.ID.Symbol);// expiryDate.Date);
```
Martin-Molinero, line 530: @Jay-Jay-D

Martin-Molinero (commented):
Nice one! 💯 
Let's add a nice set of tests and some comments would be great too

Common/SymbolRepresentation.cs
```diff
@@ -141,20 +141,33 @@ public static FutureTickerProperties ParseFutureTicker(string ticker)
...
-            // As modulo works for [0, x), it's best to subtract 1 (as months are [1, 12] to convert to [0, 11]),
-            // do the modulo/integer division, then add 1 back on to get into the correct range again
-            month--;
-            year += month / 12;
-            month %= 12;
-            month++;
+            var year = doubleDigitsYear ? expiration.Year % 100 : expiration.Year % 10;// 2020
+            var month = expiration.Month; // 10
+
+            var contractMonthDelta = FuturesExpiryUtilityFunctions.ExpiresInPreviousMonth(underlying, expiration.Date);// 1
```
C-SELLERS, line 147: Looks good! 😄

Common/Securities/Future/FuturesExpiryUtilityFunctions.cs
C-SELLERS: I second the rename, kind of throws me off on in FutureSymbols. If I am not mistaken this is to calculate how many months in the past the expiry is from the contract date?

gsalaz98: Yes, will rename 👍 

And yes, it's to calculate the delta between the expiry and the contract month (positive if contract expires in month before contract month, negative if expiry is in month after contract month).

Martin-Molinero (approved):
Thanks for the tests! 👍

PR #4944, opened 2020-11-18
Bug 4927 Value Error

### LINKED ISSUE #4927: Research :  ValueError: cannot reindex from a duplicate axis
When running the provided BasicQuantBookTemplate.ipynb example in the research docker, the example stops at

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-2be348dae713> in <module>
      4 
      5 # Gets historical data of indicator
----> 6 bbdf = qb.Indicator(bb, "SPY", 360, Resolution.Daily)
      7 
      8 # drop undesired fields

ValueError: cannot reindex from a duplicate axis
```

#### Expected Behavior

The example notebook functions

#### Actual Behavior

The example notebook has a non-recoverable error

#### Reproducing the Problem

1. Follow instructions for Research docker installation
2. Start above example

#### System Information

MacOS X 10.15.7
Research docker image : ba23f2aa50f9

#### Checklist

C-SELLERS:
#### Description

After looking into the problem it became clear that the issue was not caused by pandas but by the data given to it. For the specific issue linked to this PR (#4927), in the set of indicator data points being converted to a `DataFrame` there was a property called `PercentB` for the BollingerBand indicator, this property had repeated index values after the date 4/18 which caused the Value Error. 

This property uses `CompositeIndicator` which was found to not fill forward properly and update the properties time. With these changes, if the conditions to update the `CompositeIndicator` are not met, it now creates a clone of the last point with the latest time. This way the indicator current.Time is always up to date.

#### Related Issue

Closes #4927 

#### Motivation and Context

Example research was breaking, ended up being unrelated to research but still required fixing.

N/A

All tests ran and passed.
Research notebook is working again.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Python/PandasConverter.cs
+++ b/Common/Python/PandasConverter.cs
@@ -135,18 +135,12 @@ namespace QuantConnect.Python
 
                 foreach (var kvp in data)
                 {
-                    var index = new List<DateTime>();
-                    var values = new List<double>();
-
-                    foreach (var item in kvp.Value)
-                    {
-                        index.Add(item.EndTime);
-                        values.Add((double)item.Value);
-                    }
-                    pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(values, index));
+                    var index = kvp.Value.Select(x => x.EndTime);
+                    var values = kvp.Value.Select(x => x.Value);
+                    pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(data: values, index: index));
                 }
 
-                return _pandas.DataFrame(pyDict, columns: data.Keys.Select(x => x.ToLowerInvariant()).OrderBy(x => x));
+                return _pandas.DataFrame.from_dict(pyDict);
             }
         }
 
--- a/Common/Python/PandasConverter.cs
+++ b/Common/Python/PandasConverter.cs
@@ -135,8 +135,15 @@ namespace QuantConnect.Python
 
                 foreach (var kvp in data)
                 {
-                    var index = kvp.Value.Select(x => x.EndTime);
-                    var values = kvp.Value.Select(x => x.Value);
+                    var index = new List<DateTime>();
+                    var values = new List<double>();
+
+                    foreach (var item in kvp.Value)
+                    {
+                        index.Add(item.EndTime);
+                        values.Add((double)item.Value);
+                    }
+
                     pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(data: values, index: index));
                 }
 
--- a/Common/Python/PandasConverter.cs
+++ b/Common/Python/PandasConverter.cs
@@ -140,7 +140,7 @@ namespace QuantConnect.Python
                     pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(data: values, index: index));
                 }
 
-                return _pandas.DataFrame.from_dict(pyDict);
+                return _pandas.DataFrame(pyDict, columns: data.Keys.Select(x => x.ToLowerInvariant()).OrderBy(x => x));
             }
         }
 
--- a/Indicators/CompositeIndicator.cs
+++ b/Indicators/CompositeIndicator.cs
@@ -156,6 +156,12 @@ namespace QuantConnect.Indicators
                     newLeftData = null;
                     newRightData = null;
                 }
+                else
+                {
+                    // Fill forward time
+                    Current = (IndicatorDataPoint)Current.Clone();
+                    Current.Time = MaxTime(updated);
+                }
             };
 
             Right.Updated += (sender, updated) =>
@@ -170,6 +176,12 @@ namespace QuantConnect.Indicators
                     // reset these to null after each update
                     newLeftData = null;
                     newRightData = null;
+                }
+                else
+                {
+                    // Fill forward time
+                    Current = (IndicatorDataPoint)Current.Clone();
+                    Current.Time = MaxTime(updated);
                 }
             };
         }
--- a/Tests/Indicators/CompositeIndicatorTests.cs
+++ b/Tests/Indicators/CompositeIndicatorTests.cs
@@ -93,5 +93,43 @@ namespace QuantConnect.Tests.Indicators
             Assert.AreEqual(left.PeriodsSinceMaximum, 0);
             Assert.AreEqual(right.PeriodsSinceMinimum, 0);
         }
+
+        [Test]
+        public void TimeAdvancesWithDataToBothSides()
+        {
+            var left = new Maximum("left", 2);
+            var right = new Minimum("right", 2);
+            var composite = new CompositeIndicator<IndicatorDataPoint>(left, right, (l, r) => l.Current.Value + r.Current.Value);
+
+            Assert.AreEqual(DateTime.MinValue, composite.Current.Time);
+
+            left.Update(DateTime.Today, 1m);
+            right.Update(DateTime.Today, -1m);
+
+            Assert.AreEqual(DateTime.Today, composite.Current.Time);
+
+            left.Update(DateTime.Today.AddDays(1), -1m);
+            right.Update(DateTime.Today.AddDays(1), 1m);
+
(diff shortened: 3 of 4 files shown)


Common/Python/PandasConverter.cs
```diff
@@ -135,18 +135,12 @@ public PyObject GetIndicatorDataFrame(IDictionary<string, List<IndicatorDataPoin
...
-                        values.Add((double)item.Value);
-                    }
-                    pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(values, index));
+                    var index = kvp.Value.Select(x => x.EndTime);
+                    var values = kvp.Value.Select(x => x.Value);
+                    pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(data: values, index: index));
                 }
 
-                return _pandas.DataFrame(pyDict, columns: data.Keys.Select(x => x.ToLowerInvariant()).OrderBy(x => x));
+                return _pandas.DataFrame.from_dict(pyDict);
```
Martin-Molinero, line 143: `columns: data.Keys.Select(x => x.ToLowerInvariant()).OrderBy(x => x)` the `.OrderBy` is now missing? I see we already set the key to lower above so that part is okay. Suggest avoid these changes if unrequited?

Indicators/CompositeIndicator.cs
```diff
@@ -171,6 +177,12 @@ private void ConfigureEventHandlers()
                     newLeftData = null;
                     newRightData = null;
                 }
+                else
+                {
+                    // Fill forward time
+                    Current = (IndicatorDataPoint)Current.Clone();
+                    Current.Time = MaxTime(updated);
```
Martin-Molinero, line 184: Cool! We need a nice collection of tests around this. 
I think during _each_ update, we will always come through this `if` because Right/Left not updated sequentially but first one and later the other? Maybe we need some extra logic for that case?

Martin-Molinero (changes requested):
Nice find! Let's add a nice collection of tests for the CompositeIndicator behavior around this issue

Common/Python/PandasConverter.cs
C-SELLERS: Yeah I actually wanted to test that out and see if it mixed up the columns. Turns out it doesn't so there I went ahead and added it back. I can revert to the original if you like. Yesterday when I was digging for the problem I was poking around in this function to find the error.

Indicators/CompositeIndicator.cs
C-SELLERS: I just added two tests to show how the time on the consolidator advances. One of them `TimeAdvancedWithDataToOnlyOneSide`  is red/green'd from master to this branch. I am also adding one now for checking the data point after one update (since it shouldn't change until both are received) and after.

The pre existing `if` statement that I added the `else` to takes care of them not being updated sequentially by tracking new datapoints for either side, once both have a new datapoint it will updated.

Tests/Indicators/CompositeIndicatorTests.cs
```diff
@@ -93,5 +93,43 @@ public void CallsDelegateCorrectly()
...
+
+            left.Update(DateTime.Today, 1m);
+            right.Update(DateTime.Today, -1m);
+
+            Assert.AreEqual(DateTime.Today, composite.Current.Time);
+
+            left.Update(DateTime.Today.AddDays(1), -1m);
+            right.Update(DateTime.Today.AddDays(1), 1m);
+
+            Assert.AreEqual(DateTime.Today.AddDays(1), composite.Current.Time);
```
gsalaz98, line 114: Leaving a comment for posterity: usually we'll want to save the `DateTime.Today/Now/UtcNow` in a variable to ensure that time doesn't change on us while we're running our tests. 

In this case, it's unlikely that the test will fail, but it would fail at midnight if it was ran then :)

Common/Python/PandasConverter.cs
```diff
@@ -140,7 +140,7 @@ public PyObject GetIndicatorDataFrame(IDictionary<string, List<IndicatorDataPoin
                     pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(data: values, index: index));
                 }
 
-                return _pandas.DataFrame.from_dict(pyDict);
+                return _pandas.DataFrame(pyDict, columns: data.Keys.Select(x => x.ToLowerInvariant()).OrderBy(x => x));
```
gsalaz98, line 143: I think that if the column names are not already sorted, this will give back the dataframe with the wrong column names.

Let's make a test to make sure that the column names are in the order we expect 👍

gsalaz98 (changes requested):
One concern with the ordering of the pandas columns. Might have been a pre-existing bug, but I want to make sure that we fix it now if it is one.

Common/Python/PandasConverter.cs
C-SELLERS: That's exactly what I was thinking in the conversation with Martin above, but upon testing it, it actually matched the columns to the correct values in the dictionary. I'll go ahead and figure a way to add a test to confirm that behavior.

Tests/Indicators/CompositeIndicatorTests.cs
C-SELLERS: Gotcha, better safe than sorry 😁

Common/Python/PandasConverter.cs
```diff
@@ -135,15 +135,9 @@ public PyObject GetIndicatorDataFrame(IDictionary<string, List<IndicatorDataPoin
...
-                    var values = new List<double>();
-
-                    foreach (var item in kvp.Value)
-                    {
-                        index.Add(item.EndTime);
-                        values.Add((double)item.Value);
-                    }
-                    pyDict.SetItem(kvp.Key.ToLowerInvariant(), _pandas.Series(values, index));
+                    var index = kvp.Value.Select(x => x.EndTime);
+                    var values = kvp.Value.Select(x => x.Value);
```
AlexCatarino, line 139: In PandasConverter+PandasData, we have opted to use `List` instead of `Linq` expressions for performance reasons, since we would enumerate `kvp.Value` only once.

C-SELLERS: 💯 Changed it back!

gsalaz98 (approved):

Indicators/CompositeIndicator.cs
```diff
@@ -171,6 +177,12 @@ private void ConfigureEventHandlers()
                     newLeftData = null;
                     newRightData = null;
                 }
+                else
+                {
+                    // Fill forward time
+                    Current = (IndicatorDataPoint)Current.Clone();
```
Martin-Molinero, line 183: Would like to know a bit more detailed _why_ is this required:
- Why is there `repeated index values after the date 4/18`
- Why should the composite indicator FF values, he is already triggering an event when a new value is available, so it sounds more like an issue in the consumer side 🤔

athon-millane:
@C-SELLERS I have merged your changes to [my own fork](https://github.com/athon-millane/Lean) and am still seeing the same error as #4927 in my notebook. I gather this would be because the base image isn't updated, if I want to spin my own image is this as simple as building from [this Dockerfile](https://github.com/QuantConnect/Lean/blob/master/DockerfileJupyter) with the updated code?

Thanks

C-SELLERS:
Hey Athon! Yeah you will have to build the new image for it to work, you'll have to do two actually because the research image builds off of the Lean image. 

So first build the Linux version of Lean; you can use `"/p:ForceLinuxBuild=true"` if on windows; then build the Lean Docker image with this [DockerFile](https://github.com/QuantConnect/Lean/blob/master/Dockerfile), then finally build the research image with this [DockerFile](https://github.com/QuantConnect/Lean/blob/master/DockerfileJupyter#L7) but be sure to the FROM statement to the new Lean image you created in the previous step. 

Then you should be good to go! If using the launch scripts make sure you enter your custom image name!

Martin-Molinero:
Closing for now, see https://github.com/QuantConnect/Lean/pull/4944#discussion_r530461527

PR #4953, opened 2020-11-20
Fixes inability to parse negative strike prices in SecurityIdentifier

### LINKED ISSUE #4952: Negative strike prices makes `SecurityIdentifier` throw OverflowException
#### Expected Behavior

We are able to create an option with a negative strike price
#### Actual Behavior

Negative strike prices cause the SecurityIdentifier to throw an OverflowException
#### Potential Solution

Use `long` rather than `ulong` as data type in `NormalizeStrike`
#### Reproducing the Problem

```c#
[Test]
public void NegativeStrikePriceDoesNotThrow()
{
    var future = Symbol.CreateFuture(
        "CL",
        Market.NYMEX,
        new DateTime(2020, 5, 20));

    Assert.DoesNotThrow(() =>
    {
        Symbol.CreateOption(
            future,
            Market.NYMEX,
            OptionStyle.American,
            OptionRight.Call,
            -50,
            new DateTime(2020, 4, 16));
    });
}
```
#### System Information

N/A
#### Checklist

gsalaz98:
#### Description

Fixes negative strike price parsing.

  * Adds new tests ensuring backwards compat and no throwing w/ negative
    strike prices

#### Related Issue

FIxes #4952 
#### Motivation and Context

Allow for negative strike prices, in case a future drops below 0 and a new option chain entry is added.

No

New tests
#### Types of changes

#### Checklist:


gsalaz98:
This fix isn't ready yet. We'll need to either parse the wrap-around behavior this PR enables,
or we'll need to set a negative bit flag in the strike.

Closing for now while I reassess a way to enable negative strikes.

Tests/Common/SymbolTests.cs
```diff
@@ -556,6 +557,82 @@ public void TestIfFridayLastTradingDayIsHolidaysThenMoveToPreviousThursday()
...
+                OptionStyle.American,
+                OptionRight.Call,
+                -50,
+                new DateTime(2020, 4, 16));
+
+            Assert.AreEqual(-50, option.ID.StrikePrice);
+
+            var properties = (ulong)typeof(SecurityIdentifier)
+                .GetField("_properties", BindingFlags.NonPublic | BindingFlags.Instance)
+                .GetValue(option.ID);
```
Martin-Molinero, line 580: Suggest use `ToString()` and later the `Parse()`. Let's add asserts on all the other properties too, so we are sure it's not breaking 👍

Common/SecurityIdentifier.cs
```diff
@@ -682,12 +691,39 @@ private static ulong NormalizeStrike(decimal strike, out ulong scale)
...
+            }
+
+            var encodedStrike = (long)strike;
+            if (strike < 0)
+            {
+                // Flip the sign
+                encodedStrike = -encodedStrike;
+
+                // Sets the 20th bit equal to 1
+                encodedStrike ^= 1 << 19;
```
Martin-Molinero, line 723: Curious why use `^=` and not normal or `|=`, using `^` leaves the window open for the bit not to be `1`? If it happened it would mean there is a bug in the logic though

Tests/Common/SymbolTests.cs
```diff
@@ -556,6 +557,82 @@ public void TestIfFridayLastTradingDayIsHolidaysThenMoveToPreviousThursday()
...
+                var option = Symbol.CreateOption(
+                    equity,
+                    Market.USA,
+                    OptionStyle.American,
+                    OptionRight.Call,
+                    3600.75m,
+                    new DateTime(2020, 5, 21));
+            });
+        }
+
```
Martin-Molinero, line 635: Would be nice to add test cases where the strike is off the limits and we assert it behaves as we expect and throws and exception

Martin-Molinero (commented):
Nice one! Looks good 👍 just minor comments

Tests/Common/Securities/SecurityIdentifierTests.cs
```diff
@@ -389,6 +389,347 @@ public void GenerateBase_UsesProvidedSymbol_WhenBaseDataTypeIsNull()
...
+                    -999999m,
+                    new DateTime(2020, 5, 21));
+            });
+        }
+
+        [Test]
+        public void HighPrecisionNegativeFractionalNonBoundApproachingNumberStrikePricesThrows()
+        {
+            var equity = Symbol.Create("AAPL", SecurityType.Equity, Market.USA);
+            Assert.Throws<ArgumentException>(() =>
```
Martin-Molinero, line 721: This test is the same as `HighPrecisionNegativeFractionalNumberStrikePricesThrows`. `HighPrecisionNegativeWholeNumberStrikePricesThrows` seems to cover the same case as `HighPrecisionNegativeWholeVeryOutOfBoundsNumberStrikePricesThrows`. `HighPrecisionFractionalNumberStrikePricesThrows` is the same as `HighPrecisionFractionalNonBoundApproachingNumberStrikePricesThrows`.

Let's create a single test with all these different test cases, we can add a flag to know if we expect an exception. Something like ` [TestCase("-999999", true)]` strike/throws and parse the string to decimal?

Martin-Molinero (approved):
Looks good 👍 
Minor request to join some of the unit tests

jaredbroad (dismissed):
Please store a range of symbol hashes from a CL option backtest over the period CL went negative; 

 - Let's show the AAPL options symbol hashes over the period remaining the same
 - And the CL symbol hashes changing for the period CL was negative.

Tests/Common/Securities/SecurityIdentifierTests.cs
```diff
@@ -389,6 +391,284 @@ public void GenerateBase_UsesProvidedSymbol_WhenBaseDataTypeIsNull()
...
+                    Market.USA,
+                    OptionStyle.American,
+                    OptionRight.Call,
+                    (decimal)strike, // strike decimal precision is limited to 4 decimal places only
+                    new DateTime(2020, 5, 21));
+            });
+        }
+
+        [Test, Ignore("Requires complete option data to validate chain")]
+        public void ValidateAAPLOptionChainSecurityIdentifiers()
```
Martin-Molinero, line 629: Cool! 👍

gsalaz98: For the record, prior to constructing this test, I tested validation for ALL option chains for ALL equities on a day around 2020-11-20. The test (exactly the same as this) passed successfully.

PR #4961, opened 2020-11-25
Fixes issue parsing SI data w/ unknown enum value

gsalaz98:
#### Description

A previously unknown value in the data causes crashes without any form of exception handling.
I've added exception handling + logging so that the conversion of the data can occur regardless of the enum value encountered.

#### Related Issue

N/A
#### Motivation and Context

Stability and data correctness

No

New unit test, reprocessed data for day with new enum value successfully

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Custom/SmartInsider/SmartInsiderExecutionType.cs
+++ b/Common/Data/Custom/SmartInsider/SmartInsiderExecutionType.cs
@@ -41,6 +41,11 @@ namespace QuantConnect.Data.Custom.SmartInsider
         /// Under a specific agreement between the issuer and shareholder
         /// </summary>
         [EnumMember(Value = "Off Market Agreement")]
-        OffMarket
+        OffMarket,
+
+        /// <summary>
+        /// Field is not in this enum
+        /// </summary>
+        Error
     }
 }
--- a/Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
+++ b/Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
@@ -141,7 +141,7 @@ namespace QuantConnect.Data.Custom.SmartInsider
                 }
                 catch (JsonSerializationException)
                 {
-                    Log.Error($"SmartInsiderIntention.FromRawData: New unexpected entry found {tsv[1]}. Parsed as NotSpecified.");
+                    Log.Error($"SmartInsiderIntention.FromRawData(): New unexpected entry found {tsv[1]}. Parsed as NotSpecified.");
                 }
             }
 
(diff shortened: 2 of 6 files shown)


Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
```diff
@@ -171,10 +171,54 @@ public override void FromRawData(string line)
...
+            if (!string.IsNullOrWhiteSpace(tsv[43]))
+            {
+                try
+                {
+                    Execution = JsonConvert.DeserializeObject<SmartInsiderExecution>($"\"{tsv[43]}\"");
+                }
+                catch (JsonSerializationException)
+                {
+                    Log.Error($"SmartInsiderIntention.FromRawData(): New unexpected entry found for Execution: {tsv[43]}. Parsed as Error.");
+                    Execution = SmartInsiderExecution.Error;
```
Martin-Molinero, line 184: This would mean this data point is malformed? Wonder if we should throw instead and have the caller discard it versus emitting invalid data points, what are your thoughts too @Jay-Jay-D

Common/Data/Custom/SmartInsider/SmartInsiderExecutionType.cs
```diff
@@ -41,6 +41,11 @@ public enum SmartInsiderExecution
         /// Under a specific agreement between the issuer and shareholder
         /// </summary>
         [EnumMember(Value = "Off Market Agreement")]
-        OffMarket
+        OffMarket,
+
+        /// <summary>
+        /// Field is not in this enum
+        /// </summary>
+        Error
```
Martin-Molinero, line 49: I would expect us to be adding new enum values in this PR too, the ones that triggered this error?

Common/Data/Custom/SmartInsider/SmartInsiderIntention.cs
Jay-Jay-D: This is a good approach, I think. Cases like this one are unexpected and - very - sporadic, we shouldn’t make the converter explode.

Martin-Molinero: Right 💯, my point was not making the converter explode but just discard this data point directly

gsalaz98: I'm thinking that we should model backtesting exactly as we would live trading. 

In the event of a sporadic signal, we might want to let the user deal with the data point rather than discarding it outright and introducing some sort of survivorship bias.

Jay-Jay-D (approved):
Good catch!

Common/Data/Custom/SmartInsider/SmartInsiderExecutionType.cs
gsalaz98: The value provided to us is non-standard and the value it represents can not be mapped to an expected value, nor can it be inferred.

The value in question is an error, and should be parsed as such.

Martin-Molinero (approved):

PR #4974, opened 2020-11-28
Bug 4947 OnOrderEvent exceptions

### LINKED ISSUE #4947: Exception in OnOrderEvent does not fail regression test
#### Expected Behavior

- If regression test algorithm throws an exception in OnOrderEvent the algorithm stops and it fails
#### Actual Behavior

- If regression test algorithm throws an exception in OnOrderEvent the algorithm stops but does NOT fail
#### Potential Solution

- Review `BrokerageTransactionHandler`, add test reproducing issue
#### Reproducing the Problem

```
        {
            SetStartDate(2013, 10, 07);
            SetEndDate(2013, 10, 11);
        }
        {
            if (!Portfolio.Invested)
            {
                SetHoldings(_spy, 1);
            }
        }
        {
            throw new Exception("OnOrderEvent exception");
        }
```
#### System Information

N/A
#### Checklist

C-SELLERS:
#### Description

Have `AlgorithmManager` return when a runtime error is detected, this way we can force it to return a failed state instead of overwriting it with `Completed` status when breaking out of the loop.

#### Related Issue

Closes #4947 

#### Motivation and Context

Following a `RunTimeError` status we should expect that the algorithm will not only stop but also maintain that status instead of replacing with `Completed`. 

N/A

Regressions ran and passed successfully.
New regression added that is intended to fail.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OnOrderEventExceptionRegression.cs
+++ b/Algorithm.CSharp/OnOrderEventExceptionRegression.cs
@@ -1,0 +1,80 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using QuantConnect.Orders;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Basic template algorithm simply initializes the date range and cash. This is a skeleton
+    /// framework you can use for designing an algorithm.
+    /// </summary>
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="using quantconnect" />
+    /// <meta name="tag" content="trading and orders" />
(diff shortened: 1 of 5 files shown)


Engine/AlgorithmManager.cs
```diff
@@ -689,6 +689,14 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
                 results.SendStatusUpdate(AlgorithmStatus.Deleted);
             }
 
+            //Runtime Error.
+            if (_algorithm.Status == AlgorithmStatus.RuntimeError)
+            {
+                Log.Trace("AlgorithmManager.Run(): Algorithm RuntimeError...");
+                Log.Trace("RuntimeError:" + _algorithm.RunTimeError);
```
Martin-Molinero, line 696: Not sure we need `Log.Trace("RuntimeError:" + _algorithm.RunTimeError);` since it's being logged before? Or maybe above, when detecting the runtimeError, instead of `break` we could directly use `return` as done with all the other `AlgorithmStatus.RuntimeError` cases in this loop, keep it normalized 🤔

Algorithm.CSharp/OnOrderEventExceptionRegression.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using QuantConnect.Orders;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression Algorithm for testing engine behavior with throwing errors in OnOrderEvent
+    /// Should result in a RunTimeError status.
```
Martin-Molinero, line 26: Cool! nit - maybe we can add a ref for the GH issue

Martin-Molinero (commented):
Nice! minor suggestion

Engine/AlgorithmManager.cs
C-SELLERS: Gotcha, I was little worried about line 176 in AlgorithmManager  breaking out of the loop after the error is posted and then it being set the complete again, but now that I took a deeper look at the sequence it makes sense that it would catch it on line 327 after processing all the data for that time slice.

Martin-Molinero: Yeah exactly 👍 because the status isn't being set before but by the AlgorithmManager at 329

Martin-Molinero (approved):
Nice! thanks 👍

PR #4975, opened 2020-11-30
Block SetWarmUp after Algorithm has compeleted initialization

### LINKED ISSUE #4939: SetWarmup Throws Exception When Not Called in Initialize
#### Expected Behavior
Notify users that `QCAlgorithm.SetWarmup` cannot be used after initialization. 

#### Actual Behavior
`QCAlgorithm.SetWarmup` can be called after initialization which doesn't do any warm up.

#### Potential Solution
Throw exception if `SetWarmup` is called when `IsWarmingUp` has already been flipped to `false`. The message should tell the user that this method should be called in `Initialize` only.

#### Reproducing the Problem
Call `SetWarmup` in `OnData`.

#### Checklist

C-SELLERS:
#### Description

Now we throw an error when SetWarmUp is called after Algorithm Initialization

#### Related Issue

Closes #4939

#### Motivation and Context

Shouldn't be able to mess with warmup settings after warmup has already occurred. 

N/A

All previous regressions pass, included new regression that intentionally throws and is checked for runtime error status.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/WarmUpAfterIntializeRegression.cs
+++ b/Algorithm.CSharp/WarmUpAfterIntializeRegression.cs
@@ -1,0 +1,61 @@
+﻿/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm to test warming up after initialize behavior, should throw if used outside of initialize
+    /// Reference GH Issue #4939
+    /// </summary>
+    public class WarmUpAfterIntializeRegression : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 07);  //Set Start Date
+            SetEndDate(2013, 10, 11);    //Set End Date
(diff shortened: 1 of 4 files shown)


Algorithm/QCAlgorithm.History.cs
```diff
@@ -51,6 +51,11 @@ public bool IsWarmingUp
         /// <param name="timeSpan">The amount of time to warm up, this does not take into account market hours/weekends</param>
         public void SetWarmup(TimeSpan timeSpan)
         {
+            if (!IsWarmingUp)
```
Martin-Molinero, line 54: Cool!
I think more precisely we could use the algorithms `_locked` field, which is false until after being initialized, seems more exact for the behavior we want? Think calling `SetWarmup` during warmup should also throw.

Maybe we could make this method directly call `SetWarmup(TimeSpan timeSpan, Resolution resolution)`. With resolution null. Same for `public void SetWarmup(int barCount)` -> `public void SetWarmup(int barCount, Resolution resolution)`

Algorithm.CSharp/WarmUpAfterIntializeRegression.cs
```diff
@@ -0,0 +1,61 @@
...
+            SetStartDate(2013, 10, 07);  //Set Start Date
+            SetEndDate(2013, 10, 11);    //Set End Date
+            SetCash(100000);
+            var equity = AddEquity("SPY");
+        }
+
+        public override void OnData(Slice slice)
+        {
+            // Should throw and set Algorithm status to be runtime error
+            SetWarmUp(TimeSpan.FromDays(2));
```
Martin-Molinero, line 41: Nice 👍

Tests/RegressionTests.cs
```diff
@@ -77,7 +77,8 @@ private static TestCaseData[] GetRegressionTestParameters()
             var nonDefaultStatuses = new Dictionary<string, AlgorithmStatus>
             {
                 {"TrainingInitializeRegressionAlgorithm", AlgorithmStatus.RuntimeError},
-                {"OnOrderEventExceptionRegression", AlgorithmStatus.RuntimeError}
+                {"OnOrderEventExceptionRegression", AlgorithmStatus.RuntimeError},
+                {"WarmUpAfterIntializeRegression", AlgorithmStatus.RuntimeError }
```
Martin-Molinero, line 81: A separate issue, but starting to think this end state should probably be owned by the algorithm itself, not this 3rd party 🤔

C-SELLERS: I could see that, maybe we should consider extending the IRegressionAlgorithmDefinition interface to contain a member for expected status?

Algorithm/QCAlgorithm.History.cs
C-SELLERS: Gotcha!

Martin-Molinero (approved):

PR #4978, opened 2020-12-02
Reserve Names for Pandas Mapper

### LINKED ISSUE #4886: Add Reserved Names in Pandas Converter
#### Expected Behavior
Be access `low` column in pandas Dataframe with
```python
history = self.History(symbol, 1, Resolution.Minute).low
```
#### Actual Behavior
The statement above raises the following exception:
```AttributeError : 'DataFrame' object has no attribute 'LOW R735QTJ8XC9X'```

#### Potential Solution
Add reserved names in `PandasConverter` to not convert "low" into "LOW R735QTJ8XC9X". We will need to use "LOW" instead.

#### Reproducing the Problem
```python
class NadionDynamicChamber(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2020, 1, 6)
        symbol = self.AddEquity('LOW').Symbol
        history = self.History(symbol, 1, Resolution.Minute)
        self.Quit(history.low)
```
#### Checklist

C-SELLERS:
#### Description

Force pandas mapper to not convert strings 'low' 'high' 'close' and 'open' to symbol before indexing dataframe. Now those keywords are reserved and won't attempt to be converted.

#### Related Issue

Closes #4886 

#### Motivation and Context

Pandas bug specific to stock ticker LOW, when requesting history for LOW and attempting to access history `.low` values pandas would convert the request for `.low` to symbol LOW which pandas dataframe does not have a value for thus causing the issue.

I reserved more than just low since it is possible that other tickers may fit some of those column names in the future. 

N/A

- Added a test that red/greened from master to this pr.
- Confirmed original behavior DataFrame.loc["LOW"] can return a dataframe just for LOW ticker
- Confirmed DataFrame.low is now working

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Python/PandasConverterTests.cs
+++ b/Tests/Python/PandasConverterTests.cs
@@ -288,7 +288,7 @@ def Test(dataFrame):
                 var Test = PythonEngine.ModuleFromString("testModule",
     $@"
 def Test1(dataFrame):
-    # Should not throw, access LOW ticker data
+    # Should not throw, access all LOW ticker data
     data = dataFrame.loc['LOW']
 def Test2(dataFrame):
     # Bad accessor, expected to throw
@@ -297,7 +297,7 @@ def Test3(dataFrame):
     # Bad key, expected to throw
     data = dataFrame.loc['low']
 def Test4(dataFrame):
-    # Should not throw, also access data lows
+    # Should not throw, access data column low for all tickers
     data = dataFrame.low
 ");
 
--- a/Tests/Python/PandasConverterTests.cs
+++ b/Tests/Python/PandasConverterTests.cs
@@ -256,7 +256,7 @@ def Test(dataFrame):
     if data.__len__() != 10:
         raise Exception('Expected 10 data points')
     lowData = data.low
-    if data.empty is 0:
+    if lowData.empty:
         raise Exception('LOW history low data is empty')").GetAttr("Test");
 
                 Assert.DoesNotThrow(() => test(dataFrame));
--- a/Tests/Python/PandasConverterTests.cs
+++ b/Tests/Python/PandasConverterTests.cs
@@ -264,6 +264,56 @@ def Test(dataFrame):
             }
         }
 
+        [Test]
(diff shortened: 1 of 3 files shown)


AlexCatarino (changes requested):
Thank you, @C-SELLERS .
I think we should add a test that throws an exception for completeness. For example, what if the user tries to access the "low" (lower case) key? If should throw if we are referring to the Symbol, and should not if referring to the field.

Tests/Python/PandasConverterTests.cs
```diff
@@ -223,6 +223,47 @@ public void HandlesNullableValues()
...
+                dynamic test = PythonEngine.ModuleFromString("testModule",
+    $@"
+def Test(dataFrame):
+    data = dataFrame.loc['LOW']
+    if data.empty:
+        raise Exception('LOW history data is empty')
+    if data.__len__() != 10:
+        raise Exception('Expected 10 data points')
+    lowData = data.low
+    if data.empty is 0:
```
Martin-Molinero, line 259: `if data.empty is 0` should check `lowData` instead?

C-SELLERS: My bad, thanks boss!!!!

AlexCatarino (approved):
Thank you for the extra tests! 👍

Tests/Python/PandasConverterTests.cs
```diff
@@ -223,6 +223,97 @@ public void HandlesNullableValues()
...
+def Test2(dataFrame):
+    # Bad accessor, expected to throw
+    data = dataFrame.LOW
+def Test3(dataFrame):
+    # Bad key, expected to throw
+    data = dataFrame.loc['low']
+def Test4(dataFrame):
+    # Should not throw, access data column low for all tickers
+    data = dataFrame.low
+");
```
Martin-Molinero, line 302: Sweet test! 👍

Martin-Molinero (approved):

Martin-Molinero (approved):

PR #4981, opened 2020-12-02
Api Update November 2020

### LINKED ISSUE #4983: API Changes not reflected in Lean
#### Expected Behavior

That any changes to the web API are reflected in Lean

#### Actual Behavior

Web API endpoint changes are breaking Lean API functions.

#### Potential Solution

Update API functions in Lean and merge into production

#### Reproducing the Problem

Use Backtest/* endpoints in tests.

#### System Information

N/A

#### Checklist

C-SELLERS:
#### Description

Update API functions in Lean to reflect web API changes. 

#### Related Issue

Closes #4983

#### Motivation and Context

We need both the Web API and Lean API functions to be in sync for all functionality to be maintained.

There is a functional change in the behavior of the resulting `Backtest` object.

Now instead of having a `Result` obj as a member of `Backtest` (Backtest.Result) there is no longer a result object. Instead the members originally contained in the Result obj are now included in the `Backtest` obj. 

So before where you would use `Backtest.Result.Charts` you will just use `Backtest.Charts`

API Tests

#### Types of changes

#### Checklist:


jaredbroad (approved):

C-SELLERS:
One error still persists in my tests but it is completely independent of lean and only springs up in special cases.

It is the SortinoRatio being sent by the web api in exponential format that breaks my de-serializing.
```{"Input string '7.9228162514264338e+28' is not a valid decimal. Path 'backtest.alphaRuntimeStatistics.SortinoRatio', line 1, position 16196."}```

Again this only occurs when the SortinoRatio is a very large value. It can be fixed at the endpoint as it has before, just in the latest changes to the API it must've been lost.

File: Api/Api.cs [hunk 1]
```diff
@@ -409,9 +415,53 @@ public Backtest ReadBacktest(int projectId, string backtestId)
...
+                    if (chartResponse.Success)
+                    {
+                        updatedCharts.Add(chart.Key, chartResponse.Backtest.Charts[chart.Key]);
+                    }
+                }
+
+                // Update our result
+                foreach(var updatedChart in updatedCharts)
+                {
+                    result.Backtest.Charts[updatedChart.Key] = updatedChart.Value;
```
Martin-Molinero, line 455: Why not directly update the `result.Backtest.Charts` above instead of having the intermediate `updatedCharts`?

Api/Api.cs
```diff
@@ -384,9 +384,15 @@ public Backtest CreateBacktest(int projectId, string compileId, string backtestN
...
+            BacktestResponseWrapper result;
             ApiConnection.TryRequest(request, out result);
-            return result;
+
+            // Use API Response values for Backtest Values
+            result.Backtest.Success = result.Success;
+            result.Backtest.Errors = result.Errors;
+
+            // Return only the backtest object
+            return result.Backtest;
```
Martin-Molinero, line 395: Do we **need** this wrapper? Looks fishy, incompatibility between Lean and cloud api

Common/Api/Backtest.cs
```diff
@@ -78,6 +72,44 @@ public class Backtest : RestResponse
...
+        /// <summary>
+        /// Runtime banner/updating statistics in the title banner of the live algorithm GUI.
+        /// </summary>
+        [JsonProperty(PropertyName = "runtimeStatistics", NullValueHandling = NullValueHandling.Ignore)]
+        public IDictionary<string, string> RuntimeStatistics;
+    }
+
+    /// <summary>
+    /// Wrapper class for Backtest/* endpoints JSON response
+    /// Currently used by Backtest/Read & Backtest/Create
```
Martin-Molinero, line 104: nit - Believe `&` broke the xml documentation, suggest using `and`?

Api/Api.cs
C-SELLERS: Changing the collection during enumeration threw a InvalidOperationException.r

Api/Api.cs
C-SELLERS: With the new response structure it is, we could remove the success and error members from the backtest class though and not assign them like it is now. I wanted to keep those members though for users to know what could've gone wrong in their api call, because I stuck with returning just the backtest obj instead of returning the wrapped obj in the new class that contains those details

C-SELLERS:
I implemented a custom de-serializer for the AlphaRuntimeStatistics, its rather simple since the bug was only coming from SortinoRatio. Let me know if there's anything else I need to change about it!

C-SELLERS:
I initially implemented a custom de-serializer for AlphaRuntimeStatistics, and it worked, but seemed to be too much for the issue and only addressed the SortinoRatio being a broken value and not any other decimals which "could" possibly break.

So, I thought I would scrap the initial custom de-serializer and create a small one specifically for decimals.

Now I am using the custom decimal converter I created which allows exponential notation for decimals via `decimal.Parse()` w/ `NumberStyles.float` as the style.

Common/AlphaRuntimeStatistics.cs
```diff
@@ -183,7 +184,7 @@ public decimal ReturnOverMaxDrawdown
         /// It is calculated by taking a portfolio's annualized rate of return and subtracting the risk free rate of return.
         /// </summary>
         /// <remarks>For performance we only truncate when the value is gotten</remarks>
-        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore), JsonConverter(typeof(StringDecimalJsonConverter), true)]
```
Martin-Molinero, line 187: Cool! 👍

Common/Api/Backtest.cs
```diff
@@ -78,6 +72,44 @@ public class Backtest : RestResponse
...
+        /// </summary>
+        /// <remarks>Intended for update mode -- send updates to the existing statistics in the result GUI. If statistic key does not exist in GUI, create it</remarks>
+        [JsonProperty(PropertyName = "statistics", NullValueHandling = NullValueHandling.Ignore)]
+        public IDictionary<string, string> Statistics;
+
+        /// <summary>
+        /// Runtime banner/updating statistics in the title banner of the live algorithm GUI.
+        /// </summary>
+        [JsonProperty(PropertyName = "runtimeStatistics", NullValueHandling = NullValueHandling.Ignore)]
+        public IDictionary<string, string> RuntimeStatistics;
```
Martin-Molinero, line 99: Just checking: there are some properties that the `BacktestResult` has that are now not present, should we add them? like for example `Orders/OrderEvents/ProfitLoss/RollingWindow/...`

C-SELLERS: Order/OrderEvents is not in the Backtest response so we don't have to worry about that one.

I went ahead and added RollingWindow and TotalPerformance so those are good to go.

ProfitLoss is having some issues because of a strange inconsistency in the web api. If there is no entries in the profitloss object the web api send a empty list, if there is, it sends a dictionary type (which is expected). Newtonsoft isn't happy about the empty list though when it expects a dictionary and fails to de-serialize.

PR #4997, opened 2020-12-04
Stop Emitting Insights on Delisted Securities

### LINKED ISSUE #4877: Do Not Emit Insights of Delisted Securities
#### Expected Behavior
Do not emit insights of delisted securities.

#### Actual Behavior
It's possible to emit insights of delisted securities

#### Potential Solution
N/A

#### Reproducing the Problem
```python
class ARQLInvalidInsightAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.SetCash(1000000)
        self.SetStartDate(2015, 1, 1)
        self.SetBrokerageModel(BrokerageName.AlphaStreams)
        self.AddEquity("SPY", Resolution.Daily)
        self.symbol = self.AddEquity("ARQL", Resolution.Daily).Symbol
        
    def OnData (self, Data):
        self.EmitInsights(Insight.Price(self.symbol, timedelta(1), InsightDirection.Up))
```
ARQL was delisted on 20200116 and this algorithm keeps emitting insights after that date.

#### Checklist

C-SELLERS:
#### Description

In the `QCAlgorithm` layer I added a check to verify that all insights are valid before allowing they to be emitted.
 
#### Related Issue

Closes #4877 

#### Motivation and Context

Before you could emit insights for delisted securities, now I check all the pending insights securities to see if they are still listed, if not I remove it from the insights list that will be processed and post an error in the log.

N/A

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Framework.cs
+++ b/Algorithm/QCAlgorithm.Framework.cs
@@ -431,7 +431,10 @@ namespace QuantConnect.Algorithm
                 }
                 else
                 {
-                    // If we already had an invalid insight this will have been initialized storing the valid ones.
+                    // Initialize the insight fields
+                    insights[i] = InitializeInsightFields(insights[i]);
+
+                    // If we already had an invalid insight, this will have been initialized storing the valid ones.
                     if (validInsights != null)
                     {
                         validInsights.Add(insights[i]);
--- a/Algorithm/QCAlgorithm.Framework.cs
+++ b/Algorithm/QCAlgorithm.Framework.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Algorithm.Framework.Alphas;
 using QuantConnect.Algorithm.Framework.Alphas.Analysis;
@@ -381,6 +382,20 @@ namespace QuantConnect.Algorithm
                 }
                 return;
             }
+
+            // Verify insights are valid
+            var validInsights = new List<Insight>();
+            foreach(var insight in insights){
+                if (Securities[insight.Symbol].IsDelisted)
+                {
+                    Error($"QCAlgorithm.EmitInsights(): Cannot emit insight for delisted security {insight.Symbol}");
+                } else {
+                    validInsights.Add(insight);
+                }
+            }
+
+            // Update our insights with our valids ones
+            insights = validInsights.ToArray();
 
             OnInsightsGenerated(insights.Select(InitializeInsightFields));
             ProcessInsights(insights);
--- a/Algorithm/QCAlgorithm.Framework.cs
+++ b/Algorithm/QCAlgorithm.Framework.cs
@@ -149,7 +149,8 @@ namespace QuantConnect.Algorithm
             // only fire insights generated event if we actually have insights
             if (insights.Length != 0)
             {
-                OnInsightsGenerated(insights.Select(InitializeInsightFields));
+                insights = InitializeInsights(insights);
+                OnInsightsGenerated(insights);
             }
 
             ProcessInsights(insights);
@@ -228,7 +229,7 @@ namespace QuantConnect.Algorithm
                     Log($"{Time}: RISK ADJUSTED TARGETS: {string.Join(" | ", riskAdjustedTargets.Select(t => t.ToString()).OrderBy(t => t))}");
                 }
             }
-            
+
             Execution.Execute(this, riskAdjustedTargets);
         }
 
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.Framework.cs
```diff
@@ -382,6 +383,22 @@ public void EmitInsights(params Insight[] insights)
                 return;
             }
 
+            // Verify insights are valid
+            var validInsights = new List<Insight>();
+            foreach(var insight in insights){
+                if (Securities[insight.Symbol].IsDelisted)
+                {
+                    Error($"QCAlgorithm.EmitInsights(): Cannot emit insight for delisted security {insight.Symbol}");
```
AlexCatarino, line 391: Calling `Error` will add the message to the user log and consume his quota, and it can also be a repetitive annoying message. Perhaps we should find a solution that only logs once for the first time it happens just to let the user know that the `Insight` was discarded.

C-SELLERS: I went ahead and changed the logic to post one warning, informing the user that they will be discarded

Algorithm/QCAlgorithm.Framework.cs
```diff
@@ -382,6 +384,26 @@ public void EmitInsights(params Insight[] insights)
...
             }
 
+            // Verify insights are valid
+            var validInsights = new List<Insight>();
+            foreach(var insight in insights){
+                if (Securities[insight.Symbol].IsDelisted)
+                {
+                    if (!_isEmitDelistedInsightWarningSent)
+                    {
+                        Error($"QCAlgorithm.EmitInsights(): Warning: cannot emit insights for delisted securities, these will be discarded");
```
Martin-Molinero, line 394: Cool! 👍 
We are missing to cover the insight emitted by the alpha too, see `public void OnFrameworkData(Slice slice)`. 
Suggest we create a method that performs this loop and `InitializeInsightFields` at the same time, returning an array, so we don't iterate twice over the insights. `OnInsightsGenerated` should also take an array of insights. And we call this new method in both locations.
We should add a regression test covering these two cases too

Algorithm/QCAlgorithm.Framework.cs
```diff
@@ -394,7 +398,35 @@ public void EmitInsights(params Insight[] insights)
...
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="insights"></param>
+        /// <returns></returns>
+        private Insight[] InitializeInsights(Insight[] insights)
+        {
+            // Verify insights are valid and initialize their fields
+            var validInsights = new List<Insight>();
+            foreach (var insight in insights)
```
Martin-Molinero, line 413: Cool!
This is very sensitive to performance since it's executed for each insight, suggest we do the following:
- using a simple `for(var i = 0;..)` loop, so that we do not create the enumerator using linq.
- avoid creating the list unless there is at least 1 delisted security which we should remove, so in 99.9% of the cases we do not create a new collection at all, we return provided `Insight[]`
- let's make `OnInsightsGenerated` take an `Insight[]` and avoid the `ToArray` call he has internally

nit - missing XML docs

Tests/Algorithm/Framework/QCAlgorithmFrameworkTests.cs
```diff
@@ -71,6 +71,45 @@ public void SetsInsightGeneratedAndCloseTimes()
...
+            security.IsDelisted = true;
+
+            // Trigger Alpha to emit insight
+            algorithm.OnFrameworkData(new Slice(new DateTime(2000, 01, 01), new List<BaseData>() { tick }));
+
+            // Manually emit insight
+            algorithm.EmitInsights(Insight.Price(Symbols.SPY, TimeSpan.FromDays(1), InsightDirection.Up, .5, .75));
+
+            // Should be zero because security is delisted
+            Assert.AreEqual(0, actualInsights.Count);
```
Martin-Molinero, line 110: Cool! Maybe we can make `security.IsDelisted = true;` the bool a argument for the test so we assert the count is 2 for when delisted is `false`

Martin-Molinero (commented):
Nice! almost there, some minor request

Tests/Algorithm/Framework/QCAlgorithmFrameworkTests.cs
C-SELLERS: Gotcha, made it two test cases one for delisted one for not

Algorithm/QCAlgorithm.Framework.cs
C-SELLERS: Gotcha!

AlexCatarino (approved):

Martin-Molinero (approved):
Sweet! 🍬

PR #5025, opened 2020-12-11
Noted Polygon downloaded in Readme

cnaccio:
Just added Polygon to the list of downloaded on the Readme as I almost missed it at first, until I looked at the source files.


jaredbroad (approved):
Nice thank you

PR #5028, opened 2020-12-12
Fixed QuantBook import path for CSharp example

cnaccio:
Now requires a relative path for loading QuantBook.csx


C-SELLERS (approved):
Good change for docker container! Just like `KitchenSinkCSharpQuantBookTemplate` . Tested to be sure, fits with docker changes to have notebooks in there own directory.

In fact we may even consider moving these example notebooks into a directory called Notebooks so a user coming to use the docker container doesn't even need to move the examples.

Martin-Molinero (approved):
Thanks!

Issue #4999, opened 2020-12-04
Feature Request: RVI Indicator 


shilewenuw:
#### Expected Behavior
RVI Indicator implemented
[Relevant Thread](https://www.quantconnect.com/forum/discussion/9953/does-quantconnect-have-a-relative-vigor-index-indicator/p1)

#### Actual Behavior
Not implemented

#### Potential Solution
https://www.investopedia.com/terms/r/relative_vigor_index.asp

#### Checklist

- [x ] I have completely filled out this template
- [x ] I have confirmed that this issue exists on the current `master` branch
- [x ] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x ] I have provided detailed steps to reproduce the issue

Martin-Molinero:
Closed by https://github.com/QuantConnect/Lean/pull/5041

PR #5041, opened 2020-12-16
Feature 4999 -- Adds RVI

aarjaneiro:
https://www.investopedia.com/terms/r/relative_vigor_index.asp
#### Description

Adds the Relative Vigor Index along with its signal as a sub-indicator

#### Related Issue

#4999 

#### Motivation and Context

Requested addition

No

Added test data from TradingView's implementation

#### Types of changes

- [x ] New feature (non-breaking change which adds functionality)

#### Checklist:

- [ x] I have added tests to cover my changes. 
- [ x] All new and existing tests passed.


Tests/Indicators/RelativeVigorIndexTests.cs
```diff
@@ -41,7 +41,7 @@ public void ComparesWithExternalDataRviSignal()
                 TestFileName,
                 "RVI_S",
                 (ind, expected) => Assert.AreEqual(
-                    ind.IsReady ? expected : 0,
```
aarjaneiro: I wasn't a huge fan about needing to tie the main indicator's IsReady to the signal (as one can still use the RVI without the signal for the 3 extra terms it needs to warm up). I also didn't exactly like my original implementation for dealing with the issue in this older version of the test. Instead, removing the extra data from spy_rvi was a safer equivalent to `ind.IsReady ? expected : 0`. 

One can still check signal readiness by `Signal.IsReady` but it will be independent of "ind.IsReady".

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// <summary>
+        /// An optional signal line which behaves like a slowed version of the RVI.
+        /// </summary>
+        public RviSignal Signal { get; }
+
+
+        /// <summary>
+        /// Gets the band of Closes for the RVI.
+        /// </summary>
+        private IndicatorBase<IndicatorDataPoint> CloseBand { get; }
```
C-SELLERS, line 96: nit - remove extra new lines and move class members to top of class and organize 😄

Tests/Indicators/RelativeVigorIndexTests.cs
```diff
@@ -0,0 +1,52 @@
...
+                "RVI_S",
+                (ind, expected) => Assert.AreEqual(
+                    expected,
+                    (double) ((RelativeVigorIndex) ind).Signal.Current.Value,
+                    delta: 0.06
+                ));
+        }
+    }
+   
+
```
C-SELLERS, line 51: nit - cleanup new lines here as well 👍🏽

C-SELLERS (changes requested):
Everything looks good! Just a couple nit changes requested.

Indicators/RelativeVigorIndex.cs
aarjaneiro: Sounds good -- will use the ordering of https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md from hereon!

C-SELLERS (approved):

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        /// <summary>
+        /// Gets the band of Ranges for the RVI.
+        /// </summary>
+        private IndicatorBase<IndicatorDataPoint> RangeBand { get; }
```
Martin-Molinero, line 65: nit - The Lean code style is placing privates at the top

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,210 @@
...
+                RangeBand.Update(input.Time, (e + 2 * (f + g) + h) / 6);
+            }
+
+            _previousInputs.Add(input);
+            if (CloseBand.IsReady && RangeBand.IsReady)
+            {
+                var rvi = CloseBand / RangeBand;
+                Signal?.Update(input.Time, rvi); // Checks for null before updating.
+                return rvi;
+            }
```
Martin-Molinero, line 137: Cool! 
Might want to add a check for `RangeBand` being `0` & avoid cosmic explosion

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// <param name="type">The type of Moving Average to use</param>
+        public RelativeVigorIndex(
+            string name, string signalName, int period, MovingAverageType type = MovingAverageType.Simple)
+            : base(name)
+        {
+            WarmUpPeriod = period + 3;
+            CloseBand = type.AsIndicator("_closingBand", period);
+            RangeBand = type.AsIndicator("_rangeBand", period);
+            _previousInputs = new RollingWindow<IBaseDataBar>(3);
+            Signal = new RviSignal(signalName);
```
Martin-Molinero, line 107: Wondering if we should directly no make it optional and have it there always, simplify the logic, not sure it's worth not having it (null checks, duplicated constructors, etc). Maybe can take the indicator name and directly add `_S` to it

Indicators/RelativeVigorIndex.cs
```diff
@@ -0,0 +1,210 @@
...
+            CloseBand.Reset();
+            RangeBand.Reset();
+            _previousInputs.Reset();
+            Signal?.Reset();
+        }
+
+        /// <summary>
+        /// The signal for the Relative Vigor Index, itself an indicator. 
+        /// </summary>
+        public class RviSignal : Indicator, IIndicatorWarmUpPeriodProvider
```
Martin-Molinero, line 157: Cool! Let's please move this guy to his own file with his own simple test collection too 👍

Martin-Molinero (commented):
Nice one! leaving minor suggestions

Indicators/RelativeVigorIndex.cs
jaredbroad: `RelativeVigorIndexSignal` & agree should be there all the time like MACD.

aarjaneiro: Will do -- I reckon it does not add much in terms of workload, anyways

Indicators/RelativeVigorIndex.cs
aarjaneiro: I ended up moving it to its own file but in a manner that it should only be accessible through the (main) RVI class. As such, I added `TestDivByZero` for it (to ensure it can properly avoid dividing by zero) in the test file for RVI.

Martin-Molinero (approved):
🚀

PR #5039, opened 2020-12-16
Reduce Travis Log

C-SELLERS:
#### Description

- Reduce verbosity of Nuget, and MsBuild commands
- Add ConsoleErrorLogHandler that only logs error statements
- Modify Test.AlgorithmRunner to use ConsoleErrorLogHandler
- Set Test SetupFixture (test assembly) to use ConsoleErrorLogHandler
- Replace all Console.WriteLine() statements in Tests project with Log.Trace

#### Related Issue

Closes #5033 

#### Motivation and Context

Travis logs are too long causing Travis to fail and exit out during test builds.

In its current form this should be significantly less long than before, so if we want to return some of the changes back (such as AlgorithmRunner logging) we can.

#### Requires Documentation Change

N/A

#### How Has This Been Tested?

Successful building and unit tests.

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #5049, opened 2020-12-17
Reduce Travis Logs v2

C-SELLERS:
#### Description

Continuation of #5039 

Additions:
- Reapply silencing of publishing stubs #4933
- Set test assembly method that will maintain loghandler after every test suite

Need to address:
- Change stub generator repo logging level to error only. Maybe a parameter we can pass in like `NO_DEBUG` 

#### Motivation and Context

This continues the changes made in #5039, there were a lot in there that greatly reduced the log size but I was running into a couple issues that I hope this PR will address and limit it even further.

N/A

All tests have been ran and are successful

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/AssemblyInitialize.cs
+++ b/Tests/AssemblyInitialize.cs
@@ -17,12 +17,14 @@
 using System;
 using System.IO;
 using NUnit.Framework;
+using NUnit.Framework.Interfaces;
 using QuantConnect;
 using QuantConnect.Configuration;
 using QuantConnect.Logging;
 using QuantConnect.Python;
 using QuantConnect.Util;
 
+[assembly: MaintainLogHandler()]
 [SetUpFixture]
 public class AssemblyInitialize
 {
@@ -30,18 +32,6 @@ public class AssemblyInitialize
     public void InitializeTestEnvironment()
     {
         AdjustCurrentDirectory();
-        
-        if (TestContext.Parameters.Exists("log-handler"))
-        {
-            var logHandler = TestContext.Parameters["log-handler"];
-            Log.Trace($"QuantConnect.Tests.AssemblyInitialize(): Log handler test parameter loaded {logHandler}");
-
-            Log.LogHandler = Composer.Instance.GetExportedValueByTypeName<ILogHandler>(logHandler);
-        }
-        else
-        {
-            Log.LogHandler = new ConsoleLogHandler();
-        }
     }
 
     public static void AdjustCurrentDirectory()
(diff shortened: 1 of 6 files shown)


C-SELLERS:
Travis Log after the last commit is only 7812 Lines, if this is merged we will see the buildstubs script length as well which will hopefully only make it a little longer because I quieted the dotnet call that build the stubs.

jaredbroad:
I'd prefer we "set properly" the classes who change the log handler, rather than lock it from being changed which could introduce bugs later when the logs don't change as expected.

C-SELLERS:
Gotcha, I thought that may be the case. I will have to find a way to work around the classes that do change it so we can maintain one specific loggingHandler for all the tests that don't need something specific.

ci_build_stubs.sh
```diff
@@ -55,7 +55,7 @@ function generate_stubs {
 
     NO_DEBUG="true" \
     STUBS_VERSION="$TRAVIS_TAG" \
-    dotnet run $LEAN_DIR $RUNTIME_DIR $STUBS_DIR
+    dotnet run -v q $LEAN_DIR $RUNTIME_DIR $STUBS_DIR
```
jmerle, line 58: I've just ran this locally and I don't think it does what you hope it will do (it won't get rid of the ~300 "Generating ..." messages). I think the only way to silence those messages is to set the stubs generator's minimum logging level to `Error` if the `NO_DEBUG` environment variable is set to `true`. This can be done by changing `Level.Info` to `Level.Error` in https://github.com/QuantConnect/quantconnect-stubs-generator/blob/master/QuantConnectStubsGenerator/Program.cs#L29.

jmerle:
You can reduce the Travis logs on tag builds by an additional ~2,000 lines if you revert the changes made in #4933. These changes were only needed temporarily to debug what turned out to be an authentication issue, but were never reverted after that issue was fixed.

C-SELLERS:
Thank you Jasper!!

Tests/AssemblyInitialize.cs
```diff
@@ -70,3 +60,43 @@ public static void AdjustCurrentDirectory()
             });
     }
 }
+
+[AttributeUsage(AttributeTargets.Assembly)]
+public class MaintainLogHandlerAttribute : Attribute, ITestAction
```
Martin-Molinero, line 65: Cool! 👍

Martin-Molinero (approved):

Martin-Molinero (commented):
Mind updating the `AlgorithmRunner.RunLocalBacktest()` so that locally regression tests log output to console too as before, this change https://github.com/QuantConnect/Lean/pull/5039/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.sh&file-filters%5B%5D=.yml#diff-9c19e625bd0c67dd45cf92ae2fcce0e1c00858b6e136012955afe2aaec5069adR94 @C-SELLERS

C-SELLERS:
Ouch switching AlgorithmRunner back to ConsoleLogHandler doubled the travis log to 28k 😬

Issue #2853, opened 2019-01-23
RenkoBar Consolidator not functioning as expected


simonsonjack:
#### Expected Behavior
Consolidator will build bars at appropriate intervals according to the argument passed to RenkoConsolidator(10) -- in this case, $10.

#### Actual Behavior
Despite a number of occasions where there is a $10 difference, the consolidator fails to build bars until late in the time series -- that is, there are instances where the stock price fluctuates around $121 when it's starting price was around $111.

#### Potential Solution
Examine and update the behavior of the Renko Consolidator.

#### Reproducing the Problem
This code will reproduce the problem and show a RenkoBar plot overlaid with an asset price plot.
```
def Initialize(self):

        self.SetStartDate(2012,1, 1)  #Set Start Date
        self.SetEndDate(2013,1,1)    #Set End Date
        self.SetCash(100000)           #Set Strategy Cash

        self.AddEquity("SPY", Resolution.Hour)
        renko_bar = RenkoConsolidator(10)
        renko_bar.DataConsolidated += self.OnDataConsolidated
        self.SubscriptionManager.AddConsolidator('SPY', renko_bar)
        
        renko_plot = Chart('Renko')
        renko_plot.AddSeries(Series('Renko Bar', SeriesType.Bar, 4))
        renko_plot.AddSeries(Series('Stock', SeriesType.Bar, 0))        
    
    def OnData(self, data):
        
        if data.Dividends.ContainsKey('SPY'):
            return
        self.Plot('Renko', 'Stock', data['SPY'].Close)
        self.Log(str(self.Time) + '   SPY Ceiling: $' + str(np.ceil(data['SPY'].Close)))
        self.Log(str(self.Time) + '   SPY Floor: $' + str(np.floor(data['SPY'].Close)))
            
    def OnDataConsolidated(self, sender ,bar):
        self.Plot('Renko', 'Renko Bar', bar.Close)
        self.Log(str(self.Time) + '   Renko Bar: $' + str(bar.Close))
```

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

adriantorrie:
@simonsonjack Is this still an issue, or has it been inherently fixed as part of #3764?

amijinxed:
Still a problem.

PR #5052, opened 2020-12-17
4553 - Refactor RenkoConsolidator

### LINKED ISSUE #4553: RenkoConsolidator Division By Zero

aarjaneiro:
Refactored:
```
                         ↗ RenkoConsolidator
 RenkoConsolidator
                         ↘ WickedRenkoConsolidator
```
#### Description

Created a base class from which both inherit. 

#### Related Issue

Closes #4553 
#### Motivation and Context

Calling RenkoType.Wicked separated initialization and update conditions, better to have separate classes altogether.

No

Tests as implemented for the Renko Consolidator
#### Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [x ] Refactor (non-breaking change which improves implementation)

#### Checklist:

- [x ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x ] I have added tests to cover my changes. 


Common/Data/Consolidators/BaseRenkoConsolidator.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// </summary>
+        public Type OutputType => typeof(RenkoBar);
+
+
+        /// <summary>
+        /// Updates this consolidator with the specified data
+        /// </summary>
+        /// <param name="data">The new data for the consolidator</param>
+        public abstract void Update(IBaseData data);
+
```
C-SELLERS, line 171: nit - extra new lines

Common/Data/Consolidators/BaseRenkoConsolidator.cs
```diff
@@ -0,0 +1,228 @@
...
+using System;
+using Python.Runtime;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Data.Consolidators
+{
+    /// <summary>
+    /// The base class for the RenkoConsolidator class between <see cref="RenkoType"/>s.
+    /// </summary>
+    public abstract class BaseRenkoConsolidator : IDataConsolidator
```
C-SELLERS, line 25: Nice!

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -122,264 +64,51 @@ public RenkoConsolidator(decimal barSize, Func<IBaseData, decimal> selector, Fun
...
-                        _highRate = limit;
-
-                        Falling(data);
-                    }
-                }
-            }
-        }
-
-        private void UpdateClassic(IBaseData data)
+        public override void Update(IBaseData data)
```
C-SELLERS, line 81: 👍🏽

Common/Data/Consolidators/BaseRenkoConsolidator.cs
```diff
@@ -0,0 +1,228 @@
...
+            else
+            {
+                VolumeSelector = (x => 0);
+            }
+        }
+
+        /// <summary>
+        /// Gets the kind of the bar
+        /// </summary>
+        public RenkoType Type { get; }
```
C-SELLERS, line 136: nit - members above the constructor ordered by accessibility

C-SELLERS (changes requested):
Looks good to me! Just needs a once over on the nit details.

Common/Data/Consolidators/BaseRenkoConsolidator.cs
aarjaneiro: Ah, it seems I was following the wrong convention! I will reorder as in https://stackoverflow.com/questions/150479/order-of-items-in-classes-fields-properties-constructors-methods

C-SELLERS (approved):

Common/Data/Consolidators/BaseRenkoConsolidator.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RenkoConsolidator" /> class.
+        /// </summary>
+        /// <param name="barSize">The size of each bar in units of the value produced by <paramref name="selector"/></param>
+        /// <param name="selector">Extracts the value from a data instance to be formed into a <see cref="RenkoBar"/>. The default
+        /// value is (x => x.Value) the <see cref="IBaseData.Value"/> property on <see cref="IBaseData"/></param>
+        /// <param name="volumeSelector">Extracts the volume from a data instance. The default value is null which does
+        /// not aggregate volume per bar.</param>
+        /// <param name="evenBars">When true bar open/close will be a multiple of the barSize</param>
+        public BaseRenkoConsolidator(decimal barSize, PyObject selector, PyObject volumeSelector = null, bool evenBars = true)
```
Martin-Molinero, line 89: nit - `public` -> `protected`

Martin-Molinero (commented):
Thanks for the PR!
Leaving a couple of requests, basically let's abstract the base Renko consolidator a bit more, in such a way he only holds shared logic/properties/methods and each implementation owns its particular details/behavior

Martin-Molinero (changes requested):
See previous comments 👍

Martin-Molinero:
Believe stats have changed for the `WarmupIndicatorRegressionAlgorithm` & `RenkoConsolidatorAlgorithm` with this branch? Should review _why_ and revert the change causing it, if appropriate, wouldn't expect algorithm stats changed with this PR

Common/Data/Consolidators/BaseRenkoConsolidator.cs
```diff
@@ -0,0 +1,211 @@
...
+{
+    /// <summary>
+    /// The base class for the RenkoConsolidator class between <see cref="RenkoType"/>s.
+    /// </summary>
+    public abstract class BaseRenkoConsolidator : IDataConsolidator
+    {
+        internal readonly Func<IBaseData, decimal> Selector;
+        internal readonly Func<IBaseData, decimal> VolumeSelector;
+        internal readonly bool EvenBars;
+        internal RenkoBar CurrentBar;
```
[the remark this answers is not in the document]
aarjaneiro: Good catch on the Selector! I will change the access modifiers and keep to the styling as well.

aarjaneiro:
> See previous comments +1

Ended up removing the base class as little really ends up in common after refactoring (really only Renko type as a property and the epsilon checking (void) function could have been generalized in a base class).

Algorithm.CSharp/WarmupIndicatorRegressionAlgorithm.cs
```diff
@@ -17,6 +17,7 @@
 using System.Collections.Generic;
 using QuantConnect.Data;
 using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
```
Martin-Molinero, line 20: nit - unrequired new import

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -81,12 +66,11 @@ public RenkoConsolidator(decimal barSize, RenkoType type)
...
         public RenkoConsolidator(decimal barSize, bool evenBars = true)
         {
+            EpsilonCheck(barSize);
             _barSize = barSize;
             _selector = x => x.Value;
-            _volumeSelector = x => 0;
             _evenBars = evenBars;
-
-            Type = RenkoType.Classic;
+            _volumeSelector = x => 0;
```
Martin-Molinero, line 73: Believe this constructor could call the one bellow `: this(barSize, x => x.Value, x => 0, evenBars)`

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -122,205 +105,63 @@ public RenkoConsolidator(decimal barSize, Func<IBaseData, decimal> selector, Fun
...
-                OnDataConsolidated(wicko);
-
-                _openOn = _closeOn;
-                _openRate = limit;
-                _lowRate = limit;
-            }
-        }
-
-        private void Falling(IBaseData data)
+        event DataConsolidatedHandler IDataConsolidator.DataConsolidated
```
Martin-Molinero, line 154: nit - would be nice to move `IDataConsolidator.DataConsolidated` & `event EventHandler<RenkoBar> DataConsolidated` above the constructors

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -360,27 +202,36 @@ public void Scan(DateTime currentLocalTime)
...
+        {
+            DataConsolidated = null;
+            _dataConsolidatedHandler = null;
+        }
+
+        private static void EpsilonCheck(decimal barSize)
+        {
+            if (barSize < Extensions.GetDecimalEpsilon())
+            {
+                throw new ArgumentOutOfRangeException(nameof(barSize),
```
Martin-Molinero, line 217: Cool!

Common/Data/Consolidators/WickedRenkoConsolidator.cs
```diff
@@ -0,0 +1,277 @@
...
+        private DateTime _openOn;
+        private decimal _openRate;
+        private decimal _barSize;
+        private DataConsolidatedHandler _dataConsolidatedHandler;
+        private RenkoBar _currentBar;
+        private IBaseData _consolidated;
+
+        // Used for unit tests
+        internal RenkoBar OpenRenkoBar =>
+            new RenkoBar(null, _openOn, _closeOn, _barSize, _openRate, _highRate, _lowRate, _closeRate);
```
Martin-Molinero, line 42: Could we directly remove `OpenRenkoBar` and use `WorkingData` instead?

Common/Data/Consolidators/WickedRenkoConsolidator.cs
```diff
@@ -0,0 +1,277 @@
...
+
+        /// <summary>
+        /// Event handler that fires when a new piece of data is produced
+        /// </summary>
+        public event EventHandler<RenkoBar> DataConsolidated;
+
+        /// <summary>
+        /// Event handler that fires when a new piece of data is produced
+        /// </summary>
+        event DataConsolidatedHandler IDataConsolidator.DataConsolidated
```
Martin-Molinero, line 92: nit - The same comment as the classic renko, let's move these last 2 guys above the constructor

Tests/Common/Data/WickedRenkoConsolidatorTests.cs
```diff
@@ -0,0 +1,586 @@
...
+using QuantConnect.Indicators;
+using System.Collections.Generic;
+
+namespace QuantConnect.Tests.Common.Data
+{
+    [TestFixture]
+    public class WickedRenkoConsolidatorTests
+    {
+        [Test]
+        public void WickedOutputTypeIsRenkoBar()
```
Martin-Molinero, line 29: Nice! 💯

Tests/Common/Data/RenkoConsolidatorTests.cs
```diff
@@ -18,11 +18,9 @@
 using QuantConnect.Data.Consolidators;
 using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
-using System.Collections.Generic;
 
 namespace QuantConnect.Tests.Common.Data
 {
-    [TestFixture]
```
Martin-Molinero, line 25: Wondering why remove `[TestFixture]` here?

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -24,53 +24,38 @@ namespace QuantConnect.Data.Consolidators
...
+        /// </summary>
+        public Type InputType => typeof(IBaseData);
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="RenkoConsolidator"/> class using the specified <paramref name="barSize"/>.
+        /// Gets <see cref="RenkoBar"/> which is the type emitted in the <see cref="IDataConsolidator.DataConsolidated"/> event.
         /// </summary>
-        /// <param name="barSize">The constant value size of each bar</param>
-        /// <param name="type">The RenkoType of the bar</param>
-        public RenkoConsolidator(decimal barSize, RenkoType type)
```
Martin-Molinero, line 64: For a happier user experience suggest we keep this constructor, flag it as obsolete, for example: `[Obsolete("Please use WickedRenkoConsolidator type")]` and make it thrown an exception with a similar message

Martin-Molinero (changes requested):
Sweet! Almost there, leaving a few minor comments & the regression tests `WarmupIndicatorRegressionAlgorithm` & `RenkoConsolidatorAlgorithm (C#/Py)` are still failing in this branch, should review why & address/fix

Martin-Molinero:
If you're up to it, it would be great if after this PR you took a stab at this renko related bug https://github.com/QuantConnect/Lean/issues/2853

Common/Data/Consolidators/WickedRenkoConsolidator.cs
aarjaneiro: This is interesting. From what I see, this is a means of bypassing the fact that for the end user, assignment of `_currentBar` occurs upon `OnDataConsolidated`. For a property like `WorkingData`, I reckon its best to change each of `_openOn, _closeOn, _barSize, _openRate, _highRate, _lowRate, _closeRate` to `internal` than to determine what is returned based on whether or not the caller is a test. Doing this and then adding a method in WickedConsolidatorTests which returns the same value as `OpenRenkoBar` can allow removal of `OpenRenkoBar`.

Common/Data/Consolidators/RenkoConsolidator.cs
aarjaneiro: I tried that, though it seems to cause a cyclic constructor call

aarjaneiro:
> Sweet! Almost there, leaving a few minor comments & the regression tests `WarmupIndicatorRegressionAlgorithm` & `RenkoConsolidatorAlgorithm (C#/Py)` are still failing in this branch, should review why & address/fix

My next commit will focus on this now that the main "structure" provided by this refactor is essentially there.

Algorithm.CSharp/RenkoConsolidatorAlgorithm.cs
```diff
@@ -25,6 +25,7 @@ namespace QuantConnect.Algorithm.CSharp
     /// </summary>
     /// <meta name="tag" content="renko" />
     /// <meta name="tag" content="indicators" />
+    /// <meta name="tag" content="wicked"/>
```
Martin-Molinero, line 28: nit - unrequired change?

Common/Data/Consolidators/RenkoConsolidator.cs
Martin-Molinero: Please try `: this(barSize, x => x.Value, x => 0, evenBars)` should work, the other constructor isn't calling no other 👍

Common/Data/Consolidators/RenkoConsolidator.cs
```diff
@@ -98,19 +96,34 @@ public RenkoConsolidator(decimal barSize, bool evenBars = true)
...
+        /// <param name="type"></param>
+        [Obsolete("Please use the WickedRenkoConsolidator if RenkoType is not Classic")]
+        public RenkoConsolidator(decimal barSize, RenkoType type)
+        {
+            if (type != RenkoType.Classic)
+            {
+                throw new ArgumentException("Please use the WickedRenkoConsolidator type if RenkoType is not Classic");
+            }
+
+            _barSize = barSize;
```
Martin-Molinero, line 126: nit - similar let's call `: this(barSize, true)` instead

Common/Data/Consolidators/WickedRenkoConsolidator.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// </summary>
+    public class WickedRenkoConsolidator : IDataConsolidator
+    {
+        internal DateTime CloseOn;
+        internal decimal CloseRate;
+        internal decimal HighRate;
+        internal decimal LowRate;
+        internal DateTime OpenOn;
+        internal decimal OpenRate;
+        internal decimal BarSize;
```
Martin-Molinero, line 33: Let's please make these protected and create a test class `WickedRenkoConsolidatorTest` exposing them so they can be asserted on 👍

aarjaneiro: Alternatively, changing them to protected internals will expose them without needing to re-add something of the form of the removed `OpenRenkoBar`. Would that be preferable?

Martin-Molinero (approved):

Algorithm.CSharp/RenkoConsolidatorAlgorithm.cs
jaredbroad: This helps our documentation search for the types

PR #5065, opened 2020-12-21
Bug jsonrounding maxvalue minvalue

BadgerTheAdger:
Fixed bug - decimal.MinValue and MaxValue cannot be deserialized

#### Description
After serializing decimal.MinValue or MaxValue, Newtonsoft.Json can't deserialize it by default - it thinks it's too large.

#### Motivation and Context
I have encountered a bug due to which I lost 6 hours of time today! I don't want somebody else to get the same bug. I was getting very high CompoundingAnnualReturn which turned out to be greater than decimal.MaxValue and so decimal.MaxValue was serialized in a BacktestResultPacket. But it could not be deserialized because Newtonsoft.Json does not know how to handle large decimal values. So I fixed this by serializing huge decimals as strings.

- by launching the project manually and retrieving backtestpacket (now thankfully without errors)
- with the unit test I added to the PR

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Util/JsonRoundingConverter.cs
+++ b/Common/Util/JsonRoundingConverter.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using ExchangeSharp;
 using Newtonsoft.Json;
 
 namespace QuantConnect.Util
@@ -33,7 +34,7 @@ namespace QuantConnect.Util
         /// Will always return false.
         /// Gets a value indicating whether this <see cref="T:Newtonsoft.Json.JsonConverter" /> can read JSON.
         /// </summary>
-        public override bool CanRead => false;
+        public override bool CanRead => true;
 
         /// <summary>
         /// Determines whether this instance can convert the specified object type.
@@ -55,7 +56,14 @@ namespace QuantConnect.Util
         /// <param name="serializer">The calling serializer.</param>
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
         {
-            throw new NotImplementedException();
+            if (objectType == typeof(double))
+            {
+                return Double.Parse(existingValue.ToStringInvariant(), System.Globalization.CultureInfo.InvariantCulture);
+            }
+            else
+            {
+                return Decimal.Parse(existingValue.ToStringInvariant(), System.Globalization.CultureInfo.InvariantCulture);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Martin-Molinero:
Hey @BadgerTheAdger! Thanks for the PR,
Wonder through which code path you were deserializing the `BacktestResultPacket`, like an API call, or just some custom code loading from disk?
There are a couple of tests failing, believe because the changes are now serializing a string instead of a number into the JSON, suggest we avoid these type of changes. Maybe there is some place where the `JsonRoundingConverter` is not being used and should be added? Because he is already writing 4 decimal places, so it should be able to deserialize?

BadgerTheAdger:
Hey @Martin-Molinero !

> Wonder through which code path you were deserializing the BacktestResultPacket

I was deserializing it this way:
```csharp
var json = File.ReadAllText(backtestFilePath);
var backtestPacket = Newtonsoft.Json.JsonConvert.DeserializeObject<BacktestResultPacket>(json);
```

So it was saved in the file system (I have my own MessagingHandler implementation that serializes the packet and saves it into a file).

> There are a couple of tests failing, believe because the changes are now serializing a string instead of a number into the JSON, suggest we avoid these type of changes.

Well ok, the value is then deserialized as a string, so it should work. 

The issue in the tests is "Unknown GDAX symbol: BTCLTC". It's not related to this change at all. Also, I noted that another [PR](https://github.com/QuantConnect/Lean/pull/5075) had the same problem. So I believe this is a test issue rather than a PR issue.

I have my local tests with my own brokerage and they work just fine, and I was running backtests the whole day yesterday and all worked. Can you please take a look at the tests and fix them if needed? If it's indeed an issue with my code, I will take a look but I already did that and it's a very simple change, it shouldn't cause any errors and I have done testing on my side many times already.

> Because he is already writing 4 decimal places, so it should be able to deserialize

As I mentioned in the description of PR, the issue is not with decimal places, it's an issue with big numbers in general, such as decimal.MaxValue and decimal.MinValue. These constants don't have any decimal places but it doesn't work with them because the number is too big. I decided to deserialize too big numbers into strings.

Martin-Molinero:
Hey @BadgerTheAdger,
I've opened a new PR on top of this one at https://github.com/QuantConnect/Lean/pull/5084 fixing some tests and with some minor changes so I'll close this PR. Thanks for your contribution!

Issue #5098, opened 2020-12-29
Feature - Time series indicators


aarjaneiro:
#### Expected Behavior

Currently there exists no means of fitting time series models (e.g., https://en.wikipedia.org/wiki/Time_series#Models) to data other than by employing Python's statsmodels/arch packages. To add such an option for C# users and to speed up computation of such models for Python users, an implementation under Indicators would be ideal.

#### Potential Solution

1. Create an Indicator subclass for time series analysis/models and add specific implementations of it (e.g., ARMA, GARCH).
2. Use a package like Cronos.

#### Checklist

- [ x] I have completely filled out this template
- [ x] I have confirmed that this issue exists on the current `master` branch
- [ x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

PR #5099, opened 2020-12-29
Feature #5098 - Time Series Indicators 

aarjaneiro:
Adds TimeSeriesIndicator along with the base implementation of ArimaIndicator
#### Description

Adds the ability to create new subclasses by inheriting from TimeSeriesIndicator.

ArimaIndicator currently allows for fitting models of ARIMA, ARMA, AR, and MA types by means of "Two-step Regression" as in  http://mbhauser.com/informal-notes/two-step-arma-estimation.pdf.

Future extensions to ArimaIndicator in particular can include fitting by means of MLE or innovations. MLE in particular can follow easily from the fact that the Two Step method can be used to generate Var(ε)'s unbiased estimator (e.g., for `Sigma` in https://github.com/aarjaneiro/CronoSeries/blob/c870b4a325342502fba7d04d77100812f483edc0/TimeSeries/Models/ARMAModel.cs#L496)

#### Related Issue

#5098 
#### Motivation and Context

Currently there does not exist any trivial means for allowing users across the platform (C#, Python) access to fitting such models.

*Likely not*; components are documented though it might be worth including an example for the usage of Time Series indicators. In general, the usage of these indicators will differ from the typical usage of technical analysis indicators as Time Series indicators will likely be employed for the models they return (i.e., the fitted parameters and the residuals resulting from fitting such parameters). Current implementation will by default forecast one point ahead such that one can use ARIMA as they would other indicators should they so choose (perhaps a forecast length parameter could be implemented such that TimeSeriesIndicators can alternatively return an enumerable object).

Tests have been added - compares against Python's statsmodels ARIMA under their conditional sum of squares method. Note, however, that statsmodels (or any other major package) does not have a pure two-step fitting method, though this is similar (the main difference being that there is no MLE used in the two-step method). Thus the test delta is extended to encompass the maximum residual between the two (the average difference being ≈1.36). As a compromise, another test has been added to guarantee that prediction error for the next value of Close is less than or equal to those from the statsmodels-generated predictions. 

The code for the statsmodels predictions:
```python
import numpy as np
import pandas as pd
# import matplotlib.pyplot as plt

import statsmodels.api as sm
from statsmodels.tsa.arima_model import ARIMA

data = pd.read_csv("spy_rvi.csv")
close = data["close"]
newdata = data[["date","open","high","low","close","volume"]]
newdata.columns = ["Date","Open","High","Low","Close","Volume"]

## Rolling
dayahead = [ARIMA(close[(i-50):i], 
	order=(1, 1, 1)).fit(method="css-mle").forecast(1) for i in range(50,len(close))]

## Prepend
final = [0 for i in range(50)]
for i in dayahead:
    final.append(i[0][0])

## Optional: Plot rolling day-ahead forecasts (better to prepend up to 51 for this)
# plt.plot(final)
# plt.plot(close[:len(final)])
# plt.show()

newdata = newdata.iloc[:len(final),:]
newdata["ARIMA"] = final

newdata.to_csv("spy_arima.csv")
``` 
Alternatively, a direct test can and should be incorporated to compare directly against statsmodels' MLE method when MLE is itself added as a fit method.  
#### Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] Refactor (non-breaking change which improves implementation)
- [ x] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ x] New feature (non-breaking change which adds functionality)
- [ x] Breaking change (fix or feature that would cause existing functionality to change)
- [ x] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [ x] My code follows the code style of this project.
- [ x] I have added tests to cover my changes. 
- [ x] All new and existing tests passed.
- [ x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`


Indicators/TimeSeriesIndicator.cs
```diff
@@ -0,0 +1,158 @@
...
+            DifferenceSeries(int d, double[] series)
+        {
+            _diffHeads = new double[series.Length - d];
+            if (d == 0)
+            {
+                return null;
+            }
+
+            var localSeries = series;
+            for (var j = 1; j <= d; j++)
```
Martin-Molinero, line 73: Wondering if some of these methods couldn't be replaced by a call to `Math.` or `MathNet.Numerics.` for example, maybe it's already implemented?

Let's please add a couple of specific unit tests for each of these different methods 👍. Feels like the core implementation could be a static methods? Easier to test and reuse.

`var j = 1` starting from `0` would avoid the `-1` bellow?
nit - let's remove the new line after `protected double[]`, same bellow at `InverseDifferencedSeries`

Indicators/TimeSeriesIndicator.cs
```diff
@@ -0,0 +1,158 @@
...
+        public int WarmUpPeriod { get; }
+
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IndicatorDataPoint input)
+        {
+            return 0m;
```
Martin-Molinero, line 55: nit - could just be abstract?

Indicators/TimeSeriesIndicator.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        /// <summary>
+        /// Gets a flag indicating when this indicator is ready and fully initialized
+        /// </summary>
+        public override bool IsReady { get; }
+
+        /// <summary>
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public int WarmUpPeriod { get; }
```
Martin-Molinero, line 46: Should be able to directly remove `public override bool IsReady { get; }`
`WarmUpPeriod ` should be abstract so derivative can override. Using `new` (at `ArimaIndicator.WarmUpPeriod`) is usually a code smell (something is wrong 😄)
nit - let's move properties above constructor

Indicators/ArimaIndicator.cs
```diff
@@ -0,0 +1,307 @@
...
+        /// AR coefficient.
+        /// </summary>
+        private readonly int _p;
+
+        /// <summary>
+        /// MA Coefficient.
+        /// </summary>
+        private readonly int _q;
+
+        private readonly RollingWindow<double> _rollingData;
```
Martin-Molinero, line 54: Is there a reason to use `double` instead of `decimals`?

Indicators/ArimaIndicator.cs
```diff
@@ -0,0 +1,307 @@
...
+
+        private List<double> _residuals;
+
+        /// <summary>
+        /// A dictionary, indexed by "AR" and "MA", containing their respective, fitted parameters.
+        /// </summary>
+        public Dictionary<string, double[]> Parameters;
```
Martin-Molinero, line 61: Are these going to be consumed by a user of the indicator?
Should they directly be 3 properties of the indicator AR/MA & intercept?

Martin-Molinero (changes requested):
Nice work!

Indicators/ArimaIndicator.cs
aarjaneiro: I did so due to Math.Net preferring doubles and _rollingData not itself being user-accessible

Indicators/ArimaIndicator.cs
aarjaneiro: These should be accessible imo in case one wishes to implement the fitted ARIMA parameters for some other common purposes other than forecasting (e.g., starting parameters for fitting a SARIMA).

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -103,6 +103,30 @@ public AccumulationDistributionOscillator ADOSC(Symbol symbol, int fastPeriod, i
...
+        /// Creates a new ARIMA indicator.
+        /// </summary>
+        /// <param name="symbol">The symbol whose ARIMA indicator we want</param>
+        /// <param name="p">AR order</param>
+        /// <param name="d">Difference order</param>
+        /// <param name="q">MA order</param>
+        /// <param name="period">Size of the rolling series to fit onto</param>
+        /// <param name="resolution">The resolution</param>
+        /// <returns>The ARIMA indicator for the requested symbol over the specified period</returns>
+        public ArimaIndicator ARIMA(Symbol symbol, int p, int d, int q, int period, Resolution? resolution = null)
```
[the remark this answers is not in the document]
aarjaneiro: While other names can be used (Autoregressive order (p) Differencing order (d) Moving average order (q)), I initially chose the abbreviated forms for the args and constructor because "[ARIMA(p,d,q)x(P,D,Q)m](https://v8doc.sas.com/sashtml/ets/chap7/sect8.html)" (for the generalized SARIMA, a potential extension class) is usually well understood in time series analysis and thus I figured it could be faster specify/understand for the end user.

Indicators/TimeSeriesIndicator.cs
aarjaneiro: Unfortunately the implemented Math.Net does not offer much in terms of Time Series. Newer versions have implemented some useful functions however (e.g. https://github.com/mathnet/mathnet-numerics/blob/64aa01e33c10b5063057128739d60c34ddc373ea/src/Numerics/Statistics/Correlation.cs#L46)

Good idea w.r.t. implementing tests for these -- probably should have had a TimeSeriesIndicator test file!  

Lastly, the `var j  = 1` can be changed such that `j + 1 <= d`.

Algorithm.CSharp/TimeSeriesIndicatorExampleAlgorithm.cs
```diff
@@ -0,0 +1,65 @@
...
+using QuantConnect.Indicators;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm to test the behaviour of ARMA versus AR models at the same order of differencing.
+    /// In particular, an ARIMA(1,1,1) and ARIMA(1,1,0) are instantiated while orders are placed if their difference
+    /// is sufficiently large (which would be due to the inclusion of the MA(1) term).
+    /// </summary>
+    public class AutoRegressiveMovingAverageIndicatorExample : QCAlgorithm
```
Martin-Molinero, line 27: Cool! `AutoRegressiveMovingAverageIndicatorExample` doesn't match file `TimeSeriesIndicatorExampleAlgorithm.cs` we should rename the file name 👍.
Let's please add `IRegressionAlgorithmDefinition` and the python equivalent algorithm 👍

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -103,6 +103,30 @@ public AccumulationDistributionOscillator ADOSC(Symbol symbol, int fastPeriod, i
             return accumulationDistributionOscillator;
         }
 
+        /// <summary>
+        /// Creates a new ARIMA indicator.
+        /// </summary>
+        /// <param name="symbol">The symbol whose ARIMA indicator we want</param>
+        /// <param name="arOrder">AR order</param>
+        /// <param name="diffOrder">Difference order</param>
+        /// <param name="maOrder">MA order</param>
```
Martin-Molinero, line 112: > AR order | Difference order | MA order 

We should improve the documentation on what these are

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+    /// An Autoregressive Intergrated Moving Average (ARIMA) is a time series model which can be used to describe a set of data.
+    /// In particular,with Xₜ representing the series, the model assumes the data are of form
+    /// (after differencing <see cref="_diffOrder" /> times):
+    /// <para>
+    ///     Xₜ = c + εₜ + ΣᵢφᵢXₜ₋ᵢ +  Σᵢθᵢεₜ₋ᵢ
+    /// </para>
+    /// where the first sum has an upper limit of <see cref="_arOrder" /> and the second <see cref="_maOrder" />.
+    /// </summary>
+    public class AutoRegressiveIntegratedMovingAverage : TimeSeriesIndicator, IIndicatorWarmUpPeriodProvider
+
```
Martin-Molinero, line 35: nit - empty line

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// </summary>
+        public double Intercept;
+
+        /// <summary>
+        /// Differencing coefficient (d). Determines how many times the series should be differenced before fitting the
+        /// model.
+        /// </summary>
+        private readonly int _diffOrder;
+
+        private readonly bool _intercept;
```
Martin-Molinero, line 58: nit - privates should be at the top

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// </summary>
+        public override int WarmUpPeriod { get; }
+
+        /// <summary>
+        /// The variance of the residuals (Var(ε)) from the first step of <see cref="TwoStepFit" />.
+        /// </summary>
+        public double ArResidualError { get; private set; }
+
+        /// <summary>
+        /// The variance of the residuals (Var(ε)) from the second step of <see cref="TwoStepFit" />.
```
Martin-Molinero, line 90: Like these docs! 💯

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <para>
+        ///     Xₜ = c + εₜ + ΣᵢφᵢXₜ₋ᵢ +  Σᵢθᵢεₜ₋ᵢ
+        /// </para>
+        /// where the first sum has an upper limit of <see cref="_arOrder" /> and the second <see cref="_maOrder" />.
+        /// This particular constructor fits the model by means of <see cref="TwoStepFit" /> for a specified name.
+        /// </summary>
+        /// <param name="name">The name of the indicator</param>
+        /// <param name="arOrder">AR order -- p</param>
+        /// <param name="diffOrder">Difference order -- d</param>
+        /// <param name="maOrder">MA order -- q</param>
```
Martin-Molinero, line 105: Let's expand a bit what MA/AR and difference order are

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <param name="period">Size of the rolling series to fit onto</param>
+        public AutoRegressiveIntegratedMovingAverage(
+            int arOrder,
+            int diffOrder,
+            int maOrder,
+            int period,
+            bool intercept
```
Martin-Molinero, line 159: `bool intercept` isn't used, missing docs

Indicators/TimeSeriesIndicator.cs
Martin-Molinero: hm this static method isn't being used and it's duplicated bellow?

Tests/Indicators/TimeSeriesIndicatorTests.cs
```diff
@@ -0,0 +1,71 @@
...
+                    default:
+                        test[i] = 2;
+                        break;
+                }
+            }
+
+            double[] heads;
+            var differencer = TimeSeriesIndicator.DifferenceSeries(1, test, out heads);
+            Assert.AreEqual(-1, differencer.Sum());
+            Assert.AreEqual(test.Sum(), TimeSeriesIndicator.InverseDifferencedSeries(differencer, heads).Sum());
```
Martin-Molinero, line 46: Cool! `InverseDifferencedSeries` should have it's own unit test 👍, each unit test should assert a single component.
Think could move the following into a static private helper method in this test class:
```
var test = new double[100];
for (var i = 0; i < test.Length; i++)
{
    switch (i % 2)
    {
        case 0:
            test[i] = 0;
            break;
        default:
            test[i] = 1;
            break;
    }
}
```

Indicators/TimeSeriesIndicator.cs
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        /// <summary>
+        /// Returns a series where each spot is taken by the cumulative sum of all points up to and including
+        /// the value at that spot in the original series.
+        /// </summary>
+        /// <param name="series">Series to cumulatively sum over.</param>
+        /// <param name="reverse">Whether to reverse the series before applying the cumulative sum.</param>
+        /// <returns>Cumulatively summed series.</returns>
+        protected static List<double> CumulativeSum(List<double> series, bool reverse = false)
```
Martin-Molinero, line 126: `CumulativeSum` is missing some unit test 👍

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+                if (_arOrder > 0)
+                {
+                    for (var i = 0; i < _arOrder; i++) // AR Parameters
+                    {
+                        summants += ArParameters[i] * arrayData[i];
+                    }
+                }
+
+                if (_maOrder > 0)
+                    for (var i = 0; i < _maOrder; i++) // MA Parameters
```
Martin-Molinero, line 280: nit - missing brackets

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IndicatorDataPoint input)
+        {
+            _rollingData.Add((double) input.Value);
+            if (_rollingData.IsReady)
+            {
+                var arrayData = _rollingData.ToArray();
+                TwoStepFit(arrayData);
+                arrayData = _diffOrder > 0 ? DifferenceSeries(_diffOrder, arrayData) : arrayData;
```
Martin-Molinero, line 269: `arrayData = _diffOrder > 0 ? DifferenceSeries(_diffOrder, arrayData) : arrayData;` is also done by `TwoStepFit` internally, could we do it once and pass it in?

Indicators/AutoRegressiveIntegratedMovingAverage.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// Fits the model by means of implementing the following pseudo-code algorithm (in the form of "if{then}"):
+        /// <code>
+        /// if diffOrder > 0 {Difference data diffOrder times}
+        /// if arOrder > 0 {Fit the AR model Xₜ = ΣᵢφᵢXₜ; ε's are set to residuals from fitting this.}
+        /// if maOrder > 0 {Fit the MA parameters left over  Xₜ = c + εₜ + ΣᵢφᵢXₜ₋ᵢ +  Σᵢθᵢεₜ₋ᵢ}
+        /// Return: φ and θ estimates.
+        /// </code>
+        /// http://mbhauser.com/informal-notes/two-step-arma-estimation.pdf
+        /// </summary>
+        protected void TwoStepFit(double[] series) // Protected for any future inheritors (e.g., SARIMA)
```
Martin-Molinero, line 185: Suggest trying to separate this method into smaller logical parts it's quite complex/hard to follow. For example something like `UpdateArResidualError()`, you will know best for sure 👍

Tests/Indicators/TimeSeriesIndicatorTests.cs
```diff
@@ -0,0 +1,71 @@
...
+                    case 0:
+                        test[i] = 0;
+                        break;
+                    default:
+                        test[i] = 1;
+                        break;
+                }
+            }
+
+            Console.WriteLine(test.Sum());
```
Martin-Molinero, line 66: Suggest replace `Console.WriteLine` with an assert 👍. Shouldn't call `Console` directly, but `Log.` which is controlled by the configuration of the test run

Martin-Molinero (changes requested):
Thanks for the updates! Getting close, leaving some requests 👍

Indicators/TimeSeriesIndicator.cs
aarjaneiro: I gave it a static method and a non-static implementation such that one can utilize it outside of the context of indicators (and thus, for example, allowing us to use it in a test where we aren't concerned with initializing an ARIMA).
e.g., 
```
  double[] heads;
            var differencer = TimeSeriesIndicator.DifferenceSeries(1, test, out heads);        
```

Tests/Indicators/AutoregressiveIntegratedMovingAverageTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                var arimas = testValues[i - 50];
+                ssIndicator += Math.Pow((double) (arimas[0] - test), 2);
+                ssTest += Math.Pow((double) (arimas[1] - test), 2);
+                betweenMethods.Add(Math.Abs(arimas[0] - arimas[1]));
+            }
+
+            Console.Write($"Prediction error for AutoRegressiveIntegratedMovingAverage: {ssIndicator}\n");
+            Console.Write($"Prediction error for External: {ssTest}\n");
+            Console.Write(
+                $"Mean difference between methods: {betweenMethods.Average()}, Std: {betweenMethods.ToDoubleArray().StandardDeviation()} ");
```
Martin-Molinero, line 97: Let's replace `Console.Write` for asserts

Martin-Molinero (approved):

PR #5103, opened 2020-12-30
Refactor 4065 expose buying power model initial and maintenance margin

mchandschuh:
We need to expose these functions in order to support the upcoming `IPositionGroupBuyingPowerModel`

#### Description

* Refactor HasSufficientBuyingPowerForOrder implementations

> Adds Sufficient and Insufficient helper methods to HashSufficientbuyingPowerForOrderParameters
enabling syntax like:

``` csharp
// we have sufficient buying power
return parameters.Sufficient()

// we don't have sufficient buying power
returnparameters.Insufficient("Reason we don't have enough")
```

* IBuyingPowerModel: Add margin functions Maintenance/Initial/ForOrder

> These were originally hidden in an effort to only expose what's necessary
for the engine to perform its work. Additionally, we encapsulated all of
the method arguments into parameters classes to prevent having to break
anyone in the future. Not including these foundational methods turns out to
be an oversight. These methods are not required by the engine, but rather by
other models. Another possible solution here is to add an additional abstraction
and include these methods on this new abstraction. BuyingPowerModel would then
explicitly implement these methods and models that depend on them would require
two code paths, one for when the buying power model implements this interface
and another for when it doesn't.

> Tests were additionally updated to remove test model implementations created for
the sole purpose of exposing these private methods.

* Add NullBuyingPowerModel

> Provides an implementation of IBuyingPowerModel that essentially
disables margin checks and ensures all orders are not failed due
to insufficient buying power. Technically, orders can still fail,
but it requires the entire portfolio to be underwater which would
be exceedingly hard to do if this model is cofigured onall secrites.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Python/BuyingPowerModelPythonWrapper.cs
+++ b/Common/Python/BuyingPowerModelPythonWrapper.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
  *
@@ -143,6 +143,43 @@ namespace QuantConnect.Python
             using (Py.GIL())
             {
                 _model.SetLeverage(security, leverage);
+            }
+        }
+
+        /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="security">The security to compute maintenance margin for</param>
+        /// <returns>The maintenance margin required for the </returns>
+        public decimal GetMaintenanceMargin(Security security)
+        {
+            using (Py.GIL())
+            {
+                return _model.GetMaintenanceMargin(security);
+            }
+        }
+
+        /// <summary>
+        /// The margin that must be held in order to increase the position by the provided quantity
+        /// </summary>
+        public decimal GetInitialMarginRequirement(Security security, decimal quantity)
+        {
+            using (Py.GIL())
+            {
+                return _model.GetInitialMarginRequirement(security, quantity);
+            }
+        }
+
+        /// <summary>
+        /// Gets the total margin required to execute the specified order in units of the account currency including fees
+        /// </summary>
+        /// <param name="parameters">An object containing the portfolio, the security and the order</param>
+        /// <returns>The total margin in terms of the currency quoted in the order</returns>
+        public decimal GetInitialMarginRequiredForOrder(InitialMarginRequiredForOrderParameters parameters)
+        {
+            using (Py.GIL())
+            {
+                return _model.GetInitialMarginRequiredForOrder(parameters);
             }
         }
     }
--- a/Common/Securities/BuyingPowerModel.cs
+++ b/Common/Securities/BuyingPowerModel.cs
@@ -236,14 +236,15 @@ namespace QuantConnect.Securities
             // short circuit the div 0 case
             if (parameters.Order.Quantity == 0)
             {
-                return new HasSufficientBuyingPowerForOrderResult(true);
+                return parameters.Sufficient();
             }
 
             var ticket = parameters.Portfolio.Transactions.GetOrderTicket(parameters.Order.Id);
             if (ticket == null)
             {
-                var reason = $"Null order ticket for id: {parameters.Order.Id}";
-                return new HasSufficientBuyingPowerForOrderResult(false, reason);
+                return parameters.Insufficient(
+                    $"Null order ticket for id: {parameters.Order.Id}"
+                );
             }
 
             if (parameters.Order.Type == OrderType.OptionExercise)
@@ -267,23 +268,24 @@ namespace QuantConnect.Securities
 
                     // we continue with this call for underlying
                     return underlying.BuyingPowerModel.HasSufficientBuyingPowerForOrder(
-                        new HasSufficientBuyingPowerForOrderParameters(parameters.Portfolio, underlying, newOrder));
+                        parameters.ForUnderlying(newOrder)
+                    );
                 }
 
-                return new HasSufficientBuyingPowerForOrderResult(true);
+                return parameters.Sufficient();
             }
 
             // When order only reduces or closes a security position, capital is always sufficient
             if (parameters.Security.Holdings.Quantity * parameters.Order.Quantity < 0 && Math.Abs(parameters.Security.Holdings.Quantity) >= Math.Abs(parameters.Order.Quantity))
             {
-                return new HasSufficientBuyingPowerForOrderResult(true);
+                return parameters.Sufficient();
             }
 
             var freeMargin = GetMarginRemaining(parameters.Portfolio, parameters.Security, parameters.Order.Direction);
             var initialMarginRequiredForOrder = GetInitialMarginRequiredForOrder(
-                new InitialMarginRequiredForOrderParameters(parameters.Portfolio.CashBook,
-                    parameters.Security,
-                    parameters.Order));
+                new InitialMarginRequiredForOrderParameters(
+                    parameters.Portfolio.CashBook, parameters.Security, parameters.Order
+            ));
 
             // pro-rate the initial margin required for order based on how much has already been filled
             var percentUnfilled = (Math.Abs(parameters.Order.Quantity) - Math.Abs(ticket.QuantityFilled)) / Math.Abs(parameters.Order.Quantity);
@@ -291,14 +293,13 @@ namespace QuantConnect.Securities
 
             if (Math.Abs(initialMarginRequiredForRemainderOfOrder) > freeMargin)
             {
-                var reason = Invariant($"Id: {parameters.Order.Id}, ") +
+                return parameters.Insufficient(Invariant($"Id: {parameters.Order.Id}, ") +
                     Invariant($"Initial Margin: {initialMarginRequiredForRemainderOfOrder.Normalize()}, ") +
-                    Invariant($"Free Margin: {freeMargin.Normalize()}");
-
-                return new HasSufficientBuyingPowerForOrderResult(false, reason);
+                    Invariant($"Free Margin: {freeMargin.Normalize()}")
+                );
             }
 
-            return new HasSufficientBuyingPowerForOrderResult(true);
+            return parameters.Sufficient();
         }
 
         /// <summary>
--- a/Common/Securities/BuyingPowerModel.cs
+++ b/Common/Securities/BuyingPowerModel.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
  *
@@ -134,7 +134,7 @@ namespace QuantConnect.Securities
         /// </summary>
         /// <param name="parameters">An object containing the portfolio, the security and the order</param>
         /// <returns>The total margin in terms of the currency quoted in the order</returns>
-        protected virtual decimal GetInitialMarginRequiredForOrder(
+        public virtual decimal GetInitialMarginRequiredForOrder(
             InitialMarginRequiredForOrderParameters parameters)
         {
             //Get the order value from the non-abstract order classes (MarketOrder, LimitOrder, StopMarketOrder)
@@ -156,7 +156,7 @@ namespace QuantConnect.Securities
         /// </summary>
         /// <param name="security">The security to compute maintenance margin for</param>
         /// <returns>The maintenance margin required for the </returns>
-        protected virtual decimal GetMaintenanceMargin(Security security)
+        public virtual decimal GetMaintenanceMargin(Security security)
         {
             return security.Holdings.AbsoluteHoldingsValue * _maintenanceMarginRequirement;
         }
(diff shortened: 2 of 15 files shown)


Common/Securities/BuyingPowerModel.cs
```diff
@@ -267,38 +270,38 @@ public virtual HasSufficientBuyingPowerForOrderResult HasSufficientBuyingPowerFo
...
+                return parameters.Sufficient();
             }
 
             var freeMargin = GetMarginRemaining(parameters.Portfolio, parameters.Security, parameters.Order.Direction);
             var initialMarginRequiredForOrder = GetInitialMarginRequiredForOrder(
-                new InitialMarginRequiredForOrderParameters(parameters.Portfolio.CashBook,
-                    parameters.Security,
-                    parameters.Order));
+                new InitialMarginRequiredForOrderParameters(
+                    parameters.Portfolio.CashBook, parameters.Security, parameters.Order
```
mchandschuh, line 289: I'd like to go through all of these model implementations and remove all the noise from the parameter and result classes. One pattern already used on some is result methods, such as this `Sufficient()` method directly on the parameter type. Another pattern to clean up calls to other model functions is implicit conversions. In this case, it would be a implicit conversion defined in `HasSufficientBuyingPowerForOrderParameters` that converts to `InitialMarginRequiredForOrderParameters` resulting in noise reduction:
``` csharp
var initialMarginRequiredForOrder = GetInitialMarginRequiredForOrder(parameters);
```

Common/Securities/HasSufficientBuyingPowerForOrderResult.cs
```diff
@@ -40,5 +40,21 @@ public HasSufficientBuyingPowerForOrderResult(bool isSufficient, string reason =
...
+        {
+            return new HasSufficientBuyingPowerForOrderResult(true);
+        }
+
+        /// <summary>
+        /// Creates a new result indicating that there is insufficient capital for the contemplated order
+        /// </summary>
+        public static HasSufficientBuyingPowerForOrderResult Insufficient(string reason)
+        {
+            return new HasSufficientBuyingPowerForOrderResult(false, reason);
```
Martin-Molinero, line 57: Believe these two new methods are unused, think we are using the similar ones added at `HasSufficientBuyingPowerForOrderParameters.cs`

Common/Securities/NullBuyingPowerModel.cs
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Provides an implementation of <see cref="IBuyingPowerModel"/> that uses an absurdly low margin
+    /// requirement to ensure all orders have sufficient margin provided the portfolio is not underwater.
+    /// </summary>
```
Martin-Molinero, line 21: Cool!
The only doubt with this guy is `SetHoldings` behavior because it will return a big number of holdings for say buying power of 1, since it uses the margin requirement and will still enforce margin requirements, it's just quantities are going to be **big**. See `PortfolioTarget.Percent()-> GetMaximumOrderQuantityForTargetBuyingPower`, maybe `GetMaximumOrderQuantityForTargetBuyingPower` should throw an invalid operation exception or reduce the quantity size returned somehow?

Common/Securities/IBuyingPowerModel.cs
```diff
@@ -37,6 +37,27 @@ public interface IBuyingPowerModel
...
+        /// <param name="security">The security to compute maintenance margin for</param>
+        /// <returns>The maintenance margin required for the </returns>
+        decimal GetMaintenanceMargin(Security security);
+
+        /// <summary>
+        /// The margin that must be held in order to increase the position by the provided quantity
+        /// </summary>
+        /// <param name="security">The security to compute initial margin for</param>
+        /// <param name="quantity">The change in the contemplated change in shares</param>
+        decimal GetInitialMarginRequirement(Security security, decimal quantity);
```
Martin-Molinero, line 52: Not sure if worth it but following the pattern of the methods below and for future proof, wondering if we should make these new methods return a `Result` class type and receive a `Parameters` class type? Maybe the result class isn't as needed but the parameters one might? 🤔

Martin-Molinero (commented):
Looks good! 👍 just minor comments/concerns

Common/Python/BuyingPowerModelPythonWrapper.cs
```diff
@@ -145,5 +145,42 @@ public void SetLeverage(Security security, decimal leverage)
...
+                return _model.GetInitialMarginRequirement(security, quantity);
+            }
+        }
+
+        /// <summary>
+        /// Gets the total margin required to execute the specified order in units of the account currency including fees
+        /// </summary>
+        /// <param name="parameters">An object containing the portfolio, the security and the order</param>
+        /// <returns>The total margin in terms of the currency quoted in the order</returns>
+        public decimal GetInitialMarginRequiredForOrder(InitialMarginRequiredForOrderParameters parameters)
```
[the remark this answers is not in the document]
mchandschuh: Thanks! That was a total oversight on my part 👍

Common/Securities/HasSufficientBuyingPowerForOrderResult.cs
mchandschuh: Great catch -- I'll re-review usages of any added methods. I think I initially added them to the result yielding:
``` csharp
return HasSufficientBuyingPowerForOrderResult.Insufficient()
```
and then remembered what we did for `BuyingPowerParameters` where the parameters simply have a `Result` method. Applying this pattern  yields what I ended up deciding on:
``` csharp
return parameters.Insufficient();
```

Thanks!

Common/Securities/NullBuyingPowerModel.cs
mchandschuh: Intersting. I did consider set holdings but appears I overlooked how it actually works. Maybe for the get max functions we do things more-or-less normally?

Common/Securities/IBuyingPowerModel.cs
mchandschuh: I had considered this as well --if we're confident that the number of potential broken users is near-zero then I think this is a good move -- the concern is if users do have to make this change, before it's just `protected` -> `public` -- but if we're changing the signatures/etc then that might be a little more difficult.

mchandschuh: Could add something to `BuyingPowerModel` that keeps the protected virtual and checks for overrides and provides an `[Obsolete]` error message via compile?

Martin-Molinero: > I had considered this as well --if we're confident that the number of potential broken users is near-zero then I think this is a good move 

I think so, because since these methods were protected only usages will be custom buying power models really which already would get broken by the `protected` -> `public` change?

Common/Securities/IBuyingPowerModel.cs
```diff
@@ -37,6 +37,26 @@ public interface IBuyingPowerModel
...
+        /// </summary>
+        /// <param name="parameters">An object containing the security and quantity</param>
+        InitialMargin GetInitialMarginRequirement(InitialMarginParameters parameters);
+
+        /// <summary>
+        /// Gets the total margin required to execute the specified order in units of the account currency including fees
+        /// </summary>
+        /// <param name="parameters">An object containing the portfolio, the security and the order</param>
+        /// <returns>The total margin in terms of the currency quoted in the order</returns>
+        InitialMargin GetInitialMarginRequiredForOrder(InitialMarginRequiredForOrderParameters parameters);
```
Martin-Molinero, line 58: Sweet 💯

Martin-Molinero (approved):
Cool! The only comment is about failing unit 👍

Common/Securities/IBuyingPowerModel.cs
```diff
@@ -37,6 +37,26 @@ public interface IBuyingPowerModel
         /// <param name="leverage">The new leverage</param>
         void SetLeverage(Security security, decimal leverage);
 
+        /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="parameters">An object containing the security</param>
+        /// <returns>The maintenance margin required for the </returns>
```
Martin-Molinero, line 44: nit - probably a copy pasta but `The maintenance margin required for the`...`provided quantity`?

Common/Securities/InitialMargin.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Result type for <see cref="IBuyingPowerModel.GetInitialMarginRequirement"/>
+    /// and <see cref="IBuyingPowerModel.GetInitialMarginRequiredForOrder"/>
+    /// </summary>
+    public class InitialMargin
+    {
+        /// <summary>
+        /// The maintenance margin value in account currency
```
Martin-Molinero, line 25: nit - `The maintenance` -> `The initial`, same bellow in the constructor

Common/Securities/BuyingPowerModelExtensions.cs
```diff
@@ -96,5 +96,49 @@ OrderDirection direction
             // existing implementations assume certain non-account currency units, so return raw value
             return buyingPower.Value;
         }
+
+        /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="model">The buying power model</param>
+        /// <param name="security">The security</param>
+        /// <returns>The maintenance margin required for the </returns>
```
Martin-Molinero: nit - `The maintenance margin required for the `... `current security holdings`?

Common/Securities/BuyingPowerModelExtensions.cs
```diff
@@ -96,5 +96,49 @@ OrderDirection direction
...
+            return model.GetMaintenanceMargin(MaintenanceMarginParameters.ForCurrentHoldings(security));
+        }
+
+        /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="model">The buying power model</param>
+        /// <param name="security">The security</param>
+        /// <param name="quantity">The quantity of shares</param>
+        /// <returns>The maintenance margin required for the </returns>
```
Martin-Molinero: nit - `The maintenance margin required for the `-> `The initial margin required for the provided security and quantity`

Common/Securities/ConstantBuyingPowerModel.cs
```diff
@@ -0,0 +1,71 @@
...
+        public override InitialMargin GetInitialMarginRequirement(InitialMarginParameters parameters)
+        {
+            return parameters.Quantity * _marginRequiredPerUnitInAccountCurrency;
+        }
+
+        /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="parameters">An object containing the security</param>
+        /// <returns>The maintenance margin required for the </returns>
```
Martin-Molinero: `The maintenance margin required for the `

Common/Python/PythonWrapper.cs
```diff
@@ -51,10 +51,15 @@ public static void ValidateImplementationOf<TInterface>(this PyObject model)
                     }
                 }
             }
+
             if (missingMembers.Any())
             {
-                throw new NotImplementedException($"{nameof(TInterface)} must be fully implemented. Missing implementations: {string.Join(", ", missingMembers)}");
+                throw new NotImplementedException($"{nameof(TInterface)} must be fully implemented. Please implement " +
+                    $"these missing methods on {model.GetPythonType()}: {string.Join(", ", missingMembers)}"
```
Martin-Molinero: I think `model.GetPythonType()` outside of the Py.GIL might be causing the explosions

Common/Securities/IBuyingPowerModel.cs
```diff
@@ -37,6 +37,27 @@ public interface IBuyingPowerModel
...
+        MaintenanceMargin GetMaintenanceMargin(MaintenanceMarginParameters parameters);
+
+        /// <summary>
+        /// The margin that must be held in order to increase the position by the provided quantity
+        /// </summary>
+        /// <param name="parameters">An object containing the security and quantity</param>
+        /// <returns>The initial margin required for the provided security and quantity</returns>
+        InitialMargin GetInitialMarginRequirement(InitialMarginParameters parameters);
+
+        /// <summary>
```
Martin-Molinero, line 54: - These 3 new methods in the interface refer to `Margin` versus other methods that refer to `BuyingPower` which is more abstract?
This question took me to review the CashPBM which does not implement any of these 3 methods and will use the underlying implementation which might not be right, since it's margin-based?

- Having `GetInitialMarginRequiredForOrder` and `GetInitialMarginRequirement` _might_ be tricky, because this last one does not take into account order fees which the first one does, same as `HasSufficientBuyingPowerForOrder`. So the result of `GetInitialMarginRequirement` might not be completely valid by it self since its missing fees?

Common/Securities/IBuyingPowerModel.cs
```diff
@@ -37,6 +37,26 @@ public interface IBuyingPowerModel
         /// <param name="leverage">The new leverage</param>
         void SetLeverage(Security security, decimal leverage);
 
+        /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="parameters">An object containing the security and holdings quantity/cost/value</param>
+        /// <returns>The maintenance margin required for the provided holdings quantity/cost/value</returns>
+        MaintenanceMargin GetMaintenanceMargin(MaintenanceMarginParameters parameters);
```
Martin-Molinero, line 45: TODO: 
- Make `MaintenanceMargin` & `InitialMargin` use a `List<CashAmount>` instead of a plane `decimal`
- `CashBuyingPM` implement `InitialMargin` methods so that it returns in the actual currency required, not in account currency
- Rename `Margin` for `BuyingPower`
- If `GetReservedBuyingPowerForPosition` supported a provided quantity instead of just using `Holdings`, we can remove `GetMaintenanceMargin` from interface
- Switch `GetInitialMarginRequirement` to protected again, if possible
- `HasSufficientBuyingPowerForOrder` could include the initial margin requirement for order in the response and we can get rid of `GetInitialMarginRequiredForOrder`

Martin-Molinero:
Included by https://github.com/QuantConnect/Lean/pull/5511

PR #5110, opened 2021-01-01
Fix MACD WarmUpPeriod and Updating

### LINKED ISSUE #5109: Incorrect MACD WarmUpPeriod & ComputeNextValue
#### Expected Behavior

Regarding the MovingAverageConvergenceDivergence indicator:
(1) The `WarmUpPeriod` is set to
`max(fastPeriod, slowPeriod) + signalPeriod`

(2) The `Signal` is only updated when `Fast` and `Slow` are ready.

(3) The `Histogram` is only updated when `Signal.IsReady` 

#### Actual Behavior

The three conditions above are not true. See [this related thread](https://www.quantconnect.com/forum/discussion/10229/moving-average-convergance-divergence-warm-up-period/p1).

#### Potential Solution

(1) Change [line 83](https://github.com/QuantConnect/Lean/blob/03efc1b735bb8b841eb04434f537ac3d798f5877/Indicators/MovingAverageConvergenceDivergence.cs#L83) to
`Math.Max(fastPeriod, slowPeriod) + signalPeriod`

(2 & 3) Add `IsReady` guards before updating `Signal` and `Histogram`.

#### Reproducing the Problem

Run [this backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_a50cbc073f15ad4aa0b4e4983109b15e.html) in the cloud. The logs show the MACD indicator `IsReady` when the internal `Fast` and `Slow` indicators are not ready.

#### System Information

Windows 10. Run in QC Cloud.

#### Checklist

DerekMelchin:
#### Description

The WarmUpPeriod of the MACD is corrected to include all internal indicators. In addition, we now only update the `Signal` and `Histogram` when the internal indicators used in their computations are ready.

#### Related Issue

Close #5109

#### Motivation and Context

Before this change, the MACD indicator could be ready while it's internal indicators are not ready. The backtest shared in the related issue demonstrates this.

No.

The new indicator was run in the Cloud in[ this backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_720865fb20fe671956a8afec7c7499da.html). The logs show the indicator now has the correct `IsReady` state.

#### Types of changes

#### Checklist:

- [ ] I have added tests to cover my changes.  This probably needs a test added, but I'm not sure how to add unit tests to LEAN yet.

### DIFF
--- a/Indicators/MovingAverageConvergenceDivergence.cs
+++ b/Indicators/MovingAverageConvergenceDivergence.cs
@@ -79,14 +79,14 @@ namespace QuantConnect.Indicators
         {
             if (fastPeriod >= slowPeriod)
             {
-                throw new ArgumentException("MovingAverageConvergenceDivergence: fastPeriod must be less than slowPeriod", "fastPeriod, slowPeriod");
+                throw new ArgumentException("MovingAverageConvergenceDivergence: fastPeriod must be less than slowPeriod", $"{nameof(fastPeriod)}, {nameof(slowPeriod)}");
             }
             
             Fast = type.AsIndicator(name + "_Fast", fastPeriod);
             Slow = type.AsIndicator(name + "_Slow", slowPeriod);
             Signal = type.AsIndicator(name + "_Signal", signalPeriod);
             Histogram = new Identity(name + "_Histogram");
-            WarmUpPeriod = Math.Max(fastPeriod, slowPeriod) + signalPeriod - 1;
+            WarmUpPeriod = slowPeriod + signalPeriod - 1;
         }
 
         /// <summary>
--- a/Indicators/MovingAverageConvergenceDivergence.cs
+++ b/Indicators/MovingAverageConvergenceDivergence.cs
@@ -80,7 +80,7 @@ namespace QuantConnect.Indicators
             Slow = type.AsIndicator(name + "_Slow", slowPeriod);
             Signal = type.AsIndicator(name + "_Signal", signalPeriod);
             Histogram = new Identity(name + "_Histogram");
-            WarmUpPeriod = signalPeriod;
+            WarmUpPeriod = Math.Max(fastPeriod, slowPeriod) + signalPeriod;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Indicators/MovingAverageConvergenceDivergence.cs
```diff
@@ -95,8 +96,16 @@ protected override decimal ComputeNextValue(IndicatorDataPoint input)
...
-            Signal.Update(input.Time, macd);
-            Histogram.Update(input.Time, macd - Signal.Current.Value);
+            if (Fast.IsReady && Slow.IsReady)
+            {
+                Signal.Update(input.Time, macd);
+            }
+                
+            if (Signal.IsReady)
+            {
+                Histogram.Update(input.Time, macd - Signal.Current.Value);
```
Martin-Molinero, line 106: Believe this `if` could be moved into the one above. 
Instead of consuming the `IsReady` properties, we could directly use the returned value from the Update calls since we already have them

Martin-Molinero (commented):
Nice!
Some unit tests are failing because of this change, think we should add some new test too

Indicators/MovingAverageConvergenceDivergence.cs
```diff
@@ -80,7 +81,7 @@ public MovingAverageConvergenceDivergence(string name, int fastPeriod, int slowP
             Slow = type.AsIndicator(name + "_Slow", slowPeriod);
             Signal = type.AsIndicator(name + "_Signal", signalPeriod);
             Histogram = new Identity(name + "_Histogram");
-            WarmUpPeriod = signalPeriod;
+            WarmUpPeriod = Math.Max(fastPeriod, slowPeriod) + signalPeriod - 1;
```
jaredbroad, line 84: Should we throw if `fast >= slow`? It'll result in zero or negative `var macd = Fast.Current.Value - Slow.Current.Value;` then weirdness in `Histogram.Update(input.Time, macd - Signal.Current.Value);`

DerekMelchin: Yes, good point. I'll add this tomorrow.

Martin-Molinero (approved):

PR #5111, opened 2021-01-02
Bug insufficient buying power unrounded amount

BadgerTheAdger:
When buying power check is made, not rounded amounts are used. But some brokerages (I believe all of them) have limitations such as LotSize, meaning that one can't place an amount with a precision higher than lotsize's one. Buying power check should take that into consideration. 

#### Description
I have added some extension that gets precision out of lot size, and then I rounded quantity variables to this precision in the HasBuyingPowerModel code

#### Motivation and Context
I've been running some paper trading with an algorithm that places trades with 100% of the available amount. So it was pretty close to encountering the bloody buying power model. But it shouldn't in theory, anyway. But not in this case - in my case the model was checking if I had 10000 BTC but it calculated (base amount * price) something like 9999.999999999999999999998. Needless to say, that's equal to 10000, considering the lot size which was something like 0.0000001.

So I have added some rounding to lotSize so that the model doesn't compare wrong values. 

I tested privately in my private fork, I did not add unit tests feel free to add them but I think it's a pretty simple change and I tested it by running paper & backtests

#### Types of changes

#### Checklist:

- [ ] I have added tests to cover my changes.  - I tested in my private fork manually


Common/Securities/CashBuyingPowerModel.cs
```diff
@@ -118,6 +125,8 @@ public override HasSufficientBuyingPowerForOrderResult HasSufficientBuyingPowerF
                     GetMaximumOrderQuantityForTargetBuyingPower(
                         new GetMaximumOrderQuantityForTargetBuyingPowerParameters(parameters.Portfolio, parameters.Security, targetPercent)).Quantity * GetOrderPrice(parameters.Security, parameters.Order);
 
+                maximumQuantity = Math.Round(maximumQuantity, amountPrecision);
```
Martin-Molinero, line 128: `GetMaximumOrderQuantityForTargetBuyingPower` is already using the lotSize to determine the correct quantity, this shouldn't be required or it's a bug in the `GetMaximumOrderQuantityForTargetBuyingPower` implementation

Common/Securities/CashBuyingPowerModel.cs
```diff
@@ -81,6 +82,11 @@ public override HasSufficientBuyingPowerForOrderResult HasSufficientBuyingPowerF
                 orderQuantity = parameters.Order.AbsoluteQuantity;
             }
 
+            var amountPrecision = parameters.Security.SymbolProperties.LotSize.GetDecimalPlaces();
+
+            totalQuantity = Math.Round(totalQuantity, amountPrecision);
+            orderQuantity = Math.Round(orderQuantity, amountPrecision);
```
Martin-Molinero, line 88: Interesting, _I think_ this might be because the provided argument in `HasSufficientBuyingPowerForOrderParameters` aren't respecting the lot size

Common/Securities/CashBuyingPowerModel.cs
```diff
@@ -89,7 +95,8 @@ public override HasSufficientBuyingPowerForOrderResult HasSufficientBuyingPowerF
             if (parameters.Order.Direction == OrderDirection.Sell)
             {
                 // can sell available and non-reserved quantities
-                isSufficient = orderQuantity <= totalQuantity - openOrdersReservedQuantity;
+                var availableQuantity = Math.Round(totalQuantity - openOrdersReservedQuantity, amountPrecision);
```
Martin-Molinero, line 98: I think `totalQuantity - openOrdersReservedQuantity` shouldn't require rounding, since the current holding and the open orders should already respect the lot size, else it means there is some other bug

Martin-Molinero (commented):
Hey Mark!
Thanks for the PR. I think we do need tests for these changes, covering the use case and expected behavior. Curious if you are using python algorithms?
_I suspect_ that it's the parameters passed into `HasSufficientBuyingPowerForOrder` which are not respecting the lot size. But not sure how is this affecting behavior since this method just returns true or false?

Common/Securities/CashBuyingPowerModel.cs
BadgerTheAdger: I have code that does round to lot size before calling the HasSufficientBuyingPowerForOrderParameters. I can add some tests but that will be some time on weekends I believe

In case it's a buy, we multiply the values (parameters.Order.AbsoluteQuantity * GetOrderPrice(parameters.Security, parameters.Order)) which could cause rounding issues. 

Imagine these values: 20.5 * 2.01  - they produce 41.2005 which has the precision of 4, while the parameters have the precision of 1 and 2. So it's known that values with lesser precision can result in values with greater precision after division or multiplication has been performed on them. 

That's why we need to round order quantity which is quantity converted into quote currency. And after that conversion, the precision could become greater than the lot size.

The reason why we need to round total quantity - that's because it's not always valid in the context of lot size, for example after an order has been filled we make calculations in the ProcessFill method, but these calculations do not respect the lot size, again the mul/div operations can cause values of greater precision, which is not taken into consideration in the ProcessFill method. But now I am thinking that we shouldn't really round total quantity but perhaps better fix the error in the ProcessFill methods

Common/Securities/CashBuyingPowerModel.cs
BadgerTheAdger: Hm ok I will need to add tests that reproduce the issues but based on the code I can already see an issue, the code below does not respect the lot size, it does not round after multiplication has been performed
image

Common/Securities/CashBuyingPowerModel.cs
BadgerTheAdger: I agree, it's a fix of a symptom vs actual issue.

I think total quantity can have invalid value because of Portfolio manipulations (ProcessFill method) that don't respect lot size i.e they don't round amounts after multiplications.

But here it's a '-' so shouldn't require Math.Round

BadgerTheAdger:
Also, I have spotted that Limit order results do not round amounts to lotSize, but market orders do

PR #5113, opened 2021-01-04
Fix BTC future price multiplier

Martin-Molinero:


jaredbroad (approved):

gsalaz98 (changes requested):
IB says that this contract has a multiplier of 5x. I think this might either be an issue with the raw data or the conversion of the data.

gsalaz98 (approved):
My apologies, I thought I was looking at the SPDB

Martin-Molinero:
Thanks for verifying this @gsalaz98 👍

PR #5114, opened 2021-01-04
Refactors Market Fill of Equity Fill Model 

AlexCatarino:
#### Description
- Create `GetBidPrice` and `GetAskPrice` methods to request the most suitable price to fill market orders.
- Fix unit tests to show that the new implementation prevents using non-market data to fill the order.
- Fix regression test: the previous market fill method would consider the last tick from the last collection of ticks that was a tick of trade type. With this PR, the model looks for the last tick with quote type. 

#### Related Issue
Ref.: #4567 

Existing unit and regression tests.
- Market Fill does not change the core current behavior. However once accepts market data now while it could accept any other type that can set the security price.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Orders/Fills/EquityFillModel.cs
+++ b/Common/Orders/Fills/EquityFillModel.cs
@@ -543,7 +543,7 @@ namespace QuantConnect.Orders.Fills
 
         /// <summary>
         /// Get current bid price for subscribed data
-        /// This method will try to get the most recent ask price data, so it will try to get tick quote first, then quote bar.
+        /// This method will try to get the most recent bid price data, so it will try to get tick quote first, then quote bar.
         /// If no quote, tick or bar, is available (e.g. hourly data), use trade data with preference to tick data.
         /// </summary>
         /// <param name="asset">Security which has subscribed data types</param>
--- a/Common/Orders/Fills/EquityFillModel.cs
+++ b/Common/Orders/Fills/EquityFillModel.cs
@@ -14,12 +14,14 @@
 */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
 using QuantConnect.Python;
 using QuantConnect.Orders.Fees;
 using QuantConnect.Securities;
+using QuantConnect.Util;
 
 namespace QuantConnect.Orders.Fills
 {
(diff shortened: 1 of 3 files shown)


Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -460,6 +463,144 @@ public virtual OrderEvent MarketOnCloseFill(Security asset, MarketOnCloseOrder o
...
+                    return tradeBar.Close;
+                }
+            }
+
+            throw new InvalidOperationException($"Cannot get ask price to perform fill for {asset.Symbol} because no market data subscription were found.");
+        }
+
+        /// <summary>
+        /// Get current bid price for subscribed data
+        /// This method will try to get the most recent ask price data, so it will try to get tick quote first, then quote bar.
```
Martin-Molinero, line 546: nit - `most recent ask price data` should be `bid`

Martin-Molinero (commented):
Looks good!
Just `EquityTickQuoteAdjustedModeRegressionAlgorithm` is failing, think because there might be some difference with how `master` is getting the ticks. Seems to me the PR approach of using `GetAll<Tick>` is actually better

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -600,6 +610,35 @@ private decimal GetBidPrice(Security asset, out DateTime endTime)
...
+            if (subscribedTypes.Contains(typeof(TradeBar)))
+            {
+                var tradeBar = asset.Cache.GetData<TradeBar>();
+                if (tradeBar != null)
+                {
+                    return tradeBar;
+                }
+            }
+
+            throw new InvalidOperationException($"Cannot perform fill for {asset.Symbol} because no trade data were found.");
```
Martin-Molinero: Not 100% sure about throwing here, what if no trade has happened yet? as long as there is a subscription to TradeBar or Tick, don't think it should throw?

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -391,9 +391,21 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
+            {
+                return fill;
+            }
 
-            fill.FillPrice = GetPricesCheckingPythonWrapper(asset, order.Direction).Open;
+            // For non-daily resolution, check whether the last trade bar
+            // If the first data after the market opens is a quote bar, the open of the last trade bar is not the open
+            // for the current day, but the last open in cache
+            var resolution = Parameters.ConfigProvider.GetSubscriptionDataConfigs(asset.Symbol).GetHighestResolution();
+            if (resolution != Resolution.Daily && !asset.Exchange.DateTimeIsOpen(lastTradeBar.EndTime))
```
Martin-Molinero: `asset.Exchange.DateTimeIsOpen(lastTradeBar.EndTime)` will be false of the bar ending at the market close, which I think should be a valid bar to use? maybe we could just compare the lastTradeBar date > order date?

Martin-Molinero (commented):
Cool! Leaving a couple of minor concerns

Martin-Molinero (approved):

PR #4941, opened 2020-11-17
Bug 4925 daylight out of order bar

AdalyatNazirov:
#### Description

Reverted PRs https://github.com/QuantConnect/Lean/pull/3708, https://github.com/QuantConnect/Lean/pull/4700

behind these changes a few main ideas:
1) In most cases in Lean we calculate `EndTime=Time + resolution`. it means that `EndTime` is built in the same TZ as `Time`, because simple arithmetic operation doesn't take into account TZ itself. It works for standard situation. But broken when it comes to specific days when we switch TZ to Daylight/Standard.

That's why both `potentialbarEndTime` and `next.EndTime` should be in the same TZ (Daylight/Standard). because of  `next.EndTime` is built in 'current TZ' we also need to build  `potentialbarEndTime` in 'current'. To do that  it's necessary to `RoundDown` origin day and then add interval, while those PRs rounded down the day + interval, which would not work for daily resolution because we missed the 2am of Sunday where the movement actually happens. In the https://github.com/QuantConnect/Lean/pull/4700 we tried to compensate that by adding `daylightMovement` but didn't worked for minute resolution on pre-market data

2) Usually `Time` and `EndTime`  are in the same TZ, but if we consider that specific day of Daylight movement, we can find that `Time` is one TZ, but `EndTime` in another. In `master` we ignore that, so ff'ed bar `Time` and `EndTime` are always from the same TZ, but it breaks 'next' step. To fix it we need apply their own TZ to `Time` and `EndTime`

There are a few examples (DataTimeZone = UTC-5; Exchange TimeZone = America/new York)

**Standard Time** 
00:00:00 => 1.00:00:00
image

**Daylight Time**
01:00:00 => 1.01:00:00
image

### **EDGE CASES**

**Standard => Daylight**
00:00:00 => 1.01:00:00
image

**Daylight => Standard**
020:00:00 => 1.19:00:00
image

Worth to mention that `NodaTime` consider 1 AM - 2 AM as interval _after_ daylight move, so it's sensitive for DataTZ and ExchangeTZ, for example

DataTZ = UTC; ExchangeTZ = America/New York
020:00:00 => 1.19:00:00

DataTZ = UTC-5; ExchangeTZ = America/New York
01:00:00 => 1.01:00:00

#### Related Issue

Closes #4925 

#### Motivation and Context

On minute resolution error was thrown for pre-market data `FillForwardEnumerator received data out of order`.

#### Requires Documentation Change

N/A

#### How Has This Been Tested?

[...]

PR #5118, opened 2021-01-05
FillForward enumerator does not loop infinitely if fillforward resolution is different from data resolution

### LINKED ISSUE #5116: FillForward enumerator infinite loop
#### Expected Behavior

- FillForward enumerator does not loop infinitely
#### Actual Behavior

- FillForward enumerator is looping infinitely and blocking the backtest. Started happening with https://github.com/QuantConnect/Lean/pull/4941
#### Potential Solution

N/A
#### Reproducing the Problem

```

    def Initialize(self):
        self.SetStartDate(2007, 1, 1)
        self.AddEquity('QQQ', Resolution.Hour)
        self.AddForex('EURUSD', Resolution.Daily, Market.FXCM)
```
#### System Information

N/A
#### Checklist

AdalyatNazirov:
#### Description

Calculate `potentialBarEndTime` in data time zone. `next.EndTime` is also replaced with `next.Time' converted into data time zone.

It's worth to consider this PR as continuation of https://github.com/QuantConnect/Lean/pull/4941.  There are two things that were not fixed in original PR.

During reading and parsing data from file we consider them as data in data time zone, and then convert them into exchange time zone. `TradeBar.EndTime` swallow Daylight Saving Time hour because is built using simple arithmetic operations (+/-).

That's why comparison `potentialBarEndTime` and `next.EndTime` isn't allowed - they can be in different TZ. To solve this problem we convert and compare dates built on top of `Start Time`.

Another change is that  `fillforward.EndTime` became fixed, and `fillforward.Time` is calculated by rewinding bar size. 

#### Related Issue

Closes #5116 

#### Motivation and Context

FillForward enumerator is looping infinitely and blocking the backtest. Started happening with https://github.com/QuantConnect/Lean/pull/4941

N/A

More unit tests. Created unit tests around of Daylight Saving time changes for different fillforward and data resolutions.

#### Types of changes

#### Checklist:


Engine/DataFeeds/Enumerators/FillForwardEnumerator.cs
```diff
@@ -300,7 +300,8 @@ protected virtual bool RequiresFillForwardData(TimeSpan fillForwardResolution, B
                 // it means that we have to calculate potentialBadEndTime in the same TZ
                 // by adding interval in daily resolution we move to next day
                 // in order to track daylight movement need to round-down origin day (before + Interval)
-                var potentialBarEndTime = RoundDown(item.ReferenceDateTime, item.Interval) + item.Interval;
+                var potentialBarEndTime = (RoundDown(item.ReferenceDateTime, item.Interval).ConvertTo(Exchange.TimeZone, _dataTimeZone) + item.Interval)
```
Martin-Molinero, line 303: Believe we should update the comments above and the _why_ we are doing these operations

jaredbroad (commented):
Pending 2 unit tests demonstrating the "skip" or "repeat" of the 2am Sunday bar during a timezone change. Expect the 2 am to not repeat due to how LEAN ignore <= bars; but should definitely see it _skip_ the bar when time moves forward.

mchandschuh:
FYI -- time comparisons should _always_ use end time

Martin-Molinero (approved):
Thanks for all gray matter put into this one 👍

PR #5125, opened 2021-01-05
Fix column headings in Data/equity/readme.md

RileyMShea:
* Swapped incorrect column header positions in readme.md table


Martin-Molinero (approved):
Thanks!

Jay-Jay-D (approved):

PR #5129, opened 2021-01-06
Fixes Market On Open Fill of Equity Fill Model

AlexCatarino:
#### Description
Only use trade data (`Tick` with `TickTrade` type or `TradeBar`) to get the open price, since MOO is filled with the opening auction price. Ensure that this method doesn't use trade data from before the market opens for high-resolution data case.

- Change regression tests to reflect the bug fix.
  - In the `ExtendedMarketHoursHistoryRegressionAlgorithm`, MOO was filled with extended market hours.
  - In the `RegressionAlgorithm`, MOO was filled with open prior to the order. The algorithm now has one order less, since the last MOO would need to wait another day to be filled.
  - All other changes are the consequence of using `TradeBar` instead of `QuoteBar`.

#### Related Issue
Ref.: #4567 

Unit and regression tests.

#### Types of changes

#### Checklist:


Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -379,22 +379,58 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
             if (asset.Exchange.DateTimeIsOpen(localOrderTime) && localOrderTime.Date == asset.LocalTime.Date)
             {
                 return fill;
             }
 
+            // If the first data after the market opens is a quote bar (high resolution only),
+            // the open of the last trade bar is not the open for the current day, but the last open in cache.
+            // We need to verify whether the trade data is from the open market.
+            var resolution = Parameters.ConfigProvider.GetSubscriptionDataConfigs(asset.Symbol).GetHighestResolution();
+            if (resolution < Resolution.Hour && !asset.Exchange.DateTimeIsOpen(openTime))
```
Martin-Molinero, line 421: Suggest we move this change upwards into each tick or tradeBar `if` case, don't think we need to rely on the `.GetHighestResolution()`, we don't know the subscriptions it's looking at. For the `tick` case we can directly check `asset.Exchange.DateTimeIsOpen()`. For the tradeBar case we could call `tb.Period.ToHigherResolutionEquivalent()` and get the exact resolution of the bar we are going to use, think it would be a more transparent decision/logic

Martin-Molinero (commented):
Cool!
Leaving a minor suggestion and I think should mention in the PR description why all of the stats changed, only `ExtendedMarketHoursHistoryRegressionAlgorithm.cs` is mentioned

Tests/Common/Orders/Fills/EquityFillModelTests.cs
```diff
@@ -435,6 +436,80 @@ public void PerformsMarketOnOpenUsingOpenPrice()
...
+            }, typeof(Tick));
+
+            fill = model.MarketOnOpenFill(security, order);
+            Assert.AreEqual(0, fill.FillQuantity);
+
+            // One trade and quote is received after the market is open, but there is trade prior to that with different price
+            security.Update(new List<Tick>
+            {
+                new Tick(time.AddMinutes(-1), Symbols.SPY, "", "", 100, 0.9m),
+                new Tick(time, Symbols.SPY, "", "", 100, expected),
```
Martin-Molinero, line 504: Cool!

Martin-Molinero (approved):

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -379,9 +379,41 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
+                    fill.FillPrice = tradeBar.Open;
+                }
+            }
+
             var localOrderTime = order.Time.ConvertFromUtc(asset.Exchange.TimeZone);
-            if (currentBar == null || localOrderTime >= currentBar.EndTime) return fill;
+            if (localOrderTime >= endTime) return fill;
 
             // if the MOO was submitted during market the previous day, wait for a day to turn over
             if (asset.Exchange.DateTimeIsOpen(localOrderTime) && localOrderTime.Date == asset.LocalTime.Date)
```
Martin-Molinero, line 419: What if `TradeBar` was resolution Hour and Daily, check on 405 would be skipped, asset local time could already be market open of the next day by the trade data would be of the previous? 🤔

Martin-Molinero (approved):

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -379,12 +379,49 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
+                    {
+                        return fill;
+                    }
+
+                    endTime = tradeBar.EndTime;
+                    fill.FillPrice = tradeBar.Open;
+                }
+            }
+
+            if (localOrderTime >= endTime) return fill;
```
jaredbroad, line 419: Minute data:  Submission Previous Day / Same Day 12am / 8am / 930am, first bar sets time at X - End time 931am
Hour data:     Submission Previous Day / Same Day 12am / 8am / 930am, first bar sets time at X  - End time 10am
Daily data:     Submission Previous Day / Same Day 12am / 8am / 930am, first bar sets time at X  - End time 12am tomorrow.

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -379,12 +379,49 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
 
             // if the MOO was submitted during market the previous day, wait for a day to turn over
-            if (asset.Exchange.DateTimeIsOpen(localOrderTime) && localOrderTime.Date == asset.LocalTime.Date)
+            // The date of the order and the trade data end time cannot be the same.
+            // Note that the security local time can be ahead of the data end time.
+            if (asset.Exchange.DateTimeIsOpen(localOrderTime) && localOrderTime.Date == endTime.Date)
             {
                 return fill;
             }
```
jaredbroad, line 427: This also blocks premarket submissions

jaredbroad (changes requested):

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -379,12 +379,49 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
+            if (subscribedTypes.Contains(typeof(Tick)))
+            {
+                // Get the first valid (non-zero) tick of trade type from an open market
+                var trade = asset.Cache.GetAll<Tick>().FirstOrDefault(x =>
+                    x.TickType == TickType.Trade && x.Price > 0 && asset.Exchange.DateTimeIsOpen(x.Time));
+                if (trade != null)
+                {
+                    endTime = trade.EndTime;
+                    fill.FillPrice = trade.Price;
+                }
```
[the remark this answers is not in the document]
Jay-Jay-D: Yes, tick data have flags, from there we can know wich one is the opening acution price.

Common/Orders/Fills/EquityFillModel.cs
AlexCatarino: I added this unit test to address this comment:
https://github.com/QuantConnect/Lean/pull/5129/commits/78ddf8af110a33aafb2d0fcf31a269d0f3561572#diff-18938dbae43240eab6eee79fdd2221594549c57b2b4937ba737188dd5197d9d5R556

PR #5132, opened 2021-01-06
Fix email new line symbol

Marrri4ka:
- Notification system looks like this:
msg = 'First line\nSecond line\n Third line\n'
self.Notify.Email('lorem@quantconnect.com', 'Hi!', msg)
self.Quit(msg)

- Symbol '\n'  is converted to <br> in the terminal, but not in email.  Symbol' \n' was also a new line in email.

- looks like this is the problem: https://stackoverflow.com/questions/16063312/new-lines-r-n-are-not-working-in-email-body

- the solution in this pr suggest to add new header

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Notifications/NotificationManager.cs
+++ b/Common/Notifications/NotificationManager.cs
@@ -78,6 +78,11 @@ namespace QuantConnect.Notifications
             {
                 return false;
             }
+            
+            if (headers == null) {
+                headers = new Dictionary<string, string>();
+            }
+            headers['Content-Type'] = 'text/plain';
 
             var email = new NotificationEmail(address, subject, message, data, headers);
             Messages.Enqueue(email);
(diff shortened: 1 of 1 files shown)


Common/Notifications/NotificationManager.cs
```diff
@@ -78,6 +78,11 @@ public bool Email(string address, string subject, string message, string data =
             {
                 return false;
             }
+            
+            if (headers == null) {
+                headers = new Dictionary<string, string>();
+            }
+            headers['Content-Type'] = 'text/plain';
```
Martin-Molinero, line 85: Not sure we should override the `Content-Type`, what if the user wants a custom/different content type, maybe this could be the default value if `headers` doesn't contain `'Content-Type'`?

Marrri4ka: Fixed.

Martin-Molinero:
Available to test at `test-email-default-content-type-e6360b5-0a1fc2b v10265`

gaviles:
After talking with @Martin-Molinero we decided that this is an QC Web API concern.

PR #5131, opened 2021-01-06
Fixes Market On Close Fill of Equity Fill Model

AlexCatarino:
#### Description
Only use trade data (`Tick` with `TickTrade` type or `TradeBar`) to get the close price, since MOC is filled with the closing auction price. Ensure that this method doesn't use trade data from the previous day by comparing the trade data end time and the next market close.

- Fix unit tests to show that the new implementation only fills with trade data from the current close market.
- Change `UpdateOrderRegressionAlgorithm` regression tests to reflect the bug fix, since it was using the closing price of the previous day.

#### Related Issue
Ref.: #4567 

#### How Has This Been Tested?
Unit and regression tests.

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #5133, opened 2021-01-06
Fixes Stop Market Fill of Equity Fill Model

AlexCatarino:
#### Description
Only use trade data (`Tick` with `TickTrade` type or `TradeBar`) to get the high and low prices, since stop market orders are triggered with high and low prices. The fill price is the worst-case scenario between the stop price and the bid/ask, unless there is a gap and the open price is used instead.
https://www1.interactivebrokers.com/en/index.php?f=609

- Fix unit tests to show that the new implementation only fills with high and low values of trade data.
- Change `StopLossOnOrderEventRegressionAlgorithm` regression tests to reflect the bug fix. The change results from the price gap handling as the both TradeBar and QuoteBar high and low triggers the stop loss.

**Note**: This PR also changes statistics of `UpdateOrderRegressionAlgorithm`. However, since #5131 also does, it would be better to address it once #5131 is merged.

#### Related Issue
Ref.: #4567 
Closes #4546

Unit and regression tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/StopLossOnOrderEventRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/StopLossOnOrderEventRegressionAlgorithm.cs
@@ -73,16 +73,16 @@ namespace QuantConnect.Algorithm.CSharp
         public Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>
         {
             {"Total Trades", "2"},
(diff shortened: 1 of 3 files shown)


Martin-Molinero:
Closing for now until https://github.com/QuantConnect/Lean/pull/5129 is shipped/address first

PR #5142, opened 2021-01-07
Adds FuturesOptionsUnderlyingMapper and FuturesListing

gsalaz98:
#### Description

  This adds a mapper for Futures options that have a different
  contract month between the FOP and the underlying future, as is
  observed in the contracts ZB, ZC, ZS, ZT, ZW, HG, SI, GC.
  An example of this is GC (e.g. OGH21 -> GCJ21). Other contracts follow
  different rules, which have been included as part of this commit.

  * Adds unit tests for FuturesOptionsUnderlyingMapper

  * Adds unit and regresssion tests for underlying Futures mapping for FOPs

#### Related Issue

N/A
#### Motivation and Context

Allows us to map the FOP to the correct underlying, which is critical for processing/loading of the data.

Maybe, one Future can have multiple FOPs in their option chain.

New unit tests, regression algorithm
#### Types of changes

#### Checklist:


Common/Securities/FutureOption/FuturesOptionsUnderlyingMapper.cs
```diff
@@ -0,0 +1,172 @@
...
+                var newFutureContractMonth = _underlyingFuturesOptionsRules[futureTicker](contractMonth, date);
+                if (newFutureContractMonth == null)
+                {
+                    // This will only happen when we search the Futures chain for a given contract and no
+                    // closest match could be made, i.e. there are no futures in the chain that come after the FOP's
+                    // contract month.
+                    return null;
+                }
+
+                contractMonth = newFutureContractMonth.Value;
```
Martin-Molinero, line 93: So this means that the response of `FuturesOptionsExpiryFunctions.GetFutureContractMonth` isn't always correct, what about the other usages, like IB, I think they shouldn't be directly calling it? but instead calling this new mapper class. 

~We should add a big comment at `FuturesOptionsExpiryFunctions.GetFutureContractMonth` about this and probably point to this class~. Or actually even **better** we should have a single method that returns the correct future contract month directly 🤔 else it's like a bug waiting to happen.

Might we worth reviewing if there are any other similar cases where these helper methods are returning a buggy response which requires the caller to know about it and adjust correctly with another call

Common/Securities/FutureOption/FuturesOptionsUnderlyingMapper.cs
```diff
@@ -0,0 +1,172 @@
...
+            return futureOptionContractMonth;
+        }
+
+        /// <summary>
+        /// Sets the contract month to the third month for the first 3 months, then begins using the <see cref="ContractMonthEvenOddMonth"/> rule.
+        /// </summary>
+        /// <param name="futureOptionContractMonth">Future option contract month. Note that this is not the expiry of the Future Option.</param>
+        /// <param name="oddMonths">True if the Future Option's underlying future contract month is on odd months, false if on even months. Only used for months greater than 3 months</param>
+        /// <returns></returns>
+        protected DateTime ContractMonthYearStartThreeMonthsThenEvenOddMonthsSkipRule(DateTime futureOptionContractMonth, bool oddMonths)
```
Martin-Molinero, line 162: Believe should make these methods directly `private` since there doesn't seem to be a reason for them to be `protected`?

Martin-Molinero (commented):
Nice work! Great tests

Brokerages/InteractiveBrokers/InteractiveBrokersSymbolMapper.cs
```diff
@@ -151,15 +151,21 @@ public Symbol GetLeanSymbol(string brokerageSymbol, SecurityType securityType, s
...
+                        var future = FuturesOptionsUnderlyingMapper.GetUnderlyingFutureFromFutureOption(
+                            GetLeanRootSymbol(brokerageSymbol),
+                            market,
+                            expirationDate,
+                            DateTime.Now);
+
+                        if (future == null)
+                        {
+                            // This is the worst case scenario, because we didn't find a matching futures contract for the FOP.
+                            // Note that this only applies to CBOT symbols for now.
```
Martin-Molinero, line 163: This shouldn't happen now that the rules are in place right?

Tests/Common/Securities/Futures/FuturesListingsTests.cs
```diff
@@ -0,0 +1,150 @@
...
+                        new DateTime(2021, 9, 14),
+                        new DateTime(2021, 12, 14),
+                        new DateTime(2022, 3, 14),
+                        new DateTime(2022, 5, 13),
+                        new DateTime(2022, 7, 14),
+                        new DateTime(2022, 9, 14),
+                        new DateTime(2022, 12, 14),
+                        new DateTime(2023, 3, 14),
+                        new DateTime(2023, 5, 12),
+                        new DateTime(2023, 7, 14)
```
Martin-Molinero, line 144: Nice! 👍

Tests/Common/Securities/Options/OptionChainProviderTests.cs
```diff
@@ -22,6 +22,7 @@
 using QuantConnect.Interfaces;
 using QuantConnect.Lean.Engine.DataFeeds;
 using QuantConnect.Securities.Future;
+using QuantConnect.Securities.FutureOption;
```
Martin-Molinero, line 25: An unrequired import?

Common/Securities/FutureOption/FuturesOptionsUnderlyingMapper.cs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        /// <summary>
+        /// Gets the theoretical (i.e. intermediate/naive) future contract month if we assumed a 1-1 mapping
+        /// between FOPs contract months and Futures contract months, i.e. they share the same contract month.
+        /// </summary>
+        /// <param name="canonicalFutureSymbol">Canonical future Symbol</param>
+        /// <param name="futureOptionExpirationDate">Future Option Expiration Date</param>
+        /// <returns>Contract month assuming that the Future Option and Future share the same contract month</returns>
+        public static DateTime GetFutureContractMonthNoRulesApplied(Symbol canonicalFutureSymbol, DateTime futureOptionExpirationDate)
```
Martin-Molinero, line 200: nit - let's make this method private

Common/Securities/Future/FuturesListings.cs
```diff
@@ -0,0 +1,225 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Example:
+        ///
+        ///   (from: https://www.cmegroup.com/trading/agricultural/grain-and-oilseed/wheat_contract_specifications.html)
+        ///   "15 monthly contracts of Mar, May, Jul, Sep, Dec listed annually following the termination of trading in the July contract of the current year."
+        ///
+        /// This would equate to a cycle of [3, 5, 7, 9, 12], a limit of 15, and the contract month == 7.
+        /// </remarks>
+        public class FuturesListingCycles
```
Martin-Molinero, line 200: nit - can be private 👍

Common/Securities/Future/FuturesListings.cs
```diff
@@ -0,0 +1,225 @@
...
+namespace QuantConnect.Securities.Future
+{
+    /// <summary>
+    /// Helpers for getting the futures contracts that are trading on a given date.
+    /// This is a substitute for the <see cref="BacktestingFutureChainProvider"/>, but
+    /// does not outright replace it because of missing entries. This will resolve
+    /// the listed contracts without having any data in place. We follow the listing rules
+    /// set forth by the exchange to get the <see cref="Symbol"/>s that are listed at a given date.
+    /// </summary>
+    public static class FuturesListings
```
Martin-Molinero, line 29: Nice one 💯

Common/Securities/Future/FuturesListings.cs
```diff
@@ -0,0 +1,225 @@
...
+                quarterlyContracts[i] = firstFutureContractMonth.AddMonths(-firstFutureContractMonth.Month + quarterlyContractMonth);
+                quarterlyContractMonth += 3;
+            }
+
+            return quarterlyContracts
+                .OrderBy(t => t)
+                .Select(t => FuturesExpiryFunctions.FuturesExpiryFunction(canonicalFuture)(t))
+                .Where(e => e >= time)
+                .Select(e => Symbol.CreateFuture(canonicalFuture.ID.Symbol, canonicalFuture.ID.Market, e))
+                .ToArray();
```
Martin-Molinero, line 116: We could directly perform these operations in the loop above and avoid iterating twice, same for `MonthlyContractListings`, and we shouldn't need the order by?

Martin-Molinero (approved):
Looks good, leaving minor comments 👍

Brokerages/InteractiveBrokers/InteractiveBrokersSymbolMapper.cs
gsalaz98: It still can, if we request an expiry for a future that hasn't been listed (think: expiry super far ahead into the future) or does not have a FuturesListing rule whenever it hits the Serial contracts lookup in the FuturesOptionsUnderlyingMapper, though that should never happen.

Martin-Molinero (approved):

PR #5158, opened 2021-01-14
Refactor Default Benchmark

### LINKED ISSUE #5143: Make algorithm benchmark per brokerage model
#### Expected Behavior

- Default benchmark does not cause live trading issues
#### Actual Behavior

- Since the default benchmark is SPY it causes live trading issue when trading with crypto brokerages or generally brokerages for which the user doesn't have a SPY subscription (like IB)
#### Potential Solution

- Make default algorithm benchmark part of the brokerage model. Implement accordingly for the different models
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

C-SELLERS:
#### Description

Have Brokerage models set the default benchmark for their brokerage.

#### Related Issue

Closes #5143 

#### Motivation and Context

> Since the default benchmark is SPY it causes live trading issue when trading with crypto brokerages or generally brokerages for which the user doesn't have a SPY subscription (like IB)

Instead of using a default benchmark for all brokerages this allows a unique benchmark to be used by any brokerage implementation through overriding `GetBenchmark()` in any `IBrokerageModel`. All it needs to do is return a `IBenchmark` and it will be set for the algorithm. 

Any brokerage model that does not override the function will take on the default benchmark (SPY).

All regressions pass as expected.
Added a new regression to test the case of 0 value benchmark for C# and Python

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -558,17 +558,23 @@ namespace QuantConnect.Algorithm
             // if the benchmark hasn't been set yet, set it
             if (Benchmark == null)
             {
+                // If we don't have a symbol fetch it from the Brokerage model
                 if (_benchmarkSymbol == null)
                 {
-                    _benchmarkSymbol = QuantConnect.Symbol.Create("SPY", SecurityType.Equity, Market.USA);
+                    _benchmarkSymbol = BrokerageModel.DefaultBenchmark;
                 }
 
-                var security = Securities.CreateSecurity(_benchmarkSymbol,
-                    new List<SubscriptionDataConfig>(),
-                    leverage: 1,
-                    addToSymbolCache:false);
+                // Before we set it, ensure the benchmark symbol is not null
+                // Brokerage implementations may set it to null to have no benchmark
+                if (_benchmarkSymbol != null)
+                {
+                    var security = Securities.CreateSecurity(_benchmarkSymbol,
+                        new List<SubscriptionDataConfig>(),
+                        leverage: 1,
+                        addToSymbolCache: false);
 
-                Benchmark = new SecurityBenchmark(security);
+                    Benchmark = new SecurityBenchmark(security);
+                }
             }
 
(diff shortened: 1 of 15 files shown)


C-SELLERS (commented):
This works as it is right now but it is not smooth yet,  

An alternative option I am considering is having the the creation of the `IBenchmark` handled inside of the `IBrokerageModel` at a function like `GetBenchmark` so that it can be overridden by its inheritors.

Then we can have the case where if `_benchmarkSymbol == null` we can just use `Benchmark = BrokerageModel.GetBenchmark`

then the else is using the already defined `_benchmarkSymbol` as it is right now, creating the symbol, security, and then function.

jaredbroad (approved):

C-SELLERS:
Sorry for two force pushes, had to rebase twice. The first one to get the new csproj format and then the second because master was broken from a different issue after the first time.

Its all fixed now though so no more rebases are necessary.

C-SELLERS (commented):
Ready for review 👍🏽

Algorithm/QCAlgorithm.cs
```diff
@@ -1106,7 +1087,15 @@ public void SetBenchmark(Symbol symbol)
...
-            _benchmarkSymbol = symbol;
+
+            // Create a security from this symbol
+            var security = Securities.CreateSecurity(symbol,
+                new List<SubscriptionDataConfig>(),
+                leverage: 1,
+                addToSymbolCache: false);
+
+            // Create and set our SecurityBenchmark
+            Benchmark = new SecurityBenchmark(security);
```
Martin-Molinero, line 1098: Nice! 👍

Martin-Molinero (approved):
Looks good minor findings 👍

C-SELLERS (commented):
Review addressed 👍🏽 Thinking I might make a Python regression that matches the C# one for the sake of testing the wrapper.

Martin-Molinero (approved):
Nice work! 💯

PR #5177, opened 2021-01-15
bug-4954  Polygon Data Downloader : Added Quote & Trade Tick Downloads with SIP Mapping

TightLinesOfCode:
#### Description
Updated the language version in each project to be the LATEST C# version.  This was needed for coalescing null values
SIP Mapping logic needs work, but now adds Row Conditions and uses Suspicious Flag.
Trades and Quotes are now working
Added x64 for projects to eliminate out of memory errors
Added parallel foreach and commented out in downloader because Tick classes aren't thread safe on dictionary inserts
Extended HistoryRequest.cs for polygon parameters to flow through instead of config.json lookup which would vastly slow down the system
Extended Tick.cs for polygon specific parameters related to faster downloads

#### Motivation and Context
Added better logic and SIP mapping for polygon downloader

No documentation changes necessary

Ran this many times and downloaded all of 2020 data and then verified the results manually

#### Types of changes
- [ ] Refactor (non-breaking change which improves implementation)     Refactored the json downloader to handle null json values

#### Checklist:
- [X] I have added tests to cover my changes.   Just fixed the bug and verified the data


jaredbroad:
Hey @TightLinesOfCode, thank you for the PR. We'll need to do substantial rework but can guide you through if you're up to it.

To start; strip all changes not in /Toolbox or  ToolBox/Polygon. Try and think of yourself as a heart surgeon, not a carpetener. Given the risks of interfering with live trading, we make all changes step by step.

Tick data is raw in lean; the filtering is at your algorithm level, not the infrastructure level. In separate PR's we'll need to review and discuss the addition of new properties and ensuring they're relevant and well-integrated first. 

Anything not specifically Polygon related needs to be separate PRs.

Issue #5181, opened 2021-01-17
Set x64 as default build configuration for all projects on C# solution


TightLinesOfCode:
#### Expected Behavior
x64 is set as the default build configuration

#### Actual Behavior
ANYCPU with the default 32 bit build is configured

#### Potential Solution
Change configuration, and also fix the output directories so that the data directory and any hard coded directories paths are still valid.   C# will put the build into a x64 directory, so that need to fixed to match the ANYCPU build

#### Reproducing the Problem
Run the code from the current master branch

#### System Information
C#

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

PR #5145, opened 2021-01-08
CSProj Update

C-SELLERS:
#### Description

Update all CSProj files to the new format

Includes:
- Migration from `packages.config` to `PackageReference` -> Reduces work of managing packages to a single line per package in proj file.
- Migration to project sdk `Microsoft.NET.Sdk` -> Moves us to .Net standard proj format, allowing us to target more frameworks in the future as well as cleans up project files significantly. This is the largest change by far as it impacts the entire structure of the proj files. Including removal of a lot of default settings to reduce clutter.
- Compile globbing -> instead of relying on compile include statements all files in the project are by default compiled, to not compile we must use `Compile Remove`
- Reduction of certain build settings into a overhead default that utilizes dynamic vars such as `$(Configuration)` to further reduce clutter.
- Use `$(NugetPackageRoot)` variable instead of relative pathing for specific dll finds. -> With the change to packages, the storage is no longer in ./packages, so we needed to have a dynamic path variable to address this.
- Test project now includes `Microsoft.NET.Test.Sdk` allowing for `dotnet test` as well as supporting testing in VS16 project (new version this PR moves us too)
- Open some internal test classes to public for composer use (There was issues in linux testing regarding accessing these classes)

#### Related Issue

Closes #1686

#### Motivation and Context

To move forward with supporting dotnet core we need to update our project files to the new .net standards.

#### Requires Documentation Change

N/A

#### How Has This Been Tested?

- dotnet build works
- msbuild works
- All regressions pass locally on Windows 10
- Travis test set runs successfully in Docker container and on Ubuntu 20.04
- C# and Py Live trading tested in cloud and locally*
- C# and Py Backtesting tested in cloud and locally*
- C# and Py Research tested in cloud and locally*
- C# and Py Debugging tested in cloud and locally*

*locally refers to windows 10 environment and cloud refers to QuantConnect.com

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
[...]

PR #5182, opened 2021-01-18
feature-5181-set x64 as default build configuration for all projects on c# solution

TightLinesOfCode:
#### Description
Added the x64 solution configuration and also adjusted each project to have x64 for build.
Removed the ANYCPU configuration from solution to enhance consistency for users downloading the software

#### Related Issue
https://github.com/QuantConnect/Lean/issues/5181

#### Motivation and Context
Solves out of memory issues related to massively parallel downloads and lean data writer.   Also enhances the speed of the overall application by using a 64 bit address space

None

-Built Solution with Debug and Release and verified output folders to be the same.
-Ran the default Lean Launcher project with no errors in Debug and Release mode.

#### Types of changes

- [X] Performance (non-breaking change which improves performance. Please add associated performance test and results)

#### Checklist:
- [NA] I have added tests to cover my changes. 
- [no] All new and existing tests passed.    ( Default lean repository does not pass all of the unit tests )


Martin-Molinero:
Hey @TightLinesOfCode! Thanks for the PR, interesting
> Solves out of memory issues related to massively parallel downloads and lean data writer. Also enhances the speed of the overall application by using a 64 bit address space

Mind expanding **why**, references are welcome and believe we should also test the statements somehow.
I don't think there is a performance difference really. Compiled code should be the same. See https://stackoverflow.com/questions/11062028/performance-of-a-c-sharp-application-built-on-anycpu-vs-x64-platform-on-a-64-bit/11062102#11062102 or https://stackoverflow.com/questions/516730/what-does-the-visual-studio-any-cpu-target-mean/516740#516740 or https://medium.com/@trapdoorlabs/c-target-platforms-x64-vs-x86-vs-anycpu-5f0c3be6c9e2

Thanks to the PR, I'm realizing though that after the recent change https://github.com/QuantConnect/Lean/pull/5145 the `Platform target` for the launcher/toolbox/report in master is `x86` for the AnyCpu/debug or release build configurations. While previously it was `AnyCpu`. We will fix this ASAP @C-SELLERS.

TightLinesOfCode:
@Martin-Molinero  @C-SELLERS    One of the issues I ran into is the massively parallel downloads of data which easily grow larger than the small address space of 32-bit compilation.   Also, for speed increases, the 64-bit will have a dramatic increase later when the the code is refactored for speed.   This becomes especially true for the back testing engine when a user might want to keep 8000 tickers in memory for large scale AI pipelines and computation.

But generally, since 32-bit compilation is for backward compatibility, it doesn't really make sense to have it for high performance code.

jaredbroad:
> But generally, since 32-bit compilation is for backward compatibility, it doesn't really make sense to have it for high-performance code.

I think @Martin-Molinero 's point is specifically AnyCPU runs in x64 when it's available, and the research above points to that there is no speed up as it's already running in x64 when it's available.

TightLinesOfCode:
The platform target is set to x86 and the main problem is the addressable memory space.    For large data analysis, the x86 default will cause a problem.       It's not just about speed, except when windows has to swap memory out from disk.

Issue #4274, opened 2020-04-03
Test stat OrderListHash is non deterministic


Martin-Molinero:
#### Expected Behavior

- OrderListHash  is deterministic
#### Actual Behavior

- OrderListHash uses `string.GetHashCode()` which is not deterministic since it relies on the implementation that could change from one runtime version to another. https://docs.microsoft.com/en-us/dotnet/api/system.string.gethashcode?view=netcore-1.1#System_String_GetHashCode
#### Potential Solution

- Use MD5 hash instead which is deterministic (or another)
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #5192, opened 2021-01-19
Setup GitHub Actions CI

itsbalamurali:
#### Description

Setup and use GitHub Actions CI along with Travis CI.

#### Motivation and Context

Setup and use GitHub Actions CI along with Travis CI.

#### Types of changes

#### Checklist:


itsbalamurali:
@Martin-Molinero any idea on the seg fault?

Martin-Molinero:
> 
> 
> @Martin-Molinero any idea on the seg fault?

Sounds like some python configuration is wrong 🤔, seems affected by https://github.com/QuantConnect/Lean/issues/4274 too.

Martin-Molinero:
**What if instead of setting the environment up you used Lean docker foundation image, might be possible https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer. Avoid any setup will be faster too 🚀**. `quantconnect/lean:foundation` https://hub.docker.com/repository/docker/quantconnect/lean



> @Martin-Molinero any idea on the seg fault?

hm it might be using mono 6.12? Suggest adding a `mono --version` print comment at the end of the mono setup

image

itsbalamurali:
@Martin-Molinero currently the Docker image is being pulled Docker Hub which is terribly slow (8mins) instead i suggest that we mirror Docker images to GitHub Packages (I can raise and another P.R for it to build on git tag and master branch and push to GitHub packages) which would cut down pipeline time.

itsbalamurali:
@Martin-Molinero & @jaredbroad i think we are good for merging this P.R

Martin-Molinero (approved):
This is great! 🚀 
GitHub also has an action cache https://github.com/actions/cache, but it should be cleared when we update the foundation for example, can't be cached for ever.
Think this is a great first step 💯

.github/workflows/gh-actions.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches: ['*']
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ubuntu-16.04
```
Martin-Molinero, line 11: Wonder why `ubuntu-16.04` seems like an old image?

itsbalamurali: That's what is being used on Travis CI (Since, I was having issues with latest Mono runtime). I had choose this one to try out.

PR #5194, opened 2021-01-19
Add primary exchage to map files

Jay-Jay-D:
#### Description

This PR incorporates the primary exchange of each equity in its map file.
Also adds a `PrimaryExchange` property into the `Equity:Security` implementation.

#### Related Issue

#5129 

#### Motivation and Context

The motivation is to ensure that all resolutions fill Market On Open orders at the same price. 

All resolutions - except Tick - have the auction opening price in the first bar after the market's open. In tick resolution, the first trade tick after the market's open is never the tick with the auction opening price. It usually appears after a few milliseconds for high volume securities.

On the other hand, it is common to receive multiple auction opening ticks, one for each exchange where the security trades.

To correctly get the trade tick, we need to get the tick trade with the auction opening price from the security's primary exchange. That is the reason we need to know the security's primary exchange.

N/A

Adding new tests for the new features and all test passed to check backward compatibility.

#### Types of changes

#### Checklist:


Common/Data/Auxiliary/MapFileRow.cs
```diff
@@ -29,20 +30,36 @@ public class MapFileRow : IEquatable<MapFileRow>
...
+        public PrimaryExchange PrimaryExchange { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MapFileRow"/> class.
+        /// </summary>
+        public MapFileRow(DateTime date, string mappedSymbol, char primaryExchange=' ')
+        {
+            Date = date;
+            MappedSymbol = mappedSymbol.LazyToUpper();
+            PrimaryExchange = Exchanges.GetPrimaryExchange(primaryExchange);
```
Martin-Molinero, line 52: Cool!
Let's make this constructor call the other one 👍

Common/Data/Auxiliary/MapFileRow.cs
```diff
@@ -70,7 +87,14 @@ public static IEnumerable<MapFileRow> Read(string path)
...
             var csv = line.Split(',');
-            return new MapFileRow(DateTime.ParseExact(csv[0], DateFormat.EightCharacter, null), csv[1]);
+            var primaryExchange = PrimaryExchange.UNKNOWN;
+            if (csv.Length == 3)
+            {
+                primaryExchange = Exchanges.GetPrimaryExchange(Convert.ToChar(csv[2], CultureInfo.InvariantCulture));
+            }
+
+            return new MapFileRow(DateTime.ParseExact(csv[0], DateFormat.EightCharacter, null), csv[1],
+                Convert.ToChar((byte) primaryExchange));
```
Martin-Molinero, line 97: Could we skip the new constructor if we just passed in `primaryExchange` directly?

Common/Data/Market/Tick.cs
```diff
@@ -47,10 +49,37 @@ public class Tick : BaseData
         public decimal Quantity = 0;
 
         /// <summary>
-        /// Exchange we are executing on. String short code expanded in the MarketCodes.US global dictionary
+        /// Exchange code this tick came from <see cref="Exchanges"/>
         /// </summary>
         [ProtoMember(12)]
```
Martin-Molinero, line 54: > [ProtoMember(12)]

Should still be on the `string Exchange` for now

Common/Data/Market/Tick.cs
```diff
@@ -47,10 +49,37 @@ public class Tick : BaseData
...
+        public byte ExchangeCode
+        {
+            get
+            {
+                return _exchangeCode;
+            }
+            set
+            {
+                _exchangeCode = value;
+                _exchange = ((PrimaryExchange) value).ToString();
```
Martin-Molinero, line 64: > _exchange = ((PrimaryExchange) value).ToString();

Cool! let's add a test where value isn't in `PrimaryExchange` enum, think we should handle this case 👍

Common/Global.cs
```diff
@@ -23,6 +25,33 @@
 
 namespace QuantConnect
 {
+    public enum PrimaryExchange : byte
+    {
+        UNKNOWN=32,
```
Martin-Molinero, line 30: Curious why just not 0, empty char vs `' '` space char?
Let's add some xml doc for the new enum 👍

Common/Global.cs
```diff
@@ -679,40 +708,107 @@ public enum DataNormalizationMode
...
+        }
+
+        /// <summary>
+        /// Gets the exchange name for the given code number
+        /// </summary>
+        /// <remarks>Useful for performance</remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static PrimaryExchange GetName(int exchangeCode)
+        {
+            return (PrimaryExchange) exchangeCode;
```
Martin-Molinero, line 810: Let's just remove this guy if unused 👍

Common/MapFilePrimaryExchangeProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect
+{
+    /// <summary>
+    /// Implementation of IPrimaryExchangeProvider from map files. 
+    /// </summary>
+    class MapFilePrimaryExchangeProvider : IPrimaryExchangeProvider
```
Martin-Molinero, line 26: Cool!
Maybe we can move this guy into `Data/Auxiliary` with the other map file classes

Common/MapFilePrimaryExchangeProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        private readonly Dictionary<SecurityIdentifier, PrimaryExchange> _primaryExchangeBySid;
+        public MapFilePrimaryExchangeProvider(IMapFileProvider mapFileProvider, string market = "USA")
+        {
+            _primaryExchangeBySid = new Dictionary<SecurityIdentifier, PrimaryExchange>();
+            foreach (var mapFile in mapFileProvider.Get(market))
+            {
+                var sid = SecurityIdentifier.GenerateEquity(mapFile.FirstDate, mapFile.FirstTicker, market);
+                var primaryExchange = (PrimaryExchange) mapFile.Last().PrimaryExchange;
+                _primaryExchangeBySid[sid] = primaryExchange;
```
Martin-Molinero, line 36: Think we should do this on demand when queried, based on the provided SID market, so we avoid doing this if not trading equities. We might even just cache on demand into the `_primaryExchangeBySid` so we don't hold all of the symbols in it always. What do you think?

Common/Securities/Equity/Equity.cs
```diff
@@ -60,6 +60,11 @@ public bool Shortable
         /// <returns>Zero if not shortable, null if infinitely shortable, or a number greater than zero if shortable</returns>
         public long? TotalShortableQuantity => ShortableProvider.ShortableQuantity(Symbol, LocalTime);
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public PrimaryExchange PrimaryExchange;
```
Martin-Molinero, line 66: nit - Missing docs

Common/Securities/SecurityService.cs
```diff
@@ -130,7 +133,12 @@ public class SecurityService : ISecurityService
             switch (symbol.ID.SecurityType)
             {
                 case SecurityType.Equity:
-                    security = new Equity.Equity(symbol, exchangeHours, quoteCash, symbolProperties, _cashBook, _registeredTypes, cache);
+                    var primaryExchange = PrimaryExchange.UNKNOWN;
+                    if (_primaryExchangeProvider?.GetPrimaryExchange(symbol.ID) != null)
+                    {
+                        primaryExchange = (PrimaryExchange)Enum.Parse(typeof(PrimaryExchange), _primaryExchangeProvider.GetPrimaryExchange(symbol.ID));
```
Martin-Molinero, line 139: Suggest we push this logic into the `_primaryExchangeProvider` implementation instead of returning `null`, will make for a better usage 👍

Tests/Common/Data/Auxiliary/MapFileTests.cs
```diff
@@ -81,5 +82,64 @@ public void GenerateMapFileCSV()
...
+            var rowParts = mapFileRow.Split(',');
+            var expectedMapFileRow = new MapFileRow(
+                DateTime.ParseExact(rowParts[0], DateFormat.EightCharacter, CultureInfo.InvariantCulture),
+                rowParts[1]);
+            // Act
+            var actualMapFileRow = MapFileRow.Parse(mapFileRow);
+            // Assert
+            Assert.AreEqual(PrimaryExchange.UNKNOWN, actualMapFileRow.PrimaryExchange);
+            Assert.AreEqual(expectedMapFileRow, actualMapFileRow);
+        }
```
Martin-Molinero, line 143: Sweet!

Martin-Molinero (commented):
Nice work! Leaving a few requests here and there. Also let's add a couple of simple unit tests for the `Tick.cs` class changes 👍

Common/Global.cs
Jay-Jay-D: >Curious why just not 0, empty char vs ' ' space char?
Good question, I think it's was more a matter of taste, the closes we can get to a empty char is `'\0'`, the `null char`.
Seems to me that space char was more clear than the null char as default value for the  `MapFileRow `constructor.

Martin-Molinero (approved):

Martin-Molinero (commented):
Looks good 👍  but didn't expect regression algorithms statistics to change, maybe `OrderListHash` if the ticker changed. but the other statistics related to the performance of the algorithm why?

Martin-Molinero (approved):

Issue #4954, opened 2020-11-22
PolygonDataQueueHandler.cs  not fully implemented


TightLinesOfCode:
#### Expected Behavior
A standard data model for equity ticks and a supporting architecture that is not vendor specific

#### Actual Behavior
The tick class is not supporting the full SIP standard.

#### Potential Solution
Map a proper data model for equity ticks and bars that maps to SIP.    Determine logic for suspicious ticks so that if this used in the LEAN algorithm it will work for any tickers.

#### Reproducing the Problem
Basic tick creation doesn't have room for the exchange timestamp

#### System Information
C#

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

TightLinesOfCode:
Also, the conditions mappings and exchange were not fully implemented in the PolygonDataQueueHandler.cs

yield return new Tick(time, request.Symbol, string.Empty, string.Empty, row.BidSize, row.BidPrice, row.AskSize, row.AskPrice);

TightLinesOfCode:
I forked and branch and will be working on this

TightLinesOfCode:
@Martin-Molinero    Martin, can you help me with this one question regarding the sale conditions ?       I need to know if your existing vendor is using multiple sale conditions, or just one.    The other equity data feeds have multiple sale conditions from the SIP Processing.     I need to determine the scope of the code I am going to fix.   I already have all of the Polygon APIs built in a multithreaded, api throttled architecture and am working on adding this to the LEAN code. 

I also have a RocksDB implementation I can add, but I really need to figure out scope

jaredbroad:
Hi @TightLinesOfCode, tick data is not filtered in LEAN. We provided in a raw form and advise only use for experts who know how to interpret those fields themselves. Please implement the Polygon IQDH to not filter any ticks.  

Please remember a nice separation of PRs; if wanting to contribute a RocksDB please make it a separate issue and PR. 

Please also respect the contributing guidelines and templates and create complete fully filled out issues in the future.

TightLinesOfCode:
[...]

PR #5204, opened 2021-01-21
Feature 4954 polygon data queue handler.cs not fully implemented

TightLinesOfCode:
#### Description
-Added correct API for dates and iteration through requested dates
-Double check NULL results before downloading data
-Added error handling for 404 errors from Polygon.IO for weekend and non business dates
-Added logic to deal with NULL results from Holidays
-Added quote conditions and trade sale conditions
-Added naive suspicious trade & quote logic
-Fixed incorrect API logic for pagination results
-Hard coded a 10,000 result limit
-Hard coded a service point connection limit of 4, for open API connections

#### Related Issue
https://github.com/QuantConnect/Lean/issues/4954 

#### Motivation and Context
Bugs in the Polygon Quote & Trade Tick downloader

No changes

Ran successfully over entire 2020 tickers for multiple tickers.
Verified Nanosecond Tick times and Prices to ensure correct behavior
Verified UTC time converted to DATA time correctly

#### Types of changes

#### Checklist:
- [NA] I have added tests to cover my changes.     Only changes to the tick download and this was verified manually

### DIFF
--- a/ToolBox/Polygon/PolygonDataQueueHandler.cs
+++ b/ToolBox/Polygon/PolygonDataQueueHandler.cs
@@ -525,26 +525,20 @@ namespace QuantConnect.ToolBox.Polygon
                 // Skips Saturday and Sunday
                 if (!start.IsCommonBusinessDay())
                 {
-#if DEBUG
-                    Log.Trace("Weekends are not processed " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
-#endif
+                    Log.Debug("Weekends are not processed " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
                     start = start.Date.AddDays(1);
                     break;
                 }
                 else
                 {
-#if DEBUG
-                    Log.Trace("Processing " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
-#endif
+                    Log.Debug("Processing " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
                 }
 
                 using (var client = new WebClient())
                 {
                     var url = $"{HistoryBaseUrl}/v2/ticks/stocks/nbbo/{request.Symbol.Value}/{start.Date:yyyy-MM-dd}?apiKey={_apiKey}&timestamp={offsetStart}&limit={_apiResultsLimit}";
 
-#if DEBUG
-                    Log.Trace(url);
-#endif
+                    Log.Debug(url);
 
                     string response = "";
                     try
@@ -554,16 +548,14 @@ namespace QuantConnect.ToolBox.Polygon
                     catch (WebException e)
                     {
                         isException = true;
-                        offsetStart = offsetEnd;
                     }
 
-                    if ((response != "") && (isException == false) && (response != null))
+                    if ((!response.IsNullOrEmpty()) && (isException == false))
                     {
                         var obj = JObject.Parse(response);
                         var objTicks = obj["results"];
                         if ((objTicks.Type == JTokenType.Null) || (!objTicks.Any()))
                         {
-
                             offsetStart = offsetEnd;   // Set offsetStart end of the time to Finish the while loop
                         }
                         else
@@ -603,7 +595,7 @@ namespace QuantConnect.ToolBox.Polygon
                                 {
                                     quoteConditions = string.Join("", row.Conditions.Select(p => String.Format(System.Globalization.CultureInfo.GetCultureInfo("en-US"), "{0:X2}", p)));
                                 }
-                                // Adds the exchange the Tick.   Prefers the bid exchange and assumes mutually exclusive bid/ask exchange IDs on quote
+                                // Adds the exchange to the Tick.   Prefers the bid exchange and assumes mutually exclusive bid/ask exchange IDs on quote
                                 string exchange = "";
                                 if (row.BidExchange != null)
                                 {
@@ -648,27 +640,21 @@ namespace QuantConnect.ToolBox.Polygon
                 // Skips Saturday and Sunday
                 if (!start.IsCommonBusinessDay())
                 {
-#if DEBUG
-                    Log.Trace("Weekends are not processed " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
-#endif
+                    Log.Debug("Weekends are not processed " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
                     start = start.Date.AddDays(1);
                     break;
                 }
                 else
                 {
-#if DEBUG
-                    Log.Trace("Processing " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
-#endif
+                    Log.Debug("Processing " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
                 }
 
                 using (var client = new WebClient())
                 {
                     var url = $"{HistoryBaseUrl}/v2/ticks/stocks/trades/{request.Symbol.Value}/{start.Date:yyyy-MM-dd}?apiKey={_apiKey}&timestamp={offsetStart}&limit={_apiResultsLimit}";
 
-#if DEBUG
-                    Log.Trace(url);
-#endif
-                    
+                    Log.Debug(url);
+
                     string response = "";
                     try
                     {
@@ -720,7 +706,7 @@ namespace QuantConnect.ToolBox.Polygon
                                 }
 
                                 // Convert Sale Conditions to Hex and concatenate
-                                string saleConditions = "";
+                                var saleConditions = "";
                                 if (row.Conditions != null)
                                 {
                                     saleConditions = string.Join("", row.Conditions.Select(p => String.Format(System.Globalization.CultureInfo.GetCultureInfo("en-US"), "{0:X2}", p)));
--- a/ToolBox/Polygon/PolygonDataQueueHandler.cs
+++ b/ToolBox/Polygon/PolygonDataQueueHandler.cs
@@ -594,11 +594,6 @@ namespace QuantConnect.ToolBox.Polygon
                                     _suspiciousFlag = (PolygonQuoteMappings.isSuspiciousQuote(row.Conditions[0]));
                                 }
 
-#if DEBUG
-                                Log.Trace($"Results count for tickers {objTicks.Count()}");
-#endif
-
-
                                 string quoteConditions = "";
                                 if (row.Conditions != null)
                                 {
@@ -662,6 +657,10 @@ namespace QuantConnect.ToolBox.Polygon
                 {
                     var url = $"{HistoryBaseUrl}/v2/ticks/stocks/trades/{request.Symbol.Value}/{start.Date:yyyy-MM-dd}?apiKey={_apiKey}&timestamp={offsetStart}&limit={_apiResultsLimit}";
 
+#if DEBUG
+                    Log.Trace(url);
+#endif
+                    
                     string response = "";
                     try
                     {
--- a/ToolBox/Polygon/PolygonDataQueueHandler.cs
+++ b/ToolBox/Polygon/PolygonDataQueueHandler.cs
@@ -48,6 +48,7 @@ namespace QuantConnect.ToolBox.Polygon
         private const string HistoryBaseUrl = "https://api.polygon.io";
 
         private readonly string _apiKey = Config.Get("polygon-api-key");
+        private readonly int _apiResultsLimit = 10000;
 
         private readonly IDataAggregator _dataAggregator = Composer.Instance.GetExportedValueByTypeName<IDataAggregator>(
             Config.Get("data-aggregator", "QuantConnect.Lean.Engine.DataFeeds.AggregationManager"));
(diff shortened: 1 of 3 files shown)


TightLinesOfCode:
@jaredbroad    Is this code ok ?

jaredbroad:
👋 Hey @TightLinesOfCode, do you mind reviewing this PR again in light of this code (https://github.com/QuantConnect/Lean/pull/5227/files).

This way the Polygon ticks can match LEAN ones 🥇

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
                     {
-                        var row = objTick.ToObject<EquityQuoteTickResponse>();
-
-                        var utcTime = Time.UnixNanosecondTimeStampToDateTime(row.ExchangeTimestamp);
+                        isException = true;
+                        offsetStart = offsetEnd;
+                    }
 
-                        if (utcTime < start)
+                    if ((response != "") && (isException == false) && (response != null))
```
Martin-Molinero, line 560: `&& (response != null)` will never be true

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+                        response = client.DownloadString(url);
                     }
-
-                    foreach (var objTick in objTicks)
+                    catch (WebException e)
                     {
-                        var row = objTick.ToObject<EquityQuoteTickResponse>();
-
-                        var utcTime = Time.UnixNanosecondTimeStampToDateTime(row.ExchangeTimestamp);
+                        isException = true;
```
Martin-Molinero, line 556: Suggest logging the exception here as an error, `offsetStart = offsetEnd;` isn't required

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+                                {
+                                    quoteConditions = string.Join("", row.Conditions.Select(p => String.Format(System.Globalization.CultureInfo.GetCultureInfo("en-US"), "{0:X2}", p)));
+                                }
+                                // Adds the exchange the Tick.   Prefers the bid exchange and assumes mutually exclusive bid/ask exchange IDs on quote
+                                string exchange = "";
+                                if (row.BidExchange != null)
+                                {
+                                    exchange = row.BidExchange.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US"));
+                                }
+                                else if (row.AskExchange != null)
```
Martin-Molinero, line 612: `row.BidExchange` & `row.AskExchange` can't ever be null they are defined as an int

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+                                {
+                                    exchange = row.AskExchange.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US"));
```
Martin-Molinero, line 614: We should use the exchanges defined by https://github.com/QuantConnect/Lean/pull/5194/files at https://github.com/QuantConnect/Lean/blob/master/Common/Global.cs#L37

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
                 {
-                    var offset = Time.DateTimeToUnixTimeStampNanoseconds(start);
-                    var url = $"{HistoryBaseUrl}/v2/ticks/stocks/trades/{request.Symbol.Value}/{start.Date:yyyy-MM-dd}?apiKey={_apiKey}&timestamp={offset}";
-
-                    var response = client.DownloadString(url);
+#if DEBUG
+                    Log.Trace("Weekends are not processed " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
+#endif
+                    start = start.Date.AddDays(1);
+                    break;
```
Martin-Molinero, line 655: Similar comment about `start` and replacement of `#if DEBUG` 👍

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+                                // Look for suspicious trades.   In this case, it uses the SIP mapping for updates of high/low otherwise marks as suspicious
+                                bool suspiciousFlag = false;
+                                // Check SIP Mappings for update high/low and mark as suspicious if not update/high low.    This needs more sophistication
+                                if (row.Conditions != null)
+                                {
+                                    suspiciousFlag = !Polygon.PolygonSIPTradeMappings.TradeUpdateHighLow(row.Conditions[0]);
+                                }
+
+                                // Convert Sale Conditions to Hex and concatenate
+                                string saleConditions = "";
```
Martin-Molinero, line 723: nit - code style, we usually use `var` all around

ToolBox/Polygon/PolygonDownloaderProgram.cs
```diff
@@ -54,35 +58,50 @@ public static void PolygonDownloader(IList<string> tickers, string securityTypeS
...
-
+                        
+                        // Get the correct time zones
                         var exchangeTimeZone = marketHoursDatabase.GetExchangeHours(market, symbol, securityType).TimeZone;
                         var dataTimeZone = marketHoursDatabase.GetDataTimeZone(market, symbol, securityType);
 
-                        foreach (var tickType in tickTypes)
+                        // Set the UTC times for correctness of the architecture
+                        var utcStartDate = fromDate.ConvertToUtc(dataTimeZone);  // TimeZones.NewYork);
+                        var utcEndDate = toDate.ConvertToUtc(dataTimeZone);  // TimeZones.NewYork);
```
Martin-Molinero, line 81: nit - commented code `// TimeZones.NewYork);`

ToolBox/Polygon/PolygonDownloaderProgram.cs
```diff
@@ -54,35 +58,50 @@ public static void PolygonDownloader(IList<string> tickers, string securityTypeS
...
+                                        {
+                                            x.Time = x.Time.ConvertTo(exchangeTimeZone, dataTimeZone);
+                                            return x;
+                                        }
+                                    );
 
-                            // Save the data
-                            var writer = new LeanDataWriter(resolution, symbol, dataDirectory, tickType);
-                            writer.Write(data);
+                                if ((data != null))
```
Martin-Molinero, line 98: data will never be null? might be empty though, suggest adding `ToList()` after the select and checking `count` here instead

ToolBox/Polygon/PolygonDownloaderProgram.cs
```diff
@@ -54,35 +58,50 @@ public static void PolygonDownloader(IList<string> tickers, string securityTypeS
...
-                var endDate = toDate.ConvertToUtc(TimeZones.NewYork);
 
+                // Get hours
                 var marketHoursDatabase = MarketHoursDatabase.FromDataFolder();
 
                 // Create an instance of the downloader
                 using (var downloader = new PolygonDataDownloader())
                 {
+                    // Server side parameters for multithreaded simultaneous API connections
+                    ServicePointManager.DefaultConnectionLimit = 4;
```
Martin-Molinero, line 69: > ServicePointManager.DefaultConnectionLimit = 4;

Don't think this is required because we are not requesting in parallel?

ToolBox/Polygon/PolygonDownloaderProgram.cs
```diff
@@ -54,35 +58,50 @@ public static void PolygonDownloader(IList<string> tickers, string securityTypeS
...
-                                        return x;
-                                    }
-                                );
+                            foreach (var tickType in tickTypes)
+                            {
+                                // Download the data
+                                var data = downloader.Get(symbol, resolution, utcStartDate.AddDays(i), utcEndDate.AddDays(i + 1), tickType)
+                                    .Select(x =>
+                                        {
+                                            x.Time = x.Time.ConvertTo(exchangeTimeZone, dataTimeZone);
```
Martin-Molinero, line 93: cool! 👍

Martin-Molinero (commented):
Welcome to Lean! Leaving a few comments

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: @Martin-Molinero Sometimes the response is null.   This is a peculiarity of the Polygon API and has to be here to deal with a null response that can't be parsed by the JSON parser.

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: @Martin-Molinero The Polygon API is NULL for the exchange since it separates the bid quote from the ask quote and can be different exchanges.   My understanding is that the LEAN data model combines these two into one field called Exchange

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: will fix

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: Yes, we need to request in Parallel to increase the bandwidth throughput.   The Polygon servers allow for 100 concurrent connections.

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+            while (offsetStart < offsetEnd)
             {
+                // Skips Saturday and Sunday
+                if (!start.IsCommonBusinessDay())
+                {
+#if DEBUG
+                    Log.Trace("Weekends are not processed " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
+#endif
+                    start = start.Date.AddDays(1);
+                    break;
```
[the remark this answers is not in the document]
TightLinesOfCode: The offset is used in the Pagination for the API and this design pattern is specific to the Polygon API behavior.   This code iterates through each day at a time and the start offset is only for the pagination.

You can see from the REST endpoint that it is based on a day.    https://polygon.io/docs/get_v2_ticks_stocks_trades__ticker___date__anchor

This is one of the problems of the original code that has been fixed with this revision.

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: This code is required for the REST API to function properly.   This is specific to the Polygon API and is the desired behavior to finish iterating through the pagination.

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,235 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+                }
+                else
+                {
+#if DEBUG
+                    Log.Trace("Processing " + start.ToString(System.Globalization.CultureInfo.GetCultureInfo("en-US")));
+#endif
```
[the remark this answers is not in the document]
TightLinesOfCode: Will fix-

ToolBox/Polygon/PolygonDataQueueHandler.cs
Martin-Molinero: Right, my point is that we are first checking for `(response != "")`, suggest directly replacing both for `string.IsNullOrEmpy()`?

ToolBox/Polygon/PolygonDataQueueHandler.cs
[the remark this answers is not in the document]
Martin-Molinero: Right, my point was that as is I don't think it's behaving as expected, `break` will just cause it to break the while loop and return, and the result of `start = start.Date.AddDays(1)` isn't being used?

ToolBox/Polygon/PolygonDataQueueHandler.cs
Martin-Molinero: if `isException` is `true` setting `offsetStart = offsetEnd` isn't required because it will go to the else case at line 623 and perform `offsetStart = offsetEnd`?

ToolBox/Polygon/PolygonDataQueueHandler.cs
Martin-Molinero: If the Exchanges can be NULL in the response they should be nullable fields in the `EquityQuoteTickResponse` class, else they will keep their default value of `0`

ToolBox/Polygon/PolygonDownloaderProgram.cs
Martin-Molinero: Cool! So maybe the following foreach could be parallel for each? I was suggesting removing it because it's not required/used as is today

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: Ahh.. Yes ... thanks.   I'll fix it

ToolBox/Polygon/PolygonDataQueueHandler.cs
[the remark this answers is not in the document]
TightLinesOfCode: I'll take another look.   The loop offset iterates over 1 day at a time.     So, if the day is not a common business day, then the offset loop stops and the outer loop should process the next day.

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: The Tick class ( and perhaps other areas of the code ) are not thread safe.   I had initially used a Parallel ForEach in my earlier versions, but this broke the system.    I will remove it for now.

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: will fix

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: thanks, will fix.

Martin-Molinero:
Hey @TightLinesOfCode, any update on these reviews https://github.com/QuantConnect/Lean/pull/5204#issuecomment-771269827 & https://github.com/QuantConnect/Lean/pull/5204#pullrequestreview-581539908 thanks 👍

TightLinesOfCode:
Still working on these.

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: Fixed

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: Done.

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: This is a server wide configuration, so it is also important if I want to run two separate APIs at the same time, so I think I should leave it in here.

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: I spent some time thinking about this, and I believe that this code is the correct behavior.   The JSON parser will set the value to null if there is not a JSON field returned from the API.  The API doesn't return a value for that field.  If I am confused about the behavior, please let me know.

TightLinesOfCode:
@Martin-Molinero I just changed the requested fixes, but I left the bid exchange behavior as it is.    I made comments in the conversations above to explain.    Please let me know if you want me to dig deeper on any of these.

ToolBox/Polygon/PolygonDataQueueHandler.cs
Martin-Molinero: >  The JSON parser will set the value to null if there is not a JSON field returned from the API

The parse will set the field to it's default value, for `int` it's 0, for it to be `null` the property should be `public int?` nullable see https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types

ToolBox/Polygon/PolygonDownloaderProgram.cs
Martin-Molinero: `ToList()` should be applied in the response of `downloader.Get()` so we reuse it when calling `writer.Write(data);`, else the enumeration will be triggered more than once

ToolBox/Polygon/PolygonDownloaderProgram.cs
Martin-Molinero: > This is a server wide configuration, so it is also important if I want to run two separate APIs at the same time

Believe the setting affects only objects created in this process `Gets or sets the maximum number of concurrent connections allowed by a ServicePoint object.`  see https://docs.microsoft.com/en-us/dotnet/api/system.net.servicepointmanager.defaultconnectionlimit?view=net-5.0. Don't think it's required

ToolBox/Polygon/PolygonDataQueueHandler.cs
```diff
@@ -507,105 +508,221 @@ private IEnumerable<Tick> GetCryptoTradeTicks(HistoryRequest request)
...
+                }
+
                 using (var client = new WebClient())
                 {
-                    var offset = Time.DateTimeToUnixTimeStampNanoseconds(start);
-                    var url = $"{HistoryBaseUrl}/v2/ticks/stocks/nbbo/{request.Symbol.Value}/{start.Date:yyyy-MM-dd}?apiKey={_apiKey}&timestamp={offset}";
+                    var url = $"{HistoryBaseUrl}/v2/ticks/stocks/nbbo/{request.Symbol.Value}/{start.Date:yyyy-MM-dd}?apiKey={_apiKey}&timestamp={offsetStart}&limit={_apiResultsLimit}";
 
-                    var response = client.DownloadString(url);
+                    Log.Debug(url);
```
Martin-Molinero, line 541: Let's make sure these logs follow the conversion of starting with the method name they are called from, for example: `Log.Debug(url);` -> `Log.Debug($"PolygonDataQueueHandler.GetEquityQuoteTicks(): target url {url}");`

ToolBox/Polygon/PolygonDataQueueHandler.cs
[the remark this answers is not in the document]
Martin-Molinero: This isn't behaving as expected, the `break` will exit the `while (offsetStart < offsetEnd)`

ToolBox/Polygon/PolygonSIPMappings.cs
```diff
@@ -0,0 +1,220 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace QuantConnect.ToolBox.Polygon
+{
+    // Provides information for SIP sale conditions according to the Polygon Data Vendor
+    // https://polygon.io/glossary/us/stocks/conditions-indicators
+    //
+    public static class PolygonSIPTradeMappings
```
Martin-Molinero, line 12: We should modify this class so it maps Polygon trade conditions into Leans format see https://github.com/QuantConnect/Lean/pull/5227/files and those values should be the ones to be stored in the `Tick.SaleCondition`

Martin-Molinero:
> 
> 
> @Martin-Molinero I just changed the requested fixes, but I left the bid exchange behavior as it is. I made comments in the conversations above to explain. Please let me know if you want me to dig deeper on any of these.

Hey @TightLinesOfCode! Thanks for the updates 💯, still a few to go I think, posted a few new comments/replies & unresolved some comments.

ToolBox/Polygon/PolygonDataQueueHandler.cs
TightLinesOfCode: I will try that and see what happens.   Thanks.

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: ok

ToolBox/Polygon/PolygonDataQueueHandler.cs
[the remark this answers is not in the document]
TightLinesOfCode: @Martin-Molinero     This is the expected behavior.   The Polygon API is setup for processing only one day at a time.   When this breaks, the calling method returns and processes the next day, which resets the offsetStart and offsetEnd.

downloader.Get(symbol, resolution, utcStartDate.AddDays(i), utcEndDate.AddDays(i + 1), tickType)

ToolBox/Polygon/PolygonDownloaderProgram.cs
TightLinesOfCode: @Martin-Molinero     On days where this is no data returned from the API call, the returned response is a null value.   Are you wanting this NULL check done in a different part of the code ?

Martin-Molinero (changes requested):
See previous comments

ToolBox/Polygon/PolygonDataQueueHandler.cs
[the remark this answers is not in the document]
jaredbroad: That might work for your application @TightLinesOfCode that is not the API exposed in this method; it is expected the method would return a complete history request with all the days. That looping and selection of the active days should be done internally. 

We should not rely on the caller's application of the interface to implement it or it will not be portable to other applications.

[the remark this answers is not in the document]
jaredbroad: PS: Ideally this should be using the Market Hours DB; what if Monday is a holiday; we can look ahead and remove this loop.

jaredbroad (changes requested):
Some changes required.

jaredbroad:
Closing for now but please feel free to reopen with the review comments addressed

PR #5207, opened 2021-01-22
OpenInterest Bug Fixes

### LINKED ISSUE #5093: Algorithm Manager Iterates Slice with Same Timestamp After AddOptionContract
#### Expected Behavior
The algorithm does not iterate Slice with the same timestamp.

#### Actual Behavior
After adding an `Option` security with `AddOptionContract`, `OnData` is triggered with the same `IAlgorithm.Time`. 

#### Potential Solution
N/A

#### Reproducing the Problem
```csharp
    public class AddOptionContractAlgorothm : QCAlgorithm {
	private Symbol _symbol;
	private DateTime _lastDatetime = DateTime.MinValue;
		
            SetStartDate(2020, 6, 1);
            
            var spy = AddEquity("SPY");
            spy.SetDataNormalizationMode(DataNormalizationMode.Raw);
            _symbol = spy.Symbol;
        }

        public override void OnData(Slice data) {
            if (Time.Ticks - _lastDatetime.Ticks < 1000)
                throw new Exception($"Repeated time: {Time}");
            
            var underlyingPrice = Securities[_symbol].Price;

            var contractSymbol = OptionChainProvider.GetOptionContractList(_symbol, Time)
            	.Where(x => x.ID.StrikePrice - underlyingPrice > 0)
            	.OrderBy(x => x.ID.Date)
		.FirstOrDefault();
            
            if (contractSymbol != null) {
            	var contract = AddOptionContract(contractSymbol);
            	_lastDatetime = Time;
            }
        }
    }
}
```

#### Checklist

C-SELLERS:
#### Description

- OpenInterest will be added as an internal feed and will not be added by default in history requests
- HistoryRequests and SubscriptionRequest will use AlwaysOpen exchange
  for open interest requests. Adding unit test reproducing the issue
- Adding `BaseDataRequest` to avoid duplication logic.

- `OptionOpenInterestRegressionAlgorithm.cs` stats changed because now they are calculated at a different time, because open interest data points are not being filtered by the exchange hours.

#### Related Issue

Closes #5093 & data requests filtering out valid open interest data points
Related to https://github.com/QuantConnect/Lean/pull/1002/files

#### Motivation and Context

This PR seeks to address some of the complications with Subscription data triggering a second timeslice for one DateTime. 

In the main branch when data is added in OnData() (meaning a timeslice has been emitted) the subscription is set to start +1 Tick after the current algorithm time to offset it from the current slice time, this was introduced in #4879, and was successful for equities because the first piece of data would arrive on the appropriate increment (resolution of the security). The problem with Options though, is we also have data for OpenInterest & Delisting data, and because this data is handled differently then other datatypes, we emit old data that has been filled forward, triggering the bug where we see a TimeSlice + 1 Tick ahead of the last one.

~The filtering of data in this PR seeks to reject datapoints that are before the StartTime, thus stopping the emitting of that datapoint at +1 Tick. Now this caused a problem with the algorithm manager recognizing delistings so we had to let that piece of data through or else automatic option exercise would fail.~ The PR sets OpenInterest so that it never triggers an `OnData` call by itself and will avoid OpenInterest from being filtered by exchange hours

The following is from testing the algorithm in the linked issue to see at what point the exception was triggered and what the contents of the offending TimeSlice were.
```
	LEGEND:
	D - Delisting
	OC - Option Chain
	OI - Open Interest
	PR:
	- 8/6 12AM +1 Tick D
	- 8/7 9:31 + 1 Tick D
	- 8/10 9:31 + 1 Tick D
	- 8/11 9:31 + 1 Tick D
	- 8/12 9:31 + 1 Tick D
	MASTER:
	- 8/6 12AM + 1 Tick D
	- 8/7 9:31 + 1 Tick OC, OI, D
	- 8/10 9:31 + 1 Tick D
	- 8/11 9:31 + 1 Tick OC, OI, D
        - 8/12 9:31 + 1 Tick OC, OI, D
```
As you can see the Option Chain and Open Interest data no longer emits at that time, but as we cannot filter out the delisting without causing more issues it remains.

Other attempts at solutions were made closer to the creation of the subscription but unfortunately due to some limitations in the Datafeed stack its difficult to control a subscriptions start time without a refactor to the subscription process. Because the function that create new subscription for the algo has no say in the startTime (AddSecurity only creates the SubscriptionDataConfig), and the actual creation of the subscription requests with the start time is in multiple places (UserDefinedUniverse, UniverseSelection, DataManager, etc.) There is not a unified way to increment the start time ONLY when the algorithm is already running.

N/A

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/Enumerators/SubscriptionFilterEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/SubscriptionFilterEnumerator.cs
@@ -127,7 +127,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
                     try
                     {
                         // Ensure the data time is after subscription start time
-                        if (current.EndTime < _startTime)
+                        if (current.EndTime < _startTime && current.DataType != MarketDataType.Auxiliary)
                         {
                             continue;
                         }
--- a/Engine/DataFeeds/Enumerators/SubscriptionFilterEnumerator.cs
+++ b/Engine/DataFeeds/Enumerators/SubscriptionFilterEnumerator.cs
@@ -37,6 +37,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
 
         private readonly bool _liveMode;
         private readonly Security _security;
+        private readonly DateTime _startTime;
         private readonly DateTime _endTime;
         private readonly bool _extendedMarketHours;
         private readonly SecurityExchange _exchange;
@@ -54,9 +55,9 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
         /// <param name="liveMode">True if live mode</param>
         /// <returns>A new instance of the <see cref="SubscriptionFilterEnumerator"/> class that has had it's <see cref="DataFilterError"/>
         /// event subscribed to to send errors to the result handler</returns>
-        public static SubscriptionFilterEnumerator WrapForDataFeed(IResultHandler resultHandler, IEnumerator<BaseData> enumerator, Security security, DateTime endTime, bool extendedMarketHours, bool liveMode)
+        public static SubscriptionFilterEnumerator WrapForDataFeed(IResultHandler resultHandler, IEnumerator<BaseData> enumerator, Security security, DateTime startTime, DateTime endTime, bool extendedMarketHours, bool liveMode)
         {
-            var filter = new SubscriptionFilterEnumerator(enumerator, security, endTime, extendedMarketHours, liveMode);
+            var filter = new SubscriptionFilterEnumerator(enumerator, security, startTime, endTime, extendedMarketHours, liveMode);
             filter.DataFilterError += (sender, exception) =>
             {
                 Log.Error(exception, "WrapForDataFeed");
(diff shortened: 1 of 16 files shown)


Engine/DataFeeds/Enumerators/SubscriptionFilterEnumerator.cs
```diff
@@ -124,6 +126,12 @@ public bool MoveNext()
                 {
                     try
                     {
+                        // Ensure the data time is after subscription start time
+                        if (current.EndTime < _startTime && current.DataType != MarketDataType.Auxiliary)
```
Martin-Molinero, line 130: Comment `Why` `current.DataType != MarketDataType.Auxiliary`

Engine/DataFeeds/SubscriptionDataReader.cs
```diff
@@ -365,7 +365,9 @@ public bool MoveNext()
                         }
                     }
 
-                    if (instance.EndTime < _periodStart)
+                    // To deal with OpenInterest data that only comes in once per day, we will check it against the periodStart date
```
Martin-Molinero, line 368: comment + `it will later be filtered correctly based on the start time by the SubscriptionFilterEnumerator which is after the FFEnumerator`

Martin-Molinero (commented):
Cool! where are the tests though? 😁 
Failing tests in travis:
```
1) Failed : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTickTypesAtTickResolution

  No data was emitted

  Expected: True

  But was:  False

  at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAllTickTypesAtTickResolution () [0x00000] in <a593e78d278342c4a736859cc4b3a0e0>:0 

2) Failed : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAuxiliaryDataAtTickResolution

  Expected: True

  But was:  False

  at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.HandlesAuxiliaryDataAtTickResolution () [0x00000] in <a593e78d278342c4a736859cc4b3a0e0>:0 

3) Failed : QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.SuspiciousTicksAreNotFilteredAtTickResolution

  Expected: True

  But was:  False

  at QuantConnect.Tests.Engine.DataFeeds.LiveTradingDataFeedTests.SuspiciousTicksAreNotFilteredAtTickResolution () [0x00000] in <a593e78d278342c4a736859cc4b3a0e0>:0 
```

PR #5251, opened 2021-02-02
Refactor Api Downloader Use FilePath Key

### LINKED ISSUE #5021: ApiDataProvider Downloading Wrong Data TickType
#### Expected Behavior

ApiDataProvider would download Quote or Trade data depending on the subscription data being fetched

#### Actual Behavior

For Equities it will download Trade data no matter what, even if the case of quote data subscription.

See 
`Api.cs`'s `DownloadData()` line 799 -> Will download and create a relative path 
`QuantConnect.Util.LeanData` line 737 -> Relative path is always trade data? I wonder if this is because we know that equities is always trade in QC data repo?

#### Potential Solution

There are a few ways to go about this but the number one thing that comes to mind is changing the way we request the download data, I think we need to include the ticktype in requesting the link. That way when I make a request for DownloadData I will include the ticktype. If it doesn't exist then I wouldn't need to download anything, whereas right now it just downloading trade data again that has likely already been downloaded before. (in the request for trade data)

As it stands I don't know that the Api "data/read" endpoint accepts a `ticktype` field. So I think this will take some cooperation between the web team and Lean team to fix. 

Furthermore "data/read" endpoint will reply to non-existent data requests with a link anyways see #4994, I think this also will need to be addressed. PR #5020 at least handles the bad link case and stops it from going to a corrupt file, but I think it should be stopped prior when requesting the link.

#### Reproducing the Problem

Debug Lean with ApiDataProvider, a basic algorithm with SPY subscribed, and have the trade data in your QC account:
- Breakpoint on `ApiDataProvider` at line 54,
- Wait for key = ***_quote.zip 
- Step through and watch the download process.

#### System Information

N/A

#### Checklist

C-SELLERS:
###WIP: Awaiting changes to Web API and testing.

#### Description

- Modify Api interface and ApiDataProvider to request data using the filepath key, which contains all the required information such as symbol, resolution, date, ticktype. 

#### Related Issue

Closes #5021 

#### Motivation and Context

Currently there are issues using the ApiDataProvider where you cannot specify the TickType and it will download whatever is available, even if lean is making a request for tradebar it may download quote, even if we already have it locally. Now this PR seeks to simplify the entire process by passing the FilePath key, which is standardized in Lean to contain all the required information about the data. 

API documentation will have to updated as now `DownloadData` and `ReadDataLink` only accept one param `filePath`. This path can be generated using `symbol` `date` `resolution` and `ticktype` with `LeanData.GenerateRelativeZipFilePath()`.

**WIP**

#### Types of changes

#### Checklist:


C-SELLERS:
Example request JSON from test `RestApiTests.BacktestingData_CanBeDownloadedAndSaved_Successfully()`

```
Resource = "data/read"
Method = POST
Parameters =
      {application/json={"format":"link","filePath":"../../../Data/forex/oanda/minute/eurusd/20131011_quote.zip"}}
```

Engine/DataFeeds/ApiDataProvider.cs
```diff
@@ -88,19 +88,19 @@ public static bool IsOutOfDate(Resolution resolution, string filepath)
...
-        /// <param name="symbol"></param>
-        /// <param name="date"></param>
-        /// <param name="resolution"></param>
+        /// <param name="filePath">The path to store the file</param>
+        /// <param name="symbol">The symbol we are downloading</param>
+        /// <param name="date">The date of the data</param>
+        /// <param name="resolution">The resolution of the data</param>
         /// <returns>A FileStream of the data</returns>
-        private FileStream DownloadData(string filepath, Symbol symbol, DateTime date, Resolution resolution)
+        private FileStream DownloadData(string filePath, Symbol symbol, DateTime date, Resolution resolution)
```
Martin-Molinero, line 96: We don't need the arguments `Symbol symbol, DateTime date, Resolution resolution` anymore really, think we should remove them if possible

Martin-Molinero (approved):
Cool!

Common/Data/Auxiliary/FactorFileZipHelper.cs
```diff
@@ -0,0 +1,76 @@
...
+                    let factorFile = SafeRead(filename, lines)
+                    let mapFile = mapFileResolver.GetByPermtick(factorFile.Permtick)
+                    where mapFile != null
+                    let sid = SecurityIdentifier.GenerateEquity(mapFile.FirstDate, mapFile.FirstTicker, market)
+                    let symbol = new Symbol(sid, mapFile.Permtick)
+                    select new { Symbol = symbol, FactorFile = factorFile }
+                )
+                .DistinctBy(x => x.Symbol)
+                .ToDictionary(x => x.Symbol, x => x.FactorFile);
+            file.DisposeSafely();
```
Martin-Molinero, line 52: `file.DisposeSafely()` think it's better that this is handled by the caller which is requesting the stream, same for `MapFileZipHelper`

Engine/DataFeeds/ApiDataProvider.cs
```diff
@@ -40,38 +60,114 @@ public class ApiDataProvider : IDataProvider
...
+            {
+                // Log URL to go accept terms
+                throw new Exception($"Must agree to terms at {_dataPrices.AgreementUrl}, before using the ApiDataProvider");
+            }
+
+            // Verify we have the balance to maintain our purchase limit, if not adjust it to meet our balance
+            _balance = organization.Credit.BalanceQCC;
+            if (_balance < _purchaseLimit)
+            {
+                Log.Trace("ApiDataProvider(): Purchase limit is greater than balance." +
```
Martin-Molinero, line 101: I think this should be more like a `Log.Error` or even throw argument exception?

Engine/DataFeeds/ApiDataProvider.cs
```diff
@@ -40,38 +60,114 @@ public class ApiDataProvider : IDataProvider
...
                 {
-                    return DownloadData(key, symbol, date, resolution);
+                    if (_purchaseLimit < price)
+                    {
+                        throw new Exception($"Cost {price} for {filePath} data exceeds remaining purchase limit: {_purchaseLimit}");
+                    }
+
+                    // Update our purchase limit and balance.
+                    _purchaseLimit -= price;
+                    _balance -= price;
```
Martin-Molinero, line 130: nit - not sure we need to keep track of the balance too?

Engine/DataFeeds/ApiDataProvider.cs
```diff
@@ -40,38 +60,114 @@ public class ApiDataProvider : IDataProvider
...
-            Log.Error("ApiDataProvider.Fetch(): failed to parse key {0}", key);
+            Log.Error($"ApiDataProvider.Fetch(): failed to fetch data at {filePath}");
             return null;
         }
 
+        private bool NeedToDownload(string filePath)
+        {
+            // Todo: TryParsePath logs error, should we do something to mute this?
+            // For files that TryParsePath can extract symbol, date, resolution (not factor files, map files, etc)
+            if (LeanData.TryParsePath(filePath, out Symbol symbol, out DateTime date, out Resolution resolution))
```
Martin-Molinero, line 150: `LeanData.TryParsePath` is able to parse map/factor file path?

Engine/DataFeeds/ApiDataProvider.cs
```diff
@@ -40,38 +60,114 @@ public class ApiDataProvider : IDataProvider
...
 
-                // Use the file already on the disk
-                return new FileStream(key, FileMode.Open, FileAccess.Read, FileShare.Read);
+                return DownloadData(filePath);
+            }
+
+            // Use the file already on the disk
+            if (File.Exists(filePath))
+            {
+                return new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read);
```
Martin-Molinero, line 139: Maybe it could inherit from `DefaultDataProvider` and call the base implementation here which skips checking if the file exists on disk, since it's already done by `FileStream`

Engine/DataFeeds/BacktestingOptionChainProvider.cs
```diff
@@ -27,6 +27,13 @@ namespace QuantConnect.Lean.Engine.DataFeeds
     /// </summary>
     public class BacktestingOptionChainProvider : IOptionChainProvider
     {
+        private IDataProvider _dataProvider;
+
+        public BacktestingOptionChainProvider(IDataProvider dataProvider)
```
Martin-Molinero, line 32: nit - missing docs, same for FutureChainProvider

Engine/DataFeeds/BacktestingOptionChainProvider.cs
```diff
@@ -51,32 +58,33 @@ public IEnumerable<Symbol> GetOptionContractList(Symbol underlyingSymbol, DateTi
...
-            if (!fileExists)
+            if (stream == null)
             {
                 Log.Trace($"BacktestingOptionChainProvider.GetOptionContractList(): File not found: {zipFileName}");
                 yield break;
             }
 
             // generate and return the contract symbol for each zip entry
-            var zipEntryNames = Compression.GetZipEntryFileNames(zipFileName);
+            var zipEntryNames = Compression.GetZipEntryFileNames(stream);
```
Martin-Molinero, line 87: Missing to dispose of the `stream`?

Tests/Api/ApiTestBase.cs
```diff
@@ -0,0 +1,47 @@
...
+        internal Api.Api ApiClient;
+
+        /// <summary>
+        /// Run once before any RestApiTests
+        /// </summary>
+        [OneTimeSetUp]
+        public void Setup()
+        {
+            TestAccount = Config.GetInt("job-user-id", 1);
+            TestToken = Config.Get("api-access-token", "ec87b337ac970da4cbea648f24f1c851");
```
Martin-Molinero, line 39: Cool class!
nit - `ec87b337ac970da4cbea648f24f1c851` isn't valid right?

Tests/TestGlobals.cs
```diff
@@ -0,0 +1,41 @@
...
+using QuantConnect.Data.Auxiliary;
+using QuantConnect.Interfaces;
+using QuantConnect.Lean.Engine.DataFeeds;
+
+namespace QuantConnect.Tests
+{
+    /// <summary>
+    /// Global static class for some core providers used by various tests
+    /// </summary>
+    public static class TestGlobals
```
Martin-Molinero, line 25: Cool!

Tests/Engine/DataFeeds/SubscriptionDataReaderTests.cs
```diff
@@ -46,7 +46,8 @@ public void DoesNotEmitDataBeyondTradableDate(string data, bool shouldEmitSecond
             var start = new DateTime(2019, 12, 9);
             var end = new DateTime(2019, 12, 12);
 
-            var mapFileProvider = new LocalDiskMapFileProvider();
+            var dataProvider = new DefaultDataProvider();
```
Martin-Molinero, line 49: Could use `TestGlobals.DataProvider`?

Tests/Common/Data/Auxiliary/LocalZipFactorFileProviderTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            var tmp = "./tmp.zip";
+
+            _zipFilePath = Path.Combine(Globals.DataFolder, $"equity/usa/factor_files/factor_files_{date:yyyyMMdd}.zip");
+
+            // Have to compress to tmp file or else it doesn't finish reading all the files in dir
+            QuantConnect.Compression.ZipDirectory(path, tmp);
+            File.Move(tmp, _zipFilePath);
+
+            FactorFileProvider = new LocalZipFactorFileProvider();
+            FactorFileProvider.Initialize(TestGlobals.MapFileProvider, TestGlobals.DataProvider);
```
Martin-Molinero, line 45: Nice! Still missing a few test here right? And the MapFile Zip provider test 😁

Tests/Api/DataTests.cs
```diff
@@ -0,0 +1,140 @@
...
+using QuantConnect.Api;
+using QuantConnect.Util;
+
+namespace QuantConnect.Tests.API
+{
+    /// <summary>
+    /// API Data endpoint tests
+    /// </summary>
+    [TestFixture]
+    class DataTests : ApiTestBase
```
Martin-Molinero, line 29: nit - missing `public`?
There are a few test failing in the CI travis/GH think because of missing ignores that were there before?

Tests/Api/OrganizationTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.IsTrue(list.Count > 0);
+        }
+
+        [Test]
+        public void ReadOrganization()
+        {
+            var organization = ApiClient.ReadOrganization(TestOrganization);
+            Assert.IsNotNull(organization.Credit);
+        }
+    }
```
Martin-Molinero, line 53: Maybe we can add some more asserts for these tests 👍

Tests/Api/ProjectTests.cs
```diff
@@ -0,0 +1,283 @@
...
+using System.Web;
+using NUnit.Framework;
+using QuantConnect.Api;
+
+namespace QuantConnect.Tests.API
+{
+    /// <summary>
+    /// API Project endpoints, includes some Backtest endpoints testing as well 
+    /// </summary>
+    class ProjectTests : ApiTestBase
```
Martin-Molinero, line 29: nit - public

Tests/Engine/DataProviders/ApiDataProviderTests.cs
```diff
@@ -51,5 +43,20 @@ public void OutOfDateTest(Resolution resolution, int days, bool expected)
...
+        {
+            var dataProvider = new ApiDataProvider();
+            var path = Path.Combine(Globals.DataFolder, file);
+            var stream = dataProvider.Fetch(path);
+
+            Assert.IsNotNull(stream);
+            stream.Dispose();
+            
+            Assert.IsTrue(File.Exists(path));
+        }
```
Martin-Molinero, line 60: We should add some tests for map files, also the NeedUpdate logic checking how old files are and assert it doesn't download twice

Martin-Molinero (changes requested):
Awesome stuff! Leaving a few requests and comments 👍

Engine/DataFeeds/ApiDataProvider.cs
C-SELLERS: I'm cool with Log.Error 👍🏽  I think its good to have this behavior because if they didn't set a "Purchase Limit" , then this sets their balance as limit.

Tests/Api/DataTests.cs
C-SELLERS: Yup, need to go back in and place all the ignores 👍🏽

Tests/TestGlobals.cs
C-SELLERS: Ty 😁

Common/Api/Serialization/ProductJsonConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (jObject == null)
+            {
+                return null;
+            }
+
+            return new Product
+            {
+                Type = _productTypeMap[jObject["name"].Value<string>()],
+                Items = jObject["items"].ToObject<List<ProductItem>>()
+            };
```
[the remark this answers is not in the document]
C-SELLERS: 😁

Engine/DataFeeds/ApiDataProvider.cs
C-SELLERS: No it can't thats why I have the else clause that just checks that file date. If you pass it a factorfile path it logs an error like so:

```
Error: System.ArgumentException: Requested value 'factor_files' was not found.
   at System.Enum.TryParseByName(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, UInt64& result)
   at System.Enum.TryParseInt32Enum(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Int32 minInclusive, Int32 maxInclusive, Boolean ignoreCase, Boolean throwOnFailure, TypeCode type, Int32& result)
   at System.Enum.TryParse(Type enumType, String value, Boolean ignoreCase, Boolean throwOnFailure, Object& result)
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase)
   at QuantConnect.Util.LeanData.TryParsePath(String fileName, Symbol& symbol, DateTime& date, Resolution& resolution) in C:\Users\Colton\Desktop\Working\QuantConnect\Repos\Lean\Common\Util\LeanData.cs:line 1020
   ```

It doesn't crash but I don't love that it posts the error. Wondering if it would be best to add a optional param bool to flag posting the error. True by default but can set to false in this case.

C-SELLERS: ✅

Martin-Molinero (approved):

PR #5258, opened 2021-02-04
Random fixes from position group work

mchandschuh:
#### Description
This branch sits on top of PR #5103 and contains miscellaneous fixes implemented while working on the core position group feature.

#### Related Issue
See #4065

#### Motivation and Context

Remove noise from the main #4065 pull request.

#### Requires Documentation Change

No.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

Issue #4065, opened 2020-02-05
Options margin modelling


Martin-Molinero:
#### Expected Behavior

- Options margin modelling is as close to reality as possible and allows usage of option strategies to reduce margin.
#### Actual Behavior

- To be reviewed
#### Potential Solution

- Review, refactor and add more tests
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

mchandschuh:
Here's my notes from researching margin requirements with a special focus on options. I've read through FINRA Rules 2360 and 4210, as well as the parts of Regulation T that concern themselves with margin with a particular focus on offsetting transactions. I know this is a long post, but I wanted it to be a complete and sufficient reference for the forthcoming development work to improve LEAN's option margin modeling.

#### FINRA: [Rule 2360: Options](https://www.finra.org/rules-guidance/rulebooks/finra-rules/2360)
FINRA Rule 2360 defines regulations specific to trading option contracts, including how holding different classes of options can offset margin requirements. The below excerpt is also accompanied by an example to illustrate how this works in practice.

##### FINRA: Rule 2360: Options - (a) Definitions
This section aims at providing key definitions used in other section and is reproduced here for completeness and to remove any ambiguity in the following definitions and discussions thereof.

> (3) Class of Options — The term "class of options" means all option contracts of the same type of option covering the same underlying security or index.
> (10) Covered — The term "covered" in respect of a short position in a call option contract means that the writer's obligation is secured by a "specific deposit" or an "escrow deposit," meeting the conditions of Rules 610(e) or 610(g), respectively, of the rules of The Options Clearing Corporation, or the writer holds in the same account as the short position, on a unit-for-unit basis, a long position either in the underlying security or in an option contract of the same class of options where the exercise price of the option contract in such long position is equal to or less than the exercise price of the option contract in such short position. The term "covered" in respect of a short position in a put option contract means that the writer holds in the same account as the short position, on a unit-for-unit basis, a long position in an option contract of the same class of options having an exercise price equal to or greater than the exercise price of the option contract in such short position.
[...]

PR #5259, opened 2021-02-04
Implement Position Groups

mchandschuh:
#### Description

This PR sits on top of #5258 and #5103.

These changes fundamentally change the basis for how LEAN views margin from being security-centric to position-centric. Positions can be grouped into `IPositionGroup` instances and position groups can be of different types. This PR provides the core position group types and the `SecurityPositionGroupBuyingPowerModel` to handle the buying power modeling for the 'default' position group which is 'ungrouped'. **ALL** positions are grouped. Any ungrouped quantities are placed into a default group dedicated for that security and the modeling falls back on the security's defined models. Upcoming changes will add an `OptionStrategyPositionGroupBuyingPowerModel` as well as wire up the `OptionsMatcher` as an `IPositionGroupResolver`. Position group resolvers inspect the algorithm's holdings and try to determine the best way to group securities in order to decrease total margin requirements. For example, you can decrease margin requirements when writing calls if you own the underlying stock (covered calls). Before this change LEAN would evaluate the margin requirements of each independently (underlying and short call). This leads to modelling errors and removes many of the benefits brokers provide to encourage hedged positions.

#### Related Issue

See #4065

#### Motivation and Context

The main motivation is to improve overall modeling fidelity and to enable LEAN to compute margin requirements for groups instead of securities individually. The next PR in this line of work will wire in the `OptionsMatcher` which will try to group options and underlyings in such a way as to minimize maintenance margin requirements, thereby allowing the algorithm to take positions it otherwise wouldn't be able to due to insufficient margin.

Another upcoming change in this line of work is the implementation of the `ComboOrder` which will support buying/selling position groups as a unit and will require brokerage updates.

YES! I believe these and upcoming changes warrant a dedicated section in the documentation.

Regression and unit tests -- backwards compatibility proved via `BuyingPowerModelComparator` which ensures that the `SecurityPositionGroupBuyingPowerModel` returns the same values as the security's `IBuyingPowerModel`.

#### Types of changes

#### Checklist:

- [x] All new and existing tests passed. [need to inspect python tests via travis, unable to run locally due to known environment issues -- it's not that I can't get it to work, it's that by getting it to work it breaks things I use in my full time job)

### DIFF
--- a/Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
+++ b/Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
  *
@@ -108,7 +108,7 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
             // we normalize the target buying power by the leverage so we work in the land of margin
             var targetFinalMarginPercentage = adjustedPercent / security.BuyingPowerModel.GetLeverage(security);
 
-            var positionGroup = algorithm.Portfolio.Positions.CreateDefaultGroup(security);
+            var positionGroup = algorithm.Portfolio.Positions.GetOrCreateDefaultGroup(security);
             var result = positionGroup.BuyingPowerModel.GetMaximumLotsForTargetBuyingPower(
                 algorithm.Portfolio, positionGroup, targetFinalMarginPercentage
             );
--- a/Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
+++ b/Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using QuantConnect.Interfaces;
 using QuantConnect.Securities;
+using QuantConnect.Securities.Positions;
 using static QuantConnect.StringExtensions;
 
 namespace QuantConnect.Algorithm.Framework.Portfolio
@@ -107,8 +108,9 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
             // we normalize the target buying power by the leverage so we work in the land of margin
             var targetFinalMarginPercentage = adjustedPercent / security.BuyingPowerModel.GetLeverage(security);
 
-            var result = security.BuyingPowerModel.GetMaximumOrderQuantityForTargetBuyingPower(
-                new GetMaximumOrderQuantityForTargetBuyingPowerParameters(algorithm.Portfolio, security, targetFinalMarginPercentage, silenceNonErrorReasons:true)
+            var positionGroup = algorithm.Portfolio.Positions.CreateDefaultGroup(security);
+            var result = positionGroup.BuyingPowerModel.GetMaximumLotsForTargetBuyingPower(
+                algorithm.Portfolio, positionGroup, targetFinalMarginPercentage
             );
 
             if (result.IsError)
@@ -122,7 +124,8 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
 
             // be sure to back out existing holdings quantity since the buying power model yields
             // the required delta quantity to reach a final target portfolio value for a symbol
-            var quantity = result.Quantity + (returnDeltaQuantity ? 0 : security.Holdings.Quantity);
+            var lotSize = security.SymbolProperties.LotSize;
+            var quantity = result.NumberOfLots * lotSize + (returnDeltaQuantity ? 0 : security.Holdings.Quantity);
 
             return new PortfolioTarget(symbol, quantity);
         }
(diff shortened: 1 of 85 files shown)


Common/Securities/Positions/PositionManager.cs
```diff
@@ -0,0 +1,202 @@
...
+        public void ResolvePositionGroups()
+        {
+            if (_requiresGroupResolution)
+            {
+                _requiresGroupResolution = false;
+                // TODO : Replace w/ special IPosition impl to always equal security.Quantity and we'll
+                // use them explicitly for resolution collection so we don't do this each time
+
+                Groups = ResolvePositionGroups(new PositionCollection(_securities
+                    .Where(kvp => kvp.Value.Invested).ToImmutableDictionary(
```
Martin-Molinero, line 155: `.Where(kvp => kvp.Value.Invested)` Will remove uninvested securities which seems different from the behavior at `securities.CollectionChanged +=` where we add all securities with the default position group?

Common/Securities/Positions/PositionManager.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        public IPositionGroup CreateDefaultGroup(Security security)
+        {
+            var key = CreateDefaultKey(security);
+            return Groups[key];
+        }
+
+        private void HoldingsOnQuantityChanged(object sender, SecurityHoldingQuantityChangedEventArgs e)
+        {
+            _requiresGroupResolution = true;
```
Martin-Molinero, line 199: Here we differ the calculation of the group resolution which seems different than what is being done on a `Fill`, wonder why? Should both be deferred instead?

Common/Securities/DefaultMarginCallModel.cs
```diff
@@ -148,19 +149,19 @@ protected virtual SubmitOrderRequest GenerateMarginCallOrder(Security security,
             // compute the amount of quote currency we need to liquidate in order to get within margin requirements
             var deltaAccountCurrency = totalUsedMargin - totalPortfolioValue;
 
-            var currentlyUsedBuyingPower = security.BuyingPowerModel.GetReservedBuyingPowerForPosition(
-                new ReservedBuyingPowerForPositionParameters(security)).AbsoluteUsedBuyingPower;
+            var positionGroup = Portfolio.Positions.CreateDefaultGroup(security);
+
+            var currentlyUsedBuyingPower = positionGroup.BuyingPowerModel.GetReservedBuyingPowerForPositionGroup(Portfolio, positionGroup);
```
Martin-Molinero, line 154: Remember we talked about this case being a bit of a PITA. The value of `currentlyUsedBuyingPower` might not be accurate because it's using the `DefaultGroup` and this security could actually be part of another group/S determined by the resolver, right? Those groups could be accessed through the collection at the `PositionManager` right?

Martin-Molinero (commented):
Pretty cool stuff! 🙌 
Sorry for the late review, leaving a few questions and minor comments

Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
```diff
@@ -107,8 +108,9 @@ public static IPortfolioTarget Percent(IAlgorithm algorithm, Symbol symbol, deci
             // we normalize the target buying power by the leverage so we work in the land of margin
             var targetFinalMarginPercentage = adjustedPercent / security.BuyingPowerModel.GetLeverage(security);
 
-            var result = security.BuyingPowerModel.GetMaximumOrderQuantityForTargetBuyingPower(
-                new GetMaximumOrderQuantityForTargetBuyingPowerParameters(algorithm.Portfolio, security, targetFinalMarginPercentage, silenceNonErrorReasons:true)
+            var positionGroup = algorithm.Portfolio.Positions.CreateDefaultGroup(security);
```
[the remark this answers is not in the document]
mchandschuh: I've changed this method to be called `GetOrCreateDefaultGroup` which better describes the behavior. Here's the implementation in `PositionManager` and then `PositionGroupCollection`'s `this[]` indexer:
``` csharp
// https://github.com/QuantConnect/Lean/blob/refactor-4065-position-groups/Common/Securities/Positions/PositionManager.cs#L191
public IPositionGroup CreateDefaultGroup(Security security)
{
    var key = CreateDefaultKey(security);
    return Groups[key];
}

// https://github.com/QuantConnect/Lean/blob/refactor-4065-position-groups/Common/Securities/Positions/PositionGroupCollection.cs#L130
public IPositionGroup this[PositionGroupKey key]
{
    get
    {
        IPositionGroup group;
        if (!TryGetGroup(key, out group))
        {
            return new PositionGroup(key, key.CreateEmptyPositions());
        }

        return group;
    }
}
```

IlshatGaripov:
Michael suppose portfolio is a compact closed set . Then there are the series of mappings done by alphas from subset of portfolio set to a set of shares. What will be the position group here ?

IlshatGaripov:
I suppose that portfolio falls into non intersecting subsets and union constitute the portfolio set. Single Alpha maps from a such single subset into Sequence of shares . This map should meet certain conditions .. like margin requirements .

Issue #452, opened 2016-07-04
Add .Net Core Support


gaoxiaojun:
Microsoft has been released .Net Core 1.0.0 and SDK Preview2,

jaredbroad:
Thanks @gaoxiaojun! Do you mind giving an example use case to use it in LEAN?

bnmanager:
@jaredbroad @gaoxiaojun 

**Universal Windows Apps (UWP)**, (of which I am currently developing an app for), use `.NET Core` as their foundation. Therefore, many useful libraries found in console or WPF programs are simply incompatible with .NET Core. Currently, QuantConnect is one of those.

Also,`.NET Core` is open source and can run on `Windows/Linux/Apple`.

jaredbroad:
Thank you @gaoxiaojun @bnmanager. I'll leave this open but from my reading it seems like a mammoth & risky task to convert the platform across. We can start a branch if there is someone to lead the effort.

For now LEAN is cross platform Windows/Linux/Apple via Mono. The only limitation to date is the winforms UX project which is windows only.

omencat:
Instead of targeting .net core, the easiest way today to go about this would be to target the [.net standard 2.0 surface](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/). Using the [.net portability analyzer](https://github.com/Microsoft/dotnet-apiport/blob/master/docs/VSExtension/README.md) extension, it looks like QuantConnect.Algorithm.CSharp is 98.47% complete. QuantConnect.Indicators is already 100% complete.

lnaie:
yes, just wait for dns 2 release and you should be 99% good to go.

wklm:
Any news on that?

jaredbroad:
If you click subscribed you'll be notified when there are updates to the
ticket

wklm:
I was asking, because if there's no one currently working on it I could give it a try

snielsson:
Hi, what is the current status of this ?

mamatopian:
Hey guys, any news?

omidkrad:
Hi, is migrating to .NET Core in the plans at all?

omidkrad:
Or maybe a more appropriate compilation target is [.NET Standard 2.0+](https://dotnet.microsoft.com/platform/dotnet-standard).

Leonardo-Ferreira:
[...]

PR #5261, opened 2021-02-05
Updates CPU performance metrics gathering approach

gsalaz98:
#### Description

PerformanceCounters are not supported cross-platform in .NET Core/5.0, which requires the use of an alternative method in the gathering of performance metrics. 

Since no exposed .NET API exists to gather these metrics without blocking, a new thread is created to block the time necessary for the CPU performance to be calculated.

Other attempts made before this was decided as the solution were using `EventCounter` and other related classes in the Microsoft.Diagnostics namespace. 
#### Related Issue

.NET Core/5.0 migration #452 
#### Motivation and Context

Cross-platform code for eventual .NET Core/5.0 migration

No

Cloud testing (backtesting, research) local testing (backtesting C#)
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/OS.cs
+++ b/Common/OS.cs
@@ -20,7 +20,6 @@ using System.IO;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
-using Log = QuantConnect.Logging.Log;
 using static QuantConnect.StringExtensions;
 
 namespace QuantConnect
@@ -153,6 +152,7 @@ namespace QuantConnect
             /// <summary>
             /// CPU usage as a percentage (0-100)
             /// </summary>
+            /// <remarks>Float to avoid any atomicity issues</remarks>
             public float CpuPercentage { get; private set; }
 
             /// <summary>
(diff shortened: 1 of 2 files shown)


Common/OS.cs
```diff
@@ -160,5 +138,68 @@ public static decimal CpuUsage
...
+                    var startTime = DateTime.UtcNow;
+                    var startCpuUsage = Process.GetCurrentProcess().TotalProcessorTime;
+                    var stopWatch = Stopwatch.StartNew();
+
+                    if (_exitEvent.Wait(1000))
+                    {
+                        return;
+                    }
+
+                    stopWatch.Stop();
```
Martin-Molinero, line 182: `stopWatch` isn't used I think? Using `DateTime.UtcNow` which seems better 👍

Common/OS.cs
```diff
@@ -160,5 +138,68 @@ public static decimal CpuUsage
...
+                    var endTime = DateTime.UtcNow;
+                    var endCpuUsage = Process.GetCurrentProcess().TotalProcessorTime;
+
+                    var cpuUsedMs = (endCpuUsage - startCpuUsage).TotalMilliseconds;
+                    var totalMsPassed = (endTime - startTime).TotalMilliseconds;
+                    var cpuUsageTotal = cpuUsedMs / totalMsPassed;
+
+                    CpuPercentage = cpuUsageTotal * 100;
```
Martin-Molinero, line 190: We will need some access control because written to the double isn't guaranteed to be atomic, meaning it can be read while being modified, see https://stackoverflow.com/a/11745701/7334720 https://stackoverflow.com/a/3677373/7334720 . Can use some interlock exchange for example 👍

Martin-Molinero (approved):
Thanks! You should make a public repo just holding this guy 😁

Martin-Molinero (approved):

PR #5263, opened 2021-02-05
Fix for "Engine does not warmup live algorithms" issue

### LINKED ISSUE #5262: Engine does not warmup live algorithms
#### Expected Behavior
The engine feeds data to warm up every security added by user.

#### Actual Behavior
Historical warm up data does not reach the algorithm and consolidators.

#### Potential Solution
Will be posted as PR with more detailed description.

#### Reproducing the Problem
Run live algorithm with Bitfinex brokerage, algorithm inherits QCAlgorithm.Framework design paradigm, securities are added via the UniverseSelectionModel.

#### System Information
Lean on stand-alone machine, Windows10/WindowsServer2019.

#### Checklist

IlshatGaripov:
#### Description
There are few things here related to addressed issue I have notied:

1) First and most improtant thing here is if we deploy an alogirhtm written using Agorithm.Framework (and probably the classical way too) in a live mode. Then Universe selection, and security additions and non-internal subscriptionDataConfigs additions are happening only once we unwrap synchronizer's data streming and run the first time slice results all the way through the Run method, to handle SecrutyChanges and add securities. Otherwise we can't construct proper history request. Thus we come up with design when Stream() spins off the synchronizer first, let it start streaming, and then calls HistoryWarmpUp requests after a couple of cycles in appropriate moment. I believe such design is quite in the spirit of laid down architecture, because there are the following comments can be found in code :

// this is hand-over logic, we spin up the data feed first and then request
// the history for warmup, so there will be some overlap between the data

2) If synchronizer is not launched before HistoryProcessor called, this line of codes does not find all the required subscriptions:
https://github.com/QuantConnect/Lean/blob/master/Engine/AlgorithmManager.cs#L806
and in result the produced timeslice contains empty bars  . 
And this line of code may be missing some symbols from yet no initialized Securities:
https://github.com/QuantConnect/Lean/blob/master/Algorithm/QCAlgorithm.History.cs#L168

3) Next thing is I made securities that are being created in Cash.cs for currency conversions Non-Tradable. Because they are not really created for trading here. And used the following filter:
var hasTradableSecurities = algorithm.Securities.Values.Any(s => s.IsTradable);
to define the moment when required Security initialization is done and when is the moment we can call history processor inside the SynchronizerProcessor.

#### Related Issue
Closes #5262 

#### Motivation and Context
An issue with live algorithm

NA

This fixes the described issue with local algorithm. But it's worth to think of all possible scenarios and behaviours that these changes may break, I would appreciate a review from core Lean developers.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -613,11 +613,12 @@ namespace QuantConnect.Algorithm
 
                 foreach (var config in GetMatchingSubscriptions(x, typeof(BaseData), resolution))
                 {
+                    if (config.IsInternalFeed) continue;
                     var request = _historyRequestFactory.CreateHistoryRequest(config, startAlgoTz, endAlgoTz, GetExchangeHours(x), resolution);
 
                     // apply overrides
                     var res = GetResolution(x, resolution);
-                    if (fillForward.HasValue) request.FillForwardResolution = fillForward.Value ? res : (Resolution?)null;
+                    if (fillForward.HasValue) request.FillForwardResolution = fillForward.Value ? res : (Resolution?) null;
                     if (extendedMarket.HasValue) request.IncludeExtendedMarketHours = extendedMarket.Value;
 
                     requests.Add(request);
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -167,14 +167,17 @@ namespace QuantConnect.Algorithm
         /// <returns></returns>
         public IEnumerable<HistoryRequest> GetWarmupHistoryRequests()
         {
+            var symbolsToWarmup = Securities.Values
+                .Where(s => s.IsTradable).Select(s => s.Symbol);
+
             if (_warmupBarCount.HasValue)
             {
-                return CreateBarCountHistoryRequests(Securities.Keys, _warmupBarCount.Value, _warmupResolution);
+                return CreateBarCountHistoryRequests(symbolsToWarmup, _warmupBarCount.Value, _warmupResolution);
             }
             if (_warmupTimeSpan.HasValue)
             {
                 var end = UtcTime.ConvertFromUtc(TimeZone);
-                return CreateDateRangeHistoryRequests(Securities.Keys, end - _warmupTimeSpan.Value, end, _warmupResolution);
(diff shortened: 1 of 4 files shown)


IlshatGaripov:
Few more comments to the last commit . The previously suggested flag for the moment when is the right time to call HistoryProcessor 
 hasTradableSecurities = algorithm.Securities.Values.Any(s => s.IsTradable);
seems to be not taking into account all scenarious. 
Consider algorithm launched and has open positions or open orders in ETHUSD. Then they will be added as unrequested Secutiy and marked is Tradable, which is correct,, and added to UserDefinedUniverse . And so hasTradableSecurities  may work up before Security changes were properly handler by Algorithm manager, in result Algorithm.Framework  OnSecuritiesChanged may not be called before HistoryProcessor is stearted. OnSecuritiesChanged is the place where usually symbol data and various consolidators may be initialized -> in result they do not get updated with historical data -> this is a bug. Need to make sure OnSecuritiesChanged  is called before start historical data processing..

2) And second need to filter out internal subscriptions in CreateDateRangeHistoryRequests. Consider example. ETH is used as collaterial in margin trading, USD is a base currency. Then ETHUSD is being added as an internal subscription. On top a used adds ETHUSD as a manual defined universe security -> result two subscriptions -> and two historical requests -> which is not quite what is required

IlshatGaripov:
@mchandschuh Michael, please kindly could you find a little time to check it out - I think improvements needed here affecting the core Lean - I would be open to cover the required parts by tests. Just wanteв to learn in general how such significant changes can be integrated what needs to be done

Martin-Molinero:
Thanks for taking a stab at this one @IlshatGaripov! It's a hard one we've deferred so far, leaving some thoughts bellow:

In the current approach calling `synchronizer.StreamData()`, to trigger selection, and later fetching the historical data is a tricky one, because the selection itself will use current data/time, and later it will feed older data into the algorithm so there is kind of a look ahead there. Also it's relying on the first slice to go through and set `_appliedSecurityChanges` and I think this slice could also potentially bring _current time_ data points that will be feed into the algorithm.
I think the ideal world solution would be deeper into the data stack, where the frontier time itself starts from the past and we inject/provide historical data points into each enumerator as it's required, and for backtesting it wouldn't even really need to perform a history request

Martin-Molinero:
Closing PR now - Let's continue discussing potential approaches in the associated issue 👍

IlshatGaripov:
> Closing PR now - Let's continue discussing potential approaches in the associated issue 👍

Sounds good 👍

PR #5298, opened 2021-02-15
Implements Coin API HistoryProvider

IlshatGaripov:
#### Description
Coin Api has a nice source for historical data - very fast - just what you need to warm up live algorithms. Unfortauntely limited by tarrifs' daily requests limit

#### Related Issue
NA

#### Motivation and Context
Live crypto algorithms deployment improvement

NA

Tests are provided 

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/ToolBox/CoinApiHistoryProviderTests.cs
+++ b/Tests/ToolBox/CoinApiHistoryProviderTests.cs
@@ -30,7 +30,7 @@ namespace QuantConnect.Tests.ToolBox
     {
         private static readonly Symbol _CoinbaseBtcUsdSymbol = Symbol.Create("BTCUSD", SecurityType.Crypto, Market.GDAX);
         private static readonly Symbol _BitfinexBtcUsdSymbol = Symbol.Create("BTCUSD", SecurityType.Crypto, Market.Bitfinex);
-        private readonly CoinApiDataQueueHandler _coinApiDataQueueHandler = new CoinApiDataQueueHandler(); 
+        private readonly CoinApiDataQueueHandlerMock _coinApiDataQueueHandler = new CoinApiDataQueueHandlerMock(); 
 
         // -- DATA TO TEST --
         private static TestCaseData[] TestData => new[]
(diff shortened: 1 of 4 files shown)


Tests/ToolBox/CoinApiHistoryProviderTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            };
+
+            var slices = _coinApiDataQueueHandler.GetHistory(historyRequests, TimeZones.Utc).ToArray();
+            
+            if (isNonEmptyResult)
+            {
+                // Slices are not empty
+                Assert.IsNotEmpty(slices);
+                // And are ordered by time
+                Assert.That(slices, Is.Ordered.By("Time"));
```
Martin-Molinero, line 66: Cool! Thanks for adding the tests 👍 
Think we should assert the data too, for example:
- each slice should contain the requested symbol
- the data should: 
     - be of the requested resolution
     - in the requested date range
     - of the requested type
- we should assert the data count is expected

What's the behavior for:  Resolution.Tick, Resolution.Second & Quote type of history requests, let's assert these cases too 👍

ToolBox/CoinApi/Messages/HistoricalDataMessage.cs
```diff
@@ -0,0 +1,66 @@
...
+        [JsonProperty("trades_count")]
+        public int TradesCount { get; set; }
+
+        /// <summary>
+        /// Converts CoinApi OHLCV to QuantConnect TradeBar 
+        /// </summary>
+        /// <returns></returns>
+        public TradeBar ToTradeBar()
+        {
+            return new TradeBar()
```
Martin-Molinero, line 54: `Period` should be set I believe? Let's directly use one of the TradeBar constructors think it's less error-prone, for example it would have shown the missing period being set 👍

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -326,5 +341,94 @@ private void OnError(object sender, Exception e)
...
+            var remaining = GetHttpHeaderValue(response, "x-ratelimit-remaining");
+
+            Log.Trace($"CoinApiDataQueueHandler.TraceRestUsage(): Used {used}, Remaining {remaining}, Total {total}");
+        }
+
+        private string GetHttpHeaderValue(IRestResponse response, string propertyName)
+        {
+            return response.Headers.ToList()
+                .Find(x => x.Name == propertyName)
+                .Value.ToString();
```
Martin-Molinero, line 431: This will throw null ref if the property name is not present, suggest adding a check for it
```
        [Test]
        public void NullReference()
        {
            var l = new List<Parameter>();
            var tt = l.Find(s => s.Name == "pinocho").Value.ToString();
        }
```

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -326,5 +341,94 @@ private void OnError(object sender, Exception e)
...
+                Log.Error($"CoinApiDataQueueHandler.GetHistory(): Invalid security type {historyRequest.Symbol.SecurityType}");
+                yield break;
+            }
+
+            if (historyRequest.Resolution == Resolution.Tick)
+            {
+                Log.Error("CoinApiDataQueueHandler.GetHistory(): Historical ticks can not be provided, only OHLCV timeseries");
+                yield break;
+            }
+
```
Martin-Molinero, line 377: Should also add a check for the tick type, since current implementation only supports Trade information, not Quote

Martin-Molinero (changes requested):
Thanks @IlshatGaripov for this nice PR!
Leaving a few requests and minor findings, almost there

IlshatGaripov:
HI @Martin-Molinero , thank for this review,  to use TradeBar constructor is really good suggestion because the Symbol was also missing and data is not pumped because of it. I will be addressing the changes just shortly. thanks

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
```diff
@@ -326,5 +341,113 @@ private void OnError(object sender, Exception e)
...
+            // Perform a check the number of bars requested, this must not exceed the limit of 100k
+            var dataRequestedCount = (historyRequest.EndTimeUtc - historyRequest.StartTimeUtc).Ticks 
+                                     / historyRequest.Resolution.ToTimeSpan().Ticks;
+
+            if (dataRequestedCount > 100000)
+            {
+                Log.Error("CoinApiDataQueueHandler.GetHistory(): The requested data range is too large, " +
```
Martin-Molinero, line 390: Think the history request could be paginated and performed in more than one call?

Martin-Molinero (commented):
Thanks for the updates, looks good! Leaving a final request 👍

ToolBox/CoinApi/CoinApiDataQueueHandler.cs
IlshatGaripov: Hi @Martin-Molinero thanks for review. I will fix this on Monday, sorry little bit busy week for me here..

Martin-Molinero: Thanks! 👍

Martin-Molinero (approved):

PR #5309, opened 2021-02-17
Net50

xposure:
All the information can be found here https://github.com/QuantConnect/Lean/issues/452 if the QuantConnect team wants to see .net5 support to truly be added, somebody else will have to pull up their seelves and do the proper unit tests and verifying python works correctly (looking for additional help).

As of now this PR does the following...

 - Cleaned up all the csproj files
 - Created submodules for projects that aren't updated to .net5
 - Added Directory.Build.props to help streamline the build process and remove the duplication
 - Removed Interactive Brokers and FXCM, they need updated packages
 - I've added docker support that uses staged builds
 - I've added a dotnet template to make getting started easier
 - Removed all nuspec files and switched over to `dotnet pack`
 - Docker images and packages are stored in my accounts until there is a determination if this can be merged

Issues:
 - I had to rebrand some things to `Xposure` to avoid conflicts with the real QuantConnect packages. These would be replaced once merged.
 - RDotNet and WebSocketSharp needs to be updated if you want to use them as a package and not submodule
 - Python is kind of janky, there are windows/linux/osx references to a nuget base path, these should really just come from a proper quantconnect pythonnet using a package that has all these variations needed. This would further simplify the csproj files.
 - I looked at the original docker image which has a lot of deps installed, I highly doubt python will work and would need fixed in my docker image, c# works fine.

TODOS:
 - Add skaffold to enable easy docker debugging straight from vscode
 - Move package versioning to Directory.Build.props

You can review the readme I have updated here. https://github.com/xposure/lean/tree/net50 with additional information


jaredbroad:
Thank you for the PR @xposure. We've been slowly and carefully migrating to .NET core and testing each step of the way for the last 2-3 months now. I appreciate your contribution but in the future recommend connecting with one of the core dev team to coordinate efforts so we can avoid lost work.

xposure:
Can I ask what has taken 2-3 months? This is a days worth of work and is already capable of running c# on .net 5 through docker, kubernetes and skaffold. 

Do you have some sort of roadmap that I missed that is tracking this progress? From what I can tell, nobody has commented anything on the actual issue for the last 4 years.

xposure:
I will continue my work, if anyone is interested in using .net 5 today with c#, its available.

cmoski:
In my estimation, it doesn't make a lot of sense to even bother with core at this point, as the MS roadmap is to merge core into .NET5.

Unless there is 'stability problem' with dotnet5 or a mass of users who need NET Standard 2.0, seems to me to be a step behind.

Just 2c form someone who has been waiting for support of either NET5 or core for a long, long time. (PRs on this are sisyphean task)

xposure:
@jaredbroad Could you please answer my questions here so that your team can work with me to get .NET5 sooner than later. From the merges I have seen it looks like you are working on .NET standard which is about 4 years too late. I would really like to help out and provide my expertise as this isn't my first conversion project and they've been much larger than this by several magnitudes. The only thing your contributors would need to help with is the python side of things and getting proper Fxcm + IB packages.

xposure:
@Martin-Molinero I'm pinging you on this as well since I've seen that you have been actively involved in the transition, perhaps you could help as well?

PR #5351, opened 2021-03-02
Refactors Capacity Estimation and moves estimation to main event loop

gsalaz98:
#### Description

The capacity estimation has been moved from the report generator and wired directly into Lean via the ResultHandler. In addition, the capacity estimation strategy has changed to account for errors in the previous iteration of the capacity estimation.

The capacity estimation in the report generator has been updated to use the statistics results in the BacktestResult packet.

The `$0.00` Capacity Regression test cases are caused by one of the following reasons:
  
  * Zero volume for the assets being traded in the algorithm
  * Option exercise before snapshot date, causing the asset to be delisted before calculating capacity
  * Delisting of asset before snapshot date
  * No trades placed by the algorithm

Further changes:

  * Moves old tests to regression algorithms
  * Adds Estimated Capacity statistic
  * Moves old capacity estimation tests to Algorithm.CSharp

The PR has been separated into different commits per massive updates for the reviewer's ease of reviewing :) 

#### Related Issue

N/A, Alpha Streams were reporting inconsistent capacities ranging from negative values to insane numbers.
#### Motivation and Context

Fixes Alpha Streams capacity estimates

If we intend to document how the capacity estimation works publicly (I think we should eventually), then maybe

Regression tests + sanity checks, capacity estimation test cases
#### Types of changes

#### Checklist:


jaredbroad (approved):

Engine/Results/SymbolCapacity.cs
```diff
@@ -0,0 +1,279 @@
...
+        /// <param name="algorithm"></param>
+        /// <param name="symbol"></param>
+        public SymbolCapacity(IAlgorithm algorithm, Symbol symbol)
+        {
+            _algorithm = algorithm;
+            Security = _algorithm.Securities[symbol];
+            _symbol = symbol;
+            _holding = _algorithm.Portfolio[_symbol];
+
+            var resolution = Security.Subscriptions
```
Martin-Molinero, line 104: `_resolution` might not be static, meaning securities could be added with different resolutions.
Let's replace usage of `Security.Subscriptions` for `algorithm.SubscriptionManager.SubscriptionDataConfigService.GetSubscriptionDataConfigs(symbol)` to get the subscriptions, the first isn't intended to be used by the engine really.

Engine/Results/CapacityEstimate.cs
```diff
@@ -0,0 +1,167 @@
...
+{
+    /// <summary>
+    /// Estimates dollar volume capacity of algorithm using all Symbols in the portfolio.
+    /// </summary>
+    public class CapacityEstimate
+    {
+        private readonly IAlgorithm _algorithm;
+        private readonly Dictionary<Symbol, SymbolCapacity> _capacityBySymbol;
+        private List<SymbolCapacity> _monitoredSymbolCapacity;
+        private HashSet<SymbolCapacity> _monitoredSymbolCapacitySet;
```
Martin-Molinero, line 34: Why do we need so many collections for the `SymbolCapacity`? Note that the `HashSet<SymbolCapacity>` is using the reference for uniqueness

Engine/Results/SymbolCapacity.cs
```diff
@@ -0,0 +1,279 @@
...
+        /// Creates a new SymbolCapacity object, capable of determining market capacity for a Symbol
+        /// </summary>
+        /// <param name="algorithm"></param>
+        /// <param name="symbol"></param>
+        public SymbolCapacity(IAlgorithm algorithm, Symbol symbol)
+        {
+            _algorithm = algorithm;
+            Security = _algorithm.Securities[symbol];
+            _symbol = symbol;
+            _holding = _algorithm.Portfolio[_symbol];
```
Martin-Molinero, line 102: `_algorithm.Portfolio[_symbol]` -> `Security.Holding` to avoid another dictionary access

Engine/Results/SymbolCapacity.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            return null;
+        }
+
+        /// <summary>
+        /// Signals a reset for the <see cref="MarketCapacityDollarVolume"/> and <see cref="SaleVolume"/>
+        /// </summary>
+        public void Reset()
+        {
+            _resetMarketCapacityDollarVolume = true;
```
Martin-Molinero, line 275: Why do we need the `_resetMarketCapacityDollarVolume` flag, can't we directly set to `0` the `_marketCapacityDollarVolume` here?

Engine/Results/SymbolCapacity.cs
```diff
@@ -0,0 +1,279 @@
...
+            QuoteBar quote;
+            if (_algorithm.CurrentSlice.QuoteBars.TryGetValue(_symbol, out quote))
+            {
+                // Fake a tradebar for quote data using market depth as a proxy for volume
+                var volume = (quote.LastBidSize + quote.LastAskSize) / 2;
+                volume = _symbol.SecurityType == SecurityType.Forex
+                    ? _forexMinuteVolume
+                    : volume;
+
+                return new TradeBar(
```
Martin-Molinero, line 257: Feels like we need some consolidator + a rolling window

Martin-Molinero (commented):
Leaving a few requests, would be nice if `CapacityEstimate` & `SymbolCapacity` had some unit tests too

Engine/Results/SymbolCapacity.cs
Martin-Molinero: ah let's also rename these `DollarVolume` all around, since it isn't true I believe because they are in account currency?

Engine/Results/SymbolCapacity.cs
Martin-Molinero: Nevermind me didn't realize we were doing this only around an order event 👍

Engine/Results/CapacityEstimate.cs
Martin-Molinero: Maybe monitored could directly be a state in the `SymbolCapacity` and avoid having another collection?

Algorithm.CSharp/CapacityTests/MonthlyRebalanceDaily.cs
```diff
@@ -0,0 +1,107 @@
...
+            {"Profit-Loss Ratio", "0.90"},
+            {"Alpha", "-0.7"},
+            {"Beta", "-0.238"},
+            {"Annual Standard Deviation", "0.386"},
+            {"Annual Variance", "0.149"},
+            {"Information Ratio", "-0.11"},
+            {"Tracking Error", "0.712"},
+            {"Treynor Ratio", "2.442"},
+            {"Total Fees", "$38.99"},
+            {"Estimated Strategy Capacity", "$19000000.00"},
```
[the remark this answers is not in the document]
gsalaz98: Weakest link is the ticker `C` in this case. Average volume is 20 million shares traded per day for C at roughly $70 per share on average, according to https://finance.yahoo.com/quote/C/?p=C

This gives us roughly $140 million traded per day, which scaled down by the 2% multiplier in `SymbolCapacity._resolutionScaleFactor` gives us $28000000 (28 million) USD traded, roughly in the same ballpark as this figure we're getting back.

Hourly data will factor in only one hour of trading, which will lead to a lower capacity estimate, though we scale it less with the scaling factor defined in the `SymbolCapacity._resolutionScaleFactor` (5%). Same calcuation as above (but daily volume divided by 6.5 (hours in market)), we get back roughly $10000000 USD in market capacity, which is also close to the result we have in the hourly data.

Engine/Results/CapacityEstimate.cs
gsalaz98: We have multiple collections to avoid having to perform an O(n) lookup whenever we're wanting to check whether a particular SymbolData instance is being "monitored" (i.e. it is currently aggregating market dollar volume for its capacity calculation). This way we ensure that we're performing a O(1) lookup, which would prevent any sort of looping.

Engine/Results/SymbolCapacity.cs
gsalaz98: I believe we want to maintain the capacity calculation up until a new order comes in, in case that we take a snapshot in between trades. We wouldn't want the capacity to be zero, since it'd show up as the smallest asset in the snapshot logic in `CapacityEstimate` if we reset it immediately.

I'm unable to find any terminology that diverges from Dollar Volume to describe the volume in account currency unfortunately. I'll make sure the comment addresses that point however

jaredbroad: We don't want to reset it immediately; we need to only reset it if there is a future trade so we can "remember" the capacity of previous trades which occurred outside the sample window

Engine/Results/SymbolCapacity.cs
jaredbroad: Elected for a simpler 2 bar comparison since it was a fake bar, and only needed a couple of data points to estimate the average volume for the period.

Martin-Molinero (approved):

Engine/Results/CapacityEstimate.cs
jaredbroad: Just a note for posterity; it was done to avoid a look-up of a public property on the SymbolCapacity objects on each loop. So we just check if there are any monitored assets; update those, and continue if nothing to do (99%) of the time.

PR #5352, opened 2021-03-02
Implements PivotPointsHighLow indicator + tests

IlshatGaripov:
#### Description
This PR implements PivotPointsHighLow  indicator:
https://www.fidelity.com/learning-center/trading-investing/technical-analysis/technical-indicator-guide/pivot-points-high-low

#### Related Issue
Related discussions:
https://www.quantconnect.com/forum/discussion/8000/indicator-pivot-point-high-low/p1
https://www.quantconnect.com/forum/discussion/399/extended-pivot-points-indicator

#### Motivation and Context
Implements missing indicator, that can be widely used by traders

Perhaps add a message to forum if someone new addresses the thread by the topic

Tests are included

#### Types of changes

#### Checklist:


Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,248 @@
...
+        /// Invokes NewPivotPointFormed event
+        /// </summary>
+        /// <param name="pivotPointsEventArgs"></param>
+        protected virtual void OnNewPivotPointFormed(PivotPointsEventArgs pivotPointsEventArgs)
+        {
+            NewPivotPointFormed?.Invoke(this, pivotPointsEventArgs);
+        }
+
+        // Will replace an array passed by reference with a new one, containing new pivot point,
+        // where the the order of the elements will be from nearest to farthest in time
```
jaredbroad, line 162: - Does this grow forever in size? If so let's make it capped to save ram.
 - I think you're describing a rolling window here? `RollingWindow`

Tests/TestData/spy_pivot_pnt_hl.txt
```diff
@@ -0,0 +1,293 @@
+Date,Open,High,Low,Close,Volume,PPHL
+2020-01-02T14:30:00Z,323.54,324.89,322.53,324.87,59253833,0
```
jaredbroad, line 2: Where does this test data come from? Does it satisfy the LEAN requirement of a third-party export/validation?

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,248 @@
...
+    {
+        /// <summary>
+        /// High pivot point
+        /// </summary>
+        High = 1,
+
+        /// <summary>
+        /// Low pivot point
+        /// </summary>
+        Low = -1
```
jaredbroad, line 226: Please add `Both`, and `None` here too so the result of the indicator can be parsed as an enum.

Tests/TestData/spy_pivot_pnt_hl.txt
jaredbroad: It might be nice to chart these in excel to highlight the pivots working.

jaredbroad (changes requested):

IlshatGaripov:
Hi @jaredbroad  thanks for valuable comments, I have applied the fixes. As for the test data concerned - it's been scraped from the TradingView. And this is exactly the chart to reflect the testing data point: https://www.tradingview.com/chart/MeUKFmug/ . Hope it opens well. Also in recent experiments, I faced the usefulness of defining separate length high and lengh low to define the highs and low pivot points accordingly, so I am making the appropriate changes to constructor and methdos to enable that separation. Thx.

IlshatGaripov:
Hi @jaredbroad please could you review this after the last fixes . I think this is a good to go considering we are going to do extensive research using this indicator in next few month - if any bugs i will be fixing on a way - idealistic plan is to build learning algo to recognize Elliot waves on pivot points .. let me know if any more severe inaccuracies in code . Thx

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <param name="length">The length parameter here defines the number of surround bars that we compare against the current bar high and lows for the max/min </param>
+        public PivotPointsHighLow(int length)
+            : this($"PivotPointsHighLow({length})", length, length)
+        { }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="PivotPointsHighLow"/> indicator
+        /// </summary>
+        /// <param name="lengthHigh">The number of surrounding bars whose high values should be less than the current one's for the high pivot point be assigned</param>
+        /// <param name="lengthLow">The number of surrounding bars whose lows values should be less than the current one's for the low pivot point be assigned</param>
```
Martin-Molinero, line 52: I think `should be less` should be higher/bigger? Same bellow

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+            var middlePoint2 = window[_lengthLow];
+            var isLow = false;
+            for (var k = 0; k < (2 * _lengthLow + 1); k++)
+            {
+                if (k == _lengthLow) continue;
+
+                isLow = window[k].Low > middlePoint2.Low;
+
+                if (!isLow) break;
+            }
```
Martin-Molinero, line 98: Think it would be nice to unit test this logic, could create a test `PivotPointsHighLow` which exposes this method to call it directly and assert the behavior. 
I find this `var middlePoint2 = window[_lengthLow];` a bit suspicious it's indirectly relating the high & low length through the window period. For example, what if `_lengthHigh` is 1000 & `_lengthLow` is 1,  with `window[_lengthLow]` we would be reading old values in the window

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+    /// Pivot Points (High/Low), also known as Bar Count Reversals, indicator.
+    /// https://www.fidelity.com/learning-center/trading-investing/technical-analysis/technical-indicator-guide/pivot-points-high-low
+    /// </summary>
+    public class PivotPointsHighLow : WindowIndicator<IBaseDataBar>
+    {
+        private readonly int _lengthHigh;
+        private readonly int _lengthLow;
+
+        // Indicator will keep information of the last 100 pivot points
+        private readonly RollingWindow<PivotPoint> _pivotPoints = new RollingWindow<PivotPoint>(100);
```
Martin-Molinero, line 33: This rolling window feels kind of related to feature https://github.com/QuantConnect/Lean/issues/5407, not 100% related to the indicator itself I think. These 100 value count should probably be an argument of `PivotPointsHighLow` I think

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        /// <summary>
+        /// Get all pivot points, in the order such that first element in collection is the nearest to the present date
+        /// </summary>
+        /// <returns>An array of low and high pivot points. Ordered by time in descending order.</returns>
+        /// <remarks>Returned array can be empty if no points have been registered yet/</remarks>
+        public PivotPoint[] GetAllPivotPointsArray()
+        {
+            // Get all pivot points within rolling wind. collection as an array
+            return _pivotPoints?.Take(_pivotPoints.Count).ToArray();
```
Martin-Molinero, line 163: `_pivotPoints` can't be null? Don't think we need the null check

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    /// Pivot Points (High/Low), also known as Bar Count Reversals, indicator.
+    /// https://www.fidelity.com/learning-center/trading-investing/technical-analysis/technical-indicator-guide/pivot-points-high-low
+    /// </summary>
+    public class PivotPointsHighLow : WindowIndicator<IBaseDataBar>
```
Martin-Molinero, line 27: Missing to override `Reset` which should clear `_pivotPoints`?
Missing to add helper method at `QCAlgorithm.Indicators`?

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <param name="lengthLow">The number of surrounding bars whose lows values should be less than the current one's for the low pivot point be assigned</param>
+        public PivotPointsHighLow(string name, int lengthHigh, int lengthLow) :
+            base(name, 2 * Math.Max(lengthLow, lengthHigh) + 1)
+        {
+            _lengthHigh = lengthHigh;
+            _lengthLow = lengthLow;
+        }
+
+        /// <inheritdoc />
+        protected override decimal ComputeNextValue(IReadOnlyWindow<IBaseDataBar> window, IBaseDataBar input)
```
Martin-Molinero, line 71: nit - Sorry  `/// <inheritdoc />` isn't supported at the moment by our doc builder, let's replace it for the full `<summary>`

Indicators/PivotPointsHighLow.cs
```diff
@@ -0,0 +1,247 @@
...
+            if (isLow)
+            {
+                var point = new PivotPoint(PivotPointType.Low, middlePoint2.Low, middlePoint2.EndTime);
+                _pivotPoints.Add(point);
+                OnNewPivotPointFormed(new PivotPointsEventArgs(point));
+
+                return -1m;
+            }
+
+            return 0m;
```
Martin-Molinero, line 132: Let's replace returning numbers directly by returning the Enum values instead 👍

Martin-Molinero (commented):
Thanks, cool indicator! 💯 Leaving a couple of requests

IlshatGaripov:
Hi @Martin-Molinero thank you very much for review, do appreciate it - very important indicator in ongoing research :smile: Will reply to all the comments ASAP

Indicators/PivotPointsHighLow.cs
IlshatGaripov: I think null conditional operator takes care of it here

Indicators/PivotPointsHighLow.cs
IlshatGaripov: I changed to enum types but still these values require conversion to decimal due to the method signature that is being overriden

Indicators/PivotPointsHighLow.cs
IlshatGaripov: This has been fixed in 6e13947 -> Implements separate rolling windows to calculate highs and low. You are correct the logic should be separate for highs and lows, as they may vary significantly in length

Martin-Molinero (approved):

Martin-Molinero (approved):
Looks good! Pushed minor update and adding another unit test 👍

IlshatGaripov:
Hi @Martin-Molinero thank you for master patchings! :blush: glad to see it merged

PR #5375, opened 2021-03-09
Do not restrict unrequested securities by 1.0m leverage

IlshatGaripov:
#### Description
I had live crypto algorithm stalled this morning trying Submit Order
The reason is that margin information was not correctly calculated in SecurityPortfolioManager.

Consider the following scenario: 
Algorithm launched when there was a holding on BTCUSD. -> 
BTCUSD is added as unrequested security in BrokerageSetupHandler, with leverage set to 1.0m.
Later this position is being closed.
There comes some new insight, new position on BTCUSD is being opened. 
Portfolio construction model assuming that all securities have been assigned the required leverage, 
by SetSecurityInitializer(s => ... ), calculates the target 
which cannot be fulfilled, because leverage is 1.0m and algorithm considers funds not enough

#### Related Issue
NA

#### Motivation and Context
To fix my live algo issue

NA

I have logged the leverage values before and after the fix. They were as expected after the fix.

#### Types of changes

#### Checklist:


Engine/Setup/BrokerageSetupHandler.cs
```diff
@@ -461,6 +461,10 @@ private void AddUnrequestedSecurity(IAlgorithm algorithm, Symbol symbol)
             if (!algorithm.Portfolio.ContainsKey(symbol))
             {
                 var resolution = algorithm.UniverseSettings.Resolution;
+                var fillForward = algorithm.UniverseSettings.FillForward;
+                var leverage = algorithm.UniverseSettings.Leverage;
+                var extendedHours = algorithm.UniverseSettings.ExtendedMarketHours;
+
                 if (!_notifiedDefaultResolutionUsed)
                 {
                     // let's just send the message once
```
Martin-Molinero, line 470: We should update the message we send to the user 👍

Martin-Molinero (commented):
Looks good, we should add some simple unit test for `BrokerageSetupHandlerTests` we already have one for the resolution being used

Martin-Molinero (approved):

Issue #5155, opened 2021-01-12
Build Warnings Overload


C-SELLERS:
#### Expected Behavior

We ignore the warnings that are acceptable in our codebase to reduce build logs and improve visibility into new warnings and errors.

#### Actual Behavior

2000+ warnings in VS build

#### Potential Solution

Decide and then suppress acceptable warnings

Also maybe handle the warnings that are unacceptable.

#### Reproducing the Problem

Build Lean

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

C-SELLERS:
Closing this issue because of the merges that wiped a lot of the warnings. There will still be some ongoing but they are more difficult and involve some breaking changes.

C-SELLERS:
With the recent upgrade to dotnet 5 framework a lot of new warnings have been generated. It would be great to clear these up so I am going to re-open this issue.

Martin-Molinero:
In master we are currently at 
<img width="368" alt="image" src="https://github.com/QuantConnect/Lean/assets/18473240/55696c6f-2c95-4d6a-8c40-deba97a1d654">

Martin-Molinero:
Still in progress

PR #5441, opened 2021-04-01
Remove Obsolete QCAlgorithm.OnEndOfDay()

C-SELLERS:
#### Description

- Remove Obsolete `QCAlgorithm.OnEndOfDay()` functions and replace with `OnEndOfDay(Symbol)` across regressions.
- Place deprecation log message for users of obsolete EOD function.
- Refactor EOD scheduling logic to only schedule for what is implemented
- Add new python extension `GetPythonArgCount` to fetch a python methods parameter size
- Have `AlgorithmPythonWrapper` discern between which EOD is implemented via the new extension method.

#### Related Issue

#5155

#### Motivation and Context

Part of cleaning up lean #5155 

Any documentation referring to this method should already be up to date since we deprecated it a while back, but if not we should clean it up.

All test are passing
Addition of another unit test to assert EOD overload behavior in Python

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
@@ -56,14 +56,6 @@ namespace QuantConnect.Algorithm.CSharp
         }
 
         /// <summary>
-        /// Obsolete overload to be removed.
-        /// </summary>
-        public override void OnEndOfDay()
-        {
-            _onEndOfDayCallCount++;
-        }
-
-        /// <summary>
         /// We expect it to be called for the universe selected <see cref="Symbol"/>
         /// and the post initialize manually added equity <see cref="Symbol"/>
         /// </summary>
@@ -114,10 +106,6 @@ namespace QuantConnect.Algorithm.CSharp
             if (_onEndOfDayIbmCallCount != 1)
             {
                 throw new Exception($"OnEndOfDay(IBM) unexpected count call {_onEndOfDayIbmCallCount}");
-            }
-            if (_onEndOfDayCallCount != 4)
-            {
-                throw new Exception($"OnEndOfDay() unexpected count call {_onEndOfDayCallCount}");
             }
         }
 
--- a/Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
@@ -30,7 +30,6 @@ namespace QuantConnect.Algorithm.CSharp
         private int _onEndOfDaySpyCallCount;
         private int _onEndOfDayBacCallCount;
         private int _onEndOfDayIbmCallCount;
-        private int _onEndOfDayCallCount;
 
         /// <summary>
         /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
--- a/AlgorithmFactory/Python/Wrappers/AlgorithmPythonWrapper.cs
+++ b/AlgorithmFactory/Python/Wrappers/AlgorithmPythonWrapper.cs
@@ -110,14 +110,7 @@ namespace QuantConnect.AlgorithmFactory.Python.Wrappers
                             {
                                 // Since we have a EOD method implemented
                                 // Determine which one it is by inspecting its arg count
-                                int argCount;
-                                var pyArgCount = PythonEngine.ModuleFromString(Guid.NewGuid().ToString(),
-                                "from inspect import signature\n" +
-                                    "def GetArgCount(method):\n" +
-                                    "   return len(signature(method).parameters)\n"
-                                ).GetAttr("GetArgCount").Invoke(endOfDayMethod);
-                                pyArgCount.TryConvert(out argCount);
-
+                                var argCount = endOfDayMethod.GetPythonArgCount();
                                 switch (argCount)
                                 {
                                     case 0: // EOD()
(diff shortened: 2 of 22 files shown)


C-SELLERS:
Instead of completely removing the functionality we decided to remove just the usages across Lean regression examples and add a debug statement to the logs to alert users that the functionality will be removed after August 2021.

I did have to make some tweaks to the `BaseRealTimeHandler` to make the messages not show up unless they have implemented `EndOfDay` in the algorithm. 

For C# this works perfectly, discerning between 
EOD() <-Bad and will post message
EOD(Symbol) <- Good will *not* post message

For Python we currently are unable to determine the parameters of the implementation so if they implement either, the message will show up for them. 😢

jaredbroad (approved):

C-SELLERS:
Small test fixes required before merge

Algorithm.CSharp/OnEndOfDayRegressionAlgorithm.cs
```diff
@@ -55,14 +55,6 @@ public override void Initialize()
             });
         }
 
-        /// <summary>
-        /// Obsolete overload to be removed.
-        /// </summary>
-        public override void OnEndOfDay()
-        {
-            _onEndOfDayCallCount++;
```
Martin-Molinero, line 63: `_onEndOfDayCallCount` is unused now and can be removed?

Engine/RealTime/ScheduledEventFactory.cs
```diff
@@ -90,6 +90,7 @@ public static ScheduledEvent EveryAlgorithmEndOfDay(IAlgorithm algorithm, IResul
                 try
                 {
                     algorithm.OnEndOfDay();
+                    algorithm.Debug("Usage of QCAlgorithm.OnEndOfDay() without a symbol will be deprecated August 2021. Always use a symbol when overriding this method: OnEndOfDay(symbol)");
```
Martin-Molinero, line 93: I think this should be logged **once** at the start, here we are logging it on each call?

Engine/RealTime/BaseRealTimeHandler.cs
```diff
@@ -101,16 +100,30 @@ protected void Setup(DateTime start, DateTime end, Language language, DateTime?
...
+                    _implementsOnEndOfDay = true;
+                }
             }
             else if (language == Language.Python)
             {
                 var wrapper = Algorithm as AlgorithmPythonWrapper;
                 _implementsOnEndOfDaySymbol = wrapper.IsOnEndOfDayImplemented;
+
+                // Since we can't differentiate the methods in Python we have to set this to the same
+                _implementsOnEndOfDay = _implementsOnEndOfDaySymbol;
```
Martin-Molinero, line 118: When we get the python method using `GetPythonMethod` isn't there some property of it we could use to determine this? for example the argument count

Engine/RealTime/ScheduledEventFactory.cs
C-SELLERS: Gotcha, moved it to only log once when creating the event.

Engine/RealTime/BaseRealTimeHandler.cs
C-SELLERS: I found a way around this, we can use python lib inspect which is included in the python install to get the methods sig and get a count on the parameters 👍🏽

C-SELLERS:
Python now correctly discerns between the EOD method implemented by using arg count to tell the difference. Its not perfect, but for this application it is enough.

AlgorithmFactory/Python/Wrappers/AlgorithmPythonWrapper.cs
```diff
@@ -100,7 +105,29 @@ public AlgorithmPythonWrapper(string moduleName)
...
+                            if (endOfDayMethod != null)
+                            {
+                                // Since we have a EOD method implemented
+                                // Determine which one it is by inspecting its arg count
+                                int argCount;
+                                var pyArgCount = PythonEngine.ModuleFromString(Guid.NewGuid().ToString(),
+                                "from inspect import signature\n" +
+                                    "def GetArgCount(method):\n" +
+                                    "   return len(signature(method).parameters)\n"
+                                ).GetAttr("GetArgCount").Invoke(endOfDayMethod);
```
Martin-Molinero, line 118: Nice! This could be an extension method too?
What would happen if both are implemented?

C-SELLERS: If both are implemented, the latest defined takes precedence. It is Python that does this, basically overwriting the first definition with this one. Meaning no two functions can share a name with different sigs

C-SELLERS (commented):
Everything looks ready to go 👍🏽  

- Addressed reviews
- Added unit tests
- Moved GetPythonArgCount to extension method

Common/Extensions.cs
```diff
@@ -435,6 +435,24 @@ public static dynamic GetPythonMethod(this PyObject instance, string name)
             }
         }
 
+        /// <summary>
+        /// Get a python methods arg count
+        /// </summary>
+        /// <param name="method">The Python method</param>
+        /// <returns>Count of arguments</returns>
+        public static int GetPythonArgCount(this PyObject method)
```
Martin-Molinero, line 443: Suggest adding `using (Py.GIL())` think these operations will probably need it

Martin-Molinero (approved):
Nice! minor comment and should be ready to go 👍

PR #5466, opened 2021-04-07
Fix Breaking Unit Test

C-SELLERS:
#### Description

Fix breaking unit test ` DelistedEventEmitted_Equity`

#### Related Issue

N/A

#### Motivation and Context

This is a semi-regularly failing unit test. These changes are to permanently fix it

- Adjust timeout from 5 to 10 seconds
- Adjust timestep from 8 hours to 6 hours
- Log beginning and end time under Error so it appears in Travis, can be changed later but is good for debug if the issue reemerges 

Although we bumped up the timeout to 10 seconds the test has shown to pass in a range of 3-4 seconds w/ Travis. Sometimes it may take slightly longer due to VM conditions

The time step was adjusted because we believe 8 hour steps are slightly too fast for the delisting event to be processed and emitted while the `ConsumeBridge` loop advanced time very quickly. Because of this we believe the loop advanced up to `endTime` and exited before the the events were even processed and appropriate delisting events created.

#### Requires Documentation Change

N/A

#### How Has This Been Tested?

It is a test

Travis runs on this PR: 5/5 passed
Github runs on this PR: 3/3 passed

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #5471, opened 2021-04-09
LiveDataBasedDelistingEventProvider Fix 

C-SELLERS:
#### Description

This PR fixes a bug with the `LiveDataBasedDelistingEventProvider` which was actually the cause of `DelistedEventEmitted_Equity` failing regularly.

The problem was that when trying to create a `LiveDelistingEventProviderEnumerator`, using the `TryCreate()` function, it created a provider (`LiveDataBasedDelistingEventProvider`) that in construction would immediately subscribe to the given `dataQueueHandler` and start consuming data. In this instance this consumption would set the delisting date for a security as soon as it received the data to do so, but at the same time when we finally create the `LiveDelistingEventProviderEnumerator` it calls on the `Provider`'s `Initialize()` where it pulls the delisting date from a map file. This creates a race condition on the `delistingDate` value because if it sets the delisting date first, but then `Initialize()` is called, we overwrite the delisting date we just received from the datafeed.

The data from our test `DelistedEventEmitted_Equity` was emitting regularly enough, that sometimes we would have processed the delisting data before `Initialize()` was called, which then overwrites the delisting date.

Even though the delisting warning data point continued to emit over and over again in this test, there is a `IdentityDataConsolidator` feeding this provider that would recognize the point is repeated and *NOT* push it through more than once.

#### Related Issue

Related PR #5466 

#### Motivation and Context

Very annoying unit test breaking inconsistently finally resolved.

N/A

Tests all pass.

To prove this issue in master, all one needs to do is add a small `Thread.Sleep()` between 
`LiveDelistingEventProviderEnumerator.TryCreates()` creation of the `delistingEventProvider` and `enumerator`. If you do this the problem reproduce *every* time, instead of just once in awhile

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/Enumerators/DelistingEventProvider.cs
+++ b/Engine/DataFeeds/Enumerators/DelistingEventProvider.cs
@@ -48,7 +48,7 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
         /// <param name="factorFile">The factor file to use</param>
         /// <param name="mapFile">The <see cref="MapFile"/> to use</param>
         /// <param name="startTime">Start date for the data request</param>
-        public void Initialize(
+        public virtual void Initialize(
             SubscriptionDataConfig config,
             FactorFile factorFile,
             MapFile mapFile,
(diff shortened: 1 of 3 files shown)


Martin-Molinero (approved):
Nice find!

PR #5488, opened 2021-04-15
Address Missing StableCoin Pairs in Crypto Exchanges

### LINKED ISSUE #5478: EnsureCurrencyDataFeed Exception With USDC-USD
#### Expected Behavior
Do not throw exceptions for missing USDC-USD pair. Their conversion rate is 1.

#### Actual Behavior
Following exception:
```
System.ArgumentException: In order to maintain cash in USDC you are required to add a subscription for Forex pair USDCUSD or USDUSDC
```
in live mode with Coinbase Pro.

#### Reproducing the Problem
```python
class FocusedYellowFrog(QCAlgorithm):

    def Initialize(self):
        self.SetBrokerageModel(BrokerageName.GDAX, AccountType.Cash)
        self.SetStartDate(2020, 10, 11)
        self.SetCash(10000)
        self.SetCash("USDC", 0.00000036)
```

#### Checklist

C-SELLERS:
#### Description

Address issues surrounding adding stable coins to algorithms without direct trading pairs for conversion.

Even though an exchange may not offer a direct conversion from a given account currency and stablecoin (ex.  GDAX USD -> USDC) we still want to be able to override some handling of these cases. In this specific example we know at GDAX USD -> USDC is a consistent 1-1 conversion. But at another exchange such as Bitfinex there is a tradable security that reflects small variances in USDC price.

To let these both exist and not have the GDAX case throw we add a condition that if there is no currency conversion security found, we first check a new static variable `*StableCoinsWithoutPairs` in the Currency class that will tell us if a given market should allow the `Cash` (USDC or other stablecoin) to be added with a 1-1 conversion without tracking a security.

*Maybe needs a better name, something more clear? Possibly relating to the fact they are all 1-1 conversion pairs without securities, or maybe a way to add this to SPD? 

I also added two missing pairs to the GDAX SPDB for
- USDC/EUR
- USDC/GBP

Source: https://pro.coinbase.com/markets

#### Related Issue

Closes #5478 

#### Motivation and Context

Reference issue above

The problem is caused by underlying brokerages not having a direct pair for stablecoins from account currency. If they did Lean would generate a subscription to these assets and update its conversion rate as data for that pair flows through the engine.

This solution seeks to bypass the issue for stablecoins that we know hold a direct 1-1 conversion rate with the account currency.

Following this solution we will likely need to consider multi-leg conversions since there isn't a direct path on some of these exchanges.

N/A

Added a unit test that asserts behavior of this issue with a range of cases for each crypto exchange, account currency, and possible stablecoins.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -282,6 +282,14 @@ namespace QuantConnect.Securities
                 }
             }
 
+            // Special case; even if we don't have USDC-USD pair, still allow USDC cash because 1-1 with USD
+            if (Symbol == "USDC")
+            {
+                ConversionRateSecurity = null;
+                ConversionRate = 1.0m;
+                return null;
+            }
+
             // if this still hasn't been set then it's an error condition
             throw new ArgumentException($"In order to maintain cash in {Symbol} you are required to add a " +
                 $"subscription for Forex pair {Symbol}{accountCurrency} or {accountCurrency}{Symbol}"
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -282,8 +282,10 @@ namespace QuantConnect.Securities
                 }
             }
 
-            // Special case; even if we don't have USDC-USD pair, still allow USDC cash because 1-1 with USD
-            if (Symbol == "USDC")
+            // Special case for crypto markets without direct pairs (They wont be found by the above)
+            // This allows us to add cash for "StableCoins" that are 1-1 with our account currency without needing a conversion security.
+            // So far only Binance and Coinbase don't have USDCUSD pairs but can hold USDC
+            if (Currencies.StableCoinsWithoutPairs.Contains(QuantConnect.Symbol.Create(normal, SecurityType.Crypto, marketMap[SecurityType.Crypto])))
             {
                 ConversionRateSecurity = null;
                 ConversionRate = 1.0m;
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -284,7 +284,7 @@ namespace QuantConnect.Securities
 
             // Special case for crypto markets without direct pairs (They wont be found by the above)
             // This allows us to add cash for "StableCoins" that are 1-1 with our account currency without needing a conversion security.
-            // So far only Binance and Coinbase don't have USDCUSD pairs but can hold USDC
+            // Check out the StableCoinsWithoutPairs static var for those that are missing their 1-1 conversion pairs
             if (Currencies.StableCoinsWithoutPairs.Contains(QuantConnect.Symbol.Create(normal, SecurityType.Crypto, marketMap[SecurityType.Crypto])))
             {
                 ConversionRateSecurity = null;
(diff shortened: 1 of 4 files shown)


Common/Securities/Cash.cs
```diff
@@ -282,6 +282,14 @@ public void SetAmount(decimal amount)
                 }
             }
 
+            // Special case; even if we don't have USDC-USD pair, still allow USDC cash because 1-1 with USD
+            if (Symbol == "USDC")
+            {
+                ConversionRateSecurity = null;
+                ConversionRate = 1.0m;
```
StefanoRaggi, line 289: I think this wouldn't work for for users with non-USD account currency.

Shouldn't this case already be handled here?
https://github.com/QuantConnect/Lean/blob/master/Common/Securities/Cash.cs#L238-L247

Martin-Molinero (commented):
Please add some unit & regression test around the issue so we can assert the fix 👍

Common/Securities/Cash.cs
AlexCatarino: We should probably cover other [stablecoins](https://en.wikipedia.org/wiki/Stablecoin). E.g.:
- `USD`: `USDC`, `USDT`
- `EUR`: `EURS`
- `GBP`: `BGBP`

StefanoRaggi: Another option here: https://github.com/QuantConnect/Lean/issues/2060

gsalaz98: Regarding Stefano's linking to #2060, I've heard that the Bellman-Ford algorithm can be used to solve for the optimal currency conversion path (with the possibility of taking into consideration loss or gain of account equity because of non-efficient conversion rates between pairs) from point `A -> [b, [c, [d, [...]]]] -> Ω` 

This might be an avenue we might want to explore if we decide to implement a fix using that issue's description as the solution

Martin-Molinero: As Stefano noted the bigger underlying issue here is https://github.com/QuantConnect/Lean/issues/2060. We decided to add this special case handling for stable coins which don't have any real trading pair and for which #2060 wouldn't work correctly because of it. It will work only when the account currency matches too, same as the rest of the supported cases really until #2060 is fixed. Think this last part is still missing though, what about different account currencies @C-SELLERS? We should add the account currency dimension to the tests 👍

Data/symbol-properties/symbol-properties-database.csv
```diff
@@ -444,6 +444,8 @@ gdax,UMAEUR,crypto,UMA-Euro,EUR,1,0.001,0.001,UMA-EUR
 gdax,UMAGBP,crypto,UMA-British Pound,GBP,1,0.001,0.001,UMA-GBP
 gdax,UMAUSD,crypto,UMA-United States Dollar,USD,1,0.001,0.001,UMA-USD
 gdax,UNIUSD,crypto,Uniswap-United States Dollar,USD,1,0.0001,0.000001,UNI-USD
+gdax,USDCEUR,crypto,USDC-Eur,EUR,1,0.001,0.01
+gdax,USDCGBP,crypto,USDC-GBP,GBP,1,0.001,0.01
```
Martin-Molinero, line 448: Missing the last column?

Tests/Common/Securities/CashTests.cs
```diff
@@ -601,9 +601,101 @@ public void EnsureCurrencyDataFeedDoesNothingWithUnsupportedCurrency()
...
+            var cash = new Cash(stableCoin, 10m, 1m);
+            cashBook.Add(cash.Symbol, cash);
+
+            var subscriptions = new SubscriptionManager();
+            var dataManager = new DataManagerStub(TimeKeeper);
+            subscriptions.SetDataManager(dataManager);
+            var abcConfig = subscriptions.Add(Symbols.SPY, Resolution.Minute, TimeZone, TimeZone);
+            var securities = new SecurityManager(TimeKeeper);
+            securities.Add(
+                Symbols.SPY,
```
Martin-Molinero, line 617: Cool tests!
Is adding SPY a requirement for the test, why? if not, let's just remove it 💯

Martin-Molinero (approved):
Looks good, minor comments only 👍

Tests/Common/Securities/CashTests.cs
C-SELLERS: Dropped it :grin:

Issue #4520, opened 2020-06-18
Greeks Always 0 With BlackScholes and Integral Price Models


DerekMelchin:
#### Expected Behavior

All price models should return non-zero greek values.

#### Actual Behavior

BlachScholes and Integral models return zero-value greeks.

#### Potential Solution

Possibly an issue with QuantLib library.

#### Reproducing the Problem

View the logs of [this backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_56b78eff244f704a058e7cb6ef38055d.html).

#### System Information

Run in the QC web terminal. Windows 10.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

ikamanu:
Hi, what is the timeline for a fix? We are considering QC for a suite of projects, and will need to get greeks when using Black Scholes.  

TBH I'm quite surprised this doesn't seem to be a high priority issue.

DerekMelchin:
Hi ikamanu,

There is not currently an ETA for this functionality.

Best,
Derek Melchin

automaton82:
I am also surprised Black Scholes, the most common model used, can't provide the greeks for options in QC.

It's a pretty limiting problem because the greeks don't seem to ever match the brokerages or other platforms in backtest?

Shaggleford:
The following issue, once resolved, would indirectly resolve this issue for live mode.
https://github.com/QuantConnect/Lean/issues/5289

Jtcollins:
Any ETA on this one? This bug is a dealbreaker. I do like the QC platform but unfortunately, I'd have to look to alternatives without this.

automaton82:
> Any way we can get this prioritized? This bug is a dealbreaker. I do like the QC platform but unfortunately, I'd have to look to alternatives without this.

I honestly just implemented it myself. Their Greeks implementation and pricing is incredibly slow. I suggest calculating it yourself, there's lot of C# and Python Black-Scholes examples out there.

ikamanu:
[...]

PR #5495, opened 2021-04-19
Adds StandardDeviationOfReturns configurability and improves greeks warmup for Futures/Index Options

gsalaz98:
#### Description

The defaults set for the period/update frequency depending on resolution can be updated, I'd suggest reviewer(s) to review them to make sure they're reasonable defaults and sound/not buggy.

  * Makes `StandardDeviationOfReturns` configurable by users, so that
    greeks can be loaded according to user expectations and the series
    of returns that they'd like to compute for `n` periods and timespan
    of `T`, as well as resolution of the data in live mode.

  * Changes resolution to max resolution available for the default
    volatility model created for the security. Usually this only applies
    to live mode, but if creating an instance of the
    `StandardDeviationOfReturns` volatility model and no `updateFrequency`
    is provided, the resolution's time span will be used as the default
    value. Backwards compatibility for equities is maintained.

  * Changes defaults for `StandardDeviationOfReturnsVolatilityModel`
    to warmup greeks faster for other derivative asset types

  * Improves comments on `StandardDeviationOfReturns` for clarity on how
    to use the volatility model for end users

#### Related Issue

Related to #4520 - this should allow for better configurability for end users, though some educational material might be required to allow users to effectively load/warmup greeks in their algorithm
#### Motivation and Context

User experience, making sure greeks warm up faster results in less confusion in usage of this part of Lean.

Thinking out loud: we might be able to automatically warm up greeks from the get-go so that greeks are never zero. I wonder if this is an avenue we'd like to explore @Martin-Molinero @jaredbroad 

I think we should have some documentation/usage guidelines around the greeks API, since it's not very obvious that they're required to warm up (by default a month) before Greeks are non-zero.

Futures/Index Options Greeks regression algorithms, ran existing unit tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/Volatility/BaseVolatilityModel.cs
+++ b/Common/Securities/Volatility/BaseVolatilityModel.cs
@@ -17,7 +17,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Data;
+using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
+using QuantConnect.Util;
 
 namespace QuantConnect.Securities.Volatility
 {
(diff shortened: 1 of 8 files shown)


Common/Securities/Volatility/StandardDeviationOfReturnsVolatilityModel.cs
```diff
@@ -131,16 +152,25 @@ public override IEnumerable<HistoryRequest> GetHistoryRequirements(Security secu
                 configuration.DataTimeZone);
             var utcStartTime = localStartTime.ConvertToUtc(security.Exchange.TimeZone);
 
+            var bar = security.GetLastData();
+            if (bar == null)
+            {
+                bar = typeof(TradeBar).GetBaseDataInstance();
+                bar.Symbol = security.Symbol;
+            }
```
Martin-Molinero, line 160: > var bar = security.GetLastData();

This seems a bit indirect, what if there was custom data in it? Suggest getting the data type from the actual subscriptions configurations above, could check for TradeBar/QuoteBar being in it, same comment for `QCAlgorithm.Universe.cs`.

I think `RelativeStandardDeviationVolatilityModel` & `StandardDeviationOfReturnsVolatilityModel` should use the same implementation of `GetHistoryRequirements`. Think that `RelativeStandardDeviationVolatilityModel` should also allow specifying some custom resolution for the history request, should be standarized.

Martin-Molinero (commented):
Nice improvement 👍 Leaving minor requests

Common/Securities/Volatility/BaseVolatilityModel.cs
```diff
@@ -69,5 +71,68 @@ DateTime utcTime
...
+            if (SubscriptionDataConfigProvider == null)
+            {
+                throw new InvalidOperationException(
+                    "BaseVolatilityModel.GetHistoryRequirements(): " +
+                    "SubscriptionDataConfigProvider was not set."
+                );
+            }
+
+            var configurations = SubscriptionDataConfigProvider
+                .GetSubscriptionDataConfigs(security.Symbol)
```
Martin-Molinero, line 99: Let's add `OrderBy(config => config.TickType)` to be explicit and deterministic on the data type we choose for the history request, which will make it continue using `TradeBar`, if **available** in it

Martin-Molinero (approved):
Looks good! last minor request

PR #5501, opened 2021-04-22
Fix IsReady and WarmUpPeriod for RateOfChange

### LINKED ISSUE #5491: Review RateOfChange IsReady And WarmUpPeriod
#### Expected Behavior
`RateOfChange` is ready after `Period + 1` datapoints. E.g. if period is 1, then `RateOfChange` will only be ready with 2 data points.

#### Actual Behavior
`RateOfChange` is ready after `Period` datapoints. E.g. if period is 1, then `RateOfChange` is ready with 1 data point and the indicator value is zero.

#### Potential Solution
Redefine `IsReady` and `WarmUpPeriod`.

#### Reproducing the Problem
```csharp
[Test]
public void IsReadyCheck()
{
    var roc = new RateOfChange(1);
    Assert.AreEqual(2, roc.WarmUpPeriod);
    var reference = new DateTime(2021, 4, 16);
    roc.Update(reference, 2);
    Assert.False(roc.IsReady);
    roc.Update(reference.AddDays(1), 4);
    Assert.True(roc.IsReady);
    Assert.AreEqual(1, roc.Current.Value);
}
```
#### Checklist

david-acker:
#### Description
This PR fixes a bug in the <code>RateOfChange</code> indicator that caused its <code>IsReady</code> property to be true too early which resulted in the first value for indicator being zero.

Rather than being set as ready after <code>Period + 1</code> datapoints, <code>RateOfChange</code> was incorrectly being set as ready after only <code>Period</code> datapoints. 

#### Related Issue
Closes #5491.

#### Motivation and Context
Provides the correct expected behavior for the indicator.

No.

The issue is no longer reproduced when using the test code provided in #5491 to reproduce the problem. All existing tests from <code>RateOfChangeTests</code> still pass.

#### Types of changes

#### Checklist:


Martin-Molinero (commented):
Thanks for the PR!
- There are a few unit tests failing because of this change which require some reviewing 
- We should add _new_ unit tests around this bug, asserting it's behaving as expected, like the one in the reported issue

Martin-Molinero (commented):
Nice!
Last request, let's add some unit tests for all the indicators affected so we can assert the behavior and double check it makes sense, since it wasn't until now, for the `IsReady/WarmUpPeriod`

Think `BasicTemplateAlgorithm` changed the `Estimated Strategy Capacity` in this PR?

PR #5515, opened 2021-04-30
Removes Assets that Are Not Invested and Not Tradable From Capacity Valculations

AlexCatarino:
#### Description
Removes assets that are not invested and not tradable from capacity calculations. In the previous calculations, assets that were not invested were part of the calculations until the end of the execution. 
This PR also adds the smallest capacity asset to the results to help developers understand why their algorithms have low capacity.

#### Motivation and Context
Bug fix.

Regression test. Also an Alpha in the market with a capacity of $250k being updated to $4mi.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/AddAlphaModelAlgorithm.cs
+++ b/Algorithm.CSharp/AddAlphaModelAlgorithm.cs
@@ -130,7 +130,7 @@ namespace QuantConnect.Algorithm.CSharp
             {"Tracking Error", "0.369"},
             {"Treynor Ratio", "-4.159"},
             {"Total Fees", "$14.46"},
-            {"Estimated Strategy Capacity", "$38000000.00"},
+            {"Estimated Strategy Capacity", "$39000000.00"},
             {"Fitness Score", "0.408"},
             {"Kelly Criterion Estimate", "16.438"},
             {"Kelly Criterion Probability Value", "0.314"},
--- a/Algorithm.CSharp/AddOptionContractExpiresRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/AddOptionContractExpiresRegressionAlgorithm.cs
@@ -139,7 +139,7 @@ namespace QuantConnect.Algorithm.CSharp
             {"Tracking Error", "0.058"},
             {"Treynor Ratio", "2.117"},
             {"Total Fees", "$2.00"},
-            {"Estimated Strategy Capacity", "$45000000.00"},
+            {"Estimated Strategy Capacity", "$46000000.00"},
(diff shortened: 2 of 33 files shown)


Algorithm.CSharp/BasicTemplateFuturesHistoryAlgorithm.cs
```diff
@@ -161,6 +161,7 @@ public override void OnOrderEvent(OrderEvent orderEvent)
             {"Treynor Ratio", "0"},
             {"Total Fees", "$0.00"},
             {"Estimated Strategy Capacity", "$0"},
+            {"Lowest Capacity Asset", " 0"},
```
Martin-Molinero, line 164: This guy doesn't look right? There are other similar cases bellow

Algorithm.CSharp/MeanVarianceOptimizationFrameworkAlgorithm.cs
```diff
@@ -104,6 +104,7 @@ public override void OnOrderEvent(OrderEvent orderEvent)
             {"Treynor Ratio", "3.236"},
             {"Total Fees", "$26.11"},
             {"Estimated Strategy Capacity", "$25000000.00"},
+            {"Lowest Capacity Asset", "AIG R735QTJ8XC9X"},
```
Martin-Molinero, line 107: Personally I don't like the inconsistency in the way we are showing this, basically cached symbols or not

C-SELLERS: Went ahead and used Symbol.Value so we don't rely on the cache to be set for `ToString()` format to be consistent.

I chose this over Symbol.ID.ToString() because its behavior is unique if it has an underlying and isn't as easily identifiable as the Symbol.Value.

Ex. 
A GC Futures Contract is the "Lowest Capacity Asset"
Symbol.ID.ToString() -> GC VOFJUCDY9XNH
Symbol.Value -> GC26G14

ES Future Option Contract 
Symbol.ID.ToString() -> ES 31C3JQTOYO9T0|ES XCZJLC9NOB29
Symbol.Value -> ES20H20  200320P03150000

TWX Option
TWX (AOL Previously)
Symbol.ID.ToString() -> AOL VRKS95ENLBYE|AOL R735QTJ8XC9X
Symbol.Value -> TWX   140621C00042500

As you can see the `Value` is a lot more specific on exactly what contract or option you are dealing with. Then in the case of all equities it will use the newest ticker for that where as the SID will output the original listing ticker.

Martin-Molinero (approved):

PR #5529, opened 2021-05-07
Consolidate SetupHandlers

### LINKED ISSUE #5150: Review Console and Backtesting ISetupHandler
#### Expected Behavior

- ISetupHandler implementations are not duplicated
#### Actual Behavior

- Console and Backtesting setup handlers are both used for backtesting, there is a lot of code duplication among them, what is the reason we need both, are they different?
#### Potential Solution

- Review and consolidate Console & backtesting setup handler
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

C-SELLERS:
#### Description

- Reduce duplicated code between setuphandlers by consolidating BacktestingSetupHandler and ConsoleSetupHandler

#### Related Issue

Closes #5150 

#### Motivation and Context

These two handlers were almost identical in implementation. The only important unique factors about them are their MaxOrder and MaximumRuntime variables which are set differently. The address this difference I used virtual functions `GetMaximumOrders` and `GetMaximumRuntime`.

N/A

WIP

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/Setup/BacktestingSetupHandler.cs
+++ b/Engine/Setup/BacktestingSetupHandler.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
  *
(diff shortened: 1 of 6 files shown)


C-SELLERS:
One thing that was **NOT** in `BacktestingSetupHandler` that was in `ConsoleSetupHandler` was a call to set `algorithm.SetAvailableDataTypes(GetConfiguredDataFeeds());`. I went ahead and added this to `BacktestingSetupHandler.Setup()` because it just updates the algorithms acceptable ticktypes for each security type, and if none are in the config it does nothing. Can confirm that it loads in default values anyways.

C-SELLERS:
It appears after adding `SetAvailableDataTypes` we are failing to pass all regressions, will address asap.

Engine/Setup/BaseSetupHandler.cs
```diff
@@ -152,5 +153,23 @@ public static void LoadBacktestJobAccountCurrency(IAlgorithm algorithm, Backtest
...
+        /// Get the available data feeds from config.json,
+        /// If none available, throw an error
+        /// </summary>
+        public static Dictionary<SecurityType, List<TickType>> GetConfiguredDataFeeds()
+        {
+            var dataFeedsConfigString = Config.Get("security-data-feeds");
+
+            if (!dataFeedsConfigString.IsNullOrEmpty())
+            {
+                Dictionary<SecurityType, List<TickType>> dataFeeds = new Dictionary<SecurityType, List<TickType>>();
```
Martin-Molinero, line 167: nit - don't think this line is required? The `new` resulting object will be overriden bellow

Engine/Setup/BacktestingSetupHandler.cs
```diff
@@ -262,17 +245,51 @@ public bool Setup(SetupHandlerParameters parameters)
...
+            {
+                MaxOrders = 10000;
+            }
+            else
+            {
+                MaxOrders = int.MaxValue;
+                MaximumRuntime += MaximumRuntime;
+            }
+
+            MaxOrders = job.Controls.BacktestingMaxOrders;
```
Martin-Molinero, line 271: This looks fishy, `MaxOrders` value is overriden by `job.Controls.BacktestingMaxOrders` so what's the `if/else` above for? `MaximumRuntime += MaximumRuntime;` is also fishy 😁 should be in the call to `GetMaximumRuntime()` I think

Engine/Setup/BaseSetupHandler.cs
C-SELLERS: Missed that in the refactor when I let it pass null back, ty 😁

Engine/Setup/BacktestingSetupHandler.cs
C-SELLERS: 💯

Martin-Molinero (approved):
Awesome consolidation!

PR #5552, opened 2021-05-13
Support 2-leg currency conversions

### LINKED ISSUE #2060: Support indirect/multi-leg currency conversions ( A->{Bridge Currencies}->B )

jmerle:
#### Description

This PR is a continuation of #2146. It includes the currency conversion changes in that PR applied on the current master, with some clean-up and more tests.

This PR changes the conversion rate calculation of `Cash` objects so that it considers 2-leg currency conversions on top of the 1-leg currency conversions that we currently check for. For example, if the account currency is set to ETH and LTC cash is added, the conversion rate from LTC to ETH can now be calculated using a combination of pairs like LTCUSD and ETHUSD, instead of throwing an error for there not being an LTCETH or an ETHLTC security.

`Cash.SecuritySymbol` and `Cash.ConversionRateSecurity` have been made obsolete because it no longer holds true that there is at most one conversion rate security. These properties still return the details of the first conversion rate security if there is one to preserve backwards compatibility. The replacement for these properties is `Cash.CurrencyConversion.ConversionRateSecurities`.

I initially tried rebasing the commits in #2146 on top of the current master, but the 3 year difference caused so many conflicts that I eventually re-applied the changes on the current master by hand instead. Despite the missing Git history, credit for some of the changes in this PR goes to @viliwonka.

#### Related Issue

Closes #2060

#### Motivation and Context

Makes it possible to use symbols that don't directly map to the account currency.

Maybe? I don't know if the current documentation says anything about the obsoleted properties.

Unit & regression tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -22,6 +22,7 @@ using QuantConnect.Data.UniverseSelection;
 using QuantConnect.Interfaces;
 using QuantConnect.Logging;
 using static QuantConnect.StringExtensions;
+using QuantConnect.Securities.CurrencyConversion;
 
 namespace QuantConnect.Securities
 {
@@ -32,7 +33,6 @@ namespace QuantConnect.Securities
     {
         private decimal _conversionRate;
         private bool _isBaseCurrency;
-        private bool _invertRealTimePrice;
 
         private readonly object _locker = new object();
 
@@ -42,19 +42,8 @@ namespace QuantConnect.Securities
         /// </summary>
         public event EventHandler Updated;
 
-        /// <summary>
-        /// Gets the symbol of the security required to provide conversion rates.
-        /// If this cash represents the account currency, then <see cref="QuantConnect.Symbol.Empty"/>
-        /// is returned
-        /// </summary>
-        public Symbol SecuritySymbol => ConversionRateSecurity?.Symbol ?? QuantConnect.Symbol.Empty;
-
-        /// <summary>
-        /// Gets the security used to apply conversion rates.
-        /// If this cash represents the account currency, then null is returned.
-        /// </summary>
         [JsonIgnore]
-        public Security ConversionRateSecurity { get; private set; }
+        public ICurrencyConversion CurrencyConversion { get; private set; }
 
         /// <summary>
         /// Gets the symbol used to represent this cash
@@ -114,16 +103,15 @@ namespace QuantConnect.Securities
         /// Updates this cash object with the specified data
         /// </summary>
         /// <param name="data">The new data for this cash object</param>
-        public void Update(BaseData data)
+        public void Update()
         {
             if (_isBaseCurrency) return;
 
-            var rate = data.Value;
-            if (_invertRealTimePrice)
+            if (CurrencyConversion != null)
             {
-                rate = 1/rate;
+                ConversionRate = CurrencyConversion.Update();
             }
-            ConversionRate = rate;
+
             OnUpdate();
         }
 
@@ -168,7 +156,7 @@ namespace QuantConnect.Securities
         /// <param name="accountCurrency">The account currency</param>
         /// <param name="defaultResolution">The default resolution to use for the internal subscriptions</param>
         /// <returns>Returns the added <see cref="SubscriptionDataConfig"/>, otherwise null</returns>
-        public SubscriptionDataConfig EnsureCurrencyDataFeed(SecurityManager securities,
+        public List<SubscriptionDataConfig> EnsureCurrencyDataFeed(SecurityManager securities,
             SubscriptionManager subscriptions,
             IReadOnlyDictionary<SecurityType, string> marketMap,
             SecurityChanges changes,
@@ -179,14 +167,14 @@ namespace QuantConnect.Securities
         {
             // this gets called every time we add securities using universe selection,
             // so must of the time we've already resolved the value and don't need to again
-            if (ConversionRateSecurity != null)
+            if (CurrencyConversion != null)
             {
                 return null;
             }
 
             if (Symbol == accountCurrency)
             {
-                ConversionRateSecurity = null;
+                CurrencyConversion = null;
                 _isBaseCurrency = true;
                 ConversionRate = 1.0m;
                 return null;
@@ -195,26 +183,11 @@ namespace QuantConnect.Securities
             // we require a security that converts this into the base currency
             string normal = Symbol + accountCurrency;
             string invert = accountCurrency + Symbol;
+
+            // existing securities
             var securitiesToSearch = securities.Select(kvp => kvp.Value)
                 .Concat(changes.AddedSecurities)
                 .Where(s => s.Type == SecurityType.Forex || s.Type == SecurityType.Cfd || s.Type == SecurityType.Crypto);
-
-            foreach (var security in securitiesToSearch)
-            {
-                if (security.Symbol.Value == normal)
-                {
-                    ConversionRateSecurity = security;
-                    return null;
-                }
-                if (security.Symbol.Value == invert)
-                {
-                    ConversionRateSecurity = security;
-                    _invertRealTimePrice = true;
-                    return null;
-                }
-            }
-
-            // if we've made it here we didn't find a security, so we'll need to add one
 
             // Create a SecurityType to Market mapping with the markets from SecurityManager members
             var markets = securities.Select(x => x.Key)
@@ -244,45 +217,9 @@ namespace QuantConnect.Securities
             {
                 // currency not found in any tradeable pair
                 Log.Error($"No tradeable pair was found for currency {Symbol}, conversion rate to account currency ({accountCurrency}) will be set to zero.");
-                ConversionRateSecurity = null;
+                CurrencyConversion = null;
                 ConversionRate = 0m;
                 return null;
-            }
-
-            var potentials = potentialEntries
-                .Select(x => QuantConnect.Symbol.Create(x.Key.Symbol, x.Key.SecurityType, x.Key.Market));
-
-            var minimumResolution = subscriptions.Subscriptions.Select(x => x.Resolution).DefaultIfEmpty(defaultResolution).Min();
-
-            foreach (var symbol in potentials)
-            {
-                if (symbol.Value == normal || symbol.Value == invert)
-                {
-                    _invertRealTimePrice = symbol.Value == invert;
-                    var securityType = symbol.ID.SecurityType;
-
-                    // use the first subscription defined in the subscription manager
-                    var type = subscriptions.LookupSubscriptionConfigDataTypes(securityType, minimumResolution, false).First();
-                    var objectType = type.Item1;
-                    var tickType = type.Item2;
-
-                    // set this as an internal feed so that the data doesn't get sent into the algorithm's OnData events
-                    var config = subscriptions.SubscriptionDataConfigService.Add(symbol,
-                        minimumResolution,
-                        fillForward: true,
-                        extendedMarketHours: false,
-                        isInternalFeed: true,
-                        subscriptionDataTypes: new List<Tuple<Type, TickType>> { new Tuple<Type, TickType>(objectType, tickType) }).First();
-
-                    var security = securityService.CreateSecurity(symbol,
-                        config,
-                        addToSymbolCache: false);
-
-                    ConversionRateSecurity = security;
-                    securities.Add(config.Symbol, security);
-                    Log.Trace($"Cash.EnsureCurrencyDataFeed(): Adding {symbol.Value} for cash {Symbol} currency feed");
-                    return config;
-                }
             }
 
             // Special case for crypto markets without direct pairs (They wont be found by the above)
@@ -291,15 +228,51 @@ namespace QuantConnect.Securities
             if (marketMap.ContainsKey(SecurityType.Crypto)
                 && Currencies.StableCoinsWithoutPairs.Contains(QuantConnect.Symbol.Create(normal, SecurityType.Crypto, marketMap[SecurityType.Crypto])))
             {
-                ConversionRateSecurity = null;
+                CurrencyConversion = null;
                 ConversionRate = 1.0m;
                 return null;
             }
 
-            // if this still hasn't been set then it's an error condition
-            throw new ArgumentException($"In order to maintain cash in {Symbol} you are required to add a " +
-                $"subscription for Forex pair {Symbol}{accountCurrency} or {accountCurrency}{Symbol}"
-            );
+            var requiredSecurities = new List<SubscriptionDataConfig>();
+
+            var potentials = potentialEntries
+                .Select(x => QuantConnect.Symbol.Create(x.Key.Symbol, x.Key.SecurityType, x.Key.Market));
+
+            var minimumResolution = subscriptions.Subscriptions.Select(x => x.Resolution).DefaultIfEmpty(defaultResolution).Min();
+
+            var makeNewSecurity = new Func<Symbol, Security>(symbol =>
+            {
+                var securityType = symbol.ID.SecurityType;
+
+                // use the first subscription defined in the subscription manager
+                var type = subscriptions.LookupSubscriptionConfigDataTypes(securityType, minimumResolution, false).First();
+                var objectType = type.Item1;
+                var tickType = type.Item2;
+
+                // set this as an internal feed so that the data doesn't get sent into the algorithm's OnData events
+                var config = subscriptions.SubscriptionDataConfigService.Add(symbol,
+                                                                             minimumResolution,
+                                                                             fillForward: true,
+                                                                             extendedMarketHours: false,
+                                                                             isInternalFeed: true,
+                                                                             subscriptionDataTypes: new List<Tuple<Type, TickType>> { new Tuple<Type, TickType>(objectType, tickType) }).First();
+
+                var newSecurity = securityService.CreateSecurity(symbol,
+                                                                 config,
+                                                                 addToSymbolCache: false);
+
+                Log.Trace($"Cash.EnsureCurrencyDataFeed(): Adding {symbol.Value} for cash {Symbol} currency feed");
+
+                securities.Add(symbol, newSecurity);
+
+                requiredSecurities.Add(config);
+
+                return newSecurity;
+            });
+
+            CurrencyConversion = SecurityCurrencyConversion.LinearSearch(Symbol, accountCurrency, securitiesToSearch, potentials, makeNewSecurity);
+
+            return requiredSecurities;
         }
 
         /// <summary>
--- a/Common/Securities/Cash.cs
+++ b/Common/Securities/Cash.cs
@@ -33,6 +33,7 @@ namespace QuantConnect.Securities
     {
         private decimal _conversionRate;
         private bool _isBaseCurrency;
+        private bool _conversionRateNeedsUpdate;
 
         private readonly object _locker = new object();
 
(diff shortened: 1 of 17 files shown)


Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
```diff
@@ -0,0 +1,268 @@
...
+        /// <exception cref="ArgumentException">Thrown when no conversion from sourceCurrency to destinationCurrency can be found</exception>
+        public static SecurityCurrencyConversion LinearSearch(
+            string sourceCurrency,
+            string destinationCurrency,
+            IList<Security> existingSecurities,
+            IEnumerable<Symbol> potentialSymbols,
+            Func<Symbol, Security> makeNewSecurity)
+        {
+            var allSymbols = existingSecurities.Select(sec => sec.Symbol).Concat(potentialSymbols)
+                .Where(x => x.SecurityType == SecurityType.Crypto || x.Value.Length == 6)
```
Martin-Molinero, line 154: > x.Value.Length == 6

hm this looks fishy?

Believe we already filter in the caller for `.Where(s => s.Type == SecurityType.Forex || s.Type == SecurityType.Cfd || s.Type == SecurityType.Crypto)` ? `potentialSymbols` is created using forex/cfd and crypto too

We could filter `allSymbols` for `if (!potentialConversionRateSymbol.PairContainsCurrency(sourceCurrency))` once I think here, and avoid doing it on each loop bellow, note that the `ToList()` call here is already looping over all of this once

Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
```diff
@@ -0,0 +1,268 @@
...
+                    Security existingSecurity2;
+                    if (securitiesBySymbol.TryGetValue(potentialConversionRateSymbol2, out existingSecurity2))
+                    {
+                        steps.Add(new Step(existingSecurity2, middleCurrency == quoteCurrency));
+                    }
+                    else
+                    {
+                        steps.Add(
+                            new Step(makeNewSecurity(potentialConversionRateSymbol2), middleCurrency == quoteCurrency));
+                    }
```
Martin-Molinero, line 258: We could add some private helper method that performs this `if/else` which we do 3 times

Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
```diff
@@ -0,0 +1,268 @@
...
+                if (lastData == null)
+                {
+                    return;
+                }
+
+                var price = lastData.Price;
+                if (price == 0m)
+                {
+                    return;
+                }
```
Martin-Molinero, line 114: If one of the rate securities does not have data and one does, what would happen? A nice unit test to add, I think it should keep a 0 value probably

Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
```diff
@@ -0,0 +1,268 @@
...
+
+                if (potentialConversionRateSymbol.CurrencyPairDual(sourceCurrency) != destinationCurrency)
+                {
+                    continue;
+                }
+
+                var steps = new List<Step>(1);
+
+                var inverted = potentialConversionRateSymbol.ComparePair(sourceCurrency, destinationCurrency) ==
```
Martin-Molinero, line 183: Seems we could replace `PairContainsCurrency/CurrencyPairDual & ComparePair` for a single call, would be nice if the 2step bellow and 1 step here followed the same pattern to determine if matched, inverted or not

Common/Util/CurrencyPairUtil.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="baseCurrency">The output base currency</param>
+        /// <param name="quoteCurrency">The output quote currency</param>
+        public static void DecomposeCurrencyPair(Symbol currencyPair, out string baseCurrency, out string quoteCurrency)
+        {
+            if (currencyPair == null)
+            {
+                throw new ArgumentException("Currency pair must not be null");
+            }
+
+            if (currencyPair.SecurityType == SecurityType.Crypto)
```
Martin-Molinero, line 44: What about CFD security types?

Common/Util/CurrencyPairUtil.cs
```diff
@@ -0,0 +1,182 @@
...
+            InverseMatch
+        }
+
+        /// <summary>
+        /// Returns how two currency pairs are related to each other
+        /// </summary>
+        /// <param name="pairA">The first pair</param>
+        /// <param name="pairB">The second pair</param>
+        /// <returns>The <see cref="Match"/> member that represents the relation between the two pairs</returns>
+        public static Match ComparePair(this Symbol pairA, Symbol pairB)
```
Martin-Molinero, line 114: This method is only used in tests, let's remove it 👍

Martin-Molinero (commented):
Nice one 💯 leaving a few requests and concerns

Common/Securities/Cash.cs
```diff
@@ -43,18 +44,27 @@ public class Cash
...
 
         /// <summary>
-        /// Gets the symbol of the security required to provide conversion rates.
+        /// Gets the symbol of the first security required to provide conversion rates.
         /// If this cash represents the account currency, then <see cref="QuantConnect.Symbol.Empty"/>
         /// is returned
         /// </summary>
-        public Symbol SecuritySymbol => ConversionRateSecurity?.Symbol ?? QuantConnect.Symbol.Empty;
+        [Obsolete("This property has been made obsolete, use the symbols on the securities in CurrencyConversion.ConversionRateSecurities instead. This property only returns the symbol of the first security used to calculate conversion rates, there may be more.")]
+        public Symbol SecuritySymbol =>
```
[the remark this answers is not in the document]
jmerle: `JsonIgnore`'d this guy, added a `SecuritySymbols` member that contains all conversion rate security symbols and is not ignored

Common/Util/CurrencyPairUtil.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// <param name="currency">The currency to look for</param>
+        /// <returns>True if currency is the base or quote currency of pair, false if not</returns>
+        public static bool PairContainsCurrency(this Symbol pair, string currency)
+        {
+            string baseCurrency;
+            string quoteCurrency;
+
+            DecomposeCurrencyPair(pair, out baseCurrency, out quoteCurrency);
+
+            return baseCurrency == currency || quoteCurrency == currency;
```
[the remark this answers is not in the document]
jmerle: Replaced with `baseCurrency.Equals(currency, StringComparison.InvariantCultureIgnoreCase)` because I think it's prettier than `string.Compare(...) == 0`, and added some tests for it

Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
jmerle: Changed it to use a conversion rate of 0 if data for any of the steps is missing

Common/Securities/Cash.cs
```diff
@@ -43,18 +44,37 @@ public class Cash
...
+        [Obsolete("This property has been made obsolete, use CurrencyConversion.ConversionRateSecurities instead.\n" +
+            "This property only returns the first security used to calculate conversion rates, there may be more.")]
+        public Security ConversionRateSecurity => CurrencyConversion?.ConversionRateSecurities.First();
+
+        /// <summary>
+        /// Gets the symbols of the securities required to provide conversion rates.
+        /// If this cash represents the account currency, then an empty enumerable is returned.
+        /// </summary>
+        public IEnumerable<Symbol> SecuritySymbols =>
+            CurrencyConversion?.ConversionRateSecurities.Select(x => x.Symbol) ?? new List<Symbol>(0);
```
Martin-Molinero, line 71: nit - can replace `new List<Symbol>(0)` for static `Enumerable.Empty<Symbol>()`

Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
```diff
@@ -0,0 +1,283 @@
...
+        /// <exception cref="ArgumentException">Thrown when no conversion from sourceCurrency to destinationCurrency can be found</exception>
+        public static SecurityCurrencyConversion LinearSearch(
+            string sourceCurrency,
+            string destinationCurrency,
+            IList<Security> existingSecurities,
+            IEnumerable<Symbol> potentialSymbols,
+            Func<Symbol, Security> makeNewSecurity)
+        {
+            var allSymbols = existingSecurities.Select(sec => sec.Symbol).Concat(potentialSymbols)
+                .Where(CurrencyPairUtil.IsDecomposable);
```
Martin-Molinero, line 149: nit - should ToList() this result since we might iterate up to 3 times through it bellow

Common/Securities/CurrencyConversion/SecurityCurrencyConversion.cs
```diff
@@ -0,0 +1,283 @@
...
+
+        /// <summary>
+        /// Checks whether a symbol converts between two currencies, and whether it does so in inverse or not
+        /// </summary>
+        /// <param name="symbol">The symbol to check for</param>
+        /// <param name="sourceCurrency">The currency to convert from</param>
+        /// <param name="destinationCurrency">The currency to convert to</param>
+        /// <param name="inverted">The parameter that contains whether the conversion is inverted, or False if the conversion is not possible with the given symbol</param>
+        /// <returns></returns>
+        private static bool SymbolConvertsCurrencies(
```
Martin-Molinero, line 239: This method is decomposing `Symbol symbol` 3 times on each call it does, we should do it just once

Martin-Molinero (commented):
Almost there, minor requests 👍

Common/Securities/Cash.cs
jmerle: This appears to be incompatible with Newtonsoft.Json. Changing it to `Enumerable.Empty<Symbol>()` causes deserialization of a `Cash` instance in `LiveResult.CashBook` in [`LiveResult_NullChartPoint_IsSkipped`](https://github.com/QuantConnect/Lean/blob/master/Tests/Report/ResultDeserializationTests.cs#L99) to fail because Newtonsoft.Json tries to cast the empty `SecuritySymbols` enumerable to an `IList`:



A similar bug was reported in Newtonsoft.Json in https://github.com/JamesNK/Newtonsoft.Json/issues/2034, but that one was fixed in Newtonsoft.Json 12.0.3, which is what all our projects appear to use. I cannot reproduce this error in an empty C# project, and debugging it inside LEAN doesn't get me anywhere, so I suggest keeping `new List<Symbol>(0)` for now as that does work.

Martin-Molinero (approved):

Engine/AlgorithmManager.cs
```diff
@@ -295,15 +295,11 @@ public void Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer syn
...
-                foreach (var kvp in algorithm.Portfolio.CashBook)
+                foreach (var cash in algorithm.Portfolio.CashBook.Values.Where(x => x.CurrencyConversion != null))
                 {
-                    var cash = kvp.Value;
-                    var updateData = cash.ConversionRateSecurity?.GetLastData();
-                    if (updateData != null)
-                    {
-                        cash.Update(updateData);
-                    }
+                    cash.Update();
```
Martin-Molinero, line 300: We could eventually avoid this call once the security itself has an update event 👍

PaoloOranges:
Hi guys, I've just merged a branch where I have my algo/strategy with master and I found that this merge breaks my algo. Is there any breaking change related to this?
the error that I'm getting from Lean is:
``20210606 21:40:20.508 ERROR:: AlgorithmManager.Run(): Portfolio value is less than or equal to zero, stopping algorithm.``

from the log I see this if it can be considered helpful.
``
20210606 21:47:25.082 TRACE:: BaseSetupHandler.SetupCurrencyConversions():
Symbol      Quantity    Conversion = Value in USD
USD: $           0.00 @       1.00 = $0.0
ETH: Ξ           0.00 @       0.00 = $0
EUR: €        1000.00 @       1.21 = $1210.00
-------------------------------------------------
CashBook Total Value:                $1210.00

20210606 21:47:25.086 TRACE:: SetUp Backtesting: User: 40630 ProjectId: 0 AlgoId: PaoloFastMinuteCryptoAlgorithm
20210606 21:47:25.086 TRACE:: Dates: Start: 04/01/2021 End: 05/15/2021 Cash: ¤1,210.00
20210606 21:47:25.088 TRACE:: BacktestingResultHandler(): Sample Period Set: 16.20
20210606 21:47:25.089 TRACE:: Time.TradeableDates(): Security Count: 3
20210606 21:47:25.091 TRACE:: Config.GetValue(): forward-console-messages - Using default value: True
20210606 21:47:25.093 TRACE:: JOB HANDLERS:
20210606 21:47:25.093 TRACE::          DataFeed:     QuantConnect.Lean.Engine.DataFeeds.FileSystemDataFeed
20210606 21:47:25.093 TRACE::          Setup:        QuantConnect.Lean.Engine.Setup.ConsoleSetupHandler
20210606 21:47:25.094 TRACE::          RealTime:     QuantConnect.Lean.Engine.RealTime.BacktestingRealTimeHandler
20210606 21:47:25.094 TRACE::          Results:      QuantConnect.Lean.Engine.Results.BacktestingResultHandler
20210606 21:47:25.094 TRACE::          Transactions: QuantConnect.Lean.Engine.TransactionHandlers.BacktestingTransactionHandler
20210606 21:47:25.095 TRACE::          Alpha:        QuantConnect.Lean.Engine.Alphas.DefaultAlphaHandler
20210606 21:47:25.095 TRACE::          ObjectStore:  QuantConnect.Lean.Engine.Storage.LocalObjectStore
20210606 21:47:25.095 TRACE::          History Provider:     QuantConnect.Lean.Engine.HistoricalData.SubscriptionDataReaderHistoryProvider
20210606 21:47:25.106 TRACE:: AlgorithmManager.Run(): Begin DataStream - Start: 01/04/2021 00:00:00 Stop: 15/05/2021 23:59:59
20210606 21:47:25.117 TRACE:: AlgorithmManager.Stream(): WarmupHistoryRequest: EURUSD 8G: Start: 25/03/2021 00:00:00 End: 01/04/2021 00:00:00 Resolution: Hour
20210606 21:47:25.117 TRACE:: AlgorithmManager.Stream(): WarmupHistoryRequest: ETHUSD XJ: Start: 25/03/2021 00:00:00 End: 01/04/2021 00:00:00 Resolution: Hour
20210606 21:47:25.117 TRACE:: AlgorithmManager.Stream(): WarmupHistoryRequest: ETHEUR: Start: 25/03/2021 00:00:00 End: 01/04/2021 00:00:00 Resolution: Hour
20210606 21:47:25.160 TRACE:: Debug: Launching analysis for PaoloFastMinuteCryptoAlgorithm with LEAN Engine v2.5.0.0
20210606 21:47:25.616 ERROR:: AlgorithmManager.Run(): Portfolio value is less than or equal to zero, stopping algorithm.
20210606 21:47:25.617 TRACE:: AlgorithmManager.Run(): Firing On End Of Algorithm...
20210606 21:47:25.618 TRACE:: Engine.Run(): Exiting Algorithm Manager
20210606 21:47:25.637 TRACE:: Debug: Algorithm warming up...
20210606 21:47:25.637 TRACE:: Log: EUR amount in Portfolio: 1000 - Initialized to Sold
20210606 21:47:25.639 TRACE:: Debug: AlgorithmManager.Run(): Portfolio value is less than or equal to zero, stopping algorithm.
20210606 21:47:25.639 TRACE:: Log: 25/03/2021 01:00:00 - TotalPortfolioValue: 0.00
25/03/2021 01:00:00 - CashBook: Symbol      Quantity    Conversion = Value in USD
USD: $           0.00 @       1.00 = $0.0
ETH: Ξ           0.00 @    1591.22 = $0.00
EUR: €        1000.00 @       0.00 = $0
-------------------------------------------------
CashBook Total Value:                $0.00

20210606 21:47:25.639 TRACE:: FileSystemDataFeed.Exit(): Start. Setting cancellation token...
20210606 21:47:25.640 TRACE:: Debug: Algorithm Id:(PaoloFastMinuteCryptoAlgorithm) completed in 0.52 seconds at 0k data points per second. Processing total of 8 data points.
``

Martin-Molinero:
Hey @PaoloOranges!
Sorry to hear that, could you provide an example algorithm reproducing the issue you are experiencing, will help debug and fix the issue much faster, since we are currently having trouble reproducing it.
Thanks!

PaoloOranges:
Hello Martin, sure. it's enough to Init the algorithm as the following:
```
public class PaoloFastMinuteCryptoAlgorithm : QCAlgorithm
    {
        private const string CryptoName = "ETH";
        private const string CurrencyName = "EUR";
        private const string SymbolName = CryptoName + CurrencyName;

        private Symbol _symbol = null;

        /// <summary>
        /// Initialise the data and resolution required, as well as the cash and start-end dates for your algorithm. All algorithms must initialized.
        /// </summary>
        {
            SetBrokerageModel(BrokerageName.GDAX, AccountType.Cash);
            SetTimeZone(NodaTime.DateTimeZone.Utc);

            Resolution resolution = Resolution.Hour;

            SetStartDate(2021, 4, 1); // Set Start Date
            SetEndDate(2021, 6, 7); // Set End Date

            SetCash(CurrencyName, 1000, 1.21m);
#if DEBUG
            SetCash("USD", 0);
#endif

            _symbol = AddCrypto(SymbolName, resolution, Market.GDAX).Symbol;
            SetBenchmark(_symbol);
            
            SetWarmUp(TimeSpan.FromDays(7));

        }

        /// <summary>
        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
        /// </summary>
        /// <param name="data">Slice object keyed by symbol containing the stock data</param>
        {

        }

        {
            if(orderEvent == null)
            {
                return;
            }

            Debug(Time + " " + orderEvent);          
        }

        {
            Log($"{Time} - TotalPortfolioValue: {Portfolio.TotalPortfolioValue}");
            Log($"{Time} - CashBook: {Portfolio.CashBook}");
        }
    }
```

jmerle:
Hi Paolo,

We were able to reproduce your issue, assuming you have local ETHUSD and ETHEUR data but no recent hourly EURUSD data. What happens is that you set the initial conversion rate from `CurrencyName` (EUR) to account currency (USD) to 1.21, but when we try to update this conversion rate there is no data for EURUSD so it resets to 0. This brings your portfolio value to 0, which causes the algorithm to crash.

Judging by the `SetCash("USD", 0)` call it looks like you want your algorithm to run like `CurrencyName` is the account currency. If this is the case, you can fix this issue by replacing the two `SetCash` calls with the following:
```csharp
SetAccountCurrency(CurrencyName);
SetCash(1000);
```

This sets the account currency of the algorithm to EUR so you no longer need EURUSD data to calculate an EUR -> USD conversion rate.

Alternatively, you can also fix this issue by making sure you have recent EURUSD data covering the timespan of the backtest. This data must have the same resolution as the fastest resolution of any of the subscribed securities (hourly in this case).

PaoloOranges:
Hi Jasper, thanks for the quick feedback.
Your assumptions are correct. I have both `ETHUSD` and `ETHEUR` but no `EURUSD` in my local data. And it's true also about `SetCash("USD", 0)`, I think it was the answer coming from a question in the forum asking how to have only Euro account, few months (or even years) ago. Basically that was a workaround.

PR #5567, opened 2021-05-17
Fixes SEC rate limiting issue by adding user-agent and reducing clustering of requests

gsalaz98:
#### Description

We were being rate-limited before when we made requests to SEC website. To prevent this from happening, SEC requires a user-agent to be set to identify ourselves. Reference: https://www.sec.gov/developer

```
To ensure that SEC.gov remains available to all users, we reserve the right to block IP addresses that submit excessive requests. The SEC does not allow "unclassified" bots or automated tools to crawl the site. Any request that has been identified as part of an unclassified bot or an automated tool outside of the acceptable policy will be managed to ensure fair access for all users. See the SEC's Web Site Privacy and Security Policy.
```

  * We add a user-agent header to all outgoing requests for SEC
    as they require it to identify the client, otherwise we're rate
    limited.

  * Use single `HttpClient` instance to make use of HTTP keep-alive and establish user-agent for requests

  * Some adjustments to how we wait for work to be done, and RateGate
    rate of requests. We wait for all requests now/don't use tasks
    anymore. Increases stability of SEC downloading

To reviewer, this is a bit of a noisy PR, but only few changes were made, just a lot of new nesting causes git to say a lot has changed.

#### Related Issue

N/A
#### Motivation and Context

Enable smooth downloading of SEC data

No

Locally. Was unable to make requests without being rate limited properly (even at 5 req/s), until I added the user-agent header. Then the SEC download went super smooth.
#### Types of changes

#### Checklist:

### DIFF
--- a/ToolBox/SECDataDownloader/SECDataDownloader.cs
+++ b/ToolBox/SECDataDownloader/SECDataDownloader.cs
@@ -25,6 +25,7 @@ using QuantConnect.Logging;
 using QuantConnect.Util;
 using System.Globalization;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading;
 
 namespace QuantConnect.ToolBox.SECDataDownloader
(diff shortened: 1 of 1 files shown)


ToolBox/SECDataDownloader/SECDataDownloader.cs
```diff
@@ -319,24 +339,30 @@ public void Download(string rawDestination, DateTime start, DateTime end)
...
+                            Log.Trace(
+                                $"SECDataDownloader.Download(): Rate limited downloading CIK-ticker mappings - retrying in 10s");
+                            Thread.Sleep(10000);
+                        }
                     }
                 }
             }
+            finally
+            {
+                _client.DisposeSafely();
```
Martin-Molinero, line 364: If we are going to dispose of the client at the end of `Download()` we should create a new one per each call at the top or we should make the class IDiposable and dispose on the client there, else two consecutive calls would break. Think this means we are not expecting consecutive calls most of the cases so we could just go for the first approach of creating the client at the top of the method 👍

Martin-Molinero (commented):
Cool fix! Minor request 👍

Martin-Molinero (approved):

PR #5582, opened 2021-05-20
CFD Capacity Estimate

### LINKED ISSUE #5553: CFD Algorithms Have Capacity of Zero
#### Expected Behavior
Non-zero capacity for all asset types even CFD that has no volume information.

#### Actual Behavior
CFD algorithms have a capacity of $0.

#### Potential Solution
Use Forex assumption.

#### Reproducing the Problem
```python
class SquareAsparagusGull(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2020, 11, 12)
        self.AddCfd("XAUUSD", market = Market.Oanda)

    def OnData(self, data):
            self.SetHoldings("XAUUSD", 1)
```
#### Checklist

C-SELLERS:
#### Description

- Use forex method of setting volume for estimation on CFDs
- Update regression

#### Related Issue

Closes #5553 

#### Motivation and Context

CFD algorithms had 0 estimated capacity, this pr addresses that by using a static volume for estimation.

N/A

Regressions all passing (1 updated)

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CfdTimeZonesRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/CfdTimeZonesRegressionAlgorithm.cs
@@ -95,7 +95,7 @@ namespace QuantConnect.Algorithm.CSharp
             {"Tracking Error", "0.006"},
             {"Treynor Ratio", "0"},
             {"Total Fees", "$0.00"},
-            {"Estimated Strategy Capacity", "$82000000000.00"},
+            {"Estimated Strategy Capacity", "$660000.00"},
             {"Lowest Capacity Asset", "DE30EUR 8I"},
             {"Fitness Score", "0.002"},
             {"Kelly Criterion Estimate", "0"},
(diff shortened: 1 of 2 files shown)


Algorithm.CSharp/CfdTimeZonesRegressionAlgorithm.cs
```diff
@@ -95,7 +95,7 @@ public override void OnData(Slice data)
             {"Tracking Error", "0.006"},
             {"Treynor Ratio", "0"},
             {"Total Fees", "$0.00"},
-            {"Estimated Strategy Capacity", "$0"},
+            {"Estimated Strategy Capacity", "$82000000000.00"},
```
Martin-Molinero, line 98: 82 billion, don't think this is accurate enough?

C-SELLERS: Yeah, its the static factor we are using for both forex and cfd
```_forexMinuteVolume = 25000000m;```

Not sure what value is best 🤔

C-SELLERS: Just looked into this a bit, apparently Forex markets makeup almost all trading volume daily explaining why the static forex volume value is so high.

Goofy picture attached below:
image

After looking at some charts for CFD with volumes I think we can use a very different number. In Fact I am seeing ~300k daily not minute like forex. Maybe I could use a unique estimation value for  CFDs like 200 per minute to get a more realistic value. (200 is roughly 300k a day broken into minutes)

PR #5600, opened 2021-05-25
Bitfinex Fee Model Fix

### LINKED ISSUE #5561: Bitfinex Fee Discount to Fill Quantity Results in Quantity That Does Not Obey Lot Size
#### Expected Behavior
The quantity of a position should have the same decimal places of the lot size.

#### Actual Behavior
In Bitfinex brokerage model, the executed amount is subtracted by the fee, since the fee may not respect the same lot size, the final fee quantity will have different decimal places. E.g:

Buy 1 NEOUSD (Lot size: 0.00000001)
1st fill: executed amount -> 0.23826819, fee -> 0.0484018001166,
    fill quantity -> 0.23826819 - 0.04840180**01166** = 0.18986638**98834**
2nd fill: executed amount ->  0.76173181, fee -> 0.1547534345196
    fill quantity -> 0.76173181 - 0.15475343**45196** = 0.60697837**54804**
0.23826819 + 0.76173181 = 1

#### Checklist

C-SELLERS:
#### Description

- Create `GetDecimalPlaces()` extension method for decimals
- Adjust Bitfinex fee model to round fee to lot size
- Adjust and add unit tests

#### Related Issue

Closes #5561

#### Motivation and Context

Incorrect modeling of Bitfinex fees which causes issues with order fill sizes. The fee should be rounded to lot size decimal length always. I verified this behavior in bitfinex by using a practice account.

Using BTC (Lot size `0.00000001`) Taker orders (0.2% fee)
Case 1:
- Order 0.00024444 
- Fee = Order * 0.002 = 0.00000048888 -> Rounded to lot size = 0.00000049
image

Case 2:
- Order 0.00021111 
- Fee = Order * 0.002 = 0.00000042222 -> Rounded to lot size = 0.00000042
image

N/A

All regressions passing,
Created a unit test to assert rounding behavior is as expected.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Orders/Fees/BitfinexFeeModel.cs
+++ b/Common/Orders/Fees/BitfinexFeeModel.cs
@@ -69,9 +69,9 @@ namespace QuantConnect.Orders.Fees
 
             unitPrice *= security.SymbolProperties.ContractMultiplier;
 
-            // Determine fee and round to lot size decimal places
-            var decimalPlaces = security.SymbolProperties.LotSize.GetDecimalPlaces();
-            var fee = Math.Round(unitPrice * order.AbsoluteQuantity * feePercentage, decimalPlaces);
+            // Determine fee and round decimal places to lotsize
+            var fee = unitPrice * order.AbsoluteQuantity * feePercentage;
+            fee -= fee % security.SymbolProperties.LotSize;
             return new OrderFee(new CashAmount(
                 fee,
                 security.QuoteCurrency.Symbol));
--- a/Common/Orders/Fees/BitfinexFeeModel.cs
+++ b/Common/Orders/Fees/BitfinexFeeModel.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
  *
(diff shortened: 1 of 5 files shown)


Common/Orders/Fees/BitfinexFeeModel.cs
```diff
@@ -67,9 +69,11 @@ public override OrderFee GetOrderFee(OrderFeeParameters parameters)
 
             unitPrice *= security.SymbolProperties.ContractMultiplier;
 
-            // apply fee factor, currently we do not model 30-day volume, so we use the first tier
+            // Determine fee and round to lot size decimal places
+            var decimalPlaces = security.SymbolProperties.LotSize.GetDecimalPlaces();
+            var fee = Math.Round(unitPrice * order.AbsoluteQuantity * feePercentage, decimalPlaces);
```
Martin-Molinero, line 74: Maybe something like `quantity -= quantity % security.SymbolProperties.LotSize;` might be simpler rounding logic? we use that approach in the rest of lean like: `BuyingPowerModel/OrderSizing`

Think we should review the live trading implementation at https://github.com/QuantConnect/Lean/blob/master/Brokerages/Bitfinex because I think the bug was originally reported for live trading and I don't think we use the feeModel directly? The fix here is still valid 👍

jaredbroad (dismissed):

C-SELLERS:
Before merging this PR we did seek out information from Bitfinex directly regarding the actually application of their fee model to see if the fee is truely rounded when applied to the user account balance. As it turns out, it is not actually rounded and values that are below the lot size are actually applied to the users balance.

Meaning that in the example in the PR description where:
- Order 0.00024444 Fee = Order * 0.002 = 0.00000048888 
- Resulting BTC = .00024444 - 0.00000049 = 0.00024395 <- Incorrect, the fee is not actually rounded as the web view shows
- Resulting BTC = .00024444 - 0.00000048888 = 0.00024395112 <- This is the actual reflected balance in Bitfinex

This is somewhat problematic because we can't place an order that includes that  .112 x 10^-8 BTC, but it IS there.

C-SELLERS:
After the findings above, we will not be merging this PR as the fee model is correct in not rounding the fee to lotsize. Lean will not throw if attempting to sell the entire BTC holdings in a case like this but just round off to lot size. 

```
Warning: Due to brokerage limitations, orders will be rounded to the nearest lot size of 0.00000001
```

PR #5656, opened 2021-06-12
Research BollingerBands Value Error

### LINKED ISSUE #4927: Research :  ValueError: cannot reindex from a duplicate axis

C-SELLERS:
#### Description

Address an issue with BollingerBands where the indicator continues to emit updates when one of its properties has gone stale because of an arithmetic error.

The error only occurs when the data is continually filled forward (or just no value change) long enough  for the standard deviation to become zero, causing the BB's PercentB to have a 0 in the denominator. When this happens PercentB no longer updates and it's current point remains the same (time and value) while BB continually emits updates.

To address it, I overrode a function that the indicator base uses to determine if a indicator should send updates `ValidateAndComputeNextValue`. Now when we know that the value will be stale (STD == 0) we will not allow BB to emit updates. 

Although it can just remain stale and BB could to continue emit updates in this condition, this can causes issues for those consuming the property values. A specific example would be the issue linked below.

Note: Also in this case where STD == 0 the other properties (mid, lower, upper bands) would all be equal so really the BB is useless and it is fair to say that it doesn't need to be emitting in this case.

#### Related Issue

Closes #4927 

#### Motivation and Context

This problem was found in Py research when using our default notebook to run SPY data through a BB indicator. In the case where only local data is available, the SPY data is filled forward after reaching the end of the included data. (As of right now 3/31/21 is the last SPY daily point) 

Once 30 (BB period) tradable days have passed, the BB's STD becomes 0, because each point was the same. (5/12/21 is the date this occurs) At this point BB's PercentB would stop updating, but BB continued to emit as it did not do anything to validate its properties are all valid.

The consumer, `QuantBook.Indicator()`, would fetch all the BB's properties values and store them for a dataframe each update it got pushed from BB. When the PercentB became stale, it still collected this value as a new value and would store it in the list to later be added to the dataframe.

But when it would create the dataframe with its collection of indicator datapoints, the percent B list would contain many instances of the same point with the same timestamp. Then pandas would throw a value error because the index (timestamp) was appearing more than once. 

N/A

All regressions passing

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/BollingerBands.cs
+++ b/Indicators/BollingerBands.cs
@@ -139,10 +139,10 @@ namespace QuantConnect.Indicators
             // Update our Indicators
             var value = ComputeNextValue(input);
 
-            // In the case STD = 0, we know that the our PercentB indicator will fail to update.
-            // This is because the denominator will be 0, when this is the case we do not want
-            // the BollingerBands to emit an update because its PercentB property will be stale.
-            return StandardDeviation.Current.Value == 0
+            // If we are ready and STD = 0, we know that the our PercentB indicator will fail to update.
+            // This is because the denominator will be 0. When this is the case we do not want the
+            // BollingerBands to emit an update because its PercentB property will be stale.
+            return IsReady && StandardDeviation.Current.Value == 0
                 ? new IndicatorResult(value, IndicatorStatus.MathError)
                 : new IndicatorResult(value);
         }
--- a/Indicators/BollingerBands.cs
+++ b/Indicators/BollingerBands.cs
@@ -139,9 +139,9 @@ namespace QuantConnect.Indicators
             // Update our Indicators
             var value = ComputeNextValue(input);
 
-            // If we are ready and STD = 0, we know that the our PercentB indicator will fail to update.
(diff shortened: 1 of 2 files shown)


Martin-Molinero (approved):
Looks good!
Where is the unit test? 😄

Issue #5636, opened 2021-06-07
Can't Trade Options on BRK.B Even Though the Data Is In Place


DerekMelchin:
#### Expected Behavior

We can trade options for securities that we have option data for.

#### Actual Behavior

In [this project](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_631d8c78dfe4b25aca9306516807ebd3.html), the notebook shows the data is in place, but it's not being attached to slice.OptionChains during a backtest. This is not the case when we replace BRKB with TSLA in the algorithm.
 It's strange because upon completion of the backtest, the algorithm logs

> completed in 1.10 seconds at 44k data points per second. Processing total of 48,771 data points.

so we know some data is being processed.

Additionally, [this backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_fb0a03b63a0b45af0aa1cff959fe21a8.html) shows that when we try to buy an option for BRK.B after adding it with the `OptionChainProvider`, we get 

> Backtest
 Handled Error: BRKB  220121P00210000: The security does not have an
accurate price as it has not yet received a bar of data. Before placing a
 trade (or using SetHoldings) warm up your algorithm with SetWarmup, or
use slice.Contains(symbol) to confirm the Slice object has price before
using the data. Data does not necessarily all arrive at the same time so
 your algorithm should confirm the data is ready before using it. In
live trading this can mean you do not have an active subscription to the
 asset class you're trying to trade. If using custom data make sure
you've set the 'Value' property

Although, this error isn't thrown when we run the same algorithm with TSLA instead of BRK.B.

#### Potential Solution

N/A

#### Reproducing the Problem

Run the backtest above.

#### System Information

QC Cloud

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

C-SELLERS:
This seems to be two problems playing into one,

The reason in research we are able to get these values is specifically because you are using "BRKB" as the ticker when adding the option. Then when you use `GetOptionHistory`, it hasn't updated that ticker to be what it is today "BRK.B"(bug 1) , "luckily" for this case our data on the other end is under "BRKB" (bug 2). So it successfully retrieves open interest data which is all you need for the history request to print out the contracts.

[...]

PR #5664, opened 2021-06-15
Option Chain Provider Symbol Map Resolve

C-SELLERS:
#### Description

Have the backtesting Option chain provider resolve and check the underlying symbols map file for different tickers to ensure that it is able to fetch data from the appropriate location

#### Related Issue

Related #5636 

#### Motivation and Context

This issue is most likely to occur in research but also in algorithm history requests that call on GetOptionContracts.

Basically if the symbol is not resolved the data key the option chain provider creates may be incorrect, leading to it failing to retrieve the data.

N/A

Tested locally with research and algorithms
Added reproducing unit tests (red/green from master/this branch)

Running regressions currently

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/DataFeeds/BacktestingOptionChainProvider.cs
+++ b/Engine/DataFeeds/BacktestingOptionChainProvider.cs
@@ -56,11 +56,20 @@ namespace QuantConnect.Lean.Engine.DataFeeds
                 throw new NotSupportedException($"BacktestingOptionChainProvider.GetOptionContractList(): SecurityType.Equity, SecurityType.Future, or SecurityType.Index is expected but was {underlyingSymbol.SecurityType}");
             }
 
-            // Resolve any mapping before requesting option contract list
-            var mapFileResolver = _mapFileProvider.Get(underlyingSymbol.ID.Market);
-            var mapFile = mapFileResolver.ResolveMapFile(underlyingSymbol.Value, date);
-            var ticker = mapFile.GetMappedSymbol(date, underlyingSymbol.Value);
-            var mappedSymbol = underlyingSymbol.UpdateMappedSymbol(ticker);
+            // Resolve any mapping before requesting option contract list for equities
+            Symbol mappedSymbol;
+            if (underlyingSymbol.SecurityType.RequiresMapping())
+            {
+                var mapFileResolver = _mapFileProvider.Get(underlyingSymbol.ID.Market);
+                var mapFile = mapFileResolver.ResolveMapFile(underlyingSymbol.Value, date);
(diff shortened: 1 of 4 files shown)


Engine/DataFeeds/BacktestingOptionChainProvider.cs
```diff
@@ -51,13 +56,19 @@ public IEnumerable<Symbol> GetOptionContractList(Symbol underlyingSymbol, DateTi
                 throw new NotSupportedException($"BacktestingOptionChainProvider.GetOptionContractList(): SecurityType.Equity, SecurityType.Future, or SecurityType.Index is expected but was {underlyingSymbol.SecurityType}");
             }
 
+            // Resolve any mapping before requesting option contract list
+            var mapFileResolver = _mapFileProvider.Get(underlyingSymbol.ID.Market);
+            var mapFile = mapFileResolver.ResolveMapFile(underlyingSymbol.Value, date);
```
Martin-Molinero, line 61: Think should send in `.ID.Symbol` here, can see other users of `.ResolveMapFile()`

Engine/DataFeeds/BacktestingOptionChainProvider.cs
```diff
@@ -51,13 +56,19 @@ public IEnumerable<Symbol> GetOptionContractList(Symbol underlyingSymbol, DateTi
                 throw new NotSupportedException($"BacktestingOptionChainProvider.GetOptionContractList(): SecurityType.Equity, SecurityType.Future, or SecurityType.Index is expected but was {underlyingSymbol.SecurityType}");
             }
 
+            // Resolve any mapping before requesting option contract list
```
Martin-Molinero, line 59: Maybe add a _why_ comment here 👍

Martin-Molinero (commented):
Cool! Minor comment, let's assert behavior for indexes/futures just in case too

Engine/DataFeeds/BacktestingOptionChainProvider.cs
gsalaz98: Adding on my review to this particular section of code, unrelated to MM's comment:

Are we sure we want to map futures/index options? 

The OptionChainProvider implementations were adjusted to generically support any kind of option chain being requested from it, regardless of SecurityType. 

Maybe we could have (or already have) some sort of extension method that checks if a Symbol or SecurityType requires mapping? 

Not a change that has to be added, but I do want to make sure we don't map Symbols for assets that don't need to be.

Martin-Molinero (approved):

PR #5679, opened 2021-06-17
Fixes Market On Open Fill of Equity Fill Model

AlexCatarino:
#### Description
Only use trade data (`Tick` with `TickTrade` type or `TradeBar`) to get the open price, since MOO is filled with the opening auction price. Ensure that this method doesn't use trade data from before the market opens for high-resolution data case.

- Change regression tests to reflect the bug fix.
  - In the `ExtendedMarketHoursHistoryRegressionAlgorithm`, MOO was filled with extended market hours.
    - On master, the algorithm places 18 market orders before the market opens. They are all converted into MOO. At 9:30 the MOO are filled, so the Portfolio.Invested condition is true and no more orders are placed.
    - In new, the algorithm places 18 market orders before the market opens. They are all converted into MOO. At 9:30, the MOC are not filled, so the algorithm places a market order (19th order) which is not converted into MOO because the market is open and it is filled. At 9:31, the MOO are filled
  - In the `RegressionAlgorithm`, MOO was filled with open prior to the order. The algorithm now has one order less, since the last MOO would need to wait another day to be filled.
  - All other changes are the consequence of using `TradeBar` instead of `QuoteBar`.

#### Related Issue
Ref.: #4567 
Closes #5129

Unit and regression tests.

#### Types of changes

#### Checklist:


Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -493,12 +495,58 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
+                }
+            }
+
+            if (localOrderTime >= endTime) return fill;
 
             // if the MOO was submitted during market the previous day, wait for a day to turn over
-            if (asset.Exchange.DateTimeIsOpen(localOrderTime) && localOrderTime.Date == asset.LocalTime.Date)
+            // The date of the order and the trade data end time cannot be the same.
+            // Note that the security local time can be ahead of the data end time.
+            if (asset.Exchange.DateTimeIsOpen(localOrderTime) && localOrderTime.Date == endTime.Date)
```
Martin-Molinero, line 549: Not sure this is right always? see from IB https://www.interactivebrokers.eu/en/index.php?f=41323
image

Martin-Molinero (commented):
Cool!

I'm thinking we could create a unit test which runs an algorithm trading SPY with MOO but running the same test with different resolutions and the resulting fill prices should be the same?
`FillForwardEnumeratorTests` does a similar approach

Common/Orders/Fills/EquityFillModel.cs
AlexCatarino: This PR will accept orders that are submitted until 9:30 (regular market closed).
See unit tests [here](https://github.com/QuantConnect/Lean/pull/5679/files#diff-18938dbae43240eab6eee79fdd2221594549c57b2b4937ba737188dd5197d9d5R690)
I think we could add the extra 2-minute, but I guess it's an edge case for Nasdaq.

Martin-Molinero: Your right! Ignore me 👍

Tests/Common/Orders/Fills/EquityFillModelTests.cs
```diff
@@ -516,54 +518,234 @@ public void PerformsStopMarketFillSell()
...
+        [TestCase(Resolution.Hour, 4, 0, 0, 8, 0)]
+        [TestCase(Resolution.Hour, 4, 8, 0, 8, 0)]
+        [TestCase(Resolution.Hour, 4, 9, 30, 8, 0, true)]
+        [TestCase(Resolution.Daily, 3, 17, 0, 8, 0)]
+        [TestCase(Resolution.Daily, 4, 0, 0, 8, 0)]
+        [TestCase(Resolution.Daily, 4, 8, 0, 8, 0)]
+        [TestCase(Resolution.Daily, 4, 9, 30, 8, 0)]
+        public void PerformsMarketOnOpenUsingOpenPriceWithDifferentOrderSubmissionDateTime(Resolution resolution, int day, int hour, int minute, int ref_hour, int ref_minute, bool nextDay = false)
+        {
+            var model = new EquityFillModel();
```
Martin-Molinero, line 704: Nice!
Would be nice to add these tests I think, where ref time is after market open too:
```
        [TestCase(Resolution.Minute, 4, 9, 30, 9, 31, true)]
        [TestCase(Resolution.Hour, 4, 9, 30, 11, 0, true)]
```


> var model = new EquityFillModel();

Don't think should be required? Can do `equity.FillModel`? Closer to the real use case too 👍

Martin-Molinero (approved):
Think the changes are solid 👍 leaving minor request

PR #5698, opened 2021-06-23
Feature add additional web subscription data sources

### LINKED ISSUE #5697: Expand subscription data source to support other web protocols such as gRPC, etc.
#### Expected Behavior
When adding custom data from the web, we are limited to rest calls. I would expect to be able to define our own stream reader that handles the communication.

#### Actual Behavior
You can only choose from an enum that creates the respective stream reader later on in the engine.

#### Potential Solution
Instead of using an enum. Change Subscription data source to an abstract class. The new abstract class will take a Func that returns a stream reader that will be called in the engine to match the current behavior of stream reader initialization in the engine. Derived classes will be created to create cleaner constructors. Derived classes are LocalFileSubscriptionDataSource, RemoteFileSubscriptionDataSource, WebSubscriptionDataSource, and RestSubscriptionDataSource. RestSubscriptionDataSource is derived from WebSubscriptionDataSource. WebSubscriptionDataSource should be used for any additional web protocols a person would want to use.

#### Checklist

bmello4688:
#### Description
Instead of using an enum in SubscriptionDataSource to create the stream reader in the engine, change Subscription data source to an abstract class. The new abstract class will take a Func that returns a stream reader that will be called in the engine to match the current behavior of stream reader initialization in the engine. Derived classes will be created to create cleaner constructors. Derived classes are LocalFileSubscriptionDataSource, RemoteFileSubscriptionDataSource, WebSubscriptionDataSource, and RestSubscriptionDataSource. RestSubscriptionDataSource is derived from WebSubscriptionDataSource. WebSubscriptionDataSource should be used for any additional web protocols a person would want to use.

#### Related Issue
Closes #5697

#### Motivation and Context
In my case, I wanted to create a custom data feed that uses a gRPC stream and I couldn't do it without modifying the lean engine. I feel I was able to abstract out the stream reading pretty nicely. 

Need to update documentation around creating custom data feeds, universes, etc. Anything that uses the basedata class to return a SubscriptionDataSource.

Unit tests and running the lean engine locally.

#### Types of changes

#### Checklist:

- [x] I have added tests to cover my changes.  Existing tests should suffice


Brokerages/QuantConnect.Brokerages.csproj
```diff
@@ -40,7 +40,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="NodaTime" Version="3.0.5" />
     <PackageReference Include="QuantConnect.IBAutomater" Version="2.0.39" />
-    <PackageReference Include="RestSharp" Version="106.6.10" />
+    <PackageReference Include="RestSharp" Version="106.11.7" />
```
Martin-Molinero, line 43: We should avoid updating dependency unless required or if it's the purpose of the PR itself

Common/Data/RestSubscriptionDataSource.cs
```diff
@@ -0,0 +1,20 @@
+using QuantConnect.Data;
```
Martin-Molinero, line 1: nit - license and documentation missing in new files

Martin-Molinero (commented):
Interesting idea! It seems like a breaking change for any implemented custom data outside of Lean though, it might be possible to make it backwards compatible. Another option could be directly adding a new `SubscriptionTransportMedium` and it's `StreamReader` implementation

bmello4688:
What do you think now? I added backwards compatibility with Obsolete attribute tags and log errors so people will get notified they should migrate over to the new way. I did not want to keep adding on to the SubscriptionDataSource object because it will end up becoming overly bloated with anything new that needs to be added to it.

Martin-Molinero:
Hey @bmello4688!
Sorry for the delay coming back to you
> I did not want to keep adding on to the SubscriptionDataSource object because it will end up becoming overly bloated with anything new that needs to be added to it.

These are core Lean engine classes really, we only have a hand full which cover most use cases we know off, and don't really change.
If you think there is some missing let's please go ahead and add it as a new `SubscriptionTransportMedium` with it's collection of tests and docs so it can be used by all too, which would be awesome! 💯
I'll close this PR for now, since it's fallen outdated since we've moved out the implementations of the custom data sources

bmello4688:
Hi Martin,

Yea I understand. It does cut through everything and I know it would be a pain to change to. That’s why I added the backwards compatibility so people could still use the old way or the new way. In my opinion, web protocols change a lot, which is why I didn’t want to just add a Grpc to the SubscriptionTransportMedium, but I guess in the future, it could always be addressed. I’ll add a new PR with gRPC support.

Brian

Martin-Molinero:
> Hi Martin,
> 
> Yea I understand. It does cut through everything and I know it would be a pain to change to. That’s why I added the backwards compatibility so people could still use the old way or the new way. In my opinion, web protocols change a lot, which is why I didn’t want to just add a Grpc to the SubscriptionTransportMedium, but I guess in the future, it could always be addressed. I’ll add a new PR with gRPC support.
> 
> Brian

Ty Brian, sounds great.

PR #5702, opened 2021-06-24
MaximumOrderQuantity Bug Fix

### LINKED ISSUE #5687: GetMaximumOrderQuantityForTargetBuyingPower Calculates Incorrect Quantity for Edge Cases
#### Expected Behavior
The calculated order quantity represents a weight that is equal to or less than the target.

#### Actual Behavior
In some edge cases, the weight is slightly higher. 

#### Potential Solution
Review the calculations. I would suggest factoring in the quantity holding only at the end.

#### Reproducing the Problem
```csharp
public void SetHoldings_LongerToLong_HighConstantFeeStructure(decimal leverage)
{
    Security msft;
    var algo = GetAlgorithm(out msft, leverage, 10000);
    //Set price to $25
    Update(msft, 25);
    //75% cash spent on 3000 MSFT shares.
    algo.Portfolio.SetCash(25000);
    algo.Portfolio[Symbols.MSFT].SetHoldings(25, 3000);
    var actual = algo.CalculateOrderQuantity(Symbols.MSFT, 0.5m);
    // TPV =  Cash + Holdings  - Fees  - Buffer => Target = TVP * 0.5
    // TPV = 25000 + 25 * 3000 - 10000 - 250 = 89750 => 89750 * 0.5 = 44875
    // Final Quantity = Target / Unit - Holdings Quantity
    // Final Quantity = 44875 / 25 - 3000 = 1795.0 - 3000 = -1205
    Assert.AreEqual(-1204m, actual);
    // 3000 - 1204 = 1796. Multiply by unit 1796 * 25 = 44900. Weight = 44900 / 89750 (TPV) = 0.50027855 > 0.5
    Assert.IsTrue(HasSufficientBuyingPowerForOrder(actual, msft, algo));
}
```
#### Checklist

C-SELLERS:
#### Description

- Fix issue where resulting order quantity would place us higher than the target holding percentage
- Refactor logic to streamline process, and hopefully make it easier to read.

#### Related Issue

Closes #5687

#### Motivation and Context

This bug was a interesting one, basically with the previous implementation it was possible to get values out of `BuyingPowerModel.GetMaximumOrderQuantityForTargetBuyingPower()` that would actually put us over our target holding percentage. This is because once we got our order margin and target margin set we would not consider the total holdings or whether the order was positive/negative for adjustments. (Negative orders would still be rounded down, when really you would want up.)

There was a couple of solutions possible, such as factoring in our quantity holding at the end of the loop and verify that it didn't put us over the margin, then adjust on the next loop. But because of the complexity of the function I worked out a refactor that simplified the process that would not require all values in absolutes, and would always converge on a value that would satisfy our overall margin portfolio target.

For a specific definition of satisfying our portfolio target:
- TargetHoldingsMargin : Amount in dollars the target holdings will cost -> TargetHoldings * UnitMargin 
- TargetMarginAllocated : Amount of portfolio dollars dedicated to this holding -> Target Percentage * (TotalPortfolioValue - RequiredFreeBuyingPower)
✅ TargetHoldingsMargin <= TargetMarginAllocated : Target holdings MUST be less than portfolio target, these values are absolutes so even if we have a short TargetMarginAllocated of -970, we require the TargetHoldingsMargin to be > -970.

It actually reduced the length of the function by ~30 lines and I think its a little more concise because we focus on the end target being the total margin allocated to this asset, which the percentage makes clear. We start by calculating a maximized value for order size that will meet this allocation, and then adjust down accordingly after calculating fees in the `do while` loop. 

I also left in the precaution for an infinite loop where it stops converging and quantity remains the same. This shouldn't happen, but better to be safe. Also there is an addition if the adjustment brings the order quantity to zero, we will immediately return 0 and the reason why.

N/A

Some tests need validation and adjustments

#### Types of changes

#### Checklist:


Common/Securities/BuyingPowerModel.cs
```diff
@@ -358,141 +358,113 @@ public virtual HasSufficientBuyingPowerForOrderResult HasSufficientBuyingPowerFo
...
-            // compute the initial order quantity
-            var orderQuantity = absFinalOrderMargin / absUnitMargin;
+            // First pass, calculate:
+            // - Final holdings quantity
+            // - Order quantity to get to our final holdings
+            // - Total target holdings margin requirement
+            var finalHoldingsQuantity = (targetMarginAllocated / unitMargin)
+                .DiscretelyRoundBy(parameters.Security.SymbolProperties.LotSize);
+            var orderQuantity = finalHoldingsQuantity - parameters.Security.Holdings.Quantity;
+            var targetHoldingsMargin = finalHoldingsQuantity * unitMargin;
```
Martin-Molinero, line 386: `finalHoldingsQuantity * unitMargin` what about lotSize for the resulting operation?

Common/Securities/BuyingPowerModel.cs
```diff
@@ -358,141 +358,113 @@ public virtual HasSufficientBuyingPowerForOrderResult HasSufficientBuyingPowerFo
...
-                string reason = null;
-                if (!parameters.SilenceNonErrorReasons)
+                string reason;
+                if (finalHoldingsQuantity == parameters.Security.Holdings.Quantity)
                 {
-                    reason = $"The order quantity is less than the lot size of {parameters.Security.SymbolProperties.LotSize} " +
-                             "and has been rounded to zero.";
+                    reason =
+                        $"Already at target holding {finalHoldingsQuantity} for {parameters.Security.Symbol.Value}" +
+                        $" Current holdings: {parameters.Security.Holdings.Quantity}";
```
Martin-Molinero, line 396: For performance we shouldn't create the string if we are not going to use it as before, this happens quite often I believe

Common/Securities/BuyingPowerModel.cs
C-SELLERS: Not needed for that guy, he is a dollar value for the target holdings margin used if we were to have finalHoldingsQuantity. 😁

Martin-Molinero:
FYI: `Failed:     7` unit tests

Tests/Common/Securities/PatternDayTradingMarginBuyingPowerModelTests.cs
```diff
@@ -86,9 +86,9 @@ public void SetLeverageTest()
         public void VerifyOpenMarketLeverage()
         {
             // Market is Open on Tuesday, Feb, 16th 2016 at Noon
-
+            // SPY @ $100 * 100 Shares / Leverage (4) = 2500
             var leverage = 4m;
-            var expected = 100 * 100m / leverage + 1;
```
C-SELLERS, line 91: These `+1`'s were from the fee ($1) that got changed to 0. 

This was intentionally changed to zero because other tests in the same set, asserted expected results that did not include calculations for a fee. Specifically `VerifyMarginCallOrderLongClosedMarket`, `VerifyMarginCallOrderShortOpenMarket`, `VerifyMarginCallOrderShortClosedMarket`, & `VerifyMarginCallOrderShortOpenMarket`.

To match their expected calculation I removed the fee from the security all these test share. 👍🏽

C-SELLERS (commented):
Ready for review 👍🏽

C-SELLERS:
Closing this because we opted for a lighter refactor version @ #5727

Issue #5721, opened 2021-06-30
Some testcasesource tests prevent all tests from running in visual studio 2019


bmello4688:
#### Expected Behavior
Should be able to run all tests from test explorer

#### Actual Behavior
tests cases exit without running

#### Potential Solution
Some of the test case source tests are causing the problem. There seems to be a mismatch between versions of Nunit and the use of test case source.

#### Reproducing the Problem
Try to run all.

#### System Information

Windows 10
VS 2019
No Nunit test extensions installed. (Using nuget)

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #5722, opened 2021-06-30
fixed problems with testcasesource

bmello4688:
#### Description
There seems to be a breaking change with the nunit test case source between versions or usage of the extensions in visual studio. The current documentation says no extensions should be used and only nuget should be used to pull down the nuget package. This experience is based on that.

#### Related Issue
#5721

#### Motivation and Context
We should be able to run all tests from the test explorer in visual studio to make sure there are no tests failing. I currently cannot do that without fixing the problem with the test case source.

No

In visual studio, I can run all tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Algorithm/AlgorithmSetHoldingsTests.cs
+++ b/Tests/Algorithm/AlgorithmSetHoldingsTests.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
  * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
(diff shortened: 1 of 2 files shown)


Martin-Molinero (commented):
Hello!
Mind expanding on what's/why the issue and how is it being solved? Referenced issue is not too detailed. Changes are dropping setting the test case name?

bmello4688:
Sorry for the lack of detail. I updated the comments a bit. I'm not entirely sure why there is a disconnect in the usage of test case source and the implementing test class.

Martin-Molinero (approved):
Your right! It's setting the name that causes the failure for some reason, ty 👍

PR #5747, opened 2021-07-07
Relative Daily Volume Indicator

### LINKED ISSUE #5746: Feature Relative Volume Indicator
from(https://www.tradingview.com/script/n0f50JKv-Relative-Volume-at-Time/)
The Relative Volume Indicator is a ratio of current volume from open to the time of day to x days prior volume. The calculation is as such: current cumulative volume from open to the time of day / cumulative average volume from open to the time of day of x previous days.

#### Checklist

jovuribe:
#### Description

Added Relative Volume Indicator. The indicator measures current cumulative volume from open to the time of day to x days prior volume.

Calculation: cumulative volume from open to the time of day / cumulative average volume from open to time of day for x prior days

#### Related Issue

Closes #5746

#### Motivation and Context

Requested

No

Ran tests on an older version of lean (NET 4.7) and passed all tests. However, no test cases for all indicators pass when I switch to NET 5.0.

Test data from TradingView: https://www.tradingview.com/script/n0f50JKv-Relative-Volume-at-Time/

#### Types of changes

#### Checklist:


ikamanu:
Excited to see this landing!

ikamanu:
Eager to get started using on this indicator. Any thoughts on when this will be reviewed & merged? Possibly this weekend?
cc: @Martin-Molinero

ikamanu:
Gentle nudge on this. It's the missing piece to a strategy I am developing.

Indicators/RelativeVolume.cs
```diff
@@ -0,0 +1,154 @@
...
+
+            if (curValue != 0) { relativeValues.Add(curValue); }
+
+            if (relativeValues.Count == 0)
+            {
+                throw new ArgumentException("Need atleast one day of data", nameof(WarmUpPeriod));
+            }
+
+            var relativeAverage = relativeValues.Average();
+            var relativeVolume = todaysTotal / relativeAverage;
```
Martin-Molinero, line 138: `relativeAverage` could be calculated only when a new day passes?

Indicators/RelativeVolume.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public int WarmUpPeriod { get; }
+
+        /// <summary>
+        /// Initializeds a new instance of the RelativeVolume class using the specified period
+        /// </summary>
+        /// <param name="period">The period over which to perform to computation</param>
+        public RelativeVolume(int period = 14)
+            : this($"RVOL({period})", period)
```
Martin-Molinero, line 50: What if the user creates the indicator with Minute resolution and a period of 10, I don't think it would behave as expected?

I think more generically the indicator could rely on the user provided period and **resolution** to determine to window time span to average volume on, right?

Martin-Molinero (commented):
Cool!
Leaving a few comments but in general concerned of the indicator relation between it's period and the fact it's implemented as a daily indicator internally not being related to the resolution used on creation.
The indicator could be adjusted to not be daily but use the resolution of the user to determine that?

Indicators/RelativeVolume.cs
jovuribe: Hi Martin, thank you for the feedback. I will start making these changes. If a user were to use a minute resolution and period of 10, an ArgumentException would be thrown stating that at least one day of data Is required. If the user were to use a minute resolution they would need a period of at least 390 minutes or one trading day. 

In terms of determining a window time span to average volume, it is based on days. Here is a drawing that @ikamanu drew clearing up a similar question I had: 

image

Martin-Molinero: Right 👍 my point was why not remove the hard coded `based on days` notion and just use the resolution the user provided as input passing it into the indicator, that could be the timespan?

>  If a user were to use a minute resolution and period of 10, an ArgumentException would be thrown stating that at least one day of data Is required

This wouldn't be very user friendly, cause also it's happening at runtime, after create the indicator while updating him?

ikamanu:
Looks good from my POV.

summerlinryan:
Looking forward to hopefully seeing this PR reviewed/accepted. A RVOL indicator seems like an excellent addition that could be used by tons of strategies.

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,141 @@
...
+                    cummulateValue += dataList[timeList.IndexOf(pair)];
+                }
+                currentDay = pair.Day;
+            }
+
+            if (cummulateValue != 0) { relativeValues.Add(cummulateValue); }
+
+            if (relativeValues.Count == 0)
+            {
+                throw new ArgumentException("Need atleast one day of data", nameof(WarmUpPeriod));
```
Martin-Molinero, line 119: Just adding it to the basic template algorithm throws this exception (`atleast` typo), don't think this is correct behavior.

Seems what would work here is holding a SMA instance per time of a day? Each new data point we get the SMA to use and once we get it's value we can update it with the current cumulative volume?

Martin-Molinero (commented):
Cool! Like the idea of defining it as a daily volume indicator 👍, still don't think the relation between the period, resolution and warmup period is correct and throwing exceptions at runtime.

Suggest using an SMA per time of the day (like Dic<Time, SMA>) to handle the historical volume average, can add these on demand, plus holding the current cumulative volume and update the previous time based SMA?
Leaving a few typo and inconsistencies findings too

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,141 @@
...
+            _rollingData.Add(input.Volume);
+
+            if (!IsReady) { return 0; }
+
+            var todaysTotal = _currentVolume.Sum();
+            var timeList = _rollingTime.ToList();
+            var dataList = _rollingData.ToList();
+            int listCount = timeList.Count;
+            var currentDay = -1;
+            int listDifference = (listCount - WarmUpPeriod) - 1;
```
Martin-Molinero, line 94: `int` -> can just use var in these cases

> if (listDifference != -1) { timeList.Skip(listDifference); }

In Lean code style usually we place each curly Braces in a new line

Indicators/RelativeDailyVolume.cs
jovuribe: It would be difficult to use an SMA as the period for the SMA would be unknown. Instead, I moved the ArgumentException to where the indicator is initialized. Please reference the method: CheckLength.

Indicators/RelativeVolume.cs
jovuribe: This needs to be calculated every time a new data point comes in. For example, if the Resolution is Hour, period is 14, and its 1:00pm. We need the "realtiveVolumes" from 2 days ago and 1 day ago from open to 1:00pm.

Martin-Molinero: > if the Resolution is Hour, period is 14, and its 1:00pm. We need the "realtiveVolumes" from 2 days ago and 1 day ago from open to 1:00pm.

Since this is the `RelativeDailyVolume` indicator I would expect period 14 to mean 14 days? As of current implementation the indicator does not know what resolution the user input. `"realtiveVolumes" from 2 days ago` not sure how '2' days ago is reached here?

> This needs to be calculated every time a new data point comes in.

What if you just kept a SMA instance per each time seen? that would avoid having to calculate anything I believe, it would all be hidden internally by the SMA?

Indicators/RelativeDailyVolume.cs
Martin-Molinero: > It would be difficult to use an SMA as the period for the SMA would be unknown.

Isn't the period of the RDV the amount of days, I'd expect it to be? If not, what is the period representing?

ikamanu:
Almost there...

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public int WarmUpPeriod { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the RelativeDailyVolume class using the specified period
+        /// </summary>
+        /// <param name="period">The period over which to perform the computation</param>
+        /// /// <param name="resolution">The resolution over which to perform to computation</param>
+        public RelativeDailyVolume(int period = 2, Resolution resolution = Resolution.Daily)
```
Martin-Molinero, line 51: `Resolution resolution = Resolution.Daily` isn't used I believe?

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,145 @@
...
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            if (input.Time.Day != _previousDay)
+            {
+                var cummulativeVolume = 0;
+                foreach (var pair in _currentData)
+                {
+                    /// Arbitrary year, month, day
+                    DateTime timeBar = new DateTime(1, 1, 1, pair.Key.Hour, pair.Key.Minute, pair.Key.Second);
+                    if (!_relativeData.ContainsKey(timeBar))
```
Martin-Molinero, line 85: Cool!
- Can replace `ContainsKey` & `[]` for single `TryGetValue`
- Believe can replace `DateTime timeBar = new DateTime(1, 1, 1, pair.Key.Hour, pair.Key.Minute, pair.Key.Second);` by directly using `pair.Key.TimeOfDay`

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                _currentData[input.Time] = input.Volume;
+            }
+
+            if (!IsReady)
+            {
+                return 0;
+            }
+
+            /// Arbitrary year, month, day
```
Martin-Molinero, line 107: nit `///` comments should use `//`

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,145 @@
...
+            DateTime currentTimeBar = new DateTime(1, 1, 1, input.Time.Hour, input.Time.Minute, input.Time.Second);
+            var denominator = 0.0m;
+            if (!_relativeData.ContainsKey(currentTimeBar))
+            {
+                /// If there is no historical data for the current time, get most recent historical data
+                /// This may come into play for crypto assets or a circuit breaker event
+                var relativeDataKeys = _relativeData.Keys.ToList();
+                relativeDataKeys.Sort((x, y) => DateTime.Compare(x, y));
```
Martin-Molinero, line 115: Can avoid the need for this sorting directly using a sorted dictionary and going through it?

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    if (relativeDataKeys[i] > currentTimeBar)
+                    {
+                        denominator = _relativeData[relativeDataKeys[i - 1]].Current.Value;
+                    }
+                }
+            }
+            else
+            {
+                denominator = _relativeData[currentTimeBar].Current.Value;
```
Martin-Molinero, line 126: Same comment about replacing ContainsKey & [] for single TryGetValue

Indicators/RelativeDailyVolume.cs
```diff
@@ -0,0 +1,145 @@
...
+                    if (!_relativeData.ContainsKey(timeBar))
+                    {
+                        _relativeData[timeBar] = new SimpleMovingAverage(WarmUpPeriod);
+                    }
+                    cummulativeVolume += (int)pair.Value;
+                    _relativeData[timeBar].Update(pair.Key, cummulativeVolume);
+                }
+                _currentData.Clear();
+                _previousDay = input.Time.Day;
+                _days += 1;
```
Martin-Molinero, line 94: Just checking the logic is expected here, `_days` is starting from -1, to reach `IsReady => _days == WarmUpPeriod;` means `WarmUpPeriod+1`?

Martin-Molinero (commented):
Almost there! Implementation is much cleaner now 👍, leaving minor comments

ikamanu:
@Martin-Molinero : thanks for the thorough rounds of review.  To save time going back and forth, would it be helpful to take longer to do an even deeper dive, one-time? That way @jovuribe  can make as many desired changes in one revision, rather than piecemeal over several rounds. 
I'm sure that's not your intended review cadence, but it might be worth being even more exhaustive in this review round.  Can you spot any other potential issues?

Martin-Molinero:
Hey @ikamanu,
> To save time going back and forth, would it be helpful to take longer to do an even deeper dive, one-time?

Each review I've done has been as deep as I could 💯, the implementation approach has changes after each review, requiring a new deep review that's why it's taken time

> Can you spot any other potential issues?

No, I think current approach looks good! We've discussed it offline too.
Current review is about improving this approach and not refactoring it as previous were 👍

jovuribe:
Thanks @Martin-Molinero and @ikamanu for sticking with me on this. Will make changes and push during my lunch.

ikamanu:
Ah, understood. @Martin-Molinero. Thanks again for the help.  
PS: We have at least one more indicator PR landing in the coming weeks --we'll apply all this feedback as well!

Indicators/RelativeDailyVolume.cs
jovuribe: Yep

Martin-Molinero (approved):

ikamanu:
Thank you!

PR #5768, opened 2021-07-15
Feature Python Virtual Environment Support

### LINKED ISSUE #3189: Add `TensorFlow 2.0` Python Library to Foundation
Add [TensorFlow 2](https://www.tensorflow.org):
`pip install tensorflow==2.1.0`

Add Tensorflow [Model Optimizatoin](https://github.com/tensorflow/model-optimization)

Add [Cvxpylayers](https://github.com/cvxgrp/cvxpylayers) (previously in #3982)

#### Checklist

### LINKED ISSUE #3950: Adds N-BEATS Python Library
#### Potential Solution
Add [N-BEATS](https://github.com/philipperemy/n-beats): Neural basis expansion analysis for interpretable time series forecasting

#### Checklist

### LINKED ISSUE #4857: Library Request: PyTorch Geometric, Deep Graph Library, and StellarGraph
#### Expected Behavior
Support [PyTorch Geometric](https://pytorch-geometric.readthedocs.io/en/latest/), [Deep Graph Library](https://github.com/dmlc/dgl), [StellarGraph](https://github.com/stellargraph/stellargraph).

#### Actual Behavior
Does not support these machine learning libraries.

#### Checklist 

C-SELLERS:
#### Description

- Add support for python virtual environments in Lean, allowing overriding of base libraries with py venv directory config value ("python-venv") for local use or job packet var ("sPythonVirtualEnvironment") for cloud.
     - Refactor AddPythonPaths to prepend to path to use for this solution
     - Add static ActivatePythonVirtualEnvironment to prepend the virtual environment libraries to PythonPath, so that they are used.

- Add "venv" virtual environment to foundation image, acts a shell install that gets mounted on later.
- Add Jupyter Kernel for this venv to allow switching to this set of libraries

#### Related Issue

Closes #3189
Closes #4857
Closes #3950
Closes https://github.com/QuantConnect/Lean/issues/6291

Regressions and unit tests ✅ 
Image testing successful using new foundation image with venv config value. ✅ 
Cloud testing ✅ 

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -3038,7 +3038,7 @@ namespace QuantConnect
         /// <returns>Substring with start removed</returns>
         public static string RemoveFromStart(this string s, string start)
         {
-            if (!s.IsNullOrEmpty() && !start.IsNullOrEmpty() && s.StartsWith(start, StringComparison.InvariantCulture))
+            if (!string.IsNullOrEmpty(s) && !string.IsNullOrEmpty(start) && s.StartsWith(start, StringComparison.InvariantCulture))
             {
                 return s.Substring(start.Length);
             }
--- a/Common/Extensions.cs
+++ b/Common/Extensions.cs
@@ -3020,9 +3020,27 @@ namespace QuantConnect
         /// <returns></returns>
         public static string RemoveFromEnd(this string s, string ending)
         {
-            if (s.EndsWith(ending))
+            if (s.EndsWith(ending, StringComparison.InvariantCulture))
             {
                 return s.Substring(0, s.Length - ending.Length);
+            }
+            else
+            {
+                return s;
+            }
+        }
+
+        /// <summary>
+        /// Returns a new string in which specified start in the current instance is removed.
+        /// </summary>
+        /// <param name="s">original string value</param>
+        /// <param name="start">the string to be removed</param>
+        /// <returns>Substring with start removed</returns>
+        public static string RemoveFromStart(this string s, string start)
+        {
+            if (!s.IsNullOrEmpty() && !start.IsNullOrEmpty() && s.StartsWith(start, StringComparison.InvariantCulture))
+            {
+                return s.Substring(start.Length);
             }
             else
             {
--- a/Common/Packets/AlgorithmNodePacket.cs
+++ b/Common/Packets/AlgorithmNodePacket.cs
@@ -112,9 +112,9 @@ namespace QuantConnect.Packets
         public string Version;
 
         /// <summary>
-        /// Virtual environment to use for PythonNet libraries
+        /// Virtual environment directory to use for PythonNet libraries
         /// </summary>
-        [JsonProperty(PropertyName = "PyVirtualEnv")]
+        [JsonProperty(PropertyName = "sPythonVirtualEnvironment")]
         public string PythonVirtualEnvironment;
 
         /// <summary>
--- a/Common/Packets/AlgorithmNodePacket.cs
+++ b/Common/Packets/AlgorithmNodePacket.cs
@@ -112,6 +112,12 @@ namespace QuantConnect.Packets
         public string Version;
 
         /// <summary>
+        /// Virtual environment to use for PythonNet libraries
+        /// </summary>
+        [JsonProperty(PropertyName = "PyVirtualEnv")]
+        public string PythonVirtualEnvironment;
+
+        /// <summary>
         /// An algorithm packet which has already been run and is being redelivered on this node.
         /// In this event we don't want to relaunch the task as it may result in unexpected behaviour for user.
         /// </summary>
--- a/Common/Packets/AlgorithmNodePacket.cs
+++ b/Common/Packets/AlgorithmNodePacket.cs
@@ -23,7 +23,7 @@ namespace QuantConnect.Packets
     /// <summary>
     /// Algorithm Node Packet is a work task for the Lean Engine
     /// </summary>
-    public class AlgorithmNodePacket : Packet
+    public class AlgorithmNodePacket : PythonEnvironmentPacket
     {
         /// <summary>
         /// Default constructor for the algorithm node:
@@ -110,12 +110,6 @@ namespace QuantConnect.Packets
         /// </summary>
         [JsonProperty(PropertyName = "sVersion")]
         public string Version;
-
-        /// <summary>
-        /// Virtual environment directory to use for PythonNet libraries
-        /// </summary>
-        [JsonProperty(PropertyName = "sPythonVirtualEnvironment")]
-        public string PythonVirtualEnvironment;
 
         /// <summary>
         /// An algorithm packet which has already been run and is being redelivered on this node.
--- a/Common/Python/PythonInitializer.cs
+++ b/Common/Python/PythonInitializer.cs
@@ -106,31 +106,38 @@ namespace QuantConnect.Python
         /// these modules must be reloaded by reload() from importlib library</remarks>
         public static void ActivatePythonVirtualEnvironment(string pathToVirtualEnv)
         {
-            if (pathToVirtualEnv != null)
+            if (pathToVirtualEnv == null)
             {
-                pathToVirtualEnv = pathToVirtualEnv.TrimEnd('/').TrimEnd('\\');
-                var pathsToPrepend = new List<string>();
-                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
-                {
-                    // For linux we need to know the python version to determine the lib folder containing our packages
-                    // Compare our PyDLL to the directory names under the lib directory and get a match
-                    var pyDll = Environment.GetEnvironmentVariable("PYTHONNET_PYDLL");
-                    var version = Path.GetFileNameWithoutExtension(pyDll);
-                    var libDir = Directory.GetDirectories($"{pathToVirtualEnv}/lib")
-                        .Select(d => new DirectoryInfo(d).Name)
-                        .First(x => version.Contains(x, StringComparison.InvariantCulture));
-
-                    pathsToPrepend.Add($"{pathToVirtualEnv}/lib/{libDir}");
-                    pathsToPrepend.Add($"{pathToVirtualEnv}/lib/{libDir}/site-packages");
-                }
-                else
-                {
-                    pathsToPrepend.Add($"{pathToVirtualEnv}\\Lib");
-                    pathsToPrepend.Add($"{pathToVirtualEnv}\\Lib\\site-packages");
-                }
-
-                AddPythonPaths(pathsToPrepend);
+                return;
             }
+
+            if(!Directory.Exists(pathToVirtualEnv)){
+                Log.Error($"PythonIntializer.ActivatePythonVirtualEnvironment(): Path {pathToVirtualEnv} to virtual environment does not exist");
+                return;
+            }
+
+            pathToVirtualEnv = pathToVirtualEnv.TrimEnd('/').TrimEnd('\\');
+            var pathsToPrepend = new List<string>();
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                // For linux we need to know the python version to determine the lib folder containing our packages
+                // Compare our PyDLL to the directory names under the lib directory and get a match
+                var pyDll = Environment.GetEnvironmentVariable("PYTHONNET_PYDLL");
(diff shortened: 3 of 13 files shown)


DockerfileLeanFoundation
```diff
@@ -265,5 +265,13 @@ RUN wget https://cdn.quantconnect.com/h2o/h2o-3.30.0.3.zip && \
 # Remove black-listed packages
 RUN pip uninstall -y s3transfer
 
+# Install a Python Virtual Environment to contain upgraded libraries for users to opt into (Pandas, Tensorflow)
+RUN python -m venv /venv && \
+    . /venv/bin/activate && \
+    pip install --upgrade pip && \
+    pip install \
+    pandas==1.1.5 \
```
Martin-Molinero, line 273: Should use `pip install --no-cache-dir`, don't think there is a need to upgrade `pip`? Name here is `venv`? Let's give it a more significant name 👍

Queues/JobQueue.cs
```diff
@@ -147,7 +148,8 @@ public AlgorithmNodePacket NextJob(out string location)
                 BacktestId = algorithmId,
                 Language = Language,
                 Parameters = parameters,
-                Controls = controls
+                Controls = controls,
+                PythonVirtualEnvironment = Config.Get("python-venv", null)
```
Martin-Molinero, line 152: Cool! think default is already `null` no need to provide it?

Launcher/Program.cs
```diff
@@ -100,6 +101,12 @@ static void Main(string[] args)
                 throw new ArgumentException(jobNullMessage);
             }
 
+            // Activate our PythonVirtualEnvironment if provided
+            if (job.PythonVirtualEnvironment != null)
```
Martin-Molinero, line 105: Just in case can use `string.IsNullOrEmpty` check

Common/Python/PythonInitializer.cs
```diff
@@ -57,31 +60,82 @@ public static void Initialize()
...
-            else
+        }
+
+        /// <summary>
+        /// "Activate" a virtual Python environment by prepending its library storage to PythonNets
+        /// path. This allows the libraries in this venv to be selected prior to our base install.
+        /// Requires PYTHONNET_PYDLL to be set to base install.
+        /// </summary>
+        /// <remarks>If a module is already loaded, Python will use its cached version first
+        /// these modules must be reloaded by reload() from importlib library</remarks>
```
Martin-Molinero, line 121: >       /// <remarks>If a module is already loaded, Python will use its cached version first
 >       /// these modules must be reloaded by reload() from importlib library</remarks>

Not sure we need to worry about supporting this use case? Should always be switch before starting? or maybe research use case 🤔

Common/Packets/AlgorithmNodePacket.cs
```diff
@@ -117,6 +117,12 @@ public string AlgorithmId
         [JsonProperty(PropertyName = "sVersion")]
         public string Version;
 
+        /// <summary>
+        /// Virtual environment to use for PythonNet libraries
+        /// </summary>
+        [JsonProperty(PropertyName = "PyVirtualEnv")]
```
Martin-Molinero, line 123: nit, think `PyVirtualEnv` should follow property name convention in the class, `sPythonVirtualEnvironment` ?

Common/Python/PythonInitializer.cs
```diff
@@ -57,31 +60,82 @@ public static void Initialize()
...
                 {
-                    _pendingPathAdditions.AddRange(paths);
+                    dynamic sys = Py.Import("sys");
+                    var locals = new PyDict();
+                    locals.SetItem("sys", sys);
+
+                    // Insert any pending start of path additions
+                    if (!_pendingStartOfPathAdditions.IsNullOrEmpty())
+                    {
+                        PythonEngine.Exec(string.Join(";", _pendingStartOfPathAdditions.Select(s => $"sys.path.insert(0, '{s}')"))
```
Martin-Molinero, line 93: Think using `;` or `:` would depend on the OS?

Common/Python/PythonInitializer.cs
```diff
@@ -57,31 +60,82 @@ public static void Initialize()
...
+                            .Replace('\\', '/'), null, locals.Handle);
 
-                    PythonEngine.Exec($"import sys;{code}");
-                    _pendingPathAdditions.Clear();
+                        _pendingEndOfPathAdditions.Clear();
+                    }
+
+                    // Update our path cache
+                    pathCache = PythonEngine.Eval("sys.path", null, locals.Handle).As<List<string>>();
+                    locals.Dispose();
```
Martin-Molinero, line 110: Would be sweet if we can directly avoid the need of the cache 👍

Martin-Molinero (commented):
Awesome! leaving minor requests

Martin-Molinero:
We can review pending libraries that were blocked because of backwards compatibility/breaking changes, can see a few at https://github.com/QuantConnect/Lean/pull/5536

DockerfileLeanFoundation
C-SELLERS: I did need to upgrade pip in order to find TF 2.5.0, but will do the rest 😁

Common/Python/PythonInitializer.cs
C-SELLERS: Yeah little grey area here.

In the use case of launching lean, our implementation calls this before python is initialized so this problem does not apply. But when I was trying to use a unit test to prove the override, when the test assembly loaded it initialized PythonNet and imported pandas somewhere in our python project.

When it got to my unit test and I called this ActivatePythonVirtualEnvironment it worked on TF, since it hadn't been imported yet, but with Pandas it used the already loaded module.

Common/Python/PythonInitializer.cs
C-SELLERS: Not applicable here because we are just modifying Pythons sys.path directly with appending/insert calls. The `;` is just to separate those calls into individual python statements in one line.

Common/Python/PythonInitializer.cs
C-SELLERS: We don't need it per say, but I noticed in testing that we were adding the same directories more than once. Seemed like a lightweight solution to keep us from  duplicating dirs on our PythonPath on accident.

Common/Python/PythonInitializer.cs
C-SELLERS: Just thought it may be nice to leave as a remark for anyone who messes around with this. 😁

Common/Python/PythonInitializer.cs
C-SELLERS: Could always get the list<strings> on the fly and remove any duplicates before appending/inserting 😁

Martin-Molinero (approved):

DockerfileLeanFoundation
```diff
@@ -265,5 +270,11 @@ RUN wget https://cdn.quantconnect.com/h2o/h2o-3.30.0.3.zip && \
 # Remove black-listed packages
 RUN pip uninstall -y s3transfer
 
+# Install a shell Python Virtual Environment, where we will mount libraries at runtime
+# Selected in Lean with directory via JSON JobPacket "sPythonVirtualEnvironment" or config "python-venv" value
+RUN python -m venv /venv --system-site-packages && . /venv/bin/activate \
+    && python -m ipykernel install --name=venv \
+    && deactivate
```
Martin-Molinero, line 277: Ah missing the ARM foundation version of these changes 👍

Martin-Molinero (changes requested):

C-SELLERS (commented):
Tested in cloud and ready to go 👍🏽

C-SELLERS:
Rebased and ready 👍🏽

Common/Extensions.cs
```diff
@@ -3029,6 +3029,24 @@ public static string RemoveFromEnd(this string s, string ending)
...
 
+        /// <summary>
+        /// Returns a new string in which specified start in the current instance is removed.
+        /// </summary>
+        /// <param name="s">original string value</param>
+        /// <param name="start">the string to be removed</param>
+        /// <returns>Substring with start removed</returns>
+        public static string RemoveFromStart(this string s, string start)
+        {
+            if (!s.IsNullOrEmpty() && !start.IsNullOrEmpty() && s.StartsWith(start, StringComparison.InvariantCulture))
```
Martin-Molinero, line 3040: nit - should use `string.IsNullOrEmpty` else it's  converting the string to an enumerable and enumerating it 😬

Common/Python/PythonInitializer.cs
```diff
@@ -59,30 +61,83 @@ public static void Initialize()
...
+        /// <summary>
+        /// "Activate" a virtual Python environment by prepending its library storage to Pythons
+        /// path. This allows the libraries in this venv to be selected prior to our base install.
+        /// Requires PYTHONNET_PYDLL to be set to base install.
+        /// </summary>
+        /// <remarks>If a module is already loaded, Python will use its cached version first
+        /// these modules must be reloaded by reload() from importlib library</remarks>
+        public static void ActivatePythonVirtualEnvironment(string pathToVirtualEnv)
+        {
+            if (pathToVirtualEnv == null)
```
Martin-Molinero, line 109: nit - maybe string.IsNullOrEmpty here too?

Common/Python/PythonInitializer.cs
```diff
@@ -59,30 +61,83 @@ public static void Initialize()
...
+            }
+
+            if(!Directory.Exists(pathToVirtualEnv)){
+                Log.Error($"PythonIntializer.ActivatePythonVirtualEnvironment(): Path {pathToVirtualEnv} to virtual environment does not exist");
+                return;
+            }
+
+            pathToVirtualEnv = pathToVirtualEnv.TrimEnd('/').TrimEnd('\\');
+            var pathsToPrepend = new List<string>();
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
Martin-Molinero, line 121: > OSPlatform.Linux

Ah wonder what about mac/OSX? 

Suggest adding a log in both `if/else` stating the actual final path added?

DockerfileLeanFoundation
```diff
@@ -82,9 +87,7 @@ RUN pip install --no-cache-dir      \
     numba==0.46                     \
     setuptools-git==1.2             \
     xarray==0.15.1                  \
-    plotly==4.7.1                   \
-    jupyterlab==3.2.6               \
-    tensorflow==1.15.2              \
+    jupyterlab==2.1.2               \
```
Martin-Molinero, line 90: whops `jupyterlab==2.1.2` doesn't seem right?

Launcher/Program.cs
```diff
@@ -69,6 +70,22 @@ static void Main(string[] args)
...
+            if (!string.IsNullOrEmpty(job.PythonVirtualEnvironment))
+            {
+                var path = "/venv";
+                Log.Trace($"Engine.Main(): Activating Python Virtual Environment at {path}");
+                PythonInitializer.ActivatePythonVirtualEnvironment(path);
+            }
+
+            // Activate our PythonVirtualEnvironment if defined in config
+            var configPythonVenv = Config.Get("python-venv");
+            if(!string.IsNullOrEmpty(configPythonVenv))
```
Martin-Molinero, line 83: nit - this should probably be an else? Maybe `job.PythonVirtualEnvironment` should just be the default value of `Config.Get("python-venv")` ?

Martin-Molinero (commented):
Awesome!

Common/Python/PythonInitializer.cs
C-SELLERS: Mac appears to be the same structure as Linux 👍🏽  Adding it to the if case, also added the log 👍🏽

Launcher/Program.cs
C-SELLERS: These guys are separate because of their use case.

The job.PythonVirtualEnvironment is really just for the cloud to flag that there is a env mounted. No matter what value it passes in, we know that the path in deployment will be "/venv".

Where as config value `python-venv` is for the local user, whos folder could live anywhere and they can define it in their config.json.

DockerfileLeanFoundation
C-SELLERS: Good catch, ty ty

Martin-Molinero (approved):

Martin-Molinero:
Partially replaced by https://github.com/QuantConnect/Lean/pull/6556

PR #5773, opened 2021-07-16
Update BinanceBrokerage to handle more than 512 symbols

### LINKED ISSUE #5794: ToolBox throws exceptions on simple data fetch from Binance
#### Expected Behavior
When using ToolBox for downloading data from Binance with the following simple arguments
```
--app=BinanceDownloader --from-date=20210701-00:00:00 --to-date=20210701-12:00:00 --resolution=Hour --tickers=BTCUSDT
```
`BinanceDataDownloader` should download and finish without any errors.

#### Actual Behavior
`BinanceBrokerage` throws two unnecessary exceptions inside `BinanceDownloaderProgram.DataDownloader` method.

The first exception throws at  line 184 of  WebSocketClientWrapper.cs
```C#
protected virtual void OnClose(WebSocketCloseData e)
{
    Log.Trace($"WebSocketClientWrapper.OnClose(): Connection closed (IsOpen:{IsOpen}, State:{_client.State}): {_url}");
    Closed?.Invoke(this, e);
}
```
`_client.State` causes the exception because _client is null.

The second exception throws at  line 358 of BinanceRestApiClient.cs
```C#
 public void StopSession()
{
    if (string.IsNullOrEmpty(SessionId))
    {
        throw new Exception("BinanceBrokerage:UserStream. listenKey wasn't allocated or has been refused.");
    }
...
```
`SessionId` is null and it should be because no call is made with API keys. But exception throws anyway.

#### Potential Solution
The first exception has a simple fix:
```c#
Log.Trace($"WebSocketClientWrapper.OnClose(): Connection closed (IsOpen:{IsOpen}, State:{_client?.State}): {_url}");
```

For the second exception may be throwing an exception is not necessary at all :
```C#
public void StopSession()
{
    if (!string.IsNullOrEmpty(SessionId))
    {
       var request = new RestRequest(UserDataStreamEndpoint, Method.DELETE);
      request.AddHeader(KeyHeader, ApiKey);
      request.AddParameter(
          "application/x-www-form-urlencoded",
          Encoding.UTF8.GetBytes($"listenKey={SessionId}"),
          ParameterType.RequestBody
      );
      ExecuteRestRequest(request);   
    }           
}
```

#### Reproducing the Problem
```
QuantConnect.ToolBox.exe --app=BinanceDownloader --from-date=20210701-00:00:00 --to-date=20210701-12:00:00 --resolution=Hour --tickers=BTCUSDT
```

There is no test available for `Brokerages`

#### System Information
N/A

#### Checklist

StefanoRaggi:
#### Description
- The `BinanceBrokerage` has been updated to use the new `BrokerageMultiWebSocketSubscriptionManager` to handle more than `512` symbols (Binance only supports `1024` subscriptions per single websocket connection).
- The `BrokerageMultiWebSocketSubscriptionManager` can be configured to distribute symbols on multiple connections based on the 24h trade count, using the `"binance-maximum-websocket-connections"` config setting (defaults to zero, where no symbol weighting is performed and additional websocket connections are created as needed).

Closes #5794 

#### Motivation and Context
- Allow `BinanceBrokerage` to support more than `512` subscribed symbols

- Tested locally, subscribing all `1503` symbols available (both with and without symbol weighting)

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
+++ b/Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using QuantConnect.Data;
 using QuantConnect.Logging;
 using QuantConnect.Util;
@@ -79,7 +80,10 @@ namespace QuantConnect.Brokerages
                 // symbol weighting enabled, create all websocket instances
                 for (var i = 0; i < _maximumWebSocketConnections; i++)
                 {
-                    _webSocketEntries.Add(new BrokerageMultiWebSocketEntry(symbolWeights, _webSocketFactory()));
+                    var webSocket = _webSocketFactory();
+                    webSocket.Open += OnOpen;
+
+                    _webSocketEntries.Add(new BrokerageMultiWebSocketEntry(symbolWeights, webSocket));
                 }
             }
         }
@@ -144,6 +148,8 @@ namespace QuantConnect.Brokerages
         /// </summary>
         private IWebSocket GetWebSocketForSymbol(Symbol symbol)
         {
+            BrokerageMultiWebSocketEntry entry;
+
             lock (_locker)
             {
                 if (_webSocketEntries.All(x => x.SymbolCount >= _maximumSymbolsPerWebSocket))
@@ -155,6 +161,8 @@ namespace QuantConnect.Brokerages
 
                     // symbol limit reached on all, create new websocket instance
                     var webSocket = _webSocketFactory();
+                    webSocket.Open += OnOpen;
+
                     _webSocketEntries.Add(new BrokerageMultiWebSocketEntry(webSocket));
                 }
 
@@ -166,19 +174,19 @@ namespace QuantConnect.Brokerages
                         ? -1
                         : Math.Sign(x.TotalWeight - y.TotalWeight));
 
-                var entry = _webSocketEntries.First();
-
-                if (!entry.WebSocket.IsOpen)
-                {
-                    Connect(entry.WebSocket);
-                }
-
-                entry.AddSymbol(symbol);
-
-                Log.Trace($"BrokerageMultiWebSocketSubscriptionManager.GetWebSocketForSymbol(): added symbol: {symbol} to websocket: {entry.WebSocket.GetHashCode()} - Count: {entry.SymbolCount}");
-
-                return entry.WebSocket;
+                entry = _webSocketEntries.First();
             }
+
+            if (!entry.WebSocket.IsOpen)
+            {
+                Connect(entry.WebSocket);
+            }
+
+            entry.AddSymbol(symbol);
+
+            Log.Trace($"BrokerageMultiWebSocketSubscriptionManager.GetWebSocketForSymbol(): added symbol: {symbol} to websocket: {entry.WebSocket.GetHashCode()} - Count: {entry.SymbolCount}");
+
+            return entry.WebSocket;
         }
 
         private void Connect(IWebSocket webSocket)
@@ -205,7 +213,7 @@ namespace QuantConnect.Brokerages
 
                 if (!connectedEvent.WaitOne(ConnectionTimeout))
                 {
-                    throw new Exception("BrokerageMultiWebSocketSubscriptionManager.Connect(): WebSocket connection timeout.");
+                    throw new TimeoutException("BrokerageMultiWebSocketSubscriptionManager.Connect(): WebSocket connection timeout.");
                 }
             }
             finally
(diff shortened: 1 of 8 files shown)


Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
```diff
@@ -0,0 +1,251 @@
...
+            {
+                connectedEvent.Set();
+            };
+
+            webSocket.Open += onOpenAction;
+
+            if (_connectionRateLimiter != null && !_connectionRateLimiter.WaitToProceed(TimeSpan.Zero))
+            {
+                _connectionRateLimiter.WaitToProceed();
+            }
```
Martin-Molinero, line 208: Could just be `_connectionRateLimiter?.WaitToProceed();` instead?

Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
```diff
@@ -0,0 +1,251 @@
...
+                {
+                    if (entry.WebSocket == webSocket && entry.Symbols.Count > 0)
+                    {
+                        Log.Trace($"BrokerageMultiWebSocketSubscriptionManager.Connect(): WebSocket opened - Resubscribing existing symbols: {entry.Symbols.Count}");
+
+                        Task.Factory.StartNew(() =>
+                        {
+                            foreach (var symbol in entry.Symbols)
+                            {
+                                _subscribeFunc(webSocket, symbol, TickType.Trade);
```
Martin-Molinero, line 243: `TickType.Trade` what about other tick types, shouldn't be hard coded I believe

Brokerages/WebSocketClientWrapper.cs
```diff
@@ -208,10 +208,15 @@ private async Task HandleConnection(CancellationTokenSource connectionCts)
...
                         var message = Encoding.UTF8.GetString(messageData.Data);
-                        OnMessage(new WebSocketMessage(message));
+                        OnMessage(new WebSocketMessage(this, message));
                     }
                 }
                 catch (OperationCanceledException) { }
+                catch (WebSocketException ex)
+                {
+                    OnError(new WebSocketError(ex.Message, ex));
+                    Thread.Sleep(2000);
```
Martin-Molinero, line 218: Could use `connectionCts.Token.WaitHandle.WaitOne(2000);`, so if cancelled exits immediately?

Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
```diff
@@ -0,0 +1,251 @@
...
+
+                entry = _webSocketEntries.First();
+            }
+
+            if (!entry.WebSocket.IsOpen)
+            {
+                Connect(entry.WebSocket);
+            }
+
+            entry.AddSymbol(symbol);
```
Martin-Molinero, line 185: Think `entry` isn't thread safe, since it uses a list internally, `AddSymbol` `RemoveSymbol` should be better wrapped by the lock, or maybe better could make the entry thread safe?

Martin-Molinero (commented):
Cool! Leaving a couple of minor comments 👍

Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
```diff
@@ -202,7 +202,7 @@ private void Connect(IWebSocket webSocket)
 
             webSocket.Open += onOpenAction;
 
-            if (_connectionRateLimiter != null && !_connectionRateLimiter.WaitToProceed(TimeSpan.Zero))
+            if (_connectionRateLimiter is { IsRateLimited: false })
```
Martin-Molinero, line 205: Interesting expression!

Martin-Molinero (approved):
Looking good, minor comment 👍

Brokerages/BrokerageMultiWebSocketSubscriptionManager.cs
```diff
@@ -0,0 +1,251 @@
...
+        /// <param name="tickType">Type of tick data</param>
+        protected override bool Subscribe(IEnumerable<Symbol> symbols, TickType tickType)
+        {
+            Log.Trace($"BrokerageMultiWebSocketSubscriptionManager.Subscribe(): {string.Join(",", symbols.Select(x => x.Value))}");
+
+            foreach (var symbol in symbols)
+            {
+                var webSocket = GetWebSocketForSymbol(symbol);
+
+                _subscribeFunc(webSocket, symbol);
```
Martin-Molinero, line 104: We should make these guy `_subscribeFunc` `_unsubscribeFunc` return a `bool` and use it to determine the result of this method, not hard coded `true` response 👍

Martin-Molinero (approved):

Brokerages/WebSocketClientWrapper.cs
```diff
@@ -94,6 +96,17 @@ public void Connect()
                             Log.Trace($"WebSocketClientWrapper connection task ended: {_url}");
                         },
                         _cts.Token);
```
Martin-Molinero, line 98: We should add `TaskCreationOptions.LongRunning` for this task 👍

Martin-Molinero (approved):

Issue #3735, opened 2019-10-16
[Bug] Fxcm and Oanda not setting Tags/Comments correctly


HeroKato-Developer:
#### Expected Behavior
Using any Order function placing, limits, stops or market orders, does not actually apply the comment/tag to the order sent.
Both Oanda and Fxcm do not work.

#### Actual Behavior
Order Get sent without any comment/tag

#### Checklist

- [X] I have completely filled out this template
- [X] I have confirmed that this issue exists on the current `master` branch
- [X] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [X] I have provided detailed steps to reproduce the issue

jaredbroad:
Duplicate of https://github.com/QuantConnect/Lean/issues/5777

PR #5778, opened 2021-07-17
Fix to restore orders tags on restart in live mode

### LINKED ISSUE #5777: Custom orders tags are not restored when the algorithm is restarted
#### Expected Behavior
When algorithm is restarted is live mode and there are any open orders remain, submitted with custom tag, these order tags are are restored and be the same as user defined.

#### Actual Behavior
Brokerage implementations restore the open orders on restart but the tag field of these orders is not updated accordingly - it is not the same as user assigned.

#### Potential Solution
To be submitted in separate PR

IlshatGaripov:
#### Description
User can store some critical information in order tags, as an example we currently have algorithm in development containing some scenarios triggered on order filled action. Stored in order tags. If algorithm in live mode accidently stops - because of the bug or to deploy a newer version - it's important that orders tags for existing orders are preserved.

_Sorry I am branching from the last commit in master supporting Net Framework. We use Lean as part of larger solution . in which migration to .NET is in still progress_

#### Related Issue
Fixes #5777 

Includes tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
+++ b/Engine/TransactionHandlers/BrokerageTransactionHandler.cs
@@ -16,14 +16,18 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Threading;
+using Newtonsoft.Json;
 using QuantConnect.Brokerages;
+using QuantConnect.Configuration;
 using QuantConnect.Interfaces;
 using QuantConnect.Lean.Engine.Results;
 using QuantConnect.Logging;
 using QuantConnect.Orders;
 using QuantConnect.Orders.Fees;
+using QuantConnect.Orders.Serialization;
 using QuantConnect.Securities;
 using QuantConnect.Securities.Positions;
 using QuantConnect.Util;
@@ -58,7 +62,7 @@ namespace QuantConnect.Lean.Engine.TransactionHandlers
 
         private Thread _processingThread;
         private readonly CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
-
+        private List<SerializedOrder> _openOrdersCache = new List<SerializedOrder>();
         private readonly ConcurrentQueue<OrderEvent> _orderEvents = new ConcurrentQueue<OrderEvent>();
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Engine/TransactionHandlers/BrokerageTransactionHandler.cs
```diff
@@ -673,6 +716,17 @@ public void HandleOrderRequest(OrderRequest request)
...
         }
 
+        /// <summary>
+        /// Loads in cache information of existing orders stored in json file
+        /// </summary>
+        protected void RefreshOpenOrdersCache()
+        {
+            if (File.Exists(OpenOrdersFilePath))
+            {
+                _openOrdersCache = JsonConvert.DeserializeObject<List<SerializedOrder>>(File.ReadAllText(OpenOrdersFilePath));
```
Martin-Molinero, line 726: Interesting!
Currently Lean is expected to be restoring the orders from the brokerage itself, see brokerage setup handler, I'd expect the order retrieve to be valid/complete, isn't the underlying bug here a brokerage implementation which isn't returning the tags set on the orders (like https://github.com/QuantConnect/Lean/issues/3735) ?

IlshatGaripov: It seems that not all brokerages allow to keep the order tags. 
Tradier does seem to have that option https://documentation.tradier.com/brokerage-api/trading/place-equity-order
IB has many additional fields where additional info could be stored , I guess OrderMiscOptions could work for the purpose https://interactivebrokers.github.io/tws-api/classIBApi_1_1Order.html#a83f6142af0b975ae26c85573d0daee2e and it's just that in current IB implementation in Lean tags are not stored/restored.

At the same time I found there are other brokerages, example crypto, that miss that field for the order:
CoinBase https://docs.pro.coinbase.com/#place-a-new-order
Binance https://binance-docs.github.io/apidocs/spot/en/#new-order-trade
so I decided we could add the feature that would work for all the brokerage

_Note: current implementation restores tag for limit orders only , I didn't pay attention to the fact there could also be other types that are submitted and not filled when algo restarts, they are LimitIfTouched and StopOrders I guess, if I am not missing any others? This I will fix._

Martin-Molinero:
Hey @IlshatGaripov! 
Sorry for the delay coming back to this one,
Live Lean deployments are already storing their state every hour today and at the end of the algorithm too, this state file contains orders, it's at `{AlgorithmId}.json`, so that would avoid the need to store them again. The only missing piece is loading up the tags from this file.
I think ideally this could come through the `Job.BrokerageData` collection, similar to https://github.com/QuantConnect/Lean/blob/master/Brokerages/Brokerage.cs#L237 & https://github.com/QuantConnect/Lean/blob/master/Brokerages/Brokerage.cs#L208 it could provide open orders and the brokerage implementation could use it to pick up the tags or even the open orders themselves. Following the pattern of holdings and cash balance.
Closing current approach for now, let me know what you think!

IlshatGaripov:
Hi @Martin-Molinero sorry yes I should have closed this earlier, I ve been working hard on one algo in june and july, this was a day trading algo on tick resolution, where it was essential to restore the tags in live mode otherwise it led to losses. But it could not show as well good performance in live as in backtest, so I had to switch to another algo and could not finish this in timely manner. I will probably return to that algo and will finish then with this issue. thanks!

Issue #5790, opened 2021-07-23
TrailingStopRiskManagementModel Only Handles Long Positions


AlexCatarino:
#### Expected Behavior
`TrailingStopRiskManagementModel` handles risk for both long and short positions.

#### Actual Behavior
`TrailingStopRiskManagementModel` only handles risk for long positions.

#### Potential Solution
Implement logic to handle risk for short positions.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

davidackroyd99:
Please could I be assigned to this issue?

Issue #2716, opened 2018-11-29
Extends Trailing Stop Risk Management Model for short positions


AlexCatarino:
#### Expected Behavior
Trailing Stop Risk Management Model should work for long and short positions.

#### Actual Behavior
Trailing Stop Risk Management Model only supports long positions.

#### Potential Solution
Implement the short side logic.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

ofirshm:
class TrailingStopPerSecurity(RiskManagementModel):
    '''Provides an implementation of IRiskManagementModel that limits the drawdown per holding to the specified percentage'''

    def __init__(self, maximumDrawdownPercent = 0.05):
        '''Initializes a new instance of the MaximumDrawdownPercentPerSecurity class
        Args:
            maximumDrawdownPercent: The maximum percentage drawdown allowed for any single security holding'''
        self.maximumDrawdownPercent = -abs(maximumDrawdownPercent)
        self.asset_best_pnl = {}
        
    def ManageRisk(self, algorithm, targets):
        '''Manages the algorithm's risk at each time step
        Args:
            algorithm: The algorithm instance
            targets: The current portfolio targets to be assessed for risk'''
        syms = [kvp.Value.Symbol for  kvp in algorithm.Securities]
        new_assets = {key:val for key,val in self.asset_best_pnl.items() if key in syms}
        self.asset_best_pnl = new_assets

        targets = []
         
        for kvp in algorithm.Securities:
            security = kvp.Value
            symbol = security.Symbol
            if symbol not in self.asset_best_pnl.keys():
               self.asset_best_pnl[symbol] = security.Holdings.UnrealizedProfitPercent
            self.asset_best_pnl[symbol] = np.maximum(self.asset_best_pnl[symbol], security.Holdings.UnrealizedProfitPercent)
            if not security.Invested:
                continue

            pnl = security.Holdings.UnrealizedProfitPercent -  self.asset_best_pnl[symbol]
            if pnl < self.maximumDrawdownPercent:
                # liquidate
                targets.append(PortfolioTarget(symbol, 0))

        return targets

PR #5791, opened 2021-07-23
Fix for Trailing Stop Risk Management Model

IlshatGaripov:
#### Description
I am making contribution as separate file so it's easier to review/test. I had it in local files. If solution is correct this could be used to replace existing TrailingStopRiskManagementModel

#### Related Issue
Closes https://github.com/QuantConnect/Lean/issues/5790
Closes https://github.com/QuantConnect/Lean/issues/2716

#### Motivation and Context
To fix an issue

Used module in couple local algorithms for trailing stop

#### Types of changes

- [ X] Refactor (non-breaking change which improves implementation)

#### Checklist:


Algorithm.Framework/Risk/TrailingDrawdownFromMaxPnl.cs
```diff
@@ -0,0 +1,87 @@
...
+                if (!security.Invested)
+                {
+                    if (_trailing.ContainsKey(symbol))
+                    {
+                        _trailing.Remove(symbol);
+                    }
+                    continue;
+                }
+
+                var pnl = security.Holdings.UnrealizedProfitPercent * 100;
```
Martin-Molinero, line 63: Cool idea 👍

Martin-Molinero (commented):
Thanks for the contribution! 👍
Should update existing `TrailingStopRiskManagementModel` C#/Py implementations and add new tests to cover the new behavior

Algorithm.Framework/Risk/TrailingStopRiskManagementModel.cs
```diff
@@ -53,37 +53,36 @@ public override IEnumerable<IPortfolioTarget> ManageRisk(QCAlgorithm algorithm,
...
                     continue;
                 }
 
-                // Check for new highs and update - set to tradebar high
-                if (_trailingHighs[symbol] < security.High)
+                // Check for new high and update
+                if (_trailing[symbol] < profitPercent)
                 {
-                    _trailingHighs[symbol] = security.High;
+                    _trailing[symbol] = profitPercent;
```
Martin-Molinero, line 75: `_trailing[symbol] = profitPercent;` should probably start from 0 here? 

Suggest directly using `TryGetValue()` instead of 3 access: `ContainsKey` + `[symbol]` twice
Same for python 👍

Martin-Molinero (commented):
Cool!
> add new tests to cover the new behavior

Let's please add a new unit test to cover the new feature.
*Needs a rebase

Algorithm.Framework/Risk/TrailingStopRiskManagementModel.cs
IlshatGaripov: Hey @Martin-Molinero thanks for review, good point, have fixed it. Hope python version is done right, I am not a big expert lol) As for the unit test - should this be another regression algorithm? or regular test? Sorry didn't quite get it. Thanks,

Martin-Molinero: Hey! 
> Hey @Martin-Molinero thanks for review, good point, have fixed it. Hope python version is done right, I am not a big expert lol)

Looks good 💯 

> As for the unit test - should this be another regression algorithm? or regular test? Sorry didn't quite get it. Thanks,

Would be nice to add unit tests, example for another risk model at https://github.com/QuantConnect/Lean/blob/master/Tests/Algorithm/Framework/Risk/MaximumDrawdownPercentPerSecurityTests.cs

Martin-Molinero: Reminder: branch needs a rebase from master

Algorithm.Framework/Risk/TrailingStopRiskManagementModel.cs
```diff
@@ -27,15 +27,15 @@ namespace QuantConnect.Algorithm.Framework.Risk
...
+        private readonly Dictionary<Symbol, decimal> _trailing = new Dictionary<Symbol, decimal>();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TrailingStopRiskManagementModel"/> class
         /// </summary>
         /// <param name="maximumDrawdownPercent">The maximum percentage relative drawdown allowed for algorithm portfolio compared with the highest unrealized profit, defaults to 5% drawdown per security</param>
         public TrailingStopRiskManagementModel(decimal maximumDrawdownPercent = 0.05m)
         {
-            _maximumDrawdownPercent = -Math.Abs(maximumDrawdownPercent);
+            _maximumDrawdownPercent = Math.Abs(maximumDrawdownPercent);
```
AlexCatarino, line 38: Have you considered a different `maximumDrawdownPercent` for long and short positions (e.g.: `maximumDrawdownPercentLong` and `maximumDrawdownPercentLong`)? It would give more flexibility to this model. The parameters could be `decimal?` and if we use
```csharp
var model = TrailingStopRiskManagementModel(0.05m, null)
```
would turn off the trailing stop for short positions and 
```csharp
var model = TrailingStopRiskManagementModel(null, 0.05m)
```
would turn off the trailing stop for long positions.

IlshatGaripov: I think that in most cases, the same logic is used to generate both long and short signals, i.e mirrored reversed, and this usually makes these insights equally relevant. I believe there are many other cases where this separation could be useful. To keep the things simple, I 've only done testing with uniform trailing stop for both long/short. Also possible to note, that sometimes it may work better to trail the stops by various indicators that change over time. I am more supporter of this concept.

Martin-Molinero (approved):
Looks good to me 👍
- Regression test stats need an update
- One of the new unit tests is currently failing

IlshatGaripov:
@Martin-Molinero thanks for review . I have applied a fix . Sorry was a small mistake, as for the units mine are green

image

It's strange. May you please check again? Thank you!

> Looks good to me 👍
> 
> * Regression test stats need an update
> * One of the new unit tests is currently failing

Martin-Molinero:
Hey!
> It's strange. May you please check again? Thank you!

You can check the result of GH, for example the last https://github.com/QuantConnect/Lean/runs/3290103735

Tests/Algorithm/Framework/Risk/MaximumDrawdownPercentPerSecurityTests.cs
```diff
@@ -32,11 +32,9 @@ public class MaximumDrawdownPercentPerSecurityTests
         [TestCase(Language.CSharp, 0.1, false, 0, 0, false)]
         [TestCase(Language.CSharp, 0.1, true, -50, 1000, false)]
         [TestCase(Language.CSharp, 0.1, true, -100, 1000, false)]
-        [TestCase(Language.CSharp, 0.1, true, -150, 1000, true)]
         [TestCase(Language.Python, 0.1, false, 0, 0, false)]
         [TestCase(Language.Python, 0.1, true, -50, 1000, false)]
         [TestCase(Language.Python, 0.1, true, -100, 1000, false)]
-        [TestCase(Language.Python, 0.1, true, -150, 1000, true)]
```
Martin-Molinero, line 39: Why removing this test cases for the other risk management model, shouldn't be affecting it?

Martin-Molinero (changes requested):

IlshatGaripov:
> Hey!
> 
> > It's strange. May you please check again? Thank you!
> 
> You can check the result of GH, for example the last https://github.com/QuantConnect/Lean/runs/3290103735

Hey Martin, these two tests were the ones failing in that build run by the link above:

image

I've then tested it locally , these two removed were indeed failing

Martin-Molinero:
> 
> 
> > Hey!
> > > It's strange. May you please check again? Thank you!
> > 
> > 
> > You can check the result of GH, for example the last https://github.com/QuantConnect/Lean/runs/3290103735
> 
> Hey Martin, these two tests were the ones failing in that build run by the link above:
> 
> image
> 
> I've then tested it locally , these two removed were indeed failing

They are passing in `master` https://github.com/QuantConnect/Lean/actions?query=branch%3Amaster+. It's because of the `SecurityHolding` change

IlshatGaripov:
> They are passing in `master` https://github.com/QuantConnect/Lean/actions?query=branch%3Amaster+. It's because of the `SecurityHolding` change.
> Edit: We should fix the unit tests failing, not remove them 👌

I got it now. Thanks! I am making the appropriate changes

Martin-Molinero (approved):
Thanks for the fix!

Issue #5678, opened 2021-06-17
Scheduled Events Don't Start In Current Month When Live Trading


DerekMelchin:
#### Expected Behavior

When we set a monthly Scheduled Event, it starts firing in the current month.

#### Actual Behavior

When we set a monthly Scheduled Event, it starts firing next month.

#### Potential Solution

N/A

#### Reproducing the Problem

Run [this algorithm](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_94fb101df51e754fb8f51b0aeb056893.html) live. It sets monthly Scheduled Events to occur around the current day.
```
self.Schedule.On(self.DateRules.MonthStart(10),self.TimeRules.At(19, 23), self.RebalancingCode)
# ...
self.Schedule.On(self.DateRules.MonthStart(20),self.TimeRules.At(19, 23), self.RebalancingCode)
```
Even though today is the 17th, the live logs show that all of the Scheduled Events start firing in the following month.

![image](https://user-images.githubusercontent.com/38889814/122485208-94e35280-cf93-11eb-976b-8a8bd64f2057.png)

#### System Information

Run with the lean cli. Window 10

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #5808, opened 2021-07-29
Pandas Remapper Refactor

C-SELLERS:
#### Description

Refactor pandas mapper as a part of #5678

Broke this change out into its own PR since it applies to old and new pandas.

#### Related Issue

#5678

#### Motivation and Context

Attempting to support newer pandas version > 1 caused a lot of problems with the previous implementation. So I went about finding a new way to do mapping to Symbol ID indexes. 

I was able to do this by directly wrapping internally shared index functions that do the fetching of values. In this wrapping I remap the args and attempt to fetch the index with the new values. If this fails pandas will attempt to fall back to the previous arg request just in case (supports user defined pandas frames with tickers)

#### Performance improvements
Research Indexing
```
Cell 1:
qb = QuantBook()
qb.AddEquity("SPY")
qb.AddEquity("TSLA")
qb.AddEquity("AAPL")
qb.AddEquity("GOOG")
qb.AddEquity("BAC")
qb.AddEquity("CHWY")
qb.AddEquity("BABA")
qb.AddEquity("RKT")
qb.AddForex("EURUSD")
qb.AddCrypto("BTCUSD")

history = qb.History(qb.Securities.Keys, 360, Resolution.Daily)

Cell 2:
%%time
for seq in range(1000):
    for key in qb.Securities.Keys:
        history.loc[key]
        history.loc[key.Value]
        history.loc[str(key.ID)]
```

Results from cell 2 %%time:
```
---------- NEW -----------
Research Indexing
CPU times: user 19.6 s, sys: 84.1 ms, total: 19.7 s
CPU times: user 19.5 s, sys: 156 ms, total: 19.7 s
CPU times: user 19.9 s, sys: 123 ms, total: 20 s
CPU times: user 20.2 s, sys: 79.5 ms, total: 20.2 s
CPU times: user 19.9 s, sys: 123 ms, total: 20 s
CPU times: user 20.1 s, sys: 217 ms, total: 20.3 s

---------- OLD -----------
Research Indexing
CPU times: user 1min 47s, sys: 693 ms, total: 1min 47s
CPU times: user 1min 48s, sys: 834 ms, total: 1min 48s
CPU times: user 1min 49s, sys: 930 ms, total: 1min 50s
CPU times: user 1min 49s, sys: 974 ms, total: 1min 50s
CPU times: user 1min 45s, sys: 958 ms, total: 1min 46s
```

`HistoryRequestBenchmark` Python Algorithm
```
---------- NEW -----------
45.86 seconds at 40k data points per second
40.77 seconds at 45k data points per second
47.54 seconds at 38k data points per second
53.84 seconds at 34k data points per second

---------- OLD -----------
62.89 seconds at 29k data points per second
59.16 seconds at 31k data points per second
68.05 seconds at 27k data points per second
66.98 seconds at 27k data points per second
```

N/A

All regressions and unit tests passing with:
- current pandas (0.25.3)
- pandas 1.15

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Python/PandasData.cs
+++ b/Common/Python/PandasData.cs
@@ -59,86 +59,8 @@ namespace QuantConnect.Python
             {
                 using (Py.GIL())
                 {
-                    // this python Remapper class will work as a proxy and adjust the
-                    // input to its methods using the provided 'mapper' callable object
-                    _pandas = PythonEngine.ModuleFromString("remapper",
-                        @"import pandas as pd
-from pandas.core.indexes.frozen import FrozenList as pdFrozenList
-
-from clr import AddReference
-AddReference(""QuantConnect.Common"")
-from QuantConnect import *
-
-def mapper(key):
-    '''Maps a Symbol object or a Symbol Ticker (string) to the string representation of
-    Symbol SecurityIdentifier.If cannot map, returns the object
-    '''
-    keyType = type(key)
-    if keyType is Symbol:
-        return str(key.ID)
-    if keyType is str:
-        reserved = ['high', 'low', 'open', 'close']
-        if key in reserved:
-            return key
-        kvp = SymbolCache.TryGetSymbol(key, None)
-        if kvp[0]:
-            return str(kvp[1].ID)
-    if keyType is list:
-        return [mapper(x) for x in key]
-    if keyType is tuple:
-        return tuple([mapper(x) for x in key])
-    if keyType is dict:
-        return { k: mapper(v) for k, v in key.items()}
-    return key
-
-def wrap_function(f):
-    '''Wraps function f with g.
-    Function g converts the args / kwargs to use alternative index keys
-      and then calls original function with the mapped args
-    '''
-    def wrapped_function(*args, **kwargs):
-
-        # Map args & kwargs, if wrapped function fails because key, execute with original
-        # Allows for df, Series, etc indexing for keys like 'SPY' if they exist
-        try:
-            newargs = args
-            newkwargs = kwargs
-
-            if len(args) > 1:
-                newargs = mapper(args)
-            if len(kwargs) > 0:
-                newkwargs = mapper(kwargs)
-
-            result = f(*newargs, **newkwargs)
-            return result
-        except KeyError:
-            pass
-
-        result = f(*args, **kwargs)
-        return result
-
-    wrapped_function.__name__ = f.__name__
-    return wrapped_function
-
-# Wrap all core __getItem__ and loc functions that are shared, yet still throw key errors if index not found
-pd.core.indexing._LocationIndexer.__getitem__ = wrap_function(pd.core.indexing._LocationIndexer.__getitem__)
-pd.core.indexing._ScalarAccessIndexer.__getitem__ = wrap_function(pd.core.indexing._ScalarAccessIndexer.__getitem__)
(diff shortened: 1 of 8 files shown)


Common/Python/PandasData.cs
```diff
@@ -95,220 +88,55 @@ public PandasData(object data)
...
-
-setattr(modules[__name__], 'concat', wrap_function(pd.concat))");
+    wrapped_function.__name__ = f.__name__
+    return wrapped_function
+
+pd.core.indexing._LocationIndexer.__getitem__ = wrap_function(pd.core.indexing._LocationIndexer.__getitem__)
+pd.core.indexing._ScalarAccessIndexer.__getitem__ = wrap_function(pd.core.indexing._ScalarAccessIndexer.__getitem__)
+pd.core.indexes.base.Index.get_loc = wrap_function(pd.core.indexes.base.Index.get_loc)
+pd.core.indexes.base.Index.__contains__ = wrap_function(pd.core.indexes.base.Index.__contains__)
+pd.core.series.Series.__getitem__ = wrap_function(pd.core.series.Series.__getitem__)
```
Martin-Molinero, line 127: Maybe a small comment of why these would be cool 👍

Tests/Python/PandasConverterTests.cs
```diff
@@ -2652,6 +2712,11 @@ public void BackwardsCompatibilitySeries_set_value(string index, bool cache = fa
         {
             if (cache) SymbolCache.Set("SPY", Symbols.SPY);
 
+            // set_value deprecated in newer pandas
+            if(_newerPandas) {
+                return;
+            }
+
             using (Py.GIL())
```
Martin-Molinero, line 2720: Cool! Makes me think we should be adding new unit tests for the new stuff in pandas?

Martin-Molinero (commented):
Awesome! Love the simplification here, minor comments 👍

Tests/Python/PandasConverterTests.cs
C-SELLERS: I went through the unit tests that use deprecated functions and ensured their replacements are all covered. All these cases were just using old deprecated ways of indexing instead of stricter iloc, loc, iat, & at indexing. These are already being tested 👍🏽

There were a couple tests that were made to test other functions but used the deprecated indexing, so I refactored them to use the stricter indexers to be sure we are testing the method intended for both new and old pandas. 😁  They are all good now!

Common/Python/PandasData.cs
```diff
@@ -59,256 +59,8 @@ public PandasData(object data)
...
-wrap_special_function('expanding', Series, Expanding)
-wrap_special_function('expanding', DataFrame, Expanding)
-wrap_special_function('rolling', Series, Rolling, Window)
-wrap_special_function('rolling', DataFrame, Rolling, Window)
-
-CreateSeries = pd.Series
-
-setattr(modules[__name__], 'concat', wrap_function(pd.concat))");
+                    // Use our PandasMapper class that modifies pandas indexing to support tickers, symbols and SIDs
+                    _pandas = PythonEngine.ImportModule("PandasMapper");
```
Martin-Molinero, line 63: Beautiful ❤️

Martin-Molinero (approved):

Common/Python/PandasData.cs
C-SELLERS: 🙌🏽

PR #5823, opened 2021-08-04
Change Binance Benchmark to BTCUSDC

AlexCatarino:
#### Description
`BTCUSD` is not supported by Binance (see Symbol Properties Dabase)

#### Motivation and Context
Bug fix and improve Brokerage modeling.

The following algorithm raises an exception on master, and passes with this change
```python
class BinanceAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.SetBrokerageModel(BrokerageName.Binance, AccountType.Cash)
```
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Brokerages/BinanceBrokerageModel.cs
+++ b/Common/Brokerages/BinanceBrokerageModel.cs
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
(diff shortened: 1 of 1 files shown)


gsalaz98:
Comparing BTCUSDT vs BTCUSDC, it seems the USDT pair has higher trading volume than USDC by $1.9 Billion USD(T|C). Since they're both stablecoins pegged to the USD, this comparison is 1:1.

Maybe we should use USDT for the benchmark, but realistically it doesn't even matter since they're both essentially the same :smile:

image
image

Martin-Molinero (approved):

Issue #5835, opened 2021-08-08
RelativeDailyVolume EnableAutomaticIndicatorWarmUp


jovuribe:
#### Expected Behavior

When EnableAutomaticIndicatorWarmUp is set to True, no warm-up should be needed.

#### Actual Behavior

Does not warm up correctly

#### Potential Solution

Changing the resolution in Indicator.cs -> RelativeDailyVolume -> WarmUpindicator to a fixed Resolution of Resolution.Daily

#### Reproducing the Problem

Run C# or Python algorithm with EnableAutomaticIndicatorWarmUp set to True

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

jaredbroad:
Remove the WarmUpPeriod Provider for the indicator in the meantime.

PR #5836, opened 2021-08-09
EnableAutomaticIndicatorWarmUp change in RelativeDailyVolume Indicator

jovuribe:
#### Description

Changed:
WarmUpIndicator(symbol, relativeDailyVolume, Resolution.Daily);
to
WarmUpIndicator(symbol, relativeDailyVolume, Resolution.Daily); // Fixed WarmUp Resolution because this indicator is based on days
#### Related Issue

#5835 

#### Motivation and Context

Fixs the automatic warmup of the RDV indicator

N/A

Test cases and works on a locally developed algorithm

#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1528,7 +1528,7 @@ public RelativeDailyVolume RDV(Symbol symbol, int period = 2, Resolution resolut
 
             if (EnableAutomaticIndicatorWarmUp)
             {
-                WarmUpIndicator(symbol, relativeDailyVolume, resolution);
+                WarmUpIndicator(symbol, relativeDailyVolume, Resolution.Daily); // Fixed WarmUp Resolution because this indicator is based on days
```
Martin-Molinero, line 1531: Cool!
Please add a test reproducing the issue and being fixed because of this change.
The default resolution of the indicators input data, line 1523 Daily, should probably be hourly? Else I don't think it would really be useful/valid? Seems like `RDV` should probably throw an argument exception if resolution is Daily

Martin-Molinero (changes requested):

Algorithm.Python/RelativeDailyVolumeAlgorithm.py
```diff
@@ -0,0 +1,52 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from clr import AddReference
+AddReference("System")
+AddReference("QuantConnect.Algorithm")
+AddReference("QuantConnect.Common")
```
Martin-Molinero, line 17: All these imports should not be required since included by `from AlgorithmImports import *`

Martin-Molinero (changes requested):
Believe `RelativeDailyVolume` indicator implementation is not deterministic. https://github.com/QuantConnect/Lean/blob/master/Indicators/RelativeDailyVolume.cs#L80 Dictionary enumeration order is not guaranteed, should probably be a sorted dictionary what we want here

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1515,20 +1515,24 @@ public RelativeVigorIndex RVI(Symbol symbol, int period, MovingAverageType movin
...
+                throw new ArgumentException("RelativeDailyVolume Indicator must have a resolution less than Resolution.Daily", nameof(resolution));
+            }
             var name = CreateIndicatorName(symbol, $"RDV({period})", resolution);
             var relativeDailyVolume = new RelativeDailyVolume(name, period);
             RegisterIndicator(symbol, relativeDailyVolume, resolution, selector);
 
             if (EnableAutomaticIndicatorWarmUp)
             {
-                WarmUpIndicator(symbol, relativeDailyVolume, resolution);
+                WarmUpIndicator(symbol, relativeDailyVolume, Resolution.Daily);
```
Martin-Molinero, line 1535: This isn't right, Resolution.Daily here is telling lean that the indicator needs daily resolution bars which isn't true, see exception at line 1527, should be using what the user provided as `resolution`. Maybe the bug is the amount of bars the indicator is telling Lean it needs?

Martin-Molinero (changes requested):
Hey @jovuribe! please give this a local debug session to iron it out 👍

Martin-Molinero:
Hey @jovuribe! Any update with this issue fix?

jovuribe:
The python regression test is failing with the error being that the OrderListHash is different than expected. However, when running this locally, the OrderListHash is correct: 
image

jovuribe:
> Hey @jovuribe! Any update with this issue fix?

Yep, made a comment below.

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1515,20 +1515,29 @@ public RelativeVigorIndex RVI(Symbol symbol, int period, MovingAverageType movin
...
+            var numberOfBars = (TimeSpan.FromDays(1) / resolution.ToTimeSpan()) * period;
 
             if (EnableAutomaticIndicatorWarmUp)
             {
-                WarmUpIndicator(symbol, relativeDailyVolume, resolution);
+                var history = History(symbol, (int)numberOfBars, resolution);
+                foreach(TradeBar tradeBar in history)
+                {
+                    relativeDailyVolume.Update(tradeBar);
+                }
```
Martin-Molinero, line 1540: Shouldn't duplicate the purpose/responsibility of `WarmUpIndicator`, the _indicator_ should return a valid number of bars required to warmup it up

Martin-Molinero (changes requested):

Algorithm/QCAlgorithm.Indicators.cs
jovuribe: Please see changes. Still passes locally but not on GitHub.

Martin-Molinero:
Hey @jovuribe!
Locally I seem to be getting the same result as GH does, can you retry with a clean repository copy? Attaching my [orderList.txt](https://github.com/QuantConnect/Lean/files/7173836/orderList.txt) just before it's hashed so you can compare see https://github.com/QuantConnect/Lean/blob/master/Common/Extensions.cs#L475

Martin-Molinero:
Hey @jovuribe! Were you able to test this out?

Martin-Molinero:
Closing for now, please re open when updated/addressed

PR #5841, opened 2021-08-09
Performs update

IlshatGaripov:
#### Description
Update Binance symbols properties by running --app=binancesymbolpropertiesupdater
Just to be on a safe side and make sure all the alt coins are existent - before uploading historical data for the various new alt coins


StefanoRaggi (approved):
@IlshatGaripov Thanks for the update 👍

PR #5846, opened 2021-08-10
More logging for BinanceRestApiClient

IlshatGaripov:
#### Motivation and Context
Makes historical data download process more interactive by adding the log messages to trace the current execution stage.

### DIFF
--- a/Brokerages/Binance/BinanceRestApiClient.cs
+++ b/Brokerages/Binance/BinanceRestApiClient.cs
@@ -319,6 +319,11 @@ namespace QuantConnect.Brokerages.Binance
                 var klines = JsonConvert.DeserializeObject<object[][]>(response.Content)
                     .Select(entries => new Messages.Kline(entries))
                     .ToList();
+
+                var windowStartTime = (new DateTime(1970, 1, 1)).AddMilliseconds(klines.First().OpenTime);
+                var windowEndTime = (new DateTime(1970, 1, 1)).AddMilliseconds(klines.Last().OpenTime + resolutionInMs);
+                Log.Trace($"Received [{symbol}] data for timeperiod from {windowStartTime.ToStringInvariant()} to {windowEndTime.ToStringInvariant()}..");
+
                 if (klines.Count > 0)
                 {
                     startMs = klines.Last().OpenTime + resolutionInMs;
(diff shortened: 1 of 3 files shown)


Martin-Molinero (changes requested):
Hey @IlshatGaripov!
Don't think we should Trace log in the while loop for each history request. Suggest replacing for something like the following, else it would get to verbose quickly
```
            if (Log.DebuggingEnabled)
            {
                Log.Debug();
            }
```

Martin-Molinero:
Hey!
> GDAXDownloader.cs

Think the key here is the `Downloader` part -> it's more of a one time execution for a short lived process. Versus a history request in a live algorithm which can happen multiple times in a context of a long running process, logs will explode as you are showing in the snapshots. `PolygonDataQueueHandler` should not be logging that much.
Also it takes me to the reason of why logging _each little sub request_ of a single history request, which it's more of a debugging operation not Trace

IlshatGaripov:
Makes sense  👍 . I have changed it to Debug level messages for both Binance and Polygon. Thanks!

Brokerages/Binance/BinanceRestApiClient.cs
```diff
@@ -317,6 +317,11 @@ public IEnumerable<Messages.Kline> GetHistory(Data.HistoryRequest request)
                 var klines = JsonConvert.DeserializeObject<object[][]>(response.Content)
                     .Select(entries => new Messages.Kline(entries))
                     .ToList();
+
+                var windowStartTime = (new DateTime(1970, 1, 1)).AddMilliseconds(klines.First().OpenTime);
+                var windowEndTime = (new DateTime(1970, 1, 1)).AddMilliseconds(klines.Last().OpenTime + resolutionInMs);
+                Log.Debug($"Received [{symbol}] data for timeperiod from {windowStartTime.ToStringInvariant()} to {windowEndTime.ToStringInvariant()}..");
```
Martin-Molinero, line 323: Should move this inside the next `if (klines.Count > 0)`, since `klines` might be empty? I bet we have helper methods for this `(new DateTime(1970, 1, 1)).AddMilliseconds()`

IlshatGaripov:
_Note: still getting build issue I was going to investigate it during the week_
image

Martin-Molinero (approved):

PR #5848, opened 2021-08-10
Pandas Mapper Recursion Bug Fix

C-SELLERS:
#### Description

Pandas mapper indexing would sometimes recurse infinitely before crashing. After reproducing the issue it became clear that it was caused by wrapping our base `__contains__` method with the same wrapper as our indexers.

Since the __contains__ returns true or false it does not throw a KeyError like the index functions, leading to conditions under which an underlying dataframe could contain values like "SPY" but because its mapped as "SPY ******" before checking it would return false, causing different pathways to be followed, (I.E. `DataFrame.__getitem__`, failing to `key in self.columns`, causing a infinite recurse until crash)

The solution was to make a special wrapper for functions like __contains__ that checks both the original and mapped keys and returns if either are true, or false if both fail.

This is a better way to break up the Pandas mapper when we consider the index functions we wrapped all share the same KeyError behavior and __contains__ does not.

I also tacked on a more informative error for our wrapper_keyerror_function that prints the keys it attempted.

Note: I added a few to-dos to further cleanup our testing framework I created for pandas. I opt-ed for this implementation because it is much simpler to test it directly with a debugger in Python.

#### Related Issue

N/A

#### Motivation and Context

Read description

N/A

All tests passing, added additional unit tests that are red -> green

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Python/PandasIndexingTests.cs
+++ b/Tests/Python/PandasIndexingTests.cs
@@ -40,7 +40,7 @@ namespace QuantConnect.Tests.Python
         }
 
         [Test]
-        public void TestIndexingDataFrameWithList()
+        public void IndexingDataFrameWithList()
         {
             using (Py.GIL())
             {
@@ -49,7 +49,7 @@ namespace QuantConnect.Tests.Python
         }
 
         [Test]
-        public void TestContainsUserMappedTickers()
+        public void ContainsUserMappedTickers()
         {
             using (Py.GIL())
             { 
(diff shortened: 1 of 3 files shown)


Tests/Python/PandasIndexingTests.cs
```diff
@@ -43,5 +47,29 @@ public void TestIndexingDataFrameWithList()
                 Assert.DoesNotThrow((() => _pandasIndexingTests.test_indexing_dataframe_with_list()));
             }
         }
+
+        [Test]
+        public void TestContainsUserMappedTickers()
```
Martin-Molinero, line 52: nit - Don't think tests shouldn't have `Test` in their name

Martin-Molinero (approved):
Sweet! leaving minor comments

PR #5913, opened 2021-09-09
Fixes Market On Close Fill of Equity Fill Model

AlexCatarino:
#### Description
Only use trade data (`Tick` with `TickTrade` type or `TradeBar`) to get the closing price, since MOC is filled with the closing price.

- Change regression tests to reflect the bug fix. All other changes are the consequence of using `TradeBar` instead of `QuoteBar`.

#### Related Issue
Ref.: #4567

Unit and regression tests.

#### Types of changes

#### Checklist:


Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -587,11 +587,52 @@ public virtual OrderEvent MarketOnCloseFill(Security asset, MarketOnCloseOrder o
...
+                var trade = ticks
+                    .Where(x => !string.IsNullOrWhiteSpace(x.SaleCondition))
+                    .LastOrDefault(x =>
+                        x.TickType == TickType.Trade && x.Price > 0 && x.ExchangeCode == primaryExchange &&
+                        (x.ParsedSaleCondition == officialClose || x.ParsedSaleCondition == closingPrints));
+
+                // If there is no tick with the desired information, we wait one minute
+                if (trade == null)
+                {
+                    trade = ticks.LastOrDefault(x => x.Price > 0);
```
Martin-Molinero, line 611: Seems `trade` here might not really be of `Trade` type?

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -587,11 +587,52 @@ public virtual OrderEvent MarketOnCloseFill(Security asset, MarketOnCloseOrder o
...
+                    {
+                        return fill;
+                    }
+                }
+
+                fill.FillPrice = trade.Price;
+            }
             // make sure the exchange is open/normal market hours before filling
-            if (!IsExchangeOpen(asset, false)) return fill;
+            else if (!IsExchangeOpen(asset, false))
```
Martin-Molinero, line 621: > if (asset.LocalTime < nextMarketClose)

At line 586, means asset time would already be past market close of the order,

> if (!IsExchangeOpen(asset, false))

Requires the market to be open to skip this `if`?
Meaning we really are waiting for the next market open to fill the MOC order, for resolution != Tick?

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -587,11 +587,52 @@ public virtual OrderEvent MarketOnCloseFill(Security asset, MarketOnCloseOrder o
...
+            else if (subscribedTypes.Contains(typeof(TradeBar)))
+            {
+                var tradeBar = asset.Cache.GetData<TradeBar>();
+                if (tradeBar == null)
+                {
+                    return fill;
+                }
+
+                fill.FillPrice = tradeBar.Close;
+            }
```
Martin-Molinero, line 634: What about the final `else` case here? We have no `TradeBar` and not tick subscription?

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -587,11 +587,52 @@ public virtual OrderEvent MarketOnCloseFill(Security asset, MarketOnCloseOrder o
...
+            else if (!IsExchangeOpen(asset, false))
+            {
+                return fill;
+            }
+            else if (subscribedTypes.Contains(typeof(TradeBar)))
+            {
+                var tradeBar = asset.Cache.GetData<TradeBar>();
+                if (tradeBar == null)
+                {
+                    return fill;
```
Martin-Molinero, line 630: `return fill;` should we timeout here, similar to the market fill and use quote information? We are timing out with Tick resolution few lines above, think we should probably add some message to the fill about it?

Common/Orders/Fills/EquityFillModel.cs
AlexCatarino: `IsExchangeOpen` tests if the market is open (same as 586), if not, check the bar size, this handles large-sized bars (hours/days) in order to fill at that time, not on the next open.

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -590,59 +590,79 @@ public virtual OrderEvent MarketOnCloseFill(Security asset, MarketOnCloseOrder o
...
-                if (trade == null)
+                // The official close is recorded after the market is closed,
+                // We can only use this information if extended market hours are included
+                if (Parameters.ConfigProvider.GetSubscriptionDataConfigs(asset.Symbol).IsExtendedMarketHours())
                 {
-                    trade = ticks.LastOrDefault(x => x.Price > 0);
-                    if (trade == null || (trade.EndTime - nextMarketClose).TotalMinutes < 1)
+                    var primaryExchange = (byte) ((Equity) asset).PrimaryExchange;
+                    var officialClose = (uint) (TradeConditionFlags.Regular | TradeConditionFlags.OfficialClose);
+                    var closingPrints = (uint) (TradeConditionFlags.Regular | TradeConditionFlags.ClosingPrints);
```
Martin-Molinero, line 607: Think it should still _try_ to get the official closing trade even if `IsExtendedMarketHours` is false, the difference would be in one case it can wait a bit in the other it wouldn't? Thinking if the implementation of `MarketOnOpenFill` should also have a timeout (probably the same one) for waiting for the official open tick similar case as this

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -498,18 +497,39 @@ public virtual OrderEvent MarketOnOpenFill(Security asset, MarketOnOpenOrder ord
...
+                // If there are only quotes, we will need to test for the tick type before we assign the fill price
+                if (tick == null)
+                {
+                    var previousOpen = asset.Exchange.Hours
+                        .GetMarketHours(asset.LocalTime)
+                        .GetMarketOpen(TimeSpan.Zero, false);
 
-                if (trade != null)
+                    tick = trades.LastOrDefault() ?? asset.Cache.GetAll<Tick>().LastOrDefault();
+                    if ((tick?.EndTime.TimeOfDay - previousOpen)?.TotalMinutes < 1)
```
Martin-Molinero, line 521: `1` wondering if we should make this a configurable value

Martin-Molinero (approved):
Nice! awesome test suite

PR #5922, opened 2021-09-10
Adds support for getting a Lean symbol based on FIGI, CUSIP, ISIN, SEDOL

gsalaz98:
#### Description

Adds support for Symbol lookup based on a unique, industry standardized, security identifier such as FIGI, CUSIP, ISIN, SEDOL.
#### Related Issue

N/A
#### Motivation and Context

Ergonomics of resolving industry standardized security identifiers to usable Lean symbols.

Added new tests to ensure the mapping of tickers occurs based on the trading date provided, as well as parsing validation.
#### Types of changes

#### Checklist:


Common/Securities/SecurityDefinition.cs
```diff
@@ -0,0 +1,106 @@
...
+        public SecurityIdentifier SecurityIdentifier { get; set; }
+        
+        /// <summary>
+        /// The CUSIP of the security
+        /// </summary>
+        public string CUSIP { get; set; }
+       
+        /// <summary>
+        /// The FIGI of the security
+        /// </summary>
```
Martin-Molinero, line 42: nit - Loved the verbose/informative of the QCAlgorithm xml logs -> `The composite Financial Instrument Global Identifier (FIGI) of an asset` maybe we can copy them over

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,163 @@
...
+        /// <param name="cusip">CUSIP</param>
+        /// <param name="tradingDate">
+        /// The date that the stock was trading at with the CUSIP provided. This is used
+        /// to get the ticker of the symbol on this date.
+        /// </param>
+        /// <returns>The CUSIP's corresponding Symbol</returns>
+        public Symbol CUSIP(string cusip, DateTime tradingDate)
+        {
+            return SecurityDefinitionToSymbol(
+                _securityDefinitions?.FirstOrDefault(x => x.CUSIP == cusip && !string.IsNullOrWhiteSpace(x.CUSIP)),
```
Martin-Molinero, line 84: `&& !string.IsNullOrWhiteSpace(x.CUSIP)` we should check the argument is not null or whitespace before calling `_securityDefinitions?.FirstOrDefault` and just return null 👍

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,163 @@
...
+            _mapFileResolver = mapFileProvider.Get(market);
+        }
+        
+        /// <summary>
+        /// Converts CUSIP into a Lean <see cref="Symbol"/>
+        /// </summary>
+        /// <param name="cusip">CUSIP</param>
+        /// <param name="tradingDate">
+        /// The date that the stock was trading at with the CUSIP provided. This is used
+        /// to get the ticker of the symbol on this date.
```
Martin-Molinero, line 78: > This is used to get the ticker of the symbol on this date.

Ah nice this is what's missing in the other docs but here `"cusip"` can be improved - Suggest just copy pasting the exact complete docs 👍

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,163 @@
...
+            if (securityDefinition == null)
+            {
+                return null;
+            }
+            
+            // Get the first ticker the symbol traded under, and then lookup the
+            // trading date to get the ticker on the trading date.
+            var mappedTicker = _mapFileResolver
+                .ResolveMapFile(securityDefinition.SecurityIdentifier.Symbol, securityDefinition.SecurityIdentifier.Date)
+                .GetMappedSymbol(tradingDate);
```
Martin-Molinero, line 156: `GetMappedSymbol` wont map if `tradingDate` is past the delisting date, maybe to be more user friendly should just return the symbol unmapped?

Tests/Common/Securities/SecurityDefinitionSymbolResolverTests.cs
```diff
@@ -0,0 +1,220 @@
...
+using System.Collections.Generic;
+using System.IO;
+using NUnit.Framework;
+using QuantConnect.Configuration;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Tests.Common.Securities
+{
+    [TestFixture]
+    public class SecurityDefinitionSymbolResolverTests
```
Martin-Molinero, line 27: Cool!

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,163 @@
...
+            }
+
+            if (mapFileProvider == null)
+            {
+                var dataProvider = Composer.Instance.GetExportedValueByTypeName<IDataProvider>(Config.Get("data-provider", "QuantConnect.Lean.Engine.DataFeeds.DefaultDataProvider"));
+                mapFileProvider = Composer.Instance.GetExportedValueByTypeName<IMapFileProvider>(Config.Get("map-file-provider", "LocalDiskMapFileProvider"));
+                mapFileProvider.Initialize(dataProvider);
+            }
+            
+            _mapFileResolver = mapFileProvider.Get(market);
```
Martin-Molinero, line 69: Don't think we need to hard code the market, we can just get it from the SID from the `SecurityDefinition` and determine the `mapFileResolver` at runtime 👍

Common/Securities/SecurityDefinition.cs
```diff
@@ -0,0 +1,106 @@
...
+        public string ISIN { get; set; }
+        
+        /// <summary>
+        /// Reads data from the specified file and converts it to a list of SecurityDefinition
+        /// </summary>
+        /// <param name="securitiesFile">File to read from</param>
+        /// <returns>List of security definitions</returns>
+        public static List<SecurityDefinition> FromCsvFile(FileInfo securitiesFile)
+        {
+            return File.ReadAllLines(securitiesFile.FullName)
```
Martin-Molinero, line 62: We should be using the `IDataProvider` instance to get the file here, since this file might not be found locally (it's not in Lean) thinking about the `ApiDataProvider` case

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,163 @@
...
+            if (!SecurityDefinition.TryFromCsvFile(securityDefinitionFile, out _securityDefinitions))
+            {
+                Log.Error($"SecurityDefinitionSymbolResolver(): No security definitions data loaded from file: {securityDefinitionFile.FullName}");
+            }
+
+            if (mapFileProvider == null)
+            {
+                var dataProvider = Composer.Instance.GetExportedValueByTypeName<IDataProvider>(Config.Get("data-provider", "QuantConnect.Lean.Engine.DataFeeds.DefaultDataProvider"));
+                mapFileProvider = Composer.Instance.GetExportedValueByTypeName<IMapFileProvider>(Config.Get("map-file-provider", "LocalDiskMapFileProvider"));
+                mapFileProvider.Initialize(dataProvider);
```
Martin-Molinero, line 66: `LocalZipFactorFileProvider` & `LocalZipMapFileProvider` don't behave well being initialized twice, they could just return if already initialized, do you mind updating them to do so? 👍

Martin-Molinero (changes requested):
Awesome improvement! Leaving a few requests

Common/Securities/SecurityDefinitionSymbolResolver.cs
gsalaz98: Making it so that we return the last known symbol now :+1:

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            var market = securityDefinition.SecurityIdentifier.Market;
+            if (!_mapFileResolvers.TryGetValue(market, out var mapFileResolver))
+            {
+                var mapFileProvider = Composer.Instance.GetExportedValueByTypeName<IMapFileProvider>(Config.Get("map-file-provider", "LocalDiskMapFileProvider"));
+                mapFileProvider.Initialize(_dataProvider);
+
+                mapFileResolver = mapFileProvider.Get(market);
+                _mapFileResolvers[market] = mapFileResolver;
+            }
```
Martin-Molinero, line 176: Think the initial approach of performing the fetching and init of the `mapFileProvider` on construction seems more accurate since it should only be done once 👍 
We don't really need a cache of the `mapFileResolver`,  the provider handles it internally, and the provider handles clearing and regenerating it's cache which wouldn't happen here 👍 - Let's just do `var mapFileResolver = mapFileProvider.Get(market);`

Common/Securities/SecurityDefinitionSymbolResolver.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// <returns>The Lean Symbol corresponding to the CUSIP number on the trading date provided</returns>
+        public Symbol CUSIP(string cusip, DateTime tradingDate)
+        {
+            if (string.IsNullOrWhiteSpace(cusip))
+            {
+                return null;
+            }
+            
+            return SecurityDefinitionToSymbol(
+                _securityDefinitions?.FirstOrDefault(x => x.CUSIP == cusip),
```
Martin-Molinero, line 79: Ah something I didn't notice before is that comparisons here are subject to case, they should probably ignore case to be more user friendly?

Martin-Molinero (approved):
Cool! Almost there 👍

Common/Securities/SecurityDefinitionSymbolResolver.cs
gsalaz98: I thought about it for a second in the middle of my previous refactor, but decided it didn't really make sense, given that everybody only uses the upper-case variants of these security identifiers.

Most likely all of these security identifiers have standardization regarding case-sensitivity. I'm not 100% sure what to do here honestly.

If we ignored case when comparing, we'd be accepting non-standard security identifier values, though it'd come at the cost of ease of use.

I'm okay with either one of these choices, just making sure that I brought it up for consideration before I make the change

Martin-Molinero (approved):

Issue #5955, opened 2021-09-30
Unusual performance drop when using consolidators with python


ArthurAsenheimer:
#### Expected Behavior

Indicator updates using user-defined consolidators are executed at similar speed as automatic updates using RegisterIndicator.

#### Actual Behavior

User-defined consolidators results in a performance drop of ~70% despite identical workload.

#### Potential Solution

Review python wrapper for consolidators.

#### Reproducing the Problem

Original report: https://www.quantconnect.com/forum/discussion/12399/tradebarconsolidator-causes-big-performance-drop/p1

Python version: 
```Python
class PyConsolidatorSpeedTest(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2010, 1, 1)  
        self.SetEndDate(2020, 1, 1)
        self.SetCash(100000) 
        
        spy = self.AddEquity("SPY", Resolution.Minute).Symbol
        self.sma = SimpleMovingAverage(100)
        
        selectUserDefinedConsolidator = True
        
        if selectUserDefinedConsolidator:
            consolidator = TradeBarConsolidator(1)
            consolidator.DataConsolidated += self.OnDataConsolidated
            self.SubscriptionManager.AddConsolidator(spy, consolidator)
        else:
            self.RegisterIndicator(spy, self.sma, Resolution.Minute)
            
        
    def OnDataConsolidated(self, sender, bar):
        self.sma.Update(bar.EndTime, bar.Close)
        

    def OnData(self, data):
        if not self.Portfolio.Invested and self.sma.IsReady:
            self.SetHoldings("SPY", 1)
```

#### Results

with RegisterIndicator (i.e. `selectUserDefinedConsolidator = False`):
![image](https://user-images.githubusercontent.com/68070194/135417355-558a8ddc-4de2-4ba5-b42c-0224f341c300.png)

with user-defined consolidator (i.e. `selectUserDefinedConsolidator = True`):
![image](https://user-images.githubusercontent.com/68070194/135417310-69a9c768-2661-4219-bdc9-9a6b783829c8.png)

C# version:

```C#
namespace QuantConnect.Algorithm.CSharp
{
    public class UpgradedAsparagusParrot : QCAlgorithm
    {
    	private SimpleMovingAverage _sma;

        public override void Initialize()
        {
            SetStartDate(2010,1,1);
            SetEndDate(2020,1,1);
            SetCash(100000);     
            
			var spy = AddEquity("SPY", Resolution.Minute).Symbol;
[...]

PR #5960, opened 2021-10-04
Adds IndicatorBatchUpdate

### LINKED ISSUE #5634: Implement History.PushThrough for Python Indicators
#### Expected Behavior
Be able to update a series of indicators with a single command.

#### Actual Behavior
Python algorithm needs to make a historical data request that returns a pandas DataFrame that will be used to update indicators. If the algorithm uses `EnableAutomaticIndicatorWarmUp` instead, it will make the same historical data request if there is more than one indicator per Symbol. Additionally, it will not warm up chained indicators (e.g.: `IndicatorExtensions.Of`), because they are not supported (there is no chain `WarmupPeriod`).
In C#, this is easily handled with 
```csharp
History(symbol, period, resolution).PushThrough(bar => indicator.Update(bar.EndTime, bar.Price));
```

#### Potential Solution
Implement a method that updates a list of indicators:
E.g.:
```python
self.IndicatorBatchUpdate(symbol, period, resolution, indicators)
```

#### Checklist

ArthurAsenheimer:
#### Description

In order to warm up indicators, Python algorithms needs to make a historical data request that returns a pandas DataFrame that will be used to update indicators. 
This approach is slow and inefficient. Python algorithms with universes that are frequently changing spend a large amount of time on warm ups. This is especially true when higher resolution data is used and/or many securities are added/removed. 

IndicatorBatchUpdate aims to solve this problem by avoiding the conversion from `Enumerable<Slice>` to `pandas.DataFrame` and providing Python developers an alternative to `history.PushThrough(data => indicator.Update(data))` that is significantly faster. 

IndicatorBatchUpdate supports various use cases. For instance, it also works for a custom quote bare indicator if the associated security type is equity and has not yet been added to the SubscriptionManager. 

#### Limitations and other notes: 
User-defined selectors are not reflected in IndicatorBatchUpdate, but ConsolidatorBatchUpdate can be used an an alternative here. The latter offers a bit more flexibility. 

Resolution.Tick is not supported due to `typeof(period) = int`, see also [line 334 in QCAlgorithm.History.cs](https://github.com/QuantConnect/Lean/blob/9579c4263bc8681dc7d8499f8f58252aa41c917c/Algorithm/QCAlgorithm.History.cs#L334). 

#### Performance analysis: 

image

image

image

#### Conclusion: 
IndicatorBatchUpdate brings a time saving of 80% compared to the currently recommended version with DataFrames. ConsolidatorBatchUpdate, on the other hand, is still slightly faster, but significantly slower than IndicatorBatchUpdate. Please refer to #5955 for more information. A solution of #5955 should bring the performance of ConsolidatorBatchUpdate close to those of IndicatorBatchUpdate. 

#### Related Issue

Closes #5634 

#### Motivation and Context

Faster and more efficient warm ups.

N/A

Regression and unit tests.

#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -2530,6 +2530,75 @@ private void WarmUpIndicatorImpl<T>(Symbol symbol, TimeSpan period, Action<T> ha
 
             SubscriptionManager.RemoveConsolidator(symbol, consolidator);
         }
+        
+        /// <summary>
+        /// Warms up a given list of indicators with historical data
+        /// </summary>
+        /// <param name="symbol">The symbol whose indicators we want</param>
+        /// <param name="indicators">The indicators we want to warm up</param>
+        /// <param name="period">The necessary period to warm up the indicators</param>
```
AlexCatarino, line 2539: In some cases, the minimum number of samples that an indicator needs to be fully warmed up is not trivial, so the user might provide a period that is not big enough. Consequently, the `period` should be defined by the maximum value of `IIndicatorWarmUpPeriodProvider.WarmUpPeriod` (see `GetIndicatorWarmUpHistory` for reference), and this parameter can be removed.

ArthurAsenheimer: I've been thinking about that, too. But what about custom python indicators which don't have a `WarmUpPeriod` as far as I know? Maybe a better solution would be setting `int period = null` as default and if period is null, we use the approach you described. What do you think?

PR #6005, opened 2021-10-22
First prototype for European market

immutagmbh:
First draft to enable submitting orders for the European market.

Currently work in progress. Hints are always welcome.


uwehartmann:
Thank you @jaredbroad. I am happy to work on Lean. You are right. Done.

jaredbroad:
The market concept in LEAN is to reflect that an asset is fungible across multiple exchanges. E.g. NASDAQ and NYSE share listings and people are able to trade freely between them. I can buy AAPL then decide to route it to NYSE or NASDAQ to sell.

Since the assets are different securities in different markets; they have different prices (or even denominating currencies) so they are stored in separate data folders. E.g. Is there an "A" ticker in LSE and MD? Should they be stored in different data folders? 

I may be wrong but I don't think a Madrid listed stock can trade on the LSE ETF exchange?  I'm not sure there is a central market the same way there is in the US? We might need different markets for each major exchange.

jaredbroad:
If keen to work on European support with LEAN please join our discord, it would be great to brainstorm it with you and make a plan for how it could be implemented. There are many pieces of this PR that would be a good start. https://www.quantconnect.com/discord

Issue #5816, opened 2021-08-02
[Crypto] Implement FTX Crypto Exchange


AlexCatarino:
#### Expected Behavior
Lean supports [FTX](https://ftx.com/) bitcoin exchange.

#### Actual Behavior
Not supported.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

#### Related:
Kraken #2003

PR #6026, opened 2021-10-28
FTX API endpoint is cinfigurable 

AdalyatNazirov:
#### Description

specify enpoint url of ftx api (can be FTX pr FTX.US). FTX and FTX.US are considered as entirely different exchanges

#### Related Issue

relates #5816 

#### Motivation and Context

FTX is running different exchange (and endpoing) for trading in USA

N/A

#### Types of changes

#### Checklist:


AdalyatNazirov:
> 🤔 I wonder if there's a different way to do this. It won't scale well to have countries carved out of the Market enum. I think we only have 255 slots. Are the FTX US BTCUSD fungible and able to be moved and traded on the FTX exchange?

initially it was the idea, but then we found that FTX and FTX.US are different - not only endpoints, but also different history, listed symbols etc. It made us to change a decision on implementation and create another brokerage - `FTXUSBrokerage: FTXBrokerage`, with overridden bits responsible for diffs. So it will even will have it's own `live-ftxus` section in config file.

Common/Brokerages/FTXUSBrokerageModel.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Market name
+        /// </summary>
+        protected override string MarketName => Market.FTXUS;
+
+        /// <summary>
+        /// Gets a map of the default markets to be used for each security type
+        /// </summary>
+        public override IReadOnlyDictionary<SecurityType, string> DefaultMarkets { get; } = GetDefaultMarkets(Market.FTXUS);
```
Martin-Molinero, line 40: Cool! I think if underlying directly used `MarketName` we wouldn't need to override it again?
nit - mind moving constructor after properties/fields

Launcher/config.json
```diff
@@ -443,6 +450,21 @@
...
       "history-provider": "BrokerageHistoryProvider"
+    },
+
+    // defines the 'live-ftxus' environment
+    "live-ftxus": {
+      "live-mode": true,
+
+      // real brokerage implementations require the BrokerageTransactionHandler
+      "live-mode-brokerage": "QuantConnect.FTXBrokerage.FTXUSBrokerage",
+      "data-queue-handler": "QuantConnect.FTXBrokerage.FTXUSBrokerage",
```
Martin-Molinero, line 461: > "data-queue-handler"

Value should be an array moving forward 👍

Launcher/config.json
```diff
@@ -194,6 +194,11 @@
   "ftx-api-secret": "",
   "ftx-api-key": "",
 
+  // FTX.US configuration
+  "ftxus-account-tier": "Tier1", // accept "Tier1", "Tier2", "Tier3", "Tier4", "Tier5", "Tier6", "VIP1", "VIP2", "VIP3", "MM1", "MM2", "MM3"
```
Martin-Molinero, line 198: They have the same tier values? cool 👍

Tests/Common/Brokerages/FTXUSBrokerageModelTests.cs
```diff
@@ -0,0 +1,35 @@
...
+*/
+
+using NUnit.Framework;
+using QuantConnect.Brokerages;
+using QuantConnect.Orders.Fees;
+
+namespace QuantConnect.Tests.Common.Brokerages
+{
+    [TestFixture]
+    public class FTXUSBrokerageModelTests : FTXBrokerageModelTests
```
Martin-Molinero, line 23: Nice!

Martin-Molinero (approved):
Sweet! leaving couple of minor comments 👍

Common/Brokerages/FTXUSBrokerageModel.cs
AdalyatNazirov: it requires `MarkerName` to be static field.

Launcher/config.json
AdalyatNazirov: Apparently it has more tiers (+7, 8, 9) and only VIP1 and VIP2. Fixed

Martin-Molinero (approved):

Issue #6056, opened 2021-11-13
Data downloader improvements for rate limit apis


bmello4688:
#### Expected Behavior

if the DownloaderDataProvider cannot get the correct ticktype/datatype then the data downloader should not fetch the data from the abstracted api. All API's have data fetching limitations, in this case it would be better to not make the call.

#### Actual Behavior

The call is made to the api and values are ignored if not the correct data type

#### Potential Solution

Pass ticktype into the api call in order for the implementing class to decide what to do.

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

PR #6057, opened 2021-11-13
Add tick type to data downloader get (price data) call in order for a…

bmello4688:
…pi's with rate limits to handle not making an api call at all.

#### Description

Add tick type to data downloader get call so that api limits are not incremented when data does not match type that lean is requesting

#### Related Issue

#6056 

#### Motivation and Context

Reduces rate limits being hit

No

Unit tests and integration tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/DataDownloaderGetParameters.cs
+++ b/Common/DataDownloaderGetParameters.cs
@@ -1,3 +1,18 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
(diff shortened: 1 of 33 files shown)


Martin-Molinero (changes requested):
Hey @bmello4688! Thanks for the PR,

To avoid future breaking changes like this one suggest we introduce a new DTO class `DataDownloaderGetParameters`, following what we've done before in the [BuyingPowerModels](https://github.com/QuantConnect/Lean/blob/master/Common/Securities/BuyingPowerModel.cs) which can hold all of the current required method arguments

Martin-Molinero (approved):
Thanks!

PR #6068, opened 2021-11-22
Add Overload in SecurityTransactionManager Methods to Support Python Functions

### LINKED ISSUE #4514: Add Overload in SecurityTransactionManager Methods to Support Python Functions
#### Expected Behavior
Be able to use Python functions in the methods that accept, for e.g.:, `Func<Order, bool>` parameter.

#### Actual Behavior
Not supported:

#### Potential Solution
Implement the overloads.

#### Reproducing the Problem
The following is not available
```python
orders = self.Transactions.GetOrders(lambda o: o.Status == OrderStatus.Filled)
```
as an equivalent of 
```csharp
var orders = Transactions.GetOrders(o => o.Status == OrderStatus.Filled);
```
#### Checklist

Marinovsky:
#### Description

- Add overload in `SecurityTransactionManager` class to support Python functions as parameter in some methods
- Add unit tests and regression tests to ensure the bug is gone
#### Related Issue

Closes #4514
#### Motivation and Context

With this changes python users will be able to pass python functions as parameters to some methods of `SecurityTransactionManager` class

This change does not require changes or updates to documentation

- In the unit test it was created a QCAlgorithm and submitted two orders for different symbols and then asserted if the number of elements returned by `GetOrderTickets()`, `GetOrders()`, `GetOpenOrders()`, `GetOpenOrderTickets()` and `GetOpenOrdersRemainingQuantity()` methods, using PyObjects filters as parameters, were two.
- It was used the regression test `OrderTicketDemoAlgorithm.cs/py` to check the error was gone. At the end of the algorithm it was asserted if number of filled orders were the expected ones(8 orders) and that there were not open orders still, using `GetOrderTickets()`, `GetOrders()`, `GetOpenOrders()`, `GetOpenOrderTickets()` and `GetOpenOrdersRemainingQuantity()` methods methods with Python functions as parameters.
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
+++ b/Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
@@ -488,10 +488,10 @@ namespace QuantConnect.Algorithm.CSharp
             }
 
             var filledOrderTickets = Transactions.GetOrderTickets(x => x.Status == OrderStatus.Filled);
-
+            Log("Order ticket types");
             foreach (var ticket in filledOrderTickets)
             {
-                Log($"Ticket symbol: {ticket.Symbol}");
+                Log($"Order ticket type: {ticket.OrderType}");
             }
         }
 
--- a/Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
+++ b/Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
@@ -509,6 +509,25 @@ namespace QuantConnect.Algorithm.CSharp
             {
                 throw new Exception($"No remaining quantiy to be filled from open orders was expected");
             }
+
+            var defaultOrders = Transactions.GetOrders();
+            var defaultOrderTickets = Transactions.GetOrderTickets();
+            var defaultOpenOrders = Transactions.GetOpenOrders();
+            var defaultOpenOrderTickets = Transactions.GetOpenOrderTickets();
+            var defaultOpenOrdersRemaining = Transactions.GetOpenOrdersRemainingQuantity();
+
+            if (defaultOrders.Count() != 10 || defaultOrderTickets.Count() != 10)
+            {
+                throw new Exception($"There were expected 10 orders and 10 order tickets");
+            }
+            if (defaultOpenOrders.Count != 0 || defaultOpenOrderTickets.Any())
+            {
+                throw new Exception($"No open orders or tickets were expected");
+            }
+            if (defaultOpenOrdersRemaining != 0m)
+            {
+                throw new Exception($"No remaining quantiy to be filled from open orders was expected");
+            }
         }
 
         /// <summary>
--- a/Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
+++ b/Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
@@ -476,22 +476,25 @@ namespace QuantConnect.Algorithm.CSharp
 
         public override void OnEndOfAlgorithm()
         {
-            var filledOrders = Transactions.GetOrders(x => x.Status == OrderStatus.Filled);
-            var openOrders = Transactions.GetOpenOrders(x => true);
-            if (filledOrders.Count() != 8)
-            {
-                throw new Exception($"There were expected 8 filled orders but there were only {filledOrders.Count()}");
-            }
-            if (openOrders.Count != 0)
-            {
-                throw new Exception($"No open order was expected but there were {openOrders.Count} orders");
-            }
+            Func<OrderTicket, bool> basicOrderTicketFilter = x => x.Symbol == symbol;
 
-            var filledOrderTickets = Transactions.GetOrderTickets(x => x.Status == OrderStatus.Filled);
-            Log("Order ticket types");
-            foreach (var ticket in filledOrderTickets)
+            var filledOrders = Transactions.GetOrders(x => x.Status == OrderStatus.Filled);
+            var orderTickets = Transactions.GetOrderTickets(basicOrderTicketFilter);
+            var openOrders = Transactions.GetOpenOrders(x => x.Symbol == symbol);
+            var openOrderTickets = Transactions.GetOpenOrderTickets(basicOrderTicketFilter);
+            var remainingOpenOrders = Transactions.GetOpenOrdersRemainingQuantity(basicOrderTicketFilter);
+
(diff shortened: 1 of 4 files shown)


Algorithm.CSharp/OrderTicketDemoAlgorithm.cs
```diff
@@ -473,6 +474,27 @@ private bool TimeIs(int day, int hour, int minute)
             return Time.Day == day && Time.Hour == hour && Time.Minute == minute;
         }
 
+        public override void OnEndOfAlgorithm()
+        {
+            var filledOrders = Transactions.GetOrders(x => x.Status == OrderStatus.Filled);
+            var openOrders = Transactions.GetOpenOrders(x => true);
```
Martin-Molinero, line 480: Cool! 
- `x => true` maybe can check the symbol is `X symbol`

- `Log()` suggest avoid logging if not required, and only using `Debug` when desired. For regression tests there is no real value for logging and causes issues really (there is a max output size, etc)

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -332,6 +364,18 @@ public List<Order> GetOpenOrders(Func<Order, bool> filter = null)
             return _orderProcessor.GetOpenOrders(x => filter(x));
         }
 
+        /// <summary>
+        /// Gets open orders matching the specified filter. Specifying null will return an enumerable
+        /// of all open orders.
```
Martin-Molinero, line 369: > Specifying null will return an enumerable of all open orders.

This will probably blow up? since `GetOpenOrders(Func<Order, bool> filter = null)` covers no filter case, don't think we need to support users sending `None`?

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -371,6 +415,17 @@ public IEnumerable<Order> GetOrders(Func<Order, bool> filter)
             return _orderProcessor.GetOrders(filter);
         }
 
+        /// <summary>
+        /// Gets all orders matching the specified filter. Specifying null will return an enumerable
+        /// of all orders.
```
Martin-Molinero, line 420: Same comment about None/null, maybe `IEnumerable<Order> GetOrders(Func<Order, bool> filter)` should have a default `null` value?

Tests/Common/Securities/SecurityTransactionManagerTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                var openOrdersTickets = transactions.GetOpenOrderTickets(basicOrderTicketFilter.ToPython());
+                var openOrdersRemaining = transactions.GetOpenOrdersRemainingQuantity(basicOrderTicketFilter.ToPython());
+                Assert.AreEqual(2, orders.Count());
+                Assert.AreEqual(2, orderTickets.Count());
+                Assert.AreEqual(0, openOrders.Count);
+                Assert.AreEqual(0, openOrdersTickets.Count());
+                Assert.AreEqual(0, openOrdersRemaining);
+                foreach(var ticket in orderTickets)
+                {
+                    Log.Trace($"Ticket symbol: {ticket.Symbol} - Status: {ticket.Status}");
```
Martin-Molinero, line 119: Same comment about logging 👍 generally tests should log rarely in debug mode only. 
I think the unit test could be simplified, really needs a `SecurityTransactionManager` and some `TestOrderProcessor`, can create it passing in the orders you want I think

Martin-Molinero (changes requested):
Nice! leaving a couple of requests

Tests/Common/Securities/SecurityTransactionManagerTests.cs
```diff
@@ -57,67 +45,43 @@ public void SetUp()
...
             var algorithm = new QCAlgorithm();
-            var securities = new SecurityManager(TimeKeeper);
-            var transactions = new SecurityTransactionManager(null, securities);
-            var transactionHandler = new BacktestingTransactionHandler();
-            var portfolio = new SecurityPortfolioManager(securities, transactions);
+            algorithm.SubscriptionManager.SetDataManager(new DataManagerStub(algorithm));
+            var spySecurity = algorithm.AddEquity("SPY");
+            var ibmSecurity = algorithm.AddEquity("IBM");
+            spySecurity.Exchange = new SecurityExchange(SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork));
+            ibmSecurity.Exchange = new SecurityExchange(SecurityExchangeHours.AlwaysOpen(TimeZones.NewYork));
```
Martin-Molinero, line 52: Believe if you specify the current algorithm time you can avoid these lines

Tests/Common/Securities/SecurityTransactionManagerTests.cs
```diff
@@ -57,67 +45,43 @@ public void SetUp()
...
 
-            algorithm.Securities = securities;
-            transactionHandler.Initialize(algorithm, new BacktestingBrokerage(algorithm), _resultHandler);
-            transactions.SetOrderProcessor(transactionHandler);
+            var transactionHandler = new BrokerageTransactionHandler();
 
-            // Adding cash: strike price times number of shares
-            portfolio.SetCash(192 * 100);
+            transactionHandler.Initialize(algorithm, new BacktestingBrokerage(algorithm), _resultHandler);
+            Thread.Sleep(250);
```
Martin-Molinero, line 60: Sleeping shouldn't be required, same about the sleep bellow 👍

Common/Securities/SecurityTransactionManager.cs
```diff
@@ -365,15 +365,25 @@ public List<Order> GetOpenOrders(Func<Order, bool> filter = null)
         }
 
         /// <summary>
-        /// Gets open orders matching the specified filter. Specifying null will return an enumerable
-        /// of all open orders.
+        /// Gets open orders matching the specified filter. However this method can be confused with the
+        /// override that takes a Symbol as parameter. For this reason it first checks if it can convert
+        /// the parameter into a symbol. If that conversion cannot be aplied it assumes the parameter is
+        /// a Python function object and not a Python representation of a Symbol.
```
Martin-Molinero, line 371: Nice catch! Can the same happen for some of the other methods here, like `GetOpenOrderTickets`? Would be nice to expand the test calls a bit to cover all possible cases like these too, another example missing to test is `GetOrders()` I think?

Martin-Molinero (approved):
Sweet! 🍭

PR #6128, opened 2021-12-23
Bug: Config.TryGetValue returns true if key is not found

omidkrad:
Unlike the method's documentation, [`Config.TryGetValue`](https://github.com/QuantConnect/Lean/blob/3d3733c0fb5a167669e99bb4cc74470c8c3f22c7/Configuration/Config.cs#L295-L329) will actually return true if `key` is not found and using `defaultValue`. It will only return false if `key` is found and fails to parse. This pull request fixes it to return false if key is not found. `value` is still set to `defaultValue`.

### DIFF
--- a/Algorithm.CSharp/ParameterizedAlgorithm.cs
+++ b/Algorithm.CSharp/ParameterizedAlgorithm.cs
@@ -28,9 +28,9 @@ namespace QuantConnect.Algorithm.CSharp
     /// <meta name="tag" content="using quantconnect" />
     public class ParameterizedAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
     {
-        // we place attributes on top of our fields or properties that should receive
+        // We place attributes on top of our fields or properties that should receive
         // their values from the job. The values 100 and 200 are just default values that
-        // or only used if the parameters do not exist
+        // are only used if the parameters do not exist.
         [Parameter("ema-fast")]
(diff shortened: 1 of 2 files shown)


omidkrad:
You can alternatively correct this method's doc instead of accepting this change since current behavior might be desired. For example:

```
/// <returns>True on successful parse or if key is not found. false only when key is found but fails to parse.</returns>
```

omidkrad:
You can alternatively correct this method's doc instead of accepting this change since current behavior might be desired. For example:

```xml
/// <returns>True on successful parse or if key is not found. false only when key is found but fails to parse.</returns>
```

| key found | parsed | return | out |
| ---- | ---- | ---- | ---- |
| no |  n/a | `true` | `defaultValue` |
| yes | yes | `true` | parsed value |
| yes | no | `false` | `defaultValue` |

jaredbroad:
Hey @omidkrad thank you for the PR: we'll review but I'm curious what was the pretext for this? Did it cause an issue somewhere else?

omidkrad:
@jaredbroad not really. I noticed the method's doc was not consistent with its behavior so I tried to fix it. But now I think the current behavior is more useful that it returns false only when it cannot parse. I updated the PR to just update the docs.

Martin-Molinero (approved):
Thanks @omidkrad!

Issue #6140, opened 2022-01-05
Add Sortino as Indicator 


jaredbroad:
#### Expected Behavior
Sortino Ratio Indicator in LEAN - `SORTINO()` helper with the base class "SortinoRatioIndicator.cs" 

`SortinoRatio(string name, int period, decimal riskFreeRate = 0.0m)`

#### Actual Behavior
No sortino ratio indicator

#### Potential Solution
Implement a sortino ratio indicator 

Follow example of https://github.com/QuantConnect/Lean/blob/master/Indicators/SharpeRatio.cs including tests and external data sources.

#### Checklist 
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues

PR #6150, opened 2022-01-10
Sortino Ratio Indicator Implementation

JerrodDev:
#### Description
Implementation of the SORTINO indicator class alongside the required unit tests and sample data. Addition of SORTINO method in QCAlgorithm.Indicators class.

#### Related Issue

 [https://github.com/QuantConnect/Lean/issues/6140](Issue)

#### Motivation and Context

No sortino ratio indicator exists, these changes implement this functionality. 

Requires addition to the available indicators that users may need to use.

This change was unit tested using the standard unit test classes and the indicator was validated using sample data with a 10 period trailing sortino ratio.

#### Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

#### Checklist:

- [ X ] My code follows the code style of this project.
- [ X ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [ X ] I have added tests to cover my changes. 
- [ X ] All new and existing tests passed.
- [ X ] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`


Tests/Indicators/SortinoRatioTests.cs
```diff
@@ -0,0 +1,72 @@
+using NUnit.Framework;
```
Martin-Molinero, line 1: Would be nice to add some unit tests with a risk rate different than default 0 value too 👍 
`new DateTime()` for it to be more realistic suggest creating a valid datetime on start and just adding a day per `i`
nit
- missing license header
- Should use `var` to the left always when possible

Indicators/SortinoRatio.cs
```diff
@@ -0,0 +1,121 @@
...
+                if (_roc.Current.Value > riskFreeRate)
+                    _roc.Current.Value = riskFreeRate;
+            };
+            var std = new StandardDeviation(period).Of(_rocForSTD);
+            var sma = _rocForSMA.SMA(period);
+            _sortinoRatio = sma.Minus(riskFreeRate).Over(std);
+
+            // define warmup value; 
+            // _roc is the base of our indicator chain + period of STD and SMA
+            WarmUpPeriod = _rocForSMA.WarmUpPeriod + _period;
```
Martin-Molinero, line 95: Cool!
- The implementation is **very** similar to Sharpe Ratio one, seems they could share their implementation with some inheritance maybe, what would change is the `StandardDeviation` vs `StandardDownsideDeviation`?
- `std/sma/_roc/_rocForSTD/_rocForSMA` following Lean coding style ideally these should be expanded to their full version, and maybe instead of their name being they type it could reflect their purpose/nature?
- Think the `Standard Downside Deviation` should be it's own indicator with his tests? Not sure the currently implementation is right, based on https://www.investopedia.com/terms/d/downside-deviation.asp should only take negative returns after subtracting the free rate?
- `_rocForSTD` ideally we shouldn't be changing the indicators value after it's being calculated, seems like this should live inside the indicator or in the consumer of it's value

Martin-Molinero (changes requested):
Thanks! Sorry for the delayed review, leaving a couple of requests
Welcome to Lean repo BTW!

Indicators/SortinoRatio.cs
AlexCatarino: I was about to suggest only one `RateOfChange` to simplify the code a little bit, and remove the recalculation of `_rocForStd`, but the idea of implementing `Standard Downside Deviation` is very good since it might be used in other applications.

JerrodDev: Thanks @AlexCatarino, I’ll have some updated changes pushed soon!

JerrodDev:
Hey Martin, apologies for dropping the ball here. I am getting over a bout of Covid and got swamped with everything I missed while down. All that is left is sourcing test data to validate Downside Deviation and Sortino, I’ll make sure to get this over to you this weekend!

Jerrod Smith
Vexor Holdings, LLC
________________________________

Indicators/StandardDownsideDeviation.cs
```diff
@@ -0,0 +1,62 @@
+using System;
```
Martin-Molinero, line 1: nit - missing license header file

Think `StandardDownsideDeviation` should have unit tests of it's own

Tests/Indicators/SortinoRatioTests.cs
Martin-Molinero: These seem unaddressed @JerrodDev

Martin-Molinero (changes requested):
Nice improvement! There are still a couple of requests though

Issue #4729, opened 2020-09-18
SetWarmup to Fetch history from disk in live mode


captholley:
#### Expected Behavior

SetWarmup() request at initialization performs an API call while in live mode to the exchange being used at that time. The exchange I am using does not have enough historical data for the security I am trading and I would like to use data from files stored on disk the same way I would during backtest mode. I believe this may also speed up the warmup period as the data will be quicker to access and won't run into rate limiting issues.
#### Actual Behavior

SetWarmup() in live mode makes continuous API calls to retrieve history
#### Potential Solution

Allow the use of disk data, automatically switching to api if data is not available

#### Checklist

- [x ] I have completely filled out this template
- [x ] I have confirmed that this issue exists on the current `master` branch
- [x ] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

Martin-Molinero:
Hey @captholley!
Interesting use case, currently Lean only allows specifying a single history provider. A workaround for your use case is having a custom history provider wrapping both providers you want to use, a file-based one, and the brokerage one and switching on each acordengly.

PR #6187, opened 2022-02-04
Feature handle multiple history providers

rjra2611:
#### Description

- Add `HistoryProviderManager` for LEAN to use as a single history provider which works as a wrapper around multiple history providers.
- History request would fetch data from all history providers and the received data would get merged into forwarding nonduplicate data

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/4729

#### Motivation and Context

- Access historical data from more than one history provider in a single request

N/A

Tests have been added
#### Types of changes

#### Checklist:


Common/Data/Slice.cs
```diff
@@ -444,6 +446,50 @@ public override bool TryGetValue(Symbol symbol, out dynamic data)
...
 
+        /// <summary>
+        /// Merge two slice with same Time
+        /// </summary>
+        /// <param name="inputSlice">slice instance</param>
+        public void MergeSlice(Slice inputSlice)
+        {
+            if (Time != inputSlice.Time)
+            {
+                throw new Exception($"Slice with time {Time} can't be merged with given slice with different {inputSlice.Time}");
```
Martin-Molinero, line 457: - Maybe using ArgumentException or InvalidOperationException might be more accurate 👌 
- `<TradeBar>/<QuoteBar>/...` think these aren't required since they are implicitly figured out by the compiler

Common/Data/Slice.cs
```diff
@@ -444,6 +446,50 @@ public override bool TryGetValue(Symbol symbol, out dynamic data)
...
+            var tempRawDataList = inputSlice._rawDataList;
+            tempRawDataList.AddRange(_rawDataList);
+            _rawDataList = tempRawDataList; 
+            _data = new Lazy<DataDictionary<SymbolData>>(() => CreateDynamicDataDictionary(_rawDataList));
+        }
+
+        private static DataDictionary<T> UpdateCollection<T>(DataDictionary<T> baseCollection, DataDictionary<T> inputCollection)
+        {
+            if (inputCollection?.Count > 0)
+            {
```
Martin-Molinero, line 481: Cool! Can check the count of `baseCollection` if 0 could directly return `inputCollection` instead?
See `TimeSliceFactory` he will **reuse empty collections** when possible and clear them just in case on each new slice creation, so we should keep empty collections empty and just update/use the non empty ones

Common/Data/Slice.cs
```diff
@@ -444,6 +446,50 @@ public override bool TryGetValue(Symbol symbol, out dynamic data)
...
+            _splits = (Splits)UpdateCollection<Split>(_splits, inputSlice.Splits);
+            _dividends = (Dividends)UpdateCollection<Dividend>(_dividends, inputSlice.Dividends);
+            _delistings = (Delistings)UpdateCollection<Delisting>(_delistings, inputSlice.Delistings);
+            _symbolChangedEvents = (SymbolChangedEvents)UpdateCollection<SymbolChangedEvent>(_symbolChangedEvents, inputSlice.SymbolChangedEvents);
+
+            // Should keep this._rawDataList last so that selected data points are not overriden
+            // while creating _data
+            var tempRawDataList = inputSlice._rawDataList;
+            tempRawDataList.AddRange(_rawDataList);
+            _rawDataList = tempRawDataList;
```
Martin-Molinero, line 474: Similar to `UpdateCollection<T>` can check counts to be super lazy about the `AddRange` call 👍

Engine/DataFeeds/Enumerators/SynchronizingSliceEnumerator.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using QuantConnect.Data;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
+{
+    /// <summary>
+    /// Represents an enumerator capable of synchronizing other base data enumerators in time.
```
Martin-Molinero, line 25: Docs seem to need an update for this guy 👍

Engine/DataFeeds/Enumerators/SynchronizingEnumerator.cs
```diff
@@ -26,18 +26,19 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators
...
     /// This assumes that all enumerators have data time stamped in the same time zone
     /// </summary>
-    public class SynchronizingEnumerator : IEnumerator<BaseData>
+    public abstract class SynchronizingEnumerator<T> : IEnumerator<T>
     {
-        private IEnumerator<BaseData> _syncer;
-        private readonly IEnumerator<BaseData>[] _enumerators;
+        private IEnumerator<T> _syncer;
+        private readonly IEnumerator<T>[] _enumerators;
+        protected abstract DateTime GetInstanceTime<T>(T instance);
```
Martin-Molinero, line 33: - Let's add the docs for `GetInstanceTime` 👍 
- Constructors can now be protected too

Engine/DataFeeds/HistoryProviderManager.cs
```diff
@@ -0,0 +1,138 @@
...
+    /// Provides an implementation of <see cref="IHistoryProvider"/> which
+    /// acts as a wrapper to use multiple history providers together
+    /// </summary>
+    public class HistoryProviderManager : SynchronizingHistoryProvider
+    {
+        /// <summary>
+        /// Collection of history providers being used
+        /// </summary>
+        /// <remarks>Protected for testing purposes</remarks>
+        protected List<IHistoryProvider> HistoryProviders { get; } = new();
```
Martin-Molinero, line 39: - privates should go at the top
- let's move `HistoryProviderManager` under `HistoricalData` instead of `DataFeeds`

Engine/DataFeeds/HistoryProviderManager.cs
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+            var synchronizer = new SynchronizingSliceEnumerator(historyEnumerators);
+            var previousTime = DateTime.MinValue;
+            Slice previousMergedSlice = null;
+            Slice latestMergeSlice = null;
+            bool isFirstComplete = new();
+            while (synchronizer.MoveNext())
+            {
+                if (synchronizer.Current.Time > previousTime)
```
Martin-Molinero, line 112: Cool! we might add a null check for `synchronizer.Current` just in case 👍

Tests/Common/Data/SliceTests.cs
```diff
@@ -206,6 +207,122 @@ public void AccessesGenericallyByTypeAndSymbol()
             Assert.AreEqual(unlinkedDataSpy, unlinkedData);
         }
 
+        [Test]
+        public void MergeSlice()
+        {
+            TradeBar tradeBar1 = new TradeBar { Symbol = Symbols.SPY, Time = DateTime.Now };
+            TradeBar tradeBar2 = new TradeBar { Symbol = Symbols.AAPL, Time = DateTime.Now, Open = 23 };
```
Martin-Molinero, line 214: nit - `var`. Should avoid using `DateTime.Now` is possible an use `DateTime.UtcNow` instead, avoid the tz conversion and will already return utc. Ideally get time at the top once an re use it? or even just create a fixed `DateTime` instance

Tests/Common/Data/SliceTests.cs
```diff
@@ -206,6 +207,122 @@ public void AccessesGenericallyByTypeAndSymbol()
...
+            var custom1 = new FxcmVolume { DataType = MarketDataType.Base, Symbol = Symbols.MSFT };
+            var custom2 = new FxcmVolume { DataType = MarketDataType.Base, Symbol = Symbols.SBIN };
+            Slice slice6 = new Slice(DateTime.Today, new BaseData[] { custom1, custom2, tradeBar2 });
+            Slice slice5 = new Slice(DateTime.Today, new BaseData[] { tradeBar1 });
+            slice5.MergeSlice(slice6);
+            Assert.AreEqual(4, slice5.Values.Count);
+            Assert.AreEqual(2, slice5.Values.Where(x => x.DataType == MarketDataType.Base).Count());
```
Martin-Molinero, line 323: Cool! similar request as the Ticks tests, let's add custom data with the same symbol in both slices and expect it to use 1 after the merge

Tests/Engine/DataFeeds/Enumerators/SynchronizingSliceEnumeratorTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        public void WillStopIfAllEnumeratorsCurrentIsNullAndReturningTrue()
+        {
+            var stream1 = new TestEnumerator { MoveNextReturnValue = true };
+            var synchronizer = new SynchronizingSliceEnumerator(stream1);
+            while (synchronizer.MoveNext())
+            {
+                Assert.Fail();
+            }
+            Assert.IsTrue(stream1.MoveNextWasCalled);
+            Assert.Pass();
```
Martin-Molinero, line 104: Think we don't need this guy `Assert.Pass();` ? same bellow

Martin-Molinero (commented):
Cool! looks good, almost there leaving a few requests

Engine/DataFeeds/Enumerators/SynchronizingBaseDataEnumerator.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// Initializes a new instance of the <see cref="SynchronizingBaseDataEnumerator"/> class
+        /// </summary>
+        /// <param name="enumerators">The enumerators to be synchronized. NOTE: Assumes the same time zone for all data</param>
+        public SynchronizingBaseDataEnumerator(IEnumerable<IEnumerator> enumerators) : base((IEnumerable<IEnumerator<BaseData>>)enumerators)
+        {
+        }
+
+        protected override DateTime GetInstanceTime<T>(T instance)
+        {
+            return (instance as BaseData).EndTime;
```
Martin-Molinero, line 53: Shouldn't need to be casting here, since ur specifying `T`, the problem is that `<T>` in `GetInstanceTime<T>` is actually saying it's a different `T`.
Should be
```csharp
        protected abstract DateTime GetInstanceTime(T instance);


        protected override DateTime GetInstanceTime(BaseData instance)
        {
            return instance.EndTime;
        }
```

Engine/HistoricalData/HistoryProviderManager.cs
```diff
@@ -0,0 +1,143 @@
...
+                    latestMergeSlice.MergeSlice(synchronizer.Current);
+                }
+                previousMergedSlice = latestMergeSlice;
+            }
+            synchronizer.Dispose();
+            if (!isFirstComplete)
+            {
+                yield break;
+            }
+            yield return previousMergedSlice;
```
Martin-Molinero, line 140: Think we could reduce a few variables here, something like:?
```suggestion
            using var synchronizer = new SynchronizingSliceEnumerator(historyEnumerators);
            Slice latestMergeSlice = null;
            while (synchronizer.MoveNext())
            {
                if (synchronizer.Current == null)
                {
                    continue;
                }
                if (latestMergeSlice == null)
                {
                    latestMergeSlice = synchronizer.Current;
                    continue;
                }

                if (synchronizer.Current.Time > latestMergeSlice.Time)
                {
                    // a newer slice we emit the old and keep a reference of the new
                    // so in the next loop we merge if required
                    yield return latestMergeSlice;
                    latestMergeSlice = synchronizer.Current;
                }
                else
                {
                    // a new slice with same time we merge them into 'latestMergeSlice'
                    latestMergeSlice.MergeSlice(synchronizer.Current);
                }
            }

            if (latestMergeSlice != null)
            {
                yield return latestMergeSlice;
            }
```

Martin-Molinero (approved):
✨

Martin-Molinero (approved):

PR #6227, opened 2022-02-21
Reconcile the Market Opening for Futures and Scheduled Events

### LINKED ISSUE #6189: Reconcile the Market Opening and Closing For Futures and Scheduled Events
#### Expected Behavior
`AfterMarketOpen` for futures is triggered by the real market opening time.

#### Actual Behavior
Since we break the MHDB considers defined a start or the day that matches the close of the previous, causing `AfterMarketOpen` triggering at midnight.

#### Potential Solution
N/A

#### Reproducing the Problem
```python
class PensiveFluorescentYellowParrot(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2022, 1, 1)
        gdp = self.AddFuture("6B").Symbol
        
        self.Schedule.On(self.DateRules.EveryDay(gdp), 
                 self.TimeRules.AfterMarketOpen(gdp),  
                 self.EveryDayAfterMarketOpen)

    def EveryDayAfterMarketOpen(self):
        self.Log(f'{self.Time}')
```
will **not** only log the real open at 5 pm. E.g. 
```
2022-01-26 01:00:00 : | 2022-01-26 01:00:00
2022-01-27 01:00:00 : | 2022-01-27 01:00:00
```
Jan 26th, 2022 was a Wednesday, so there was no opening event at midnight. 

#### Checklist

Marinovsky:
#### Description

- Fix bug in `GetNextMarketOpen()` method. This method didn't consider the cases when the open market time was part of the last market open segment. Also, when some potential market open wasn't accepted it just jumped into the next day but didn't consider the next segment in the day if possible.
- Fix bug in `AfterMarketOpen()` method. This method processed more than once the same date because `applicator` didn't consider the date returned by `GetNextMarketOpen()` and the time period it had advanced.
- Fix wrong regression tests statistics. Debugging the failing regression and unit tests It was found that they were directly relationed with the changes in `GetNextMarketOpen()` and `AfterMarketOpen()` and that the previous returned values were wrong.

The following regression tests were also failing for the changes in `AfterMarketOpen()` and GetNextMarketOpen():
- Algorithm.CSharp/FutureOptionBuySellCallIntradayRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionCallITMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionCallOTMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionPutITMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionPutOTMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionShortCallITMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionShortCallOTMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionShortPutITMExpiryRegressionAlgorithm.cs
- Algorithm.CSharp/FutureOptionShortPutOTMExpiryRegressionAlgorithm.cs
Although, the tests above were failing for the changes in `GetNextMarketOpen()` they were also failing because `applicator` function didn't take into account the dates `GetNextMarketOpen()` have already processed looking for the next market open so, if the given date was for example a Thursday at 00:00:00 as `GetNextMarketOpen()` is not inclusive, it jumped into the next day and returned the first segment(at 00:00:00) according to the MHDB. However, as we see for the last tests this date returned by `GetNextOpenMarket()` was wrong, so the date returned by `AfterMarketOpen()` was also wrong. Here we can see an example of the difference between the logs before and after the changes:
BEFORE
image
AFTER
image
As you can see for 06/01/2020 00:00:00 the date `AfterMarketOpen()` should return should be 06/01/2020 16:30:00, which is the next segment for that day according to MHDB
#### Related issue

Closes #6189 
#### Motivation and Context

With this change users will have the correct next open market date and they won't get repeated dates

This change does not require updates to documentation

- To test `GetNextMarketOpen()` was working correctly, there where made unit tests asserting cases in which the market opened at midnight or just opened one day of the week and it kept open until Friday. Also, there were made tests to assert this method returned the next segment of the day market schedule if the open market date was the same as the given date.
- To test `AfterMarketOpen()` was working correctly, there was made a test in which for a given set of dates, the method returned the open market dates for each of them, then those dates returned by the method were compared to the expected ones(computed manually) to check they were the same.
#### Types of changes

#### Checklist:


Common/Securities/SecurityExchangeHours.cs
```diff
@@ -271,6 +296,57 @@ public DateTime GetNextMarketOpen(DateTime localDateTime, bool extendedMarket)
...
+
+            // Check the current market open time is not part or continuation of the last market open segment
+            if (marketOpen - lastMarketOpen != new TimeSpan())
+            {
+                return false;
+            }
+
+            return true;
+        }
+
```
Martin-Molinero, line 349: Cool! Seems that the changes here could be handled by LocalMarketHours internally instead, maybe a missing functionality/method , I think it would allow for a cleaner usage of the class and potentially simplify the implementation it self too, dropping all linq usage and some duplication

Algorithm.CSharp/FutureMarketOpenConsolidatorRegressionAlgorithm.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            var es = AddSecurity(SecurityType.Future, "ES");
+
+            Consolidate<BaseData>(es.Symbol, time =>
+            {
+                var date = time;
+                if (time >= new DateTime(2013, 10, 21))
+                {
+                    date = Time;
+                }
```
Martin-Molinero, line 63: This is hiding a bug I think, see https://github.com/QuantConnect/Lean/blob/master/Common/Data/Consolidators/PeriodCountConsolidatorBase.cs#L385

---

On end of algorithm we should assert both queues are empty 👌

Algorithm.CSharp/FutureMarketOpenConsolidatorRegressionAlgorithm.cs
```diff
@@ -0,0 +1,147 @@
...
+        };
+        private Queue<DateTime> _expectedOpensQueue = new Queue<DateTime>(_expectedOpens);
+        private Queue<DateTime> _expectedClosesQueue = new Queue<DateTime>(_expectedCloses);
+
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 06);
+            SetEndDate(2013, 10, 14);
+
+            var es = AddSecurity(SecurityType.Future, "ES");
```
Martin-Molinero, line 55: > AddSecurity(SecurityType.Future

Should use `AddFuture` instead

Algorithm/QCAlgorithm.Trading.cs
```diff
@@ -937,7 +937,16 @@ private OrderResponse PreOrderChecksImpl(SubmitOrderRequest request)
 
             if (request.OrderType == OrderType.MarketOnClose)
             {
-                var nextMarketClose = security.Exchange.Hours.GetNextMarketClose(security.LocalTime, false);
+                DateTime nextMarketClose;
+
+                if (security.Exchange.Hours.IsMarketAlwaysOpen)
+                {
+                    nextMarketClose = security.LocalTime + new TimeSpan(24, 0, 0);
```
Martin-Molinero, line 944: Think we should throw here, it wont ever happen, a `MarketOnClose` order in an asset which has no market close, same comment for the `FillModel`. We should have the same for `MarketOnOpen` 👍

Common/Securities/SecurityExchangeHours.cs
```diff
@@ -226,9 +242,13 @@ public DateTime GetNextMarketOpen(DateTime localDateTime, bool extendedMarket)
         {
             var time = localDateTime;
             var oneWeekLater = localDateTime.Date.AddDays(15);
+            
+            MarketHoursSegment? lastDaySegment = null;
             do
             {
                 var marketHours = GetMarketHours(time.DayOfWeek);
+                var isEarlyCloseDay = false;
```
Martin-Molinero, line 250: > var isEarlyCloseDay = false;

Think we could remove this flag, setting `earlyCloseDateTime` to max?

Common/Securities/SecurityExchangeHours.cs
```diff
@@ -238,30 +258,61 @@ public DateTime GetNextMarketOpen(DateTime localDateTime, bool extendedMarket)
                         if (time < lateOpenDateTime)
                             return lateOpenDateTime;
 
+                        lastDaySegment = marketHours.Segments.Any() ? marketHours.Segments.Last() : null;
```
Martin-Molinero, line 261: > marketHours.Segments.Any() ? marketHours.Segments.Last() : null;

Can directly use `LastOrDefault`, same bellow

Martin-Molinero (commented):
Great improvement for Lean! Almost there

Martin-Molinero (approved):

Martin-Molinero:
Closing for now in favor of https://github.com/QuantConnect/Lean/pull/6397

PR #6267, opened 2022-03-27
Implement a log4net appender for QC

DavidThielen:
#### Description
1. Added log4net library
2. Added QuantConnectAppender

#### Motivation and Context
Better control over what is logged.

Yes, attached below.

Yes tested running on a sample program. I don't think there's any way to write unit tests for this as it's a log message.

#### Types of changes

#### Checklist:

[Using log4net on QuantConnect.docx](https://github.com/QuantConnect/Lean/files/8358597/Using.log4net.on.QuantConnect.docx)

### DIFF
--- a/Algorithm/QuantConnect.Algorithm.csproj
+++ b/Algorithm/QuantConnect.Algorithm.csproj
@@ -30,6 +30,7 @@
     <PackageLicenseFile>LICENSE</PackageLicenseFile>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="log4net" Version="2.0.14" />
     <PackageReference Include="QuantConnect.pythonnet" Version="2.0.11" />
     <PackageReference Include="MathNet.Numerics" Version="4.15.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3">
--- a/ToolBox/QuantConnect.ToolBox.csproj
+++ b/ToolBox/QuantConnect.ToolBox.csproj
@@ -40,6 +40,7 @@
     <Compile Include="..\Common\Properties\SharedAssemblyInfo.cs" Link="Properties\SharedAssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="log4net" Version="2.0.14" />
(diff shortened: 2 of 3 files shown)


DavidThielen:
Sorry, I named the branch before learning there's a naming convention.

Martin-Molinero:
Hey @DavidThielen! Welcome to Lean 🎉

Thanks for the contribution but I think it would make sense to push these changes into their own repository and nuget package, where anyone that wants to use this particular logging library, can directly reference it. And this repository can have example of how to use it in Lean etc.
If you'd like to contribute into Lean there are many reported issues to pick up from. Will close the PR and issue for now.

DavidThielen:
Hi @Martin-Molinero;

I may not be understanding something. If I want to use log4net when I run on QC's servers, is there a way I can just add log4net.dll for my use? If so, then a distinct repository makes sense. I assumed that it needed to be in master for log4net.dll to be installed/loaded.

If I can add it to my project, I'll test further using it myself, then create a distinct repository.

And when I do create that repository, do you have instructions about how I add that to LEAN? Or is this I put it in my GitHub and just let people know?

thanks - dave

Martin-Molinero:
Hey!

>  If I want to use log4net when I run on QC's servers, is there a way I can just add log4net.dll for my use?

Not at the moment sorry. I guess the underlying issue here might be missing features in the existing QC logging interface, is that the case? Which features does `log4net` have that we could add to the existing logging implementation to make it better?
We are moving Lean into a more modular/plugin approach. We try to avoid adding unrequired dependencies which will be carried around for all use cases of Lean, plus the added maintenance etc.

DavidThielen:
Hi;

I understand and agree with your desire to minimize unrequired dependencies. However, for most C# developers log4net is viewed as part of the base stack as its use is so ubiquitous. 

As to what you gain, two things. First it provides the ability in the config file to have very fine tuned control over what you log. So when your needs change on what you need to see in the log, it's trivial to change that.

Second, you can write the log to almost anything. So if you want to log to say Azure BLOB storage, again it's just some config settings.

[There's some other stuff](https://logging.apache.org/log4net/release/manual/introduction.html) others find useful like filters. But the biggie to me is the ability to quickly change what logging output I'll get.

thanks - dave

PR #6279, opened 2022-04-01
Refactor changes required for brokerages extraction

rjra2611:
#### Description

- Set price currency for open orders from `BrokerageSetupHandler.GetOpenOrders` ; Remove price currency being set from brokerages
- Use ctor to avoid using internal setters
- Remove dependence on brokerage instances from tests
  
#### Related Issue

N/A
#### Motivation and Context

Make necessary changes to start brokerage extractions from LEAN 

N/A

- Manually by checking price currency being set from brokerage setup handler
- run LEAN tests
#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/Tradier/TradierBrokerage.cs
+++ b/Brokerages/Tradier/TradierBrokerage.cs
@@ -962,7 +962,6 @@ namespace QuantConnect.Brokerages.Tradier
                 Log.Trace($"TradierBrokerage.TradierPlaceOrder(): order submitted successfully: {response.Order.Id}");
 
                 // send the submitted event
-                order.QCOrder.PriceCurrency = Currencies.USD;
                 OnOrderEvent(new OrderEvent(order.QCOrder, DateTime.UtcNow, OrderFee.Zero) { Status = OrderStatus.Submitted });
 
                 // mark this in our open orders before we submit so it's gauranteed to be there when we poll for updates
--- a/Brokerages/Tradier/TradierBrokerage.cs
+++ b/Brokerages/Tradier/TradierBrokerage.cs
@@ -998,11 +998,6 @@ namespace QuantConnect.Brokerages.Tradier
                 if (response != null && response.Errors != null && !response.Errors.Errors.IsNullOrEmpty())
                 {
                     message = "Order " + order.QCOrder.Id + ": " + string.Join(Environment.NewLine, response.Errors.Errors);
-                    if (string.IsNullOrEmpty(order.QCOrder.Tag))
-                    {
-                        var request = new UpdateOrderRequest(DateTime.UtcNow, order.QCOrder.Id, new UpdateOrderFields { Tag = message });
-                        order.QCOrder.ApplyUpdateOrderRequest(request);
-                    }
                 }
 
                 // send this error through to the console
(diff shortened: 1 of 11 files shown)


Brokerages/Tradier/TradierBrokerage.cs
```diff
@@ -1001,7 +1000,8 @@ private TradierOrderResponse TradierPlaceOrder(TradierPlaceOrderRequest order)
                     message = "Order " + order.QCOrder.Id + ": " + string.Join(Environment.NewLine, response.Errors.Errors);
                     if (string.IsNullOrEmpty(order.QCOrder.Tag))
                     {
-                        order.QCOrder.Tag = message;
+                        var request = new UpdateOrderRequest(DateTime.UtcNow, order.QCOrder.Id, new UpdateOrderFields { Tag = message });
+                        order.QCOrder.ApplyUpdateOrderRequest(request);
```
Martin-Molinero, line 1004: hm not sure this is a good idea, believe `UpdateOrderRequest` were meant to be are user driven order updates. How is this done by other brokerages?

Engine/Setup/BrokerageSetupHandler.cs
```diff
@@ -544,7 +546,11 @@ private void AddUnrequestedSecurity(IAlgorithm algorithm, Symbol symbol)
                     continue;
                 }
 
-                AddUnrequestedSecurity(algorithm, order.Symbol);
+                var security = AddUnrequestedSecurity(algorithm, order.Symbol);
+                if (security != null)
+                {
+                    order.PriceCurrency = security.SymbolProperties.QuoteCurrency;
```
Martin-Molinero, line 552: If the security was already added `AddUnrequestedSecurity` will return `null` and the price currency will not be set, but it should be.
I think it would be nice if this `if` wasn't required, `AddUnrequestedSecurity` could return the security always instead, maybe using `algorithm.Securities.TryGetValue()`?


Edit: actually even better: `order.PriceCurrency` is already being set by the BTH when creating a new order, so it would be nice to make him also set it for open orders, through `transactionHandler.AddOpenOrder`. Moving `AddUnrequestedSecurity` above the call to `AddOpenOrder` will allow the BTH to fetch the security internally like it does when `HandleSubmitOrderRequest`

Martin-Molinero (changes requested):
Cool 👍 ! leaving a couple of comments

Tests/Engine/Setup/BrokerageSetupHandlerTests.cs
```diff
@@ -79,6 +79,12 @@ public void CanGetOpenOrders()
 
             Assert.AreEqual(_transactionHandler.OrderTickets.Count, 4);
 
+            // Check Price Currency is not null
+            foreach (var order in _transactionHandler.Orders.Values)
+            {
+                Assert.IsNotEmpty(order.PriceCurrency);
```
Martin-Molinero, line 85: `Assert.IsNotEmpty((string)null);` will pass 😛 suggest using `Assert.IsFalse(string.IsNullOrEmpty((string)null));`

Martin-Molinero (approved):
Nice!

PR #6611, opened 2022-09-02
Get history for a specific bar type in Python

### LINKED ISSUE #6290: Can't Specify History Request Bar Type in Python
#### Expected Behavior

We can specify the bar type (`QuoteBar`/`TradeBar`) we want the `History` method to return in both C# and Python.

#### Actual Behavior

We can specify the bar type (`QuoteBar`/`TradeBar`) we want the `History` method to return in C#, but not in Python.

#### Potential Solution

N/A

#### Reproducing the Problem

If we run
```
qb = QuantBook()
start = datetime(2022, 4, 12)
end = datetime(2022, 4, 14)
ibm = qb.AddEquity("IBM", Resolution.Minute).Symbol
print(qb.History(ibm, start, end).empty)
print(qb.History(TradeBar, ibm, start, end).empty)
print(qb.History(QuoteBar, ibm, start, end).empty)
```
We get
> False
> True
> True

#### System Information

QC Cloud

#### Checklist

jhonabreul:
#### Description

Allowing to get history for a specific bar type from Python. The `PandasConverter.GetDataFrame()` method was not taking quote bars into account when the specified type was `QuoteBar` since `Slice.Keys` is empty in this case. `PandasConverter.GetDataFrame()` has now an optional parameter to pass in the requested data type. Omiting this parameter results in the previous behavior of going through the whole slice.

#### Related Issue

Close #6290 

#### Motivation and Context

N/A

N/A

Unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -930,7 +930,7 @@ namespace QuantConnect.Algorithm
             var requestedType = type.CreateType();
             var requests = CreateDateRangeHistoryRequests(symbols, requestedType, start, end, resolution, fillForward, extendedMarket,
                 dataMappingMode, dataNormalizationMode, contractDepthOffset);
-            return PandasConverter.GetDataFrame(History(requests.Where(x => x != null)));
+            return PandasConverter.GetDataFrame(History(requests.Where(x => x != null)), requestedType);
         }
 
         /// <summary>
@@ -955,7 +955,7 @@ namespace QuantConnect.Algorithm
             var requestedType = type.CreateType();
             var requests = CreateBarCountHistoryRequests(symbols, requestedType, periods, resolution);
 
-            return PandasConverter.GetDataFrame(History(requests.Where(x => x != null)).Memoize());
+            return PandasConverter.GetDataFrame(History(requests.Where(x => x != null)).Memoize(), requestedType);
         }
 
         /// <summary>
@@ -993,7 +993,7 @@ namespace QuantConnect.Algorithm
                 throw new ArgumentException("The specified security is not of the requested type. Symbol: " + symbol.ToString() + " Requested Type: " + requestedType.Name + " Actual Type: " + actualType);
             }
 
-            return PandasConverter.GetDataFrame(History(request).Memoize());
+            return PandasConverter.GetDataFrame(History(request).Memoize(), requestedType);
         }
 
         /// <summary>
--- a/Algorithm/QCAlgorithm.Python.cs
+++ b/Algorithm/QCAlgorithm.Python.cs
@@ -987,7 +987,7 @@ namespace QuantConnect.Algorithm
         {
             var requestedType = type.CreateType();
             var requests = CreateDateRangeHistoryRequests(new [] {  symbol }, requestedType, start, end, resolution);
-            if (requests == null || !requests.Any())
+            if (requests.IsNullOrEmpty())
             {
                 throw new ArgumentException($"No history data could be fetched. " +
                     $"This could be due to the specified security not being of the requested type. Symbol: {symbol} Requested Type: {requestedType.Name}");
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.Python.cs
```diff
@@ -986,14 +986,14 @@ public PyObject History(PyObject type, PyObject tickers, TimeSpan span, Resoluti
         public PyObject History(PyObject type, Symbol symbol, DateTime start, DateTime end, Resolution? resolution = null)
         {
             var requestedType = type.CreateType();
-            var request = CreateDateRangeHistoryRequests(new [] {  symbol }, requestedType, start, end, resolution).FirstOrDefault();
-            if (request == null)
+            var requests = CreateDateRangeHistoryRequests(new [] {  symbol }, requestedType, start, end, resolution);
+            if (requests == null || !requests.Any())
```
Martin-Molinero, line 990: nit - `.IsNullOrEmpty()`

Common/Python/PandasConverter.cs
```diff
@@ -226,5 +213,74 @@ private PyObject MakeIndicatorDataFrame(PyDict pyDict)
...
+        }
+
+        /// <summary>
+        /// Adds each slice data to the pandas data dictionary
+        /// </summary>
+        private void AddSliceDataToDict(Slice slice, IDictionary<Symbol, PandasData> sliceDataDict, ref int maxLevels)
+        {
+            foreach (var key in slice.Keys)
+            {
+                var baseData = slice[key];
```
Martin-Molinero, line 240: Instead of iterating the keys and later accessing the value, can directly iterate the slice to get the kvp pair? Same in the method bellow

Common/Python/PandasConverter.cs
```diff
@@ -226,5 +213,74 @@ private PyObject MakeIndicatorDataFrame(PyDict pyDict)
...
+                {
+                    value.Add(baseData);
+                }
+                else
+                {
+                    var ticks = slice.Ticks.ContainsKey(key) ? slice.Ticks[key] : null;
+                    var tradeBars = slice.Bars.ContainsKey(key) ? slice.Bars[key] : null;
+                    var quoteBars = slice.QuoteBars.ContainsKey(key) ? slice.QuoteBars[key] : null;
+                    value.Add(ticks, tradeBars, quoteBars);
+
```
Martin-Molinero, line 253: nit - unrequired space

Common/Python/PandasConverter.cs
```diff
@@ -226,5 +213,74 @@ private PyObject MakeIndicatorDataFrame(PyDict pyDict)
...
+        private void AddSliceDataTypeDataToDict(Slice slice, Type dataType, IDictionary<Symbol, PandasData> sliceDataDict, ref int maxLevels)
+        {
+            var isTick = dataType == typeof(Tick) || dataType == typeof(OpenInterest);
+            // Access ticks directly since slice.Get(typeof(Tick)) and slice.Get(typeof(OpenInterest)) will return only the last tick
+            var sliceData = isTick ? slice.Ticks : slice.Get(dataType);
+
+            foreach (var key in sliceData.Keys)
+            {
+                var baseData = sliceData[key];
+                var value = GetPandasDataValue(sliceDataDict, key, baseData, ref maxLevels);
```
Martin-Molinero, line 270: Believe that because baseData is dynamic value is being returned as dynamic, suggest replacing `var` here by the `PandasData` type we know it is, to avoid late binding when possible, same for method above. See https://stackoverflow.com/questions/7829518/dynamic-parameter-causes-compiler-to-think-method-return-is-dynamic for example

Martin-Molinero (commented):
Sweet! Minor requests

Common/Python/PandasConverter.cs
jhonabreul: We cannot iterate using the kvp, because it will enumerate on each base data instead of each symbol. For instance, if the slice has 3 ticks for a given symbol, enumerating the slice will result on three iterations where the symbol will be the same but the value will be each tick

Martin-Molinero (approved):

PR #6617, opened 2022-09-08
Handle non-unique multi-index error in history dataframe creation

### LINKED ISSUE #4297: Non-Unique Multi-Index Error in PandasConverter
#### Expected Behavior
Be able to make a history request of tick-resolution data in Python.

#### Actual Behavior
The following exception is thrown:
```
Exception: cannot handle a non-unique multi-index!
```
This happens since a `Slice.Ticks` object can have more than one `Tick` with the same timestamp because of the resolution of the timestamp.

#### Potential Solution
If a series of ticks have the same DateTime, only include the last tick. 

#### Reproducing the Problem
```python
aapl = qb.AddEquity("AAPL", Resolution.Tick)
start_time = datetime(2020,3,20)
end_time = datetime(2020,3,21)
data = qb.History(['AAPL'], start_time, end_time, Resolution.Tick)
```
or
```python
from QuantConnect.Python import *

class PandasConverterAlgorithm(QCAlgorithm):
    def Initialize(self):
        self.SetStartDate(2019, 10, 1)
        self.AddForex("EURUSD", Resolution.Minute, Market.Oanda)
        self.SetWarmup(10)
        self.window = RollingWindow[Slice](10)

    def OnData(self, slice):
        self.window.Add(slice)
        if self.window.Count == 10:
            df = self.PandasConverter.GetDataFrame(self.window)
```
#### Checklist

jhonabreul:
#### Description

Requesting tick history for securities with both trade and quote subscriptions, the result contains both trade and quote ticks. Converting this slices into a dataframe was resulting in a non-unique multi-index error. This was caused when there were quote and trade ticks with same timestamps, the series would have diferent indexes depending on the tick type (tick quotes have ask/bid size/price series, while trade ticks don't) which resulted in Pandas trying to reindex when creating the dataframe from the series.

Now we treat all ticks as if they had all the columns, regardless of the tick type, adding null (NaN) to the columns that don't apply (ask/bid size/price for trade ticks, for instance). If only trades are passed, some columns like ask/bid size/price will end up having only NaN values, making that columns to be ignored when creating the dataframe, as expected.

#### Related Issue

Closes #4297 

#### Motivation and Context

N/A

N/A

Unit tests and regression algorithm

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Python/PandasDataFrameFromMultipleTickTypeTickHistoryRegressionAlgorithm.py
+++ b/Algorithm.Python/PandasDataFrameFromMultipleTickTypeTickHistoryRegressionAlgorithm.py
@@ -1,0 +1,44 @@
+# QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+# Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


Martin-Molinero (approved):
🚀

PR #6725, opened 2022-11-03
Data monitor

jhonabreul:
#### Description

New `DataMonitor` class to keep track of data requests statistics:
- Total and successful/failed requests
- Universe data requests
- Request rates
- Report generation with summary of these stats.

#### Related Issue

N/A

#### Motivation and Context

N/A

N/A

Unit tests, backtests.

#### Types of changes

#### Checklist:


Common/DataMonitor.cs
```diff
@@ -0,0 +1,265 @@
...
+        private Thread _requestRateCalculationThread;
+        private CancellationTokenSource _cancellationTokenSource;
+
+        private string _succeededDataRequestsFileName;
+        private string _failedDataRequestsFileName;
+
+        /// <summary>
+        /// Directory location to store results
+        /// </summary>
+        public string ResultsDestinationFolder { get; }
```
Martin-Molinero, line 56: Could be a private field? Seems we don't need a public property

Common/DataMonitor.cs
```diff
@@ -0,0 +1,265 @@
...
+        /// <param name="report">The data monitor report to be stored<param>
+        private void StoreDataMonitorReport(DataMonitorReport report)
+        {
+            if (report == null)
+            {
+                return;
+            }
+
+            var path = GetFilePath("data-monitor-report.json");
+            var data = JsonConvert.SerializeObject(report, Formatting.None  );
```
Martin-Molinero, line 238: `Formatting.None  );` nit spaces

`GetFilePath` is adding the timestamp after the file extension, ideally it's before it

Common/DataMonitor.cs
```diff
@@ -0,0 +1,265 @@
...
+        private void Initialize()
+        {
+            if (_requestRateCalculationThread != null)
+            {
+                return;
+            }
+            
+            _cancellationTokenSource = new CancellationTokenSource();
+
+            _requestRateCalculationThread = new Thread(() =>
```
Martin-Molinero, line 172: Cool! we should take a lock here to make sure we start a single one, after the initial null check, there could be multiple threads fetching files at the same time 👍

Martin-Molinero (commented):
Nice! Leaving couple of minor requests 👍

Common/DataMonitor.cs
jhonabreul: My  bad with the timestamp, it has to be before the tghe extension 🎯

Martin-Molinero (approved):

PR #6839, opened 2023-01-05
Feature add algorithm state information in result packets

rjra2611:
#### Description

- Adds algorithm state information in the result packets of backtesting and live deployments
#### Related Issue

N/A
#### Motivation and Context

Helps other products with better information regarding the state of the algorithm

N/A

Tested by creating errors in the below cases for both backtesting and live deployments with the following language:
- Python
  - Initialization
  - OnData
  - Syntax error 
  - algorithm not available
- Csharp
  - Initialization
  - OnData
  - Syntax error (doesn't compile)
  - algorithm not available (doesn't compile)
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Result.cs
+++ b/Common/Result.cs
@@ -72,7 +72,7 @@ namespace QuantConnect
         public IDictionary<string, string> RuntimeStatistics;
 
         /// <summary>
-        /// State of the reuslt packet.
+        /// State of the result packet.
         /// </summary>
         [JsonProperty(PropertyName = "State", NullValueHandling = NullValueHandling.Ignore)]
         public IDictionary<string, string> State;
--- a/Common/Result.cs
+++ b/Common/Result.cs
@@ -72,6 +72,12 @@ namespace QuantConnect
         public IDictionary<string, string> RuntimeStatistics;
 
         /// <summary>
+        /// State of the reuslt packet.
+        /// </summary>
+        [JsonProperty(PropertyName = "State", NullValueHandling = NullValueHandling.Ignore)]
+        public IDictionary<string, string> State;
+
+        /// <summary>
         /// Server status information, including CPU/RAM usage, ect...
         /// </summary>
         [JsonProperty(PropertyName = "ServerStatistics", NullValueHandling = NullValueHandling.Ignore)]
--- a/Engine/Engine.cs
+++ b/Engine/Engine.cs
@@ -236,6 +236,7 @@ namespace QuantConnect.Lean.Engine
                         initializeComplete = false;
                         //Get all the error messages: internal in algorithm and external in setup handler.
                         var errorMessage = string.Join(",", algorithm.ErrorMessages);
+                        string stackTrace = "";
                         errorMessage += string.Join(",", AlgorithmHandlers.Setup.Errors.Select(e =>
                         {
                             var message = e.Message;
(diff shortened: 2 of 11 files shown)


Common/Result.cs
```diff
@@ -71,6 +71,12 @@ public class Result
         [JsonProperty(PropertyName = "RuntimeStatistics", NullValueHandling = NullValueHandling.Ignore)]
         public IDictionary<string, string> RuntimeStatistics;
 
+        /// <summary>
+        /// State of the reuslt packet.
```
Martin-Molinero, line 75: typo

Engine/Engine.cs
```diff
@@ -541,4 +545,23 @@ private static MarketHoursDatabase StaticInitializations()
         }
 
     } // End Algorithm Node Core Thread
+
+    public class InitializeException : Exception
+    {
+        private readonly string stackTrace;
```
Martin-Molinero, line 551: `stackTrace` not following code style, should be `_stackTrace`. Can make this class private 👌 . Should add some simple xml documentation about this class

Engine/Results/BaseResultsHandler.cs
```diff
@@ -142,6 +142,11 @@ public abstract class BaseResultsHandler
         /// </summary>
         protected Dictionary<string, string> RuntimeStatistics { get; }
 
+        /// <summary>
+        /// State of the result packet
+        /// </summary>
+        protected Dictionary<string, string> State { get; }
```
Martin-Molinero, line 148: This is not used?

Engine/Results/BaseResultsHandler.cs
```diff
@@ -667,6 +672,23 @@ protected virtual void SampleCapacity(DateTime time)
...
+        /// </summary>
+        protected Dictionary<string, string> GetAlgorithmState(string endTime = "")
+        {
+            var state = new Dictionary<string, string>
+            {
+                ["Status"] = Algorithm.Status.ToStringInvariant(),
+                ["StartTime"] = StartTime.ToStringInvariant(),
+                ["Hostname"] = _hostName,
+                ["EndTime"] = endTime,
+                ["RuntimeError"] = Algorithm?.RunTimeError?.Message ?? "",
```
Martin-Molinero, line 686: Here it's doing a null check for the algorithm `Algorithm?` but in the other cases not, should probably null check it all of them?
Why the inner exception bellow?

Engine/Engine.cs
```diff
@@ -236,18 +236,22 @@ public void Run(AlgorithmNodePacket job, AlgorithmManager manager, string assemb
...
                                 var err = interpreter.Interpret(e.InnerException);
-                                message += interpreter.GetExceptionMessageHeader(err);
+                                var stackMessage = interpreter.GetExceptionMessageHeader(err);
+                                message += stackMessage;
+                                stackTrace += stackMessage;
                             }
                             return message;
                         }));
                         Log.Error("Engine.Run(): " + errorMessage);
+                        algorithm.SetRuntimeError(new InitializeException(errorMessage, stackTrace), "Result Handler");
```
Martin-Molinero, line 254: - `SetRuntimeError` is also logging an error, so we can probably remove `Log.Error` above to avoid the duplication
- `"Result Handler"` context isn't accurate? Should be more like Algorithm Initialization

Queues/JobQueue.cs
```diff
@@ -213,7 +214,8 @@ public AlgorithmNodePacket NextJob(out string location)
                 Language = Language,
                 Parameters = parameters,
                 Controls = controls,
-                PythonVirtualEnvironment = Config.Get("python-venv")
+                PythonVirtualEnvironment = Config.Get("python-venv"),
+                HostName = Environment.MachineName
```
Martin-Molinero, line 218: We don't really need these changes I think? given the baseResulthandler is already getting MachineName if hostname isn't set?

Martin-Molinero (changes requested):
Cool! Leaving a few comments 👍

Martin-Molinero (approved):
Nice! ✨

PR #6852, opened 2023-01-11
TensorFlow Placeholder Fix

ChadBett:
Fixed this error from happening when running the TensorFlow example:

tf.placeholder() is not compatible with eager execution.
  at placeholder
    raise RuntimeError("tf.placeholder() is not compatible with "
 in array_ops.py: line 3341
  at NetTrain
    xs = tf.compat.v1.placeholder(tf.float32 in /QuantConnect/backtesting/./cache/algorithm/project/main.py: line 57


Martin-Molinero (approved):
Thanks @ChadBett!

PR #6804, opened 2022-12-15
Constant volume bar

LouisSzeto:
#### Description
The creation of `VolumeRenkoConsolidator` allow consolidation by constant volume bar.

#### Related Issue
closes #3754

#### Motivation and Context
`ClassicRenkoConsolidator`/`RenkoConsolidator` cannot create constant volume bar.

#### Requires Documentation Change
https://www.quantconnect.com/docs/v2/writing-algorithms/consolidating-data/consolidator-types

#### How Has This Been Tested?
Added unit test, following the style of `ClassicRenkoConsolidator`.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [X] My code follows the code style of this project.
- [X] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [X] I have added tests to cover my changes. 
- [X] All new and existing tests passed.
- [X] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #6854, opened 2023-01-12
Volume Renko Consolidator & Bar

### LINKED ISSUE #3754: Value Consolidator
#### Expected Behavior

Value Consolidator, for Volume and Dollar Bars as described in the book Advances in Financial Machine Learning, and as implemented by mlfinlab:

https://www.quantconnect.com/forum/discussion/5853/mlfinlab-python-package-released-advances-in-financial-machine-learning/ 

#### Actual Behavior

Discussion forum replies suggest RenkoConsolidator. However, RenkoBars do not accumulate value of previous bars.

Wanted:

```
bars = []
data = [1, 2, 3, 4, 5]
total = 0
bar = { 'open': 0, 'close': 0, 'high': 0, 'low': 0}
for d in data:
    total += d
    if total < threshold:
        if bar['open'] == 0:
            bar['open'] = d
        ...
    else:
        bar['close'] == d
        bars.append(bar)
        total = 0
        bar =  { 'open': 0, 'close': 0, 'high': 0, 'low': 0}
```

#### Checklist

LouisSzeto:
#### Description
The creation of `VolumeRenkoConsolidator` allow consolidation by constant volume bar.

#### Related Issue
closes #3754
continuation of #6804

#### Motivation and Context
Current consolidators cannot create constant volume bar.

Additional section needed in https://www.quantconnect.com/docs/v2/writing-algorithms/consolidating-data/consolidator-types

Added unit tests, following the style of `ClassicRenkoConsolidator`, and regression algorithms, following the style of `ClassicRenkoConsolidator`.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/VolumeRenkoConsolidatorAlgorithm.cs
+++ b/Algorithm.CSharp/VolumeRenkoConsolidatorAlgorithm.cs
@@ -38,7 +38,7 @@ namespace QuantConnect.Algorithm.CSharp
         public override void Initialize()
         {
             SetStartDate(2013, 10, 7);
-            SetEndDate(2014, 10, 11);
+            SetEndDate(2013, 10, 11);
             SetCash(100000);
             
             _spy = AddEquity("SPY", Resolution.Minute).Symbol;
@@ -109,7 +109,7 @@ namespace QuantConnect.Algorithm.CSharp
         /// <summary>
         /// Data Points count of all timeslices of algorithm
         /// </summary>
-        public long DataPoints => 895162;
+        public long DataPoints => 698706;
 
         /// <summary>
         /// Data Points count of the algorithm history
@@ -124,31 +124,31 @@ namespace QuantConnect.Algorithm.CSharp
             {"Total Trades", "227"},
             {"Average Win", "0.25%"},
             {"Average Loss", "-0.05%"},
-            {"Compounding Annual Return", "-0.833%"},
-            {"Drawdown", "2.800%"},
+            {"Compounding Annual Return", "-48.349%"},
+            {"Drawdown", "3.000%"},
             {"Expectancy", "-0.191"},
             {"Net Profit", "-0.841%"},
-            {"Sharpe Ratio", "-0.129"},
-            {"Probabilistic Sharpe Ratio", "9.022%"},
+            {"Sharpe Ratio", "-0.956"},
+            {"Probabilistic Sharpe Ratio", "41.341%"},
             {"Loss Rate", "87%"},
             {"Win Rate", "13%"},
             {"Profit-Loss Ratio", "5.15"},
-            {"Alpha", "-0.008"},
-            {"Beta", "0.05"},
-            {"Annual Standard Deviation", "0.021"},
-            {"Annual Variance", "0"},
-            {"Information Ratio", "-1.311"},
-            {"Tracking Error", "0.087"},
-            {"Treynor Ratio", "-0.055"},
+            {"Alpha", "-2.225"},
+            {"Beta", "1.009"},
+            {"Annual Standard Deviation", "0.234"},
+            {"Annual Variance", "0.055"},
+            {"Information Ratio", "-33.229"},
+            {"Tracking Error", "0.066"},
+            {"Treynor Ratio", "-0.222"},
             {"Total Fees", "$767.26"},
-            {"Estimated Strategy Capacity", "$4300000.00"},
+            {"Estimated Strategy Capacity", "$4000000.00"},
             {"Lowest Capacity Asset", "SPY R735QTJ8XC9X"},
-            {"Fitness Score", "0.412"},
+            {"Fitness Score", "0.025"},
             {"Kelly Criterion Estimate", "0"},
             {"Kelly Criterion Probability Value", "0"},
-            {"Sortino Ratio", "-0.111"},
-            {"Return Over Maximum Drawdown", "-0.3"},
-            {"Portfolio Turnover", "0.881"},
+            {"Sortino Ratio", "-7.077"},
+            {"Return Over Maximum Drawdown", "-19.135"},
+            {"Portfolio Turnover", "47.249"},
             {"Total Insights Generated", "0"},
             {"Total Insights Closed", "0"},
             {"Total Insights Analysis Completed", "0"},
--- a/Algorithm.CSharp/VolumeRenkoConsolidatorAlgorithm.cs
+++ b/Algorithm.CSharp/VolumeRenkoConsolidatorAlgorithm.cs
@@ -1,0 +1,159 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
+using QuantConnect.Indicators;
+using QuantConnect.Interfaces;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Demostrates the use of <see cref="VolumeRenkoConsolidator"/> for creating constant volume bar
+    /// </summary>
+    /// <meta name="tag" content="renko" />
+    /// <meta name="tag" content="using data" />
+    /// <meta name="tag" content="consolidating data" />
+    public class VolumeRenkoConsolidatorAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        private Symbol _spy, _ibm;
+        private VolumeRenkoConsolidator _tradebarVolumeConsolidator, _tickVolumeConsolidator;
+        private SimpleMovingAverage _sma = new SimpleMovingAverage(10);
+        
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 7);
+            SetEndDate(2014, 10, 11);
+            SetCash(100000);
+            
+            _spy = AddEquity("SPY", Resolution.Minute).Symbol;
+            _tradebarVolumeConsolidator = new VolumeRenkoConsolidator(1000000);
+            _tradebarVolumeConsolidator.DataConsolidated += (sender, bar) => {
+                _sma.Update(bar.EndTime, bar.Value);
+                Log($"SPY {bar.Time} to {bar.EndTime} :: O:{bar.Open} H:{bar.High} L:{bar.Low} C:{bar.Close} V:{bar.Volume}");
+            };
+            
+            _ibm = AddEquity("IBM", Resolution.Tick).Symbol;
+            _tickVolumeConsolidator = new VolumeRenkoConsolidator(1000000);
+            _tickVolumeConsolidator.DataConsolidated += (sender, bar) => {
+                Log($"IBM {bar.Time} to {bar.EndTime} :: O:{bar.Open} H:{bar.High} L:{bar.Low} C:{bar.Close} V:{bar.Volume}");
+            };
+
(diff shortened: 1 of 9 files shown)


Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,143 @@
...
+                close = rate;
+            }
+            else
+            {
+                throw new ArgumentException("VolumeRenkoConsolidator() must be used with TradeBar or Tick data.");
+            }
+
+            if (_firstTick)
+            {
+                _firstTick = false;
```
Martin-Molinero, line 109: - Can avoid the need for this flag just by checking if `_currentBar` is null or not?
- What if the first bar already covers the expected volume? we should emit right away, seems we are waiting for the next data point
- Seems `_volumeLeftOver` can just be a local variable?

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,143 @@
...
+                _currentBar = new VolumeRenkoBar(data.Symbol, data.Time, data.EndTime, _barSize, open, high, low, close, volume);
+            }
+            else
+            {
+                _volumeLeftOver = _currentBar.Update(data.EndTime, high, low, close, volume);
+                while (_volumeLeftOver >= 0)
+                {
+                    OnDataConsolidated(_currentBar);
+                    _currentBar = new VolumeRenkoBar(data.Symbol, data.EndTime, data.EndTime, _barSize, close, high, low, close, 0);
```
Martin-Molinero, line 118: > new VolumeRenkoBar(data.Symbol, data.EndTime, data.EndTime

Sending `EndTime` twice 

> close, high, low, close, 0

Sending `close` twice

- Should create the new `VolumeRenkoBar` in a single location, not 2 lines to avoid these types of issues

Common/Data/Market/VolumeRenkoBar.cs
```diff
@@ -0,0 +1,143 @@
...
+*/
+
+using System;
+
+namespace QuantConnect.Data.Market
+{
+    /// <summary>
+    /// Represents a bar sectioned not by time, but by some amount of movement in volume
+    /// </summary>
+    public class VolumeRenkoBar : BaseData, IBaseDataBar
```
Martin-Molinero, line 23: Seems pretty similar to the existing `RenkoBar`, do we need both? can we make them inherit each other? or maybe we need some base abstract share class?

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,143 @@
...
+                if (((Tick)data).TickType != TickType.Trade)
+                {
+                    return;
+                }
+
+                volume = ((Tick)data).Quantity;
+                open = rate;
+                high = rate;
+                low = rate;
+                close = rate;
```
Martin-Molinero, line 100: > close = rate;

Maybe can avoid the `rate` variable and directly call it close at the top?

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <summary>
+        /// Event invocator for the DataConsolidated event. This should be invoked
+        /// by derived classes when they have consolidated a new piece of data.
+        /// </summary>
+        /// <param name="consolidated">The newly consolidated data</param>
+        protected void OnDataConsolidated(VolumeRenkoBar consolidated)
+        {
+            DataConsolidated?.Invoke(this, consolidated);
+            Consolidated = consolidated;
+        }
```
Martin-Molinero, line 141: Instead of having the method can directly call base `OnDataConsolidated` ?

Common/Data/Market/VolumeRenkoBar.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <param name="time">The current data end time</param>
+        /// <param name="high">The current data high value</param>
+        /// <param name="low">The current data low value</param>
+        /// <param name="close">The current data close value</param>
+        /// <param name="volume">The current data volume</param>
+        /// <returns>The excess volume that the current bar cannot absorb</returns>
+        public decimal Update(DateTime time, decimal high, decimal low, decimal close, decimal volume)
+        {
+            if (Type == RenkoType.Wicked)
+                throw new InvalidOperationException("A \"Wicked\" RenkoBar cannot be updated!");
```
Martin-Molinero, line 119: This would never happen by definition? This isn't an argument `Type = RenkoType.Classic;` it shouldn't even be part of the class? does a Wicked Volumen Renko Bar exist?

Common/Data/Market/VolumeRenkoBar.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            // can't update a closed renko bar
+            if (IsClosed) return 0m;
+            if (Start == DateTime.MinValue) Start = time;
```
Martin-Molinero, line 123: > if (Start == DateTime.MinValue) Start = time;

This can't happen either? let's remove it if so 👍

Algorithm.CSharp/VolumeRenkoConsolidatorAlgorithm.cs
```diff
@@ -0,0 +1,159 @@
...
+    public class VolumeRenkoConsolidatorAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        private Symbol _spy, _ibm;
+        private VolumeRenkoConsolidator _tradebarVolumeConsolidator, _tickVolumeConsolidator;
+        private SimpleMovingAverage _sma = new SimpleMovingAverage(10);
+        
+        public override void Initialize()
+        {
+            SetStartDate(2013, 10, 7);
+            SetEndDate(2014, 10, 11);
```
Martin-Molinero, line 40: The repo doesn't have such a big data range, think we can reduce it to the minimum we need?

Tests/Common/Data/VolumeRenkoConsolidatorTests.cs
```diff
@@ -0,0 +1,240 @@
...
+
+            consolidator.Update(new Tick(reference.AddHours(3), Symbol.Empty, String.Empty, String.Empty, 2m, 2m));
+            Assert.IsNotNull(bar);
+
+            Assert.AreEqual(1m, bar.Open);
+            Assert.AreEqual(3m, bar.High);
+            Assert.AreEqual(1m, bar.Low);
+            Assert.AreEqual(2m, bar.Close);
+            Assert.AreEqual(reference.AddHours(3), bar.EndTime);
+            Assert.IsTrue(bar.IsClosed);
```
Martin-Molinero, line 63: To be safe should assert all of the properties startTime/symbol/etc/the volume! same for the other tests 👍

Martin-Molinero (changes requested):
Cool! Leaving a few requests 👍

Algorithm.CSharp/VolumeRenkoConsolidatorAlgorithm.cs
```diff
@@ -0,0 +1,159 @@
...
+            _tradebarVolumeConsolidator = new VolumeRenkoConsolidator(1000000);
+            _tradebarVolumeConsolidator.DataConsolidated += (sender, bar) => {
+                _sma.Update(bar.EndTime, bar.Value);
+                Log($"SPY {bar.Time} to {bar.EndTime} :: O:{bar.Open} H:{bar.High} L:{bar.Low} C:{bar.Close} V:{bar.Volume}");
+            };
+            
+            _ibm = AddEquity("IBM", Resolution.Tick).Symbol;
+            _tickVolumeConsolidator = new VolumeRenkoConsolidator(1000000);
+            _tickVolumeConsolidator.DataConsolidated += (sender, bar) => {
+                Log($"IBM {bar.Time} to {bar.EndTime} :: O:{bar.Open} H:{bar.High} L:{bar.Low} C:{bar.Close} V:{bar.Volume}");
```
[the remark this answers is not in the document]
LouisSzeto: I'll leave the `Consolidate<VolumeRenkoBar>` first and create a issue instead. 👌

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
LouisSzeto: I'll inherit the base virtual method directly

Common/Data/Market/VolumeRenkoBar.cs
LouisSzeto: Sure. I think a base class works better, or it would be weird intuitively if `RenkoBar` is inherited from `VolumeRenkoBar`.

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
LouisSzeto: The input type is different?

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <summary>
+        /// Gets <see cref="VolumeRenkoBar"/> which is the type emitted in the <see cref="IDataConsolidator.DataConsolidated"/> event.
+        /// </summary>
+        public override Type OutputType => typeof(VolumeRenkoBar);
+
+        /// <summary>
+        /// Gets the most recently consolidated piece of data. This will be null if this consolidator
+        /// has not produced any data yet.
+        /// </summary>
+        public IBaseData Consolidated
```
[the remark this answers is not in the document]
LouisSzeto: `DataConsolidated` will still be included to make the consolidated output bar a `VolumeRenkoBar` object, but this `Consolidated` will be deleted

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
LouisSzeto: I create a `Rollover` method in the `VolumeRenkoBar` class to use information from the current bar to create a new bar (like a `Clone`, but modified)

Common/Data/Consolidators/DataConsolidator.cs
```diff
@@ -58,7 +58,7 @@ public void Update(IBaseData data)
         /// </summary>
         public IBaseData Consolidated
         {
-            get; private set;
+            get; protected set;
```
Martin-Molinero, line 61: This change can be avoided, take a look at how `PeriodCountConsolidatorBase` is doing it 👍

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        /// <summary>
+        /// Gets <see cref="VolumeRenkoBar"/> which is the type emitted in the <see cref="IDataConsolidator.DataConsolidated"/> event.
+        /// </summary>
+        public override Type OutputType => typeof(VolumeRenkoBar);
+
+        /// <summary>
+        /// Event handler that fires when a new piece of data is produced
+        /// </summary>
+        public event EventHandler<VolumeRenkoBar> DataConsolidated;
```
Martin-Molinero, line 43: Sadly need to add `new`, see `PeriodCountConsolidatorBase`. Created related issue https://github.com/QuantConnect/Lean/issues/6867

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            if (_currentBar == null)
+            {
+                _currentBar = new VolumeRenkoBar(data.Symbol, data.Time, data.EndTime, _barSize, open, high, low, close, 0);
+            }
+            var volumeLeftOver = _currentBar.Update(data.EndTime, high, low, close, volume);
+            while (volumeLeftOver >= 0)
+            {
+                OnDataConsolidated(_currentBar);
+                _currentBar = _currentBar.Rollover();
```
Martin-Molinero, line 101: Don't think rollover is accurate always, it will reuse the same high/low of the consolidated bar but those could of be older of previous bars, it should use the current data to create the new bar

Common/Data/Market/RenkoBar.cs
```diff
@@ -20,54 +20,10 @@ namespace QuantConnect.Data.Market
     /// <summary>
     /// Represents a bar sectioned not by time, but by some amount of movement in a value (for example, Closing price moving in $10 bar sizes)
     /// </summary>
-    public class RenkoBar : BaseData, IBaseDataBar
+    public class RenkoBar : BaseRenkoBar
     {
         /// <summary>
-        /// Gets the kind of the bar
-        /// </summary>
-        public RenkoType Type { get; private set; }
```
Martin-Molinero, line 28: Nice clean up! ✨

Common/Data/Consolidators/VolumeRenkoConsolidator.cs
LouisSzeto: yah, I will use `Close` for both attributes on rolling over

PR #6971, opened 2023-02-16
Custom data type history API usage in Python

### LINKED ISSUE #6927: Can't Make History Requests in Python for Custom/Alt Data Objects
#### Expected Behavior

In Python, like in C#, we can make a history request that returns data in our requested data type.
```
history = self.History[MyCustomDataType](self.symbol, 30, Resolution.Daily)
```

#### Actual Behavior

In C#, we have no issues.

In Python, we can request a DataFrame. If we request the data be in our a list of the data type, when we try to iterate through the result, the algorithm just times out.
```
history = self.History[MyCustomDataType](self.symbol, 30, Resolution.Daily)
# This line just causes a time-out:
[x for x in history]
```
> Algorithm.Initialize() Error: Execution Security Error: Operation timed out - 5 minutes max. Check for recursive loops. Stack Trace: Execution Security Error: Operation timed out - 5 minutes max. Check for recursive loops.

#### Potential Solution

N/A

#### Reproducing the Problem

Python: https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_b004df9c56245296dc6dfbf4897d1598.html

C#: https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_86cec936cde21ef37da9eee228301674.html

#### System Information

QC Cloud

#### Checklist

jhonabreul:
#### Description

Using the History API that returns lists instead of data frames with Python custom data types was having two issues: returning empty lists and timing out when enumerating the result in Python. 

These changes solve both issues by using the `Slice.Get(Type)` method instead of the generic `Slice.Get<T>()`, so that it can match the custom data in the slices to the actual Python class.

The timeout was being caused by a deadlock in the custom class readers data access, which is solved by wrapping Python custom data history requests results using `WrapPythonDataHistory`.

#### Related Issue

Closes #6927 

Unit and regression tests.

#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.History.cs
```diff
@@ -399,7 +401,7 @@ public IEnumerable<T> History<T>(Symbol symbol, DateTime start, DateTime end, Re
             where T : IBaseData
         {
             var requests = CreateDateRangeHistoryRequests(new[] { symbol }, typeof(T), start, end, resolution);
-            return GetDataTypedHistory<T>(symbol, requests);
+            return GetDataTypedHistory<T>(requests, symbol).Memoize();
```
Martin-Molinero, line 404: Maybe we can push `.Memoize();` into the `GetDataTypedHistory` implementations, so all these consumers don't need to worry about it?

Algorithm/QCAlgorithm.History.cs
```diff
@@ -624,34 +626,91 @@ public BaseData GetLastKnownPrice(Security security)
...
             }
 
             var slices = History(requests);
 
+            // If T is a custom data coming from Python (a class derived from PythonData), T will get here as PythonData
+            // and not the actual custom type. We take care of this especial case by using a dynamic version of GetDataTypedHistory that
+            // receives the Python type, and we get it from the history requests.
+            if (type == typeof(PythonData))
+            {
+                return GetPythonCustomDataTypeHistory(slices, requests.First().DataType, symbol).OfType<T>();
```
Martin-Molinero, line 650: - `requests.First()` ideally we should get around iterating twice over the generation of the history requests
- `requests.First().DataType` maybe we can pass the requests as a list to `GetPythonCustomDataTypeHistory` so internally it determines what to use? Similar comment as before, avoiding consumers needing to be smart about it

Algorithm/QCAlgorithm.History.cs
```diff
@@ -624,34 +626,91 @@ public BaseData GetLastKnownPrice(Security security)
...
+            where T : IBaseData
+        {
+            var slices = History(requests.Where(x => x != null));
+
+            if (typeof(T) == typeof(PythonData))
+            {
+                return GetPythonCustomDataTypeHistory(slices, requests.First().DataType).OfType<DataDictionary<T>>().Memoize();
+            }
+
+            return slices.Get<T>();
```
Martin-Molinero, line 682: - `History` is returning a `Memoize` already, we should probably use the private method which doesn't?
- This method has 2 returns, first using `Memoize`, second doesn't. Maybe can have a single return to avoid these issues?

Algorithm/QCAlgorithm.History.cs
```diff
@@ -624,34 +626,91 @@ public BaseData GetLastKnownPrice(Security security)
...
+        private IEnumerable<dynamic> GetPythonCustomDataTypeHistory(IEnumerable<Slice> slices, Type pythonType, Symbol symbol = null)
+        {
+            var history = slices.Select(x => x.Get(pythonType));
+
+            if (symbol == null)
+            {
+                return history;
+            }
+
+            return history.Where(x => x.ContainsKey(symbol)).Select(x => x[symbol]);
```
Martin-Molinero, line 706: Maybe some of these could go into the `SliceExtensions` along side the other Get implementations? Potentially there could be a single method which covers both use cases and we directly call that one?

Algorithm.Python/CustomDataTypeHistoryAlgorithm.py
```diff
@@ -0,0 +1,73 @@
...
+            raise Exception("History request returned no data")
+
+        self._assertHistoryData(history)
+
+        history2 = list(self.History[CustomDataType]([self.symbol], 48, Resolution.Hour))
+
+        if len(history2) != len(history):
+            raise Exception("History requests returned different data")
+
+        self._assertHistoryData([y.values()[0] for y in history2])
```
Martin-Molinero, line 37: Nice! ✨

Algorithm.CSharp/CustomDataTypeHistoryAlgorithm.cs
```diff
@@ -0,0 +1,170 @@
...
+            var history2 = History<CustomDataType>(new[] { _symbol }, 48, Resolution.Hour).ToList();
+
+            if (history2.Count != history.Count)
+            {
+                throw new Exception("History requests returned different data");
+            }
+
+        }
+
+        public class CustomDataType : DynamicData
```
Martin-Molinero, line 58: nit - could this class be private? to avoid exposing it to others if not required 👍

Martin-Molinero (commented):
Awesome 💪 
Leaving a couple of minor requests, also seems like a couple of unit test are failing

Martin-Molinero:
This bug probably affects python custom data research history requests too, would be nice to add a simple research notebook, like the added algorithm & make it part of the `ResearchRegressionTests` CI ✨

Martin-Molinero (approved):
Awesome! 🚀

PR #6977, opened 2023-02-17
Adds NullBuyingPowerModel

### LINKED ISSUE #3157: Implement NullMarginModel for Live Trading
#### Expected Behavior
Live trading should lean on the brokerage margin limitations and not impose them in LEAN.

#### Actual Behavior
LEAN models restrict brokerage margin.

#### Potential Solution
1) Improve option margin models
2) and Implement null margin model to disable restrictions for live trading.

#### Reproducing the Problem
Live trading straddles.

#### Checklist

AlexCatarino:
#### Description
The `NullBuyingPowerModel` considers that we have sufficient buying power for all orders.

Adds an example using a bull call spread since an equity buy and hold would not show the impact of this model in the position group buying power model.

#### Related Issue
Closes #3157

#### Motivation and Context
The buying power model prevents orders in live trading if it considers there is not sufficient buying power. If users want to submit the orders, they can use this model to bypass the validation. 

Yes. List this model on
https://www.quantconnect.com/docs/v2/writing-algorithms/reality-modeling/buying-power

Regression tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/NullBuyingPowerModel.cs
+++ b/Common/Securities/NullBuyingPowerModel.cs
@@ -21,6 +21,16 @@ namespace QuantConnect.Securities
     public class NullBuyingPowerModel : BuyingPowerModel
     {
         /// <summary>
+        /// Gets the margin currently allocated to the specified holding
+        /// </summary>
+        /// <param name="parameters">An object containing the security</param>
+        /// <returns>The maintenance margin required for the provided holdings quantity/cost/value</returns>
+        public override MaintenanceMargin GetMaintenanceMargin(MaintenanceMarginParameters parameters)
+        {
+            return new MaintenanceMargin(decimal.Zero);
+        }
+
+        /// <summary>
         /// Check if there is sufficient buying power to execute this order.
         /// </summary>
         /// <param name="parameters">An object containing the portfolio, the security and the order</param>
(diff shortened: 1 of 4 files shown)


Common/Securities/NullBuyingPowerModel.cs
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+namespace QuantConnect.Securities
+{
+    /// <summary>
+    /// Provides a buying power model considers that there is sufficient buying power for all orders
+    /// </summary>
+    public class NullBuyingPowerModel : BuyingPowerModel
```
Martin-Molinero, line 21: Would of expected it to use the https://github.com/QuantConnect/Lean/blob/master/Common/Securities/ConstantBuyingPowerModel.cs with a small value because it's not only about placing a new trading but the margin used to maintain it without getting a margin call?

AlexCatarino: Users could use the `MarginCallModel.Null` to handle the margin calls. Regardless, I  added the `NullBuyingPowerModel.GetMaintenanceMargin` returning a `MaintenanceMargin` of zero. It means the total margin used is always zero and margin calls will not be triggered.

This feature is inspired by the  `ConstantBuyingPowerModel`. I didn't inherit from that class, because `ConstantBuyingPowerModel.GetInitialMarginRequirement` gives us a very small initial margin that leads to a very large quantity if we use `SetHoldings` or `CalculateOrderQuantity`. If the algorithm is using `SetHoldings` the author would have to rewrite it.

Martin-Molinero (approved):
Nice!

PR #6999, opened 2023-02-22
Normalization of Minimum Variance Optimizer's output weight

### LINKED ISSUE #4120: MeanVarianceOptimizationPCM bugdet constraint
#### Expected Behavior

- `MeanVarianceOptimizationPortfolioConstructionModel` respects budget constraint
#### Actual Behavior
- Current budget constraint used by `MaximumSharpeRatioPortfolioOptimizer` and `MinimumVariancePortfolioOptimizer` is the sum of all portfolio weights should be equal to 1. But since weights can be negative too, it allows situations like the following: Asset A: -1 Asset B: 1 , Asset C: -1 , Asset D : 1 ... etc as long as they sum up to 1, with 100 assets it would mean trading 100x your TPV
- Throws NaN exception because cannot convert double.NaN to decimal. In this case, we should probably consider 0 weight instead:
```csharp
    public class NaNBug : QCAlgorithm {
		    SetStartDate(2020, 1, 1);

            AddUniverse(
            	coarse => coarse
		        	.Where(x => x.HasFundamentalData)
		        	.OrderByDescending(x => x.DollarVolume)
		        	.Take(100)
		        	.Select(x => x.Symbol),
		        fine => fine
		        	.OrderByDescending(x => x.ValuationRatios.PERatio)
		        	.Take(5)
		        	.Select(x => x.Symbol));

            SetAlpha(new HistoricalReturnsAlphaModel(14));
            SetPortfolioConstruction(new MeanVarianceOptimizationPortfolioConstructionModel());
        }
    }
}
```

#### Potential Solution

- Refactor `IPortfolioOptimizer` implementations to respect a realistic budget constraint
- Refactor `MeanVarianceOptimizationPCM` to enforce a realistic budget constraint on the output of the `IPortfolioOptimizer`, maybe scaling the result down proportionally (similar to https://github.com/QuantConnect/Lean/blob/master/Algorithm.Framework/Portfolio/InsightWeightingPortfolioConstructionModel.cs#L121)
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

LouisSzeto:
#### Description
Additional normalization with total absolute weight assigned by the optimizer allows Mean Variance PCM or other related PCMs to respect the overall budget constraint.

#### Related Issue
closes #4120 

#### Motivation and Context
Current implementation does not respect the realistic budget constraint and have margin error due to excessive negative weights

NA

Regression test

#### Types of changes

#### Checklist:


Algorithm.Framework/Portfolio/MinimumVariancePortfolioOptimizer.py
```diff
@@ -63,7 +63,7 @@ def Optimize(self, historicalReturns, expectedReturns = None, covariance = None)
                        constraints = constraints,                                 # Constraints definition
                        method='trust-constr')   # Optimization method:  trust-region algorithm for constrained optimization
 
-        return opt['x'] if opt['success'] else x0
+        return opt['x'] / np.sum(abs(opt['x'])) if opt['success'] else x0
```
AlexCatarino, line 66: This solution will break the budget constraint:  Σw = 1 will become Σw <= 1. Perhaps it's possible to add another constraint where Σ|w| <= 1, and the minimize function would find a solution. I am not sure how this is done in C# using this library, but for python we should be able to use something like:

```python
        constraints = [
            {'type': 'eq', 'fun': lambda weights: self.get_budget_constraint(weights)},
            {'type': 'eq', 'fun': lambda weights: self.get_absolute_budget_constraint(weights)},   # new constraint
            {'type': 'eq', 'fun': lambda weights: self.get_target_constraint(weights, expectedReturns)}]

    def get_absolute_budget_constraint(self, weights):
        '''Defines a budget constraint: the sum of the absolute weights equals unity'''
        return np.sum(np.abs(weights)) - 1
```

AlexCatarino (changes requested):
Thank you, @LouisSzeto!
I think we need to do some research to find a solution that doesn't break the budget constraint: Σw = 1. It might be simple in Python, but we need to dig for a solution with Accord.

Additionally, if you find any spelling errors in the classes, please address them.

AlexCatarino (approved):

PR #7005, opened 2023-02-23
Fixes EquityFillModel.FillModel

AlexCatarino:
#### Description
Use `Tick` with `TickType.Trade` or `TradeBar` information to fill limit orders.

- Adds Unit Tests:
The `LimitFill` method should not fill using `QuoteBar` or `Tick` with `TickType.Quote` type.
Adds missing test for tick data (`TickType.Trade`).

- Update Regression Test Expected Statistics:
The regression tests changed because of different fills.
The `ExtendedMarketTradingRegressionAlgorithm` has different number of trades because of an extra fill on the 4th order generated by TradeBar with a Low lower than than the QuoteBar.Ask Low:

> 20230222 13:56:24.251 TRACE:: Log: Time: 10/10/2013 12:01:00 OrderID: 4 EventID: 2 Symbol: SPY Status: Filled Quantity: 10 FillQuantity: 10 FillPrice: 143.8998 USD

> asset.Cache.GetData<QuoteBar>().ToString()
"SPY: Bid: O: 144.2457 Bid: H: 144.2629 Bid: **L: 144.2457** Bid: C: 144.2629 Ask: O: 144.2543 Ask: H: 144.2889 Ask: **L: 144.2543** Ask: C: 144.2889 "

> asset.Cache.GetData<TradeBar>().ToString()
"SPY: O: 144.2543 H: 144.4532 **L: 143.4156** C: 144.2716 V: 75423"

#### Related Issue
Closes #6973

#### Motivation and Context
Fix modeling bug.

#### Requires Documentation Change
No.

#### How Has This Been Tested?
New unit tests exposing the bug.
Regression tests.

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

Issue #963, opened 2017-06-10
ImmediateFillModel limit order fill price assumption very optimistic


4real:
I'm not really talking about the fact that the order is assumed immediately filled when there is a price match (which is the point of the class), but the price at which the fill takes place:

```
// fill at the worse price this bar or the limit price, this allows far out of the money limits
// to be executed properly
fill.FillPrice = Math.Min(prices.High, order.LimitPrice);
```

More likely, when we're dealing with not frequently traded stocks, the order will in reality be filled at the limit price, which can differ significantly from the high point of the next bar. So it's possible to write an intraday algorithm that trades on the distortion caused by ImmediateFillModel... Why is this a problem when volume is ignored? Because even if you make a simple modification to a copy of ImmediateFillModel to e.g. ignore trades below a certain volume the fill price problem remains, and the later you detect issues like these the more time you waste.

What's problematic about the code in question: The Min/Max allows one case to be properly handled (that otherwise would be pessimistically dealt with) but makes another case very common for limit orders (order remaining passive in limit book) optimistically erroneously dealt with. I much prefer pessimism over optimism because the latter leads to untradable algorithms.

Suggested fix:
```
fill.FillPrice = order.LimitPrice;
```
Alternatively, if keeping old behavior for immediate fills is important, make some logic to determine if this is the first bar a fill could have occurred, and if so apply old fill price function.

Decided to post this directly here rather than on forum since it's a straightforward fix if accepted.

Petter

jaredbroad:
Tracking this here https://github.com/QuantConnect/Lean/issues/6973

PR #7012, opened 2023-02-24
Fixes EquityFillModel.LimitIfTouchedFill

AlexCatarino:
#### Description
- Adds Unit Tests for Limit-If-Touched Order Fill to cover the following features following #7005:
  - The `LimitIfTouchedFill` method should not fill using `QuoteBar` or `Tick` with `TickType.Quote` type **unless** the trigger price and the limit price are potentially attained or penetrated in the same bar. In this case, we need to use the closing (quote) price to test the limit fill since it's certain that the closing price doesn't come before the trigger price.
  - Adds missing test for tick data (`TickType.Trade`).
  - Favorable gap test (see #963).
- Fixes EquityFillModel LimitIfTouchedFill
  - Use the entire tick history to find the trigger price
  - Fill price is the limit price (as `LimitFill`)
- Updates LimitIfTouchedRegressionAlgorithm
  - Update the expected order tickets using the limit price as the fill price to reflect the change to the optimistic fill.

#### Related Issue
Ref. #7005

#### Motivation and Context
Fix modeling bug.

No.

New unit tests exposing the bug.
Regression tests.

#### Types of changes

#### Checklist:


Tests/Common/Orders/Fills/EquityFillModelTests.LimitIfTouchedFill.cs
```diff
@@ -0,0 +1,472 @@
...
+
+            // Time jump => limit reached, security bought
+            // |---> First, ensure that price data are not used to fill
+            security.SetMarketPrice(new TradeBar(Noon, Symbols.SPY, 100m, 100m, 99m, 99m, 100));
+            fill = model.LimitIfTouchedFill(security, order);
+            Assert.AreEqual(0, fill.FillQuantity);
+            Assert.AreEqual(0, fill.FillPrice);
+            Assert.AreEqual(OrderStatus.None, fill.Status);
+
+            // |---> Lastly, ensure that quote data used to fill
```
Martin-Molinero, line 97: Cool! seems comments are off?
Would be great if the first part of the test which are duplicated come from a shared private method?

Common/Orders/Fills/EquityFillModel.cs
```diff
@@ -112,7 +112,7 @@ public override OrderEvent LimitIfTouchedFill(Security asset, LimitIfTouchedOrde
                         if (askCurrent <= order.LimitPrice)
```
Martin-Molinero, line 112: >The LimitIfTouchedFill method should not fill using QuoteBar or Tick with TickType.Quote type

Just double checking, `LimitIfTouchedFill` is filling using quote data but getting triggered using trade data. The PR description describes another behavior?

PR #7161, opened 2023-04-03
Accounting for the position quantity in margin requirements

### LINKED ISSUE #7153: Cannot liquidate part of an option strategy naked short position
#### Expected Behavior

When holding a naked short position for an option strategy, the algorithm is able to liquidate part of that position.

#### Actual Behavior

When holding a naked short position for an option strategy (say `group quantity=-5`), the algorithm is not able to place a reverting order to liquidate part of the position (`quantity < |strategyQuantity|`. E.g. `quantity = 4` in this case).

#### Potential Solution

A potential hint: look into the `PositionGroupBuyingPowerModel`, specifically into how the available and delta buying power are calculated in [`HasSufficientBuyingPowerForOrder`](https://github.com/QuantConnect/Lean/blob/4d3211957631e085fe299c8452c794c180a4f165/Common/Securities/Positions/PositionGroupBuyingPowerModel.cs#L142), which is used by the transaction handler to check where an order can be submitted and filled. The condition `deltaBuyingPower <= availableBuyingPower` is not met for this cases, resulting in unsufficient buying power.

#### Reproducing the Problem

Run this algorithm:

```python
import itertools

class SleepyTanMule(QCAlgorithm):

    def Initialize(self):        
        self.SetStartDate(2015, 12, 24)
        self.SetEndDate(2015, 12, 24)
        self.SetCash(150000)
        
        equity = self.AddEquity("GOOG", leverage=4)
        option = self.AddOption(equity.Symbol)
        self._optionSymbol = option.Symbol

        option.SetFilter(lambda u: u.Strikes(-2, +2).Expiration(0, 180))

    def OnData(self, data):
            for chain in data.OptionChains.Values:
                callContracts = [contract for contract in chain if contract.Right == OptionRight.Call]
                callContracts = [(key, list(group)) for key, group in itertools.groupby(callContracts, key=lambda x: x.Expiry)]
                callContracts.sort(key=lambda x: x[0])
                callContracts = callContracts[0][1]
                callContracts.sort(key=lambda x: x.Strike)

                if len(callContracts) < 3:
                    break

                expiry = callContracts[0].Expiry
                lowerStrike = callContracts[0].Strike
                middleStrike = callContracts[1].Strike
                higherStrike = callContracts[2].Strike

                self._optionStrategy = OptionStrategies.Straddle(self._optionSymbol, higherStrike, expiry)

                self._strategyQuantity = -5
                self.Order(self._optionStrategy, self._strategyQuantity)  
        else:
            self.Order(self._optionStrategy, abs(self._strategyQuantity) - 1)
```

Se the orders and logs:

image

#### System Information

Local environment in Windows 11 and QC Cloud

#### Checklist

jhonabreul:
#### Description

When shorting options and options strategies with no initial holdings, the margin buying power model allowed us to short too many contracts, making us go avobe the available margin. This is because it was using the existing security holdings to calculate the margin requirements for any existing or potential position.

This fixes this by pssing the actual quantity down to the margin requirement calculation method.

- Quantity is reduced in initial orders when there are no holdings, because the models where allowing highier quantities than actually margin was available for. For some algorithms, the order quantities were reduced to match this, as well as cash was increased for some others for them to be able to hold the positions.
- Note about `Algorithm.CSharp/OptionChainConsistencyRegressionAlgorithm.cs`: the orders list hash changed. The reason is the following: this algorithm runs for a single day, placing a market order and then liquidating it on market close. Right after that, the condition `!Portfolio.Invested` is true again, so the orders are placed again but never filled because the algorithm is completed. The last MOC order is invalidated with this changes because of insufficient buying power, hence the hash change.

##### Shorting options outside of margin in IB for validation

image
image

#### Related Issue

Closes #7153 

Unit tests, regression algorithm.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Common/Securities/OptionMarginBuyingPowerModelTests.cs
+++ b/Tests/Common/Securities/OptionMarginBuyingPowerModelTests.cs
@@ -465,6 +465,61 @@ namespace QuantConnect.Tests.Common.Securities
             Assert.AreEqual(expectedOrderSize, result.Quantity);
         }
 
+        [TestCase(0)]
+        [TestCase(-10)]
+        public void GetsMaintenanceMarginForAPotentialShortPositionWithoutInitialHoldings(decimal initialHoldings)
+        {
+            // Computing the maintenance margin for a potential position is useful because it will be used to check whether there is
+            // enough available buying power to open said new position.
+
+            const decimal price = 14m;
+            const decimal underlyingPrice = 196m;
+            var tz = TimeZones.NewYork;
+
+            var equity = new QuantConnect.Securities.Equity.Equity(
+                SecurityExchangeHours.AlwaysOpen(tz),
+                new SubscriptionDataConfig(typeof(TradeBar), Symbols.SPY, Resolution.Minute, tz, tz, true, false, false),
+                new Cash(Currencies.USD, 0, 1m),
+                SymbolProperties.GetDefault(Currencies.USD),
+                ErrorCurrencyConverter.Instance,
+                RegisteredSecurityDataTypesProvider.Null
+            );
+            equity.SetMarketPrice(new Tick { Value = underlyingPrice });
+
+            var optionCall = new Option(
+                SecurityExchangeHours.AlwaysOpen(tz),
+                new SubscriptionDataConfig(
+                    typeof(TradeBar),
+                    Symbols.SPY_C_192_Feb19_2016,
+                    Resolution.Minute,
+                    tz,
+                    tz,
+                    true,
+                    false,
+                    false
+                ),
+                new Cash(Currencies.USD, 0, 1m),
+                new OptionSymbolProperties("", Currencies.USD, 100, 0.01m, 1),
+                ErrorCurrencyConverter.Instance,
+                RegisteredSecurityDataTypesProvider.Null
+            );
+            optionCall.SetMarketPrice(new Tick { Value = price });
+            optionCall.Underlying = equity;
+            optionCall.Holdings.SetHoldings(price, initialHoldings);
+
+            var buyingPowerModel = new OptionMarginModel();
+
+            if (initialHoldings == 0)
+            {
+                // No holdings for the option, so no maintenance margin expected
+                Assert.AreEqual(0m, buyingPowerModel.GetMaintenanceMargin(optionCall));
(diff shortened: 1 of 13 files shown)


Tests/Common/Securities/OptionMarginBuyingPowerModelTests.cs
```diff
@@ -412,31 +412,26 @@ public void NonAccountCurrency_GetBuyingPower(decimal nonAccountCurrencyCash)
...
+        [TestCase(-2, 838, 0.40)] // Short to Long (-2 + 838 = 836)
+        [TestCase(2, 834, 0.40)] // Long To Longer (2 + 834 = 836)
 
         // ~0.04% Target (~400). This is below the needed margin for one unit. We end up at 0 holdings for all cases.
-        [TestCase(-31, 31, 0.0004)] // Short to Long (-31 + 31 = 0)
         [TestCase(0, 0, 0.0004)] // Open Long (0 + 0 = 0)
-        [TestCase(31, -31, 0.0004)] // Long To Longer (31 + -31 = 0)
+        [TestCase(-2, 2, 0.0004)] // Short to Long (-2 + 2 = 0)
+        [TestCase(2, -2, 0.0004)] // Long To Longer (2 + -2 = 0)
         public void CallOTM_MarginRequirement(int startingHoldings, int expectedOrderSize, decimal targetPercentage)
```
Martin-Molinero, line 435: Nice! ✨

Tests/Common/Securities/OptionStrategyPositionGroupBuyingPowerModelTests.cs
```diff
@@ -0,0 +1,224 @@
...
+            );
+            _putOption.Underlying = _equity;
+
+            _portfolio.Securities.Add(_equity.Symbol, _equity);
+            _portfolio.Securities.Add(_callOption.Symbol, _callOption);
+            _portfolio.Securities.Add(_putOption.Symbol, _putOption);
+        }
+
+        [Test]
+        public void TestHasSufficientBuyingPower([Values] bool withInitialHoldings)
```
Martin-Molinero, line 101: cool!

Martin-Molinero (approved):

PR #7173, opened 2023-04-11
Fix SMA first value calculation

### LINKED ISSUE #6336: ExponentialMovingAverage first value calculation
#### Expected Behavior

When testing TSI and TRIX indicators (which depend on EMA) against TradingView data, tests don't pass. According to [TradingView's description of EMA](https://www.tradingview.com/scripts/ema/?solution=43000502589#:~:text=For%20the%20first%20EMA%2C%20we%20use%20the%20SMA(previous%20day)%20instead%20of%20EMA(previous%20day).), the first value after the indicator is warmed up (Samples == period) should be calculated as a SMA of the first N=period elements. Other references also indicate this ([StockCharts](https://school.stockcharts.com/doku.php?id=technical_indicators:moving_averages#:~:text=First%2C%20calculate%20the%20simple%20moving%20average%20for%20the%20initial%20EMA%20value.%20An%20exponential%20moving%20average%20(EMA)%20has%20to%20start%20somewhere%2C%20so%20a%20simple%20moving%20average%20is%20used%20as%20the%20previous%20period%27s%20EMA%20in%20the%20first%20calculation), [Investopedia 1](https://www.investopedia.com/terms/e/ema.asp#:~:text=Calculating%20the%20EMA%20requires%20one%20more%20observation%20than%20the%20SMA.%20Suppose%20that%20you%20want%20to%20use%2020%20days%20as%20the%20number%20of%20observations%20for%20the%20EMA.%20Then%2C%20you%20must%20wait%20until%20the%2020th%20day%20to%20obtain%20the%20SMA.%20On%20the%2021st%20day%2C%20you%20can%20then%20use%20the%20SMA%20from%20the%20previous%20day%20as%20the%20first%20EMA%20for%20yesterday.), [Investopedia 2](https://www.investopedia.com/ask/answers/122314/what-exponential-moving-average-ema-formula-and-how-ema-calculated.asp#:~:text=There%20are%20three,Compute%20the%20SMA), [Sciencing](https://sciencing.com/calculate-moving-range-4963173.html#:~:text=Since%20you%20have%20to%20start%20your%20calculations%20somewhere%2C%20the%20initial%20value%20for%20your%20first%20EMA%20calculation%20will%20actually%20be%20an%20SMA), [Corporate Finance Institute](https://corporatefinanceinstitute.com/resources/knowledge/trading-investing/exponential-moving-average-ema/#:~:text=The%20formula%20for%20calculating%20the%20EMA%20is%20as%20follows%3A), [TheTradingBible](https://thetradingbible.com/exponential-moving-average-ema-indicator#:~:text=Easy.%20In%20the%20first%20EMAs%20required%20in%20your%20formula%20you%27ll%20use%20a%20Simple%20Moving%20Average%20as%20the%20initial%20value.%C2%A0), [WarrorTrading](https://www.warriortrading.com/exponential-moving-average/#:~:text=Since%20an%20EMA,divide%20by%2020.), [Groww](https://groww.in/p/exponential-moving-average), [Ashtrade](https://asthatrade.com/blogs/what-is-exponential-moving-average-ema/)).

After changing ExponentialMovingAverage to return 0 during the warmup period, the SMA of the first N=period samples as the first elements and then using the [EMA formula](https://github.com/QuantConnect/Lean/blob/0bc087954e9358565eb9cffed53a508f01015580/Indicators/ExponentialMovingAverage.cs#L99) for the following samples, tests pass for the TradingView data.

#### Actual Behavior

Currently, Lean's EMA indicator returns identity for the first sample and then uses the formula for the following ones, even during the warmup period. This way of computing EMA is described [here](https://en.wikipedia.org/wiki/Moving_average#:~:text=The%20EMA%20for,be%20calculated%20recursively%3A) and that is how is is implemented in [CRAN's pracma package](https://github.com/cran/pracma/blob/12e26b7898f3fb6da81e710947bff1bfaf513f81/R/movavg.R#L33-L36) in R.

#### Potential Solution

The idea would be to study which way of computing EMA is the right one. In case the SMA approach is correct, the solution would include changing EMA indicator to wait for the warmup period and then start the EMA calculation with the SMA of the first N=period samples.

Also, this would require checking that all the indicators that use or depend on EMA pass their own tests as well.

#### System Information

N/A

#### Checklist

jhonabreul:
#### Description

As [TALib](https://sourceforge.net/p/ta-lib/code/HEAD/tree/branches/ta-lib/ta-lib/c/src/ta_func/ta_EMA.c#l226) and [TradingView](https://es.tradingview.com/?solution=43000502589#:~:text=For%20the%20first%20EMA%2C%20we%20use%20the%20SMA(previous%20day)%20instead%20of%20EMA(previous%20day).) do, the first value for the EMA (right after warm up is finished) is commonly calculated as the SMA of the first period. Right after that, the SMA formula is used using the first SMA as the seed or previous EMA value.

Test data for the EMA and other indicators based on it was exported from TradingView.

#### Related Issue

Closes #6336 

#### Types of changes

#### Checklist:


Martin-Molinero (approved):
Looks good ty 👍

Grosner300:
It appears to me that the fixed EMA calculates the first value in the EMA calculation as the SMA of the future n values in the series. 

So given a historical series of 2,4,6,8,10, the first value the EMA is using is 6. 

I believe the EMA should use, as its first value, an SMA from the 5 periods BEFORE the 1st period (e.g. 2). 

Can you recheck and confirm if the formula should again be adjusted? I use Trading View and a brokerage site to compute and EMA value, and believe QuantConnect should produce the same result (given that the input data and smoothing factor is the  the same)

jaredbroad:
> The "first" value that I am referring to above is the first value in the warmup Period
If this is the behavior you need, is there a reason not to go back ten days in the warm-up period?

Technically isn't it warmed up with the SMA-5 first result? (6) Can you please confirm this "EMA should use, as its first value, an SMA from the 5 periods BEFORE the first period (e.g. 2)." - with an export from Trading View of say SPY data-EMA5?

Grosner300:
Jared, Thank you very much for responding!

The EMA warmup probably affects dependent indicators too. Going back an extra few days is a workaround but may not IMHO be the best approach. I think the indicator should work correctly out of the box and consistently match third party calculations where it makes sense. 

I am included two images: 

1. Trading View FCX January 2019-01-04 5 Day EMA value: 10.45. 
2. QuantConnect Local Lean engine's FCS value (10.406) as determined during a program debug

- I verified the close prices are identical between my data and TV Data. 
- I compared the TV calculations to the TOS platform and they are consistent! 
- I realize the differences are small, in this case.

Hopefully this can help in reproducing what I am see. 

I believe the cause for the differences is that QuantConnect is calculating the very first EMA the SMA of THE FUTURE n PERIODS. I believe the correct calculation is to use the PAST n PERIODS. 

TradingView
FCX-TradingView-5 Day FastEMA-20190104

QuantConnect:
FCX-QuantConnect-5 Day FastEMA-20190104

AlexCatarino:
Hi @Grosner300 ,

> I think the indicator should work correctly out of the box and consistently match third party calculations where it makes sense.

When we add or fix an indicator on LEAN, we validate it with unit tests with external data generated by third-party calculation.  For example, Jhonathan created  [spy_ema.csv](https://github.com/QuantConnect/Lean/pull/7173/files#diff-e37e175f38e7cedf1e523d89ad49aee656d34490408454a481fb891ed10c1ab3) using TA-Lib for this pull request. 

If you could create a new `spy_ema.csv` by exporting data from TradingView, we could run our unit tests to verify/debug and/or fix the implementation. We cannot replace the `spy_ema.csv` generated by TA-Lib with manual calculations.

Grosner300:
The [spy_ema.csv](https://github.com/QuantConnect/Lean/pull/7173/files#diff-e37e175f38e7cedf1e523d89ad49aee656d34490408454a481fb891ed10c1ab3) file contains monthly EMA calculations beginning on 1/1/1993. The EMA values are correct, assuming that the EMA calculation always begins at the earliest point of data possible. 

I did notice that TradingView uses all available data to create the starting point for the EMA, so its values match those in the spy_ema.csv file. However, QuantConnect does not use the earliest starting point. Instead, it uses an alternative calculation that takes into account the lack of data at the beginning of the time series. This results in different EMA values than those calculated by TradingView. This alternative calculation is  documented in [Investopedia] (https://www.investopedia.com/ask/answers/122314/what-exponential-moving-average-ema-formula-and-how-ema-calculated.asp) and another website: [How to calculate exponential moving average in Excel](https://www.spreadsheetweb.com/how-to-calculate-exponential-moving-average-in-excel/).  

To match the EMA values on websites like TradingView, QuantConnect would need to calculate the EMA using all available split-adjusted data from the start of time the ticker begins trading. However, the current QuantConnect methodology is also acceptable, so it should be preserved. It would be nice, however, if an alternate Long Term EMA (LEMA?) could be calculated that uses all available data for the ticker.

During my analysis, I found an informative [video](https://youtu.be/SQ2TgUQoLq0?si=0D5GhoZZrwDymddZ&t=241) that sheds more light on why exponential moving averages are used and different ways they are calculated.

Hopefully, this explains why there are differences between what websites like Trading View might show for an EMA and what is shown from QuantConnect.

AlexCatarino:
Hi @Grosner300 ,

You can make a historical data request to fetch all available data for the ticker and update the EMA accordingly:

```python
symbol = self.AddEquity("SPY").Symbol
ema = self.EMA(symbol, 10, Resolution.Daily)
history = selh.History[TradeBar](symbol, datetime(1998,1,1), self.Time, Resolution.Daily)
for bar in history:
    self.ema.Update(bar.EndTime, bar.Close)
```

PR #7221, opened 2023-04-26
Multiple option position groups liquidation

### LINKED ISSUE #7205: Error buying and then selling bull call and bear put spreads
#### Expected Behavior
Should be able to buy and then sell bull call and bear put spreads

#### Actual Behavior
Get: Backtest Handled Error: Order Error: id: 6, Error executing margin models: Unable to create group for orders: [5,6]

#### Potential Solution
This may be related to issue #6891 or #7054 or #7184 
Source of error may be: https://github.com/QuantConnect/Lean/blob/761ea2571ac9ed5fbb5d22304b81baa36e9f70df/Common/Securities/Positions/OptionStrategyPositionGroupResolver.cs#L60

#### Reproducing the Problem
`

    public class BuyThenSellSpreadsError : QCAlgorithm
    {
        private Symbol symbol;
        private int step;
        private decimal atm_strike;
        OptionStrategy bullCallSpread;
        OptionStrategy bearPutSpread;

        {
            SetStartDate(2022, 1, 1);
            SetEndDate(2022, 2, 1);
            SetCash(500000);
            var option = AddOption("SPY", Resolution.Minute);
            symbol = option.Symbol;
            option.SetFilter(universe => universe.Strikes(-10, 10).Expiration(0, 1));
            step = 1;
        }

        {
            // we will only execute the 2 steps: 
            // 1. buy bear put and bull call spreads
            // 2. sell the spreads bought in step 1
            if (step > 2)
            {
                return;
            }

            var chain = data.OptionChains.get(symbol, null);
            if (chain == null || !chain.Any())
            {
                return;
            }

            if (step == 1)
            {
                atm_strike = chain.OrderBy(x => Math.Abs(x.Strike - chain.Underlying.Price)).ToList()[0].Strike;
            }

            var contracts = (from i in chain where i.Strike == atm_strike select i).ToList();
            if (contracts.Count == 0)
            {
                return;
            }

            // step 1: buy 1 atm bull call and bear put spreads
            if (step == 1)
            {
                // buy spreads
                bullCallSpread = OptionStrategies.BullCallSpread(symbol, atm_strike, atm_strike + 1, Time.Date);
                Buy(bullCallSpread, 1);

                bearPutSpread = OptionStrategies.BearPutSpread(symbol, atm_strike, atm_strike - 1, Time.Date);
                Buy(bearPutSpread, 1);
                step += 1;
            }
            else if (step == 2)
            {
                // step 2: attempt to sell the spreads bought in step1 - error occurs: 
                // Backtest Handled Error: Order Error: id: 6, Error executing margin models: Unable to create group for orders: [5,6]
                Sell(bullCallSpread, 1);
                Sell(bearPutSpread, 1);
                step += 1;
            }
        }

        {
            foreach (var x in Portfolio)
            {
                if (x.Value.Invested)
                {
                    Log("{x.Key} :: {x.Value.Quantity}");
                }
            }
        }

        {
            var order = Transactions.GetOrderById(orderEvent.OrderId);
            if (orderEvent.Status == OrderStatus.Filled)
            {
                Debug("{self.Time}: {order.Type}: {orderEvent}");
            }
        }
    }
`

#### Checklist

- [X ] I have completely filled out this template
- [ X] I have confirmed that this issue exists on the current `master` branch
- [ X] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [X ] I have provided detailed steps to reproduce the issue

jhonabreul:
#### Description

When holding multiple position groups of the same underlying, reverting one of the groups was not possible because of the presence of the other position groups, which caused no matching groups to be formed for the order. Now we can detect these situations and group the positions accordingly.

##### Important note abuot liquidation

`QCAlgorithm.Liquidate()` when called without symbol, will liquidate each invested securitiy placing market orders individually for each one of them, regardless of whether there are position groups like option strategies. This can cause issues because the margin requirements to hold a full strategy can be lower than the one required to hold only a subset of its legs (like after liquidating the first leg).

Take for example a Butterfly Call Strategy, which has a margin requirement of zero (0). If the long legs are liquidated first, there might be the case where the available margin is less than the required to hold the short leg, so the liquidation of the long legs would fail with unsifficient buying power.

This is what happended with the `ComboOrderAlgorithm` 's and that is the reason for the initial cash increase. 

This was **tested using IB TWS** locally trading a large quantity of a Butterfly Call strategy and then trying to close the position for the first leg. The account margin is not enough for the remaning contracts, so the order is not accepted:

image

#### Related Issue

Closes #7205 

Regression algorithm

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/ComboLegLimitOrderAlgorithm.cs
+++ b/Algorithm.CSharp/ComboLegLimitOrderAlgorithm.cs
@@ -85,9 +85,9 @@ namespace QuantConnect.Algorithm.CSharp
             {"Tracking Error", "0"},
             {"Treynor Ratio", "0"},
             {"Total Fees", "$75.00"},
-            {"Estimated Strategy Capacity", "$8000.00"},
+            {"Estimated Strategy Capacity", "$100000.00"},
             {"Lowest Capacity Asset", "GOOCV W78ZERHAOVVQ|GOOCV VP83T1ZUHROL"},
-            {"Portfolio Turnover", "698.54%"},
+            {"Portfolio Turnover", "30.16%"},
             {"OrderListHash", "2b64aec759a089d23ccf9722d6b87ccd"}
         };
     }
--- a/Algorithm.CSharp/ComboLimitOrderAlgorithm.cs
+++ b/Algorithm.CSharp/ComboLimitOrderAlgorithm.cs
@@ -154,7 +154,7 @@ namespace QuantConnect.Algorithm.CSharp
             {"Total Fees", "$85.00"},
             {"Estimated Strategy Capacity", "$5000.00"},
             {"Lowest Capacity Asset", "GOOCV W78ZERHAOVVQ|GOOCV VP83T1ZUHROL"},
-            {"Portfolio Turnover", "1893.90%"},
+            {"Portfolio Turnover", "60.92%"},
             {"OrderListHash", "186986b27fac082600a064a8a59df604"}
         };
     }
--- a/Algorithm.CSharp/ComboMarketOrderAlgorithm.cs
+++ b/Algorithm.CSharp/ComboMarketOrderAlgorithm.cs
@@ -74,9 +74,9 @@ namespace QuantConnect.Algorithm.CSharp
             {"Tracking Error", "0"},
             {"Treynor Ratio", "0"},
             {"Total Fees", "$75.00"},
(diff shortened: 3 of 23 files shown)


Martin-Molinero (approved):
Nice work 👍

Issue #7207, opened 2023-04-24
Improve Export Signal Examples


AlexCatarino:
#### Expected Behavior
The examples should generate a signal on deployment.

#### Actual Behavior
The current examples do not generate signals on deployment since they depend on trading criteria based on indicators.

#### Potential Solution
Review the algorithm's logic.

#### Reproducing the Problem
Deploy the algorithms in live mode, and verify that no order (therefore signal) is generated in the first OnData call.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #7233, opened 2023-05-03
Update Numerai Demo Algorithms

DerekMelchin:
#### Description

Update Numerai Demo Algorithms

#### Related Issue

#7207

#### Motivation and Context

Updates `NumeraiSignalExportDemonstrationAlgorithm` to demonstrate how to use the proper universe and schedule signals for when Numerai expects them.

Remove `NumeraiPortfolioSignalExportDemonstrationAlgorithm` because we can't use `SetTargetPortfolioFromPortfolio` with Numerai. We need to send signals to Numerai before market opens. If we try to rebalance the portfolio before the market opens (and we don't have extended market hours), our orders are converted to MOC orders. So if we use `SetTargetPortfolioFromPortfolio`, it sends signals to Numerai using the current portfolio holdings, not the holdings we want at market open.

N/A

QC Cloud

#### Types of changes

#### Checklist:


Algorithm.CSharp/NumeraiSignalExportDemonstrationAlgorithm.cs
```diff
@@ -31,184 +31,91 @@ namespace QuantConnect.Algorithm.CSharp
...
+            var denominator = symbols.Count() * (symbols.Count() + 1) / 2.0m; // sum of 1, 2, ..., symbols.Count()
+            symbols = symbols.OrderBy(symbol => symbol);
+            var targets = new List<PortfolioTarget>();
+            for (var i = 0; i < symbols.Count(); i++)
             {
-                SetHoldings("SPY", 0.1);
-                _targets[0] = new PortfolioTarget(Portfolio["SPY"].Symbol, (decimal)0.1);
-                SignalExport.SetTargetPortfolio(_targets);
-                _firstCall = false;
+                targets.Add(new PortfolioTarget(symbols.ToList()[i], (i + 1) / denominator));
```
Martin-Molinero, line 88: `symbols` is an enumerable, every operation `.Count()/.ToList()/.OrderBy` is re enumerating the linq statement. Suggest moving `OrderBy` at the top same to the `ToList`

I'm a bit confused by `var denominator = symbols.Count() * (symbols.Count() + 1) / 2.0m;` & `(i + 1) / denominator)`, for example for `10` symbols this will be 1/55, 2/55 until reaching 10? maybe just simpler to do `1 / count` for all of them?
Similar comments for python algorithm 👍

Martin-Molinero (commented):
Nice! minor comments 👍

Algorithm.CSharp/NumeraiSignalExportDemonstrationAlgorithm.cs
DerekMelchin: We can't give an equal weight to each security because Numerai requires at least one of them to be unique.
In this example, we give a linear allocation to each asset. This is correct:
> for 10 symbols this will be 1/55, 2/55 until reaching 10?

Martin-Molinero (approved):

PR #7314, opened 2023-06-12
Indicators history rolling window

### LINKED ISSUE #3372: Add RollingWindow History on All Indicators
#### Expected Behavior
Rolling window (history), should be available just by specifying history length in constructor or calling some method with this parameter after initialisation.
This is quite heavily used feature, given number of question for this topic. It is available in implicit form similar to described one in most other trading/backtesting platforms...
.
#### Actual Behavior
To add rolling window functionality, there is need to explicitly add values to explicitly created rolling window

#### Potential Solution
Example usage: 
self.RSI_Ind = self.RSI("AAPL", RSI_Period, history= 50)
...
if self.RSI_Ind.history.IsReady:
  self.Debug ("Previous RSI value is: {}".format(self.RSI_Ind.history[1]))

jhonabreul:
#### Description

Implement a rolling window in `IndicatorBase` to keep track of indicator's history for a user-defined period.

- The window size can be set/overriden using `indicator.Window = <value>;`.
- The number of currently store history points in the window can be accessed through `indicator.WindowCount`.
- Indicators are now enumerables, so the latest indicator value can be found on index 0 (`indicator[0]`) and the oldest value on index `indicator.WindowCount - 1`.
- Indexing an indicator history outside the current count `null`.
- Indexing an indicator history outside the window size (`indicator[i]` where `i >= indicator.Window`) will return null and resize the window to `i + 1`.

#### Related Issue

Closes #3372 

We may need to add documentation for this new indicator's feature.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/IndicatorHistoryAlgorithm.cs
+++ b/Algorithm.CSharp/IndicatorHistoryAlgorithm.cs
@@ -58,6 +58,12 @@ namespace QuantConnect.Algorithm.CSharp
                 if (_sma[0] < _sma[_sma.WindowCount - 1])
                 {
                     Buy(_symbol, 100);
+
+                    // Let's log the SMA values for the last 14 days, for demonstration purposes on how it can be enumerated
+                    foreach (var dataPoint in _sma)
+                    {
+                        Log($"SMA @{dataPoint.EndTime}: {dataPoint.Value}");
+                    }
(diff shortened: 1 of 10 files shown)


Martin-Molinero (approved):
Awesome! 🚀

PR #7321, opened 2023-06-16
Fix `CustomPartialFillModelAlgorithmFails.cs/py`

### LINKED ISSUE #5694: CustomPartialFillModelAlgorithm Throws Error During Short Trades
#### Expected Behavior

CustomPartialFillModelAlgorithm runs without error.

#### Actual Behavior

Whenever the CustomPartialFillModelAlgorithm places short trades, we get
`Insufficient buying power to complete order`

#### Potential Solution

Adjust the CustomPartialFillModel to work with long and short trades.

#### Reproducing the Problem

Run the [CustomPartialFillModelAlgorithm](https://github.com/QuantConnect/Lean/blob/master/Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs).

#### System Information

QC Cloud web IDE

#### Checklist

Marinovsky:
#### Description

The algorithm `CustomPartialFillModelAlgorithm.cs` was not working as expected with short orders because in its Fill model, the variable `absoluteRemaining` instead of being decreased was being increased since the `FillQuantity` in each call to this method was negative. Therefore when this amount was substracted to `absoluteRemaining`, it was actually added. Hence the method created more partial orders than expected, specifically: 580 short orders for SPY with Quantity -10.
#### Related Issue

Closes #5694 
#### Motivation and Context

With this change user will have a working example algorithm for custom partial fill models

This change does not require changes to documentation

Both example algorithms `CustomPartialFillModelAlgorithm.cs/py` were converted into regression algorithms
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
+++ b/Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
@@ -114,7 +114,7 @@ namespace QuantConnect.Algorithm.CSharp
         /// <summary>
         /// This is used by the regression test system to indicate which languages this algorithm is written in.
         /// </summary>
-        public Language[] Languages { get; } = { Language.CSharp };
+        public Language[] Languages { get; } = { Language.CSharp, Language.Python };
 
         /// <summary>
         /// Data Points count of all timeslices of algorithm
--- a/Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
+++ b/Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
@@ -53,7 +53,7 @@ namespace QuantConnect.Algorithm.CSharp
 
             if (Time.Day > 10 && _holdings.Quantity <= 0)
             {
-                MarketOrder(_spy, 100, true);
+                MarketOrder(_spy, 105, true);
             }
             else if (Time.Day > 20 && _holdings.Quantity >= 0)
             {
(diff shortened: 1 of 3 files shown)


Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
```diff
@@ -89,20 +90,71 @@ public override OrderEvent MarketFill(Security asset, MarketOrder order)
                 // Set this fill amount
                 fill.FillQuantity = Math.Sign(order.Quantity) * 10;
```
jhonabreul, line 91: We could take this chance here and get the max of this quantity and te signed remaining one, for cases where the original quantity is not a multiple of 10. Just to make the example more accurate.
Same for Python version.

jhonabreul (changes requested):

jhonabreul:
Nice, thank you! Leaving a minor comment above

Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
```diff
@@ -87,11 +87,13 @@ public override OrderEvent MarketFill(Security asset, MarketOrder order)
                 // Create the object
                 var fill = base.MarketFill(asset, order);
 
-                // Set this fill amount
-                fill.FillQuantity = Math.Sign(order.Quantity) * 10;
+                // Set the fill amount to the maximum 10-multiple smaller than the order.Quantity for long orders
+                // Set the fill amount to the minimum 10-multiple greater than the order.Quantity for short orders
+                fill.FillQuantity = Math.Sign(order.Quantity) * 10m * Math.Floor(Math.Abs(order.Quantity) / 10m);
```
jhonabreul, line 92: This would fill 100 the first time. You can just leave this as `Math.Sign(order.Quantity) * 10;` and just get the min between it and the remaining: `Math.Sign(order.Quantity) * Math.Min(10m, absoluteRemaining)`. This would fill 10 at a time and the remaining 5 (or whatever the remaning is) at the end.

Algorithm.CSharp/CustomPartialFillModelAlgorithm.cs
```diff
@@ -53,7 +53,7 @@ public override void OnData(Slice data)
 
             if (Time.Day > 10 && _holdings.Quantity <= 0)
             {
-                MarketOrder(_spy, 100, true);
+                MarketOrder(_spy, 105, true);
```
jhonabreul, line 56: 💯

jhonabreul (changes requested):
Thanks! Just leaving another minor comment on the custom partial fill model 👌

jhonabreul (approved):

PR #7336, opened 2023-06-27
Fix bug in VolumeWeightedAveragePriceIndicator

### LINKED ISSUE #4964: VolumeWeightedAveragePriceIndicator reset
#### Expected Behavior

- `VolumeWeightedAveragePriceIndicator.Reset()` leaves residual unrequired `CompositeIndicator` instances on each call
#### Actual Behavior

- `VolumeWeightedAveragePriceIndicator.Reset()` calls `WeightedBy()` but does not dispose of the previous `WeightedBy` -> `CompositeIndicator` which will continue to receive updates
#### Potential Solution

- `VolumeWeightedAveragePriceIndicator`: Instead of creating a new `WeightedBy` could we reset it?
- `CompositeIndicator` should be disposed of correctly when no longer used.
- Review usages of the `CompositeIndicator` and make sure they are all disposed of correctly. This bug is probably happening in other use cases too.
#### Reproducing the Problem

`VolumeWeightedAveragePriceIndicatorTests.ComparesAgainstExternalDataAfterReset`
#### System Information

N/A
#### Checklist

Marinovsky:
#### Description

When this indicator was reset, instead of reset its VolumeWeightedAveragePrice composite indicator, it created a new one. So the previous was never disposed. In order to fix this, there was created a new indicator called `ResetCompositeIndicator.cs` that takes a composite indicator and an action to execute once the given indicator is reset. Therefore, when `Reset()` method is called in `VolumeWeightedAveragePriceIndicator`, the `VolumeWeightedAveragePriceIndicator` is reset instead of re-assigned. Immediately after , this `ResetCompositeIndicator` executes the given action to reset variables used in `IndicatorExtensions.WeightedBy()` method.

On the other hand, `IndicatorExtensions.cs` was reviewed to ensure this bug was not in other usages of `CompositeIndicator`
#### Related Issue

Closes #4964 
#### Motivation and Context

With this change, LEAN will no create a new composite indicator once `VolumeWeightedAveragePriceIndicator` is reset. Additionally, users will be able to assign an action to execute once certain indicator is reset.

This change does not require updates to documentation

There was created a unit test that asserted `VolumeWeightedAveragePrice` field in `VolumeWeightedAveragePriceIndicator.cs` was the same before and after the indicator was reset
#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/IndicatorExtensions.cs
+++ b/Indicators/IndicatorExtensions.cs
@@ -70,7 +70,7 @@ namespace QuantConnect.Indicators
         /// <param name="weight">Indicator that provides the average weights</param>
         /// <param name="period">Average period</param>
         /// <returns>Indicator that results of the average of first by weights given by second</returns>
-        public static ResetCompositeIndicator WeightedBy<T, TWeight>(this IndicatorBase<T> value, TWeight weight, int period)
+        public static CompositeIndicator WeightedBy<T, TWeight>(this IndicatorBase<T> value, TWeight weight, int period)
             where T : IBaseData
             where TWeight : IndicatorBase<IndicatorDataPoint>
         {
@@ -98,8 +98,7 @@ namespace QuantConnect.Indicators
                 denominator.Update(consolidated);
             };
 
-            var compositeIndicator = numerator.Over(denominator);
-            var resetCompositeIndicator = new ResetCompositeIndicator(compositeIndicator, () => {
+            var resetCompositeIndicator = numerator.Over(denominator, () => {
                 x.Reset();
                 y.Reset();
                 numerator.Reset();
@@ -220,10 +219,18 @@ namespace QuantConnect.Indicators
         /// </remarks>
         /// <param name="left">The left indicator</param>
         /// <param name="right">The right indicator</param>
+        /// <param name="extraResetAction">Optional aciton to execute once the composite indicator is reset</param>
         /// <returns>The ratio of the left to the right indicator</returns>
-        public static CompositeIndicator Over(this IndicatorBase left, IndicatorBase right)
+        public static CompositeIndicator Over(this IndicatorBase left, IndicatorBase right, Action extraResetAction = null)
         {
-            return new (left, right, (l, r) => r.Current.Value == 0m ? new IndicatorResult(0m, IndicatorStatus.MathError) : new IndicatorResult(l.Current.Value / r.Current.Value));
+            if (extraResetAction == null)
+            {
+                return new(left, right, (l, r) => r.Current.Value == 0m ? new IndicatorResult(0m, IndicatorStatus.MathError) : new IndicatorResult(l.Current.Value / r.Current.Value));
+            }
+            else
+            {
+                return new ResetCompositeIndicator(left, right, (l, r) => r.Current.Value == 0m ? new IndicatorResult(0m, IndicatorStatus.MathError) : new IndicatorResult(l.Current.Value / r.Current.Value), extraResetAction);
+            }
         }
 
         /// <summary>
--- a/Indicators/IndicatorExtensions.cs
+++ b/Indicators/IndicatorExtensions.cs
@@ -98,7 +98,7 @@ namespace QuantConnect.Indicators
                 denominator.Update(consolidated);
             };
 
-            var resetCompositeIndicator = numerator.Over(denominator, () => {
+            var resetCompositeIndicator = new ResetCompositeIndicator(numerator, denominator, GetOverIndicatorComposer(), () => {
                 x.Reset();
                 y.Reset();
                 numerator.Reset();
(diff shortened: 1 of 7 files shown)


Indicators/IndicatorExtensions.cs
```diff
@@ -70,7 +70,7 @@ public static T Of<T>(this T second, IIndicator first, bool waitForFirstToReady
         /// <param name="weight">Indicator that provides the average weights</param>
         /// <param name="period">Average period</param>
         /// <returns>Indicator that results of the average of first by weights given by second</returns>
-        public static CompositeIndicator WeightedBy<T, TWeight>(this IndicatorBase<T> value, TWeight weight, int period)
+        public static ResetCompositeIndicator WeightedBy<T, TWeight>(this IndicatorBase<T> value, TWeight weight, int period)
```
jhonabreul, line 73: `ResetCompositeIndicator` can be a `CompositeIndicator` that overrides the `Reset` method, that way we don't change the `WeightedBy` API (keep returning `CompositeIndicator`) and users can still access every member of the `CompositeIndicator` as per usual.

Indicators/ResetCompositeIndicator.cs
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+*/
+
+using QuantConnect.Data;
+using System;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    /// Class that extends CompositeIndicator to execute a given action once is reset
```
jhonabreul, line 22: As commented in `WeightedBy` (and also in this summary), we can extend `CompositeIndicator`.

Indicators/ResetCompositeIndicator.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// Gets current value of the given CompositeIndicator
+        /// </summary>
+        public IndicatorDataPoint Current => _compositeIndicator.Current;
+
+        /// <summary>
+        /// Gets a flag indicating when the given CompositeIndicator is ready
+        /// </summary>
+        public override bool IsReady => _compositeIndicator.IsReady;
```
jhonabreul, line 44: It should also override the remaining `CompositeIndicator` properties so that the users can access the inner composite indicator instance properties.

Indicators/ResetCompositeIndicator.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            _compositeIndicator.Reset();
+            _extraResetAction.Invoke();
+        }
+
+        /// <summary>
+        /// Updates the given CompositeIndicator
+        /// </summary>
+        /// <param name="input"></param>
+        /// <returns></returns>
```
jhonabreul, line 82: Missing param and returns comments

Tests/Indicators/VolumeWeightedAveragePriceIndicatorTests.cs
```diff
@@ -82,5 +82,34 @@ public override void ResetsProperly()
...
+            foreach (var data in TestHelper.GetTradeBarStream(TestFileName))
+            {
+                indicator.Update(data);
+            }
+            Assert.IsTrue(indicator.IsReady);
+
+            var lastVWAP = indicator.GetVolumeWeightedAveragePrice();
+            indicator.Reset();
+            var newVWAP = indicator.GetVolumeWeightedAveragePrice();
+            Assert.IsTrue(Object.ReferenceEquals(lastVWAP, newVWAP));
```
jhonabreul, line 100: Perhaps assert new state (number of samples/not ready) of the VWAP after reset

jhonabreul (changes requested):
Nice! Leaving a few comments:

- Could also add unit tests for this new `ResetCompositeIndicator` class, creating a composite indicator (any), wrapping it up with this class, updating until ready, resetting and asserting the state of each single indicator/part that composes it.
- Also could assert that each `CompositeIndicator` property when accessed in an instance of `ResetCompositeIndicator` is properly bypassed to the inner composite indicator. 
- Having these tests might be enough to revert changes in `Indicators/VolumeWeightedAveragePriceIndicator.cs` and keeping _vwap private and as a `CompositeIndicator`.

jhonabreul (approved):
Thank you! 🚀

Indicators/VolumeWeightedAveragePriceIndicator.cs
```diff
@@ -74,9 +74,7 @@ public VolumeWeightedAveragePriceIndicator(string name, int period)
         /// </summary>
         public override void Reset()
         {
-            _price.Reset();
-            _volume.Reset();
```
Martin-Molinero, line 78: We should still reset `_price/_volume` I think?

Martin-Molinero (approved):
Nice! minor comment 👍

Martin-Molinero (approved):

PR #7360, opened 2023-07-07
Make indicators work with RenkoBars and VolumeRenkoBars

### LINKED ISSUE #331: Make indicators work with Renko bars

Marinovsky:
#### Description

- Make `BaseRenkoBar` inherit from `TradeBar`
- Remove repeated fields in `BaseRenkoBar` such as: `Open, High, Low` and `Close`
- Add regression test explaining how to update an indicator with a RenkoBar
- Add unit tests asserting any indicator which inherits from `IndicatorBase<TradeBar>, IndicatorBase<IBaseDate>, BarIndicator` or `IndicatorBase<IBaseDataBar>` can receive `RenkoBar`'s and `VolumeRenkoBar`'s
#### Related Issue

Closes #331 
#### Motivation and Context

With this change users will be able to use indicators with `RenkoBar`'s and `VolumeRenkoBar`'s

This change may need updates to documentation advising indicators can be updated with `RenkoBar`'s or `VolumeRenkoBar`'s

- There was created one regression algorithm which updates three indicators with `RenkoBar's`. On the end of the algorithm it asserts all the indicators are ready and their current value is different from zero.
- Using the data from the test file defined by each indicator, it was asserted the indicator could receive as input `RenkoBar`'s and `VolumeRenkoBar`'s
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/IndicatorWithRenkoBarsRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/IndicatorWithRenkoBarsRegressionAlgorithm.cs
@@ -1,0 +1,139 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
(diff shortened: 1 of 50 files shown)


Martin-Molinero (approved):
Nice work! 💪

PR #7343, opened 2023-06-29
Fix Security.IsMarketOpen for cases outside regular MH and last bar

jhonabreul:
#### Description

This fixes `Security.IsMarketOpen` to just use the `SecurityExchange` market hours check.
The `FillModel` and `EquityFillModel` have their own additional bar checks for long bar cases when market is not actually open.

In master, `QCAlgorithm.IsMarketOpen` would return true from market open up until midnight, because the last bar (ending an 4PM for SPY for instance) will make it return true until date changes. With the changes, it will return true only whitin the segments in the MHDB.

Master, minute SPY subscription, charting IsMarketOpen return value every 30 minutes:
chart (7)

PR, minute SPY subscription, charting IsMarketOpen return value every 30 minutes:
chart (6)

For daily subscriptions, in master it returns true since market open on monday untill Friday midnight. While with the changes again, only returns true when market is actually open.

Master, daily SPY subscription, charting IsMarketOpen return value every 60 minutes:
chart (4)

PR, daily SPY subscription, charting IsMarketOpen return value every 60 minutes:
chart (5)

#### Related Issue

Closes #7340 

#### How Has This Been Tested?

Unit tests and regression algorithm.

#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #7363, opened 2023-07-10
Fix `IsExchangeIpen` bar check in `FillModel` and `EquityFillModel`

jhonabreul:
#### Description

This fixes the bar check for resolutions < daily in `FillModel.IsExchangeOpen`. Also, `EquityFillModel` now uses the base class `IsExchangeOpen` method to avoid duplicate code.

For daily resolution, we considere market open for fill purposes if the current bar end time is on the same day as the asset's local time.

For other resolutions, we check that the local time is within the current bar end time in order to avoid false positive results until midnight (see charts in #7343).

#### Types of changes

#### Checklist:


Martin-Molinero (approved):
Thank you 👍

PR #7385, opened 2023-07-18
Implement RangeConsolidator

### LINKED ISSUE #3079: Implement Range Bar Indicator
#### Expected Behavior
In a range bar, all price bars are of equal length. If you had a 10 pip range bar, and EURUSD was trading at 1.1000, then the range bar would close only when EURUSD crossed above 1.1010. Then, a new bar opens at 1.1010. Now, if EURUSD fell to 1.1005 and then closed at 1.1015, a second range bar is formed with the third range bar opening at 1.1015. Now, EURUSD rises to 1.1020 and then falls to 1.1010. A third range bar opens at 1.1020 and closes at 1.1010. Finally, EURUSD rises to 1.1015 and then falls to 1.1005. A fourth range bar opens at 1.1015 closing at 1.1005.

In essence, the close of a bar is always equal to the open of the next bar. The range is measured as a high and low and not close.

#### Potential Solution
Build new Indicator.

#### Checklist

Marinovsky:
#### Description

Range bars are quite unique since they allow the user to analyze price movement of the financial instrument, while reducing market noise. Range bars don’t take time into consideration and are therefore able to create the so-called clean chart effect. A new bar is formed when price reaches a certain value set by the user. As you can see, this approach is quite different from the usual one as it completely eliminates the time factor.

How are the Range bars built?
- Range bar has an opening price (Open). The price movement is recorded when the bar is being formed. During that time the size of the bar is growing from the Open point either up or down, depending on the price.
- The difference between the highest (High) and the lowest (Low) bar value is considered as price movement.
- Whenever the High – Low point reaches the Range interval, the formation of the bar finishes.   
- Further:
-- If the price moves in the [Low, High] range, the closing point (Close) of the bar changes.
-- When the price crosses the [Low, High] range, the bar closes and the new bar starts to form.
For more information see  (https://www.tradingview.com/blog/en/range-bars-now-available-tradingview-8078/).

The following tasks were done:
- Create `BaseTimelessConsolidator.cs`, an abstract class from which timeless consolidators can inherit
- Create `RangeConsolidator.cs` and `ClassicRangeConsolidator.cs`, two consolidators which inherit from `BaseTimelessConsolidator` since it's behavior similar. Indeed, the unique difference between them is that `ClassicRangeConsolidator` does not use intermediate/phantom bars
- Add unit and regression tests to assert the new consolidators are working as expected

Finally, a chart of the consolidated bars of SPY at daily resolution from 10/07/2013 to 10/11/2013 is shown below:
image
As you can see, the difference between the `High` and `Low` values of each RangeBar is always the same across all the bars.
#### Related Issue

Closes #3079 
#### Motivation and Context

With this change users will be able to use RangeBar's in their strategies

This change may require an update to documentation announcing the new available type of consolidator 

- It was created a unit test that consolidated RangeBar's based on a given list of datapoints, then the values of each RangeBar were asserted to check they were the expected ones.
- It was created a regression algorithm which consolidated SPY data at Daily and Tick resolution from 10/07/2013 to 10/11/2013, then it was asserted at least one RangeBar was consolidated. In the same way, it was asserted the difference between the `High` and `Low` value of each bar were always the same, as well as the volume was never zero for any RangeBar produced by `ClassicRangeConsolidator`
#### Types of changes

#### Checklist:


Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -0,0 +1,223 @@
...
+namespace QuantConnect.Data.Consolidators
+{
+    /// <summary>
+    /// This consolidator can transform a stream of <see cref="IBaseData"/> instances into a stream of <see cref="RangeBar"/>
+    /// </summary>
+    public class RangeConsolidator : BaseTimelessConsolidator
+    {
+        private bool _firstTick;
+        private decimal _minimumPriceVariation;
+        protected RangeBar CurrentRangeBar;
```
jhonabreul, line 30: Maybe could directly use `CurrentBar`

Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// <param name="range">The constant value size of each bar, where the unit is the minimum price variation of the RangeBar's symbol</param>
+        public RangeConsolidator(decimal range)
+            : this(range, x => x.Value)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RangeConsolidator" /> class.
+        /// </summary>
+        /// <param name="range">The constant value size of each bar, where the unit is the minimum price variation of the RangeBar's symbol</param>
```
jhonabreul, line 85: Clarify the purpose of the range parameter: the number of minimum price variations a bar consists of.

Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Checks if the current RangeBar being created has closed. If that is the case, it consolidates
+        /// the bar, resets the current RangeBar and stores the needed information from the last bar to
+        /// create the next bar
+        /// <see cref="UpdateBar(DateTime, decimal, decimal)"/>
+        /// </summary>
+        protected override void CheckIfBarIsClosed()
```
jhonabreul, line 145: Maybe the name could be `ConsolidateIfBarIsClosed`. Just to make things clearer. A `Check..` method might be thougth of as something returning a boolean.

Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// <param name="data">The new data for the bar</param>
+        protected override void CreateNewBar(IBaseData data)
+        {
+            var currentValue = Selector(data);
+            var volume = VolumeSelector(data);
+            var open = currentValue;
+
+            if (_firstTick)
+            {
+                _minimumPriceVariation = SymbolPropertiesDatabase.FromDataFolder().GetSymbolProperties(data.Symbol.ID.Market, data.Symbol, data.Symbol.ID.SecurityType, "USD").MinimumPriceVariation;
```
jhonabreul, line 166: We could have the symbol properties DB in a static property or method. Look at `OptionStrategies.SymbolPropertiesDatabase`.

Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// <param name="volume">Volume of the given data</param>
+        private void Rising(DateTime time, decimal currentValue, decimal volume)
+        {
+            while (currentValue - CurrentRangeBar.Low > RangeSize)
+            {
+                CurrentRangeBar.Update(time, currentValue, volume);
+                OnDataConsolidated(CurrentRangeBar);
+                CurrentRangeBar = new RangeBar(CurrentRangeBar.Symbol, CurrentRangeBar.EndTime, RangeSize, CurrentRangeBar.Close + _minimumPriceVariation, 0);
+            }
+            CurrentRangeBar = null;
```
jhonabreul, line 189: Just a thougth but, it could be possible to get by without `Rising` and `Falling`? 
The loop stopping condition might be `while (CurrentBar.IsClosed)` and the open price passed to `new RangeBar()` might be something like `isRising ? CurrentBar.High + _minimumPriceVariation : CurrentBar.Low - _minimumPriceVariation`. 

Then this code might be moved to `UpdateBar` and there would be no need for the 3 cases there, it would be abstracted into just one loop. I think it would handle the 3 cases: (1 and 2) rising avobe/falling bellow `2*range` (more than one bar will be formed) or (3) just one bar will be formed.

Common/Data/Consolidators/ClassicRangeConsolidator.cs
```diff
@@ -0,0 +1,77 @@
...
+*/
+
+using System;
+using QuantConnect.Data.Market;
+using Python.Runtime;
+
+namespace QuantConnect.Data.Consolidators
+{
+    /// <summary>
+    /// This consolidator can transform a stream of <see cref="IBaseData"/> instances into a stream of <see cref="RangeBar"/>
```
jhonabreul, line 23: Could add the difference between this and `RangeConsolidator` in the classes descriptions.

Algorithm.CSharp/ClassicRangeConsolidatorAlgorithm.cs
```diff
@@ -0,0 +1,42 @@
...
+using System;
+using QuantConnect.Data.Consolidators;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Example algorithm of how to use the ClassicRangeConsolidator
+    /// </summary>
+    public class ClassicRangeConsolidatorAlgorithm : RangeConsolidatorAlgorithm
```
jhonabreul, line 25: Could this one assert that when price variation is high enough, we don't get the intermediate bars?

Algorithm.Python/RangeConsolidatorAlgorithm.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.AddEquity("SPY")
+        rangeConsolidator = self.CreateRangeConsolidator()
+        rangeConsolidator.DataConsolidated += self.OnDataConsolidated
+        self.firstDataConsolidated = None;
+
+        self.SubscriptionManager.AddConsolidator("SPY", rangeConsolidator)
+
+    def OnEndOfAlgorithm(self):
+        if self.firstDataConsolidated == None:
```
jhonabreul, line 35: Could use `is None` instead of `== None` in Python as a general rule.

Algorithm.CSharp/RangeConsolidatorAlgorithm.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        protected virtual void OnDataConsolidated(Object sender, RangeBar rangeBar)
+        {
+            if (_firstDataConsolidated == null)
+            {
+                _firstDataConsolidated = rangeBar;
+            }
+
+            if (Math.Abs(rangeBar.Low - rangeBar.High) != 1m)
```
jhonabreul, line 61: `High - Low` wouldn't require `Abs` here, right? Same in Python

Algorithm.CSharp/RangeConsolidatorAlgorithm.cs
```diff
@@ -0,0 +1,123 @@
...
+    public class RangeConsolidatorAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        private RangeBar _firstDataConsolidated;
+        protected virtual Resolution UniversalResolution => Resolution.Daily;
+
+        public override void Initialize()
+         {
+            SetStartDate(2013, 10, 07);
+            SetEndDate(2013, 10, 11);
+            UniverseSettings.Resolution = UniversalResolution;
```
jhonabreul, line 36: Maybe pass the resolution to `AddEquity` instead?

Common/Data/Consolidators/ClassicRenkoConsolidator.cs
```diff
@@ -138,89 +129,40 @@ public ClassicRenkoConsolidator(decimal barSize, RenkoType type)
...
-                    }
-                }
-                else
-                {
-                    _volumeSelector = x => 0;
-                }
-            }
+            EpsilonCheck(barSize);
+            _barSize = barSize;
+            _evenBars = evenBars;
```
jhonabreul, line 136: Call `this()` instead of base and we can remove the duplicate body of this constructor

jhonabreul (changes requested):
Nice work! 🚀  Leaving some comments 👇

Common/Data/Consolidators/RangeConsolidator.cs
Marinovsky: Yeah, in the beginning I wanted to use that property instead of creating new field, but since `CurrentBar` is of type `IBaseData` It doesn't have `Volume` so I cannot use it in the inner overridden methods of `ClassicRangeConsolidator` and `RangeConsolidator`. On the other hand, though I could have used a private variable named `_currentBar`, it wouldn't have worked since I also needed to access that variable from a child class called `ClassicRangeConsolidator` and if it was protected, it's name should have been in PascalCase, therefore being equal to the property.

Algorithm.CSharp/ClassicRangeConsolidatorAlgorithm.cs
Marinovsky: It indeed does! In `OnDataConsolidated()` method, after calling the base implementation of this method it checks all the bars have non-zero volume since intermediate/phantom bars have zero volume. However, thanks for reminding me to write a comment in `RangeConsolidator.cs` explaining this property of intermediate/phantom RangeBar's 💯

Common/Data/Consolidators/ClassicRenkoConsolidator.cs
Marinovsky: Hmm but how that could be done if the other constructor receives selectors of type `Func<IBaseData, decimal>` and these ones are `PyObjects`?

jhonabreul: You're right 🚀

Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -115,39 +123,34 @@ public RangeConsolidator(decimal range)
...
-        protected override void CheckIfBarIsClosed()
-        {
-            if (CurrentRangeBar != null && CurrentRangeBar.IsClosed)
+            CurrentRangeBar.Update(time, currentValue, volume);
+            while (CurrentRangeBar.IsClosed)
             {
                 OnDataConsolidated(CurrentRangeBar);
-                CurrentRangeBar = null;
+                CurrentRangeBar = new RangeBar(CurrentRangeBar.Symbol, CurrentRangeBar.EndTime, RangeSize, isRising ? CurrentRangeBar.High + _minimumPriceVariation : CurrentRangeBar.Low - _minimumPriceVariation, 0);
+                CurrentRangeBar.Update(time, currentValue, Math.Abs(CurrentRangeBar.Low - currentValue) > RangeSize ? 0 : volume); // Intermediate/phantom RangeBar's have zero volume
```
jhonabreul, line 153: For the phantom bars volume, should this check depend on `isRising`? Something like: if true, `currentValue - CurrentBar.Low` else `CurrentBar.High - currentValue`.

Common/Data/Consolidators/ClassicRangeConsolidator.cs
```diff
@@ -20,22 +20,29 @@
 namespace QuantConnect.Data.Consolidators
 {
     /// <summary>
-    /// This consolidator can transform a stream of <see cref="IBaseData"/> instances into a stream of <see cref="RangeBar"/>
+    /// This consolidator can transform a stream of <see cref="IBaseData"/> instances into a stream of <see cref="RangeBar"/>.
+    /// The difference between this consolidator and <see cref="RangeConsolidator"/>, is that this last one creates intermediate/
+    /// phantom RangeBar's (RangeBar's with zero volume) if the price rises up or falls down by above/below two times the range
```
jhonabreul, line 25: I think phantom bars are created if price rises/falls above/below the RangeSize, right? Not 2*RangeSize

Tests/Common/Data/RangeConsolidatorTests.cs
```diff
@@ -74,14 +74,15 @@ protected virtual decimal[][] GetRangeConsolidatorExpectedValues()
...
-                    new decimal[]{ 89.5m, 89m, 90m, 90m, 20m},
-                    new decimal[]{ 90.5m, 90m, 91m, 91m, 20m},
-                    new decimal[]{ 91.5m, 90.5m, 91.5m, 90.5m, 10m},
-                    new decimal[]{ 90m, 90m, 91m, 91m, 20m},
+                    new decimal[]{ 90.46m, 89.46m, 90.46m, 89.46m, 10m},
+                    new decimal[]{ 89.45m, 89m, 90m, 90m, 10m},
+                    new decimal[]{ 90.01m, 90m, 91m, 91m, 20m},
+                    new decimal[]{ 91.01m, 90.5m, 91.5m, 90.5m, 10m},
+                    new decimal[]{ 90.49m, 90m, 91m, 91m, 20m},
                     new decimal[]{ 91.01m, 91.01m, 92.01m, 92.01m, 0m }
```
jhonabreul: Why did this changed?

jhonabreul (changes requested):
Thanks! Just leave a few other minor comments.

Common/Data/Consolidators/RangeConsolidator.cs
Marinovsky: Interestingly enough, it doesn't need, because just the bar is just created a line above, so the all its values `Open, High, Low` and `Close` are the same

Common/Data/Consolidators/ClassicRangeConsolidator.cs
Marinovsky: It looks like, but if the price rises above the `currentBar.Low + BarSize` but not above `currentBar.Low + 2 * BarSize`, it will only close one bar, and create another, but this last one won't be closed yet. In that case, both the bar that closed and the bar that was created are real bars. If the price would have risen above `currentBar.Low + 2 * BarSize` It woudln't have just closed one bar, but even created and closed another one, that last one bar is the one that would be phantom bar.

Tests/Common/Data/RangeConsolidatorTests.cs
Marinovsky: These stats changed because of the change that you proposed that would abstract much more the UpdateBar() method. Indeed, I noticed that on other platforms, their RangeBar charts that use intermediate/phantom bars never had a gap between bars, even if the gap was smaller than the size of the bar. With that change, this condition will indirectly be fulfilled, therefore keeping consistency with other platforms. I'm going to add the websites from which I found this information in the PR description 👌

jhonabreul (approved):

Common/Data/Consolidators/RangeConsolidator.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                return;
+            }
+
+            CurrentBar.Update(time, currentValue, volume);
+            while (CurrentBar.IsClosed)
+            {
+                OnDataConsolidated(CurrentBar);
+                CurrentBar = new RangeBar(CurrentBar.Symbol, CurrentBar.EndTime, RangeSize, isRising ? CurrentBar.High + _minimumPriceVariation : CurrentBar.Low - _minimumPriceVariation, 0);
+                CurrentBar.Update(time, currentValue, Math.Abs(CurrentBar.Low - currentValue) > RangeSize ? 0 : volume); // Intermediate/phantom RangeBar's have zero volume
```
Martin-Molinero, line 137: > Math.Abs(CurrentBar.Low - currentValue) > RangeSize ? 0 : volume

This seems fishy I think, why using the Low and not the High for example?

Martin-Molinero (commented):
Awesome! Almost there, leaving a couple of small requests👍

Common/Data/Consolidators/RangeConsolidator.cs
Marinovsky: Because they're the same since the bar has just been created one live above, therefore `High, Low` and `Close` properties have the same value as `Open`.

PR #7410, opened 2023-08-01
Enable Report generator to use custom HTML template

Marinovsky:
#### Description

- Remove `template.crisis.html` file. Now its content is in `template.html` as a comment
- Modify `CrisisReportElement.cs` and `ParametersReportElement.cs` to use a given template
- Enable Report program, to receive a custom HTML file to use as a template
- Modify `template.html` to contain parameters and crisis templates as a comment withtin
- Extract the parameters and crisis templates from HTML template using Regex
- Add unit tests to make sure regex works in different potential scenarios
#### Related Issue

N/A
#### Motivation and Context

With this changes, users will be able to customize the HTML template of their Reports with just one single HTML file

This change definitely requires an update to documentation to brief the users on how the format of the custom HTML must be

It was asserted the Report generator worked as expected in the following scenarios:
- When no custom HTML template file is given
- When a custom HTML template file is given, but it doesn't exist
- When a custom HTML template file is given and it actually exists
- When a CSS override file and a custom HTML template file are given
- When just a CSS override file is given but not a custom HTML template file
- When a custom HTML template file is given but not a CSS override file
- When neither a custom HTML template file nor a CSS override file are given
- Regex success and fails in the expected scenarios
#### Types of changes

#### Checklist:


Martin-Molinero (approved):
Awesome 🚀

PR #7423, opened 2023-08-07
Fix Overflow bug in RandomDataGenerator

### LINKED ISSUE #7400: Random Data Generator OverflowException
#### Expected Behavior

- No unhandled exception is thrown
#### Actual Behavior

```
Unhandled exception. System.OverflowException: Value was either too large or too small for an Int64.
   at System.Decimal.ToInt64(Decimal d)
   at QuantConnect.Util.LeanData.Scale(Decimal value) in 
/LeanCloud/CI.Builder/bin/Debug/src/QuantConnect/Lean/Common/Util/LeanData.cs:line 914
   at QuantConnect.Util.LeanData.GenerateLine(IBaseData data, SecurityType securityType, Resolution 
resolution) in /LeanCloud/CI.Builder/bin/Debug/src/QuantConnect/Lean/Common/Util/LeanData.cs:line 470
   at QuantConnect.Data.LeanDataWriter.Write(IEnumerable1 source) in 
/LeanCloud/CI.Builder/bin/Debug/src/QuantConnect/Lean/Common/Data/LeanDataWriter.cs:line 161
   at QuantConnect.ToolBox.RandomDataGenerator.RandomDataGenerator.Run() in 
/LeanCloud/CI.Builder/bin/Debug/src/QuantConnect/Lean/ToolBox/RandomDataGenerator/RandomDataGenerator.cs:l
ine 260
```
#### Potential Solution

- Review why/when these invalid data points are generated, so we can avoid them
#### Reproducing the Problem

```sh
lean data generate --start 20180101 --symbol-count 1000
```
#### System Information

N/A
#### Checklist

Marinovsky:
#### Description

The bug was raised because of two main reasons:
1. The prices generated by `RandomValueGenerator.cs` were not upper bounded
2. The `DividendSplitMapGenerator.FinalSplitFactor` wasn't lower bounded
Therefore, if the price generated was large and the `FinalSplitFactor` was really small, the price generated divided by that small `FinalSplitFactor` turned out to be really large, even larger than the maximum `long` value

To fix this issue, the prices generated were upper bounded by 1000000 the same as the `FinalSplitFactor`, which was bounded between 0.001 and 1, so that all the values generated would be between the minimum `MinumumPriceMovement` for all the equities and 100000000. In order to bound the `FinalSplitFactor` we found that `FinalSplitFactor` was calculated with this formula
$$\text{Final Split Factor} := \Pi_{i=1}^{n} (psf_{i}),$$
where $psf_{i}$ is the value of `previousSplitFactor` in the $i$-th split month. It was found that, in the worst case, $n=2m$, where $m$ is the number of months from the start date to the end date (The range for which the data should be generated), since a split could occur each month, and sometimes, there was $2$ splits per month. Therefore, the formula for the `FinalSplitFactor` in the worst of the cases was
$$\text{Final Split Factor} := \Pi_{i=1}^{2n} (psf_{i}).$$
Now, if we wanted to guarantee that `FinalSplitFactor` was never going to be smaller than 0.001 we needed to ensure that $$0.001^{\frac{1}{2m}} < psf_{i}\quad i=1,\dots,2m,$$
since, once we have this inequality, if we multiply all of them we have that
$$0.001 < \Pi_{i=1}^{2m} psf_{i}.$$
On the other hand, we also needed to ensure that `FinalSplitFactor` was bounded by $1$, so that if the given price was really small, after being divided by the `FinalSplitFactor`, the price wasn't even smaller. With a similar reasoning, we concluded that if
each $psf_{i}$ was smaller than 1, the `FinalSplitFactor` would be also smaller than 1.

After bounding both the prices and the `FinalSplitFactor`, we generated data for 1 year, 6 months and 3 months for 1000 equity symbols. Here are some of the factor files of those symbols containing the split factor and the split dates:
imagen
imagen
#### Related Issue

Closes #7400 
#### Motivation and Context

With this change the users will be able to generate random data without errors

N/A

There was created different unit test asserting everything worked as expected
#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/ToolBox/RandomDataGenerator/DividendSplitMapGeneratorTests.cs
+++ b/Tests/ToolBox/RandomDataGenerator/DividendSplitMapGeneratorTests.cs
@@ -1,0 +1,74 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using NUnit.Framework;
+using QuantConnect.ToolBox.RandomDataGenerator;
+using System;
+
+namespace QuantConnect.Tests.ToolBox.RandomDataGenerator
+{
+    [TestFixture]
+    public class DividendSplitMapGeneratorTests
+    {
+        [TestCase(240, 0.9857119009006162)]
+        [TestCase(120, 0.9716279515771061)]
+        [TestCase(60, 0.9440608762859234)]
+        [TestCase(12, 0.7498942093324559)]
+        [TestCase(6, 0.5623413251903491)]
+        [TestCase(3, 0.31622776601683794)]
+        [TestCase(1, 0.03162277660168379)]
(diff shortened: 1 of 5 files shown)


Martin-Molinero (approved):
Thank you! 🚀

PR #7426, opened 2023-08-09
Change default American Option Pricing model to Binomial CRR model

### LINKED ISSUE #7408: Default option pricing model does not work on American put options
#### Expected Behavior
Theoretical Price and Greeks should be accurate without need of specifying the option pricing model

#### Actual Behavior
Greeks are inaccurate for American put options using the default `BjerksundStensland` model, as the 1993 version cannot handle discrete dividends.

#### Potential Solution
Change the default model to another one that supports the function.

#### Reproducing the Problem
The put option delta values are incorrect in the following backtest:
https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_8c215227f6ef63ae8e93ca520952d3d9.html

#### System Information
NA

#### Checklist

LouisSzeto:
#### Description
Change default American Option Pricing model to Binomial CRR model, as it is the most accurate and computational efficient model when compared to IB's modeled Greeks. Same has been compared for European options to select BSM.

#### Related Issue
closes #7408

#### Motivation and Context
The current default model `BjerksundStensland` is not working to calculate American put's delta

Previous unit test values were generated, but now it is fetched from official IB 

https://www.quantconnect.com/docs/v2/writing-algorithms/reality-modeling/options-models/pricing#06-Default-Behavior

- Unit test for error margin comparison
- regression test on the necessity of model change

#### Types of changes

- [X] Performance (non-breaking change which improves performance. Please add associated performance test and results)

#### Checklist:


Algorithm.CSharp/OptionGreeksRegressionAlgorithm.cs
```diff
@@ -0,0 +1,148 @@
...
+using System;
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm reproducing issue #5610
```
Martin-Molinero, line 25: - Comment is off I think?
- `error = 0.05m;` think this should be a %

Algorithm.CSharp/OptionGreeksRegressionAlgorithm.cs
```diff
@@ -0,0 +1,148 @@
...
+                    else if (symbol == _itmPutSymbol)
+                    {
+                        expected = -0.18395m;
+                    }
+                    else
+                    {
+                        expected = -0.99989m;
+                    }
+
+                    if (delta >= expected + error || delta <= expected - error)
```
Martin-Molinero, line 87: You are not really asserting all the expected symbols were in the chain, think it's worth adding 👍

Data/option/usa/minute/spy/20230803_quote_american.zip
Martin-Molinero, line 1: These new data files are too big, we should delete the option contracts we don't need and just leave the 4 we are using, some for the trade file below👍

Tests/Common/Securities/OptionPriceModelTests.cs
```diff
@@ -425,38 +427,129 @@ public void ThrowsIfOptionStyleIsNotSupportedByQLPricingModel(string qlModelName
...
             var greeks = results.Greeks;
 
             // Expect minor error due to interest rate and dividend yield used in IB
-            Assert.AreEqual((double)greeks.Delta, (double)ibDelta, 0.005);
-            Assert.AreEqual((double)greeks.Gamma, (double)ibGamma, 0.005);
-            Assert.AreEqual((double)greeks.Theta / 365.25, (double)ibTheta, 0.005);
-            Assert.AreEqual((double)greeks.Vega, (double)ibVega, 0.005);
-            Assert.AreEqual((double)greeks.Rho, (double)ibRho, 0.005);
+            Assert.AreEqual((double)greeks.Delta, (double)ibDelta, 0.07);
+            Assert.AreEqual((double)greeks.Gamma, (double)ibGamma, 0.07);
```
Martin-Molinero, line 457: 1- Seems to be worse in this case?
2- Shouldn't we still compare `Rho/Vega/etc`, why delete them?
3- Think the accepted delta should be a % from the value itself, so it's relative? Same for the unit test below

Martin-Molinero (commented):
Thank you! Leaving some minor comments 👍

Tests/Common/Securities/OptionPriceModelTests.cs
LouisSzeto: 1. The previous data is generated from a spreadsheet, the current ones are really from IB
2. The Rho/Vega is not available when I fetch these data
3. The greeks themselves are already derivatives/relatives to the underlying (like %), so although they look like comparing in absolute values, they are actually relative error!

Tests/Common/Securities/OptionPriceModelTests.cs
```diff
@@ -425,38 +427,129 @@ public void ThrowsIfOptionStyleIsNotSupportedByQLPricingModel(string qlModelName
...
+                    // Check the option Greeks are valid
+                    var greeks = results.Greeks;
+
+                    // Expect minor error due to interest rate, bid/ask price and dividend yield used in IB
+                    // And approximation error using Black Calculator if the original pricing model fails
+                    Assert.AreEqual((double)greeks.Delta, Parse.Double(datum[3]), delta);
+                    Assert.AreEqual((double)greeks.Gamma, Parse.Double(datum[4]), delta);
+                    // We do not test Vega as the tangent line of price change over volatility is steep for deep ITM/OTM
+                    // We do not test Theta as the tangent line of price change over time change is steep for near-expiry options
+                    // No model will be estimating these quantities well
```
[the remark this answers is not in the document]
LouisSzeto: Same reason above 😄

Algorithm.CSharp/OptionGreeksRegressionAlgorithm.cs
LouisSzeto: - Yup, thanks for pointing out 👍 
- Refer to comment below

Martin-Molinero (approved):
Thank you! 👍

PR #7436, opened 2023-08-18
Approximate order price for options when using IB Fee model

Marinovsky:
#### Description

So far IB Fee model was taking into account the order price for computing the fees when the security was an option. However, just IB Fee model worked in that fashion, none of the other Fee models used the order price. The problem with this is that sometimes the order given to the fee model hadn't been updated so its price was 0, and therefore the commission rate was always 0.25.

To solve this, we modify the IB Fee model to be more clever. By approximating the order price with the `Security.Price`, `Security.Bid/AskPrice`, `StopPrice` or `LimitPrice`, depending on the order type, IB Fee model can now give an estimation of the fees to be charged for each order.

As always, to cover this changes unit tests were made and the already existent regression tests were updated. Finally, it's worth mentioning this PR is related with #7432 since it was discovered when solving the issue that opened the last mentioned PR.
#### Related Issue

N/A
#### Motivation and Context

With this change users will have an estimation of the fees charged for each order with options using IB brokerage

N/A

There was already a unit test that asserted the fees charged for orders with options were correct. That test was modified to assert the fees returned were the expected ones with different order prices and types.
#### Types of changes

#### Checklist:


Martin-Molinero (approved):
Thank you!

PR #7456, opened 2023-09-06
`Insight` and `PortfolioTarget` tag

### LINKED ISSUE #6346: Ability to tag `Insight` and `PortfolioTarget` similar to `Order.Tag`
#### Expected Behavior

We can add a tag to an insight or portfolio target similar to `Order.Tag`. 

#### Actual Behavior

Not supported.

#### Potential Solution

Implement the property `Tag` for insights and portfolio targets so that we can access it via `Insight.Tag` and `PortfolioTarget.Tag`. 

#### Reproducing the Problem

Try adding a tag to an insight or portfolio target, for example: ` insight.Tag = "Hello World!"`. 
This throws an error (*unknown property "Tag"*). 
    

#### Motivation and Context
Tagging can provide additional information and help us to better understand the processes of an algorithm. 
In order to comply with the separation of concerns principle, we should develop the framework modules in a way that makes them independent from each other. Although this is powerful principle, it makes debugging/logging much more difficult. The use of tags represents a good compromise in my opinion as it does not violate the separation of concerns principle (we don't have to use tags for decision making) and at the same time it can provide additional information. 

Here is an application scenario: 
Let's consider an `AlphaModel` that emit insights based an moving average crossovers + trailing stop. This means there can be more than one reason why an insight direction changed. It can be because of a moving average crossover or because of the trailing stop. We could  add the tag `insight.Tag = "Trailing stop triggered!"` and pass it on to the execution model and there we could add it as a classical tag to the order and see it in the order history once the backtest finished. 

The same goes for portfolio targets. `PortfolioTarget.Tag` could include information on why the target has changed (could be rebalancing, or insight direction change or something else). 

   
    

#### System Information

QC Cloud 

#### Checklist

jhonabreul:
#### Description

This adds a new `Tag` property to `Insight` and `IPortfolioTarget` so additional information can be carried on by these classes, similar to order's tags.

#### Related Issue

Closes #6346 

Regression algorithms.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
+++ b/Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
@@ -45,14 +45,21 @@ namespace QuantConnect.Algorithm.Framework.Portfolio
         public decimal Quantity { get; }
 
         /// <summary>
+        /// Portfolio target tag with additional information
+        /// </summary>
+        public string Tag { get; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="PortfolioTarget"/> class
         /// </summary>
         /// <param name="symbol">The symbol this target is for</param>
         /// <param name="quantity">The target quantity</param>
-        public PortfolioTarget(Symbol symbol, decimal quantity)
+        /// <param name="tag">The target tag with additional information</param>
+        public PortfolioTarget(Symbol symbol, decimal quantity, string tag = "")
         {
             Symbol = symbol;
             Quantity = quantity;
+            Tag = tag;
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


Common/Algorithm/Framework/Portfolio/PortfolioTarget.cs
```diff
@@ -67,6 +74,19 @@ public static IPortfolioTarget Percent(IAlgorithm algorithm, Symbol symbol, doub
...
 
+        /// <summary>
+        /// Creates a new target for the specified percent
+        /// </summary>
+        /// <param name="algorithm">The algorithm instance, used for getting total portfolio value and current security price</param>
+        /// <param name="symbol">The symbol the target is for</param>
+        /// <param name="percent">The requested target percent of total portfolio value</param>
+        /// <param name="tag">The target tag with additional information</param>
+        /// <returns>A portfolio target for the specified symbol/percent</returns>
+        public static IPortfolioTarget Percent(IAlgorithm algorithm, Symbol symbol, double percent, string tag)
```
Martin-Molinero, line 85: Wondering if we can directly just add `, string tag = ""` to the constructor above like in the other cases?

Martin-Molinero (approved):
Nice! Minor comment 👍

PR #7475, opened 2023-09-20
Solve bug in Tick Consolidator for Quote and Trade Tick types

### LINKED ISSUE #7450: Tick Consolidator not working as expected with TickType.Trade when no quote data is available
#### Expected Behavior

In the example below `OnTick` should be called even if there is no quote tick data available.

#### Actual Behavior

`OnTick` is not being called as the consolidator subscribed to the quote tick subscription even when the tick type is specified. 

#### Potential Solution

The provided tick type is currently ignored when the consolidator is being added to the subscription and the default tick type is being used. The issue can be fixed by only using the default tick type when it's not manually specified. 

https://github.com/QuantConnect/Lean/blob/76c0ab01ee7ad5a2a67e122a8f54ddb14a7ac498/Common/Data/SubscriptionManager.cs#L277-L281https://github.com/QuantConnect/Lean/blob/76c0ab01ee7ad5a2a67e122a8f54ddb14a7ac498/Common/Data/SubscriptionManager.cs#L164-L168
#### Reproducing the Problem

Running the following backtest with only trade tick data available.
``` csharp
 public class Example : QCAlgorithm
  {
      {
          SetStartDate(2023, 08, 01); 
          SetEndDate(2023, 08, 02); 

          var crypto = AddCrypto("BTCUSDT", Resolution.Tick, Market.Binance);
          var consolidator = Consolidate(crypto.Symbol, Resolution.Tick, TickType.Trade, (Tick tick) =>OnTick(tick));
      }

      private void OnTick(Tick tick)
      {
          //This will never be called when there is no quote tick data available even though TickType.Trade was selected
      }
  }
```
#### System Information

Not relevant

#### Checklist

Marinovsky:
#### Description

So far, when the user tried to consolidate tick data with a desired tick type, if the symbol was a Cfd, Forex or Crypto, by default it always picked a subscription with Quote tick type and thus ignored the desired tick type given by the user. Now, if the user has provided a desired tick type, LEAN will try to look for a subscription with that tick type. If not found, an exception will be thrown. To cover these changes, unit tests and regression tests have been made to assert the bug is actually solved by theses changes.
#### Related Issue

Closes #7450 
#### Motivation and Context

With this change users will be able to consolidate tick data with a desired tick type

N/A

For the unit tests, it was asserted that the method `SubscriptionManager.IsSubscriptionValidForConsolidator()` returned the expected results in different scenarios that involve different resolutions, tick types, and output types.

For the regression tests, an algorithm was created that consolidated tick data with two requested tick types, namely Trade and Quote. Every time a new tick data was found, it asserted if the tick type was Quote or Trade according to the method in which the new tick data was found. Finally, at the end of the algorithm it was asserted that at least one tick data of each type was found, if not an exception was thrown.
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CorrectConsolidatedBarTypeForTickTypesAlgorithm.cs
+++ b/Algorithm.CSharp/CorrectConsolidatedBarTypeForTickTypesAlgorithm.cs
@@ -29,7 +29,6 @@ namespace QuantConnect.Algorithm.CSharp
     {
         private bool _quoteTickConsolidatorCalled;
         private bool _tradeTickConsolidatorCalled;
-        private bool _openInterestTickConsolidatorCalled;
 
         public override void Initialize()
         {
@@ -40,7 +39,6 @@ namespace QuantConnect.Algorithm.CSharp
 
             Consolidate<QuoteBar>(symbol, TimeSpan.FromMinutes(1), TickType.Quote, QuoteTickConsolidationHandler);
             Consolidate<TradeBar>(symbol, TimeSpan.FromMinutes(1), TickType.Trade, TradeTickConsolidationHandler);
-            Consolidate<TradeBar>(symbol, TimeSpan.FromMinutes(1), TickType.OpenInterest, OpenInterestTickConsolidationHandler);
         }
 
         public override void OnData(Slice slice)
@@ -62,11 +60,6 @@ namespace QuantConnect.Algorithm.CSharp
             {
                 throw new Exception("TradeTickConsolidationHandler was not called");
             }
-
-            if (!_openInterestTickConsolidatorCalled)
-            {
-                throw new Exception("OpenInterestTickConsolidationHandler was not called");
-            }
         }
 
         private void QuoteTickConsolidationHandler(QuoteBar consolidatedBar)
@@ -77,11 +70,6 @@ namespace QuantConnect.Algorithm.CSharp
         private void TradeTickConsolidationHandler(TradeBar consolidatedBar)
         {
             _tradeTickConsolidatorCalled = true;
-        }
-
-        private void OpenInterestTickConsolidationHandler(TradeBar consolidatedBar)
-        {
-            _openInterestTickConsolidatorCalled = true;
         }
 
         /// <summary>
--- a/Common/Data/SubscriptionManager.cs
+++ b/Common/Data/SubscriptionManager.cs
@@ -142,10 +142,11 @@ namespace QuantConnect.Data
 
 
         /// <summary>
-        ///     Add a consolidator for the symbol
+        /// Add a consolidator for the symbol
         /// </summary>
         /// <param name="symbol">Symbol of the asset to consolidate</param>
         /// <param name="consolidator">The consolidator</param>
+        /// <param name="tickType">Desired tick type of the subscription</param>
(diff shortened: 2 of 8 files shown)


Common/Data/SubscriptionManager.cs
```diff
@@ -268,17 +275,25 @@ public void SetDataManager(IAlgorithmSubscriptionManager subscriptionManager)
...
+                if (desiredTickType == null)
+                {
+                    var tickType = LeanData.GetCommonTickTypeForCommonDataTypes(
                     consolidator.OutputType,
                     subscription.Symbol.SecurityType);
 
-                return subscription.TickType == tickType;
+                    return subscription.TickType == tickType;
+                }
+                else if (subscription.TickType != desiredTickType)
```
jhonabreul, line 293: If subscription's tick type is the same as the desired tick type, we coudl return true, we don't really need the method to continue doing any more checks outside of this `if`.
Maybe do this in a single check? Like doing `var tickType = desiredTickType ?? LeanData.GetCommonTickTypeForCommonDataTypes(consolidator.OutputType, subscription.Symbol.SecurityType);` and then just `return subscription.TickType == tickType;`?

Tests/Common/Data/SubscriptionManagerTests.cs
```diff
@@ -265,6 +265,147 @@ public void GetsCustomSubscriptionDataTypes()
             Assert.AreEqual(expected, SubscriptionManager.IsSubscriptionValidForConsolidator(subscription, consolidator));
         }
 
+        [TestCase(SecurityType.Future, Resolution.Tick, typeof(Tick), TickType.Trade, null, typeof(TradeBar), true)]
```
jhonabreul, line 268: Since resolution and dataType are always `Resolution.Tick` and `typeof(Tick)`, they could be removed from the test parameters

Algorithm.CSharp/CorrectConsolidatedBarTypeForTickTypesAlgorithm.cs
```diff
@@ -29,7 +29,6 @@ public class CorrectConsolidatedBarTypeForTickTypesAlgorithm : QCAlgorithm, IReg
     {
         private bool _quoteTickConsolidatorCalled;
         private bool _tradeTickConsolidatorCalled;
-        private bool _openInterestTickConsolidatorCalled;
```
jhonabreul, line 32: Why was the open interest tick consolitator code removed here and in the Python algo?

jhonabreul (commented):
Nice! 💯 Leaving a couple of minor comments

Common/Data/SubscriptionManager.cs
Marinovsky: Hmm I see, that could be an option. However, I still wonder why don't you take into consideration this last check `consolidator.InputType.IsAssignableFrom(subscription.Type)`

Algorithm.CSharp/CorrectConsolidatedBarTypeForTickTypesAlgorithm.cs
Marinovsky: Oh because Open Interest is just for futures and options, not for equities

Martin-Molinero (approved):
Nice! Thank you 🚀

PR #7498, opened 2023-10-03
Adds `SimpleCustomFillModel` to `CustomModelsAlgorithm`

AlexCatarino:
#### Description
The simple fill model shows how to implement a simpler version of the most popular order fills: Market, Stop Market, and Limit.

This model was tested on QuantConnect Cloud and will serve as an additional example since we don't have an example that does not reuse the base class method.

#### Related Issue
N/A

#### Motivation and Context
Documentation/Examples

No. We already point to `CustomModelsAlgorithm`.

QuantConnect Cloud.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Python/CustomModelsAlgorithm.py
+++ b/Algorithm.Python/CustomModelsAlgorithm.py
@@ -134,6 +134,14 @@ class SimpleCustomFillModel(FillModel):
         fill.FillPrice = fill_price
         return fill
 
+    def _get_trade_bar(self, asset, orderDirection):
+        trade_bar = asset.Cache.GetData[TradeBar]()
+        if trade_bar: return trade_bar
+
+        # Tick-resolution data doesn't have TradeBar, use the asset price
+        price = asset.Price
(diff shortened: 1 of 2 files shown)


Algorithm.Python/CustomModelsAlgorithm.py
```diff
@@ -117,3 +117,56 @@ def HasSufficientBuyingPowerForOrder(self, parameters):
...
+            return self._set_order_event_to_filled(fill, stop_price, order.Quantity)
+
+        return fill
+
+    def LimitFill(self, asset, order):
+        fill = self._create_order_event(asset, order)
+        if order.Status == OrderStatus.Canceled: return fill
+        limit_price = order.LimitPrice
+
+        trade_bar = asset.Cache.GetData[TradeBar]()
```
Martin-Molinero, line 165: Suggest adding checks in case trade_bar is None/null but this won't work for tick resolution. Should we maybe just rely on `SecurityCache.BidPrice/AskPrice` for this simple version?

Martin-Molinero (approved):
Nice! 👍

PR #7500, opened 2023-10-04
Fix bug with custom data subscription with the same symbol from multiple exchanges

### LINKED ISSUE #7495: Custom data subscription not working with the same symbol from multiple exchanges
#### Expected Behavior

All the custom data subscriptions are properly working and 4 data entries are available in each slice (see repro example)

#### Actual Behavior

Custom data subscriptions for the same symbol from multiple exchanges are not working as expected. The custom data subscription is not created for every symbol but only the first one and is being reused by the others. 

#### Potential Solution

I don't have enough information on backward compatibility and the initial design reasons to make a proper suggestion to fix it. 
Not using the US market but the same market as in the source symbol seems to work, but this might have further implications I'm not aware of. Especially after reading https://github.com/QuantConnect/Lean/pull/3484

#### Reproducing the Problem

```csharp
public class Test : QCAlgorithm
{
    {
        SetStartDate(2023, 01, 1); // Set Start Date
        SetEndDate(new DateTime(2023, 01, 01).AddMinutes(2)); // Set Start Date
        SetAccountCurrency("USDT", 10000);

        var market1 = AddCryptoFuture("BTCUSDT", Resolution.Minute, Market.Bybit, false);
        var data1 = AddData<MyCustomData>(market1.Symbol, Resolution.Minute, DateTimeZone.Utc, false);

        var market2 = AddCryptoFuture("BTCUSDT", Resolution.Minute, Market.Binance, fillForward: false);
        var data2 = AddData<MyCustomData>(market2.Symbol, Resolution.Minute, DateTimeZone.Utc, false);
    }


    {
        // I'm expecting 4 data entries here but it's always only 3
        // It seems like the symbol being created is the same for both custom data subscriptions as it's created for the US market
        // Registering the data source using the ticker instead of the underlying symbol works but that way we're loosing the market information 
        // in the BaseData making reading the data more complicated. 
        // Another solution would be to use a different data type per symbol as the data type because the final symbol consists of {underlying_ticker}.{data_type_name}

        Log($"Data: {slice.Count} Expected: 4");
    }
}
```
#### System Information

N/A
#### Checklist

Marinovsky:
#### Description

Before these changes, a user could not have custom data subscriptions with the same symbol but from different exchanges, since all the custom data subscriptions had as market `Market.USA`. With these changes, the custom data subscriptions will now have the market of the underlying security.
#### Related Issue

Closes #7495 
#### Motivation and Context

With this change, users will be able to have multiple custom data subscriptions from different exchanges

Since no regression test failed after these changes, there's no need to update the documentation

There was created a regression algorithm which created 4 subscriptions, were two of them were custom ones, but each of them had a different underlying. Every time an slice was received, it was asserted it contained values for all the subscriptions. Finally, at the end of the algorithm it was asserted at least one data point had been received.
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/CustomDataWorksWithDifferentExchangesRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/CustomDataWorksWithDifferentExchangesRegressionAlgorithm.cs
@@ -1,0 +1,145 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using System;
+using System.Collections.Generic;
+
+namespace QuantConnect.Algorithm.CSharp
+{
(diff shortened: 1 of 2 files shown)


Martin-Molinero (approved):
Thanks! 🚀

Issue #6983, opened 2023-02-20
Implements Correlation as Lean Indicator


AlexCatarino:
#### Expected Behavior
Lean supports Correlation as a Lean Indicator which takes a Symbol object, a benchmark Symbol, and a look-back period. 

#### Actual Behavior
Not supported. See #83

#### Potential Solution
Implement this indicator. 

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

LouisSzeto:
[spy_qqq_corr.csv](https://github.com/QuantConnect/Lean/files/10822689/spy_qqq_corr.csv)

LouisSzeto:
[spy_qqq_corr.csv](https://github.com/QuantConnect/Lean/files/13114332/spy_qqq_corr.csv)
columns: "SPY close" "QQQ close" "annual correlation"

script:
```
import talib

qb = QuantBook()

spy = qb.AddEquity("SPY", Resolution.Daily).Symbol
qqq = qb.AddEquity("QQQ", Resolution.Daily).Symbol

history = qb.History([spy, qqq], datetime(1998, 1, 1), datetime(2023, 2, 23))

close = history.loc[spy].close

# Assumes no risk free rate
history2 = pd.concat([close, history.loc[qqq].close], axis=1).dropna()
history2.columns = [spy, qqq]

corr = history2.rolling(252).corr().unstack(1)[spy][qqq].dropna().to_frame()
corr.columns = ["corr"]
corr["spy close"] = close
corr["qqq close"] = history2.iloc[:, 1]
corr = corr.iloc[:, [1, 2, 0]].dropna()
corr.to_csv("spy_qqq_corr.csv", header=False)
```

PR #7540, opened 2023-10-27
Beta Indicator Supports QuoteBar

AlexCatarino:
#### Description
`Beta` indicator supports `QuoteBar`.
`Beta` must support `QuoteBar` from Forex/CDF and `QuoteBarConsolidator`.

#### Related Issue
Ref.: #6985 #7538

#### Motivation and Context
The `TradeBar` indicator type imposes limitations that we don't need, since `Beta` doesn't require the `Volume`.

#### Requires Documentation Change
Auto-gen.

#### How Has This Been Tested?
New unit test.

#### Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #7543, opened 2023-10-28
correlation as lean indicator #6983

grzegorzwodnicki:
#### Description
Added Correlation indicator

#### Related Issue
#6983 

#### Motivation and Context
This change allows users to calculate the correlation between 2 tickers.

This change requires a documentation update with an indicator description.

I wrote the test and compared with data calculated by 3rd part sofware.

#### Types of changes

#### Checklist:


jaredbroad:
Thank you for the PR @grzegorzwodnicki ! Please update to use the test data provided in the original linked issue; [spy_qqq_corr.csv](https://github.com/QuantConnect/Lean/files/10822689/spy_qqq_corr.csv). This also includes the code Louis used to generate the correlations. 

He does not have both types of correlation but that's fine to include your test data as well for the second type.

grzegorzwodnicki:
Thank you, @jaredbroad, for your guidance!

I've noticed that the original file, spy_qqq_corr.csv, was manipulated using dropna() which, as you pointed out, removes records with NaN values. In our case, it's crucial to maintain the first 252 records intact for accurate correlation calculations.

To align with this requirement, I’ll proceed to regenerate the CSV file. This will ensure that we have a complete set of data, preserving the initial 252 records for calculating the first correlation value accurately.

I will use the same source of data and the same code in Python with the talib library.

grzegorzwodnicki:
Just commited spy_qqq test and also Spearman method test.
To generate data I used below code:
```
import pandas as pd
history_spy = pd.read_csv("https://github.com/QuantConnect/Lean/raw/master/Data/equity/usa/daily/aapl.zip",
                          index_col=0, names=["Open_spy", "High_spy", "Low_spy", "Close_spy", "Volume_spy"], compression='zip')
history_qqq = pd.read_csv("https://github.com/QuantConnect/Lean/raw/master/Data/equity/usa/daily/qqq.zip",
                          index_col=0, names=["Open_qqq", "High_qqq", "Low_qqq", "Close_qqq", "Volume_qqq"], compression='zip')

history2 = pd.concat([history_spy, history_qqq], axis=1).dropna()

history2['corr'] = (history2.rolling(window=252).corr().unstack(1)["Close_spy"]["Close_qqq"])

history2['corr'] = history2['corr'].fillna(0)

result_data = []
for date, row in history2.iterrows():
    result_data.append({
        "Symbol": "SPY",
        "Date": date,
        "Open": row["Open_spy"],
        "High": row["High_spy"],
        "Low": row["Low_spy"],
        "Close": row["Close_spy"],
        "Volume": row["Volume_spy"],
        "Correlation": ""
    })
    result_data.append({
        "Symbol": "QQQ",
        "Date": date,
        "Open": row["Open_qqq"],
        "High": row["High_qqq"],
        "Low": row["Low_qqq"],
        "Close": row["Close_qqq"],
        "Volume": row["Volume_qqq"],
        "Correlation": row["corr"]
    })
   
    
   

result = pd.DataFrame(result_data)

result['Date'] = pd.to_datetime(result['Date']).dt.strftime('%Y%m%d')

result.to_csv("spy_qqq_corr.csv", index=False)


```

Indicators/CorrelationIndicator.cs
```diff
@@ -0,0 +1,247 @@
...
+           
+
+            if (_correlationType == CorrelationIndicatorType.Pearson)
+            {
+                _corr = Correlation.Pearson(_targetDataPoints.Select(d=>(double)d).ToList(), _referenceDataPoints.Select(d => (double)d).ToList());
+            }
+            if (_correlationType == CorrelationIndicatorType.Spearman)
+            {
+                _corr = Correlation.Spearman(_targetDataPoints.Select(d => (double)d).ToList(), _referenceDataPoints.Select(d => (double)d).ToList());
+            }
```
jhonabreul, line 225: `Correlation.Pearson` and `Correlation.Spearman` take `IEnumerable`s, so you can drop the list conversion to avoid multiple enumerations. Could just use `Linq`'s `Cast<double>()`

Tests/Indicators/CorrelationInidicatorTests.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            var indicator = new CorrelationIndicator("testCorrelationIndicator", Symbols.AAPL, Symbols.SPX, 3);
+
+            for (int i = 0; i < 3; i++)
+            {
+                indicator.Update(new TradeBar() { Symbol = Symbols.AAPL, Low = 1, High = 2, Volume = 100, Close = i + 1, Time = _reference.AddDays(1 + i) });
+                indicator.Update(new TradeBar() { Symbol = Symbols.SPX, Low = 1, High = 2, Volume = 100, Close = i + 2, Time = _reference.AddDays(1 + i) });
+            }
+            Console.WriteLine(indicator.Current.Value);
+            Assert.AreNotEqual(0, (double)indicator.Current.Value);
```
jhonabreul, line 179: - Use Lean's logger
- Use `AreEqual` with the expected value

Indicators/CorrelationIndicator.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <summary>
+        /// Period required for calcualte correlation
+        /// </summary>
+        private decimal _period;
+        /// <summary>
+        /// Correlation type
+        /// </summary>
+        private readonly CorrelationIndicatorType _correlationType;
+
+
```
jhonabreul, line 63: nit - additional blank line

jhonabreul (changes requested):
Thank you for your contribution! Leaving a few requests...

Indicators/CorrelationIndicator.cs
grzegorzwodnicki: _corr must be double, because decimal do not support NaN value. Correlation can return NaN and it is there a risk of throwing exception.

Indicators/CorrelationIndicator.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// </summary>
+        private void ComputeCorrelation()
+        {
+
+            if (_targetDataPoints.Count < _period || _referenceDataPoints.Count < _period)
+            {
+                _correlation = 0;
+                return;
+            }
+            double _corr = 0;
```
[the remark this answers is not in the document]
grzegorzwodnicki: _corr must be double, because decimal do not support NaN value. Correlation can return NaN and it is there a risk of throwing exception.

Tests/Indicators/CorrelationIndicatorSpearmanTests.cs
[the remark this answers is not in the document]
grzegorzwodnicki: I added also _correlationType to the class, and in the inherited class just changed type and all tests are done with different method.

grzegorzwodnicki:
New code used for generating testing data:
```
import pandas as pd

def rolling_spearman(x, y):
    return x.corr(y, method='spearman')


history_spy = pd.read_csv("https://github.com/QuantConnect/Lean/raw/master/Data/equity/usa/daily/aapl.zip",
                          index_col=0, names=["Open_spy", "High_spy", "Low_spy", "Close_spy", "Volume_spy"], compression='zip')
history_qqq = pd.read_csv("https://github.com/QuantConnect/Lean/raw/master/Data/equity/usa/daily/qqq.zip",
                          index_col=0, names=["Open_qqq", "High_qqq", "Low_qqq", "Close_qqq", "Volume_qqq"], compression='zip')


history2 = pd.concat([history_spy, history_qqq], axis=1).dropna()



history2['corr_pearson'] = history2['Close_spy'].rolling(window=252).corr(history2['Close_qqq'])
history2['corr_spearman'] = history2['Close_spy'].rolling(window=252).apply(lambda x: rolling_spearman(x, history2['Close_qqq'][x.index]), raw=False)

history2['corr_pearson'] = history2['corr_pearson'].fillna(0)
history2['corr_spearman'] = history2['corr_spearman'].fillna(0)


result_data = []
for date, row in history2.iterrows():
    result_data.append({
        "Symbol": "SPY",
        "Date": date,
        "Open": row["Open_spy"],
        "High": row["High_spy"],
        "Low": row["Low_spy"],
        "Close": row["Close_spy"],
        "Volume": row["Volume_spy"],
        "Correlation_Pearson": "",
        "Correlation_Spearman": ""
    })
    result_data.append({
        "Symbol": "QQQ",
        "Date": date,
        "Open": row["Open_qqq"],
        "High": row["High_qqq"],
        "Low": row["Low_qqq"],
        "Close": row["Close_qqq"],
        "Volume": row["Volume_qqq"],
        "Correlation_Pearson": row["corr_pearson"],
        "Correlation_Spearman": row["corr_spearman"],
    })
   
    
   

result = pd.DataFrame(result_data)

result['Date'] = pd.to_datetime(result['Date']).dt.strftime('%Y%m%d')

result.to_csv("spy_qqq_corr.csv", index=False)


```

Tests/Indicators/CorrelationInidicatorTests.cs
grzegorzwodnicki: I left AreNotEqual as the test is Called: `NotEqualCorrelationValue`
The previous test is: `EqualCorrelationValue` and there is `AreEqual`.
Please let me know what to do in this situation, because in this case, we will get 2 the same tests with 2 different names.

Indicators/CorrelationIndicator.cs
```diff
@@ -0,0 +1,211 @@
...
+    /// movements.
+    /// A positive correlation indicates that the target tends to move in the same direction as the market index, 
+    /// while a negative correlation suggests an inverse relationship. A correlation close to 0 implies a weak or 
+    /// no linear relationship.
+    /// Commonly, the SPX index is employed as the benchmark for the overall market when calculating correlation, 
+    /// ensuring a consistent and reliable reference point. This helps traders and investors make informed decisions 
+    /// regarding the risk and behavior of the target security in relation to market trends.
+    /// </summary>
+
+    public class CorrelationIndicator : TradeBarIndicator, IIndicatorWarmUpPeriodProvider
```
Martin-Molinero, line 37: - Think the indicator name should just be `Correlation`, no need for the Indicator part 👌. 
- Missing to add the helper method in the `QCAlgorith.Indicators.cs`
- I actually think the indicator base type doesn't really need to be `TradeBar`, the input doesn't even need to be a Bar, just needs a `Value` which is part of the basic BaseData? Suggest looking at other generic indicators & how they handle this, like SMA

Indicators/CorrelationIndicator.cs
```diff
@@ -0,0 +1,211 @@
...
+            return _correlation;
+        }
+
+        /// <summary>
+        /// Computes the returns with the new given data point and the last given data point
+        /// </summary>
+        /// <param name="rollingWindow">The collection of data points from which we want
+        /// to compute the return</param>
+        /// <returns>The returns with the new given data point</returns>
+        private static decimal GetNewReturn(RollingWindow<decimal> rollingWindow)
```
Martin-Molinero, line 168: This method isn't used I believe?

Indicators/CorrelationIndicator.cs
```diff
@@ -0,0 +1,211 @@
...
+            return ((rollingWindow[0] / rollingWindow[1]) - 1);
+        }
+
+        /// <summary>
+        /// Computes the correlation value usuing symbols values
+        /// correlation values assing into _correlation property
+        /// </summary>
+        private void ComputeCorrelation()
+        {
+            if (_targetDataPoints.Count < _period || _referenceDataPoints.Count < _period)
```
Martin-Molinero, line 179: The called is also checking other cases where the correlation might no need to be calculated `_targetDataPoints.Samples == _referenceDataPoints.Samples && _referenceDataPoints.Count > 1`, believe these checks should be merged into a single location

Indicators/CorrelationIndicator.cs
Martin-Molinero: Agree with JA, shouldn't be doing `ToList()` for each sub enumerator, should handle required conversions. Maybe we could directly store doubles in the `_targetDataPoints/_referenceDataPoints` collections, no need to store decimals if we will only used them as double later (which is okay for this use case I thinK)

Martin-Molinero (changes requested):
Thank you! Leaving a few requests 👍

Indicators/CorrelationIndicator.cs
grzegorzwodnicki: Yes, agree that the class should be called `Correlation`, but we are using method: `Correlation` from `MathNet.Numerics.Statistics`.
For avoid writing a full namespace path and future mistakes decided to change the name of Indicator to `CorrelationIndicator`.
I can use aliases:
```
using Correlation = MathNet.Numerics.Statistics.Correlation;
using CorrelationIndicator = QuantConnect.Indicators.Correlation;
```
But I'm not sure if it will be a good practice.
I'm open to any suggestion.

grzegorzwodnicki: I tried to change `TradeBar` to `IndicatorDataPoint`, but I noticed that there is a risk that the indicator will not get a Symbol Name.
In the correlation indicator important is to separate 2 data sources using Symbol Name.

grzegorzwodnicki:
Thank you for reviewing, just committed changes and added comments.

Indicators/CorrelationIndicator.cs
jhonabreul: I think the full namespace is ok in this case, it is used just twice in the same location.
Also, could handle indicator data points without symbol by throwing an exception. It would be just like passing a bar with a symbol that is neither the target not the reference.

jhonabreul (changes requested):
Thanks again! Leaving a couple comments and also, still mising to add the helper method in `QCAlgorith.Indicators.cs`

grzegorzwodnicki:
Thank you for your reviews, just pushed 2 commits with updates.

Indicators/CorrelationIndicator.cs
jaredbroad: We can avoid always referencing the Numerics version as its not likely many people using it.

Indicators/Correlation.cs
```diff
@@ -0,0 +1,194 @@
...
+    /// movements.
+    /// A positive correlation indicates that the target tends to move in the same direction as the market index, 
+    /// while a negative correlation suggests an inverse relationship. A correlation close to 0 implies a weak or 
+    /// no linear relationship.
+    /// Commonly, the SPX index is employed as the benchmark for the overall market when calculating correlation, 
+    /// ensuring a consistent and reliable reference point. This helps traders and investors make informed decisions 
+    /// regarding the risk and behavior of the target security in relation to market trends.
+    /// </summary>
+
+    public class Correlation : BarIndicator, IIndicatorWarmUpPeriodProvider
```
AlexCatarino, line 35: Thank you for upgrading from `TradeBarIndicator` to `BarIndicator`
On this PR, we added a test for accepting `QuoteBar` which will assert this change:
https://github.com/QuantConnect/Lean/pull/7540/files#diff-0ad130eeee282f4dfc889caec0e1f31988ca69550f89eed6fb86e15ec5517995R147

grzegorzwodnicki: just pushed changed with AcceptsQuoteBarsAsInput test

Martin-Molinero (approved):
Thank you @grzegorzwodnicki! 💪

PR #7550, opened 2023-11-01
Invalidate option orders during a stock split

### LINKED ISSUE #7444: Prevent MOO for Option When Split Occurs
#### Expected Behavior
Cannot place MOO for options when the underlying split occurs.

#### Actual Behavior
If we place a MOO for options when the underlying split occurs, it opens a position and LEAN doesn't close it due to the split.

#### Potential Solution
N/A

#### Reproducing the Problem
```python
# region imports
# endregion

class Vxx(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2021, 1, 1)  
        self.SetEndDate(2021, 5, 1)
        self.vxx = self.AddEquity("VXX", Resolution.Daily)
        self.option = self.AddOption(self.vxx.Symbol, Resolution.Daily)
        self.option.SetFilter(-1, 1, 15, 45)

    def OnData(self, data: Slice):
        chain = data.OptionChains.get(self.option.Symbol)
        if not chain:
            return

        expiry = max([c.Expiry for c in chain])
        contracts = sorted([c for c in chain if c.Expiry == expiry],
            key=lambda c: c.Strike, reverse=False)

        for split in data.Splits.Values:
            if split.Type == SplitType.SplitOccurred:
                self.Debug(f'{self.Time} {split}')
                self.MarketOrder(contracts[0].Symbol, 1)

        if self.Portfolio.Invested:
            message = f'{self.Time} Underlying: {chain.Underlying.Price}. Strikes: ' + \
                ', '.join(sorted(set([str(x.Strike) for x in contracts])))
            self.Debug(message)
```

This algorithm only places one trade, and a position for an option with a pre-split strike is held. The option chain includes this option.

 #### Checklist

jhonabreul:
#### Description

Options orders are not allowed during a stock split.

Reproducing the original issue:

- Add an equity and option with daily resolution:
    - Make sure there is data for the options on the day of the split or use one of the contracts in the `Securities` collection
- Place an order on the same day the `SplitType.SplitOccurred` event is received in `OnData`
    - On this time step, the order of events is as follows:
        1. Universe selection happens [here](https://github.com/jhonabreul/Lean/blob/f5d4bc4f7b7e29f12a688fbb1720c4172dca21d8/Engine/AlgorithmManager.cs#L177-L178)
        2. Security price is split-adjusted [here](https://github.com/jhonabreul/Lean/blob/f5d4bc4f7b7e29f12a688fbb1720c4172dca21d8/Engine/AlgorithmManager.cs#L397) so the strike filter in the universe selection used the pre-split security price: the selected options have pre-split strikes.
        3. The order is placed for one of these contracts, and since there is an open order and eventually a position, the contract is [never removed](https://github.com/jhonabreul/Lean/blob/f5d4bc4f7b7e29f12a688fbb1720c4172dca21d8/Engine/DataFeeds/PendingRemovalsManager.cs#L54-L55) from the securities collection.

This is what is causing this old contract to still be available. This solves this types of problems by not allowing options orders on the same time step the underlying stock split is applied.

#### Related Issue

Closes #7444 

- Regression algorithm
- Unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OptionOrdersOnSplitRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionOrdersOnSplitRegressionAlgorithm.cs
@@ -1,0 +1,134 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System;
+using System.Linq;
+using System.Collections.Generic;
+
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+using QuantConnect.Util;
+using QuantConnect.Orders;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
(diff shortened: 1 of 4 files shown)


Martin-Molinero (approved):
Thank you 👍

Issue #6985, opened 2023-02-20
Implement Alpha Indicator


AlexCatarino:
#### Expected Behavior
Lean supports Alpha as a Lean Indicator which takes a Symbol object, a benchmark Symbol, and a look-back period. 

#### Actual Behavior
Not implemented. See #83.

#### Potential Solution
 - Implement Alpha Indicator in a similar way to Beta

Follow this interface:
```
A(Symbol target, Symbol reference, int period, Resolution? resolution = null)
Alpha(Symbol target, Symbol reference, int period)
Alpha(string name, Symbol target, Symbol reference, int period)
```

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

LouisSzeto:
This one doesn't have a talib implementation, but it could be as easy as `alpha = return - riskfree rate - beta()`, in which can be done with our existing indicators:
Return : `RateOfChange(1)`
Beta: `Beta(same window)`
Risk free rate can be implemented as an arbitrary number, or from some benchmark return series.

LouisSzeto:
[qqq_alpha.csv](https://github.com/QuantConnect/Lean/files/10822653/qqq_alpha.csv)
[spy_alpha.csv](https://github.com/QuantConnect/Lean/files/10822654/spy_alpha.csv)

anbe6083:
Hi there, I'm interested in taking this on. Please be patient with me, as I'm new to this project 😃 

Taking a look in `QuantConnect.Benchmarks`, I see two classes, `FuncBenchmark` and` SecurityBenchmark`, are these appropriate classes to use to calculate the benchmark return? Both have an Evaluate method and seem so, but just wanted to clarify.

alexander-myltsev:
@LouisSzeto I don't understand your notation :) By chance, expand it, please. Better, if you know the code implementation, share, please.

LouisSzeto:
[spy_alpha.csv](https://github.com/QuantConnect/Lean/files/13114233/spy_alpha.csv)
Columns: "SPY close" "benchmark(QQQ) close" "alpha"

[qqq_alpha.csv](https://github.com/QuantConnect/Lean/files/13114234/qqq_alpha.csv)
[...]

PR #7566, opened 2023-11-13
Alpha Indicator #6985

creatidy:
Addition of Alpha Indicator to QuantConnect LEAN

#### Description
Implemented a new Alpha indicator in LEAN. This indicator is designed to measure the excess return of a security over a benchmark or the market. It's assumed that the risk-free rate is 0 in this calculation. The implementation allows for flexibility in setting the period parameter, with separate settings available for both alpha and beta components.

#### Related Issue
Closes [#6985](https://github.com/QuantConnect/Lean/issues/6985)

#### Motivation and Context
The Alpha indicator is a critical tool for portfolio management and performance evaluation. It helps in understanding the value added by a manager's investment decisions, independent of market movements.

Yes, documentation will need to be updated to include details and examples of the new Alpha indicator.

Tests have been added to cover the new indicator. These include unit tests for various scenarios and parameters. Testing was done in a controlled environment to ensure accuracy and performance.

#### Types of changes

#### Checklist:


Tests/TestData/alpha_indicator_datatest.csv
jhonabreul, line 1: Please update to use the data in the files provided in the original issue https://github.com/QuantConnect/Lean/issues/6985#issuecomment-1777078975

Tests/Indicators/AlphaIndicatorTests.cs
```diff
@@ -0,0 +1,215 @@
...
+                else
+                {
+                    Assert.AreEqual(0.0032062252, (double)indicator.Current.Value, 0.0000000001);
+                }
+
+            }
+        }
+
+        [Test]
+        public void NotEqualAlphaValue()
```
jhonabreul, line 202: What is this test asserting?

Indicators/Alpha.cs
```diff
@@ -0,0 +1,257 @@
...
+
+            // Return of the last two data points of the window
+            // (last - previous) / previous
+            return (double)((returns[0] - returns[1]) / returns[1]);
+        }
+
+        /// <summary>
+        /// Computes the alpha of the target used in relation with the reference and stores it in the _alpha field
+        /// </summary>
+        private void ComputeAlpha()
```
jhonabreul, line 218: See the comments in the original issue, this Alpha indicator can be simplified by using our existing `Beta` and `RateOfChange` indicators.

jhonabreul (changes requested):
Thanks for your contribution! Leaving a couple of comments below. The Alpha indicator implementation could be simplified by using the existing indicators.

Also, the risk free ratio can be a parameter with a default value, like in [SharpeRatio](https://github.com/QuantConnect/Lean/blob/master/Indicators/SharpeRatio.cs#L66C81-L66C81)

Tests/TestData/alpha_indicator_datatest.csv
creatidy: Thank you for your feedback. I noticed that the data provided in the original issue didn't align with the standard calculation of Alpha I used for this implementation. According to the definition I followed, Alpha is calculated as:

Alpha = Actual Average Return (Target) − Beta × Average Return (Reference)

In your example, while the Beta and the Average Return of the Reference are correctly calculated for a 252-day period, I found that the Average Return of the Target was incorrectly determined. To maintain the accuracy and integrity of the Alpha calculation, I opted to use a different dataset that correctly represents these values.

Here is a script I've prepared independently to verify the results (including for Beta indicator):
```
import pandas as pd

df = pd.read_csv('data.csv')

# Filter by symbol, set Date, Symbol as index
amzn = df[df["Symbol"] == "AMZN"]
amzn.set_index("Date", inplace=True)
spx = df[df["Symbol"] == "SPX"]
spx.set_index("Date", inplace=True)

amzn["ret"] = amzn["Close"].pct_change()
spx["ret"] = spx["Close"].pct_change()

# Popul. variance column by using rolling window of 5
spx["variance"] = spx["ret"].rolling(5).var(ddof=0)
amzn["variance"] = amzn["ret"].rolling(5).var(ddof=0)

# join all data in dataframes by Date
amzn = amzn.join(spx["variance"], on="Date", rsuffix="_SPX")
amzn = amzn.join(spx["ret"], on="Date", rsuffix="_SPX")

spx = spx.join(amzn["variance"], on="Date", rsuffix="_AMZN")
spx = spx.join(amzn["ret"], on="Date", rsuffix="_AMZN")

# Calculate covariance by using rollong window of 5 for both
amzn["covariance"] = amzn["ret"].rolling(5).cov(amzn["ret_SPX"], ddof=0)
spx["covariance"] = spx["ret"].rolling(5).cov(spx["ret_AMZN"], ddof=0)

# Calculate beta
amzn["Beta-test"] = amzn["covariance"] / amzn["variance_SPX"]
spx["Beta-test"] = spx["covariance"] / spx["variance_AMZN"]

amzn["ret_avg"] = amzn["ret"].rolling(5).mean()
spx["ret_avg"] = spx["ret"].rolling(5).mean()
amzn = amzn.join(spx["ret_avg"], on="Date", rsuffix="_SPX")
spx = spx.join(amzn["ret_avg"], on="Date", rsuffix="_AMZN")

amzn["Alpha"] = amzn["ret_avg"] - amzn["Beta-test"] * amzn["ret_avg_SPX"]
spx["Alpha"] = spx["ret_avg"] - spx["Beta-test"] * spx["ret_avg_AMZN"]

# change index from Data to Date, Symbol
amzn.reset_index(inplace=True)
amzn.set_index(["Date", "Symbol"], inplace=True)
spx.reset_index(inplace=True)
spx.set_index(["Date", "Symbol"], inplace=True)
# join all data in dataframes by two columns: Date, Symbol
df = df.join(amzn["Alpha"], on=["Date", "Symbol"], rsuffix="_AMZN")
df = df.join(spx["Alpha"], on=["Date", "Symbol"], rsuffix="_SPX")

# Concat df Alpha to one column
df["Alpha"] = df["Alpha_AMZN"]
# Move 1 row down
df["Alpha"] = df["Alpha"].shift(1)
# Fill first 10 rows with 0
df["Alpha"][:10] = 0
# Fill with the previous value
df["Alpha"] = df["Alpha"].fillna(method="ffill")
# Fill with 0
df["Alpha"] = df["Alpha"].fillna(0)

# Drop columns
df.drop(["Alpha_AMZN", "Alpha_SPX", "Beta"], axis=1, inplace=True)

print(df)

# Save without index
df.to_csv("data-out.csv", index=False)
```

Tests/Indicators/AlphaIndicatorTests.cs
creatidy: I've copied from tests for Beta, but you're right... I'll change it to something valuable.

Indicators/Alpha.cs
creatidy: Thanks for the suggestion. I chose not to reuse the existing code initially because I thought my approach was a better fit for the specific requirements of the Alpha indicator. However, I understand the benefits of using the established Beta and RateOfChange indicators for consistency and efficiency. I'll revise my implementation to incorporate these.

creatidy:
Rebased to current master

jhonabreul (changes requested):
Thanks! Leaving minor comments. Also, let's wait for a depper review on your data generation script by @LouisSzeto

Tests/TestData/alpha_indicator_datatest.csv
jhonabreul: @LouisSzeto Do you mind taking a look at this, please?

LouisSzeto:
@creatidy Hi! Thank you for the input. Refer to the script generating checking data above, I'm assuming your raw data is in daily resolution. I had a few query want to clear:
1. Why would you use 5-day as the variance parameter, it is not the standard for any stock price model. For example, IV in option uses monthly, and Brownian Motion in stocks uses annual usually. The same goes for return as well, amost all models use annual return * 1/252. Please also set a default values for the periods in the constructors of `Alpha`.
2. I'm very concern about the `ddof=0` arguments. In any stochastic model, we almost always uses sample variance instead of population (i.e. `ddof=1`) as the second-moment estimator, since it is the uniform minimum variance unbiased estimator (or you can deduce from MLE).
3. All returns in here should actually be adjusted to risk-free rate (just like what you've done in the `Alpha.cs`). Instead of a static entry, you can use the interest rate in [here](https://github.com/QuantConnect/Lean/blob/master/Data/alternative/interest-rate/usa/interest-rate.csv).

creatidy:
@LouisSzeto Thank you for your thorough review and valuable insights. I understand your concern about standard modeling practices. I have revised the code to align with more conventional return calculations.

Here's how I've addressed the rest of your points:
1. **5-Day Variance Parameter**: The 5-day period was initially used for consistency with the beta unit tests in our test suite. I left this value for existing tests and applied 252 for new ones. 

2. **Degrees of Freedom in Variance Calculation**: I completely agree with your point on using `ddof=1` for sample variance in stochastic models. I've rewritten the script and tested the QC code against its results (and an additional Excel model). I've also conducted extensive tests to ensure the accuracy of these changes. Notably, there's a parameter order discrepancy with the TALIB library's BETA function. It seems that TALIB uses BETA(reference, target) instead of BETA(target, reference), which could account for the differences in our preliminary results. I can share the script for further review if needed.

3. **Adjustment for Risk-Free Rate**: The implementation now includes the `InterestRateProvider` for dynamic risk-free rate adjustments, enhancing the accuracy of our return calculations.

4. **Default Period Values**: Following your suggestion, I have set the default values for period arguments as `alphaPeriod=1` and `betaPeriod=252`, reflecting the standard annual period used in financial modeling.

I'm open to any further suggestions or reviews and appreciate the opportunity to refine and improve our code.

Indicators/Alpha.cs
```diff
@@ -0,0 +1,260 @@
...
+            
+            _targetSymbol = targetSymbol;
+            _referenceSymbol = referenceSymbol;
+            _alphaPeriod = alphaPeriod;
+            _betaPeriod = betaPeriod;
+
+            if (useInterestRateProvider)
+            {
+                _interestRateProvider = new InterestRateProvider();
+            }
```
jhonabreul, line 135: Hey let's go back to only having the `riskFreeRate` parameter. There is work in line that will make this not useful quite soon. The pattern in similar indicators is just having the `riskFreeRate` parameter.

jhonabreul (changes requested):
@creatidy Thanks again for taking the feedback and sorry for the back and forth.

I have just a minor comment here about the risk free rate provider usage after talking to the team.

Also, hopefully @LouisSzeto can help us here finishing the data generation review and approval 🤞

LouisSzeto:
> @creatidy Thanks again for taking the feedback and sorry for the back and forth.
> 
> I have just a minor comment here about the risk free rate provider usage after talking to the team.
> 
> Also, hopefully @LouisSzeto can help us here finishing the data generation review and approval 🤞

@creatidy Thank you for your changes. It looks good to me. (cc @jhonabreul)

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -305,6 +305,29 @@ public AugenPriceSpike APS(Symbol symbol, int period = 3, Resolution? resolution
...
+        /// <param name="target">The target symbol whose Alpha value we want</param>
+        /// <param name="reference">The reference symbol to compare with the target symbol</param>
+        /// <param name="alphaPeriod">The period of the Alpha indicator</param>
+        /// <param name="betaPeriod">The period of the Beta indicator</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="riskFreeRate">The risk free rate</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to casting the input value to a TradeBar</param>
+        /// <returns>The Alpha indicator for the given parameters</returns>
+        [DocumentationAttribute(Indicators)]
+        public Alpha A(Symbol target, Symbol reference, int alphaPeriod = 1, int betaPeriod = 252, Resolution? resolution = null, decimal riskFreeRate = 0.0m, Func<IBaseData, IBaseDataBar> selector = null)
```
jhonabreul, line 321: Can you move this up so it appears in the right alphabetical order among the indicator helper methods?

jhonabreul, line 321: Could you take a look at https://github.com/QuantConnect/Lean/pull/7594 and see how the `SharpeRaitio` indicator takes a nullable risk free ratio and decides whether to use the algorithm's risk free rate model or a constant rate? So you can make this indicator follow the same pattern. See both the `SR()` method and the `SharpeRatio` class constructors.

Indicators/Alpha.cs
```diff
@@ -0,0 +1,240 @@
...
+
+        /// <summary>
+        /// Period of the indicator - alpha
+        /// </summary>
+        private readonly decimal _alphaPeriod;
+
+        /// <summary>
+        /// Period of the indicator - beta
+        /// </summary>
+        private readonly decimal _betaPeriod;
```
jhonabreul, line 53: I think you can remove this field, since `_beta` has it and it's not used again

Indicators/Alpha.cs
```diff
@@ -0,0 +1,240 @@
...
+        /// <param name="input"></param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentException"></exception>
+        protected override decimal ComputeNextValue(IBaseDataBar input)
+        {
+            if (input is null)
+            {
+                throw new ArgumentNullException(nameof(input));
+            }
+            Symbol inputSymbol = input.Symbol;
```
jhonabreul, line 186: Prefer `var` over explicit typing in most cases

jhonabreul (changes requested):
Thanks again, @creatidy! Great work.

I apologize again for the several requests. I just have a few final comments and this will be good to go!

PR #7645, opened 2023-12-27
Solve race condition bug in LocalObjectStore

### LINKED ISSUE #7632: Object Store Delete Fails to Delete Object
#### Expected Behavior
`ObjectStore.Delete` removes file key/file.

#### Actual Behavior
In some cases, `ObjectStore.Delete` doesn't remove the key/file.

#### Potential Solution
N/A

#### Reproducing the Problem
https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_90ae3cbce345fc89886d930eaddaaf20.html

We need to add a 20s sleep to represent the calculations from an algorithm because a quick save and delete works fine.

#### System Information
QuantConnect Cloud.

#### Checklist

Marinovsky:
#### Description

The bug was raised due to a race condition when writing a file and deleting it. In order to stop it, a check condition was added at the end of the PersistData() method to remove files that have just been deleted. It's worth to mention, that the try/catch sentence was added to prevent another race condition when deleting the same file at the same time in PersistData() and Delete() methods.
#### Related Issue

Closes #7632
#### Motivation and Context

With this change users will be able to save files and then immediately delete them without any error

N/A

It was created a unit test that reproduces the mentioned race condition sleeping certain threads. Without the changes the test fails half of the times. As it is expected, after these changes, the test works correctly all of the times.
#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/Storage/LocalObjectStore.cs
+++ b/Engine/Storage/LocalObjectStore.cs
@@ -531,12 +531,12 @@ namespace QuantConnect.Lean.Engine.Storage
                         // clear the dirty flag
                         kvp.Value.SetClean();
 
+                        // This kvp could have been deleted by the Delete() method
                         if (!_storage.Contains(kvp))
                         {
                             try
                             {
                                 FileHandler.Delete(filePath);
-                                return true;
                             }
                             catch
                             {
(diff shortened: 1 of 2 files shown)


Martin-Molinero (approved):
Thank you! 🚀

PR #7758, opened 2024-02-06
Implement dividend yield model

### LINKED ISSUE #7742: Dividend Yield is not considered in ImpliedVolatility and Delta
#### Expected Behavior
According to Black-Scholes-Merton model, continuous dividend yield shall be handled when estimating IV and Delta

#### Actual Behavior
The current `ImpliedVolatility` and `Delta` indicators are not handling, so the result are diverging from IB.

#### Potential Solution
Implement dividend yield into the indicators

#### Reproducing the Problem
Current unit tests have high errors from deep ITM/OTM options when comparing with IB's number.

#### System Information
N/A

#### Checklist

LouisSzeto:
#### Description
- Implement dividend yield rate model interface for option IV & greeks calculation, in which as external plugin model from LEAN.
- A dividend yield model calculating 1-year continuous yield rate from past year's payout.
- Updated option indicators (`ImpliedVolatility` & `Delta`) and their helper methods to adopt dividend yield.

#### Related Issue
closes #7742

#### Motivation and Context
More accurate option indicators, and align methodology with other sources (e.g. IB, QL)

Extra input arguments from option indicators (`ImpliedVolatility` & `Delta`) docs, should be handled by autogen

Added unit tests and compared with 3rd party source values

#### Types of changes

#### Checklist:


Common/Data/DividendYieldProvider.cs
```diff
@@ -0,0 +1,188 @@
...
+            return dividendYield;
+        }
+
+        /// <summary>
+        /// Generate the daily historical dividend yield
+        /// </summary>
+        protected void LoadDividendYieldProvider()
+        {
+            var directory = Path.Combine(Globals.DataFolder, "equity", "usa", "factor_files", $"{_symbol.Value.ToLowerInvariant()}.csv");
+            _dividendYieldRateProvider = FromCsvFile(directory);
```
Martin-Molinero, line 99: -  Don't think we should directly access the csv, they might not be there when using the zip factor file provider, they can be in cache etc. Believe we could do something like the following:

```csharp
            var factorFileProvider = Composer.Instance.GetPart<IFactorFileProvider>();

            foreach (var factorRow in factorFileProvider.Get(_symbol))
            {
                var corporateFactor = factorRow as CorporateFactorRow;
                if (corporateFactor != null)
                {

                }
            }
```


- Also note that for future options, index options, the dividend yield provider shouldn't be really used?
- For live trading we will need to refresh the dividend yield, thinking about the long lived deployment case. Please take a look at `LocalZipFactorFileProvider` and how it solves it, think we can copy it over here 👍

Martin-Molinero (changes requested):
Nice!👍 leaving a couple of requests and concerns

Tests/Indicators/ImpliedVolatilityTests.cs
```diff
@@ -39,112 +42,113 @@ public void SetUp()
...
+        [TestCase("SPY230811P00430000", 0.015)]
+        [TestCase("SPY230811P00450000")]
+        [TestCase("SPY230811P00470000", 0.075)]
+        [TestCase("SPY230901C00430000", 0.025)]
+        [TestCase("SPY230901C00450000")]
+        [TestCase("SPY230901C00470000")]
+        [TestCase("SPY230901P00430000")]
+        [TestCase("SPY230901P00450000")]
+        [TestCase("SPY230901P00470000", 0.035)]
+        public void ComparesAgainstExternalDataAfterReset(string fileName, double errorMargin = 0.01, int column = 2)
```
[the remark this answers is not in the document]
LouisSzeto: The parameters are actually "best-guess" by us, we're still asking IB for their parameters calculation, but we already know dividend yield is an input. Will update the external source unit tests after their response. Right now, no ETA can be provided. We may just refer to QL's values for now.

Common/Data/DividendYieldProvider.cs
Martin-Molinero: The last live trading point needs to be addressed 👍

Common/Data/DividendYieldProvider.cs
```diff
@@ -119,24 +122,17 @@ protected void LoadDividendYieldProvider()
...
-
-            // skip the first header line, also skip #'s as these are comment lines
             var dividendYieldProvider = new Dictionary<DateTime, decimal>();
-            var lines = dataProvider.ReadLines(file).Where(x => !string.IsNullOrWhiteSpace(x)).ToArray();
 
             // calculate the dividend rate from each payout
-            var previousRate = 0m;
-            for (int i = lines.Length - 1; i >= 0; i--)
+            var subsequentRate = 0m;
+            foreach (var row in corporateFactors.OrderByDescending(corporateFactor => corporateFactor.Date).ToArray())
```
Martin-Molinero, line 131: `.ToArray()` can remove this 👍, xml docs of the method are off

Common/Data/DividendYieldProvider.cs
```diff
@@ -95,8 +93,13 @@ public decimal GetDividendYield(DateTime date)
...
-            _dividendYieldRateProvider = FromCsvFile(directory);
+            var factorFileProvider = Composer.Instance.GetPart<IFactorFileProvider>();
+            var corporateFactors = factorFileProvider
+                .Get(_symbol)
+                .Select(factorRow => factorRow as CorporateFactorRow)
+                .Where(corporateFactor => corporateFactor != null);
+
+            _dividendYieldRateProvider = FromCorporateFactorRow(corporateFactors);
 
             _lastDividendYieldDate = DateTime.UtcNow.Date;
```
Martin-Molinero, line 104: I don't think need to forward the last dividend yield until now, can actually just keep the last dividend yield value on the actual last date we have data for, and return it in `GetDividendYield(DateTime date)` instead of `DividendYieldRateProvider[_lastDividendYieldDate]`

Tests/Common/Data/DividendYieldProviderTests.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        [TestCase("19700306", 0.0)]   // Date in before the first date in file
+        [TestCase("20200205", 0.04147)]
+        [TestCase("20200206", 0.03355)]
+        [TestCase("20200207", 0.03355)]
+        [TestCase("20210203", 0.01676)]
+        [TestCase("20210204", 0.01239)]
+        [TestCase("20210205", 0.01239)]
+        [TestCase("20491231", 0.01239)]   // Date in far future, assuming same rate
+        public void GetDividendYieldRate(string dateString, double expected)
```
Martin-Molinero, line 32: Nice 👍  let's add a unit test for a case for which we have no factor file data to assert it's handled happily

Common/Data/DividendYieldProvider.cs
```diff
@@ -0,0 +1,192 @@
...
+
+        /// <summary>
+        /// Generate the daily historical dividend yield
+        /// </summary>
+        /// <remarks>Exposed for testing</remarks>
+        protected virtual void LoadDividendYieldProvider()
+        {
+            var factorFileProvider = Composer.Instance.GetPart<IFactorFileProvider>();
+            var corporateFactors = factorFileProvider
+                .Get(_symbol)
```
Martin-Molinero, line 132: Whops, here ur loading the factors only for the set symbol, what about the others, given the `_dividendYieldRateProvider` is static? Seems it should be a per symbol dictionary being used? like the map/factor file providers use

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -534,20 +534,26 @@ public DonchianChannel DCH(Symbol symbol, int period, Resolution? resolution = n
...
             IRiskFreeInterestRateModel riskFreeRateModel = riskFreeRate.HasValue
                 ? new ConstantRiskFreeRateInterestRateModel(riskFreeRate.Value)
                 // Make it a function so it's lazily evaluated: SetRiskFreeInterestRateModel can be called after this method
                 : new FuncRiskFreeRateInterestRateModel((datetime) => RiskFreeInterestRateModel.GetInterestRate(datetime));
-            var delta = new Delta(name, symbol, riskFreeRateModel, optionModel, ivModel);
+            IDividendYieldModel dividendYieldModel = dividendYield.HasValue
+                ? new ConstantDividendYieldModel(dividendYield.Value)
+                : symbol.ID.SecurityType == SecurityType.FutureOption || symbol.ID.SecurityType == SecurityType.IndexOption
+                ? new ConstantDividendYieldModel(0m)
+                : new DividendYieldProvider(symbol.Underlying);
```
Martin-Molinero, line 555: Mind expanding this into multiple lines for readability?

Martin-Molinero (commented):
Cool! Minor request, would be nice to add a regression algorithm adding/using & asserting the new greek indicators

Martin-Molinero (approved):
Nice! 🚀

Issue #7793, opened 2024-02-17
Custom data can't update OHLC security properties


D1mi3:
#### Expected Behavior

Custom data types should be able to update security OHLC properties if they contain user defined OHLC fields that follow a name convention (e.g: "Open", "High", etc.)

#### Actual Behavior
Custom data types update all OHLC security properties with a single fixed value from their Value attribute

#### Potential Solution
N/A

#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

PR #7808, opened 2024-02-22
PythonData OHLCV Security properties update logic

D1mi3:
Added logic to allow PythonData with OHLCV attributes to update Security properties.

#### Description

Added code in SecurityCache.cs after Tick and IBars updates. This block allows any valid OHLCV attribute contained on PythonData _storage to set _lastOHLCUpdate. 

#### Related Issue

Close https://github.com/QuantConnect/Lean/issues/7793

#### Motivation and Context

Inaccurate backtesting results for OHLCV PythonData used as price data. (e.g. Unrealistic fill prices for limit orders as high or low properties aren't taken into account).

Minor documentation changes explaining the required attribute format to correctly define OHLCV PythonData bars.

Backtesting results & unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Securities/SecurityCache.cs
+++ b/Common/Securities/SecurityCache.cs
@@ -19,6 +19,9 @@ using QuantConnect.Data;
 using QuantConnect.Data.Market;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using QuantConnect.Python;
+using System.Reflection;
+using Python.Runtime;
 using QuantConnect.Data.Fundamental;
 using QuantConnect.Data.UniverseSelection;
 using QuantConnect.Util;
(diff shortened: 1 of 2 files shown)


D1mi3:
> Hi @D1mi3 - interesting PR thank you.
> 
> Can you please shuffle the the end of the method so it doesn't impact speed, and also confirm the data is added to the cache (and thus available for fill models) with a test.
> 
> It may be easier to convert the object to an IBar instance and store that in cache vs keeping it in PythonData format.

Hi @jaredbroad, thanks for the feedback. I'll soon update the PR with the shuffling suggested. With respect to the tests you mentioned; they're already implemented, please refer to them at `SecurityCacheTests.cs#L365`.

Finally, I agree with the simplicity of just converting PythonData to IBar, since that was my first instinct. The problem of that approach is that PythonData doesn't implement the IBar interface at all, thus it's not possible to cast. I thought about making PythonData implement IBar but ended up not doing that since it could have unintended side effects.

D1mi3:
> Hi @D1mi3 - interesting PR thank you.
> 
> Can you please shuffle the the end of the method so it doesn't impact speed, and also confirm the data is added to the cache (and thus available for fill models) with a test.
> 
> It may be easier to convert the object to an IBar instance and store that in cache vs keeping it in PythonData format.

All done.

jaredbroad:
Close and making generic - base data defined func (security, basedata) and let the func define the properties.

PR #7820, opened 2024-02-28
Allow smoothing in `Implied Volatility` and add forward tree model

### LINKED ISSUE #7781: `ImpliedVolatility` does not match IB's figure
#### Expected Behavior
`ImpliedVolatility` indicator values shall be close to IB's figure

#### Actual Behavior
It is not now

#### Potential Solution
We reverse-engineer an 'intrinsic value' selector as well as forward-tree model for IV.

#### Reproducing the Problem
refer to current external data comparison unit tests

#### System Information
NA

#### Checklist

LouisSzeto:
#### Description
- Add smoothing and allow custom objective function in `Implied Volatility`
- Add binomial forward tree model

#### Related Issue
closes #7781

#### Motivation and Context
To match IB IV and Greeks, and allow overriding to match other 3rd party sources

Add description on mirror option contract and smoothing

Add unit tests and 3rd party source comparison

#### Types of changes

#### Checklist:


Tests/Indicators/ImpliedVolatilityTests.cs
```diff
@@ -44,77 +46,153 @@ public void SetUp()
...
+                    var putIndicator = new ImpliedVolatility(put, call, interestRate, dividendYield, model);
+
+                    RunTestIndicator(call, put, callIndicator, putIndicator, items, callColumn, putColumn, errorMargin);
+                }
+            }
+        }
+
+        // We are unsure about the smoothing function
+        [TestCase(7, 6)]
+        public void ComparesAgainstExternalData2(int callColumn, int putColumn, double errorMargin = 10)
```
jhonabreul, line 76: Isn't an error margin of 10 too permissive?

Tests/Indicators/OptionBaseIndicatorTests.cs
```diff
@@ -62,36 +61,59 @@ protected virtual OptionIndicatorBase CreateIndicator(QCAlgorithm algorithm)
...
-                var refValue = Parse.Double(items[column]);
+            var expected = 0d;
+            var acceptance = 0d;
+            try
+            {
+                expected = double.Parse(items[callColumn], NumberStyles.Any, CultureInfo.InvariantCulture);
+                acceptance = Math.Max(errorMargin * Math.Abs(expected), 1e-4);     // percentage error
+                Assert.AreEqual(expected, (double)callIndicator.Current.Value, acceptance);
+            }
+            catch { }
```
jhonabreul, line 101: Can it throw?

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -588,6 +588,69 @@ public DonchianChannel DCH(Symbol symbol, int period, Resolution? resolution = n
...
+                dividendYieldModel = new ConstantDividendYieldModel(dividendYield.Value);
+            }
+            else if (symbol.ID.SecurityType == SecurityType.FutureOption || symbol.ID.SecurityType == SecurityType.IndexOption)
+            {
+                dividendYieldModel = new ConstantDividendYieldModel(0m);
+            }
+            else
+            {
+                dividendYieldModel = new DividendYieldProvider(symbol.Underlying);
+            }
```
jhonabreul, line 626: Maybe there's a chance to abstract this into one method and reuse it across `Delta` methods

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -914,13 +977,56 @@ public Identity Identity(Symbol symbol, TimeSpan resolution, Func<IBaseData, dec
...
```
jhonabreul, line 1007: Same comment about abstracting and reusing it

Tests/Indicators/ImpliedVolatilityTests.cs
LouisSzeto: Hey! We just place it there as a reference, but we are not going to spend time to reverse engineer the smoothing method

Tests/TestData/greeksindicator/american/third_party_1_greeks.csv
```diff
@@ -0,0 +1,19 @@
+,,,,model_price,model_price,model_impliedVol,model_impliedVol,model_delta,model_delta,model_gamma,model_gamma,model_vega,model_vega,model_theta,model_theta,model_undPrice,TTM,interest_rate,dividend_yield
+right,,,,-1,1,-1,1,-1,1,-1,1,-1,1,-1,1,,,,
+ticker,expiry,strike,time,,,,,,,,,,,,,,,,
```
Martin-Molinero, line 3: I don't understand these headers lines? which seem to be skip while reading? Same for the file bellow

Tests/Indicators/ImpliedVolatilityTests.cs
```diff
@@ -44,77 +46,138 @@ public void SetUp()
...
-        [TestCase("SPY230811P00470000", 0.075)]
-        [TestCase("SPY230901C00430000", 0.025)]
-        [TestCase("SPY230901C00450000")]
-        [TestCase("SPY230901C00470000")]
-        [TestCase("SPY230901P00430000")]
-        [TestCase("SPY230901P00450000")]
-        [TestCase("SPY230901P00470000", 0.035)]
-        public void ComparesAgainstExternalDataAfterReset(string fileName, double errorMargin = 0.01, int column = 2)
+        [TestCase("american/third_party_1_greeks.csv")]
+        public void ComparesAgainstExternalDataAfterReset(string subPath, int callColumn = 7, int putColumn = 6, double errorMargin = 0.08)
```
Martin-Molinero, line 91: - This test is very similar to the one above, reset could be a test case argument, true/false so it's a single unit tests
- Should replace `decimal.Parse` by the `Parse.Decimal` extensions we have to avoid culture specific

Tests/Indicators/DeltaTests.cs
```diff
@@ -141,25 +140,23 @@ public void ComparesDeltaOnBSMModel(decimal price, decimal spotPrice, OptionRigh
...
+        [TestCase(2.642, 430.0, OptionRight.Call, 180, 0.264)]
+        [TestCase(27.772, 430.0, OptionRight.Put, 180, -0.556)]
         public void ComparesDeltaOnCRRModel(decimal price, decimal spotPrice, OptionRight right, int expiry, double refDelta)
         {
             // Under CRR framework
             var symbol = Symbol.CreateOption("SPY", Market.USA, OptionStyle.American, right, 450m, _reference.AddDays(expiry));
-            var indicator = new Delta(symbol, 0.0530m, 0.0153m,
-                    optionModel: OptionPricingModelType.BinomialCoxRossRubinstein,
-                    ivModel: OptionPricingModelType.BlackScholes);
+            var indicator = new Delta(symbol, 0.0403m, 0.0m, OptionPricingModelType.BinomialCoxRossRubinstein, OptionPricingModelType.BlackScholes);
```
Martin-Molinero, line 159: There's no difference with the unit tests above arguments, should there be? Seems they could be joined or share a main implementation method

Indicators/ImpliedVolatility.cs
```diff
@@ -241,33 +446,56 @@ protected override decimal ComputeNextValue(IndicatorDataPoint input)
...
+                try
+                {
+                    Func<double, double> f = (vol) => (double)(TheoreticalPrice(
+                        Convert.ToDecimal(vol), UnderlyingPrice, Strike, timeToExpiration, RiskFreeRate, DividendYield, _oppositeOptionSymbol.ID.OptionRight, _optionModel) - _oppositePrice);
+                    mirrorImpliedVol = Convert.ToDecimal(Brent.FindRoot(f, 1e-7d, 2.0d, 1e-4d, 100));
+                }
+                catch
+                {
+                    Log.Error("ImpliedVolatility.CalculateIV(): Fail to converge, returning 0.");
+                }
```
Martin-Molinero, line 493: I think the whole try catch can be moved into a method passing in the option right/price to avoid the duplication. Can create another method for `Func<double, double> f = (vol) => (double)(TheoreticalPrice(` so it's not created every time

Indicators/ImpliedVolatility.cs
```diff
@@ -175,32 +195,205 @@ public class ImpliedVolatility : OptionIndicatorBase
...
         {
         }
 
+        private bool _isReady => Price.Current.Time == UnderlyingPrice.Current.Time && Price.IsReady && UnderlyingPrice.IsReady;
+
         /// <summary>
         /// Gets a flag indicating when this indicator is ready and fully initialized
         /// </summary>
-        public override bool IsReady => HistoricalVolatility.Samples >= 2 && Price.Current.Time == UnderlyingPrice.Current.Time;
+        public override bool IsReady => UseMirrorContract ? _isReady && Price.Current.Time == _oppositePrice.Current.Time && _oppositePrice.IsReady : _isReady;
```
Martin-Molinero, line 396: I think this will cause an unexpected behavior where IsReady would toggle from true to false, depending on the time of each sub indicator being used? It seems once the condition is meet once it should stick as IsReady true until reset?

Indicators/ImpliedVolatility.cs
```diff
@@ -31,24 +32,43 @@ public class ImpliedVolatility : OptionIndicatorBase
...
+        }
+
+        /// <summary>
+        /// Set the smoothing function of IV, using both call and put IV value
+        /// </summary>
+        /// <param name="function">the smoothing function</param>
+        public void SetSmoothingFunction(PyObject function)
+        {
+            SmoothingFunction = PythonUtil.ToFunc<decimal, decimal, decimal>(function);
+        }
```
Martin-Molinero, line 58: nit - These method should be moved after the constructors below

Indicators/ImpliedVolatility.cs
```diff
@@ -175,32 +195,205 @@ public class ImpliedVolatility : OptionIndicatorBase
...
+            {
+                if (Strike > UnderlyingPrice && Right == OptionRight.Put)
+                {
+                    return mirrorImpliedVol;
+                }
+                else if (Strike < UnderlyingPrice && Right == OptionRight.Call)
+                {
+                    return mirrorImpliedVol;
+                }
+                return impliedVol;
```
Martin-Molinero, line 246: I think we should add some comments what's/why this default implementation doing 🙏

Indicators/OptionGreekIndicatorBase.cs
```diff
@@ -112,20 +112,122 @@ public abstract class OptionGreeksIndicatorBase : OptionIndicatorBase
...
+                if (Strike > UnderlyingPrice && Right == OptionRight.Put)
+                {
+                    return mirrorIV;
+                }
+                else if (Strike < UnderlyingPrice && Right == OptionRight.Call)
+                {
+                    return mirrorIV;
+                }
+                return iv;
+            });
```
Martin-Molinero, line 153: Isn't this the default smoothing function? Don't think it's implementation should be duplicated here but in a centralized location

Indicators/OptionIndicatorBase.cs
```diff
@@ -83,11 +88,21 @@ public abstract class OptionIndicatorBase : IndicatorBase<IndicatorDataPoint>, I
         /// </summary>
         public IndicatorBase<IndicatorDataPoint> Price { get; }
 
+        /// <summary>
+        /// Gets the mirror option price level, for implied volatility
+        /// </summary>
+        protected IndicatorBase<IndicatorDataPoint> _oppositePrice { get; private set; }
```
Martin-Molinero, line 94: `protected IndicatorBase<IndicatorDataPoint> _oppositePrice` following the pattern of the other indicators, should this probably be public too?

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -588,6 +588,48 @@ public DonchianChannel DCH(Symbol symbol, int period, Resolution? resolution = n
...
+        /// <param name="ivModel">The option pricing model used to estimate IV</param>
+        /// <param name="resolution">The desired resolution of the data</param>
+        /// <returns>A new Delta indicator for the specified symbol</returns>
+        [DocumentationAttribute(Indicators)]
+        public Delta D(Symbol symbol, Symbol mirrorOption, decimal? riskFreeRate = null, decimal? dividendYield = null, OptionPricingModelType optionModel = OptionPricingModelType.BlackScholes, 
+            OptionPricingModelType? ivModel = null, Resolution? resolution = null)
+        {
+            var delta = D(symbol, riskFreeRate, dividendYield, optionModel, ivModel, resolution);
+            delta.SetMirrorOptionContract(mirrorOption);
+            delta.ImpliedVolatility.SetMirrorOptionContract(mirrorOption);
```
Martin-Molinero, line 609: `Delta/ImpliedVolatility` have a constructor has `mirrorOption` overloads why do we need `SetMirrorOptionContract`, shouldn't it just be a constructor argument like the rest?
Same comment about the `SmoothingFunction`, for normalization, shouldn't it be a constructor argument?

Tests/Indicators/ImpliedVolatilityTests.cs
```diff
@@ -44,77 +46,138 @@ public void SetUp()
...
-        [TestCase("SPY230811C00470000")]
-        [TestCase("SPY230811P00430000", 0.015)]
-        [TestCase("SPY230811P00450000")]
-        [TestCase("SPY230811P00470000", 0.075)]
-        [TestCase("SPY230901C00430000", 0.025)]
-        [TestCase("SPY230901C00450000")]
-        [TestCase("SPY230901C00470000")]
-        [TestCase("SPY230901P00430000")]
-        [TestCase("SPY230901P00450000")]
-        [TestCase("SPY230901P00470000", 0.035)]
```
Martin-Molinero, line 71: Just double checking here but the error margin seems to have gone up, compared with the previous version 3% vs 8%?

Martin-Molinero (commented):
Cool! Leaving minor comments and requests 👍

Tests/TestData/greeksindicator/american/third_party_1_greeks.csv
LouisSzeto: These are files produced by python pandas when we're caching the raw data. Since the raw dataframe have 2-level columns, so it is displayed like this. Read the csv by pandas can reproduce that 2-level column dataframe

Indicators/ImpliedVolatility.cs
LouisSzeto: We intend to make it this way, since we need all 3 underlying indicators having their timestamp allign to calculate the fair price and greeks

Tests/Indicators/ImpliedVolatilityTests.cs
LouisSzeto: The previous 3% is absolute error (e.g. 15%IV +/- 3%), now it is relative error (e.g. 15%IV +/- 1.2%)

Martin-Molinero (approved):
Nice work 💪

PR #7824, opened 2024-02-29
Fix bug for NQX index options

### LINKED ISSUE #7801: "NQX" index option is not accessible through slice object.
#### Expected Behavior
We should get OptionChains from the slice object for "NQX" -> weekly contracts for the underlying index "NDX"
Reference: https://www.quantconnect.com/datasets/algoseek-us-index-options

#### Actual Behavior
We do not get OptionChains for "NQX" for all timeframes and for all resolutions but we get data using `OptionChainProvider.GetOptionContractList(option.Symbol, time)`. 
Note: We get data for all other index options using both slice object and OptionChainProvider.GetOptionContractList()

#### Potential Solution
We have the data which is confirmed by data from `OptionChainProvider.GetOptionContractList(option.Symbol, time)`. Could be a mapping issue for "NQX".

#### Reproducing the Problem
Attaching the [backtest](https://www.quantconnect.com/terminal/processCache/?request=embedded_backtest_f2290a8aaff3cad293a649771d34a12e.html) for reference.

image
data.OptionChains.get() for "NQX" is always empty.

Data for NQX is available when we use:
```
self = QuantBook()
index_symbol = self.AddIndex("NDX").Symbol
option = self.AddIndexOption(index_symbol, "NQX")
option.SetFilter(lambda universe: universe.IncludeWeeklys().Strikes(-2, 2).Expiration(0, 30))
time = datetime(2023, 6, 1)
symbols = self.OptionChainProvider.GetOptionContractList(option.Symbol, time)
self.History(symbols[:5], time, time+timedelta(2), Resolution.Daily, fillForward=False)
```
image

#### Checklist

Marinovsky:
#### Description

Since NQX is based on 1/5 of the value of NDX (see https://www.nasdaq.com/solutions/nasdaq-100-index-options), when used with the `OptionFilterUniverse.Strikes()` filter, it returned an empty list of option chains. This was because the underlying price was never found among the option strike prices. Therefore, a new column in the SPDB was added containing the value by which the option's strike price must be scaled to be compared with the underlying price in the aforementioned scenarios.
#### Related Issue

Closes #7801 
#### Motivation and Context

With this change users will get option chains for NQX when using the filter `OptionFilterUniverse.Strikes()`

N/A

I created a regression algorithm that gets 2 index option(NQX) chains while using the filter `OptionFilterUniverse.Strikes()`. Then it waits until both option contracts expire, so that it can finally assert that one of them expires ITM and the other one OTM.
#### Types of changes

#### Checklist:


Common/Data/Market/OptionContract.cs
```diff
@@ -26,6 +26,7 @@ namespace QuantConnect.Data.Market
     public class OptionContract
     {
         private Lazy<OptionPriceModelResult> _optionPriceModelResult = new(() => OptionPriceModelResult.None);
+        private readonly decimal _strikeMultipler;
```
Martin-Molinero, line 29: It doesn't seem we need to keep the private strike multiplier? Let's remove it if possible 👍

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -124,8 +127,13 @@ public OptionFilterUniverse Strikes(int minStrike, int maxStrike)
...
 
             // find the current price in the list of strikes
+            // When computing the strike prices we need to take into account
+            // that some option's strike prices are based on a fraction of
+            // the underlying. Thus we need to scale the underlying internal
+            // price so that we can find it among the strike prices
+            // using BinarySearch() method(as it is used below)
             var exactPriceFound = true;
-            var index = _uniqueStrikes.BinarySearch(UnderlyingInternal.Price);
+            var index = _uniqueStrikes.BinarySearch(UnderlyingInternal.Price / _underlyingScaleFactor);
```
Martin-Molinero, line 136: Nice! let's add some new unit tests for this case 👍

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -33,6 +33,7 @@ public class OptionFilterUniverse : ContractSecurityFilterUniverse<OptionFilterU
         private List<decimal> _uniqueStrikes;
         private bool _refreshUniqueStrikes;
         private DateTime _lastExchangeDate;
+        private decimal _underlyingScaleFactor = 1;
```
Martin-Molinero, line 36: nit - Let's make it readonly

Common/Data/Market/OptionContract.cs
```diff
@@ -199,7 +211,7 @@ public static OptionContract Create(BaseData baseData, ISecurityPrice security,
         /// <returns>Option contract</returns>
         public static OptionContract Create(Symbol symbol, Symbol underlyingSymbol, DateTime endTime, ISecurityPrice security, decimal underlyingLastPrice)
```
Martin-Molinero, line 212: The symbol argument is no longer used and can be removed?

Martin-Molinero (changes requested):
Nice! Leaving minor requests, also should review all usages of the option/symbol strike price property believe the PR is missing to update the margin/buying power models if required, we should review the margins in IB for a similar NQX index option and adjust our models accordingly

Algorithm.CSharp/BacktestingBrokerageRegressionAlgorithm.cs
```diff
@@ -96,7 +96,7 @@ public override void OnData(Slice data)
                         from optionContract in chain.OrderByDescending(x => x.Strike)
                         where optionContract.Right == OptionRight.Call
                         where optionContract.Expiry == Time.Date
-                        where optionContract.Strike < chain.Underlying.Price
+                        where optionContract.ScaledStrike < chain.Underlying.Price
```
Martin-Molinero, line 99: Since this issue only really applies for 0.01%, let's revert the changes of the example algorithms to use `ScaledStrike` instead of `Strike`, except for the new algorithm reproducing the issue 👍

Martin-Molinero (approved):

PR #7828, opened 2024-03-04
Handle insufficient margin for option exercise

### LINKED ISSUE #7815: Expand Duplicate Option Exercise Error Message to Diagnose Issue
#### Expected Behavior
Error message for duplicate option exercise would include information to reproduce the very rare bug, like a stack trace or information about the data / price / values that led to the issue.

#### Actual Behavior
At the moment, it just says to contact support but since the bug can't be reproduced easily it's impossible to make progress on. 
Runtime Error: Duplicate option exercise order request for symbol XLK 121020P00036000. Please contact support in BacktestingBrokerage.cs:line 481. 

#### Potential Solution
Capture as much information as possible and log it all. 

#### System Information

jhonabreul:
#### Description

Handling insufficient margin for automatic option exercise, which was causing the "Duplicated option assignment..." exception mentioned in the related issue #7815.

With these changes, simulated option exercises are handled using the `Brokerage.OptionNofication` event to create the option exercise order instead of going through the order processor, so that the margin check is bypassed and the order is placed and filled. After this, we expect a margin call like shown in the new regression algorithm `InsufficientBuyingPowerForAutomaticExerciseRegressionAlgorithm`.

Note: about algorithm's stats change (orders hash): since simulated options exercises are now handled like expired options exercise, without going throught the order processor to place them, some order properties are not populated, like `OrderSubmissionData` and `PriceCurrency`, but this should be taclked in a separate PR to keep this clean. 

#### Related Issue

Close #7815 

Regression algorithm

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Brokerages/OptionNotificationEventArgs.cs
+++ b/Common/Brokerages/OptionNotificationEventArgs.cs
@@ -56,9 +56,8 @@ namespace QuantConnect.Brokerages
         /// <param name="position">The new option position</param>
         /// <param name="tag">The tag to be used for the order</param>
         public OptionNotificationEventArgs(Symbol symbol, decimal position, string tag)
+            : this(symbol, position)
         {
-            Symbol = symbol;
-            Position = position;
             Tag = tag;
         }
 
--- a/Common/Brokerages/OptionNotificationEventArgs.cs
+++ b/Common/Brokerages/OptionNotificationEventArgs.cs
@@ -34,6 +34,11 @@ namespace QuantConnect.Brokerages
         public decimal Position { get; }
 
         /// <summary>
+        /// The tag that will be used in the order
+        /// </summary>
+        public string Tag { get; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="OptionNotificationEventArgs"/> class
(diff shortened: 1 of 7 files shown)


Common/Brokerages/OptionNotificationEventArgs.cs
```diff
@@ -44,12 +49,31 @@ public OptionNotificationEventArgs(Symbol symbol, decimal position)
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OptionNotificationEventArgs"/> class
+        /// </summary>
+        /// <param name="symbol">The symbol</param>
+        /// <param name="position">The new option position</param>
+        /// <param name="tag">The tag to be used for the order</param>
+        public OptionNotificationEventArgs(Symbol symbol, decimal position, string tag)
+        {
+            Symbol = symbol;
+            Position = position;
```
Martin-Molinero, line 61: Nit - we could call `base` to set these

Martin-Molinero (approved):
Nice work! 💪

Martin-Molinero (approved):
Thank you 👍

Issue #6982, opened 2023-02-20
Implements Covariance


AlexCatarino:
#### Expected Behavior
Lean supports Covariance as a Lean Indicator which takes a Symbol object, a benchmark Symbol, and a look-back period. 

#### Actual Behavior
Not supported.

#### Potential Solution
Implement this indicator. Lean currently implemented it internally in the `Beta` indicator: [Beta.cs#L160](https://github.com/QuantConnect/Lean/blob/master/Indicators/Beta.cs#L160).

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

LouisSzeto:
[spy_qqq_cov.csv](https://github.com/QuantConnect/Lean/files/10822690/spy_qqq_cov.csv)

LouisSzeto:
[spy_qqq_cov.csv](https://github.com/QuantConnect/Lean/files/13114349/spy_qqq_cov.csv)
columns: "SPY close" "QQQ close" "annual covariance"

script:
```
import talib

qb = QuantBook()

spy = qb.AddEquity("SPY", Resolution.Daily).Symbol
qqq = qb.AddEquity("QQQ", Resolution.Daily).Symbol

history = qb.History([spy, qqq], datetime(1998, 1, 1), datetime(2023, 2, 23))

close = history.loc[spy].close

# Assumes no risk free rate
history2 = pd.concat([close, history.loc[qqq].close], axis=1).dropna()
history2.columns = [spy, qqq]

cov = history2.rolling(252).cov().unstack(1)[spy][qqq].dropna().to_frame()
cov.columns = ["cov"]
cov["spy close"] = close
cov["qqq close"] = history2.iloc[:, 1]
cov = cov.iloc[:, [1, 2, 0]].dropna()
cov.to_csv("spy_qqq_cov.csv", header=False)
```

dilip0165:
Here I have implemented Variance Indicator. 

using System;
namespace QuantConnect.Indicators
{
    /// <summary>
    /// This indicator computes the n-period population variance.
    /// </summary>
    public class Variance : WindowIndicator<IndicatorDataPoint>, IIndicatorWarmUpPeriodProvider
    {
        private decimal _rollingSum;
        private decimal _rollingSumOfSquares;
[...]

PR #7843, opened 2024-03-11
Add Covariance Indicator

stevespringer:
Add indicator for covariance.

#### Description
Implement covariance indicator taking a target and reference symbol, and a look-back period.

_Note on code reuse:_ I was pointed towards the RateOfChange indicator as an example of reducing duplication, but I couldn't see way to do this without either creating a new base class (similar to WindowIndicator but with a reference window) or introducing a hacky way to hold onto the window passed into `ComputeNextValue()` by the base. I have therefore left the code as-is for now, but can will make changes if requested in review.

#### Related Issue
Closes https://github.com/QuantConnect/Lean/issues/6982

I think I'll need some input on what text to use for the summary of the new class `QuantConnect.Indicators.Covariance`. Not sure where these are sourced from as I'm a first-time contributor.

Added unit tests, including comparison with reference data generated via script in https://github.com/QuantConnect/Lean/issues/6982#issuecomment-1777089341

#### Types of changes

#### Checklist:

- [ X] I have added tests to cover my changes. 
- [ X] All new and existing tests passed.
- [ X] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`


Indicators/Covariance.cs
```diff
@@ -0,0 +1,197 @@
...
+            else if (inputSymbol == _referenceSymbol)
+            {
+                _referenceDataPoints.Add(input.Value);
+            }
+            else
+            {
+                throw new ArgumentException("The given symbol was not target or reference symbol");
+            }
+
+            if (_targetDataPoints.Samples == _referenceDataPoints.Samples && _referenceDataPoints.Count > 1)
```
Martin-Molinero, line 142: Inside `ComputeCovariance` it's also checking for `IsReady`, should only have a single check for this 👍

Indicators/Covariance.cs
```diff
@@ -0,0 +1,197 @@
...
+        private void ComputeCovariance()
+        {
+
+            if (!IsReady)
+            {
+                _covariance = 0m;
+                return;
+            }
+
+            decimal mean1 = 0;
```
Martin-Molinero, line 164: In the lean code base we usually user var cause it's cleaner and try to use description names, for example `decimal mean1 = 0;` -> `var targetMean = 0m;`

Indicators/Covariance.cs
```diff
@@ -0,0 +1,197 @@
...
+            for (int i = 0; i < windowSize; i++)
+            {
+                mean1 += _targetDataPoints[i] / windowSize;
+                mean2 += _referenceDataPoints[i] / windowSize;
+            }
+
+            for (int i = 0; i < windowSize; i++)
+            {
+                sumProduct += (_targetDataPoints[i] - mean1) * (_referenceDataPoints[i] - mean2);
+            }
```
Martin-Molinero, line 178: Take a look at the beta indicator which is calculating the covariance internally, as pointed by potential solution of the issue, I would suggest extracting that logic into this new covariance implementation and make the beta indicator the first user of it 👍

Indicators/Covariance.cs
```diff
@@ -0,0 +1,197 @@
...
+        public Covariance(string name, Symbol targetSymbol, Symbol referenceSymbol, int period)
+            : base(name)
+        {
+            // Assert the period is greater than two, otherwise the covariance can not be computed
+            if (period < 2)
+            {
+                throw new ArgumentException($"Period parameter for Covariance indicator must be greater than 2 but was {period}");
+            }
+
+            WarmUpPeriod = period + 1;
```
Martin-Molinero, line 79: Mind explaining why it requires 1 extra value than the requested period?

Tests/TestData/spy_qqq_cov.csv
```diff
@@ -0,0 +1,629 @@
...
+SPY,20200324,2349900,2441000,2338000,2431500,214693586,
+QQQ,20200324,1798800,1838300,1779500,1836600,86434644,25484181906.50064
+SPY,20200325,2450000,2563500,2397600,2467900,278538590,
+QQQ,20200325,1844900,1895100,1799800,1823000,82965783,25650271902.86495
+SPY,20200326,2492600,2628000,2490500,2612000,224153673,
+QQQ,20200326,1839300,1927200,1835600,1919000,74995622,25571440846.139008
+SPY,20200327,2532200,2608000,2510500,2534200,193638954,
+QQQ,20200327,1869000,1905100,1846600,1853000,54182139,25638656906.974693
+SPY,20200330,2556800,2623900,2535300,2616500,141153243,
+QQQ,20200330,1870600,1924900,1863500,1920400,46667707,25591109173.308296
```
Martin-Molinero, line 123: These covariance values do not seem accurate? `25591109173.308296` the order of magnitude seems off. Believe the mean formular is off

Indicators/Covariance.cs
```diff
@@ -0,0 +1,197 @@
...
+ * limitations under the License.
+*/
+
+using System;
+using QuantConnect.Data.Market;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    /// TODO: figure out what to write here
```
Martin-Molinero, line 22: Should describe what this covariance indicator is, what's it calculating/showing/etc, you can see other indicators as an example.

Martin-Molinero (changes requested):
Hey @stevespringer! Thank you for the contribution. Leaving a few requests

Indicators/Covariance.cs
stevespringer: Sorry, I was taking the lead from another indicator without considering if it's relevant here. I'll remove the plus 1

Tests/TestData/spy_qqq_cov.csv
Martin-Molinero: See my other comments https://github.com/QuantConnect/Lean/pull/7843#discussion_r1520050464

Martin-Molinero:
Closing for now, please reopen when review addressed @stevespringer

PR #7985, opened 2024-04-30
Holdings price adjustment after split

### LINKED ISSUE #7894: Stock Split When Short Causes Negative Equity With Raw Data
#### Expected Behavior

If we are short a stock when a stock split occurs, our equity value doesn't drop below zero.

#### Actual Behavior

> AlgorithmManager.Run(): Portfolio value is less than or equal to zero, stopping algorithm.

#### Potential Solution

N/A

#### Reproducing the Problem

https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_5aaa2df0fe22fb61050280d5a61c07e6.html

The research notebook of this project shows that there is no price data issue. 

TSLA and AAPL both have a stock split during the backtest period in this algorithm. If you swap the asset in the algorithm to AAPL, you'll get the same error.

The error only occurs with Raw data normalization. If you use adjusted prices, the error doesn't occur.

#### System Information

QC Cloud

#### Checklist

jhonabreul:
#### Description

Adjust all prices properties of `SecurityCache` on a split event. This makes sure that on a split, the security holdings price is set to the adjusted price, also making sure that the total portfolio value is properly calculated with the correct holdings value.

The security cache adjustment for a split is delegated to the `SecurityCache` class, which was being done by the `SecurityPortfolioManager`. The cache is now able to update all of its fields as necessary, without limiting to what the conditions in `ProcessDataPoint()` allow.

##### Affected algorithms:

- `OptionTimeSliceRegressionAlgorithm`: data points count increased by 2, which is expected:  
  - The algorithm is adding option contracts in `OnData`, being one of the conditions `x.ID.StrikePrice - underlyingPrice > 0`
  - When `OnData` is called with the split event (occurred), the underlying price was not adjusted, making it way over the strike price; no contract was found or added. 
  - Now, the price is adjusted since the split has already occurred. A contract is found at that point and added to the securities. But since the algorithm is about to end (the algorithm end date is the same as the split date), we only get 2 additional data points in the count.
- `OptionChainedAndUniverseSelectionRegressionAlgorithm`: drawdown stat change, also expected:
  - Drawdown is calculated from the algorithm equity, which is calculated from the total portfolio value.
  - At the point of the split, the drawdown is calculated with the old price (not adjusted) since that is a bug fixed here as mentioned in the description.
  - The only algorithm equity value that changes is the first one calculated after the split, because it was using the old price without adjustment, which is casing the drawdown change.
image
- `BasicTemplateOptionsFrameworkAlgorithm`: big data point count and stats change, but also expected:
  - This algorithm actually reproduces the same bug as the one attatched to the related issue: right after the split, we get `AlgorithmManager.Run(): Portfolio value is less than or equal to zero, stopping algorithm.` because the total portfolio value is calculated with the wrong holdings price.
  - Since fixed, the algorithm is allowed to continue: the framework is selecting more contracts and placing more orders.

#### Related Issue

Closes #7894 

- Unit tests
- Regression algorithms
- Existing regression algorithms and unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/BasicTemplateOptionsFrameworkAlgorithm.cs
+++ b/Algorithm.CSharp/BasicTemplateOptionsFrameworkAlgorithm.cs
@@ -139,7 +139,7 @@ namespace QuantConnect.Algorithm.CSharp
         /// <summary>
         /// Data Points count of all timeslices of algorithm
         /// </summary>
-        public long DataPoints => 993927;
+        public long DataPoints => 1847643;
 
         /// <summary>
         /// Data Points count of the algorithm history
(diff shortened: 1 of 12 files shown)


Martin-Molinero (approved):
Nice! Thank you 👍

PR #8088, opened 2024-06-12
Option Strategy Filter Universe

LouisSzeto:
#### Description
Add helper method to filter option universe by shortlisting a particular type of option strategies
e.g. `option.SetFilter(u => u.IncludeWeeklys().Strike(-2, 2).Expiration(0, 90).IronCondor(30, 5, 10)` will return 4 contracts that best fit the criteria given and be able to form an Iron Condor.

#### Related Issue
NA

#### Motivation and Context
For users that only wish to obtain certain contracts for forming an option strategy, it can reduce the number of option contract data subscribed to speed up the algorithm and avoid further filtering manually.
e.g. A single-day Iron Condor algorithm yielded a near 50% speed improvement:
- with `.IronCondor(30, 5, 10)` filter
  image
- without helper filter
  image

- Add table in option universe filter page
- Add h3 section for each option strategy page

- unit tests
- regression tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
+++ b/Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
@@ -84,7 +84,7 @@ namespace QuantConnect.Algorithm.CSharp
         /// <summary>
         /// This is used by the regression test system to indicate which languages this algorithm is written in.
         /// </summary>
-        public Language[] Languages { get; } = { Language.CSharp };
+        public List<Language> Languages { get; } = new List<Language> { Language.CSharp };
 
         /// <summary>
         /// Data Points count of all timeslices of algorithm
--- a/Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
+++ b/Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
@@ -21,6 +21,8 @@ using QuantConnect.Data;
 using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
 using QuantConnect.Securities;
+using QuantConnect.Securities.Option;
+using QuantConnect.Securities.Positions;
 
 namespace QuantConnect.Algorithm.CSharp
 {
@@ -45,6 +47,16 @@ namespace QuantConnect.Algorithm.CSharp
 
             // set our strategy filter for this option chain
             option.SetFilter(_func);
+        }
+
+        protected void AssertOptionStrategyIsPresent(string name, int? quantity = null)
+        {
+            if (Portfolio.Positions.Groups.Where(group => group.BuyingPowerModel is OptionStrategyPositionGroupBuyingPowerModel)
+                .Count(group => ((OptionStrategyPositionGroupBuyingPowerModel)@group.BuyingPowerModel).ToString() == name
+                    && (!quantity.HasValue || Math.Abs(group.Quantity) == quantity)) != 1)
+            {
+                throw new RegressionTestException($"Option strategy: '{name}' was not found!");
+            }
         }
 
         public override void OnData(Slice slice)
@@ -83,6 +95,11 @@ namespace QuantConnect.Algorithm.CSharp
         /// Data Points count of the algorithm history
         /// </summary>
         public virtual int AlgorithmHistoryDataPoints => 0;
+
+        /// <summary>
+        /// Final status of the algorithm
+        /// </summary>
+        public AlgorithmStatus AlgorithmStatus => AlgorithmStatus.Completed;
 
         /// <summary>
         /// This is used by the regression test system to indicate what the expected statistics are from running the algorithm
--- a/Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
+++ b/Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
@@ -1,0 +1,92 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using QuantConnect.Interfaces;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Base class for equity option strategy filter universe regression algorithms which holds some basic shared setup logic
+    /// </summary>
+    public abstract class OptionStrategyFilteringUniverseBaseAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
+    {
+        protected Func<OptionFilterUniverse, OptionFilterUniverse> _func;
+        protected Symbol _optionSymbol;
+        protected int _expectedCount = 0;
+
+        public override void Initialize()
+        {
+            SetStartDate(2015, 12, 24);
+            SetEndDate(2015, 12, 24);
+            SetCash(200000);
+
+            var equity = AddEquity("GOOG", leverage: 4);
+            var option = AddOption(equity.Symbol);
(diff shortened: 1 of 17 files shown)


Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -224,6 +225,483 @@ public OptionFilterUniverse PutsOnly()
...
+
+            if (nearDaysTillExpiry < 0)
+            {
+                throw new ArgumentException("CalendarSpread(): near expiry argument must be positive.");
+            }
+            
+            // Select the set strike
+            var strike = AllSymbols.OrderBy(x => Math.Abs(Underlying.Price - x.ID.StrikePrice + strikeFromAtm))
+                .First().ID.StrikePrice;
+            var contracts = AllSymbols.Where(x => x.ID.StrikePrice == strike && x.ID.OptionRight == right);
```
Martin-Molinero, line 366: Cool!
Couple of thoughts:
- Believe it's a bit sad & fragile having to duplicate the knowledge of each strategy again here, we could have a core implementation which could take a `OptionStrategyDefinition` instance for example and generically handle it there, each of these sugar names could just pass their option strategy definition to the core one and boom, we can simplify many lines and potential issues due to the duplication
- The implementations seem a bit inconsistent on when using `AllSymbols` &/or the helper methods `Where/Contracts/etc` the helper methods are overriding `AllSymbols` & calling to List (sad to do it multiple times etc) -> I think we should avoid them and just use `AllSymbols` but at the **end** of the method call ToList and override it
- Just double checking, I see ur using Abs for date/strike input operations which is strange given the input could be +/- , seems if used in OrderBy it would treat +/- result in the same way which doesn't seem right?

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -224,6 +225,483 @@ public OptionFilterUniverse PutsOnly()
...
+                return this.WhereContains( new List<Symbol> () );
+            }
+            return filtered;
+        }
+
+        /// <summary>
+        /// Sets universe of a far-OTM call, a near-OTM call, a near-OTM put, and a far-OTM put with the same expiry 
+        /// and equal strike price distance between both calls and both puts, with closest match to the criteria given
+        /// </summary>
+        /// <param name="daysTillExpiry">The desire strike price distance from the current underlying price</param>
```
Martin-Molinero, line 603: The docs are off in some cases, please rereview in detail

Algorithm.CSharp/OptionStrategyFilteringUniverseBoxSpreadRegressionAlgorithm.cs
```diff
@@ -0,0 +1,123 @@
...
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm of filtering with Box Spread option strategy and asserting it's being detected by Lean and works as expected
+    /// </summary>
+    public class OptionStrategyFilteringUniverseBoxSpreadRegressionAlgorithm : OptionStrategyFilteringUniverseBaseAlgorithm
+    {
+        public override void Initialize()
+        {
+            _func = u => u.IncludeWeeklys().Strikes(-10, +10).Expiration(0, 180).BoxSpread(30, 5);
```
Martin-Molinero, line 32: Isn't `Expiration` filter overlapping with the `daysTillExpiry` in the option strategy? For example in this example if Expiration was (30, 180) 🤔
Makes me think that `daysTillExpiry` would only be used if `Expiration` filter isn't used? which could mean we can push it to the end of the option strategy as an optional probably defaulting to null, maybe if `Expiration` wasn't use use 30 instead as default

Martin-Molinero (changes requested):
Nice! Leaving some thoughts and requests

Algorithm.CSharp/OptionStrategyFilteringUniverseBoxSpreadRegressionAlgorithm.cs
LouisSzeto: It is intentional: `Expiration` is a strict rule, but `daysTillExpiry` is looking for the best match only🤔 I don't think they are mutaully exclusive.

Common/Securities/Option/OptionFilterUniverse.cs
LouisSzeto: > Just double checking, I see ur using Abs for date/strike input operations which is strange given the input could be +/- , seems if used in OrderBy it would treat +/- result in the same way which doesn't seem right?
Yah that was intentional. We look for the best match but not a conditional best match

Algorithm.CSharp/OptionStrategyFilteringUniverseBoxSpreadRegressionAlgorithm.cs
Martin-Molinero: > Expiration is a strict rule, but daysTillExpiry is looking for the best match only🤔 

Sounds like a bug if user provides 30 `daysTIllExpiry` and it can use something bellow that?

Martin-Molinero:
Please rebase from master, the regression test interface was changed & a unit test CI bug was shipped

Algorithm.CSharp/OptionStrategyFilteringUniverseBoxSpreadRegressionAlgorithm.cs
Martin-Molinero: > .Strikes(-10, +10).Expiration(0, 180)

Let's remove this from the examples

~Sounds like a bug if user provides 30 daysTIllExpiry and it can use something bellow that?~ nevermind 👍

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -224,6 +225,483 @@ public OptionFilterUniverse PutsOnly()
...
+        }
+
+        private OptionFilterUniverse SingleContract(OptionRight right, int daysTillExpiry = 30, decimal strikeFromAtm = 0)
+        {
+            // Select the expiry as the nearest to set days later
+            var expiry = AllSymbols.OrderBy(x => Math.Abs((x.ID.Date - _lastExchangeDate.AddDays(daysTillExpiry)).Days))
+                .First().ID.Date;
+            var contracts = AllSymbols.Where(x => x.ID.Date == expiry && x.ID.OptionRight == right);
+            // Select strike price
+            var selected = contracts.OrderBy(x => Math.Abs(x.ID.StrikePrice - Underlying.Price - strikeFromAtm)).First();
```
Martin-Molinero, line 260: I don't think the expiration filter here is behaving as expected, take the follow example

```
DaysTillExpiry 4
Today 8

Contract 1 expiration date 9
Contract 2 expiration date 15
Contract 3 expiration date 20

9 - (8 + 4) =>  Abs(-3) ? -> HERE
15 - (8 + 4) => Abs(3) -> HERE
20 - (8 + 4) => Abs(8)
```

We should have 1 centralized helper method location where we handle these daysTillExpiry/strikeFromAtm and reuse them in the whole file

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -224,6 +225,483 @@ public OptionFilterUniverse PutsOnly()
...
+        }
+
+        /// <summary>
+        /// Sets universe of a single put contract with the closest match to criteria given
+        /// </summary>
+        /// <param name="daysTillExpiry">The desire days till expiry from the current time</param>
+        /// <param name="strikeFromAtm">The desire strike price distance from the current underlying price</param>
+        /// <remarks>Applicable to Naked Put, Covered Put, and Protective Put Option Strategy</remarks>
+        /// <returns>Universe with filter applied</returns>
+        public OptionFilterUniverse SinglePut(int daysTillExpiry = 30, decimal strikeFromAtm = 0)
```
Martin-Molinero, line 248: All of these strategy helpers are `Single`, should we just rename this to `NakedPut` & `NakedCall` helper methods?

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -224,6 +225,496 @@ public OptionFilterUniverse PutsOnly()
...
+        }
+
+        private DateTime GetExpiry(IEnumerable<Symbol> symbols, int minDaysTillExpiry)
+        {
+            var leastExpiryAccepted = _lastExchangeDate.AddDays(minDaysTillExpiry);
+            var accepted = symbols.Where(x => x.ID.Date >= leastExpiryAccepted);
+            if (!accepted.Any())
+            {
+                return DateTime.MaxValue;
+            }
```
Martin-Molinero, line 706: Let's just remove this and perform a FIrstOrDefault bellow to avoid enumerating twice

Algorithm.CSharp/OptionStrategyFilteringUniverseBaseAlgorithm.cs
```diff
@@ -0,0 +1,92 @@
...
+        public override void OnData(Slice slice)
+        {
+            if (!Portfolio.Invested)
+            {
+                OptionChain chain;
+                if (slice.OptionChains.TryGetValue(_optionSymbol, out chain) && chain.Any())
+                {
+                    TestFiltering(chain);
+                }
+            }
```
Martin-Molinero, line 59: Let's add an assertion of the option strategy after we trade it, in the portfolio see other regression algos that perform this already

PR #8149, opened 2024-07-05
Implement new indicator McGinleyDynamic

femtotrader:
#### Description

Implement new indicator - McGinley Dynamic

#### Related Issue
Closes #8127 

#### Motivation and Context

McGinley Dynamic wasn't implemented in QuantConnect

#### Requires Documentation Change

Doc for this indicator should be created

#### How Has This Been Tested?

comparison with talipp computation of ZLEMA(5) of SPY close prices

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves implementation)
- [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
- [x] I have added tests to cover my changes. 
- [x] All new and existing tests passed.
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

PR #8179, opened 2024-07-09
Refactors TrueRange and AverageTrueRange

### LINKED ISSUE #8171: Refactor AverageTrueRange code with TrueRange code
#### Expected Behavior

Avoid repeating code

#### Actual Behavior

https://github.com/QuantConnect/Lean/blob/2941210d20c45636b73d19cbba682345ceb10a23/Indicators/TrueRange.cs#L72-L84

https://github.com/QuantConnect/Lean/blob/2941210d20c45636b73d19cbba682345ceb10a23/Indicators/AverageTrueRange.cs#L100-L112

Looks very similar

#### Potential Solution

either

- uses TrueRange inside AverageTrueRange

or (probably harder in C#)

- creates a SmootherFactory which takes an indicator class and a moving average and return a "SmoothedIndicator" ie an indicator which can return values of internal indicator using the moving average which was defined.

This concept was experimented in Python with talipp in this PR https://github.com/nardew/talipp/pull/147/files

#### Reproducing the Problem

not concerned

#### Checklist

femtotrader:
#### Description

use TrueRange in AverageTrueRange

#### Related Issue

Closes #8171

#### Motivation and Context

Avoid code duplication

Tests AverageTrueRangeTests and TrueRangeTests are passing correctly.
Changes may have side effect on other tests (untested)

#### Types of changes

#### Checklist:


Indicators/TrueRange.cs
```diff
@@ -63,10 +63,10 @@ public TrueRange(string name)
         /// <returns>A new value for this indicator</returns>
         protected override decimal ComputeNextValue(IBaseDataBar input)
         {
-            if (!IsReady)
+            if (Samples == 1)
             {
                 _previousInput = input;
-                return 0m;
+                return input.High - input.Low;
```
Martin-Molinero, line 69: Returning `input.High - input.Low` seems okay when `samples == 1`, it does not mean the indicator is ready and warmed up though, it requires 1 extra data point given it's comparing current with previous, don't agree with the changes at WarmUpPeriod/IsReady

Think `AverageTrueRange` above should probably look whether TrueRange is ready on IsReady too

Martin-Molinero (changes requested):
Cool cleanup!
Not sure about these changes though, see original addition of TR https://github.com/QuantConnect/Lean/commit/7042411d9f3e21bf4df9a58afa8455b462c7b4b7

Indicators/TrueRange.cs
femtotrader: As exposed https://github.com/QuantConnect/Lean/pull/8149#discussion_r1670991914 my opinion is that "Ready" or not is not enough to describe an indicator state... this PR was a attempt to show case how a refactor could be done but you are right for not accepting in its current state.

I think at least 3 states are required.

 - Cold
 - WarmingUp
 - Ready

Cold could be indicator default state when no data have been received
WarmingUp when 1 data have been received (in the case of TrueRange)
Ready when 2 or more data have been received

Any opinion?

Martin-Molinero: Hey! Agree that potentially there could be other states (even failing/invalid), but if we did that, it should be handled generically for all indicators, don't think it's the scope of these changes. Let's follow current pattern, IsReady means it's fully ready, which is expected to be true when received the WarmupPeriod amount of data count, not before

Martin-Molinero: > Agree that potentially there could be other states (even failing/invalid), but if we did that, it should be handled generically for all indicators, don't think it's the scope of these changes/pr.

I see u added state for the AvgTR. This should be solve generically in it's own PR, not related to this one, after a design/brain storm I think

Issue #6861, opened 2023-01-14
Create Future & Option Universe Files


Martin-Molinero:
#### Expected Behavior

- Use universe files for  future and option universe selection and backtesting chain provider

#### Actual Behavior

- We are currently reading from the price data the available contracts.
- See https://github.com/QuantConnect/Lean/issues/6853
#### Potential Solution

- Create new universe files for future and options which hold the contracts for each day
#### Reproducing the Problem

N/A
#### System Information

N/A
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

- [x] I have provided detailed steps to reproduce the issue

Martin-Molinero:
See https://github.com/QuantConnect/Lean/issues/6863

Martin-Molinero:
Option universe files will be covered while solving https://github.com/QuantConnect/Lean/issues/1518

Martin-Molinero:
Pending todo Future and FOPs universe files

PR #8212, opened 2024-07-18
File-based options universe

jhonabreul:
#### Description

Refactor Equity and Index Option universe selection:
- Now file-based: option chains are directly read from files instead of being built from price data file names.
- Option contracts include more data at selection time, which means they can be filtered by their data:
  - Data includes Greeks, Implied Volatility and Open Interest.
  - New filters where added to allow this.
- History requests can now be used to fetch this options data, which would be like getting a full unfiltered option chain with the new additional data.

Notes about some algorithms with changed stats:
- Index `BasicTemplateSPXWeeklyIndexOptionsAlgorithm`:
  - Stats change because since options universe data is daily, an additional `Sample()` calls is made on 2021/1/9 midnight
- `IndexOptionScaledStrikeRegressionAlgorithm`:
  - Orders and stats change because now selection is immediate, so orders are place for the exact same options but an hour earlier, and price is different at that time.
- `BasicTemplateOptionsFrameworkAlgorithm`:
  - Same contracts traded, just at different time of day with a different price

#### Related Issue

Part of #6861 

#### Motivation and Context

- Performance improvements for options algorithms.
- Greeks and other data at selection time, which allows improved filtering.

New filters should be documented

- Exisiting regression algorithms and unit tests.
- New regression algorithms and unit tests.
- Cloud tests:
  - Sample options algorithms
  - Live deployments running thought multiple days with a bunch of logs:
    - Look that first selection happens right after deployment.
    - Look that selection happens everyday after data is available.
 
##### 1. Basic speed test:

Algorithm with a span of 4 years adding SPX options with minute resolution:

###### Case 1: Selecting 0 contracts in the filter function

This case allows us to test the chain loading and selection speed, without price data.

- [Master](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_ac78e40e39a772b61cab551391ad7a5a.html): ~15s
- [Branch](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_c2c019f9ca76f3d95d2969fcd714367d.html): ~4s

###### Case 2: Selecting up to 10 contracts each date

- [Master](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_c2270df6c5e69e1e99bc838e182953e2.html): ~34s
- [Branch](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_dfd716accef1df7ae04b67e8dd5e4f53.html): ~34s

Not that big of a difference once contracts are selected, but pure contracts loading is much faster.

##### 2. Filtering speed test:

SPY options algorithm of 6 month with daily resolution, doing some work in the filter function to get implied volatility values for filtering:
- Master: fetching historical data in the filter function, feeding it to an `ImpliedVolatility` indicator and using the resulting IV values to do filtering.
- Branch: using pre-calculated implied volatility in `OptionUniverse` instances.

###### Case 1: Selecting all contracts with IV in [50% IV Median, 150% IV Median]

- [Master](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_654cf6c8791ec84c791ec0cbb81d48ed.html): ~300s.
- [Branch](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_49b875ebd7378d4ca92170277cac80da.html): ~11s.

A time reduction of about 30%.

###### Case 2: Doing the IV calculations, but selecting 0 contracts.

This case allows us to test speed difference for option contracts loading and filtering only, without price data fetching and loading taken into consideration.

- [Master](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_fd623ba54a1bb8df7a3b4f1d271237d9.html): ~300s
- [Branch](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_69ddfad66aecb2b777de2d1283563017.html): ~1.5s

Selection based on greeks is now dramatically faster and simpler.

#### Types of changes

#### Checklist:


Algorithm.CSharp/AddRemoveOptionUniverseRegressionAlgorithm.cs
```diff
@@ -110,20 +111,21 @@ public override void OnData(Slice slice)
...
                     {
-                        throw new RegressionTestException($"Unexpected underlying data point {u.Underlying.EndTime} {u.Underlying}");
+                        throw new RegressionTestException($"Unexpected selection time {u.LocalTime}");
                     }
                     // find first put above market price
                     return u.IncludeWeeklys()
                         .Strikes(+1, +3)
                         .Expiration(TimeSpan.Zero, TimeSpan.FromDays(1))
-                        .Contracts(c => c.Where(s => s.ID.OptionRight == OptionRight.Put));
+                        .Contracts(c => c.Select(x => x.Symbol).Where(s => s.ID.OptionRight == OptionRight.Put));
```
Martin-Molinero, line 122: We don't need `.Select(x => x.Symbol)` given `ID` is on the OptionUniverse data too?

Algorithm.CSharp/BasicTemplateOptionsFilterAlgorithm.cs
```diff
@@ -0,0 +1,139 @@
...
+using QuantConnect.Data.Market;
+using QuantConnect.Data.UniverseSelection;
+using QuantConnect.Securities;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm with new proposed option filter API using new options universe data (greeks, implied volatility, open interest, etc).
+    /// </summary>
+    public class BasicTemplateOptionsFilterAlgorithm : QCAlgorithm
```
Martin-Molinero, line 28: Nice! think we should make it a regression algorithm to assert some of these filters, maybe can add a couple of equities and their chains?

Algorithm.CSharp/CoarseFineOptionUniverseChainRegressionAlgorithm.cs
```diff
@@ -61,7 +67,7 @@ public override void Initialize()
                 }
                 return universe.IncludeWeeklys()
                     .FrontMonth()
-                    .Contracts(universe.Take(5));
+                    .Contracts(universe.Take(5).Select(x => x.Symbol));
```
Martin-Molinero, line 70: `.Select(x => x.Symbol)` shouldn't be required here either?

Common/Data/UniverseSelection/OptionChainUniverse.cs
```diff
@@ -99,15 +99,19 @@ public override IEnumerable<Symbol> SelectSymbols(DateTime utcTime, BaseDataColl
                 return Unchanged;
             }
 
-            var availableContracts = data.Data.Select(x => x.Symbol);
+            // Temporary: this can be removed when future options universe selection is also file-based
+            var availableContractsData = data.Symbol.SecurityType != SecurityType.FutureOption
+                ? data.Data.Cast<OptionUniverse>()
+                : data.Data.Select(x => new OptionUniverse() { Symbol = x.Symbol, Underlying = data.Underlying, Time = x.Time });
```
Martin-Molinero, line 105: Should set time & endtime I think?👌

Common/Data/UniverseSelection/OptionChainUniverse.cs
```diff
@@ -99,15 +99,19 @@ public override IEnumerable<Symbol> SelectSymbols(DateTime utcTime, BaseDataColl
...
+
             // we will only update unique strikes when there is an exchange date change
-            _optionFilterUniverse.Refresh(availableContracts, data.Underlying, localEndTime);
+            _optionFilterUniverse.Refresh(availableContractsData, data.Underlying, localEndTime);
 
-            var results = Option.ContractFilter.Filter(_optionFilterUniverse);
+            var results = Option.ContractFilter.Filter(_optionFilterUniverse).Cast<OptionUniverse>();
```
Martin-Molinero, line 110: `.Cast<OptionUniverse>();` should the return type of filter be of T so we can avoid the cast?
Should the `OptionFilterUniverse` be able to convert to implicit symbol enumerable? I see there's `public static implicit operator List<Symbol>(OptionFilterUniverse universe)` why list?

Common/Securities/ISymbol.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Gets the security identifier.
+        /// </summary>
+        SecurityIdentifier ID { get; }
+
+        /// <summary>
+        /// Gets the symbol.
+        /// </summary>
+        /// <returns>The symbol</returns>
+        Symbol GetSymbol();
```
Martin-Molinero, line 33: Hm not sure the GetSymbol makes sense, why do we need it?

Common/Securities/Option/OptionFilterUniverse.cs
```diff
@@ -853,9 +1029,22 @@ public static class OptionFilterUniverseEx
...
+
+        /// <summary>
+        /// Filters universe
+        /// </summary>
+        /// <param name="universe">Universe to apply the filter too</param>
+        /// <param name="predicate">Bool function to determine which Symbol are filtered</param>
+        /// <returns>Universe with filter applied</returns>
+        public static OptionFilterUniverse Where(this OptionFilterUniverse universe, PyObject predicate)
+        {
+            using var _ = Py.GIL();
```
Martin-Molinero, line 1046: Don't think we need the Gil here, same below, since `ConvertToDelegate` internally takes it already?

Common/Securities/ContractSecurityFilterUniverse.cs
```diff
@@ -59,11 +63,45 @@ protected enum ContractExpirationType : int
...
+        internal IEnumerable<Symbol> AllSymbols
+        {
+            get
+            {
+                return _allSymbols;
+            }
+            set
+            {
+                _allSymbols = value;
+                _data = _data.Where(x => _allSymbols.Contains(x.GetSymbol()));
```
Martin-Molinero, line 102: Think maybe we should make symbols basically be a passthrough to data now, instead of having 2 sources of  truth here

Tests/Algorithm/AlgorithmAddDataTests.cs
```diff
@@ -79,47 +80,47 @@ public void DefaultDataFeeds_AreAdded_Successfully()
...
+            Assert.IsNotNull(GetMatchingSubscription(algo, forex.Symbol, typeof(QuoteBar)));
 
             // equity high resolution
             var equityMinute = algo.AddSecurity(SecurityType.Equity, "goog");
-            Assert.IsTrue(equityMinute.Subscriptions.Count() == 2);
-            Assert.IsTrue(GetMatchingSubscription(algo, equityMinute.Symbol, typeof(TradeBar)) != null);
-            Assert.IsTrue(GetMatchingSubscription(algo, equityMinute.Symbol, typeof(QuoteBar)) != null);
+            Assert.AreEqual(2, equityMinute.Subscriptions.Count());
+            Assert.IsNotNull(GetMatchingSubscription(algo, equityMinute.Symbol, typeof(TradeBar)));
+            Assert.IsNotNull(GetMatchingSubscription(algo, equityMinute.Symbol, typeof(QuoteBar)));
```
Martin-Molinero, line 90: Strange to change this in such a big PR already 😅  I see now real change here?

Research/QuantBook.cs
```diff
@@ -428,13 +428,13 @@ public dynamic GetFundamental(string ticker, string selector = null, DateTime? s
                 }
 
                 var optionFilterUniverse = new OptionFilterUniverse(option);
-                var distinctSymbols = allSymbols.Distinct();
+                var distinctSymbols = allSymbols.Distinct().Select(x => new OptionUniverse() { Symbol = x, Time = start});
```
Martin-Molinero, line 431: Think we should add a todo to cleaned this up a bit, for when we tackle futures, above we are using `OptionChainProvider.GetOptionContractList` which will basically a history request to the universe type and here we are recreating them again,

Engine/DataFeeds/DataManager.cs
```diff
@@ -222,7 +222,12 @@ public class DataManager : IAlgorithmSubscriptionManager, IDataFeedSubscriptionM
                             // If it is an add we will set time 1 tick ahead to properly sync data
                             // with next timeslice, avoid emitting now twice.
                             // We do the same in the 'TimeTriggeredUniverseSubscriptionEnumeratorFactory' when handling changes
-                            AddSubscription(new SubscriptionRequest(request, startTimeUtc: algorithm.UtcTime.AddTicks(1)));
+                            var startUtc = algorithm.UtcTime;
+                            if (!algorithm.GetLocked() && request.StartTimeUtc < startUtc)
+                            {
+                                startUtc = request.StartTimeUtc;
```
Martin-Molinero, line 228: Should add a comment why/what's going on here?

Common/Securities/IDerivativeSecurityFilterUniverse.cs
```diff
@@ -14,14 +14,29 @@
...
 
 namespace QuantConnect.Securities
 {
+#pragma warning disable CA1010, CA1710 // Generic interface should also be implemented
+#pragma warning disable CA1710 // Identifiers should have correct suffix
+
+    /// <summary>
+    /// Represents derivative symbols universe used in filtering.
+    /// </summary>
+    public interface IDerivativeSecurityFilterUniverse : IEnumerable
```
Martin-Molinero, line 28: Are u sure we need this base interface don't think it's used, the simpler the merrier 🤔

Algorithm.Python/OptionUniverseFilterOptionsDataRegressionAlgorithm.py
```diff
@@ -0,0 +1,54 @@
...
+                    # Can access the contract data here and do some filtering based on it is needed.
+                    # More complex math can be done here for filtering, but will be simple here for demonstration sake:
+                    if (contract.Greeks.Delta > 0.5 and contract.Greeks.Delta < 1.5 and
+                        contract.Greeks.Gamma > 0.0001 and contract.Greeks.Gamma < 0.0006 and
+                        contract.Greeks.Vega > 0.01 and contract.Greeks.Vega < 1.5 and
+                        contract.Greeks.Theta > -2.0 and contract.Greeks.Theta < -0.5 and
+                        contract.Greeks.Rho > 0.5 and contract.Greeks.Rho < 3.0 and
+                        contract.ImpliedVolatility > 1.0 and contract.ImpliedVolatility < 3.0 and
+                        contract.OpenInterest > 100 and contract.OpenInterest < 500)
+                    ]))
```
Martin-Molinero, line 54: Nice!
Similar comment, do we need the `.Symbol` ?

Algorithm.CSharp/OptionUniverseFilterGreeksRegressionAlgorithm.cs
```diff
@@ -0,0 +1,148 @@
...
+            */
+        }
+
+        public override void OnData(Slice slice)
+        {
+            if (slice.OptionChains.TryGetValue(_optionSymbol, out var chain) && chain.Contracts.Count > 0)
+            {
+                Log($"[{Time}] :: Recieved option chain with {chain.Contracts.Count} contracts");
+                _optionChainReceived = true;
+            }
```
Martin-Molinero, line 78: Cool!
To be safe, think would be nice to assert the chain is respecting what we requested for the different attributes, same for python 🙏

Martin-Molinero (commented):
Thank you! Nice work, leaving a few minor comments 👍

Common/Securities/IDerivativeSecurityFilterUniverse.cs
jhonabreul: This was needed to simplify casting from derived filter universe clases to their interface in methods like `https://github.com/QuantConnect/Lean/blob/master/Common/Securities/FuncSecurityDerivativeFilter.cs#L37-L38`, since casting from base classes into base generic interfaces requires knowing the generic types.

Common/Data/UniverseSelection/OptionChainUniverse.cs
jhonabreul: OptionUniverse` set EndTime on its own 
image

Common/Securities/IDerivativeSecurityFilterUniverse.cs
jhonabreul: The interface was removed 👌

PR #8228, opened 2024-07-25
Add more options to Field class

### LINKED ISSUE #6117: Adds More Options to Field Class
#### Expected Behavior
Be able to use more attributes than those that are available, e.g. BidPrice.

#### Actual Behavior
`BidPrice` options is not available.

#### Potential Solution
Potentially `BidPrice` with `Field.BidClose`. It will give additional 10 fields because of `Bid` and `Ask`. 
Or the selectors of the indicators could accept a lambda function.  

#### Reproducing the Problem
`SMA(symbol, 9, Resolution.Minute, Field.BidPrice)`

#### Checklist

Marinovsky:
#### Description

- Add more options to Field class so that the user can select properties from QuoteBars and Tick.
- Improve the implementation of the options in the Field class to avoid object creation each time certain property is called.
- Update `QCAlgorithm.Indicators` to handle the case where the selector requires the data type to be quote.
- Add regression and unit tests to cover these changes.
#### Related Issue

Closes #6117 
#### Motivation and Context

With this change users will be able to use more attributes than those that are available, e.g. BidPrice.

This change might require an update to documentation briefing the users on the new Field options.

I created two regression algorithms (in C# and Python) that create identity indicators with the new introduced selectors (using equity and option symbols). During the algorithm, it asserts that the value of the indicator matches with the given data in the slice (since the indicator is an identity indicator). I also created unit tests asserting the expected behavior but also the case where the resolution doesn't match with the selector data type. In that last case, the unit test asserted that a default selector was used.
#### Types of changes

#### Checklist:


Algorithm.CSharp/IndicatorSelectorsWorkWithDifferentOptions.cs
```diff
@@ -0,0 +1,166 @@
...
+                    return x.Current.Value != value;
+                });
+
+                if (wrongEquityIndicators.Any())
+                {
+                    throw new RegressionTestException();
+                }
+            }
+
+            if (slice.OptionChains.ContainsKey(_option.Canonical) && slice.OptionChains[_option.Canonical].TradeBars.ContainsKey(_option))
```
jhonabreul, line 84: Just a thought for good practices, perhaps doing `slice.OptionChains[_option.Canonical].TradeBars.TryGetValue(_option)` instead of `ContainsKey` + `.TradeBars[_option]`

Common/Field.cs
```diff
@@ -117,5 +217,20 @@ public static partial class Field
...
+        {
+            return x =>
+            {
+                var bar = x as Tick;
+                if (bar != null)
+                {
+                    return selector(bar);
+                }
+
+                defaultSelector = defaultSelector ?? (data => data.Value);
```
jhonabreul, line 231: nit - use `??=`

Common/Field.cs
```diff
@@ -24,28 +24,128 @@ namespace QuantConnect
...
+        private readonly static Func<IBaseData, decimal> _weighted = BaseDataBarPropertyOrValue(x => (x.High + x.Low + 2 * x.Close) / 4m);
+        private readonly static Func<IBaseData, decimal> _sevenBar = BaseDataBarPropertyOrValue(x => (2 * x.Open + x.High + x.Low + 3 * x.Close) / 7m);
+
         /// <summary>
         /// Gets a selector that selects the Open value
         /// </summary>
         public static Func<IBaseData, decimal> Open
         {
-            get { return BaseDataBarPropertyOrValue(x => x.Open); }
+            get { return _open; }
```
jhonabreul, line 52: nit - leave High, Low and Close properties next to open just for clearness and organization

Common/Field.cs
```diff
@@ -24,28 +24,128 @@ namespace QuantConnect
...
+        private readonly static Func<IBaseData, decimal> _bidLow = BaseDataBarPropertyOrValue(x => ((QuoteBar)x).Bid.Low);
+        private readonly static Func<IBaseData, decimal> _bidHigh = BaseDataBarPropertyOrValue(x => ((QuoteBar)x).Bid.High);
+        private readonly static Func<IBaseData, decimal> _askClose = BaseDataBarPropertyOrValue(x => ((QuoteBar)x).Ask.Close);
+        private readonly static Func<IBaseData, decimal> _askOpen = BaseDataBarPropertyOrValue(x => ((QuoteBar)x).Ask.Open);
+        private readonly static Func<IBaseData, decimal> _askLow = BaseDataBarPropertyOrValue(x => ((QuoteBar)x).Ask.Low);
+        private readonly static Func<IBaseData, decimal> _askHigh = BaseDataBarPropertyOrValue(x => ((QuoteBar)x).Ask.High);
+        private readonly static Func<IBaseData, decimal> _bidPrice = TickPropertyOrValue(x => ((Tick)x).BidPrice);
+        private readonly static Func<IBaseData, decimal> _askPrice = TickPropertyOrValue(x => ((Tick)x).AskPrice);
+        private readonly static Func<IBaseData, decimal> _volume = BaseDataBarPropertyOrValue(x => ((TradeBar)x).Volume);
```
jhonabreul, line 40: These casts will throw if not the right type and the default selector will not be used, right?

jhonabreul (commented):
Nice, thank you! - Left a few minor comments

Common/Field.cs
Marinovsky: Yup, that's right

Martin-Molinero (approved):
Thank you! 🚀

PR #8262, opened 2024-08-15
Fix bug for history requests with custom symbols

### LINKED ISSUE #8259: Problem retrieving CSV custom symbol history
#### Expected Behavior
Retrieve the quote history in a dataframe of a local CSV file on research notebook.

#### Actual Behavior
The dataframe is returning empty.

#### Potential Solution
The research docker image 16484 works fine, the bug started on the docker image 16487.
This [PR](https://github.com/QuantConnect/Lean/commit/621d1252b9c221a6442b11b084fdfb2e815da2cd) changing the `QCAlgorihtm.Python.History()` is from the same day of the 16487 image so maybe it introduced the bug.

#### Reproducing the Problem
Create the file `test1.csv` in `Lean/data` folder, fill it with some data rows.
Using the 16487 image and so on the `history` dataframe returns empty.

```
from datetime import datetime

class TestTradeBar(TradeBar):
    def get_source(self, config: SubscriptionDataConfig, date: datetime, is_live_mode: bool) -> SubscriptionDataSource:
        return SubscriptionDataSource(source=f'{Globals.data_folder}/test1.csv',
                                      transportMedium=SubscriptionTransportMedium.LOCAL_FILE,
                                      format=FileFormat.CSV)

    def reader(self, config: SubscriptionDataConfig, line: str, date: datetime, is_live_mode: bool) -> BaseData:
        if not line[0].isdigit():
            return None
        data = line.split(',')
        bar_time = datetime.utcfromtimestamp(int(data[0]))

        open = float(data[1])
        high = float(data[2])
        low = float(data[3])
        close = float(data[4])
        volume = int(float(data[7]))
        return TradeBar(bar_time, config.symbol, open, high, low, close, volume)


qb = QuantBook()
qb.add_data(
    type=TestTradeBar,
    ticker='TEST1',
    properties=SymbolProperties(
        description='TEST1',
        quoteCurrency='USD',
        contractMultiplier=1,
        minimumPriceVariation=0.01,
        lotSize=1,
        marketTicker='TEST1',
    ),
    exchange_hours=SecurityExchangeHours.always_open(TimeZones.NEW_YORK),
    resolution=Resolution.MINUTE,
    fill_forward=True,
    leverage=1,
)
history = qb.history(qb.securities.keys(), datetime(2024, 8, 2), datetime(2024, 8, 3))
history
```

Marinovsky:
#### Description

When the symbol used for the history request was custom, even though the data was obtained, it was discarded in the method `PandasConverter.AddSliceDataTypeDataToDict()`. This happened since `PandasConverter.GetPandasDataValue` considered a given data to be custom only if the data type was custom. Therefore, if the custom symbol inherited from Tradebar, it wasn't considered as custom data. This PR aims to solve that bug and centralize an existent method that determines if a symbol is custom or not.
#### Related Issue

Closes #8259 
#### Motivation and Context

With this change users will be able to request history data for custom symbols in research notebooks

N/A

I used the algorithm reproducing the bug from the PR to make a unit test that asserted the bug was solved.
#### Types of changes

#### Checklist:


Algorithm/QCAlgorithm.History.cs
```diff
@@ -949,7 +949,7 @@ protected IEnumerable<HistoryRequest> CreateDateRangeHistoryRequests(IEnumerable
             DataNormalizationMode? dataNormalizationMode = null, int? contractDepthOffset = null)
         {
             symbols = symbols.ToArray();
-            return CreateDateRangeHistoryRequests(symbols, GetCustomDataTypeFromSymbols((Symbol[])symbols) ?? typeof(BaseData), startAlgoTz, endAlgoTz, resolution, fillForward, extendedMarketHours,
+            return CreateDateRangeHistoryRequests(symbols, Extensions.GetCustomDataTypeFromSymbols((Symbol[])symbols) ?? typeof(BaseData), startAlgoTz, endAlgoTz, resolution, fillForward, extendedMarketHours,
```
jhonabreul, line 952: Is this cast to `Symbol[]` safe? The input `IEnumerable` might not be an array

Common/Python/PandasData.cs
```diff
@@ -134,8 +135,8 @@ public PandasData(object data)
 
             var type = data.GetType();
             _isFundamentalType = type == typeof(Fundamental);
-            IsCustomData = type.Namespace != typeof(Bar).Namespace;
             _symbol = ((IBaseData)data).Symbol;
+            IsCustomData = type.Namespace != typeof(Bar).Namespace || Extensions.GetCustomDataTypeFromSymbols(new Symbol[] { _symbol }) != null ? true : false;
```
jhonabreul, line 139: 1. Only types in `QuantConnect.Data.Market` are cosidered "not custom"? What about `QuantConnect.Data`?
2. nit - the ternary operation might not be necessary: `!= null` is enough.
3. This check is also in `Algorithm/QCAlgorithm.History.cs:1087`: maybe move it to a single location like Extensions for single source of truth

jhonabreul (commented):
Nice, ty!
Leaving some minor comments below.

Common/Python/PandasData.cs
Marinovsky: 1. The other data types present in QuantConnect.Data are `BaseData` and `DynamicData`, but they are abstract, so we cannot instantiate an object of those classes. Then, it seems that the only custom types are the ones not present in `QuantConnect.Data.Market`.

jhonabreul (approved):
Thank you!

Martin-Molinero (approved):
🚀

PR #8267, opened 2024-08-21
Fix null exception thrown in SubscriptionManager.ScanPastConsolidators

### LINKED ISSUE #8248: NullReferenceException in ScanPastConsolidators
#### Expected Behavior
LEAN doesn't throw unhandled exceptions, if possible.

#### Actual Behavior
The error message:

> 2024-08-08T17:17:31.1648326Z ERROR:: Extensions.SetRuntimeError(): Extensions.SetRuntimeError(): RuntimeError at 08/08/2024 17:17:31 UTC. Context: AlgorithmManager.Run System.NullReferenceException: Object reference not set to an instance of an object.
   at QuantConnect.Data.SubscriptionManager.ScanPastConsolidators(DateTime newUtcTime, IAlgorithm algorithm) in /Common/Data/SubscriptionManager.cs:line 263
   at QuantConnect.Lean.Engine.AlgorithmManager.Run(AlgorithmNodePacket job, IAlgorithm algorithm, ISynchronizer synchronizer, ITransactionHandler transactions, IResultHandler results, IRealTimeHandler realtime, ILeanManager leanManager, CancellationToken token) in /Engine/AlgorithmManager.cs:line 202
   at QuantConnect.Lean.Engine.Engine.<>c__DisplayClass10_1.<Run>b__9() in /Engine/Engine.cs:line 337

suggests that `priority` is `null` at [SubscriptionManager.cs:line 263](https://github.com/QuantConnect/Lean/blob/master/Common/Data/SubscriptionManager.cs#L263):
```csharp
while (_consolidatorsSortedByScanTime.TryPeek(out _, out var priority) && priority.UtcScanTime < newUtcTime)
```

#### Potential Solution
Add null check:
```csharp
while (_consolidatorsSortedByScanTime.TryPeek(out _, out var priority) && priority != null && priority.UtcScanTime < newUtcTime)
```

#### Reproducing the Problem
N/A

#### System Information
QuantConnect Cloud live trading.

#### Checklist

Marinovsky:
#### Description

When the user added/removed consolidators many times, a race condition in `SubscriptionManager.ScanPastConsolidators` method happened. Specifically, the exception was thrown since the priority queue used for the consolidators was not thread safe, so while the method `ScanPastConsolidators()` was trying to peek the top of the priority queue, another thread was adding/removing elements from it. In order to fix this bug, we added a lock every time a new consolidator had to be added, and instead of adding it directly, we waited for the next call to `ScanPastConsolidators` to add it. A unit test reproducing this bug was also added.
#### Related Issue

Closes #8248 
#### Motivation and Context

With this change, there won't be any race condition when consolidators are being added at a high rate

N/A

I created a unit test with two tasks, one calling `ScanPastConsolidators()` and the other one adding and removing consolidators. This test asserted not only that the null exception wasn't thrown but also that at then of it, the expected amount of consolidators had been added and removed.
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/SubscriptionManager.cs
+++ b/Common/Data/SubscriptionManager.cs
@@ -23,6 +23,7 @@ using QuantConnect.Data.Market;
 using QuantConnect.Interfaces;
 using QuantConnect.Util;
 using QuantConnect.Python;
+using System.Collections.Concurrent;
 
 namespace QuantConnect.Data
 {
@@ -33,6 +34,7 @@ namespace QuantConnect.Data
     {
         private readonly PriorityQueue<ConsolidatorWrapper, ConsolidatorScanPriority> _consolidatorsSortedByScanTime;
         private readonly Dictionary<IDataConsolidator, ConsolidatorWrapper> _consolidators;
+        private ConcurrentQueue<ConsolidatorScanPriority> _topPriority;
         private readonly ITimeKeeper _timeKeeper;
         private IAlgorithmSubscriptionManager _subscriptionManager;
 
@@ -65,6 +67,7 @@ namespace QuantConnect.Data
             _consolidators = new();
             _timeKeeper = timeKeeper;
             _consolidatorsSortedByScanTime = new(1000);
+            _topPriority = new();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Martin-Molinero (commented):
Thank you! 🚀

PR #8287, opened 2024-08-27
Improve `PythonIndicator.IsReady` implementation

### LINKED ISSUE #6615: `IndicatorExtensions` cannot be used with `PythonIndicator`
#### Expected Behavior
`IndicatorExtensions` can be used with `PythonIndicator` for all method (`Of`, `Over`,...)

#### Actual Behavior
For `Of`:
It will throw error of `Cannot perform runtime binding on a null reference in PythonIndicator.cs:line 143`
It looks like it didn't read the PythonIndicator's value correctly.

For others:
The combination method didn't get called at all, just like https://github.com/QuantConnect/Lean/issues/3565#issuecomment-1113254800

#### Potential Solution
NA

#### Reproducing the Problem
https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_8bbd321995671f7f48427d27a4bd29d6.html

#### System Information
QC cloud

#### Checklist

Marinovsky:
#### Description

Since the property `BaseIndicator.IsReady` does not define a setter, when a custom indicator (which inherited from PythonIndicator) was updated manually, the property `IsReady` never became true as it was handled internally by `PythonIndicator` class. This bug was noticeable when using the method `IndicatorExtensions.Of` with a second indicator because it required the first indicator to be ready in order to update the second one. To fix this bug, I improved the get implementation of the `PythonIndicator.IsReady` property so that it used the overridden python version of `IsReady` whenever it was defined. Thus, now if the user is updating manually a custom indicator, it can (and should) override the `is_ready` method to define when the indicator is ready.
#### Related Issue

Closes #6615 
#### Motivation and Context

With this change, users will be able to define when an indicator is ready

These changes might require to brief the community on how the `is_ready` method must be implemented in order to use that property

I created a regression test that creates a custom indicator that updates a second one on its time whenever it's updated (using `IndicatorExtensions.SMA`). The algorithm updates the custom indicator manually and asserts the indicator is ready at the end.
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.Python/IndicatorExtensionsSMAWithCustomIndicatorsRegressionAlgorithm.py
+++ b/Algorithm.Python/IndicatorExtensionsSMAWithCustomIndicatorsRegressionAlgorithm.py
@@ -18,25 +18,62 @@ class IndicatorExtensionsSMAWithCustomIndicatorsRegressionAlgorithm(QCAlgorithm)
     def initialize(self):
         self.set_start_date(2020, 2, 20)
         self.qqq = self.add_equity("QQQ", Resolution.DAILY).symbol
-        self.range_indicator = RangeIndicator("range")
+        self.range_indicator = RangeIndicator("range1")
         self.range_sma = IndicatorExtensions.sma(self.range_indicator, 5)
+
+        self.range_indicator_2 = RangeIndicator2("range2")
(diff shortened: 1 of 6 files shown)


Martin-Molinero (approved):
Thank you! 👍

PR #8289, opened 2024-08-28
Throw exception when base currency not found

### LINKED ISSUE #8091: Invalid Security Handling in CashBuyingPowerModel
#### Expected Behavior
If `CashBuyingPowerModel` finds an invalid security, throw an exception.

#### Actual Behavior
If `CashBuyingPowerModel` finds an invalid security, throw an exception, the order is invalid and we don't know why:
`Order Error: id: 1, Error executing margin models: Object reference not set to an instance of an object.`

In the syslog, we can read:

```
2024-06-14T20:53:43.6886306Z ERROR:: BrokerageTransactionHandler.HandleSubmitOrderRequest():  System.NullReferenceException: Object reference not set to an instance of an object.
   at QuantConnect.Securities.CashBuyingPowerModel.GetOpenOrdersReservedQuantity(SecurityPortfolioManager portfolio, Security security, Order order) in /Common/Securities/CashBuyingPowerModel.cs:line 416
```

It happens because the `BaseCurrency` of one of the securities is `null`. For example if the algorithm adds BTGUSDT crypto future:
```python
self.add_crypto_future("BTGUSDT")
```

it's not in our database, so the base currency is `null`.

#### Potential Solution
When we create a `CryptoFuture` object, we should throw an exception if the `baseCurrency` parameter is `null` ([CryptoFuture.cs#L76](https://github.com/QuantConnect/Lean/blob/master/Common/Securities/CryptoFuture/CryptoFuture.cs#L76))

#### Checklist

Marinovsky:
#### Description

I defined an exception to be thrown in the method `SecurityService.CreateSecurity()`, when the base currency is not found and the security type is CryptoFuture or Forex (which are the security types that use the base currency). Besides, some unit tests were failing after this change, because they were trying to create securities for Bybit non perpetual cryptofutures, which are not supported by LEAN currently. Thus, those non perpetual crypto futures are being removed in this PR.
#### Related Issue

Closes #8091 
#### Motivation and Context

With this change, LEAN will throw an exception whenever the base currency is not found when creating a security of type CryptoFuture or Forex. This will prevent LEAN from throwing another exception in a different part of the code where there isn't a problem

N/A

I created a unit test that attempts to create securities of types CryptoFuture and Forex using symbols not present in our SPDB. The test asserts an exception is thrown whenever we try to create a security using those symbols.
#### Types of changes

#### Checklist:


Martin-Molinero (commented):
Thank you, share minor request in slack👍

Martin-Molinero (approved):
Nice! Thank you 👍

PR #8291, opened 2024-08-29
Adds Overload to AddIndexOption

AlexCatarino:
#### Description
The new overload provides the same functionality of the `AddIndexOption(string, Resolution, string, bool)` overload which doesn't require the `AddIndex` method call and saving its `Symbol`.

#### Related Issue
N/A

#### Motivation and Context
Consistency: `AddIndexOption(string, Resolution, string, bool)`  provides this option.

Yes. Example with SPXW won't require adding the index. 

Updated regression test.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -2298,7 +2298,7 @@ namespace QuantConnect.Algorithm
         [DocumentationAttribute(AddingData)]
         public IndexOption AddIndexOptionContract(Symbol symbol, Resolution? resolution = null, bool fillForward = true)
         {
-            if (symbol.SecurityType != SecurityType.IndexOption && !symbol.IsCanonical())
+            if (symbol.SecurityType != SecurityType.IndexOption || symbol.IsCanonical())
             {
                 throw new ArgumentException("Symbol provided must be non-canonical and of type SecurityType.IndexOption");
             }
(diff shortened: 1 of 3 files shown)


Algorithm/QCAlgorithm.cs
```diff
@@ -2267,6 +2267,23 @@ public Option AddIndexOption(Symbol symbol, string targetOption, Resolution? res
...
+        /// <param name="market">Market of the index option. If no market is provided, we default to <see cref="Market.USA"/> </param>
+        /// <param name="fillForward">If true, this will fill in missing data points with the previous data point</param>
+        /// <returns>Canonical Option security</returns>
+        [DocumentationAttribute(AddingData)]
+        public Option AddIndexOption(string ticker, string targetOption, Resolution? resolution = null, string market = Market.USA, bool fillForward = true)
+        {
+            return AddIndexOption(
+                QuantConnect.Symbol.Create(ticker, SecurityType.Index, market),
+                targetOption, resolution, fillForward);
+        }
```
Martin-Molinero, line 2285: Nice! Couple of thoughts

> public Option AddOption(string underlying, Resolution? resolution = null, string market = null, bool fillForward = true, decimal leverage = Security.NullLeverage)

`ticker` -> `underlying` sounds like a good idea
`string market = Market.USA` -> `string market = null` too, see how it fetches from the brokerage model the market to use
`public Option AddIndexOption` also we should probably be returning an IndexOption security object?

PR #8303, opened 2024-09-07
fix: changed JsonFormat modifier to const

meta456trader:
Changed the JsonFormat variable modifier from const to static for consistency since it is the only non-const member in the DateFormat class. All other variables retain the const modifier as they represent compile-time constants.


Martin-Molinero:
Hey @meta456trader! Sorry we try to avoid these type of changes unless required, can have unexpected side effects for example this change doesn't compile. If interested in contributing please take a look at good first issue labels https://github.com/QuantConnect/Lean/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

Issue #8136, opened 2024-07-02
Option Position Match Finds Optimal Strategy (Lowest Margin Required)


AlexCatarino:
#### Expected Behavior
LEAN matches options legs to find the combination that requires less margin.

#### Actual Behavior
LEAN matches options legs that don't have the minimum margin requirement.

#### Potential Solution
N/A

#### Reproducing the Problem
Buy two bear call spread with different strikes: Spread 1 (A > B) > Spread 2 (C > D). On `OptionEquityBearCallSpreadRegressionAlgorithm`, add:

```csharp
callContracts = callContracts.Where(contract => contract.Strike > longCall.Strike && contract.Expiry == longCall.Expiry).ToList();
MarketOrder(callContracts[0].Symbol, -5);
AssertOptionStrategyIsPresent(OptionStrategyDefinitions.ShortButterflyCall.Name, 2);
MarketOrder(callContracts[1].Symbol, 5);
AssertOptionStrategyIsPresent(OptionStrategyDefinitions.BearCallSpread.Name, 10);
```

On `OptionEquityBaseStrategyRegressionAlgorithm`, we need to increase the strike range to +5:
```csharp
u.Strikes(-2, +5)
```

We should have two bear call spread with 10 contracts, but we have:
2 Butterfly Calls
1 Short Butterfly Call
4 Bear Call Spread

And we used the Short Butterfly Call to get the margin.

The backtest on the following thread also reproduces the issue:
https://www.quantconnect.com/forum/discussion/17414/unexpected-remaining-margin-decrease/p1

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #8312, opened 2024-09-10
Add Backspread Option Strategies

LouisSzeto:
#### Description
Add Call/Put Long/Short Backspread Option Strategies
It is a combination of other option strategies, so no option definitions or margin model required

Margin
- Call Backspread (Bear call spread + long call (0))
call-backspread-margin-ib

- Put Backspread (Bull put spread + long put (0))
put-backspread-margin-ib

- Short Call Backspread (Bull call spread (0) + naked call)
short-call-backspread-margin-ib

- Short Put Backspread (Bear put spread (0) + naked put)
short-put-backspread-margin-ib

Option Strategies

Unit tests (On definitions)
Regression tests (for margin and option strategy combinations)

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/OptionEquityCallBackspreadRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionEquityCallBackspreadRegressionAlgorithm.cs
@@ -19,6 +19,7 @@ using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
+using QuantConnect.Securities.Option;
 using QuantConnect.Securities.Option.StrategyMatcher;
 
 namespace QuantConnect.Algorithm.CSharp
@@ -50,12 +51,12 @@ namespace QuantConnect.Algorithm.CSharp
                     if (strike.Count() < 2) return;
 
                     var lowStrikeCall = callContracts.First();
-                    var highStrikeCall = callContracts.First(contract => contract.Strike > lowStrikeCall.Strike && contract.Expiry == lowStrikeCall.Expiry);
+                    var highStrikeCall = callContracts.First(contract => contract.Strike > lowStrikeCall.Strike && contract.Expiry == expiry);
 
                     var initialMargin = Portfolio.MarginRemaining;
 
-                    MarketOrder(lowStrikeCall.Symbol, -5);
-                    MarketOrder(highStrikeCall.Symbol, 10);
+                    var optionStrategy = OptionStrategies.CallBackspread(_optionSymbol, lowStrikeCall.Strike, highStrikeCall.Strike, expiry);
+                    Buy(optionStrategy, 5);
                     var freeMarginPostTrade = Portfolio.MarginRemaining;
 
                     // It is a combination of bear call spread and long call
@@ -121,7 +122,7 @@ namespace QuantConnect.Algorithm.CSharp
             {"Estimated Strategy Capacity", "$47000.00"},
             {"Lowest Capacity Asset", "GOOCV W78ZERHAOVVQ|GOOCV VP83T1ZUHROL"},
             {"Portfolio Turnover", "11.81%"},
-            {"OrderListHash", "6dbe3ae68d040c89feb283adeea5ee4c"}
+            {"OrderListHash", "6ece6c59826ea66fa7b0a1094a0021c7"}
         };
     }
 }
--- a/Algorithm.CSharp/OptionEquityCallBackspreadRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionEquityCallBackspreadRegressionAlgorithm.cs
@@ -15,12 +15,11 @@
 */
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using QuantConnect.Data;
 using QuantConnect.Data.Market;
-using System.Collections.Generic;
 using QuantConnect.Securities.Option.StrategyMatcher;
-using QuantConnect.Securities.Option;
 
 namespace QuantConnect.Algorithm.CSharp
 {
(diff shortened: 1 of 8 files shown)


Martin-Molinero:
Whops regression tests seem to be failing here @LouisSzeto

Martin-Molinero:
Regression tests still failing @LouisSzeto 👀

Launcher/config.json
```diff
@@ -9,7 +9,7 @@
   "environment": "backtesting", // "live-paper", "backtesting", "live-interactive", "live-interactive-iqfeed"
 
   // algorithm class selector
-  "algorithm-type-name": "BasicTemplateFrameworkAlgorithm",
```
Martin-Molinero, line 12: @LouisSzeto 👀

Algorithm.CSharp/OptionEquityCallBackspreadRegressionAlgorithm.cs
```diff
@@ -0,0 +1,126 @@
...
+                    var highStrikeCall = callContracts.First(contract => contract.Strike > lowStrikeCall.Strike && contract.Expiry == lowStrikeCall.Expiry);
+
+                    var initialMargin = Portfolio.MarginRemaining;
+
+                    MarketOrder(lowStrikeCall.Symbol, -5);
+                    MarketOrder(highStrikeCall.Symbol, 10);
+                    var freeMarginPostTrade = Portfolio.MarginRemaining;
+
+                    // It is a combination of bear call spread and long call (not a group)
+                    AssertOptionStrategyIsPresent(OptionStrategyDefinitions.BearCallSpread.Name, 5);
```
jhonabreul, line 62: Could also assert the call? A naked call? Same in the other algorithms

Common/Securities/Option/OptionStrategies.cs
jhonabreul, line 1: Should this file be reverted? Seems like just reordering the if-elses. Same for the option strategies definitions files

Tests/Common/Securities/Options/OptionStrategiesTests.cs
```diff
@@ -1412,5 +1412,189 @@ public void BuildsBullPutLadderStrategy()
...
+
+            var strike1 = 300m;
+            var strike2 = 325m;
+
+            // Unordered and repeated strikes
+            Assert.Throws<ArgumentException>(
+                () => OptionStrategies.CallBackspread(canonicalOptionSymbol, strike2, strike1, expiration));
+            Assert.Throws<ArgumentException>(
+                () => OptionStrategies.CallBackspread(canonicalOptionSymbol, strike1, strike1, expiration));
+        }
```
jhonabreul, line 1431: These could be test cases, like `[TestCase(300, 325)]` and `[TestCase(300, 300)]`.
Same for the rest of the tests

jhonabreul (commented):
Thanks! Leaving a couple minor comments

Common/Securities/Option/OptionStrategies.cs
LouisSzeto: I do it in purpose since the position group matcher seems to find them in order when I do the regression tests, so I make it order by the ones with the underlying, then by the number of legs

Algorithm.CSharp/OptionEquityCallBackspreadRegressionAlgorithm.cs
LouisSzeto: Not sure should it? Naked long call is not belongs to any option strategy?

Common/Securities/Option/OptionStrategies.cs
jhonabreul: Got it. Sounds like a consumer issue. It should have a sorting criteria for `OptionStrategyDefinitions.AllStrategies` so we don't rely on re-ordering this file every time we encounter this issue.

Sounds related to #8136. We should wait for that to be resolved first

Common/Securities/Option/OptionStrategyPositionGroupBuyingPowerModel.cs
```diff
@@ -120,94 +150,6 @@ public override MaintenanceMargin GetMaintenanceMargin(PositionGroupMaintenanceM
 
                 return new MaintenanceMargin(inAccountCurrency);
             }
-            else if (_optionStrategy.Name == OptionStrategyDefinitions.ProtectiveCollar.Name)
```
Martin-Molinero, line 123: The changes in this file aren't required? Let's revert them 👍

Common/Securities/Option/StrategyMatcher/OptionStrategyDefinitions.cs
```diff
@@ -305,223 +320,208 @@ public static ImmutableList<OptionStrategyDefinition> AllDefinitions
...
+        /// Sell 1 put contract without holding the underlying
         /// </summary>
-        public static OptionStrategyDefinition BullPutLadder { get; }
-            = OptionStrategyDefinition.Create("Bull Put Ladder",
-                OptionStrategyDefinition.PutLeg(-1),
-                OptionStrategyDefinition.PutLeg(+1, (legs, p) => p.Strike < legs[0].Strike,
-                                                    (legs, p) => p.Expiration == legs[0].Expiration),
-                OptionStrategyDefinition.PutLeg(+1, (legs, p) => p.Strike < legs[1].Strike,
-                                                    (legs, p) => p.Expiration == legs[0].Expiration)
+        public static OptionStrategyDefinition NakedPut { get; }
```
Martin-Molinero, line 522: These changes aren't required either? let's revert 💯

Common/Securities/Option/OptionStrategies.cs
Martin-Molinero: Let's reduce the changes to the minimum, revert anything that isn't required here. These new strategies, as sugar helper access for the users are okay, but given their margin requirements is just the sum of other 2 existing strategies let's just let the engine reuse those internally I think

Martin-Molinero (changes requested):
Cool!
Let's just review the changes and revert them to their minimum required. Adding the regression tests to master seem to pass without any of the other changes which seem risky and not adding much value

Common/Securities/Option/OptionStrategies.cs
```diff
@@ -1151,6 +1151,129 @@ DateTime expiration
...
+        /// <summary>
+        /// Method creates new Short Put Backspread strategy, that consists of two puts with the same expiration but different strikes.
+        /// It involves buying the higher strike put, while selling twice the number of the lower strike put.
+        /// </summary>
+        /// <param name="canonicalOption">Option symbol</param>
+        /// <param name="higherStrike">The strike price of the long put</param>
+        /// <param name="lowerStrike">The strike price of the short put</param>
+        /// <param name="expiration">Option expiration date</param>
+        /// <returns>Option strategy specification</returns>
+        public static OptionStrategy ShortPutBackspread(
```
Martin-Molinero, line 1267: Nice ty. Can the regression algorithms actually use these option strategies, else if feels like they are not being tested against? The unit tests are okay, but let's assert they work when used like the users will use them 👍

LouisSzeto: sure I changed to use option strategies helper class to order

Martin-Molinero (approved):
Thank you 👍

PR #8355, opened 2024-10-03
Fix bug ImmediateExecutionModel when using Crypto

### LINKED ISSUE #6793: `ImmediateExecutionModel` doesn't use proper position sizing for Binance fee model
#### Expected Behavior

Regardless of the fee model, the `ImmediateExecutionModel` achieves the target quantity of each order.

#### Actual Behavior

If we use the binance fee model, the fees are deducted from the coin/token we receive from the trade. As a result,  the `ImmediateExecutionModel` doesn't purchase the correct quantity of each coin/token after the fees are deducted.

#### Potential Solution

N/A

#### Reproducing the Problem

Run [this backtest](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_2f67389030f0a1f1fdd614ada21fa806.html). The algorithm should only trade on Jan 2nd, but it trades on Jan 2nd and Jan 3rd because it doesn't place large enough orders on Jan 2nd to achieve the target quantity.

#### System Information

QC Cloud

#### Checklist

Marinovsky:
#### Description

If we use the binance fee model(and other crypto fee models as well), the fees are deducted from the coin/token we receive from the trade. As a result, the ImmediateExecutionModel doesn't purchase the correct quantity of each coin/token after the fees are deducted. This PR aims to solve that issue by calculating the expected order fee's and adjusting the order quantity when the security type is crypto.
#### Related Issue

Closes #6793 
#### Motivation and Context

With this change, the market orders done by the `ImmediateExecutionModel` will have the correct quantity so that once they are filled, the filled quantity matches with the target quantity

N/A

I created one regression test in both C# and Python that reproduces the bug mentioned in the GH issue and asserts the filled quantity is the expected one and just 1 order is placed, instead of two.
#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/ImmediateExecutionModelWorksWithBinanceFeeModel.cs
+++ b/Algorithm.CSharp/ImmediateExecutionModelWorksWithBinanceFeeModel.cs
@@ -24,6 +24,11 @@ using System.Collections.Generic;
 
 namespace QuantConnect.Algorithm.CSharp
 {
+    /// <summary>
+    /// Regression algorithm to test ImmediateExecutionModel places orders with the
+    /// correct quantity (taking into account the fee's) so that the fill quantity
+    /// is the expected one.
+    /// </summary>
     public class ImmediateExecutionModelWorksWithBinanceFeeModel: QCAlgorithm, IRegressionAlgorithmDefinition
     {
         public override void Initialize()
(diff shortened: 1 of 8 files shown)


Martin-Molinero (approved):
Thank you 👍

PR #8410, opened 2024-11-14
Implement Reset() method

### LINKED ISSUE #8320: Allow Consolidator Reset
#### Expected Behavior
It is possible to reset consolidators.
Say we have a weekly or monthly consolidator, and a split occurs. The weekly/monthly bar will have pre- and post-split data:

Captura de ecrã 2024-09-11 231322

We can workaround this issue by removing the consolidator from the subscription_manager, and create a new one.

```python
    def reset_consolidator(self, symbol):
        self.subscription_manager.remove_consolidator(symbol, self.consolidator)
        self.consolidator: TradeBarConsolidator = self.consolidate(symbol, Calendar.WEEKLY, self.on_weekly_bar)
```

Ideally, this should be a simple `reset` method call as supported by indicators. 

#### Actual Behavior
We cannot reset indicators. 

#### Potential Solution
Add `Reset` method that sets the internal working data to null.

#### Reproducing the Problem
[backtest #c52d5e](https://www.quantconnect.com/terminal/processCache?request=embedded_backtest_c52d5eabb58016b042a2f5a65072314a.html)

#### Checklist

Marinovsky:
#### Description

Implement `Reset()` method in consolidators
#### Related Issue

Closes #8320 
#### Motivation and Context

With this change users will be able to reset a consolidator

It might be useful to inform users about this new feature

I created a base test class with a unit test that creates, updates and resets a consolidator asserting the it has been correctly been reset. This method depends on two other methods that can be overridden, so that the input type and assertion procedure can be adjusted
#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Consolidators/DataConsolidator.cs
+++ b/Common/Data/Consolidators/DataConsolidator.cs
@@ -111,7 +111,9 @@ namespace QuantConnect.Data.Consolidators
         /// <summary>
         /// Resets the consolidator
         /// </summary>
-        public abstract void Reset();
+        public virtual void Reset()
+        {
+        }
 
         /// <summary>Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.</summary>
         /// <filterpriority>2</filterpriority>
--- a/Common/Data/Consolidators/DataConsolidator.cs
+++ b/Common/Data/Consolidators/DataConsolidator.cs
@@ -108,6 +108,9 @@ namespace QuantConnect.Data.Consolidators
             Consolidated = consolidated;
         }
 
+        /// <summary>
+        /// Resets the consolidator
+        /// </summary>
         public abstract void Reset();
 
         /// <summary>Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.</summary>
--- a/Common/Data/Consolidators/DataConsolidator.cs
+++ b/Common/Data/Consolidators/DataConsolidator.cs
@@ -58,7 +58,7 @@ namespace QuantConnect.Data.Consolidators
         /// </summary>
         public IBaseData Consolidated
         {
-            get; private set;
+            get; protected set;
         }
 
         /// <summary>
@@ -107,6 +107,8 @@ namespace QuantConnect.Data.Consolidators
             // and the previous consolidated data at the same time without extra bookkeeping
             Consolidated = consolidated;
         }
+
+        public abstract void Reset();
 
         /// <summary>Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.</summary>
         /// <filterpriority>2</filterpriority>
--- a/Common/Data/Consolidators/DataConsolidator.cs
+++ b/Common/Data/Consolidators/DataConsolidator.cs
@@ -113,6 +113,7 @@ namespace QuantConnect.Data.Consolidators
         /// </summary>
         public virtual void Reset()
         {
+            Consolidated = null;
         }
 
         /// <summary>Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.</summary>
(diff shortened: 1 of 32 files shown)


Common/Data/Consolidators/IdentityDataConsolidator.cs
```diff
@@ -66,5 +66,14 @@ public override void Update(T data)
         public override void Scan(DateTime currentLocalTime)
         {
         }
+
+        /// <summary>
+        /// Resets the consolidator
+        /// </summary>
+        public override void Reset()
+        {
+            _last = default(T);
```
jhonabreul, line 75: Could be directly null, it's the default for all posible T types.

Tests/Common/Data/BaseConsolidatorTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                Assert.IsNull(consolidator.Consolidated);
+            }
+            else
+            {
+                Assert.AreEqual(previousConsolidator.Consolidated?.Value, consolidator.Consolidated?.Value);
+            }
+        }
+
+        protected virtual dynamic GetTestValues()
```
jhonabreul, line 45: Can this be `IEnumerable<IBaseData` instead of `dynamic`?

Tests/Common/Data/VolumeRenkoConsolidatorTests.cs
```diff
@@ -226,5 +226,38 @@ public void MultipleConsoldation()
...
+        protected override dynamic GetTestValues()
+        {
+            var time = new DateTime(2016, 3, 1);
+            var testValues = new List<decimal[]>
+            {
+                new decimal[]{5m, 5m}, new decimal[]{5m, 3m}, new decimal[]{5m, 7m}, new decimal[]{5m, 6m},
+                new decimal[]{5m, 5m}, new decimal[]{5m, 3m}, new decimal[]{5m, 7m}, new decimal[]{5m, 6m},
+                new decimal[]{5m, 5m}, new decimal[]{5m, 3m}, new decimal[]{5m, 7m}, new decimal[]{5m, 6m},
+                new decimal[]{5m, 5m}, new decimal[]{5m, 3m}, new decimal[]{5m, 7m}, new decimal[]{5m, 6m}
+            };
```
jhonabreul, line 239: Not used

jhonabreul (commented):
Nice! Leaving a few comments

Common/Data/Consolidators/DataConsolidator.cs
```diff
@@ -108,6 +108,13 @@ protected virtual void OnDataConsolidated(IBaseData consolidated)
             Consolidated = consolidated;
         }
 
+        /// <summary>
+        /// Resets the consolidator
+        /// </summary>
+        public virtual void Reset()
+        {
+        }
```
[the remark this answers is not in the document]
Marinovsky: I also thought on that, but Martin made me realize we need to keep backwards compatibility with any custom existing consolidator, so they compile/don't break

Common/Data/Consolidators/DataConsolidator.cs
```diff
@@ -58,7 +58,7 @@ public void Update(IBaseData data)
         /// </summary>
         public IBaseData Consolidated
         {
-            get; private set;
+            get; protected set;
```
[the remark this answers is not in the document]
Marinovsky: It could be, but there are other abstract consolidators doing the same (making `Consolidated` protected) like BaseTimelessConsolidator https://github.com/QuantConnect/Lean/blob/master/Common/Data/Consolidators/BaseTimelessConsolidator.cs#L54

Common/Data/Consolidators/IdentityDataConsolidator.cs
Marinovsky: Hmm perhaps it was the rule for previous versions of .NET but apparently now is not the case: https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0403?f1url=%3FappId%3Droslyn%26k%3Dk(CS0403)

Martin-Molinero (approved):
Thank you! 🚀

PR #8446, opened 2024-12-05
Add IB Tiered Fee Model

LouisSzeto:
#### Description
- Add IB Tiered Fee Structure Model
- Refactor the current IB fee model implementation
- Add Crypto support

#### Motivation and Context
Missing feature

Add to Writing Algorithms / Reality Modeling / Transaction Fee / Supported Models

- Unit tests
- Regression tests

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/InteractiveBrokersTieredFeeModelAlgorithm.cs
+++ b/Algorithm.CSharp/InteractiveBrokersTieredFeeModelAlgorithm.cs
@@ -25,7 +25,7 @@ namespace QuantConnect.Algorithm.CSharp
     /// </summary>
     public class InteractiveBrokersTieredFeeModelAlgorithm : QCAlgorithm, IRegressionAlgorithmDefinition
     {
-        private Symbol _spy;
+        private Symbol _spy, _aig, _bac;
         private IFeeModel _feeModel = new InteractiveBrokersTieredFeeModel();
 
         /// <summary>
@@ -33,14 +33,16 @@ namespace QuantConnect.Algorithm.CSharp
         /// </summary>
         public override void Initialize()
         {
-            SetStartDate(2020, 1, 1);   //Set Start Date
-            SetEndDate(2020, 3, 1);     //Set End Date
+            SetStartDate(2013, 10, 7);   //Set Start Date
+            SetEndDate(2013, 10, 10);     //Set End Date
             SetCash(1000000000);            //Set Strategy Cash
 
             // Set the fee model to be shared by all securities to accurately track the volume/value traded to select the correct tiered fee structure.
             SetSecurityInitializer((security) => security.SetFeeModel(_feeModel));
 
             _spy = AddEquity("SPY", Resolution.Minute, extendedMarketHours: true).Symbol;
+            _aig = AddEquity("AIG", Resolution.Minute, extendedMarketHours: true).Symbol;
+            _bac = AddEquity("BAC", Resolution.Minute, extendedMarketHours: true).Symbol;
         }
 
         public override void OnData(Slice slice)
@@ -48,15 +50,21 @@ namespace QuantConnect.Algorithm.CSharp
             // Order at different time for various order type to elicit different fee structure.
             if (slice.Time.Hour == 9 && slice.Time.Minute == 0)
             {
-                MarketOnOpenOrder(_spy, 100000);
+                MarketOnOpenOrder(_spy, 30000);
+                MarketOnOpenOrder(_aig, 30000);
+                MarketOnOpenOrder(_bac, 30000);
             }
             else if (slice.Time.Hour == 10 && slice.Time.Minute == 0)
             {
-                MarketOrder(_spy, 100000);
+                MarketOrder(_spy, 30000);
+                MarketOrder(_aig, 30000);
+                MarketOrder(_bac, 30000);
             }
             else if (slice.Time.Hour == 15 && slice.Time.Minute == 30)
             {
-                MarketOnCloseOrder(_spy, -200000);
+                MarketOnCloseOrder(_spy, -60000);
+                MarketOnCloseOrder(_aig, -60000);
+                MarketOnCloseOrder(_bac, -60000);
             }
         }
 
@@ -73,7 +81,7 @@ namespace QuantConnect.Algorithm.CSharp
         /// <summary>
         /// Data Points count of all timeslices of algorithm
         /// </summary>
-        public long DataPoints => 160;
+        public long DataPoints => 23076;
 
         /// <summary>
         /// Data Points count of the algorithm history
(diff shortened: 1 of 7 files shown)


Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
```diff
@@ -0,0 +1,782 @@
...
+                    var regulatory = _optionExchangesOrfFee.Contains(market) ? 
+                        (0.01915m + 0.0048m) * quantity :
+                        0.0048m * quantity;
+                    // Transaction Fees: SEC Transaction Fee + FINRA Trading Activity Fee (only charge on sell)
+                    var transaction = order.Quantity < 0 ? 0.0000278m * Math.Abs(order.GetValue(security)) + 0.00279m * quantity : 0m;
+                    // Clearing Fee
+                    var clearing = Math.Min(0.02m * quantity, 55m);
+                    feeResult = optionFee.Amount + regulatory + transaction + clearing;
+                    feeCurrency = optionFee.Currency;
+                    break;
```
jhonabreul, line 172: Should update ` _monthlyTradeVolume`  for these security types?

Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
```diff
@@ -0,0 +1,782 @@
...
+                MinimumFee = minimumFee;
+                MaximumFeeRate = maximumFeeRate;
+            }
+        }
+
+        /// <summary>
+        /// Determines which tier an account falls into based on the monthly trading volume of cryptos
+        /// </summary>
+        /// <remarks>https://www.interactivebrokers.com/en/pricing/commissions-cryptocurrencies.php?re=amer</remarks>
+        private static void ProcessCryptoRateSchedule(decimal monthlyCryptoTradeAmountInUSDollars, out decimal commissionRate)
```
jhonabreul, line 766: Same comment about duplication. Is there room for sharing base logic?

Algorithm.CSharp/InteractiveBrokersTieredFeeModelAlgorithm.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        public override void OnData(Slice slice)
+        {
+            // Order at different time for various order type to elicit different fee structure.
+            if (slice.Time.Hour == 9 && slice.Time.Minute == 0)
+            {
+                MarketOnOpenOrder(_spy, 30000);
+                MarketOnOpenOrder(_aig, 30000);
+                MarketOnOpenOrder(_bac, 30000);
```
jhonabreul, line 55: Are these orders filled on the next day or right away? If the latest, might as well use MarketOrder? Same for the MarketOnClose ones

jhonabreul (commented):
Nice! Leaving a few comments.

Algorithm.CSharp/InteractiveBrokersTieredFeeModelAlgorithm.cs
LouisSzeto: They should be filled right away at 9:30am, same for market on close at 4pm

Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
```diff
@@ -0,0 +1,782 @@
...
+            var order = parameters.Order;
+            var security = parameters.Security;
+
+            // Reset monthly trade value tracker when month rollover.
+            if (_lastOrderTime.Month != order.Time.Month && _lastOrderTime != DateTime.MinValue)
+            {
+                _monthlyTradeVolume = _monthlyTradeVolume.ToDictionary(kvp => kvp.Key, _ => 0m);
+            }
+            // Reprocess the rate schedule based on the current traded volume in various assets.
+            ReprocessRateSchedule(_monthlyTradeVolume[SecurityType.Equity], _monthlyTradeVolume[SecurityType.Future], _monthlyTradeVolume[SecurityType.Forex],
```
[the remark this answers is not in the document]
LouisSzeto: Nah... the fee will change by sufficient volume/dollar volume traded within that month

Common/Orders/Fees/InteractiveBrokersFeeModel.cs
```diff
@@ -374,70 +164,8 @@ private static CashAmount UnitedStatesFutureFees(Security security)
...
-            }
-
-            if (!exchangeFees.TryGetValue(symbol, out exchangeFeePerContract))
-            {
-                exchangeFeePerContract = 0.00m;
-            }
-
             // Add exchange fees + IBKR regulatory fee (0.02)
-            return new CashAmount(ibFeePerContract + exchangeFeePerContract + 0.02m, Currencies.EUR);
+            return new CashAmount(ibFeePerContract + exchangeFeePerContract + 0.02m, Currencies.USD);
```
jhonabreul, line 168: Nice catch

Tests/Common/Orders/Fees/InteractiveBrokersTieredFeeModelTests.cs
jhonabreul, line 1: Would it be possible to run at least a few of these tests for the Python model as well? Just to make sure they are always kept the same when changed

jhonabreul (commented):
Looking good! Just a couple more comments

Tests/Common/Orders/Fees/InteractiveBrokersTieredFeeModelTests.cs
Martin-Molinero: CC: @LouisSzeto 👁️

Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
```diff
@@ -0,0 +1,287 @@
...
+                    var transaction = order.Direction == OrderDirection.Sell ?
+                        0.0000278m * Math.Abs(order.GetValue(security)) + 0.00279m * quantity :
+                        0m;
+                    // Clearing Fee
+                    var clearing = Math.Min(0.02m * quantity, 55m);
+
+                    feeResult += regulatory + transaction + clearing;
+
+                    // Update the monthly value traded
+                    _monthlyTradeVolume[SecurityType.Option] += quantity * orderPrice;
```
Martin-Molinero, line 162: Whops don't think this state is correct, the lean engine can and will call `GetOrderFee` multiple times for the same order, when submitting later when processing in the backtesting brokerage, can even be more times in cases of setHoldings where it can call it N times to get the best position size. Maybe something that could be done is look at the order status, if submitted use it to account for the monthly volume, should also keep a guard not to process the same order id multiple times I think, but needs more testing and asserting to be sure it's okay
Seems should make the monthly traded volume public so it can be asserted in the regression algorithm and accessible for the user
Should adjust the algo to _also_ use SetHoldings and will make the bug of the multiple accounting more clearer,

__
btw `CalculateOptionFee` already returns `quantity * orderPrice`, what about the contract multiplier?
__
the adjustments being made here for equity and options fees, should they also be for the main IB fee model?

Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
```diff
@@ -0,0 +1,287 @@
...
+                ibFeePerContract = new[] { 0.85m, 0.65m, 0.45m, 0.25m };
+            }
+
+            if (!exchangeFees.TryGetValue(symbol, out exchangeFeePerContract))
+            {
+                exchangeFeePerContract = 1.60m;
+            }
+
+            // Add exchange fees + IBKR regulatory fee (0.02)
+            return new CashAmount(ibFeePerContract[_futureCommissionTier] + exchangeFeePerContract + 0.02m, Currencies.USD);
```
Martin-Molinero, line 252: This implementation seems like a slightly more generic version of the one at `InteractiveBrokersFeeModel` seems they can be mostly shared, where `InteractiveBrokersFeeModel` just uses tier 0?

Common/Orders/Fees/InteractiveBrokersFeeHelper.cs
```diff
@@ -0,0 +1,584 @@
...
+        }
+
+        /// <summary>
+        /// Calculate the transaction fee of a Cfd order
+        /// </summary>
+        internal static void CalculateCfdFee(Security security, Order order, out decimal fee, out string currency)
+        {
+            var value = Math.Abs(order.GetValue(security));
+            fee = 0.00002m * value; // 0.002%
+            currency = security.QuoteCurrency.Symbol;
```
Martin-Molinero, line 275: Whops careful! `GetValue` returns in account currency, which might not be related to quote currency. Same comment for the other calls to `GetValue` for forex & crypto

Martin-Molinero (changes requested):
Cool! Leaving some requests and concerns

Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
LouisSzeto: > btw CalculateOptionFee already returns quantity * orderPrice, what about the contract multiplier?
The fee is per contract, so we don't need to account for the contract multiplier

Common/Orders/Fees/InteractiveBrokersTieredFeeModel.cs
LouisSzeto: Hmm.. I would say it is the fee = tier 0 is "very coincident". Some cases the cost is completely different for fixed and tiered fee model, so we leave with this implementation in case that happens in the future.

Common/Orders/Fees/InteractiveBrokersFeeHelper.cs
LouisSzeto: CFD is direct copying from the original implementation, since it has no tier... I'll change it to `order.AbsoluteQuantity * order.Price` then.

Both Forex and Crypto fees are in USD👌

slorion:
Thanks for this addition, I just happen to need it to 👍 

IBKR mentions that "Orders where the commission cap is applied do not count towards the monthly volume tiers." I do not see this part handled in the code?

slorion:
I'm also looking into Alpaca broker and I see that regulatory fees for equities are not modeled, but they would be the same as IBKR. It's outside the scope of this PR, but worth keeping in mind that it is also needed when trading USA equities with other brokers.

ianfiske:
This hasn't moved in a few months -- is there a plan to merge this MR?

PR #8454, opened 2024-12-09
Implement ZigZag (ZZ) indicator

### LINKED ISSUE #3179: Implement ZigZag Indicator
#### Expected Behavior
ZigZag is supported.

#### Actual Behavior
ZigZag is not supported.

#### Potential Solution
Implement ZigZag indicator.

This indicator is available in Python with this package:
https://github.com/jbn/ZigZag
We could just add it to the Foundation, but it would only be available for Python algorithms and wouldn't benefit from automatic updates and warm up.

#### Checklist

JosueNina:
#### Description
This pull request introduces the ZigZag indicator, which is used to identify significant price movements and trends by connecting key high and low points in a financial time series. It helps traders and analysts visualize trend changes and potential reversals in the market.

#### Related Issue
Closes #3179 

#### Motivation and Context
The ZigZag indicator is useful for detecting trend changes and filtering out market noise, which can improve technical analysis by focusing on significant price movements.

Yes, the documentation should be updated to include usage instructions for the ZigZag indicator and explain the parameters such as sensitivity and minimum trend length.

The ZigZag indicator has been tested using unit tests that verify the calculation of the indicator over historical data. The Talipp library was used to compare values and ensure the accuracy of the indicator's output against expected results.

#### Types of changes

#### Checklist:


Indicators/ZigZag.cs
jhonabreul, line 1: You might want to implement the `Reset` method and make sure or extended the unit tests fixture to make sure reset works as expected. See other indicators for reference

Indicators/ZigZag.cs
```diff
@@ -0,0 +1,175 @@
...
+    /// filtering out noise using a sensitivity threshold and a minimum trend length.
+    /// It alternates between high and low pivots to determine market trends.
+    /// </summary>
+    public class ZigZag : BarIndicator, IIndicatorWarmUpPeriodProvider
+    {
+        /// <summary>
+        /// Stores the most recent high pivot value in the ZigZag calculation.
+        /// Updated whenever a valid high pivot is identified.
+        /// </summary>
+        public decimal HighPivot { get; set; }
```
jhonabreul, line 32: Just a thought but would the user like to know which was the last/current pivot? Either a high or a low?
Perhaps we can have a data class to hold the pivot value (a decimal) and maybe an enum for `High`/`Low`?

Indicators/ZigZag.cs
```diff
@@ -0,0 +1,175 @@
...
+        /// Updated whenever a valid low pivot is identified.
+        /// </summary>
+        public decimal LowPivot { get; set; }
+
+        /// <summary>
+        /// The sensitivity threshold for detecting significant price movements.
+        /// A decimal value between 0 and 1 that determines the percentage change required
+        /// to recognize a new pivot.
+        /// </summary>
+        private readonly decimal _sensitivity;
```
jhonabreul, line 45: nit - In Lean the common practice is to have privates at the top of the class

Tests/TestData/spy_zigzag.csv
jhonabreul, line 1: Just a thought: do you think we could have the zig zag value always in its column? Granted, tt would be repeated until the trend changes but at least the test suite would be asserting that the indicator holds that value until the trend changes

Indicators/ZigZag.cs
JosueNina: I need to expose these values to update them in case a lower or higher point is found compared to the current one. This happens because we process one value at a time, so these variables store the latest updated low/high pivots. Now, if the user only needs the last pivot, they can simply use .Current.Value from the indicator.

Tests/TestData/spy_zigzag.csv
JosueNina: I am using the talipp library, which only returns a list of values representing the pivots. What I do to generate my .csv is determine which rows correspond to those values, while all the others are None. The reason I cannot simply take the previous value is that in LEAN we process one value at a time, updating the low/high pivots on each iteration, whereas TA-Lib processes everything at once and returns a full list of pivots.

PR #8525, opened 2025-01-13
Handle zero price and target quantity case - Collective2

### LINKED ISSUE #8523: Failure to Calculate Quantity Should not Throw Collective2 Signal
#### Expected Behavior
Collective2 doesn't throw an exception and exit the algorithm when `ConvertPercentageToQuantity` can calculate the number of shares.

#### Actual Behavior
`ConvertPercentageToQuantity` throws an exception and exits the algorithm.

#### Potential Solution
Convert the exception to a warning. `ConvertPercentageToQuantity` return `null` when the price is zero, for example, which is not a problem if the quantity held is zero.

#### Reproducing the Problem
Call `set_target_portfolio_from_portfolio` when the price of some securities is zero.

#### Checklist

JosueNina:
#### Description
Handled the case where price and target quantity are zero to prevent exceptions.

#### Related Issue
Closes #8523 

#### Motivation and Context
Prevents the algorithm from throwing exceptions when the price and target quantity are zero, ensuring smoother execution.

N/A

Tested with unit tests to ensure no exceptions are thrown and the result is zero when both price and target quantity are zero.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Algorithm/Framework/Portfolio/SignalExports/Collective2SignalExport.cs
+++ b/Common/Algorithm/Framework/Portfolio/SignalExports/Collective2SignalExport.cs
@@ -209,6 +209,11 @@ namespace QuantConnect.Algorithm.Framework.Portfolio.SignalExports
         /// <returns>Number of shares hold of the given position</returns>
         protected int ConvertPercentageToQuantity(IAlgorithm algorithm, PortfolioTarget target)
         {
+            if (algorithm.Securities[target.Symbol].Price == 0 && target.Quantity == 0)
+            {
+                algorithm.Debug($"Warning: The price for {target.Symbol} is 0, and the target quantity is 0. Returning 0 as the calculated quantity.");
+                return 0;
+            }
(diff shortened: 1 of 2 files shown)


AlexCatarino:
In the `ConvertHoldingsToCollective2` method, we can add that if the security is not invested and the target quantity is zero, there is nothing to do:
```csharp
protected bool ConvertHoldingsToCollective2(SignalExportTargetParameters parameters, out List<Collective2Position> positions)
{
    _algorithm = parameters.Algorithm;
    var targets = parameters.Targets;
    positions = new List<Collective2Position>();
    foreach (var target in targets)
    {
        if (target == null)
        {
            _algorithm.Error("One portfolio target was null");
            return false;
        }

        // Don't create a signal that doesn't represent a position
        if (target.Quantity == 0 && !_algorithm.Portfolio[target.Symbol].Invested)
        {
            continue;
        }

        if (!ConvertTypeOfSymbol(target.Symbol, out string typeOfSymbol))
        {
            return false;
        }
```

Martin-Molinero (approved):

PR #8564, opened 2025-01-29
Fix options indicators data synchronization

jhonabreul:
#### Description

Introduce `MultiSymbolIndicator` class for indicators that work on multiple symbols.

This class allows to detect when the indicator is ready to calculate, that is, when data for all symbols at the same time is available. For instance, if an indicator works with daily data for an option and it's underlying, and they have different market closes, daily data would come at different times of the day for the same date. When this happens, the indicator is able to detect that both symbols have data for the same date and is ready to calculate.

Options indicators, including implied volatility and greeks now are based on this class to handle cases like this.

#### Motivation and Context

An example is RUT index, where market closes at 4pm for indices and at 4:15pm for its options.

Unit tests.
Manual runs.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -3993,7 +3993,7 @@ namespace QuantConnect.Algorithm
             }
         }
 
-        private void InitializeOptionIndicator(IndicatorBase<IndicatorDataPoint> indicator, Resolution? resolution, Symbol symbol, Symbol mirrorOption)
+        private void InitializeOptionIndicator(IndicatorBase<IBaseData> indicator, Resolution? resolution, Symbol symbol, Symbol mirrorOption)
         {
             RegisterIndicator(symbol, indicator, ResolveConsolidator(symbol, resolution, typeof(QuoteBar)));
             RegisterIndicator(symbol.Underlying, indicator, ResolveConsolidator(symbol.Underlying, resolution));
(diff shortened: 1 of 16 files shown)


Martin-Molinero (approved):
Thank you, nice work 👍, minor improvement request shared

Martin-Molinero (approved):

NoSealsInSealBeach:
Any good reason as to why the ComputeIndicator method in OptionGreekIndicatorBase got sealed with the above change ?  The ComputeIndicator method seems to have replaced the logic of the Calculate method which I was overriding in my custom classes to have my own time to expiration calculation. Now there is no straightforward way to override the time to expiration calculation in the custom greek classes derived from the Greek Indicators anymore...unless one implements their own OptionGreekIndicatorBase class but that's messy as then you will have to implement the full indicator classes for each greek as they will be deriving from that custom class...and at that point you might as well skip using QuantConnect Indicators and write your own indicators. Is it possible for you to unseal this method ? This has broken all my custom Greeks and IV classes. Thanks!!

NoSealsInSealBeach:
Also, I noticed that the ComputeIndicator in the ImpliedVolatility is unsealed so I can derive a custom class from ImpliedVolatility and use my own time to expiration calcs in the ImpliedVolatility indicator. But I am not allowed to do it for greeks ? That seems inconsistent to me...

jhonabreul:
Hi @NoSealsInSealBeach 
You are right, fixing this in https://github.com/QuantConnect/Lean/pull/8570
Thank you

Issue #8803, opened 2025-05-29
UserDefinedUniverse should have constructor that takes a Func<DateTime, IEnumerable<Symbol>>


powerdude:
#### Expected Behavior

Should be able to construct a universe using a selector that returns symbols to match the one that can return strings.  `CustomUniverse` and `CustomUniverseSelectionModel` should also be updated to use symbols instead of just strings.

#### Actual Behavior

Only has a constructor that takes strings

#### Potential Solution

A new constructor may not be necessary if it is possible to pass in option symbols (i.e. "AMZN  250207C00235000") without further downstream errors because of the ' ' in the symbol name.

```csharp
    /// <summary>
    /// Initializes a new instance of the <see cref="UserDefinedUniverse"/> class
    /// </summary>
    /// <param name="configuration">The configuration used to resolve the data for universe selection</param>
    /// <param name="universeSettings">The settings used for new subscriptions generated by this universe</param>
    /// <param name="interval">The interval at which selection should be performed</param>
    /// <param name="selector">Universe selection function invoked for each time returned via GetTriggerTimes.
    /// The function parameter is a DateTime in the time zone of configuration.ExchangeTimeZone</param>
    public UserDefinedUniverse(SubscriptionDataConfig configuration, UniverseSettings universeSettings, TimeSpan interval, Func<DateTime, IEnumerable<Symbol>> selector)
        : base(configuration)
    {
        Interval = interval;
        UniverseSettings = universeSettings;
        _selector = time =>
        {
            var selectSymbolsResult = selector(time.ConvertFromUtc(Configuration.ExchangeTimeZone));
            // if we received an unchaged then short circuit the symbol creation and return it directly
            if (ReferenceEquals(selectSymbolsResult, Unchanged))
            {
                return Unchanged;
            }
            return selectSymbolsResult;
        };
    }
```
#### Checklist

- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

Martin-Molinero:
Hey @powerdude!
Please share a simple example algorithm which requires this behavior, this same algorithm should later be used as a regression algorithm in the test suite, ACs point here is right https://github.com/QuantConnect/Lean/pull/8804#issuecomment-2920183786. Thank you

PR #8804, opened 2025-05-29
Add constructor to UserDefinedUniverse that takes a delegate to return symbols 

powerdude:
#### Description

Add new constructor that takes a `Func` that can return symbols instead of just strings for custom universes since using option symbols causes errors due to spaces in their names

#### Related Issue

#8803 

#### Motivation and Context

It is not possible to create a universe selector that returns specific option symbols due to parsing downstream that doesn't expect a space in the symbol name

no

#### Types of changes

#### Checklist:


AlexCatarino:
Hi @powerdude,
What is the use case? You need to add unit test/regression/example showing the point of this change.

Issue #8125, opened 2024-07-02
Implement Klinger Volume Oscillator


AlexCatarino:
#### Expected Behavior
Supports Klinger Volume Oscillator (see [talipp implementation](https://github.com/nardew/talipp/blob/main/talipp/indicators/KVO.py)).

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #8811, opened 2025-06-03
feature: Klinger Volume Oscillator

mattfair:
#### Description
Implements Klinger Volume Oscillator

#### Related Issue

Closes https://github.com/QuantConnect/Lean/issues/8125

#### Motivation and Context
Implements the volume oscillator

Update any documentation for supported indicators.

Verified against generated KVO(5, 10) from talipp https://github.com/QuantConnect/Lean/blob/54e58c6302f241e61e76e82e33df09e839800663/Tests/TestData/generate_reference_data_from_talipp.py#L158

#### Types of changes

#### Checklist:

- [ ] All new and existing tests passed.  -- Unrelated tests fail


Tests/Indicators/KlingerVolumeOscillatorTests.cs
```diff
@@ -0,0 +1,43 @@
+using System;
```
Martin-Molinero, line 15: Minor missing license header

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,166 @@
...
+    /// price movement and trading volume to assess the strength of market trends and identify potential
+    /// trend reversals. As a volume-based oscillator, it measures the force behind price movements by
+    /// incorporating volume data adjusted for price trends and specific conditions. Traders use the KVO
+    /// to analyze its behavior relative to price action, looking for patterns such as divergences or
+    /// crossovers that can provide insights into market trends and potential turning points.
+    /// </summary>
+    public class KlingerVolumeOscillator : TradeBarIndicator, IIndicatorWarmUpPeriodProvider
+    {
+        private readonly int _fastPeriod;
+        private readonly int _slowPeriod;
```
Martin-Molinero: Unused variables, should just be deleted

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,166 @@
...
+        /// <param name="slowPeriod">The slow EMA period.</param>
+        public KlingerVolumeOscillator(int fastPeriod, int slowPeriod)
+            : this($"KVO({fastPeriod},{slowPeriod})", fastPeriod, slowPeriod)
+        {
+        }
+
+        /// <summary>
+        /// Computes the next value of the Klinger Volume Oscillator based on the input data point.
+        /// </summary>
+        protected override decimal ComputeNextValue(TradeBar bar)
```
Martin-Molinero, line 92: I get the feeling this can be simplified a bit further, less lines and variables, see for example https://help.cqg.com/cqgic/25/default.htm#!Documents/klingervolumeoscillatorkvo.htm is the same as being done here right? `'Volume Force = VF  = (if Key Pricet > Key Pricet-1 , Volume otherwise -Volume)'`, maybe `(bar.High + bar.Low + bar.Close) / 3m;` can be pushed into a rolling window of size 2, so keep current and previous, and can get rid of a bunch of variables, including `_hasPrevBar` using isReady

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,166 @@
...
+            // Volume force:  strength of volume flow in the direction of the trend.
+            // This implementation follows Talipp's formula.
+            // Formula: volumeForce = volume * |2 * (dm / cm - 1)| * trend * 100
+            // Special case: If cm is zero, volumeForce is set to 0 to prevent division by zero.
+            // Reference: https://github.com/nardew/talipp/blob/70dc9a26889c9c9329e44321e1362c4db43dbcc3/talipp/indicators/KVO.py#L85
+            var volumeForce = cm == 0m ? 0m : bar.Volume * Math.Abs(2m * (dm / cm - 1m)) * currentTrend * 100m;
+
+            // update moving averages
+            _fastEma.Update(new IndicatorDataPoint(bar.Time, volumeForce));
+            _slowEma.Update(new IndicatorDataPoint(bar.Time, volumeForce));
```
Martin-Molinero: Should generically use EndTime instead of Time for when updating the EMAs, image a daily resolution bar, time is the start, but we need endtime which is when the value is being calculated

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,166 @@
...
+{
+    /// <summary>
+    /// The Klinger Volume Oscillator (KVO) is a technical indicator that analyzes the relationship between
+    /// price movement and trading volume to assess the strength of market trends and identify potential
+    /// trend reversals. As a volume-based oscillator, it measures the force behind price movements by
+    /// incorporating volume data adjusted for price trends and specific conditions. Traders use the KVO
+    /// to analyze its behavior relative to price action, looking for patterns such as divergences or
+    /// crossovers that can provide insights into market trends and potential turning points.
+    /// </summary>
+    public class KlingerVolumeOscillator : TradeBarIndicator, IIndicatorWarmUpPeriodProvider
```
Martin-Molinero, line 29: It seems there's an expected KVO Signal line? I think we should add it as a public property so users can access it's value

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,166 @@
...
+            {
+                // Start new flow accumulation with the previous daily movement (dm)
+                cum = _previousDm;
+            }
+            else
+            {
+                // Continue flow accumulation in the same trend direction
+                cum = _previousCm;
+            }
+            cm = cum + dm;
```
Martin-Molinero: Generically we try to expand all variable names/private fields so they can easily be understood, for example `cm` -> `cumulative`, `dm` -> etc

Indicators/KlingerVolumeOscillator.cs
mattfair: similar, however there seems to be different variations on how volume force is calculated.  Since I was using talipp generated data to validate, I chose to calculate volume force like they did. 
Agree, on cleaning things up with the rolling windows.

Indicators/KlingerVolumeOscillator.cs
mattfair: yep, I'll add parameter for signal length, but default to 13

Indicators/KlingerVolumeOscillator.cs
mattfair: So I updated to use RollingWindow, not sure if it makes things simpler.  But I also renamed variables to make it easier to follow.  The initialization logic is such that we cannot use .IsReady while we prime the windows, instead I use .Count < 2.

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,171 @@
...
+    /// crossovers that can provide insights into market trends and potential turning points.
+    /// </summary>
+    public class KlingerVolumeOscillator : TradeBarIndicator, IIndicatorWarmUpPeriodProvider
+    {
+        private readonly ExponentialMovingAverage _fastEma;
+        private readonly ExponentialMovingAverage _slowEma;
+
+        private readonly RollingWindow<decimal> _price;
+        private readonly RollingWindow<decimal> _dailyMovement;
+        private readonly RollingWindow<decimal> _comulativeMovement;
```
Martin-Molinero: typo `cumulative`

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,171 @@
...
+            WarmUpPeriod = Math.Max(fastPeriod, Math.Max(slowPeriod, signalPeriod)) + 2;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="KlingerVolumeOscillator"/> class with specified fast, slow periods.
+        /// </summary>
+        /// <param name="fastPeriod">The fast EMA period.</param>
+        /// <param name="slowPeriod">The slow EMA period.</param>
+        /// <param name="signalPeriod">The signal line period (default is 13).</param>
+        public KlingerVolumeOscillator(int fastPeriod, int slowPeriod, int signalPeriod = 13)
```
Martin-Molinero, line 84: `signalPeriod = 13` nice, let's just add this default for `KVO()` too 👌

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,171 @@
...
+                // Continue flow accumulation in the same trend direction
+                trendMovement += _comulativeMovement[0];
+            }
+            _comulativeMovement.Add(trendMovement);
+
+            // Volume force:  strength of volume flow in the direction of the trend.
+            // There are various definitions of volume force, this is what we used in our implementation:
+            // https://github.com/nardew/talipp/blob/70dc9a26889c9c9329e44321e1362c4db43dbcc3/talipp/indicators/KVO.py#L85
+            // https://www.tradingview.com/support/solutions/43000589157-klinger-oscillator/
+            var volumeForce = trendMovement == 0m ? 0m : bar.Volume * Math.Abs(2m * (todaysMovement / _comulativeMovement[0] - 1m)) * currentTrend * 100m;
```
Martin-Molinero: potential ` / _comulativeMovement[0]` division by zero, given it's high-low?
Only `_comulativeMovement[0]` is being used, why does it need size 2 window?
`var hasMovement = _comulativeMovement[0] != 0;` this is like a missing isready check too
_
could _comulativeMovement be remove maybe even, just reset _dailyMovement on trend change or something like that?

Indicators/KlingerVolumeOscillator.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            // update moving averages
+            _fastEma.Update(new IndicatorDataPoint(bar.EndTime, volumeForce));
+            _slowEma.Update(new IndicatorDataPoint(bar.EndTime, volumeForce));
```
Martin-Molinero: minor but duplicate `new IndicatorDataPoint(bar.EndTime, volumeForce)`

Indicators/KlingerVolumeOscillator.cs
Martin-Molinero: It seems to be at least ` if (_price.Count < 2) && if (_trendDirection.Count < 2)` can be replaced by IsReady?

Indicators/KlingerVolumeOscillator.cs
mattfair: cleaning up to be clearer.

PR #8877, opened 2025-07-16
Feature 6620 implement td sequential candlestick indicator

rathnagiri:
#### Description
This PR introduces the TdSequential technical indicator into the Lean framework.
It implements the TD Sequential algorithm with support for:

TD Buy/Sell Setup phases (up to 9 steps)

Perfect Setup detection

TD Buy/Sell Countdown phases (up to 13 steps)

Countdown cancellation via TDST resistance/support levels

State encoding of phase and step into a single decimal for easy usage in algorithms

The indicator is implemented as a subclass of IndicatorBase<TradeBar> and encodes the state using the format:
```
<phaseCode>.<step>
e.g., 1.04 = TD Buy Setup, Step 4
      5.09 = TD Buy Setup Perfect at Step 9
      6.13 = TD Buy Countdown Completed
```

#### Motivation and Context
This change adds support for the widely-used TD Sequential indicator, which helps traders identify trend exhaustion and potential reversal points.
The indicator is commonly used in professional trading environments and is now available to Lean algorithm authors with encoded values for seamless integration.

Yes, An update is required to include the TdSequential indicator in:
Algorithm.Indicators.cs for auto-initialization
Indicator documentation for usage patterns and interpretation of encoded values

Tests verified using broad unit tests covering all scenario outputs related to each phase of the output

#### Types of changes

#### Checklist:

Many of the tests fail out of the box without any of my changes.

### DIFF
--- a/Indicators/TomDemarkSequential.cs
+++ b/Indicators/TomDemarkSequential.cs
@@ -67,39 +67,31 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
     private const int RequiredSamples = 6;
 
     private static decimal Default => (decimal)TomDemarkSequentialPhase.None;
-
-    private int _setupStepCount;
-
-    private int _countdownStepCount;
-
     private TomDemarkSequentialPhase _currentPhase;
     
-    private decimal _resistancePrice; // highest high of the 9-bar Tom Demark Sequential buy setup (indicates resistance)
-    private decimal _supportPrice; // lowest low of the 9-bar Tom Demark Sequential sell setup (indicates support)
-
     /// <summary>
     /// Gets the current resistance price calculated during a Tom Demark Sequential buy setup.
     /// This is the highest high of the 9-bar setup and can act as a resistance level.
     /// </summary>
-    public decimal ResistancePrice => _resistancePrice;
+    public decimal ResistancePrice { get; private set; }
 
     /// <summary>
     /// Gets the current support price calculated during a Tom Demark Sequential sell setup.
     /// This is the lowest low of the 9-bar setup and can act as a support level.
     /// </summary>
-    public decimal SupportPrice => _supportPrice;
+    public decimal SupportPrice { get; private set; }
     
     /// <summary>
     /// Gets the current Setup step count in the active Tom Demark Sequential (Buy/Sell) Setup phase 
     /// (e.g., 1 to 9 in Setup), 0 if not in setup phase.
     /// </summary>
-    public int SetupPhaseStepCount => _setupStepCount;
+    public int SetupPhaseStepCount { get; private set; }
     
     /// <summary>
     /// Gets the current Countdown step count in the active TomDemark Sequential (Buy/Sell) Countdown phase
     /// (e.g., 1 to 13 in Setup), 0 if bar is not in Countdown phase.
     /// </summary>
-    public int CountdownPhaseStepCount => _countdownStepCount;
+    public int CountdownPhaseStepCount { get; private set; }
     
     /// <summary>
     /// Initializes a new instance of the <see cref="TomDemarkSequential"/> indicator.
@@ -130,7 +122,10 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
     /// </summary>
     public override void Reset()
     {
-        _setupStepCount = 0;
+        SetupPhaseStepCount = 0;
+        CountdownPhaseStepCount = 0;
+        SupportPrice = 0;
+        ResistancePrice = 0;
         _currentPhase = TomDemarkSequentialPhase.None;
         base.Reset();
     }
@@ -195,15 +190,15 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
     {
         if (current.Close >= bar2Ago.High)
         {
-            _countdownStepCount++;
-            if (_countdownStepCount == MaxCountdownCount)
+            CountdownPhaseStepCount++;
+            if (CountdownPhaseStepCount == MaxCountdownCount)
             {
                 _currentPhase = TomDemarkSequentialPhase.None;
             }
             return (decimal)TomDemarkSequentialPhase.SellCountdown;
         }
 
-        if (current.Close < _supportPrice)
+        if (current.Close < SupportPrice)
         {
             _currentPhase = TomDemarkSequentialPhase.None;
         }
@@ -215,15 +210,15 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
     {
         if (current.Close <= bar2Ago.Low)
         {
-            _countdownStepCount++;
-            if (_countdownStepCount == MaxCountdownCount)
+            CountdownPhaseStepCount++;
+            if (CountdownPhaseStepCount == MaxCountdownCount)
             {
                 _currentPhase = TomDemarkSequentialPhase.None;
             }
             return (decimal)TomDemarkSequentialPhase.BuyCountdown;
         }
 
-        if (current.Close > _resistancePrice)
+        if (current.Close > ResistancePrice)
         {
             _currentPhase = TomDemarkSequentialPhase.None;
         }
@@ -240,17 +235,17 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
     {
         if (current.Close > bar4Ago.Close)
         {
-            _setupStepCount++;
-            if (_setupStepCount == MaxSetupCount)
+            SetupPhaseStepCount++;
+            if (SetupPhaseStepCount == MaxSetupCount)
             {
                 var isPerfect = IsSellSetupPerfect(window);
                 _currentPhase = TomDemarkSequentialPhase.SellCountdown;
                 //  Check if the close of bar 9 or current bar is greater than the high two bars earlier
                 if (current.Close > bar2Ago.High)
                 {
-                    _countdownStepCount = 1;
+                    CountdownPhaseStepCount = 1;
                 }
-                _supportPrice = window.Skip(window.Count - MaxSetupCount).Take(MaxSetupCount).Min(b => b.Low);
+                SupportPrice = window.Skip(window.Count - MaxSetupCount).Take(MaxSetupCount).Min(b => b.Low);
 
                 return
                     (decimal)(isPerfect
@@ -274,16 +269,16 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
     {
         if (current.Close < bar4Ago.Close)
         {
-            _setupStepCount++;
-            if (_setupStepCount == MaxSetupCount)
+            SetupPhaseStepCount++;
+            if (SetupPhaseStepCount == MaxSetupCount)
             {
                 var isPerfect = IsBuySetupPerfect(window);
                 _currentPhase = TomDemarkSequentialPhase.BuyCountdown;
-                _resistancePrice = window.Skip(window.Count - MaxSetupCount).Take(MaxSetupCount).Max(b => b.High);
+                ResistancePrice = window.Skip(window.Count - MaxSetupCount).Take(MaxSetupCount).Max(b => b.High);
                 // check the close of bar 9 is less than the low two bars earlier.
                 if (current.Close < bar2Ago.Low)
                 {
-                    _countdownStepCount = 1;
+                    CountdownPhaseStepCount = 1;
                 }
 
                 return 
@@ -308,7 +303,7 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
         if (prevBar.Close > prevBar4Ago.Close && current.Close < bar4Ago.Close)
         {
             _currentPhase = TomDemarkSequentialPhase.BuySetup;
-            _setupStepCount = 1;
+            SetupPhaseStepCount = 1;
 
             return (decimal)TomDemarkSequentialPhase.BuySetup;
         }
@@ -317,7 +312,7 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
         if (prevBar.Close < prevBar4Ago.Close && current.Close > bar4Ago.Close)
         {
             _currentPhase = TomDemarkSequentialPhase.SellSetup;
-            _setupStepCount = 1;
+            SetupPhaseStepCount = 1;
 
             return (decimal)TomDemarkSequentialPhase.SellSetup;
         }


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -2199,6 +2199,23 @@ public TargetDownsideDeviation TDD(Symbol symbol, int period, double minimumAcce
             return targetDownsideDeviation;
         }
 
+
```
Martin-Molinero: nit - unrequired space

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -2199,6 +2199,23 @@ public TargetDownsideDeviation TDD(Symbol symbol, int period, double minimumAcce
...
+        /// updated on the symbol's subscription resolution.
+        /// </summary>
+        /// <param name="symbol">The symbol whose TD Sequential we want</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to casting the input value to a TradeBar</param>
+        /// <returns>The TD Sequential indicator for the requested symbol over the specified period</returns>
+        public TdSequential TDS(Symbol symbol, Resolution? resolution = null, Func<IBaseData, TradeBar> selector = null)
+        {
+            var name = CreateIndicatorName(symbol, "TDS", resolution);
+            var tdSequential = new TdSequential(name);
```
Martin-Molinero: > TdSequential

Let's expand the class to it's full name `TomDemarkSequential`

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
+using QuantConnect.Data.Market;
```
Martin-Molinero: minor but missing license header, same for the other file additions

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using QuantConnect.Data;
+
+namespace QuantConnect.Indicators
+{    
+    /// <summary>
+    /// Represents the TD Sequential indicator, which is used to identify potential trend exhaustion points.
```
Martin-Molinero: > TD Sequential

In the xml documentation let's please expand it's name too Tom Demark Sequential

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// sources:
+    /// https://demark.com/sequential-indicator/
+    /// https://practicaltechnicalanalysis.blogspot.com/2013/01/tom-demark-sequential.html
+    /// https://medium.com/traderlands-blog/tds-td-sequential-indicator-2023-f8675bc5d14
+    /// </summary>
+    /// <remarks>
+    /// The TD Sequential indicator is a popular technical analysis tool that helps traders identify potential reversal points
+    /// in the market. It consists of two main components: the setup phase and the countdown phase. The setup phase counts the number of consecutive bars that meet specific criteria, while the countdown phase
+    /// counts the number of bars that follow the setup phase to confirm a potential reversal.
+    /// This implementation focuses on the setup phase, counting the number of consecutive bars that meet the criteria for a bullish or bearish setup.
```
Martin-Molinero: > This implementation focuses on the setup phase

It's missing the countdown phase?

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// </remarks>
+    /// <seealso cref="IndicatorBase{T}"/>
+    /// <seealso cref="TradeBar"/>
+    public class TdSequential : IndicatorBase<TradeBar>, IIndicatorWarmUpPeriodProvider
```
Martin-Molinero: `TradeBar` This indicator does not require the `Volume` property characteristic of trade information, instead it should use BarIndicator as base which is more generic and allows for quote information as source

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+    {
+        /// <summary>
+        /// The Max Step count in Setup phase
+        /// </summary>
+        public const int MaxSetupCount = 9;
+        
+        /// <summary>
+        /// Max Step count in Countdown phase
+        /// </summary>
+        public const int MaxCountdownCount = 13;
```
Martin-Molinero: Both these const seem to be a setting which should be passed as constructor argument, right? 9/13 could be the default values used at `Algorithm/QCAlgorithm.Indicators.cs`, can see other indicators as a pattern to follow
Also don't think they should be public but just private  readonly fields

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// <param name="name"></param>
+        public TdSequential(string name) : base(name) { }
+
+        /// <inheritdoc />
+        public override bool IsReady => _bars.Count >= 5;
+
+        /// <inheritdoc />
+        public int WarmUpPeriod => 5;
+
+        /// <inheritdoc />
```
Martin-Molinero: Sorry our documentation engine does not support `/// <inheritdoc />` atm, we just use the full version instead

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// <returns>The encoded state of the TD Sequential indicator for the current bar.</returns>
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            _bars.Add(input);
+
+            if (_bars.Count < 5)
+                return Default;
+
+            if (_bars.Count > 10)
+                _bars.RemoveAt(0);
```
Martin-Molinero: Take a look at the `RollingWindow` class, it's usages/properties/functionality, will help clean this up a bit I think 👍 . Ideally when possible should always reuse existing classes/code to avoid duplication/complexity/etc
 
edit: even what if the base class of the indicator was a `WindowIndicator`? That would clean it even further I think if possible
___
These values seems a bit magical, what/why 10/5, same for WarmUpPeriod and IsReady, should this be an indicator setting too with a default value?
___
Also regarding the missing `{}` let's add them all over the place 👌  and make sure they are in new lines, eg:
```
if (something)
{
    a =10 
}
```

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+        private bool _inBuyCountdown;
+        private bool _inSellCountdown;
+
+        private decimal _tdstResistance; // higeest high of the 9-bar TD Sequential buy setup (indicates resistance)
+        private decimal _tdstSupport; // lowest low of the 9-bar TD Sequential sell setup (indicates support)
+
+        /// <summary>
+        /// Creates a new instance of <see cref="TdSequential"/> indicator
+        /// </summary>
+        /// <param name="name"></param>
```
Martin-Molinero: > /// <param name="name"></param> 

nit missing to be filled. Let's place the `{}` in a new line for each to follow the coding style 👍

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+        private int _setupCount;
+        private int _countdownCount;
+
+        private bool _inBuySetup;
+        private bool _inSellSetup;
+        private bool _inBuyCountdown;
+        private bool _inSellCountdown;
+
+        private decimal _tdstResistance; // higeest high of the 9-bar TD Sequential buy setup (indicates resistance)
+        private decimal _tdstSupport; // lowest low of the 9-bar TD Sequential sell setup (indicates support)
```
Martin-Molinero: minor but `_tdstSupport/_tdstResistance` maybe we can rename these to `_supportPrice/_resistancePrice` or level if it's not a price

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        private readonly List<TradeBar> _bars = [];
+
+        private int _setupCount;
+        private int _countdownCount;
+
+        private bool _inBuySetup;
+        private bool _inSellSetup;
+        private bool _inBuyCountdown;
+        private bool _inSellCountdown;
```
Martin-Molinero: We can't be in setup and in countdown at the same time? and I think they are not used at the same time, so we could probably merge them into a single variable?

----

Regarding the `bool` also seems like a single state machine? meaning maybe we can replace all of them with an enum using TdSequentialPhase values?

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            return (decimal)phase + (step / 100m);
+        }
+
+    }
+
+    /// <summary>
+    /// Represents the different phases of the TD Sequential indicator.
+    /// </summary>
+    public enum TdSequentialPhase
```
Martin-Molinero: Let's expand the `TdSequentialPhase` enum name too 👍

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+                _setupCount++;
+                if (_setupCount == MaxSetupCount)
+                {
+                    var isPerfect = IsBuySetupPerfect();
+                    _inBuySetup = false;
+                    _inBuyCountdown = true;
+                    _tdstResistance = _bars.Skip(_bars.Count - MaxSetupCount).Take(MaxSetupCount).Max(b => b.High);
+                    _setupCount = 0;
+
+                    return EncodeState(isPerfect ? TdSequentialPhase.BuySetupPerfect : TdSequentialPhase.BuySetup, 9);
```
Martin-Molinero: - `9` magic nine again, ideally not hardcoded
- HandleBuySetupPhase/HandleSellSetupPhase have quite a bit in common, think they could probably be merged at least some parts of them. Probably same for HandleBuyCountDown/HandleSellCountDown. I think once the variables are merged too it will be much clear the duplication and path for cleanup

QuantConnect.Lean.sln.DotSettings
```diff
@@ -30,11 +30,13 @@
...
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpUseContinuousIndentInsideBracesMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EPredefinedNamingRulesToUserRulesUpgrade/@EntryIndexedValue">True</s:Boolean>
```
Martin-Molinero: Let's revert the changes in the `QuantConnect.Lean.sln.DotSettings` file 👍

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+            var bar8 = _bars[^2];
+            var bar9 = _bars[^1];
+
+            return bar8.High >= bar6.High && bar8.High >= bar7.High ||
+                   bar9.High >= bar6.High && bar9.High >= bar7.High;
+        }
+
+        private decimal EncodeState(TdSequentialPhase phase, int step)
+        {
+            return (decimal)phase + (step / 100m);
```
Martin-Molinero: Interesting idea! so the indicator value holds both the state as integer and the step count as a decimal
I would strongly suggest though simplifying it, make the indicator return it's state enum value and just make count a single public property. Will make this code simpler and consumer code simpler too

Indicators/TdSequential.cs
```diff
@@ -0,0 +1,313 @@
...
+            var bar8 = _bars[^2];
+            var bar9 = _bars[^1];
+
+            return bar8.Low <= bar6.Low && bar8.Low <= bar7.Low ||
+                   bar9.Low <= bar6.Low && bar9.Low <= bar7.Low;
+        }
+
+        private bool IsSellSetupPerfect()
+        {
+            if (_bars.Count < MaxSetupCount) return false;
```
Martin-Molinero: Sure we need these checks 🤔, we will only be here if `_setupCount == MaxSetupCount` which means we were called at least setup count so we should have bars for all steps already

Martin-Molinero (changes requested):
Hi @rathnagiri! Thank you for the contribution 💪 

Leaving a few comments around, TLDR: I think there's some room for clean up and consolidation to make the indicator cleaner and clearer

Algorithm/QCAlgorithm.Indicators.cs
rathnagiri: done

Indicators/TdSequential.cs
rathnagiri: done

Indicators/TdSequential.cs
rathnagiri: done

Indicators/TdSequential.cs
rathnagiri: nice catch., missed updating the header after my first implementation that focussed just on the setup algos

Indicators/TdSequential.cs
rathnagiri: > Both these const seem to be a setting which should be passed as constructor argument, right? 9/13 could be the default values used at `Algorithm/QCAlgorithm.Indicators.cs`, can see other indicators as a pattern to follow Also don't think they should be public but just private readonly fields

@Martin-Molinero Thank you for the suggestion. You're right in pointing out that constants generally be private unless there’s a need for public access. I've made both MaxSetupCount and MaxCountdownCount private.

However, I've kept them as constants rather than making them configurable via the constructor. This is because the values 9 and 13 are fundamental to the Tom Demark Sequential algorithm as originally defined and are not intended to be adjusted. Making them configurable could introduce confusion or unintended behavior that deviates from the indicator's standard interpretation.

Let me know if you think it would help to still make this flexible in future use cases, but for now, keeping these values fixed seems the most correct and aligned with the intent of the algorithm.

Indicators/TdSequential.cs
rathnagiri: done

Indicators/TdSequential.cs
rathnagiri: Thanks for the great suggestion @Martin-Molinero ! I've refactored the indicator to inherit from WindowIndicator<IBaseDataBar>, which might simplify the sliding-window logic and keeps the implementation more maintainable and I need to test, If not will use RollingWindow. Looks like WindowIndicator also integrates cleanly with IIndicatorWarmUpPeriodProvider and thereby removes the need to manually manage the bar buffer, aligning better with Lean's indicator design patterns. Also for style, added {} braces across single line control blocks .,

rathnagiri: > These values seems a bit magical, what/why 10/5, same for WarmUpPeriod and IsReady, should this be an indicator setting too with a default value?

@Martin-Molinero 
I've replaced the hardcoded values (5 and 10) with named constants RequiredSamples and WindowSize, along with comments to clarify their purpose.  I’ve kept them as constants since they are intrinsic to the TD Sequential logic and not typically variable. 5 is the minimum needed for setup phases evaluation, and 10 provides enough buffer for lookbacks like perfect setups and countdowns.

I’ve kept them as constants since making them configurable might give the impression they can be tuned arbitrarily, which could lead to misinterpretation of the indicator’s purpose. Happy to revisit if you think exposing them adds anyvalue.

Feel free to suggest !

Indicators/TdSequential.cs
rathnagiri: Thanks! I've updated the <param> documentation and adjusted the constructor formatting to follow the project's style conventions

Indicators/TdSequential.cs
rathnagiri: Thanks for the suggestion!.  I have renamed the fields to _resistancePrice and _supportPrice for clarity, as they store actual support or ressitance price levels from the 9-bar setups.

Indicators/TdSequential.cs
rathnagiri: Good point ! Yes, since the setup and countdown states are mutually exclusive, I've refactored them into a single state phase based on already existing enum to simplify state tracking and avoid having to maintain several boolean flags. Also merged the two count variables as a single stepCount variable
Thanks for the suggestion!

Indicators/TdSequential.cs
rathnagiri: > * `9` magic nine again, ideally not hardcoded

Good point about the hardcoded 9. replaced it with MaxSetupCount

rathnagiri: > * HandleBuySetupPhase/HandleSellSetupPhase have quite a bit in common, think they could probably be merged at least some parts of them. Probably same for HandleBuyCountDown/HandleSellCountDown. I think once the variables are merged too it will be much clear the duplication and path for cleanup

Regarding merging the HandleBuy/SellSetupPhase and HandleBuy/SellCountdown methods: I initially explored that direction, but found that while the structure is similar, the control flow diverges in subtle but meaningful ways (e.g., Close < vs Close >, Max(High) vs Min(Low), and distinct phase transitions). Attempting to generalize them increased complexity and made the logic harder to read or follow and maintain.

Since these phases are quite domain-specific, keeping them separate improves readability and makes future enhancements or debugging easier. That said, if more logic ends up being shared over time (especially with variables or phase transitions), I’m happy to revisit and abstract out the common parts where it truly simplifies things.

Indicators/TdSequential.cs
rathnagiri: Thanks for the suggestion. I did initially considered returning just the Phase with a public Step property, but to my knowledge, it doesn't seem to align well with how the indicator is going to be consumed. The only method consumers call is ComputeNextValue, which returns a decimal per bar, so exposing Step separately isn’t feasible without altering the base class design of how these indicators are consumed.

The encoded **phase.step** format allows each bar to carry both the _state_ and its _step_ within the phase, which is especially helpful for UI/visual tools that want to mark progression (e.g., bar 1 through 9 of a Setup) directly. While this format may require brief documentation, it's compact, deterministic, and easy to parse for downstream consumers.
That said, I'm open to feedback if there are any alternative ways of consumption that i'm not aware of.

Indicators/TdSequential.cs
rathnagiri: Good point ., cleaned up that redundant condition

rathnagiri:
Hi @Martin-Molinero 

I believe I have now addressed all the review comments, including the suggestions around method consolidation, hardcoded constants, and the indicator value encoding. I’ve also added clarifications where applicable to explain some of the design decisions.

When you get a chance, please take another look and let me know if anything else needs attention.

Thanks again for your valuable feedback!

Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -2198,6 +2198,22 @@ public TargetDownsideDeviation TDD(Symbol symbol, int period, double minimumAcce
...
+        
+        /// <summary>
+        /// Creates a new TomDemark Sequential candlestick indicator for the symbol. The indicator will be automatically
+        /// updated on the symbol's subscription resolution.
+        /// </summary>
+        /// <param name="symbol">The symbol whose TD Sequential we want</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to casting the input value to a TradeBar</param>
+        /// <returns>The TomDemark Sequential indicator for the requested symbol over the specified period</returns>
+        public TomDemarkSequential TDS(Symbol symbol, Resolution? resolution = null, Func<IBaseData, TradeBar> selector = null)
```
Martin-Molinero: Minor but the selector could be IBaseDataBar + updating xml comment 👍

Indicators/TdSequential.cs
Martin-Molinero: Note that the ComputeNextValue is actually just used by the engine in 99.9% of the cases, u can take a look at `IndicatorSuiteAlgorithm/MovingAverageCrossAlgorithm` user code usually access the indicator properties or it's current value through `.current` or even just relying on the implicit conversion (indicator->decimal, which uses current). We usually prefer explicit/simple, that's why I suggest make the indicator return it's state enum value and just make count a single public property

Indicators/TomDemarkSequential.cs
```diff
@@ -0,0 +1,353 @@
...
+            }
+
+            return EncodeState(TomDemarkSequentialPhase.BuySetup, _stepCount);
+        }
+
+        ResetStepAndPhase();
+        return Default;
+    }
+
+    private decimal InitializeSetupPhase(IBaseDataBar current, IBaseDataBar bar4Ago)
```
Martin-Molinero: From my understanding I think this logic is missing a price flip check, mind double checking this? ref https://practicaltechnicalanalysis.blogspot.com/2013/01/tom-demark-sequential.html

Indicators/TomDemarkSequential.cs
```diff
@@ -0,0 +1,353 @@
...
+
+    private static bool IsSellSetupPerfect(IReadOnlyWindow<IBaseDataBar> window)
+    {
+        var bar6 = window[3];
+        var bar7 = window[2];
+        var bar8 = window[1];
+        var bar9 = window[0];
+
+        return bar8.High >= bar6.High && bar8.High >= bar7.High ||
+            bar9.High >= bar6.High && bar9.High >= bar7.High;
```
Martin-Molinero: Should these be just `>`/`<`  not sure the `=` is valid in the definition? ref https://practicaltechnicalanalysis.blogspot.com/2013/01/tom-demark-sequential.html

Indicators/TomDemarkSequential.cs
```diff
@@ -0,0 +1,353 @@
...
+    private const int RequiredSamples = 5;
+
+    private static decimal Default => EncodeState(TomDemarkSequentialPhase.None, 0);
+
+    private int _stepCount;
+
+    private TomDemarkSequentialPhase _currentPhase;
+    
+    private decimal _resistancePrice; // highest high of the 9-bar Tom Demark Sequential buy setup (indicates resistance)
+    private decimal _supportPrice; // lowest low of the 9-bar Tom Demark Sequential sell setup (indicates support)
```
Martin-Molinero: Let's make these a public readonly property, they seems interesting price points to be consumed 👍

Martin-Molinero (approved):
Thank you! Leaving some last comments

Algorithm/QCAlgorithm.Indicators.cs
rathnagiri: done

Indicators/TomDemarkSequential.cs
rathnagiri: You are right., strict comparison aligns better with the definition mentioned in the website

Indicators/TomDemarkSequential.cs
rathnagiri: Thanks! Since the values are updated during execution, I’ve added SupportPrice and ResistancePrice as read-only public properties instead of readonly fields. XML docs included—happy to tweak further if needed.

Indicators/TomDemarkSequential.cs
rathnagiri: @Martin-Molinero : I research on this further since the original version in https://demark.com/sequential-indicator/ didn’t include the price flip logic, I wrongly assumed price flip actually the initial 9bar setup. I now added this logic to detect bullish or bearish price flip's before determining the setup phase. That’s also why RequiredSamples was bumped to 6 — we need both the previous bar and the one 4 bars before that to detect a flip.

Indicators/TdSequential.cs
rathnagiri: I’ve updated ComputeNextValue() to return just the phase enum. Additionally, I’ve exposed two separate public properties for SetupPhaseStepCount and CountdownPhaseStepCount, since step values can overlap (e.g., bar 9 of the setup phase might also be bar 1 of the countdown phase and we need to track them). This distinction helps users accurately track progress using the above properties in each phase without ambiguity.

Indicators/TomDemarkSequential.cs
```diff
@@ -115,7 +130,8 @@ public TomDemarkSequential(string name = "TomDemarkSequential")
     /// </summary>
     public override void Reset()
     {
-        ResetStepAndPhase();
+        _setupStepCount = 0;
+        _currentPhase = TomDemarkSequentialPhase.None;
```
Martin-Molinero, line 129: Missing to reset some of the private fields I believe, like `_countdownStepCount/_supportPrice/_resistancePrice..`

Indicators/TomDemarkSequential.cs
```diff
@@ -87,6 +89,18 @@ public class TomDemarkSequential : WindowIndicator<IBaseDataBar>
...
+    /// Gets the current Setup step count in the active Tom Demark Sequential (Buy/Sell) Setup phase 
+    /// (e.g., 1 to 9 in Setup), 0 if not in setup phase.
+    /// </summary>
+    public int SetupPhaseStepCount => _setupStepCount;
+    
+    /// <summary>
+    /// Gets the current Countdown step count in the active TomDemark Sequential (Buy/Sell) Countdown phase
+    /// (e.g., 1 to 13 in Setup), 0 if bar is not in Countdown phase.
+    /// </summary>
+    public int CountdownPhaseStepCount => _countdownStepCount;
```
Martin-Molinero: minor might be simpler/cleaner if the property just had a public get and a private set, no need to define the backing filed manually for these -> will be able to delete a few lines

Martin-Molinero (commented):
Looks good! just minor bug in reset

Indicators/TomDemarkSequential.cs
rathnagiri: done

Indicators/TomDemarkSequential.cs
rathnagiri: @Martin-Molinero done

PR #8880, opened 2025-07-18
Log the Number of Active Subscriptions Before and After Warm Up #8779

### LINKED ISSUE #8779: Log the Number of Active Subscriptions Before and After Warm Up
#### Expected Behavior
Before and after warm-up, we are informed of the number of subscriptions, including internal.

#### Actual Behavior
We don't have this information.

#### Motivation
Knowing the number of subscriptions is essential for understanding memory consumption. For instance, if we have 200 crypto coins, LEAN will create 200 subscriptions, which is too many for an L-MICRO node.

#### Potential Solution
N/A

#### Checklist

AibekMinbaev:
#### Description

Log the Number of Active Subscriptions (including internal) Before and After Warm Up 
#### Related Issue

Closes #8779
#### Motivation and Context

NA

Yes

Tested by running WarmupAlgorithm.cs 

Log: 
TRACE:: Debug: Algorithm starting warm up...
TRACE:: Debug: Subscriptions count: 3
….
TRACE:: Debug: Algorithm finished warming up.
TRACE:: Debug: Subscriptions count: 4

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/AlgorithmManager.cs
+++ b/Engine/AlgorithmManager.cs
@@ -651,6 +651,7 @@ namespace QuantConnect.Lean.Engine
             {
                 nextWarmupStatusTime = DateTime.UtcNow.AddSeconds(1);
                 algorithm.Debug("Algorithm starting warm up...");
+                algorithm.Debug($"Subscriptions count: {algorithm.SubscriptionManager.Count}");
                 results.SendStatusUpdate(AlgorithmStatus.History, $"{warmingUpPercent}");
             }
             else
@@ -699,6 +700,7 @@ namespace QuantConnect.Lean.Engine
                     // we trigger this callback here and not internally in the algorithm so that we can go through python if required
                     algorithm.OnWarmupFinished();
                     algorithm.Debug("Algorithm finished warming up.");
+                    algorithm.Debug($"Subscriptions count: {algorithm.SubscriptionManager.Count}");
                     results.SendStatusUpdate(AlgorithmStatus.Running, "100");
                 }
                 yield return timeSlice;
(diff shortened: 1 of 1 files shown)


Engine/AlgorithmManager.cs
```diff
@@ -699,6 +700,7 @@ private IEnumerable<TimeSlice> Stream(IAlgorithm algorithm, ISynchronizer synchr
                     // we trigger this callback here and not internally in the algorithm so that we can go through python if required
                     algorithm.OnWarmupFinished();
                     algorithm.Debug("Algorithm finished warming up.");
+                    algorithm.Debug($"Subscriptions count: {algorithm.SubscriptionManager.Count}");
```
Martin-Molinero: - I think these additions should just be a `Log.Trace` call, please see other examples in this file for log format
- I don't think the log at the start/above will actually reflect universe selection initial trigger, so might not be so useful, we should probably log once after the first time slice and here

Martin-Molinero: > I don't think the log at the start/above will actually reflect universe selection initial trigger

This can be tested with universe selection based algorithm maybe csharp `FundamentalCustomSelectionTimeWarmupRegressionAlgorithm` 

__

Thanks for the help @AibekMinbaev! 💪

AibekMinbaev: @Martin-Molinero 
I made changes reflecting the review comments. Please, take a look and let me know if there any changes needed. 
I am quite new to Lean codebase, the place for logging on the first time slice might be off. Also, should contributers  squash commits for clean git history ? thank you

Martin-Molinero: Hey ! Thank you, I think would be nice to keep them in the same method, inside Stream, in the `synchronizer.StreamData` for each would be great 🙏 

>Also, should contributers squash commits for clean git history 

We squash merge so it's okay 👍

AibekMinbaev: @Martin-Molinero 
Resolved review comments, please take a look. Thank you

Issue #8881, opened 2025-07-18
Allow relaxed comparison in PivotPointsHighLow


ASAD-BE18:
## Problem
The current `PivotPointsHighLow` indicator uses strict inequality comparisons (`>` and `<`) when detecting pivot points. This means that if the central bar's high/low value equals any of its neighbors' values, it won't be identified as a pivot point. In some market conditions, this strict behavior may miss valid pivot points where price levels are tied.

Martin-Molinero:
Strict behavior seems to be the default among other implementations, but we can add a new constructor argument setting `(.., strict = true)` to enable the use of equality comparison when set to false.
This is an indicator that's been there for years, so to keep backwards compatibility at `QCAlgorithm.Inidicators` we will need to add a new overload, the new one should be called by the older one. New unit tests should be added asserting the new behavior, see PivotPointsHighLowTests

dacrowlah:
@Martin-Molinero - this is ready for review, when ever you have time.

PR #9033, opened 2025-10-13
Add optional strict parameter to PivotPointsHighLow indicator

dacrowlah:
## Description
Adds optional strict parameter (default true) to PivotPointsHighLow indicator to allow relaxed comparison for pivot point detection.

- When strict=true (default): Uses strict inequalities (> and <) - existing behavior
- When strict=false: Uses relaxed inequalities (>= and <=) - allows equal values to be detected as pivot points

## Related Issue
Addresses #8881

## Motivation and Context
Users requested the ability to detect pivot points when a bar's high/low equals surrounding bars, which the strict comparison mode rejects.

## How Has This Been Tested?
- Added 5 comprehensive unit tests using TestCase attributes
- All existing tests continue to pass (backwards compatibility verified)
- Tests use fixed reference time instead of DateTime.UtcNow for performance

## Types of changes

## Checklist:
- [x] My code follows the code style of this project
- [x] I have read the CONTRIBUTING document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] My branch follows the naming convention bug-8881-relaxed-pivot-comparison

### DIFF
--- a/Tests/Indicators/PivotPointsHighLowTests.cs
+++ b/Tests/Indicators/PivotPointsHighLowTests.cs
@@ -162,22 +162,6 @@ namespace QuantConnect.Tests.Indicators
                 Assert.AreEqual(1, highPivots.Length, "Relaxed mode should accept equal high values");
                 Assert.AreEqual(100, highPivots[0].Value);
             }
-        }
-
-        [TestCase(true)]
-        [TestCase(false)]
-        public void StrictVsRelaxedLowPivotDetection(bool strict)
-        {
-            var indicator = new PivotPointsHighLow(2, 2, strict: strict);
-            var referenceTime = new DateTime(2020, 1, 1);
-
-            // Create 5 bars where middle bar low EQUALS neighbors
-            // All bars have low = 50, which should only detect pivot in relaxed mode
-            for (var i = 0; i < 5; i++)
-            {
-                var bar = new TradeBar(referenceTime.AddSeconds(i), Symbols.AAPL, 95, 100, 50, 95, 1000);
-                indicator.Update(bar);
-            }
 
             var lowPivots = indicator.GetLowPivotPointsArray();
 
@@ -190,7 +174,7 @@ namespace QuantConnect.Tests.Indicators
             {
                 // Relaxed mode: middle bar low (50) is <= neighbors (50), so YES pivot
                 Assert.AreEqual(1, lowPivots.Length, "Relaxed mode should accept equal low values");
-                Assert.AreEqual(50, lowPivots[0].Value);
+                Assert.AreEqual(90, lowPivots[0].Value);
             }
         }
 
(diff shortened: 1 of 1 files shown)


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -1722,9 +1722,26 @@ public OnBalanceVolume OBV(Symbol symbol, Resolution? resolution = null, Func<IB
...
+        /// <param name="lengthLow">The number of surrounding bars whose low values should be more than the current bar's for the bar low to be marked as low pivot point</param>
+        /// <param name="lastStoredValues">The number of last stored indicator values</param>
+        /// <param name="resolution">The resolution</param>
+        /// <param name="selector">Selects a value from the BaseData to send into the indicator, if null defaults to casting the input value to a TradeBar</param>
+        /// <param name="strict">When true (default), uses strict inequalities (greater than and less than). When false, uses relaxed inequalities (greater than or equal and less than or equal) allowing equal values to be detected as pivot points.</param>
+        /// <returns>The PivotPointsHighLow indicator for the requested symbol.</returns>
+        [DocumentationAttribute(Indicators)]
+        public PivotPointsHighLow PPHL(Symbol symbol, int lengthHigh, int lengthLow, int lastStoredValues = 100, Resolution? resolution = null, Func<IBaseData, IBaseDataBar> selector = null, bool strict = true)
```
jhonabreul: The selector is usually the last parameter to these helper methods, so to keep the API consistent, you could add the `strict` parameter in another position.

Also, if you call `PPHL(symbol, lengthHigh, lengthLow)` I think it will not compile because of an ambiguity on the overloads resolution. Maybe a unit test might shed some light on it

dacrowlah: Re: Parameter Ordering 

I've addressed the review feedback by removing the wrapper overload (eliminating ambiguity) and reordering parameters to: `PPHL(symbol, lengthHigh, lengthLow, lastStoredValues = 100, strict = true, resolution = null, selector = null)`. The strict parameter is placed before resolution to group indicator-specific configuration together (mirroring the constructor's parameter order), while keeping selector last to maintain API consistency with other indicator helper methods throughout the codebase. 

This ordering makes common usage patterns cleaner (e.g., `PPHL(SPY, 5, 5, strict: false)`) without needing to specify null values for framework parameters. Thoughts on this approach?

jhonabreul: This is better, but still a breaking change.
If you are currently using the `PPHL` helper method with positional arguments only, it will break for you. These calls will no longer compile:

```csharp
algorithm.PPHL(symbol, 3, 3, 100, Resolution.Minute);
algorithm.PPHL(symbol, 3, 3, 100, Resolution.Minute, (x) => x);
```

I think you might need to keep the original method and add two new overloads to keep it backwards compatible:

```csharp
// Original
public PivotPointsHighLow PPHL(Symbol symbol, 
    int lengthHigh, 
    int lengthLow, 
    int lastStoredValues = 100, 
    Resolution? resolution = null, 
    Func<IBaseData, IBaseDataBar> selector = null)

// New 1
public PivotPointsHighLow PPHL(Symbol symbol, 
    int lengthHigh, 
    int lengthLow, 
    int lastStoredValues, 
    bool strict, 
    Resolution? resolution = null, 
    Func<IBaseData, IBaseDataBar> selector = null)

// New 2
public PivotPointsHighLow PPHL(Symbol symbol, 
    int lengthHigh, 
    int lengthLow, 
    bool strict, 
    Resolution? resolution = null, 
    Func<IBaseData, IBaseDataBar> selector = null)
```

jhonabreul: If there is a chance to add only one new overload and keep things backwards compatible, it would be better!

jhonabreul: I think this could work. might test to be 100% sure

```csharp
// Original modified
public PivotPointsHighLow PPHL(Symbol symbol, 
    int lengthHigh, 
    int lengthLow, 
    int lastStoredValues, 
    Resolution? resolution, 
    Func<IBaseData, IBaseDataBar> selector = null)

// New
public PivotPointsHighLow PPHL(Symbol symbol, 
    int lengthHigh, 
    int lengthLow, 
    int lastStoredValues = 100, 
    bool strict = false, 
    Resolution? resolution = null, 
    Func<IBaseData, IBaseDataBar> selector = null)
```

jhonabreul (approved):
Thank you for your contribution!

PR #9102, opened 2025-11-26
Fix inclusive loop upper bound in HurstExponent

### LINKED ISSUE #9037: HurstExponent.cs has off by one error when creating the list of time lags.
https://github.com/QuantConnect/Lean/blob/81b2d45440dd4b496d5d5f59ff86a46a600abce1/Indicators/HurstExponent.cs#L66

Two issues.

1) The maxLag argument should be required to be at least 3. Since linear regression with only a single data point is meaningless.
2) There is an off-by-one error in the code. The for loop condition should be ->

`i <= maxLag`

JosueNina:
#### Description
Fixes an off-by-one issue to ensure the intended days are fully captured (e.g., a 20-day trading month).
Although the difference in value is minimal, after reviewing the logic it should indeed include the upper bound. For example, 20 is commonly used due to the 20 trading days in a month, so the iteration should include 20.

#### Related Issue
Closes #9037 

#### Motivation and Context
N/A

N/A

It was tested using unit tests.

#### Types of changes

#### Checklist:


JosueNina:
Here is the updated script:

**Source**: https://towardsdatascience.com/introduction-to-the-hurst-exponent-with-code-in-python-4da0414ca52e
```
This is the updated script:
import talib
import pandas as pd
import numpy as np 

history = pd.read_csv("https://github.com/QuantConnect/Lean/raw/master/Data/equity/usa/daily/spy.zip",
                       index_col=0, names=["open", "high", "low", "close", "volume"])

close = history.close

def get_hurst_exponent(time_series, max_lag=20):
    lags = range(2, max_lag + 1)
    tau = [np.std(np.subtract(time_series.shift(-lag), time_series)) for lag in lags]
    reg = np.polyfit(np.log(lags), np.log(tau), 1)
    return reg[0]

# we use a rolling window of past 252 data points
hurst_exponent = close * 0
for i in range(251, len(close)):
    hurst_exponent[i] = get_hurst_exponent(close.iloc[i-251:i + 1])
    
    
hurst_exponent = hurst_exponent.replace(0, np.nan).to_frame(name="hurst_exponent")
hurst_exponent["close"] = close
hurst_exponent = hurst_exponent.iloc[:, [1, 0]]
hurst_exponent.to_csv("spy_hurst_exponent.csv", index_label="date", header=True)
```

jhonabreul (commented):
Thank you!

Issue #8356, opened 2024-10-03
Implements New Highs New Lows Indicator


AlexCatarino:
#### Expected Behavior
Support NH-NL Indicator:
> The New Highs - New Lows indicator (NH-NL) displays the daily difference between the number of stocks reaching new 52-week highs and the number of stocks reaching new 52-week lows. The NH-NL indicator generally reaches its extreme lows slightly before a major market bottom.

#### Actual Behavior
Not supported

#### Potential Solution
This indicator follows closely the ADR architecture.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #9109, opened 2025-12-04
#8356 Implemention of NHNL indicators

nemcek:
#### Description
Implemented the base NHNL indicator as well as Difference, Ratio & Volume Ration NHNL indicators.
The architecture closely follows architecture of ADR indicators.

#### Related Issue
This PR implements new indicators as in https://github.com/QuantConnect/Lean/issues/8356

#### Motivation and Context
This change adds new indicators.

Probably adding the new indicators into the list of available indicators.

Added unit tests for each new indicator.
Tests closely resemble the tests of ADR indicators.
Also created a test dava (.csv) based on the test data of ADR indicators.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Indicators/NewHighsNewLowsDifferenceTests.cs
+++ b/Tests/Indicators/NewHighsNewLowsDifferenceTests.cs
@@ -17,46 +17,22 @@ using NUnit.Framework;
 using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
 using System;
-using System.Collections.Generic;
-using System.Linq;
 
 namespace QuantConnect.Tests.Indicators
 {
     [TestFixture]
     public class NewHighsNewLowsDifferenceTests : NewHighsNewLowsTestsBase<IBaseDataBar>
     {
-        protected override IndicatorBase<IBaseDataBar> CreateIndicator()
+        protected override NewHighsNewLows<IBaseDataBar> CreateNewHighsNewLowsIndicator()
         {
             // For test purposes we use period of two
-            var nhnlDifference = new NewHighsNewLows("test_name", 2);
-            if (SymbolList.Count > 2)
-            {
-                SymbolList.Take(3).ToList().ForEach(nhnlDifference.Add);
-            }
-            else
-            {
-                nhnlDifference.Add(Symbols.AAPL);
-                nhnlDifference.Add(Symbols.IBM);
-                nhnlDifference.Add(Symbols.GOOG);
-                RenkoBarSize = 5000000;
-            }
-
-            // Even if the indicator is ready, there may be zero values
-            ValueCanBeZero = true;
-
-            return nhnlDifference;
+            return new NewHighsNewLows("test_name", 2);
         }
 
-        protected override List<Symbol> GetSymbols()
+        protected override IndicatorBase<IBaseDataBar> GetSubIndicator(IndicatorBase<IBaseDataBar> mainIndicator)
         {
-            return [Symbols.SPY, Symbols.AAPL, Symbols.IBM];
+            return (mainIndicator as NewHighsNewLows).Difference;
         }
-
-        protected override Action<IndicatorBase<IBaseDataBar>, double> Assertion => (indicator, expected) =>
-        {
-            // we need to use the Difference sub-indicator
-            base.Assertion((indicator as NewHighsNewLows).Difference, expected);
-        };
 
         [Test]
         public virtual void ShouldIgnoreRemovedStocks()
--- a/Tests/Indicators/NewHighsNewLowsRatioTests.cs
+++ b/Tests/Indicators/NewHighsNewLowsRatioTests.cs
@@ -17,46 +17,23 @@ using NUnit.Framework;
 using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
 using System;
-using System.Collections.Generic;
-using System.Linq;
 
 namespace QuantConnect.Tests.Indicators
 {
     [TestFixture]
     public class NewHighsNewLowsRatioTests : NewHighsNewLowsTestsBase<IBaseDataBar>
     {
-        protected override IndicatorBase<IBaseDataBar> CreateIndicator()
+        protected override NewHighsNewLows<IBaseDataBar> CreateNewHighsNewLowsIndicator()
         {
             // For test purposes we use period of two
-            var nhnlRatio = new NewHighsNewLows("test_name", 2);
-            if (SymbolList.Count > 2)
-            {
-                SymbolList.Take(3).ToList().ForEach(nhnlRatio.Add);
-            }
-            else
-            {
-                nhnlRatio.Add(Symbols.AAPL);
-                nhnlRatio.Add(Symbols.IBM);
-                nhnlRatio.Add(Symbols.GOOG);
-                RenkoBarSize = 5000000;
-            }
-
-            // Even if the indicator is ready, there may be zero values
-            ValueCanBeZero = true;
-
-            return nhnlRatio;
+            return new NewHighsNewLows("test_name", 2);
         }
 
-        protected override List<Symbol> GetSymbols()
-        {
-            return [Symbols.SPY, Symbols.AAPL, Symbols.IBM];
-        }
-
-        protected override Action<IndicatorBase<IBaseDataBar>, double> Assertion => (indicator, expected) =>
+        protected override IndicatorBase<IBaseDataBar> GetSubIndicator(IndicatorBase<IBaseDataBar> mainIndicator)
         {
             // we need to use the Ratio sub-indicator
-            base.Assertion(((NewHighsNewLows)indicator).Ratio, expected);
-        };
+            return (mainIndicator as NewHighsNewLows).Ratio;
+        }
 
         [Test]
         public void ShouldIgnoreRemovedStocks()
--- a/Tests/Indicators/NewHighsNewLowsTestsBase.cs
+++ b/Tests/Indicators/NewHighsNewLowsTestsBase.cs
@@ -17,7 +17,9 @@ using NUnit.Framework;
 using QuantConnect.Data.Consolidators;
 using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using static QuantConnect.Tests.Indicators.TestHelper;
 
 namespace QuantConnect.Tests.Indicators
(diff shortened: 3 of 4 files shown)


Indicators/NewHighsNewLowsIndicator.cs
```diff
@@ -0,0 +1,235 @@
...
+    /// <summary>
+    /// The New Highs - New Lows indicator displays the daily difference between 
+    /// the number of stocks reaching new highs and the number of stocks reaching new lows
+    /// in defined time period.
+    /// </summary>
+    public abstract class NewHighsNewLowsIndicator : TradeBarIndicator, IIndicatorWarmUpPeriodProvider
+    {
+        private readonly IDictionary<SecurityIdentifier, TradeBar> _currentPeriod = new Dictionary<SecurityIdentifier, TradeBar>();
+        private readonly IDictionary<SecurityIdentifier, Maximum> _rollingPreviousHighs = new Dictionary<SecurityIdentifier, Maximum>();
+        private readonly IDictionary<SecurityIdentifier, Minimum> _rollingPreviousLows = new Dictionary<SecurityIdentifier, Minimum>();
```
Martin-Molinero: I think instead of having 3 collections per SID, could just have 1 collection with a small private class holdings these 3 attributes, will make the implementation cleaner I think and runtime performance much better

Indicators/NewHighsNewLowsIndicator.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Linq;
+
+namespace QuantConnect.Indicators
+{
+    /// <summary>
+    /// The New Highs - New Lows indicator displays the daily difference between 
+    /// the number of stocks reaching new highs and the number of stocks reaching new lows
+    /// in defined time period.
+    /// </summary>
+    public abstract class NewHighsNewLowsIndicator : TradeBarIndicator, IIndicatorWarmUpPeriodProvider
```
Martin-Molinero: Generally I think 'Difference' and 'Ratio' indicators should not really exist but just be a property of this base indicator, see other indicators like `BollingerBands/AverageDirectionalIndex/etc` exposing some state through indicatorbase.
Another thought is that ideally this base indicator does not require a TradeBar (which has volume) but requires a Bar, so it would work with QuoteBars, see examples `IBaseDataBar/BarIndicator/etc`. I would keep a second indicator inheriting from this one which adds volume ratio and enforces tradebar, might need to use some T type handling I think

Minor should remove Indicator from names like `NewHighsNewLowsIndicator`

Indicators/NewHighsNewLowsIndicator.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            List<TradeBar> newHighStocks = new List<TradeBar>();
+            List<TradeBar> newLowStocks = new List<TradeBar>();
```
Martin-Molinero: Generally shouldn't create collections unless required, same comment in the `foreach` above, there's not need to `ToList`, thinking of performance should be as lazy as possible

Indicators/NewHighsNewLowsIndicator.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override IndicatorResult ValidateAndComputeNextValue(TradeBar input)
+        {
+            Enqueue(input);
+
+            if (HasMissingCurrentPeriodValue() || !HasSufficientPreviousDataForComputation())
```
Martin-Molinero: Another thought I think ideally even if not all symbols are ready we could calculate the indicator values with those that are? might even simplify the logic, the indicator state wouldn't be ready yet but have valid value at least

Indicators/NewHighsNewLowsIndicator.cs
```diff
@@ -0,0 +1,235 @@
...
+            {
+                return new IndicatorResult(0, IndicatorStatus.ValueNotReady);
+            }
+
+            decimal vNext = ComputeNextValue(input);
+            return new IndicatorResult(vNext);
+        }
+
+        private void Enqueue(TradeBar input)
```
Martin-Molinero: Ah I see the indicator is very similar to `AdvanceDeclineIndicator` generally, seems there is some room to inherit and merge maybe?

Indicators/NewHighsNewLowsIndicator.cs
nemcek: Refactored into a class.
I wasn't sure where to put that class (what standard you guys follow, because I wasn't able to find any indicator which would work with some helper classes), so I made that class as a class inside the indicator.
If I should move it somewhere else, let me know.

Indicators/NewHighsNewLowsIndicator.cs
nemcek: `Difference` and `Ratio` are now _sub_ indicators on NHNL indicator.
I also included a `NewHighsNewLows` indicator class just as a typedef so one wouldn't need to type `NewHighsNewLows<IBaseDataBar>` everytime. If this class should be moved, let me know.

It now works with `Bar` and does not enforce `TradeBar`.
New `NewHighsNewLowsVolume` indicator was added which enforces `TradeBar` and has `VolumnRatio` _sub_ indicator.

Indicators/NewHighsNewLowsIndicator.cs
nemcek: Yeah, the `.ToList()` was oversight on my side, thanks. Fixed.
Also the usage of collections is now refactored.

Indicators/NewHighsNewLowsIndicator.cs
nemcek: Honestly I'm not sure about this.
The task was that this new `NHNL` indicator should closely follow the `ADR` indicator and therefore the tests are _basically_ the same for `NHNL` and `ADR` to be sure they work the same way.

If we change this, then `ADR` and `NHNL` would **not** work the same way.
However, if you think it's ok and `NHNL` should in fact work differently in this regard, then let me know and I can change it.

Indicators/NewHighsNewLowsIndicator.cs
nemcek: Yeah, I was actually thinking about that same thing during implementation as `ADR` is basically `NHNL` with period of `1` and a different calculation.
We _could_ definitely refactor it so the `ADR` is a kind of `NHNL` indicator.
However, I believed it was out of the scope of the task and therefore didn't proceed with it. If you think this is required, then let me know and we can work out what to do with it.

Indicators/NewHighsNewLowsIndicator.cs
nemcek: Oh, and also even tho `Difference` and `Ratio` are _sub_ indicators and since the indicator still needs to return some computed value, I decided to return the value of `Difference`. If you want it to return something else, let me know.

Tests/Indicators/NewHighsNewLowsDifferenceTests.cs
```diff
@@ -0,0 +1,269 @@
...
+*/
+
+using NUnit.Framework;
+using QuantConnect.Data.Market;
+using QuantConnect.Indicators;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace QuantConnect.Tests.Indicators
```
Martin-Molinero, line 21: These test files need some love to reduce duplication, maybe some inheritance magic or just merging them?

Indicators/NewHighsNewLows.cs
```diff
@@ -0,0 +1,340 @@
...
+            if (_trackedAssets.TryGetAsset(input.Symbol, out trackedAsset)
+                && (!_currentPeriodTime.HasValue || input.Time == _currentPeriodTime))
+            {
+                trackedAsset.CurrentPeriodValue = input;
+
+                if (!_currentPeriodTime.HasValue)
+                {
+                    _currentPeriodTime = input.Time;
+                }
+            }
```
Martin-Molinero, line 226: This initial if seems to be ignoring the initial data point sent in, not updating the trackers min/max -> seems like this if can be remove just setting an initial _currentPeriodTime of min datetime, avoiding the special handling maybe?

edit: oh I see it's being used when the time passed, why not use it right away?

Indicators/NewHighsNewLows.cs
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        private bool HasMissingCurrentPeriodValue()
```
Martin-Molinero, line 229: This method can probably be removed too, should be enough checking if min/max are ready?

Indicators/NewHighsNewLows.cs
```diff
@@ -0,0 +1,340 @@
...
+                asset.RollingPreviousHigh.IsReady
+                && asset.RollingPreviousLow.IsReady);
+        }
+
+        private void UpdatePreviousValues(TrackedAsset asset, IBaseDataBar bar)
+        {
+            if (bar == null)
+            {
+                return;
+            }
```
Martin-Molinero, line 246: We don't expect Bar to ever be null, shouldn't require this check I think

Indicators/NewHighsNewLows.cs
```diff
@@ -0,0 +1,340 @@
...
+            // In case of unordered input
+            if (input.EndTime == _currentPeriodTime)
+            {
+                // Update the previous values in rolling window
+                if (_trackedAssets.TryGetAsset(input.Symbol, out trackedAsset))
+                {
+                    UpdatePreviousValues(trackedAsset, input);
+                }
+                return;
+            }
```
Martin-Molinero, line 202: Don't think the data should arrive out of order either, have u seen this behavior?

I see a mix of .Time vs .EndTime usage, should generally just use .EndTime

Indicators/NewHighsNewLowsVolume.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(TradeBar input)
+        {
+            decimal nextValue = base.ComputeNextValue(input);
```
Martin-Molinero: minor but we use `var` when possible, there are a few locations where it can be used in the PR instead 👍

Indicators/NewHighsNewLows.cs
```diff
@@ -0,0 +1,340 @@
...
+
+            // In case of new period
+            if (input.Time > _currentPeriodTime)
+            {
+                foreach (TrackedAsset asset in _trackedAssets)
+                {
+                    // We got new period and therefore we need to update
+                    // the previous value in rolling window so we can use them for calculation
+                    UpdatePreviousValues(asset, asset.CurrentPeriodValue);
+                    asset.CurrentPeriodValue = default;
```
Martin-Molinero, line 212: hm because it's set to null here, the following call to HasMissingCurrentPeriodValue would make the indicator not calculate?

Martin-Molinero (changes requested):
Hey @nemcek! Thank you for addressing review, it's looking better, I think still there's some consolidation and cleanup that can be done to avoid any unwanted issues

Indicators/NewHighsNewLowsIndicator.cs
nemcek: I was thinking about this and I think it wouldn't work.
To calculate `NHNL` you need data for each symbol in the period, because you're calculating how many of those reached their new high vs how many of those reached their new low. Therefore it wouldn't make sense to calculate it on same state with missing data for some symbols in the period.

Indicators/NewHighsNewLows.cs
nemcek: Do you mean why min/max is not updated right away, but only after time has passed (on new period)? 
That's because the algorithm works this way. Let's say we're tracking only one asset `IBM`, we care only about `High` and have window of size `2`:
1. In period 1, we get `High = 1`. State: `Window(empty), Max = 0`, `CurrentPeriodValue="High=1"`
2. In period 2, we get `High = 2`. Now we push it into rolling window. State: `Window(1), Max = 1`,  `CurrentPeriodValue="High=2"`
3. In period 3, we get `High = 3`, Now we push it into rolling window. State: `Window(1; 2), Max = 2`,  `CurrentPeriodValue="High=3"`
  a. now we can calculate `NHNL` by comparing `Window Max (2)` with `CurrentPeriodValue High (3)`
  b. we get that asset **reached** a new high

Now if we would update min/max right away, considering the same exaple, it would look like this:
1. In period 1, we get `High = 1`. We're already pushing it into rolling window. State: `Window(1), Max = 1`, `CurrentPeriodValue="High=1"`
2. In period 2, we get `High = 2`.  State: `Window(1; 2), Max = 2`,  `CurrentPeriodValue="High=2"`
3. In period 3, we get `High = 3`, State: `Window(1; 2; 3), Max = 3`,  `CurrentPeriodValue="High=3"`
  a. if we would now try to calculate `NHNL` by comparing `Window Max (3)` with `CurrentPeriodValue High (3)` we would get that the asses **didn't  reach** new high

Indicators/NewHighsNewLows.cs
nemcek: It is not enough to check only if min/max are ready.

This is used as a guard before proceeding to value computation. See `ValidateAndComputeNextValue`
```c#
            Enqueue(input);

            if (HasMissingCurrentPeriodValue() || !HasSufficientPreviousDataForComputation())
            {
                return new IndicatorResult(0, IndicatorStatus.ValueNotReady);
            }

            decimal vNext = ComputeNextValue(input);
            return new IndicatorResult(vNext);
```

To calculate `NHNL` you need data for each symbol in the period, because you're calculating how many of those reached their new high vs how many of those reached their new low. 
If we only checked for min/max readiness (`HasSufficientPreviousDataForComputation`) then when min/max are ready and we get input data for the first asset for new period, we would proceed to value computation right away with missing input data for all other assets.
Therefore it wouldn't make sense to calculate it on a state with missing data for some assets in the new period.

Indicators/NewHighsNewLows.cs
nemcek: Parameter `Bar` can be `null`
See `UpdatePreviousValues(asset, asset.CurrentPeriodValue);` where `asset.CurrentPeriodValue` can be `null`

This is for a case when a period is missing an asset and therefore the period should be ignored.
This is how `ADR` works (see `IgnorePeriodIfAnyStockMissed` in e.g. `AdvanceDeclineRatioTests.cs`) and therefore that's how `NHNL` works (see `IgnorePeriodIfAnyStockMissed` in e.g. `NewHighsNewLowsRatioTests.cs`)

Indicators/NewHighsNewLows.cs
nemcek: It seems it actually can arrive out of order, because `ADR` indicator handles this use-case explicitely.
There's this test `WarmsUpProperly` in e.g. `AdvanceDeclineRatioTests.cs`.
Therefore `NHNL` handles this use-case too, see `WarmsUpProperly` in e.g. `NewHighsNewLowsRatioTests.cs`.

Indicators/NewHighsNewLows.cs
nemcek: That's by design. 
Here we push values into the min/max rolling windows based on current period value, because new period is starting and therefore we set current period value to `null`.
The current period value is then updated further down.

Tests/Indicators/NewHighsNewLowsDifferenceTests.cs
nemcek: I moved some _shared_ logic into the test base class.
I think this should be enough. I wouldn't go into trying to pull the specific tests up into the base class as tests for each _sub_ indicator uses different setup values and assert values.

Indicators/NewHighsNewLowsVolume.cs
nemcek: changed the use of concrete types into `var`

PR #9188, opened 2026-01-05
Refactor PythonIndicator to rely on compute_next_value

### LINKED ISSUE #9085: PythonIndicator Updated is not Called on Manual Updates
#### Expected Behavior
When we set `updated` to a callback method, and update the indicator with `update`, the callback method is called as it happens to LEAN/C# indicators.

#### Actual Behavior
When we set `updated` to a callback method, and update the indicator with `update`, the callback method is not called. If we use self.register_indicator, it is.

#### Potential Solution
PythonIndicator should require implementing compute_next_value, and when update is called, it will call compute_next_value as in C#, and triggering the updated event.

#### Reproducing the Problem
`_on_update` method isn't called:

```python
from collections import deque 

class CustomIndicatorsAlgorithm(QCAlgorithm):
    def initialize(self) -> None:
        self.set_start_date(2024, 9, 1)
        self.set_end_date(2024, 12, 31)
        self.spy = self.add_equity("SPY").symbol
        self.custom_mfi = CustomMoneyFlowIndex(20)
        self.custom_mfi.updated += self._on_update

    def _on_update(self, sender, consolidated):
        self.log(str(consolidated))

    def on_data(self, slice: Slice) -> None:
        bar = slice.bars.get(self.spy)
        if bar:
            self.custom_mfi.update(bar)

class CustomMoneyFlowIndex(PythonIndicator):
    def __init__(self, period: int) -> None:
        super().__init__()
        self.value = 0
        self.previous_typical_price = 0
        self.negative_money_flow = deque(maxlen=period)
        self.positive_money_flow = deque(maxlen=period)
    
    def update(self, input: BaseData) -> bool:
        typical_price = (input.high + input.low + input.close) / 3
        money_flow = typical_price * input.volume
            
        if abs(self.previous_typical_price / typical_price - 1) < 1e-10:
            self.previous_typical_price = typical_price
        
        self.negative_money_flow.appendleft(money_flow if typical_price < self.previous_typical_price else 0)
        self.positive_money_flow.appendleft(money_flow if typical_price > self.previous_typical_price else 0)
        self.previous_typical_price = typical_price
    
        positive_money_flow_sum = sum(self.positive_money_flow)        
        total_money_flow = positive_money_flow_sum + sum(self.negative_money_flow)

        self.value = 100
        if total_money_flow != 0:
            self.value *= positive_money_flow_sum / total_money_flow    
        return len(self.positive_money_flow) == self.positive_money_flow.maxlen
```

#### Checklist

JosueNina:
#### Description
Adds a new default initialization path for custom Python indicators, requiring overriding `compute_next_value` instead of `update` and passing the instance to the base class, while preserving backward compatibility.

#### Related Issue
Closes #9085 

#### Motivation and Context
N/A

Yes.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/PythonIndicator.cs
+++ b/Indicators/PythonIndicator.cs
@@ -57,7 +57,7 @@ namespace QuantConnect.Indicators
         public PythonIndicator(PyObject indicator)
             : base(GetIndicatorName(indicator))
         {
-            InitializeIndicator(indicator);
+            SetIndicator(indicator);
         }
 
         /// <summary>
@@ -65,11 +65,6 @@ namespace QuantConnect.Indicators
         /// </summary>
         /// <param name="indicator">The python implementation of <see cref="IndicatorBase{IBaseDataBar}"/></param>
         public void SetIndicator(PyObject indicator)
-        {
-            InitializeIndicator(indicator);
-        }
-
-        private void InitializeIndicator(PyObject indicator)
         {
             _instance = indicator;
             _indicatorWrapper = new BasePythonWrapper<IIndicator>(indicator, validateInterface: false);
@@ -88,11 +83,6 @@ namespace QuantConnect.Indicators
                 return;
             }
 
-            ValidateRequiredAttributes(indicator);
-        }
-
-        private void ValidateRequiredAttributes(PyObject indicator)
-        {
             var requiredAttributes = new[] { "IsReady", "Update", "Value" };
             foreach (var attributeName in requiredAttributes)
             {


AlexCatarino:
Perhaps you should remove the examples using `update` so users don't use it. I plan on flagging it in the Docs as legacy and recommend this approach.

Martin-Molinero:
See https://github.com/QuantConnect/Lean/issues/9085#issuecomment-3719279424

PR #9371, opened 2026-04-06
Extend LeastSquaresMovingAverage to accept a benchmark

### LINKED ISSUE #6984: Extend LeastSquaresMovingAverage to Accept a Benchmark
#### Expected Behavior
Lean supports LeastSquaresMovingAverage as a Lean Indicator which takes a Symbol object, a benchmark Symbol, and a look-back period. 

#### Actual Behavior
The current implementation assumes a constant slope: [LeastSquaresMovingAverage.cs#L59](https://github.com/QuantConnect/Lean/blob/master/Indicators/LeastSquaresMovingAverage.cs#L59). See #83.

#### Potential Solution
Implement this indicator. 

#### Checklist

0xpinara:
## Description
- Added new constructors to `LeastSquaresMovingAverage` that accept a `Symbol referenceSymbol` parameter
- When a reference is provided, OLS regression uses the reference values as X instead of time indices, and projects using the latest reference value
- Added a new `LSMA(symbol, reference, period, resolution, selector)` overload in `QCAlgorithm.Indicators.cs`
- Follows the `Alpha` indicator pattern for dual-symbol data routing and registration

## Related Issue
Resolves #6984

## Motivation and Context
Previously, LSMA only regressed price against time. Users who wanted to regress against a benchmark (e.g. SPY) had no built-in way to do so. This was specifically requested by a maintainer in the issue comments.

## Requires Documentation Change
Yes — the new LSMA overload should be documented.

## How Has This Been Tested?
- All existing LSMA tests continue to pass
- `WithReferenceRegressesAgainstBenchmark` — feeds a perfect y=2x+1 relationship, verifies slope=2, intercept=1, projected value=11
- `WithReferenceIsNotReadyUntilBothWindowsFull` — verifies IsReady requires both target and reference windows
- `WithReferenceResetsProperly` — verifies full state reset
- `WithoutReferenceBehavesIdentically` — confirms no regression in original behavior

## Types of changes

## Checklist
- [x] My code follows the code style of this project
- [x] I have read the CONTRIBUTING document
- [x] I have added tests to cover my changes
- [x] All new and existing unit tests pass
- [x] My branch follows the naming convention: `feature-6984-lsma-benchmark-reference`

### DIFF
--- a/Indicators/LeastSquaresMovingAverage.cs
+++ b/Indicators/LeastSquaresMovingAverage.cs
@@ -24,6 +24,8 @@ namespace QuantConnect.Indicators
     /// The Least Squares Moving Average (LSMA) first calculates a least squares regression line
     /// over the preceding time periods, and then projects it forward to the current period. In
     /// essence, it calculates what the value would be if the regression line continued.
+    /// When a reference symbol is provided, the regression is performed against the reference
+    /// values instead of time.
     /// Source: https://rtmath.net/assets/docs/finanalysis/html/b3fab79c-f4b2-40fb-8709-fdba43cdb363.htm
     /// </summary>
     public class LeastSquaresMovingAverage : WindowIndicator<IndicatorDataPoint>, IIndicatorWarmUpPeriodProvider
@@ -32,6 +34,16 @@ namespace QuantConnect.Indicators
         /// Array representing the time.
         /// </summary>
         private readonly double[] _t;
+
+        /// <summary>
+        /// The reference symbol to regress against.
+        /// </summary>
+        private readonly Symbol _referenceSymbol = Symbol.None;
+
+        /// <summary>
+        /// Rolling window of reference symbol data points.
+        /// </summary>
+        private readonly RollingWindow<IndicatorDataPoint> _referenceWindow = new(0);
 
         /// <summary>
         /// The point where the regression line crosses the y-axis (price-axis)
@@ -47,6 +59,11 @@ namespace QuantConnect.Indicators
         /// Required period, in data points, for the indicator to be ready and fully initialized.
         /// </summary>
         public int WarmUpPeriod => Period;
+
+        /// <summary>
+        /// Gets a flag indicating when this indicator is ready and fully initialized
+        /// </summary>
+        public override bool IsReady => base.IsReady && _referenceWindow.IsReady;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LeastSquaresMovingAverage"/> class.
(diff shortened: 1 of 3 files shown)


Indicators/LeastSquaresMovingAverage.cs
```diff
@@ -33,6 +35,16 @@ public class LeastSquaresMovingAverage : WindowIndicator<IndicatorDataPoint>, II
         /// </summary>
         private readonly double[] _t;
 
+        /// <summary>
+        /// The reference symbol to regress against, if provided.
+        /// </summary>
+        private readonly Symbol _referenceSymbol;
```
AlexCatarino: If the default _referenceSymbol is Symbol.None, you can remove the null checks. See line 122.

Indicators/LeastSquaresMovingAverage.cs
```diff
@@ -33,6 +35,16 @@ public class LeastSquaresMovingAverage : WindowIndicator<IndicatorDataPoint>, II
...
+
+        /// <summary>
+        /// Rolling window of reference symbol data points.
+        /// </summary>
+        private readonly RollingWindow<IndicatorDataPoint> _referenceWindow;
```
AlexCatarino: If the default _referenceWindow is new(0),  _referenceWindow.IsReady is always true (see IsReady)

Indicators/LeastSquaresMovingAverage.cs
```diff
@@ -88,11 +146,31 @@ protected override decimal ComputeNextValue(IReadOnlyWindow<IndicatorDataPoint>
                 .OrderBy(i => i.EndTime)
                 .Select(i => Convert.ToDouble(i.Value))
                 .ToArray();
+
+            double[] xValues;
+            if (_referenceSymbol != null && _referenceWindow.IsReady)
```
AlexCatarino: If default _referenceWindow is new(0), we can use `_referenceWindow.Size != 0 && _referenceWindow.IsReady;` and also use only one if-block.

```csharp
decimal x = Period;
double intercept, slope;
 if (_referenceWindow.Size != 0 && _referenceWindow.IsReady)
{
    var xValues = _referenceWindow
        .OrderBy(i => i.EndTime)
        .Select(i => Convert.ToDouble(i.Value))
        .ToArray();
    x = _referenceWindow[0].Value;
    (intercept, slope) = Fit.Line(x: xValues , y: series);
}
else
{
    (intercept, slope) = Fit.Line(x: _t, y: series);
}
Intercept.Update(input.EndTime, intercept.SafeDecimalCast());
Slope.Update(input.EndTime, slope.SafeDecimalCast());

// Calculate the fitted value corresponding to the input
return Intercept.Current.Value + Slope.Current.Value * x;
```

AlexCatarino (changes requested):
Thank you, @0xpinara.

Please do not add multiple commits from other pull requests. Rebase with master to include only relevant changes.
I've left a few comments to improve the implementation.

0xpinara:
Hi @AlexCatarino, thanks for the feedback. I've pushed the fixes:

- Changed _referenceSymbol default to Symbol.None and _referenceWindow to new(0), all null checks are gone now
- Simplified ComputeNextValue to a single if-block using _referenceWindow.Size != 0 && _referenceWindow.IsReady with tuple deconstruction and SafeDecimalCast() like you asked
- Rebased to one commit with only the LSMA changes, sorry about that. 

Let me know if anything else needs changing.

Indicators/LeastSquaresMovingAverage.cs
```diff
@@ -70,6 +87,47 @@ public LeastSquaresMovingAverage(int period)
...
+        /// <summary>
+        /// Computes the next value of this indicator from the given state
+        /// </summary>
+        /// <param name="input">The input given to the indicator</param>
+        /// <returns>A new value for this indicator</returns>
+        protected override decimal ComputeNextValue(IndicatorDataPoint input)
+        {
+            if (input.Symbol == _referenceSymbol)
+            {
+                _referenceWindow.Add(input);
```
Martin-Molinero, line 124: Don't think this will work correctly, if main symbol comes first and later reference ( should rely on data coming in order) it wont be used for the same time, this requires some time awareness, can see similar multi symbol indicators and how they work.
Also generally I think ideally we should compare the indicator values in a test against some external source? like we usually do, at least trying to find one
Minor also but shouldn't create an unrequired rolling window `_referenceWindow = new(0);`

Martin-Molinero (changes requested):
Hey! Sharing some concerns and requests 👍

Martin-Molinero:
Closing for now until review address, feel free to reopen 👍

0xpinara:
Hey @Martin-Molinero, thanks for the feedback! I'll work on fixing these when I get some time and reopen the PR.

0xpinara:
Sorry for the long delay on this. I couldn't reopen the PR so I opened #9761
instead. It uses DualSymbolIndicator now, so the ordering problem you pointed out
is solved.

PR #9380, opened 2026-04-09
feat: add optional adanos market sentiment custom data example

alexander-schneider:
## Summary

This PR adds an optional Python custom data example for consuming structured Adanos market sentiment snapshots in LEAN.

It includes:
- a new `Algorithm.Python/AdanosMarketSentimentAlgorithm.py` example
- a project entry in `QuantConnect.Algorithm.Python.csproj`
- focused `PythonDataTests` coverage for the new custom data type

## Why

LEAN already has strong support for custom data via `AddData<T>` / `PythonData`, and this change keeps the integration in that example-oriented path rather than introducing a new core data provider.

Adanos is used here as structured alternative data for equities:
- current buzz score
- bullish percentage
- sentiment score
- simple activity / coverage counts

## Design notes

- The example is strictly optional and parameter-driven.
- No requests are made unless `adanos-api-key` is provided.
- The example now exits early outside live mode, because querying a current REST endpoint in a historical backtest would introduce look-ahead bias. The intended historical workflow is to persist daily snapshots first and then replay them from a local file or ObjectStore.
- The core LEAN data feed and provider stack are untouched.

## Parameters

- `adanos-api-key` – required to enable requests
- `adanos-source` – one of `news`, `reddit`, `x`, `polymarket`
- `adanos-lookback-days` – lookback window passed through to the stock sentiment endpoint

## Validation

Added focused tests in `Tests/Python/PythonDataTests.cs` for:
- request header propagation
- structured payload parsing
- unsupported source fallback
- `found=false` null handling

### DIFF
--- a/Algorithm.Python/AdanosMarketSentimentAlgorithm.py
+++ b/Algorithm.Python/AdanosMarketSentimentAlgorithm.py
@@ -18,7 +18,8 @@ from AlgorithmImports import *
 ### <summary>
 ### Demonstrates consuming optional Adanos market sentiment snapshots as custom equity data.
 ### This example is intended for live or research workflows that query current structured sentiment
-### from Adanos. Historical backtests should first persist daily snapshots to a local file or ObjectStore.
+### from Adanos. Historical backtests should first persist daily snapshots to a local file or ObjectStore
+### to avoid look-ahead bias from querying a current REST endpoint.
 ### Parameters:
 ### - adanos-api-key: required to enable requests
 ### - adanos-source: one of news, reddit, x, polymarket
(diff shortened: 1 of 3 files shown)


AlexCatarino:
Hi @alexander-schneider,
Do you represent Adanos? If you do, contact us at support@quantconnect.com. This dataset is better suited to [/datasets](https://www.quantconnect.com/datasets) than to this repository. Otherwise, we could find a place for it in the Docs, or you could place a working example in [/strategies](https://www.quantconnect.com/strategies/).

alexander-schneider:
Thanks, wrote you an email :)

Issue #6987, opened 2023-02-20
Implenent Parabolic SAR - Extended (SAREXT)


AlexCatarino:
#### Expected Behavior
Lean support the Parabolic SAR - Extended (SAREXT) Indicator.

#### Actual Behavior
Not supported.

#### Potential Solution
Implement SAREXT. There is scarce information on this indicator. We could look into TALIB's formulas to understand the difference from PSAR.

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

LouisSzeto:
[spy_sarext.csv](https://github.com/QuantConnect/Lean/files/10812711/spy_sarext.csv)

nbkell:
I would like to be assigned this as a first issue. Thanks!

nbkell:
@AlexCatarino I’ve recently made a PR that implements SAREXT, which is simply a generalization of PSAR (called SAR in talib) that gives you more control over the parameters and behavior. The differences are:

- SAREXT is negative if you’re in a short position, whereas PSAR is always positive regardless of current position. 
- You can now specify different acceleration factor parameters (start, increment, and max) that used according to whether you’re in a short or long position.
- You can now specify a starting value of the indicator (where the sign of the starting value dictates whether the indicator starts in a long or short position). 
- An additional parameter called offsetOnReverse, which determines a percentage of the current SAR that is first incremented/decremented from the indicator before switching between long/short positions. 

 Comments/questions regarding testing:

- To create my test .txt I just used the applied SAREXT function from tallib to spy ticker data I obtained from yahoo finance. I’ve attached the ticker data file and Python script I used to make the test file below. 

- I’d be happy to add the file posted by @LouisSzeto, but I’m not sure how to incorporate it if doesn’t include values for low, high, and close?

Finally I want to point out that the current Lean implementation for PSAR is not entirely consistent with that of SAR in talib (see the link to ta_SAR.c below). These discrepancies are:

- talib does not consider the first bar when initializing the extreme point _ep.
- talib uses the DM_MINUS function to determine if indicator starts in long/short position (currently LEAN compares close direction of first and second bars). 
- In the samples == 2 case, ComputeNextValue() should still do the first iteration of long/short updates (instead of just returning _sar computed in init()).

[...]

PR #9388, opened 2026-04-12
Add Jurik Moving Average (JMA) indicator

### LINKED ISSUE #8338: could you help add JMA indicador?
it seems JMA behaves better than EMA in some situations. and pandas_ta already has JMA indicator(pandas_ta.jma).
here are the formula info:
https://br.tradingview.com/script/nZuBWW9j-Jurik-Moving-Average/
http://jurikres.com/catalog1/ms_ama.htm
https://c.mql5.com/forextsd/forum/164/jurik_1.pdf

image
image

claygeo:
## Summary

Adds the Jurik Moving Average (JMA) indicator, a three-stage adaptive filter that produces smoother output with less lag than EMA. Requested in #8338.

**Implementation:**
- `Indicators/JurikMovingAverage.cs` — extends `Indicator`, implements `IIndicatorWarmUpPeriodProvider`
- Parameters: `period` (lookback), `phase` (-100 to 100, lag vs overshoot tradeoff), `power` (smoothing aggressiveness, default 2)
- Three-stage filter: adaptive EMA → Kalman-style velocity estimation → error correction
- `Algorithm/QCAlgorithm.Indicators.cs` — adds `JMA()` helper method

**Algorithm note:** The original JMA algorithm is proprietary (Jurik Research). This implementation follows the widely-adopted community-standard reverse-engineered formula used by [pandas_ta](https://github.com/twopirllc/pandas-ta), [TradingView](https://br.tradingview.com/script/nZuBWW9j-Jurik-Moving-Average/), and [QuanTAlib](https://github.com/mihakralj/QuanTAlib).

**Tests:**
- `JurikMovingAverageTests` inherits `CommonIndicatorTests<IndicatorDataPoint>` (10+ automated tests)
- `spy_jma.txt` reference data validated against the same formula
- `JmaComputesCorrectly` inline test with hand-computed intermediate values for independent verification

Closes #8338


Martin-Molinero (changes requested):
Hey @claygeo! Welcome to Lean 💪 

Seems the implementation of the JMA is not storing/looking at previous price values -> period arg doesn't have any effect? Please take a look at how the over moving averages work
Regarding test data `Tests/TestData/spy_jma.txt`, please share how it was generated, we usually do a python snippet using talib, mind doing so? See for example https://github.com/QuantConnect/Lean/issues/6982#issuecomment-1777089341 or https://github.com/QuantConnect/Lean/issues/6987#issuecomment-1777044713 etc

claygeo:
Thanks for the review @Martin-Molinero!

**Volatility-adaptive algorithm:** You were right that the original implementation was too simplified — the `period` parameter only affected static smoothing constants. I've rewritten the JMA to use the full community-standard algorithm (matching [pandas_ta](https://github.com/xgboosted/pandas-ta-classic)):

- Added **volatility bands** (upper/lower) that adapt to price movement
- Added **rolling volatility tracking** (10-bar sum, 65-bar average window) using `RollingWindow<decimal>`
- The smoothing factor `alpha` now **adapts per-bar** based on relative volatility: `alpha = beta^(rVolty^pow1)`, where `rVolty` varies with market conditions
- The `period` parameter controls both the base smoothing constants AND the volatility band adaptation rate through derived constants (`length1`, `pow1`, `bet`)

Note: JMA is fundamentally a recursive filter (not a windowed average like SMA/KAMA), so it extends `Indicator` rather than `WindowIndicator`. The lookback is in the volatility tracking, not in a price window.

**Test data generation:** Added `Tests/TestData/generate_jma.py` which uses `pandas_ta_classic` as the reference:

```python
pip install pandas pandas-ta-classic
python generate_jma.py
```

The script reads SPY data from `spy_kama.txt` and computes `JMA(7, phase=0)`. Test expected values in `JurikMovingAverageTests.cs` are verified against the same reference.

Tests/TestData/generate_jma.py
```diff
@@ -0,0 +1,51 @@
...
+import numpy as np
+
+# Try pandas_ta_classic first, then pandas_ta
+try:
+    from pandas_ta_classic.overlap.jma import jma
+except ImportError:
+    from pandas_ta.overlap.jma import jma
+
+# Read SPY data from the existing KAMA test file (same price data)
+df = pd.read_csv("spy_kama.txt", parse_dates=["Date"])
```
Martin-Molinero: This is a strange dependency, please follow the pattern as seen in Louis comments https://github.com/QuantConnect/Lean/pull/9388#pullrequestreview-4107272973 , and please don't commit this file just post it as a comment  👍

Martin-Molinero (changes requested):
That last commit basically reimplemented the whole indicator 😅
Please confirm your there's a human behind the PR which tested it @claygeo , tests seem to be failing

claygeo:
Hey @Martin-Molinero, yep I'm here! You're right the original was too simplified — I went back to the pandas_ta reference and realized it was missing the entire volatility adaptation system, so it needed a proper rewrite. That's why the diff was large in one shot, sorry about that.

Follow-up commit addresses your feedback:
- Removed `generate_jma.py` from committed files (Python snippet posted below instead)
- Removed the `power` parameter since the reference algorithm doesn't use it
- Added `period` validation (`ArgumentOutOfRangeException` for period < 2)
- Added tests for `phase` affecting output and invalid period values

The algorithm now matches the community-standard JMA (pandas_ta, TradingView) with volatility-adaptive bands and per-bar alpha adjustment. Tested the output against pandas_ta and it matches to 15 decimal places.

---

**Test data generation script** (`pip install numpy pandas`):

```python
"""
JMA(7, phase=0) test data generation — self-contained, numpy/pandas only.
Implements the community-standard JMA algorithm (same as pandas_ta, TradingView).
"""
import numpy as np
import pandas as pd

history = pd.read_csv(
    "https://github.com/QuantConnect/Lean/raw/master/Data/equity/usa/daily/spy.zip",
    index_col=0, names=["open", "high", "low", "close", "volume"]
)
close = history["close"].astype(float).values

_length = 7
phase = 0.0

length = 0.5 * (_length - 1)
pr = 0.5 if phase < -100 else 2.5 if phase > 100 else 1.5 + phase * 0.01
length1 = max(np.log(np.sqrt(length)) / np.log(2.0) + 2.0, 0)
pow1 = max(length1 - 2.0, 0.5)
length2 = length1 * np.sqrt(length)
bet = length2 / (length2 + 1)
beta = 0.45 * (_length - 1) / (0.45 * (_length - 1) + 2.0)

m = len(close)
jma = np.full(m, np.nan)
volty = np.zeros(m)
v_sum = np.zeros(m)

jma[0] = ma1 = uBand = lBand = close[0]
det0 = det1 = 0.0
sum_length = 10

for i in range(1, m):
    price = close[i]
    del1 = price - uBand
    del2 = price - lBand
    volty[i] = max(abs(del1), abs(del2)) if abs(del1) != abs(del2) else 0
    v_sum[i] = v_sum[i-1] + (volty[i] - volty[max(i - sum_length, 0)]) / sum_length
    avg_volty = np.average(v_sum[max(i - 65, 0) : i + 1])
    d_volty = 0 if avg_volty == 0 else volty[i] / avg_volty
    r_volty = max(1.0, min(np.power(length1, 1/pow1), d_volty))
    pow2 = np.power(r_volty, pow1)
    kv = np.power(bet, np.sqrt(pow2))
    uBand = price if del1 > 0 else price - kv * del1
    lBand = price if del2 < 0 else price - kv * del2
    alpha = np.power(beta, pow2)
    ma1 = (1 - alpha) * price + alpha * ma1
    det0 = (price - ma1) * (1 - beta) + beta * det0
    ma2 = ma1 + pr * det0
    det1 = (ma2 - jma[i-1]) * (1 - alpha)**2 + alpha**2 * det1
    jma[i] = jma[i-1] + det1

jma[0 : _length - 1] = np.nan

output = history.copy()
output["JMA_7"] = [f"{x:.12f}" if not np.isnan(x) else "" for x in jma]
output.to_csv("spy_jma.csv", header=False)
```

PR #9411, opened 2026-04-17
Use order price for crypto future initial margin checks

### LINKED ISSUE #7726: CryptoFutureMarginModel - GetInitialMarginRequirement doesn't consider order type
#### Expected Behavior
`GetInitialMarginRequirement` calculates the margin required to open a position and in cases of a limit order, it should calculate the position value based on the limit price instead of the security's current price.

#### Actual Behavior
`GetInitialMarginRequirement` calculates the margin based on the security's latest price, which leads to inaccurate buying power calculation, especially when the order type is a limit order where the limit price is much lower than the current price.  See here https://github.com/QuantConnect/Lean/blob/5611b5508ada70dda96cc2b917e9475400f781e3/Common/Securities/CryptoFuture/CryptoFutureMarginModel.cs#L77

#### Potential Solution
Maybe the `InitialMarginParameters` should take the orderticket as an input?https://github.com/QuantConnect/Lean/blob/6e2744af7cf46090bafb166be016ea350df2cb8f/Common/Securities/InitialMarginParameters.cs#L23-L33

#### Reproducing the Problem
N/A (can provide a failing limit order due to insufficient buying power due to this issue if needed, but the link to the implementation code explains the issue)

#### System Information
N/A

#### Checklist

aiSynergy37:
## Summary
- use order-derived price when computing crypto future order initial margin
- keep existing security-price-based `GetInitialMarginRequirement(InitialMarginParameters)` behavior for non-order paths
- add regression tests covering `BTCUSDT` and `BTCUSD` limit orders

## Why
Fixes incorrect buying power checks for crypto future limit orders where margin should be based on the executable order price context.

Closes #7726

## Testing
- Local environment is blocked by Windows Application Control when loading `QuantConnect.Tests.dll` (`0x800711C7`).
- Requested command to run on unrestricted runner:
  - `dotnet test Tests/QuantConnect.Tests.csproj --filter "FullyQualifiedName~CryptoFutureMarginModelTests.InitialMarginRequiredForOrderUsesLimitOrderPrice"`

### DIFF
--- a/Common/Securities/CryptoFuture/CryptoFutureMarginModel.cs
+++ b/Common/Securities/CryptoFuture/CryptoFutureMarginModel.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Linq;
 using QuantConnect.Orders;
+using QuantConnect.Orders.Fees;
 
 namespace QuantConnect.Securities.CryptoFuture
 {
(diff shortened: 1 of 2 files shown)


aiSynergy37:
Validation update:

To bypass the local Windows App Control restriction (`0x800711C7`), I ran the exact command on GitHub-hosted runners:

`dotnet test Tests/QuantConnect.Tests.csproj --filter "FullyQualifiedName~CryptoFutureMarginModelTests.InitialMarginRequiredForOrderUsesLimitOrderPrice"`

Runs:
- Ubuntu: https://github.com/aiSynergy37/Lean/actions/runs/24575309556
- Windows: https://github.com/aiSynergy37/Lean/actions/runs/24575623661

Both runs aborted before producing a test assertion result with the same host crash:
- `The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.InvalidOperationException: GIL must always be released, and it must be released from the same thread that acquired it.`
- `Test Run Aborted.`
- `Process completed with exit code 1.`

PR #9426, opened 2026-04-21
Extend LSMA to accept benchmark reference

aiSynergy37:
What: Extend `LeastSquaresMovingAverage` to regress a target symbol against a benchmark reference symbol.

Why: Issue #6984 requested an LSMA overload that accepts a benchmark reference instead of only regressing target values against a fixed time index.

How: Added reference-symbol constructors, a `QCAlgorithm.LSMA` overload, matched target/reference timestamp buffering, and reference-value regression when paired samples are ready.

Edge cases handled: Target-first and reference-first updates are supported; mismatched timestamps wait for a matching pair; reset clears pending target/reference state; unexpected target symbols are rejected.

Tests added/updated: Added LSMA reference-mode tests for target-first updates, reference-first updates, timestamp matching, and reset behavior. Build validation completed for `Tests/QuantConnect.Tests.csproj`; focused local test execution was blocked by the existing pythonnet GIL finalizer crash before test execution.

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -1366,6 +1366,25 @@ namespace QuantConnect.Algorithm
         }
 
         /// <summary>
+        /// Creates and registers a new Least Squares Moving Average instance.
+        /// </summary>
+        /// <param name="symbol">The symbol whose LSMA we seek.</param>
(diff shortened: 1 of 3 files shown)


Martin-Molinero:
Hey @aiSynergy37! Sorry for the delay here, replacing by https://github.com/QuantConnect/Lean/pull/9761 which is reusing existing multiple symbol indicator handling instead.

PR #9437, opened 2026-04-26
Fix indicator warm-up window seed

aiSynergy37:
What: Clear indicator windows after reset during indicator history warm-up.

Why: `IndicatorHistory` and `WarmUpIndicator` could leave the reset default value in an expanded public indicator window before replaying historical samples.

How: Reset the indicator window immediately after `indicator.Reset()` in the history replay path so only real historical samples populate the public window.

Edge cases handled: Expanded windows no longer retain `DateTime.MinValue` seed points; the current indicator value remains at `Window[0]` after warm-up.

Tests added/updated: Added `AlgorithmIndicatorsTests.WarmUpIndicatorDoesNotLeaveDefaultValueInWindow`. Build validation completed for `Algorithm/QuantConnect.Algorithm.csproj` and `Indicators/QuantConnect.Indicators.csproj`; focused local test execution was blocked by Windows Application Control on the test assembly.

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -4455,6 +4455,7 @@ namespace QuantConnect.Algorithm
         {
             // Reset the indicator
             indicator.Reset();
+            indicator.Window.Reset();
 
             var properties = indicator.GetType()
                 .GetProperties()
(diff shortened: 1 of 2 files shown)


Algorithm/QCAlgorithm.Indicators.cs
```diff
@@ -4455,6 +4455,7 @@ private IndicatorHistory IndicatorHistory<T>(IndicatorBase<T> indicator, IEnumer
         {
             // Reset the indicator
             indicator.Reset();
+            indicator.Window.Reset();
```
Martin-Molinero, line 4458: `indicator.Reset()` is calling `indicator.Window.Reset()`

PR #9457, opened 2026-05-05
Allow SetAccountCurrency after SetCash without throwing

AlexCatarino:
#### Description
Changes `SecurityPortfolioManager.SetAccountCurrency` so it no longer throws when called after `SetCash`. The base account currency is switched in place:

- **Different currency**: the previous `Cash` entry stays in the `CashBook` with its existing balance, the new account currency starts at zero (or `startingCash` if provided), and a trace is logged noting the residual balance held in the previous currency. This undoes the automatic amount migration performed by `CashBook.AccountCurrency`'s setter so balances stay in their original currency.
- **Same currency**: an optional `startingCash` overrides the previously set amount and a trace is logged stating the new amount and the prior amount. When `startingCash` is omitted or equals the existing amount, nothing is logged.
- **No prior `SetCash`**: behaves as before — only the existing "setting account currency to X" trace is emitted.

#### Related Issue
N/A

#### Motivation and Context
The previous behavior forced users to order `SetAccountCurrency` strictly before `SetCash`, throwing `InvalidOperationException` otherwise. Switching to a graceful in-place change with a clear log message is more forgiving and lets users override the starting amount when re-asserting the same account currency.

No.

Replaced the obsolete `CanNotChangeAccountCurrencyAfterSettingCash` test (which expected the exception) with four new tests in `SecurityPortfolioManagerTests`:
- `ChangeAccountCurrencyAfterSettingCashKeepsPreviousCash` (covers both `SetCash` overloads) — verifies the previous balance is retained in the `CashBook` after a currency change.
- `ChangeAccountCurrencyAfterSettingCashAppliesStartingCashToNewCurrency` — verifies `startingCash` applies to the new currency while the previous balance is preserved.
- `SetAccountCurrencyWithSameCurrencyOverridesStartingCash` — verifies a same-currency call with `startingCash` overwrites the prior amount.
- `SetAccountCurrencyWithSameCurrencyAndNoStartingCashKeepsExistingAmount` — verifies a no-op same-currency call leaves the existing amount untouched.

All 8 SetAccountCurrency-related tests pass locally.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Messages/Messages.Securities.cs
+++ b/Common/Messages/Messages.Securities.cs
@@ -890,13 +890,6 @@ namespace QuantConnect
                 "Cannot change AccountCurrency after adding a Security. Please move SetAccountCurrency() before AddSecurity().";
 
             /// <summary>
-            /// Returns a string message saying the AccountCurrency cannot be changed after setting cash and that the method
-            /// SetAccountCurrency() should be moved before SetCash()
-            /// </summary>
-            public static string CannotChangeAccountCurrencyAfterSettingCash =
-                "Cannot change AccountCurrency after setting cash. Please move SetAccountCurrency() before SetCash().";
-
-            /// <summary>
             /// Returns a string message saying the AccountCurrency has been changed after setting cash, reporting the
             /// remaining amount held in the previous account currency
             /// </summary>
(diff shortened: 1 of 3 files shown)


Martin-Molinero (approved):
Thank you!

PR #9497, opened 2026-05-23
Feature: add ReconnectBrokerage() to QCAlgorithm for self-healing data feeds

### LINKED ISSUE #9496: Title: [Feature Request] Expose brokerage reconnect to algorithm layer
#### Expected Behavior

Algorithms should be able to programmatically trigger a brokerage reconnect when they detect data-quality anomalies (e.g. consecutive zero-volume bars during market hours while the connection is technically alive). A method such as `self.reconnect_brokerage()` on `QCAlgorithm` would allow the algorithm to recover without human intervention or process restart.

#### Actual Behavior

There is no way for an algorithm to trigger a brokerage `Disconnect/Connect` cycle. `QCAlgorithm` holds an internal `IBrokerage` reference but does not expose it. The only workarounds are:

- `self.Quit()` — exits the algorithm entirely, disrupting live positions
- Out-of-process IPC (flag files, sockets) between the algorithm thread and the brokerage heartbeat thread — fragile and non-portable

This gap is particularly visible with IBKR, where the gateway can enter a state where heartbeats succeed but the market data feed silently freezes (bars arrive with `Volume = 0`). The algorithm layer can detect this condition but has no recourse.

#### Potential Solution

Add a single public method to `QCAlgorithm`:

```csharp
public void ReconnectBrokerage()
{
    try { Brokerage.Disconnect(); } catch (Exception) { }
    Brokerage.Connect();
}
```

Python usage:

```python
def on_consolidated_bar(self, bar):
    if self._is_data_frozen(bar):
        self.log("Data feed frozen, triggering reconnect")
        self.reconnect_brokerage()
```

The change is brokerage-agnostic and minimal — QCAlgorithm already holds the IBrokerage reference internally.

Open design questions we'd welcome feedback on before submitting a PR:

Should there be a built-in cooldown guard, or leave rate-limiting to the algorithm author?
Should the method emit a BrokerageMessageEvent before reconnecting so the live trading monitor is aware?
We are happy to contribute a PR if the team agrees on the API shape.

#### Checklist
- [x] I have confirmed that this issue exists on the current master branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues]

violet-day:
## Summary
  - Adds `ReconnectBrokerage()` to `IAlgorithm` and `QCAlgorithm`
  - Adds `SetBrokerageReconnectAction(Action)` to allow the engine to inject the actual
  Disconnect/Connect logic
  - Wires the action in `Engine.cs` after brokerage creation
  - Implements the wrapper in `AlgorithmPythonWrapper` for Python algorithms
  - 4 unit tests covering: action invoked, no-op when unregistered, null guard, order of
  operations

  ## Motivation
  Closes #9496 — algorithms running in live trading (e.g. against IBKR) can detect a silently
  frozen data feed (bars with `Volume=0`) but previously had no way to trigger a reconnect
  short of calling `Quit()`. This API lets them self-heal without human intervention.

  ## Notes
  - Brokerage-agnostic: the `Action` delegate is injected by the engine, so `QCAlgorithm` has
  no dependency on any concrete brokerage
  - No-op in backtesting (action is never registered, logs a debug message)

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -121,6 +121,8 @@ namespace QuantConnect.Algorithm
         private IStatisticsService _statisticsService;
         private IBrokerageModel _brokerageModel;
 
+        private Action _brokerageReconnectAction;
+
         private bool _sentBroadcastCommandsDisabled;
         private readonly HashSet<string> _oneTimeCommandErrors = new();
         private readonly Dictionary<string, Func<CallbackCommand, bool?>> _registeredCommands = new(StringComparer.InvariantCultureIgnoreCase);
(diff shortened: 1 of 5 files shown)


violet-day:
@Martin-Molinero

Issue #5595, opened 2021-05-24
Reorganize Schedule Event Calls in AlgorithmManager


AlexCatarino:
#### Expected Behavior
Scheduled Events are triggered after consolidators/indicators are updated.

#### Actual Behavior
Scheduled Events are triggered before consolidators/indicators are updated.

#### Potential Solution
Reorganize the calls.

#### Reproducing the Problem
```csharp
namespace QuantConnect.Algorithm.CSharp {
    public class SwimmingBlackArmadillo : QCAlgorithm {
        private int _day = 0;
		
        public override void Initialize() {
            SetStartDate(2021, 5, 1);

            AddEquity("SPY");
            SMA("SPY", 5).Updated += OnSmaUpdated;

            Schedule.On(DateRules.EveryDay("SPY"), TimeRules.AfterMarketOpen("SPY", 1),
            	() => Log($"AfterMarketOpen on {Time}")
            );
        }

        private void OnSmaUpdated(object sender, IndicatorDataPoint updated) {
            if (_day == updated.EndTime.Day) return;

            _day = updated.EndTime.Day;
            Log($"OnSmaUpdated on {Time}");
        }
    }
}
```
#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)
- [x] I have provided detailed steps to reproduce the issue

PR #9542, opened 2026-06-17
Run past consolidators before scheduled events

### LINKED ISSUE #8649: The Schedule runs before the Consolidator's Callback regardless of the time
#### Expected Behavior
The scheduled event at 1 AM should execute after the RSI indicator update, which occurs at midnight (12 AM)

#### Actual Behavior
The scheduled event at 1 AM executes before the RSI indicator update

#### Reproducing the Problem
Executes this code:

```
{
    public class CrawlingTanFrog : QCAlgorithm
    {
        private RelativeStrengthIndex _relativeStrengthIndex;
        private Symbol _spy;
        private bool _itWasUpdated;
        private int _dataPointCount;
        {
            SetStartDate(2013, 01, 01);
            SetEndDate(2013, 01, 05);
            _spy = AddEquity("SPY", Resolution.Hour).Symbol;
            _relativeStrengthIndex = new RelativeStrengthIndex(14, MovingAverageType.Wilders);
            RegisterIndicator(_spy, _relativeStrengthIndex, TimeSpan.FromDays(1));


            var history = History<TradeBar>(_spy, 20, Resolution.Daily).ToList();
            foreach (var bar in history)
            {
                _relativeStrengthIndex.Update(bar.EndTime, bar.Close);
            }
            if (!_relativeStrengthIndex.IsReady)
            {
                throw new RegressionTestException($"{_relativeStrengthIndex.Name} is not ready.");
            }
            _relativeStrengthIndex.Updated += (sender, data) =>
            {
                var updatedTime = Time;

                // RSI1 should update at midnight when precise end time is disabled
                if (updatedTime.TimeOfDay != new TimeSpan(0, 0, 0))
                {
                    throw new RegressionTestException($"{_relativeStrengthIndex.Name} must have updated at midnight, but it was updated at {updatedTime}");
                }

                _itWasUpdated = true;
            };

            _dataPointCount = 0;

            // 1 AM
            Schedule.On(DateRules.EveryDay(), TimeRules.At(1, 0, 0), TestSchedule);
        }

        private void TestSchedule()
        {
            // Expect that RSI was updated at midnight
            // Because the scheduler is set to run at 1 AM and the RSI was updated at midnight
            if (!_itWasUpdated && _dataPointCount > 0)
            {
                throw new RegressionTestException("The indicator was not updated.");
            }
        }

        {
            _dataPointCount++;
            var time = Time;
        }
    }
}
```

#### Checklist

yinli-systems:
## Description

Fixes #8649.

`AlgorithmManager.Run` currently scans past scheduled events before scanning past consolidators. When the engine catches up over a period with no data, a scheduled event can run before a due consolidator scan updates registered indicators, so the scheduled callback observes stale indicator state.

This moves the past consolidator scan ahead of `ScanPastEvents` while preserving the same performance tracking blocks. The added regression test runs `AlgorithmManager` with a single time-pulse slice and asserts that the realtime handler only scans past scheduled events after a due consolidator scan has run.

## Related

Related to #5595.

## Testing

Validated after merging the current upstream `master` (through `0136529cd`), using .NET SDK 10.0.302 on macOS Arm64:

```bash
dotnet test Tests/QuantConnect.Tests.csproj \
  --filter "FullyQualifiedName~ScheduledEventsRunAfterPastConsolidators" \
  --logger "console;verbosity=minimal"
```

Result: passed, 1 test.

The regression test was also fault-injected by temporarily restoring the old ordering. It failed at `AssertConsolidatorScannedBeforeScheduleHandler.ScanPastEvents`, confirming that the test detects the stale-indicator ordering bug. After restoring the fix, the same test passed again.

## AI assistance

AI assistance was used to revalidate the implementation, run the fault-injection check, merge current upstream changes, and refine this PR description. The resulting changes and test behavior were reviewed before submission.


yinli-systems:
Follow-up with the intended behavior this PR is trying to lock down: when a scheduled event and a consolidator are both due at the same frontier time, the consolidator should be advanced first so the scheduled event observes the latest completed bar, not stale state from the previous frontier.

The regression test covers that ordering directly. I kept the change localized to the time-slice ordering path rather than changing scheduled event semantics more broadly. Happy to adjust if there is a more appropriate Lean convention for this case.

yinli-systems:
Synced this branch with current master (104d3dcee) using the repository’s post-push merge workflow and reran it on .NET 10. The focused regression passes, all six AlgorithmManagerTests pass, and restoring the old order makes the regression fail at ScanPastEvents because the consolidator has not yet scanned. @Martin-Molinero, since you linked the related scheduling issue, could you confirm that consolidator-before-scheduled-event is the intended engine ordering?

PR #9557, opened 2026-06-23
Use risk-free rate in probabilistic Sharpe ratio

### LINKED ISSUE #9533: PSR calculation omits the risk-free rate
#### Expected Behavior

The Probabilistic Sharpe Ratio (PSR) should be statistical inference about the **same** Sharpe ratio the engine reports. The reported `SharpeRatio` is computed on excess returns (it subtracts the risk-free rate), so the PSR — defined as the probability that the true Sharpe ratio exceeds a benchmark — should also be built on the risk-free-adjusted Sharpe. A strategy whose reported (risk-free-adjusted) Sharpe ratio is strongly negative should therefore show a **low** PSR, not a near-100% one. The two statistics should be consistent with each other.

#### Actual Behavior

The reported Sharpe ratio and the PSR are built from two different Sharpe ratios:

- `SharpeRatio` uses **excess** returns — it subtracts the risk-free rate.
- The PSR's internal Sharpe (`ObservedSharpeRatio`) uses **gross** returns — `mean / standard deviation`, with no risk-free term — and compares it against a benchmark `SR* = 1 / √(tradingDaysPerYear)` (an annualized Sharpe of 1.0).

So the PSR actually estimates `P(true gross Sharpe > 1.0)`, while the reported Sharpe is the excess-over-risk-free Sharpe. For a low-volatility holding during a high-rate period the two contradict each other completely. A buy-and-hold of SHV (a 0–1 year Treasury-bill ETF, about as close to the risk-free asset as exists) reports:

- **Sharpe Ratio: −8.88**
- **Probabilistic Sharpe Ratio: 100%**
- Compounding Annual Return: 4.60%, Annual Standard Deviation: 0.25%, Max Drawdown: 0.10%

The most cash-like asset available simultaneously looks terrible (Sharpe −8.88) and certain (PSR 100%). Decomposing it: the gross annualized Sharpe is `4.60% / 0.25% ≈ 18`, which sits far above the PSR's `1.0` benchmark over ~860 observations, pinning the PSR at 100%. The reported Sharpe subtracts the risk-free rate, and because the volatility is only 0.25%, that subtraction moves the ratio by `rf / 0.25% ≈ 27` units, dragging it to −8.88.

#### Potential Solution

Compute `ObservedSharpeRatio` on risk-free-adjusted returns (subtract the per-period risk-free rate before forming `mean / standardDeviation`), so the PSR is inference about the same excess-return Sharpe the engine reports. This also matches Bailey & López de Prado (2012), whose PSR is defined on the standard excess-return Sharpe ratio.

With that change, SHV's Sharpe becomes `−8.88 / √252 ≈ −0.56`, which is far below the `1/√252 ≈ 0.063` benchmark, so the PSR would fall to ≈ **0%** — consistent with the −8.88 Sharpe instead of contradicting it.

Relevant code (`Common/Statistics`):

- `PortfolioStatistics.cs`: `SharpeRatio = Statistics.SharpeRatio(annualPerformance, AnnualStandardDeviation, riskFreeRate)` and `benchmarkSharpeRatio = 1.0d / Math.Sqrt(tradingDaysPerYear); ProbabilisticSharpeRatio = Statistics.ProbabilisticSharpeRatio(listPerformance, benchmarkSharpeRatio)`.
- `Statistics.cs`: `SharpeRatio(...) => (averagePerformance - riskFreeRate) / standardDeviation` (subtracts rf), but `ObservedSharpeRatio(...) => performanceAverage / standardDeviation` (no rf).

Also worth confirming the `rf = 0` choice inside `ObservedSharpeRatio` and the `SR* = 1/√252` benchmark were intentional and are on the same basis.

#### Reproducing the Problem

Buy-and-hold SHV from 2023-01-01 to 2026-06-01 reliably produces a strongly negative Sharpe ratio alongside a 100% PSR.

```python

class BuyAndHoldSHV(QCAlgorithm):

    def initialize(self):
        self.set_start_date(2023, 1, 1)
        self.set_end_date(2026, 6, 1)
        self.set_cash(10_000_000)
        self.add_equity("SHV", Resolution.DAILY)

    def on_data(self, data):
        if not self.portfolio.invested:
            self.set_holdings("SHV", 1)
```

Public backtest: https://www.quantconnect.cloud/backtest/281b5a27fc39d1cd448ef70792fc37c1/?theme=chrome

Result: Sharpe Ratio −8.88 and Probabilistic Sharpe Ratio 100%

#### System Information

QC Cloud

#### Checklist

yinli-systems:
#### Description

Fixes #9533.

`PortfolioStatistics` reports SharpeRatio on excess returns by subtracting the risk-free rate, but ProbabilisticSharpeRatio was still using the gross-return observed Sharpe ratio. That can make a low-volatility cash-like series report a strongly negative Sharpe ratio and a near-100% PSR at the same time.

This PR keeps the existing PSR/ObservedSharpeRatio APIs backward compatible by adding an optional per-sample `riskFreeRate = 0` parameter, then has `PortfolioStatistics` pass the annual risk-free rate deannualized by `tradingDaysPerYear`. The PSR is then inference about the same excess-return Sharpe ratio basis as the reported SharpeRatio.

#### Related Issue

#9533

#### Motivation and Context

The reported Sharpe ratio and PSR should be on the same return basis. Otherwise high-rate, low-volatility assets can show contradictory statistics.

No.

- `git diff --check`
- Formula sanity check outside Lean: the new regression sample has gross observed Sharpe around 10.9, giving PSR approximately 1.0, but after subtracting a per-sample risk-free rate of 0.00025 the observed Sharpe is around -3.23 and PSR is approximately 1.6e-12.

I could not run `dotnet test` locally because this machine does not have `dotnet`, `csc`, `mcs`, or `msbuild` installed.

#### Types of changes

- [x] New test

#### Checklist

- [x] All new and existing tests passed where runnable locally.

### DIFF
--- a/Common/Statistics/PortfolioStatistics.cs
+++ b/Common/Statistics/PortfolioStatistics.cs
@@ -309,7 +309,7 @@ namespace QuantConnect.Statistics
 
             // deannualize a 1 sharpe ratio
             var benchmarkSharpeRatio = 1.0d / Math.Sqrt(tradingDaysPerYear);
-            ProbabilisticSharpeRatio = Statistics.ProbabilisticSharpeRatio(listPerformance, benchmarkSharpeRatio).SafeDecimalCast();
+            ProbabilisticSharpeRatio = Statistics.ProbabilisticSharpeRatio(listPerformance, benchmarkSharpeRatio, (double)riskFreeRate / tradingDaysPerYear).SafeDecimalCast();
 
             ValueAtRisk99 = GetValueAtRisk(listPerformance, tradingDaysPerYear, 0.99d);
             ValueAtRisk95 = GetValueAtRisk(listPerformance, tradingDaysPerYear, 0.95d);
--- a/Common/Statistics/Statistics.cs
+++ b/Common/Statistics/Statistics.cs
@@ -195,11 +195,13 @@ namespace QuantConnect.Statistics
         /// </summary>
         /// <param name="listPerformance">The list of algorithm performance values</param>
         /// <param name="benchmarkSharpeRatio">The benchmark sharpe ratio to use</param>
+        /// <param name="riskFreeRate">The risk free rate for each performance sample</param>
         /// <returns>Probabilistic Sharpe Ratio</returns>
         public static double ProbabilisticSharpeRatio(List<double> listPerformance,
-             double benchmarkSharpeRatio)
+             double benchmarkSharpeRatio,
+             double riskFreeRate = 0)
         {
-            var observedSharpeRatio = ObservedSharpeRatio(listPerformance);
+            var observedSharpeRatio = ObservedSharpeRatio(listPerformance, riskFreeRate);
 
             var skewness = listPerformance.Skewness();
             var kurtosis = listPerformance.Kurtosis();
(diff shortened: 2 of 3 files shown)


Martin-Molinero:
Thank you @Kevin-Li-2025! Changes were included in shipped https://github.com/QuantConnect/Lean/pull/9583

PR #9563, opened 2026-06-25
Wait for fresh data before filling market orders on stale data

jhonabreul:
#### Description
Market orders could be filled on stale data, the most visible case being right at/after the market open: an order placed in the first moments of the session (e.g. a scheduled trade one second after the open) would fill on the previous trading date's bar, because the first bar of the current session has not been emitted yet.

This generalizes the fix to be independent of the time of day. A market order that would be filled on stale data now waits for fresh data when the latest available data is more than one subscribed resolution bar behind the current time (so a newer bar is still expected). This covers both:
- the market open (the first session bar has not arrived yet), and
- any intraday data gap larger than the resolution.

Details:
- `ShouldWaitForFreshDataOnStale` takes the latest data end time and the current time and is shared by `FillModel`, `FutureFillModel` and `EquityFillModel` at their stale-data guards. Coarse resolutions (hour/daily) always wait; tick never waits. Internal configurations are included when sizing the resolution bar.
- `ShouldWaitForFreshData` is intentionally left untouched at its `GetMarketFillPrice` call site (it selects bar-open vs current price and is not gated by staleness), so non-stale fill prices are unchanged.
- `EquityFillModel`'s best-effort price helpers now report the stale data end time so the gap can be measured.

The wait is gated on the data still being stale, so once a fresh bar (or its fill-forward) is available the order fills normally — it does not deadlock.

#### Related Issue
No tracked issue. Reproduced from `Algorithm.Python/TestAlgorithm.py` (market order scheduled one second after the open on minute data). Happy to open/link an issue if preferred.

#### Motivation and Context
A market order should not be filled at a past, stale price when fresh data is still expected. Doing so produces fills at incorrect prices and misleading backtest results — most notably at the open, but the same applies to any intraday gap.

No.

- `EquityFillModelTests` and `FutureFillModelTests`: new tests covering the market-open and mid-session stale cases (the order waits, then fills once fresh data within one resolution bar is available), plus the within-one-bar boundary (fills on the stale price). Verified these fail without the fix and pass with it.
- `MarketOrderStaleDataFillRegressionAlgorithm` (SPY minute, 2013-10-07 to 2013-10-11): places orders at the open and mid-session and asserts in `OnOrderEvent` that no fill ever lands on data staler than the resolution. It errors without the fix and completes with it.
- Pre-existing fill-model plumbing/data-selection tests that relied on degenerate timestamps were given fresh timestamps so they still exercise their original intent.
- Full `Orders.Fills` unit suite passes.

Note: this generalizes when a market order waits on stale data, so some regression algorithms that previously filled on stale (more than one bar old) data may shift their expected statistics; the full `RegressionTests` suite should be run in CI to update them.

#### Types of changes

#### Checklist:


jhonabreul:
## Regression Tests triage — with vs without the fill-model change

I reproduced the failing **Regression Tests** job locally and analyzed every failing **CSharp** algorithm by running each one twice:

- **with fix** = this branch as-is
- **without fix** = `Common/Orders/Fills/{FillModel,EquityFillModel,FutureFillModel}.cs` reverted to `master` (the only behavioral code in the PR), everything else identical

…then diffed the per-algorithm order events and statistics. Per the request, `Python/{name}` failures are ignored where a `CSharp/{name}` exists — all 19 Python failures are duplicates of a CSharp one, leaving **28 CSharp algorithms**.

### Root cause (one mechanism, a spectrum of severity)

The fix makes a market order **wait for fresh data** whenever the latest bar is more than one resolution-span behind the current time. That is correct in principle, but the open-source sample data the regression suite runs on is **sparse / gapped / ends at expiry-or-roll**. So "wait for fresh data" can mean:

1. fill one bar later at a fresh price → **only the statistics change** (expected), or
2. fill *many days* later when the next real bar finally appears → trade still completes but **P&L changes materially**, or
3. **never fill**, because no further bar ever arrives (expiring option / rolled-out future / data that ends) → the intended entry/exit doesn't happen, and algorithms that re-submit while flat emit an **order storm**.

The three groups below map to those three outcomes.

---

### 🟢 Group A — Expected (just regenerate `ExpectedStatistics`)

Order still fills, only the **price/time/size** changes to use fresh data. Trade completes; no broken invariant.

| Algorithm | Evidence (no-fix → with-fix) |
|---|---|
| BasicTemplateContinuousFuture | sell fill 1789.5 → 1806.25; Net 5.55% → 6.39% |
| ContinuousFuture | fill 1699.5 → 1798.25; Net 1.69% → 6.63% |
| ContinuousFutureBackMonth | fill 1692.5 → 1791.5; Net 2.29% → 7.24% |
| ContinuousBackMonthRawFuture | fill 1692.5 → 1791.5; Net 1.48% → 6.43% |
| ManualContinuousFuturesPositionRollover | fill 1809.75 → 1810.0; Net 7.579% → 7.591% |
| ManualContinuousFuturesPositionRolloverFromSymbolChangedEventHandler | fill 1809.75 → 1810.0; Net 7.579% → 7.591% |
| FuturesFramework | fill 1619.0 → 1620.75; Net −20.99% → −20.64% |
| CustomDataUsingMapFile | fill 32.59 → 29.15; Net −10.34% → +0.18% |
| FutureOptionChainFullData | same-day intraday churn, fresh-price sizing; Total Orders 450 → 304 |
| FutureOptionChainsMultipleFullData | Total Orders 450 → 304 |
| IndexOptionCallOTMExpiry | buy 0.10 → 0.05; Net −0.010% → −0.005% (expires OTM, unchanged outcome) |
| IndexOptionShortCallOTMExpiry | sell 0.10 → 0.05 |
| IndexOptionPutOTMExpiry | buy 3.7 → 0.1; Net −0.37% → −0.01% |
| IndexOptionShortPutOTMExpiry | same fill price (3.4), **fill deferred 01/05 → 01/14**; only Drawdown moves |

### 🟡 Group B — Expected mechanism, but materially different — please eyeball

The order eventually fills, but because the option's next real bar in the sample data is ~9 days out (gap 01/05 → 01/14), the deferred fill changes the trade's P&L a lot. The fix is doing what it's designed to; the magnitude is an artifact of sparse sample data. Worth confirming the new numbers are acceptable before baking them in.

| Algorithm | Evidence |
|---|---|
| IndexOptionCallITMExpiry | buy 504.8 → 620.4 (filled 9 days late); Net **+9.07% → −2.49%** |
| IndexOptionPutITMExpiry | buy 503.0 → 384.5; Net **−9.85% → +2.00%** |
| IndexOptionShortCallITMExpiry | sell 500.3 → 615.5; Net −0.95% → +0.20% |
| IndexOptionShortPutITMExpiry | sell 498.2 → 379.6; Net +0.94% → −0.25% |
| IndexOptionBuySellCallIntraday | intraday buy/sell on 01/05 → entry deferred to **01/14**, position now **held to expiry** instead of closed intraday; Win/Loss flips 0%→100% |
| AddOptionContractExpires | `Liquidate` of the exercised equity **never fills** (TWX/AOL sample data ends 06/09, order placed 06/23) → equity left open to the end; Win 50% → 0%. Same root cause as Group C, but it completes without throwing and the branch already updated its stats. |

### 🔴 Group C — Looks like a regression to guard, **not** a simple stat update

Here "wait for fresh data" means the trade **never executes** (no further bar ever arrives for an expiring/rolled/sparse instrument), and flat-and-resubmit algorithms produce an **order storm**. Two even throw their own end-of-algorithm assertions.

| Algorithm | No-fix → With-fix | Symptom |
|---|---|---|
| BasicTemplateFutureRollover | 1 order / 1 fill → **28 orders / 0 fills** | daily continuous future; the roll never executes |
| BasicTemplateContinuousFutureWithExtendedMarket | 5 / 5 → **100 orders / 0 fills** | never trades |
| BasicTemplateSPXWeeklyIndexOptions | 5 orders → **1622 orders, 810 cancels, 2 fills** | order storm (resubmits every bar while flat) |
| BasicTemplateSPXWeeklyIndexOptionsStrategy | 6 / 10 → **100 orders / 0 fills** | strategy legs never fill |
| Collective2IndexOption | 10 orders → **55 orders, 51 cancels, 4 fills** | order churn |
| TradingNotAddedOptions | 8 / 8 → **0 fills** (orders canceled) | never trades |
| FutureOptionMultipleContractsInDifferentContractMonthsWithSameUnderlyingFuture | 4 / 4 → **0 fills**, **throws** `Expected holdings at the end of algorithm, but none were found` (cs:79) | never enters |
| OptionAssignmentStatistics | 7 fills → 6 fills, **throws** `AssertTradeStatistics(): Expected 4 closed trades … Actual: 2` (cs:113) | underlying liquidation never fills |

---

### Recommendation

- **Groups A & B (20 algos):** failures are the *expected* consequence of the fix — regenerate `ExpectedStatistics` (spot-check Group B's P&L swings first).
- **Group C (8 algos):** these are *regressions introduced by the PR*, not stat updates. The wait is unbounded, so a market order can rest forever when no further bar will ever arrive (expiring option, rolled-out/delisted future), and flat-and-resubmit strategies spam orders. Consider **bounding the wait** in `ShouldWaitForFreshDataOnStale` — e.g. fall back to the best-effort (stale) fill when the subscription is ending / about to delist or the order has rested beyond N bars — so positions can still be entered and, critically, **exited**. The futures-rollover case (must be able to close an expiring contract) and the 1600-order storm are the clearest signals that an indefinite wait is too aggressive.

<sub>Method note: "without fix" reverts only the three fill-model files to `master`; the PR's other changes (new `MarketOrderStaleDataFill` algo + unit tests) don't affect these algorithms. Numbers are from local runs on the bundled sample data, same data CI uses.</sub>

Martin-Molinero (approved):

PR #9568, opened 2026-06-26
Seed runtime-added currency conversion rates immediately

JosueNina:
#### Description
If a currency is added while the algorithm is running (for example a scheduled event adds a security whose quote currency isn't in the cashbook yet), the runtime path created the conversion feed but never set its rate. The rate stayed at 0 until the first bar of the pair arrived, so any conversion before that, like a scheduled SetHoldings, would throw "The conversion rate for X is not available".

The warmup setup path already seeds these rates, but the runtime path didn't. This change makes the runtime path seed them the same way, only when a new feed is actually added, and it falls back safely when there is no data so live trading isn't affected.

#### Motivation and Context
N/A

N/A

It was tested using unit and regression tests.

#### Types of changes

#### Checklist:


jhonabreul (approved):
Thank you!

Martin-Molinero (approved):

PR #9573, opened 2026-06-30
Fix method name parsing in NoMethodMatchPythonExceptionInterpreter

jhonabreul:
#### Description

`NoMethodMatchPythonExceptionInterpreter` builds a user-facing message that names the method that failed overload resolution. It extracted that name with `LastIndexOf(" ")`, which returns the **last** space-delimited token of the pythonnet message rather than the method name. The pythonnet message has the form:

```
No method matches given arguments for RSI: (<class 'QuantConnect.Symbol'>, <class 'int'>, <class 'QuantConnect.Resolution'>)
```

so the last token is a fragment of the argument-type list (e.g. `'QuantConnect.Resolution'>)`), producing a garbled message:

> ...ensure each parameter type matches those required by the `'QuantConnect.Resolution'>)` method. Please checkout the API documentation.

The method name is now parsed from between the `"for "` keyword and the following `":"`, yielding the correct `RSI` / `SetCash` etc.

#### Related Issue

N/A — bug found while running a Python algorithm whose `initialize` called `self.rsi(symbol, 15, Resolution.DAILY)` (the third argument should be a `MovingAverageType`).

#### Motivation and Context

The interpreted exception is meant to help users fix a bad call by telling them which method's parameters didn't match. The broken parsing instead pointed at a piece of the argument list, making the guidance confusing and useless.

No.

Added two tests to `NoMethodMatchPythonExceptionInterpreterTests`:
- `VerifyMessageContainsTheMethodName` — using the existing `set_cash('SPY')` reproduction, asserts the interpreted message contains `SetCash` and not the `">)"` argument-list fragment.
- `VerifyMessageContainsTheMethodNameForOverloadedMethod` — reproduces the reported scenario via a new `no_method_match_rsi` method (`self.rsi(symbol, 15, Resolution.DAILY)`) and asserts the message contains `RSI` and not `">)"`.

Both tests fail before the fix and pass after. The full `NoMethodMatchPythonExceptionInterpreterTests` fixture passes (13/13).

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Exceptions/NoMethodMatchPythonExceptionInterpreter.cs
+++ b/Common/Exceptions/NoMethodMatchPythonExceptionInterpreter.cs
@@ -50,13 +50,36 @@ namespace QuantConnect.Exceptions
         {
             var pe = (PythonException)exception;
 
-            var startIndex = pe.Message.LastIndexOfInvariant(" ");
-            var methodName = pe.Message.Substring(startIndex).Trim();
+            var methodName = GetMethodName(pe.Message);
             var message = Messages.NoMethodMatchPythonExceptionInterpreter.AttemptedToAccessMethodThatDoesNotExist(methodName);
 
             message += PythonUtil.PythonExceptionStackParser(pe.StackTrace);
 
             return new MissingMethodException(message, pe);
+        }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


Martin-Molinero (approved):
Nice!

PR #9575, opened 2026-07-01
Fix duplicate OnData when dynamically adding an at-expiry contract

### LINKED ISSUE #9279: Double OnData Call When Add Subcription in Scheduled Event
#### Expected Behavior
OnData is called once per period.

#### Actual Behavior
OnData is called more than once when we add a subscription during a scheduled event

#### Potential Solution
N/A

#### Reproducing the Problem
https://www.quantconnect.cloud/backtest/c932358231d295109bcb90e66e26fa74/?theme=chrome

Image

#### Checklist

jonathanwu906:
#### Description

Adding a delistable contract (e.g. `AddOptionContract`) at/after its delisting date from inside a scheduled event caused `OnData` to fire twice at the same timestamp — the second slice carrying only a `Delisting.Warning`. This makes it fire once per period as expected.

#### Related Issue

Closes #9279

#### Motivation and Context

`CorporateEventEnumeratorFactory` always attaches a `DelistingEventProvider` for a `CanBeDelisted()` security. When the subscription starts on/after the delisting date (a near/at-expiry contract added dynamically, e.g. in a `Schedule.On` callback), the provider emits a `DelistingType.Warning` on its first tradable date, which arrives as a separate slice at the same time → the duplicate `OnData`.

The provider's date-based emission is intentional (covered by `DelistingEnumeratorTests`), so the fix is in the factory: skip the provider when the subscription starts on/after the delisting date.

No.

- New regression algorithm `AddOptionContractInScheduledEventRegressionAlgorithm`: fails without the fix (throws → `RuntimeError`), passes with it.
- `DelistingEnumeratorTests` pass (provider unchanged); existing option/future delisting regressions produce identical `OrderListHash`.

#### Types of changes

#### Checklist:

#### Note for reviewers

A subscription starting on/after its delisting date now receives no delisting events (both `Warning` and `Delisted`). This looks correct for an already-expired asset — happy to only skip the `Warning` if you prefer.

### DIFF
--- a/Algorithm.CSharp/AddOptionContractInScheduledEventRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/AddOptionContractInScheduledEventRegressionAlgorithm.cs
@@ -1,0 +1,138 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+using System.Linq;
+using System.Collections.Generic;
+using QuantConnect.Data;
+using QuantConnect.Interfaces;
+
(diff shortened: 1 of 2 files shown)


jonathanwu906:
@Martin-Molinero this implements a fix for #9279, based on the cause you identified — a stale `Delisting.Warning` emitted immediately for a contract added at/after its delisting date.

The fix is in `CorporateEventEnumeratorFactory`: skip attaching the `DelistingEventProvider` when the subscription starts on/after the delisting date, leaving the provider's own (tested) date-based behavior untouched. I added a regression algorithm that fails without the fix and passes with it, and the existing option/future delisting regressions are unchanged (identical `OrderListHash`).

One design question: a subscription added at/after its delisting date now receives no delisting events (both `Warning` and `Delisted`). Happy to change it to skip only the `Warning` if you'd prefer.

Would appreciate a review when you have a chance — thanks!

Martin-Molinero:
No sorry, the engine requires these warnings, don't think solution is as clear as it seems, suggest leaving for a QC team member to define potential solution first

PR #9582, opened 2026-07-02
Auto-subscribe symbols when registering indicators and consolidators

jhonabreul:
#### Description
Registering an indicator or a consolidator for a symbol that has not been subscribed to used to throw:

> Please register to receive data for symbol '...' using the AddSecurity() function.

Order submission already auto-subscribes a symbol on the user's behalf (see `GetSecurityForOrder`). This PR applies the same behavior to indicator and consolidator registration: `GetSubscription` now adds the security automatically when it has no subscription, before falling back to the error message.

Details:
- The tradability/expiry guard used by order submission was extracted into a shared `CanAutoAddSecurity` helper, now used by both `GetSecurityForOrder` and `GetSubscription`. Order submission keeps its extra `SecurityType != Index` check (indexes are not tradable), while the subscription path allows indexes since it only needs their data.
- The subscription lookup in `GetSubscription` was refactored into a `bool TryGetSubscription(..., out SubscriptionDataConfig subscription)` helper, replacing the previous try/catch control flow.

#### Related Issue
N/A

#### Motivation and Context
Users can already trade a symbol without subscribing to it (order submission subscribes it automatically). It is surprising and inconsistent that registering an indicator or consolidator for that same symbol throws instead. This makes the behavior consistent across the API.

No.

- Added `RegisterIndicatorAndConsolidatorWithoutSubscriptionRegressionAlgorithm`, which never calls `AddEquity`/`AddSecurity` and:
  - registers an SMA indicator and a 30-minute consolidator for SPY (exercising the indicator/consolidator auto-add), asserting the indicator receives data and the consolidator produces bars;
  - submits a `MarketOrder` for a separate, unsubscribed symbol (AIG) (exercising the order-submission auto-add), asserting the order is placed and filled and that the position is established.
- Verified the algorithm fails on `master` (throws during initialization) and passes with this change.
- Ran the related unit fixtures with no regressions: `AlgorithmIndicatorsTests`, `AlgorithmResolveConsolidatorTests`, `AlgorithmRegisterIndicatorTests`, `AlgorithmSubscriptionManagerRemoveConsolidatorTests`.

#### Types of changes

#### Checklist:

### DIFF
--- a/Algorithm.CSharp/RegisterIndicatorAndConsolidatorWithoutSubscriptionRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/RegisterIndicatorAndConsolidatorWithoutSubscriptionRegressionAlgorithm.cs
@@ -1,0 +1,197 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


Martin-Molinero (approved):
Nice!

PR #9627, opened 2026-07-19
Fix trade statistics for option assignment underlying fills

YadavKapil:
## Description

Physical option settlement emits an option fill and an underlying fill with the same option exercise order ID. `BrokerageTransactionHandler` previously used the parent order's security when sending both events to `TradeBuilder`, so an underlying fill that closed a trade used the option contract multiplier and quote-currency conversion rate.

For a standard US equity option, the existing `OptionAssignmentRegressionAlgorithm` reproduced an expected underlying loss of $20,000 as $2,000,000. The underlying fill quantity already contains the contract unit of trade, so applying the option multiplier again inflated closed-trade P&L, MAE, MFE, and end-trade drawdown.

This change resolves the security from `orderEvent.Symbol`, matching portfolio fill processing. Normal fills are unchanged because their order and event symbols are identical.

The existing option-assignment regression now verifies every closed underlying trade uses the underlying security's multiplier.

## Tests

- `OptionAssignmentRegressionAlgorithm` (fails before the fix, passes after)
- `OptionAssignmentStatisticsRegressionAlgorithm`
- `OptionExerciseAssignRegressionAlgorithm`

### DIFF
--- a/Algorithm.CSharp/OptionAssignmentRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionAssignmentRegressionAlgorithm.cs
@@ -81,19 +81,33 @@ namespace QuantConnect.Algorithm.CSharp
             }
         }
 
+        private Security GetSecurity(Symbol symbol)
+        {
+            if (symbol == Stock.Symbol)
+            {
+                return Stock;
+            }
+            if (symbol == CallOptionSymbol)
+            {
+                return CallOption;
+            }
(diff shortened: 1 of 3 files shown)


jhonabreul (approved):
Thank you for your contribution!
I pushed a commit with a small change: just moved the solution up earlier in the method and updated the algorithms.

Martin-Molinero (approved):

PR #9628, opened 2026-07-20
Fix CryptoFuture.IsCryptoCoinFuture() misclassifying USD-quoted linea…

### LINKED ISSUE #9574: CryptoFuture.IsCryptoCoinFuture() incorrectly classifies USD-quoted linear futures as coin-margined, breaking EU MiCA-compliant exchanges
`CryptoFuture.IsCryptoCoinFuture()` hardcodes a whitelist of quote currencies (`USDT`, `BUSD`, `USDC`) to distinguish linear from coin-margined futures. This breaks any exchange that quotes linear perpetual futures in `USD` instead of `USDT` — which is exactly the case for EU MiCA-regulated exchanges.

**Background / Regulatory Context**

Under EU MiCA, compliant exchanges are restricted from offering stablecoins such as USDT or USDC as settlement currencies for derivatives — only fiat currencies are permitted. As a result, exchanges such as Kraken Futures and OKX EU quote their linear perpetual futures in **USD** (fiat) rather than USDT or USDC. Binance and Bybit have already withdrawn from EU futures markets due to MiCA compliance requirements, so going forward, any EU-compliant futures exchange will always quote in fiat USD — making this a structural blocker for EU exchange integration.

I have been implementing community brokerage connectors for Kraken Futures and OKX EU. This bug in LEAN core is a hard blocker that prevents me from proceeding with either connector.

**Root Cause**

In `Common/Securities/CryptoFuture/CryptoFuture.cs`:

```csharp
private static bool IsCryptoCoinFuture(string quoteCurrency)
{
    return quoteCurrency != "USDT" && quoteCurrency != "BUSD" && quoteCurrency != "USDC";
}
```

This method assumes only `USDT`/`BUSD`/`USDC`-quoted futures are linear. Any fiat `USD`-quoted future is incorrectly classified as coin-margined/inverse, regardless of its actual contract type. This affects:

- `CryptoFutureHolding.GetQuantityValue()` → `HoldingsValue` and `UnrealizedProfit` use the inverse formula (`quantity / price` in base currency instead of `price * quantity` in quote currency), producing a value equal to `Quantity` instead of the correct notional USD value.
- `CryptoFutureMarginModel.GetCollateralCash()` → collateral is incorrectly attributed to base currency (e.g. XBT) instead of quote currency (USD), breaking margin calculations.
- `SecurityPortfolioManager.TotalPortfolioValue` → completely wrong, causing incorrect position sizing in live trading.

**Impact**

- `TotalPortfolioValue` shows a BTC position as worth `2.8` USD instead of `~163,000` USD.
- Position sizing is off by a factor equal to the current BTC price.

iharshlalakiya:
Fixes #9574

`IsCryptoCoinFuture()` used a quote-currency whitelist (USDT/BUSD/USDC) to distinguish linear from coin-margined futures. This broke any exchange quoting linear perpetuals in fiat USD instead of a stablecoin — the case for EU MiCA-compliant exchanges (Kraken Futures, OKX EU), which are barred from using stablecoins for derivative settlement.

Since `CryptoFutureHolding.GetQuantityValue()` and `CryptoFutureMarginModel.GetCollateralCash()` both derive their behavior from this shared classification, the misclassification cascaded into wrong holdings value, collateral attribution, and total portfolio value for USD-quoted linear futures.

`IsCryptoCoinFuture` now also considers market: Binance keeps its existing quote-currency heuristic (Binance's coin-margined contracts are also nominally quoted in USD, e.g. BTCUSD_PERP), while other markets are treated as linear/USD-margined, since Lean currently has no coin-margined contracts outside Binance.

Added tests confirming Binance's BTCUSD coin-margined classification is unchanged, and that a Kraken-market BTCUSD future is now correctly treated as linear.

### DIFF
--- a/Common/Securities/CryptoFuture/CryptoFuture.cs
+++ b/Common/Securities/CryptoFuture/CryptoFuture.cs
@@ -60,7 +60,7 @@ namespace QuantConnect.Securities.CryptoFuture
                 cache,
                 new SecurityPortfolioModel(),
                 new ImmediateFillModel(),
-                IsCryptoCoinFuture(quoteCurrency.Symbol) ? new BinanceCoinFuturesFeeModel() : new BinanceFuturesFeeModel(),
+                IsCryptoCoinFuture(symbol.ID.Market, quoteCurrency.Symbol) ? new BinanceCoinFuturesFeeModel() : new BinanceFuturesFeeModel(),
                 NullSlippageModel.Instance,
                 new ImmediateSettlementModel(),
                 Securities.VolatilityModel.Null,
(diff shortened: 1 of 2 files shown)


Martin-Molinero:
Sorry closing for now, this would break bybit,  think the solution needs a missing abstraction whether in the SPDB or brokerage model drive, to be defined.

iharshlalakiya:
Hi @Martin-Molinero, good catch on Bybit. Revised the fix: instead of "non-Binance = linear," I whitelist only the markets where the USD-quote heuristic actually holds (Binance, Bybit — verified against SPDB, since contract-multiplier doesn't distinguish Bybit's linear vs inverse contracts either). Everything else, including Kraken/OKX EU and dYdX, is treated as linear.

Also found this isn't just a future-exchange issue — dYdX is already integrated and quotes in bare USD, and CryptoFutureMarginModel.GetCollateralCash() (unlike GetQuantityValue) has no dYdX carve-out, so collateral attribution for dYdX is wrong on master today.

Added Bybit/dYdX regression tests. Is this whitelist approach fine as a fix, or did you want the bigger SPDB/brokerage-model abstraction first?

PR #9637, opened 2026-07-26
Allow subscriptions to declare their real bar period

YadavKapil:
## Description

LEAN derives a bar's length from the *name* of its resolution, so data whose stored period differs from the nominal one is misdescribed. Concretely, 30-minute bars kept in the `minute` folders are treated as 1-minute bars: `FillForwardEnumerator` manufactures ~29 synthetic bars per real bar, and every bar is stamped with an end time 29 minutes too early.

This adds an optional `barPeriod` to `SubscriptionDataConfig`, surfaced on `AddEquity`, `AddIndex`, `AddOption`, `AddIndexOption`, `AddSecurity` and `UniverseSettings`:

```csharp
AddEquity("SPY", Resolution.Minute, barPeriod: TimeSpan.FromMinutes(30));
```

It defaults to `null`, so the period continues to be derived from the resolution and **no existing algorithm changes behaviour**. `Resolution` is deliberately untouched, so data still resolves from the same files and **no data needs regenerating**.

With the period declared, `ShouldFillForward` sees a gap equal to the period rather than 30x it, so no synthetic bars are emitted when the data is complete, and genuine gaps are still filled with exactly one bar.

## Four latent `Resolution`-as-duration reads

Four places read `Resolution` where they mean a duration. They agree today because `Increment` is a pure function of `Resolution`, so **each change is provably equivalent for any subscription that does not declare a period**, and the PR includes tests asserting that equivalence across all five resolutions.

| Location | Issue |
|---|---|
| `SubscriptionCollection` | Derived the fill-forward cadence from the resolution enum, which would reinstate the synthetic bars |
| `FillModel.ShouldWaitForFreshData` | Tested `Resolution.Hour/Daily` to decide whether a resting order fills at the bar **open** rather than its close. Now `Increment > Time.OneMinute`, which selects exactly `Hour` and `Daily`. Without this, a declared 30-minute bar would fill at the close of the *following* bar |
| `FillModel` stale threshold | Sized via `GetHighestResolution().ToTimeSpan()`. Added `GetHighestResolutionSpan()`, taking the minimum `Increment` |
| `CreateConsolidator` | Compared a `TimeSpan` against `Increment` and a `Resolution` against `Resolution` in one expression. Both now compare durations |

The consolidator change also fixes a pre-existing bug: against coarser-than-nominal data, `SMA(symbol, 20, Resolution.Minute)` evaluated `Minute < Minute` as false, skipped the guard, matched the identity branch and silently returned the coarser bars labelled as minute bars. It now throws.

`Equals`/`GetHashCode` include the declared period so configs differing only by it are not deduplicated, and the copy constructor drops it when the resolution changes, since it described the previous resolution.

## Testing

New fixture `Tests/Common/Data/SubscriptionDataConfigBarPeriodTests.cs` (23 tests) covers the equivalence proofs, config identity, copy-constructor inheritance, and validation. Two new `FillForwardEnumeratorTests` assert that a declared period emits zero synthetic bars with correct end times, while a genuine gap is still filled.

Suites run green in `quantconnect/lean:foundation`:

| Filter | Result |
|---|---|
| `SubscriptionDataConfigBarPeriodTests` | 23 passed |
| `FillForwardEnumeratorTests`, `SubscriptionCollectionTests`, `SubscriptionDataConfigTests` | 315 passed |
| `Fills`, `FillModel` | 544 passed, 2 skipped |
| `Consolidator`, `HistoryRequest`, `UniverseSettings` | 669 passed |

Full solution builds clean on net10.0.

### DIFF
--- a/Algorithm/QCAlgorithm.Indicators.cs
+++ b/Algorithm/QCAlgorithm.Indicators.cs
@@ -4247,7 +4247,10 @@ namespace QuantConnect.Algorithm
 
             // verify this consolidator will give reasonable results, if someone asks for second consolidation but we have minute
             // data we won't be able to do anything good, we'll call it second, but it would really just be minute!
-            if (period.HasValue && period.Value < subscription.Increment || resolution.HasValue && resolution.Value < subscription.Resolution)
+            // Both operands are compared as durations so that subscriptions declaring an explicit bar period are
+            // validated against the real period rather than the one implied by the resolution.
+            if (period.HasValue && period.Value < subscription.Increment ||
+                resolution.HasValue && resolution.Value.ToTimeSpan() < subscription.Increment)
             {
                 throw new ArgumentException($"Unable to create {symbol} consolidator because {symbol} is registered for " +
                     Invariant($"{subscription.Resolution.ToStringInvariant()} data. Consolidators require higher resolution data to produce lower resolution data.")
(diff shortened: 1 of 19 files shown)


Martin-Molinero:
> barPeriod: TimeSpan.FromMinutes(30)

Sorry this isn't something we support atm, you should be able to use custom data, or just hourly resolution instead

PR #9639, opened 2026-07-27
Add margin-aware option strategy match selection

### LINKED ISSUE #9638: Greedy option-strategy grouping charges naked-call margin on fully-covered debit spread books; inconsistent combo order buying power checks and silent cash insolvency in backtests
#### Expected Behavior

A portfolio that holds only long call debit spreads (every short call covered 1:1 by a long call at a lower strike, same expiry) should require no maintenance margin beyond the premium already paid. `OptionStrategyPositionGroupBuyingPowerModel` gets this right for a single group: Bull Call Spread margin is `Max(longStrike - shortStrike, 0) = 0` plus the net debit as initial margin.

Consequently, opening one more small debit spread against such a portfolio should only require its net debit (~$250 for a $5-wide ATM SPY spread), through both `ComboMarketOrder` and two individual `MarketOrder` legs, and both order routes should be charged the same buying power.

#### Actual Behavior

When several overlapping debit spreads accumulate on the same underlying/expiry (e.g. a new ATM spread opened daily, so strikes overlap between days), `OptionStrategyPositionGroupResolver` re-resolves the whole per-underlying book on every order check and frequently carves it into higher-leg-count strategies instead of the minimal-margin pairing. Economically-covered shorts then land in groups that charge naked-call margin:

- A $247 net-debit 1-contract spread order is rejected with `Insufficient buying power to complete orders (Value:[528,-281]), Reason: ..., Maintenance Margin Delta: 12059.6, Free Margin: 8418.3` — ~24-49x the true risk of the trade (12,059.6 ≈ 20% × underlying value, i.e. one naked short SPY call).
- The same structure, same sizing, is sometimes accepted and sometimes rejected on different days (the greedy match outcome depends on which strikes overlap that day), so the check is inconsistent.
- `Portfolio.TotalMarginUsed` also adopts the mis-grouped value (tens of thousands of dollars for a book whose optimal grouping needs $0), which later triggers spurious margin-call liquidations.
- Because on many days the phantom margin does *not* appear, the account can keep buying spreads while `Portfolio.Cash` goes deeply negative (buying power is gated on `MarginRemaining` = TPV-based, never on cash), ending with insolvent backtests (e.g. -111% drawdown on a $30k account) instead of clean rejections.

#### Onset: correct while spreads are identical, wrong from the first overlapping pair

Instrumented runs (re-running the `GetReservedBuyingPowerImpact` contemplation inside the algorithm before each order) show the defect activating at a precise, minimal point — identically in both order routes:

- **2025-01-02 → 01-27:** every order is charged correctly. The first order sees an empty book (delta = its $290 premium); subsequent same-strike orders merge into one Bull Call Spread ×N group, margin 0, delta = premium only ($108–$301).
- **2025-01-28 — first day two different-strike spreads coexist:** held (598L/603S), new order (600L/605S). Optimal grouping is two spreads, $0 margin. The resolver instead produces `Bull Call Ladder (600L, 603S, 605S)` = **$12,257** + orphan long 598 → delta 12,552. The order still fills (free margin was ~$32k), but `TotalMarginUsed` jumps to ~$12.8k — the book's margin accounting is corrupted from the first overlapping pair onward, two spreads being sufficient to trigger it.
- **2025-01-28 → 01-31 churn:** as one more spread is added each day, the re-grouping flip-flops and reserved margin oscillates $0 → $12,758 → $12,770 → **$100** → $12,306 (a 2025-01-30 re-resolve happens to find Short Butterfly + two spreads = $100, then next day it's back to a ladder), producing alternating deltas of ±$11-12.5k for identical $250-debit trades.
- **2025-02-07 — first rejection**, once free margin has shrunk below the phantom delta. The full contemplated breakdown that day — the book held 10 long / 8 short Feb-21 calls, every short coverable by a strictly-lower long (optimal margin $0):

```
GROUP [Butterfly Call]        margin=0.0      (598@1  | 603@-2 | 608@1)
GROUP [Short Butterfly Call]  margin=100.0    (603@-1 | 604@2  | 605@-1)
GROUP [Bull Call Ladder]      margin=12403.6  (600@1  | 605@-1 | 609@-1)
GROUP [Bull Call Ladder]      margin=11812.6  (600@1  | 609@-1 | 613@-1)
GROUP [Bull Call Ladder]      margin=11812.6  (608@1  | 609@-1 | 613@-1)
GROUP [Naked Call]            margin=0.0      (598@2)
GROUP [Naked Call]            margin=0.0      (604@1)
```

contemplated = 36,128.8 vs current reserved = 24,316.2 → delta 12,059.6 → the $247 spread is rejected.

#### Combo vs individual legs

Both routes are charged identically (and correctly) until 2025-01-28, and both hit the same mis-grouping from then on. Their *outcomes* diverge at the first rejection: `ComboMarketOrder` rejects both legs atomically, while with individual `MarketOrder` legs the long has already filled when the short is rejected, leaving an unpaired long in the book. Over the same 9-month run (2025-01-01 → 2025-10-01, $30k):

| | ComboMarketOrder | individual MarketOrder legs |
| --- | --- | --- |
| total orders | 262 | 247 |
| rejected legs | 74 | 101 |
| margin-call liquidations | 3 | 5 |
| net profit | 43.7% | 48.9% |

With individual legs there is also the purest symptom: a *buy* of a single long call (max risk = its $544 premium) rejected with `Maintenance Margin Delta: 12135.5` — the delta is entirely a re-grouping artifact of positions already held, not the order's own risk.

#### Root Cause

The matcher never considers margin when choosing among valid groupings:

1. **The grouping is greedy by leg count and never looks at margin.** `OptionStrategyPositionGroupResolver.GetPositionGroups` calls `OptionStrategyMatcher.MatchOnce` (`Common/Securities/Option/StrategyMatcher/OptionStrategyMatcher.cs:47-72`), which walks the definitions in `DescendingByLegCountOptionStrategyDefinitionEnumerator` order (the `ForDefinitions` default, `OptionStrategyMatcherOptions.cs:91`), matching one strategy at a time and removing its legs — 3-leg ladders/butterflies are therefore consumed before 2-leg spreads are ever tried, and the `IOptionStrategyMatchObjectiveFunction` scoring hook is not consulted on this path at all. The in-code comment already acknowledges the gap: *"a better implementation would be to evaluate all possible matches and make decisions prioritizing positions that would require more margin if not matched"* (`OptionStrategyMatcher.cs:54-56`). Once the ladder is matched, `GetCallLadderMargin` charges `OptionMarginModel` naked margin ≈ 20% of the underlying for its uncovered short (`Common/Securities/Option/OptionStrategyPositionGroupBuyingPowerModel.cs:628`).
2. **Every order re-groups the whole book.** `OptionStrategyPositionGroupResolver.GetImpactedGroups` (`Common/Securities/Positions/OptionStrategyPositionGroupResolver.cs:157-172`) matches by underlying, so each order check re-resolves the entire per-underlying book; combined with (1), the buying-power result depends on which strikes happen to overlap that day — hence the inconsistency and the day-to-day churn of `TotalMarginUsed`.
3. **No cash gate.** Independent of grouping, `PositionGroupBuyingPowerModel.HasSufficientBuyingPowerForOrder` (`Common/Securities/Positions/PositionGroupBuyingPowerModel.cs:184-194`) only compares the margin delta with `MarginRemaining`; debit premiums are never checked against cash, so on the days the phantom margin does not appear the account runs `Portfolio.Cash` to large negative values with no margin call — silent insolvency instead of a rejection.

#### Proposed Change

Make the grouping margin-aware: pick the partition of the book that minimizes total required margin (groups' margin plus per-security margin of the unmatched remainder), which is also what real brokers' margin-optimization does — exactly the direction the existing `OptionStrategyMatcher.cs:54-56` comment sketches. The scoring hook for candidate evaluation already exists (`IOptionStrategyMatchObjectiveFunction.ComputeScore`, `Common/Securities/Option/StrategyMatcher/IOptionStrategyMatchObjectiveFunction.cs:28`), but `MatchOnce` — the path the position-group resolver uses — bypasses it. Practical caveats:

- **Cost:** margin-per-candidate inside a combinatorial search bounded by `maximumDuration` is expensive. A cheap heuristic fixes this class of case without full optimization: match risk-defined 2-leg spreads before naked-short-bearing groups (ladders, short butterflies) when both can cover the same shorts — effectively reversing today's descending-leg-count preference for short-containing strategies.
- **Context:** `ComputeScore` only sees `OptionPositionCollection`s; computing margin needs security prices/models, so a margin-aware objective (or `MatchOnce` itself) needs portfolio context threaded in.

Separately, (3) above deserves its own guard: debit premiums should be payable from settled cash regardless of how the book is grouped.

#### Reproduction

Minimal algorithm (C#, cloud defaults, no custom models): from 2025-01-01 with $30,000, each trading day open one 1-contract ATM $5-wide SPY bull call spread via `ComboMarketOrder` (nearest expiry 20-45 DTE), never close. The first overlapping-strike day (~4 weeks in) triggers the mis-grouping; within ~5 weeks orders are rejected with 5-figure `Maintenance Margin Delta` values while equivalent trades on other days fill; by late February cash is ~-$30k and margin-call liquidations of auto-exercised SPY shares begin.

```csharp
public class ComboSpreadMarginRepro : QCAlgorithm
{
    private Symbol _option;
    private DateTime _last;
    {
        SetStartDate(2025, 1, 1);
        SetEndDate(2025, 10, 1);
        SetCash(30000);
        var option = AddOption("SPY", Resolution.Minute);
        _option = option.Symbol;
        option.SetFilter(u => u.Strikes(-10, 10).Expiration(20, 45));
    }
    {
        if (Time.Date == _last.Date || !slice.OptionChains.TryGetValue(_option, out var chain)) return;
        var calls = chain.Where(c => c.Right == OptionRight.Call).ToList();
        if (calls.Count == 0) return;
        var expiry = calls.Min(c => c.Expiry);
        var atExpiry = calls.Where(c => c.Expiry == expiry).OrderBy(c => c.Strike).ToList();
        var longCall = atExpiry.OrderBy(c => Math.Abs(c.Strike - chain.Underlying.Price)).FirstOrDefault();
        var shortCall = atExpiry.FirstOrDefault(c => c.Strike == longCall.Strike + 5m);
        if (longCall == null || shortCall == null) return;
        _last = Time;
        ComboMarketOrder(new List<Leg> { Leg.Create(longCall.Symbol, 1), Leg.Create(shortCall.Symbol, -1) }, 1);
    }
}
```

Support ref: Intercom conversation 215475229332236.

AlexCatarino:
#### Description

`OptionStrategyMatcher.MatchOnce` greedily matched strategy definitions in descending leg-count order and never consulted the `IOptionStrategyMatchObjectiveFunction` hook (the long-standing TODO at the top of the class). Because every buying power check re-resolves the whole per-underlying option book through this greedy matching, a portfolio holding only fully-covered debit spreads could get carved into groups containing uncovered short legs, which are charged naked option margin.

Minimal case (from our reproduction, daily 1-contract ATM $5-wide SPY bull call spreads via `ComboMarketOrder`): holding `598C +1, 603C -1` and ordering `600C +1, 605C -1` (same expiry, SPY ≈ 600) re-groups the combined book into a Bull Call Ladder `(600, 603, 605)` plus an orphan `598C` long. The ladder's second short is charged naked call margin ≈ premium + 20% × underlying ≈ $12k, so a ~$295 net-debit, defined-risk order is rejected with `Maintenance Margin Delta` ≈ $12.5k. The correct grouping — two Bull Call Spreads — requires no margin beyond the premium, and LEAN's own spread margin formulas agree.

Consequences observed: inconsistent accept/reject on identical trades, day-to-day `TotalMarginUsed` churn ($0 → $12,758 → $100 → $12,306 on consecutive days) as re-resolution flips between carvings, spurious margin calls, and even long-only option buys (max risk = premium) rejected with ~$12k deltas that were purely re-grouping artifacts.

**The change** (implements the matcher's TODO):

- `MatchOnce` evaluates a fixed set of two candidate solutions and selects via the configured objective function:
  1. The legacy greedy match over the configured definition order — evaluated first and kept on ties, so every currently-correct grouping is unchanged.
  2. The same greedy match with definitions that leave a short leg uncovered within the strategy (naked calls/puts, ladders, short backspreads/straddles/strangles) deprioritized, so shorts are matched into covered strategies whenever the positions allow it.
- New default objective function `UncoveredShortQuantityOptionStrategyMatchObjectiveFunction`: scores a solution by the negated quantity of short option contracts left uncovered, either inside their strategy or unmatched. Uncovered shorts are charged naked option margin, typically an order of magnitude above any risk-defined strategy margin, which makes this a cheap proxy for total margin that doesn't require security prices inside the matcher. The previous default, `UnmatchedPositionCountOptionStrategyMatchObjectiveFunction`, was never consulted anywhere and remains available.
- **Coverage is strike- and expiry-aware, so the selection can only ever lower margin**, mirroring what the margin models actually honor:
  - A long on the debit side of the short strike (lower for calls, higher for puts) covers for free.
  - A long on the credit side caps the risk at the strike width instead, which for a distant long can exceed the naked charge — short 600P covered by a long 100P is a $50k spread versus ~$12k naked. Credit-side coverage therefore only counts within 10% of the short strike, the price-free stand-in for the naked margin floor of 10% of underlying value in `OptionMarginModel`.
  - A covering long must also outlive the short. A long expiring first leaves the short naked for the rest of its life, which is exactly why `OptionStrategyPositionGroupBuyingPowerModel` charges short calendar spreads the stand-alone naked short margin while ordinary calendar spreads require none. Without this the scorer read a short calendar's zero strike width as full coverage.
  - Where coverage isn't credited, the candidates tie and the previous grouping is kept.

**Why not just reorder definitions** (spreads before ladders): a true butterfly book (+1 low, −2 mid, +1 high) would then decompose into a bull call spread plus a bear call spread, charging the bear spread's strike width where the butterfly requires none. Descending-leg-count is sometimes the cheaper carve, so the fix compares solutions instead of hard-coding an order; a butterfly regression test covers this.

**Intentionally untouched** (root-cause items 2 and 3 of the issue, kept out to focus this PR):
- Premiums are still not gated on cash in `PositionGroupBuyingPowerModel.HasSufficientBuyingPowerForOrder` (compares only against `MarginRemaining`).
- `OptionStrategyPositionGroupResolver.GetImpactedGroups` still matches by underlying; it amplified this bug but is semantically defensible, and with margin-aware selection it becomes harmless.

**Pre-existing behavior worth flagging** (not introduced here, not addressed here): the matcher's result is not stable across processes. Greedy matching picks the first of several equally valid matches, and that order depends on enumeration over the symbol-keyed immutable collections in `OptionPositionCollection`, which varies with .NET's per-process randomized string hashing. Digesting the *pre-change* single-pass grouping of 15,600 multi-expiry books across four processes produced three different digests, so this predates the change; the two-candidate selection inherits it rather than causing it. It shows up mainly in books admitting several valid carves. Tracked separately in #9648, to be fixed after this merges.

#### Related Issue

Fixes #9638

#### Motivation and Context

Defined-risk option spread books were charged phantom naked-call margin, causing rejected orders, inconsistent margin usage, and spurious margin-call liquidations in backtesting and live. See the issue for the full analysis.

None.

**Behavioral change surface.** We swept 4,320 synthetic books (calls-only, puts-only, mixed, and with underlying lots at ±1/±2), running the previous algorithm and the new one on each and diffing the resulting groupings. 210 books change (4.9%), across ~100 distinct transitions, and **every transition drops a Bull Call Ladder or a Bear Put Ladder** — those two are the only strategies that ever disappear, always replaced by covered or risk-defined groups (bull/bear call and put spreads, covered calls, protective calls). Butterflies, iron condors, iron butterflies, straddles, strangles, backspreads, box spreads, calendar spreads, jelly rolls, conversions and protective collars are never affected; neither are Bear Call Ladder and Bull Put Ladder, whose net option position isn't short. (The exact transition count varies by ±1 between runs because of the pre-existing non-determinism noted above; the changed-book count and the ladders-only property are stable.)

A second sweep over 15,600 multi-expiry books, scoring every candidate with an expiry-aware uncovered count, found no book where the second candidate wins yet requires more real margin than the previous behavior.

**Performance.** Benchmarked against master per `MatchOnce` call (interleaved rounds, min of 5, alternating order; ±10% between processes):

| book | master | this PR | ratio |
|---|---|---|---|
| single spread | 4.1ms | 3.9ms | 0.93x |
| naked shorts | 7.0ms | 6.8ms | 0.96x |
| butterfly | 5.5ms | 5.1ms | 0.93x |
| iron condor | 9.7ms | 9.3ms | 0.96x |
| ladder | 14.9ms | 14.5ms | 0.98x |
| overlapping spreads | 27.3ms | 57.2ms | 2.10x |
| multi-expiry book | 29.7ms | 71.8ms | 2.42x |
| wide 8-leg book | 137.3ms | 271.8ms | 1.98x |

Books where the second pass doesn't run are slightly faster than master, since the definition ordering is now cached instead of being re-sorted on every call. The cost is confined to books where the second pass genuinely runs and changes the answer — exactly the books that were producing wrong margin before — and is bounded by the second pass itself, hence ~2x. Two optimizations keep that surface small: matching again is skipped when the first solution already leaves no more shorts uncovered than the positions can possibly cover (a long covers at most its own quantity of same-right shorts, and so does an underlying lot), which takes a plain ladder from 2.2x down to parity; and scoring takes an allocation-free fast path for every strategy with a single short leg, which is every spread, butterfly, condor, backspread and covered call. That bound reads the score as a quantity of uncovered contracts, which only the default objective function guarantees, so a custom one always gets both candidates evaluated.

**Tests.** New unit tests in `Tests/Common/Securities/Options/StrategyMatcher/OptionStrategyMatcherTests.cs`:
- `{598C +1, 600C +1, 603C -1, 605C -1}` (same expiry) matches as two Bull Call Spreads, no ladder, no leftovers.
- An interleaved book (longs 598×3, 600×2, 604×3, 608×2; shorts 603×3, 605×2, 609×2, 613×1) leaves no short contract uncovered.
- A true butterfly book still matches as Butterfly Call (guards against naive definition reordering).
- A real ladder book still matches as Bull Call Ladder (ties preserve previous behavior).
- A short call with only a distant long available stays a ladder rather than becoming a wide bear call spread, and the put-side equivalent.
- A short calendar spread's short leg scores as uncovered, and an ordinary calendar spread's does not, both rights.
- Underlying lots cover a covered call's short leg.
- A custom objective function is offered every candidate rather than being short-circuited by the default's bound.

New in `Tests/Common/Securities/OptionStrategyPositionGroupBuyingPowerModelTests.cs`: holding two overlapping bull call spreads resolves as two Bull Call Spread groups with `TotalMarginUsed == 0`; ordering an overlapping spread with cash covering only its net debit passes `HasSufficientBuyingPowerForOrder` (used to require ~$12k); a long-only call order against that book passes with cash covering only its premium.

New regression algorithm `OptionEquityOverlappingBullCallSpreadsRegressionAlgorithm` opens two overlapping debit spreads with interleaved strikes and asserts two Bull Call Spread groups, zero margin used, and margin-remaining accounting.

Every test above was verified to fail against the pre-change code (the regression algorithm ends in `RuntimeError`: *Expected two Bull Call Spread groups, found 0: Bull Call Ladder, Naked Call*), except the ones asserting behavior is preserved; those were verified by mutation instead — removing the credit-side width bound or the expiry guard makes them fail.

`dotnet build QuantConnect.Lean.sln -c Release` builds clean. Green suites: ~3,300 tests matching `StrategyMatcher|OptionEquity|PositionGroup|OptionStrategyPositionGroupBuyingPowerModel`, plus a 3,688-test run over `Strateg|MarginCall|ComboOrder`. A full `Tests` run was not executed locally.

#### Types of changes
- [x] Performance (see the benchmark table above)

#### Checklist:
- [ ] My branch follows the naming convention `bug-<issue#>-<description>`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Algorithm.CSharp/OptionEquityOverlappingBullCallSpreadsRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionEquityOverlappingBullCallSpreadsRegressionAlgorithm.cs
@@ -1,0 +1,142 @@
+/*
+ * QUANTCONNECT.COM - Democratizing Finance, Empowering Individuals.
+ * Lean Algorithmic Trading Engine v2.0. Copyright 2014 QuantConnect Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+using System;
+using System.Linq;
+using QuantConnect.Data;
+using QuantConnect.Data.Market;
+using System.Collections.Generic;
+using QuantConnect.Securities.Option;
+using QuantConnect.Securities.Positions;
+using QuantConnect.Securities.Option.StrategyMatcher;
+
+namespace QuantConnect.Algorithm.CSharp
+{
+    /// <summary>
+    /// Regression algorithm asserting that a book of two overlapping bull call debit spreads, with interleaved
+    /// strikes and the same expiration, is grouped as two margin-free bull call spreads. The greedy leg-count
+    /// descending matching used to carve this book into a bull call ladder plus an unmatched long, charging
+    /// naked call margin for the ladder's uncovered short leg on a fully covered, defined-risk book
+    /// </summary>
+    public class OptionEquityOverlappingBullCallSpreadsRegressionAlgorithm : OptionEquityBaseStrategyRegressionAlgorithm
+    {
+        /// <summary>
+        /// OnData event is the primary entry point for your algorithm. Each new data point will be pumped in here.
+        /// </summary>
+        /// <param name="slice">Slice object keyed by symbol containing the stock data</param>
+        public override void OnData(Slice slice)
+        {
+            if (!Portfolio.Invested)
+            {
+                OptionChain chain;
+                if (IsMarketOpen(_optionSymbol) && slice.OptionChains.TryGetValue(_optionSymbol, out chain))
+                {
+                    var callContracts = chain
+                        .Where(contract => contract.Right == OptionRight.Call);
+                    var expiry = callContracts.Min(x => x.Expiry);
+                    var contracts = callContracts.Where(x => x.Expiry == expiry)
+                        .DistinctBy(x => x.Strike)
+                        .OrderBy(x => x.Strike)
+                        .ToList();
+                    if (contracts.Count < 4) return;
+
+                    var initialMargin = Portfolio.MarginRemaining;
+
+                    // first debit spread: long the lowest strike, short the second highest
+                    MarketOrder(contracts[0].Symbol, 1);
+                    MarketOrder(contracts[2].Symbol, -1);
+
+                    AssertOptionStrategyIsPresent(OptionStrategyDefinitions.BullCallSpread.Name, 1);
+
+                    // second debit spread, overlapping the first: long the second lowest strike, short the highest
+                    MarketOrder(contracts[1].Symbol, 1);
+                    MarketOrder(contracts[3].Symbol, -1);
+                    var freeMarginPostTrade = Portfolio.MarginRemaining;
+
+                    // every short call is covered by a long call at a lower strike: the book must resolve into two
+                    // margin-free bull call spreads, not a bull call ladder charging naked call margin plus an orphan long
+                    var bullCallSpreadsCount = Portfolio.Positions.Groups.Count(group =>
+                        group.BuyingPowerModel is OptionStrategyPositionGroupBuyingPowerModel
+                        && group.BuyingPowerModel.ToString() == OptionStrategyDefinitions.BullCallSpread.Name);
+                    if (bullCallSpreadsCount != 2)
+                    {
+                        throw new RegressionTestException($"Expected two Bull Call Spread groups, found {bullCallSpreadsCount}: " +
+                            string.Join(", ", Portfolio.Positions.Groups.Select(group => group.BuyingPowerModel.ToString())));
+                    }
+
+                    var expectedMarginUsage = 0m;
+                    if (expectedMarginUsage != Portfolio.TotalMarginUsed)
+                    {
+                        throw new RegressionTestException($"Unexpected margin used!: {Portfolio.TotalMarginUsed}");
+                    }
+
+                    // we paid the ask and value using the assets price
+                    var priceSpreadDifference = GetPriceSpreadDifference(contracts[0].Symbol, contracts[1].Symbol,
+                        contracts[2].Symbol, contracts[3].Symbol);
+                    if (initialMargin != (freeMarginPostTrade + expectedMarginUsage + _paidFees - priceSpreadDifference))
+                    {
+                        throw new RegressionTestException("Unexpected margin remaining!");
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Data Points count of all timeslices of algorithm
+        /// </summary>
+        public override long DataPoints => 15023;
+
+        /// <summary>
+        /// Data Points count of the algorithm history
+        /// </summary>
+        public override int AlgorithmHistoryDataPoints => 0;
+
+        /// <summary>
+        /// This is used by the regression test system to indicate what the expected statistics are from running the algorithm
+        /// </summary>
+        public override Dictionary<string, string> ExpectedStatistics => new Dictionary<string, string>
+        {
+            {"Total Orders", "4"},
+            {"Average Win", "0%"},
+            {"Average Loss", "0%"},
+            {"Compounding Annual Return", "0%"},
+            {"Drawdown", "0%"},
+            {"Expectancy", "0"},
+            {"Start Equity", "200000"},
+            {"End Equity", "199756"},
+            {"Net Profit", "0%"},
+            {"Sharpe Ratio", "0"},
+            {"Sortino Ratio", "0"},
+            {"Probabilistic Sharpe Ratio", "0%"},
+            {"Loss Rate", "0%"},
+            {"Win Rate", "0%"},
+            {"Profit-Loss Ratio", "0"},
+            {"Alpha", "0"},
+            {"Beta", "0"},
+            {"Annual Standard Deviation", "0"},
+            {"Annual Variance", "0"},
+            {"Information Ratio", "0"},
+            {"Tracking Error", "0"},
+            {"Treynor Ratio", "0"},
+            {"Total Fees", "$4.00"},
+            {"Estimated Strategy Capacity", "$65000.00"},
+            {"Lowest Capacity Asset", "GOOCV W78ZERHAT67A|GOOCV VP83T1ZUHROL"},
+            {"Portfolio Turnover", "2.85%"},
+            {"Drawdown Recovery", "0"},
+            {"OrderListHash", "4f2d6ca65efe107133bf6baff5fe5512"}
+        };
+    }
+}
--- a/Algorithm.CSharp/OptionEquityOverlappingBullCallSpreadsRegressionAlgorithm.cs
+++ b/Algorithm.CSharp/OptionEquityOverlappingBullCallSpreadsRegressionAlgorithm.cs
@@ -55,13 +55,13 @@ namespace QuantConnect.Algorithm.CSharp
 
                     var initialMargin = Portfolio.MarginRemaining;
 
-                    // first debit spread: long the lowest strike, short the second highest
+                    // first debit spread: long the lowest strike, short the third lowest
                     MarketOrder(contracts[0].Symbol, 1);
                     MarketOrder(contracts[2].Symbol, -1);
 
                     AssertOptionStrategyIsPresent(OptionStrategyDefinitions.BullCallSpread.Name, 1);
 
-                    // second debit spread, overlapping the first: long the second lowest strike, short the highest
+                    // second debit spread, its strikes interleaved with the first: long the second lowest, short the fourth lowest
                     MarketOrder(contracts[1].Symbol, 1);
                     MarketOrder(contracts[3].Symbol, -1);
                     var freeMarginPostTrade = Portfolio.MarginRemaining;
--- a/Common/Securities/Option/StrategyMatcher/IOptionStrategyMatchObjectiveFunction.cs
+++ b/Common/Securities/Option/StrategyMatcher/IOptionStrategyMatchObjectiveFunction.cs
@@ -21,9 +21,9 @@ namespace QuantConnect.Securities.Option.StrategyMatcher
     public interface IOptionStrategyMatchObjectiveFunction
     {
         /// <summary>
-        /// Evaluates the objective function for the provided match solution. Solution with the highest score will be selected
-        /// as the solution. NOTE: This part of the match has not been implemented as of 2020-11-06 as it's only evaluating the
-        /// first solution match (MatchOnce).
+        /// Evaluates the objective function for the provided match solution. The solution with the highest score will be
+        /// selected as the solution. By convention, solutions that can't be improved upon score zero, the maximum, which
+        /// allows the matcher to skip evaluating additional candidate solutions.
         /// </summary>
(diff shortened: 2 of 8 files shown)


jhonabreul (approved):

AlexCatarino:
The matcher non-determinism noted in the description is now filed separately as #9648, with a minimal reproduction: a same-expiration book of `+1 595C, +1 600C, -1 605C` pairs the short with either long depending on the process, because `TryMatchOnce` takes the first of several valid matches and that order comes from hashing symbols, which .NET randomizes per process.

It predates this PR — digesting the pre-change single greedy pass over 15,600 books across four processes gave three different digests — so nothing here needs to change for it. Keeping it out of this PR to leave the diff focused on the margin defect; happy to pick it up once this one merges.

Martin-Molinero (approved):
Nice 👍

PR #9644, opened 2026-07-29
Add results analyses detecting the isolator time limit runtime errors

jhonabreul:
#### Description

Adds a new `RuntimeErrorAnalysis` family of backtest results analyses that detect algorithms terminated by the execution time limits enforced by the `Isolator`, and give the user actionable recommendations to avoid the error. The runtime error message is read from the result state (`State["RuntimeError"]`), falling back to the "Runtime Error:" log line for results that carry no state.

Three analyses, each detecting one termination reason (at most one can fire per run, all ranked as fatal with weight 100):

- `SingleTimeLoopTimeoutRuntimeErrorAnalysis` — "Algorithm took longer than N minutes on a single time loop", emitted by `AlgorithmTimeLimitManager`/`Isolator` when a single time loop exceeds the per-loop limit. Recommends: moving heavy per-event work to incremental computation (rolling windows, consolidators, indicators), reducing and filtering universes (including option/future chain `SetFilter` filters), lowering subscription counts and resolution, avoiding large history requests inside event handlers, running ML training through `Train()` (which allocates additional time), checking for infinite loops, and vectorizing with numpy/pandas for Python algorithms.
- `OperationCanceledRuntimeErrorAnalysis` — "Operation was canceled", emitted by the `Isolator` when the algorithm is asked to stop (user stop/delete) but its code keeps running past the shutdown grace period. Explains that the error is expected if the run was stopped on purpose, and otherwise recommends reducing per-event work so control returns to the engine promptly.
- `MaximumRuntimeExceededRuntimeErrorAnalysis` — the overall maximum runtime timeout, matching both the "Execution Security Error: Operation timed out - N minutes max" and "Failed to complete algorithm within N seconds" message shapes. Recommends shortening the backtest period, reducing universe size/resolution, and reviewing code inefficiencies.

The abstract `RuntimeErrorAnalysis` base owns the shared plumbing (runtime error extraction and fragment-set message matching), so future runtime-error analyses (e.g. out-of-memory) only declare their patterns, issue and solutions. The three analyses are registered in the final `ResultsAnalyzer` chain only: these errors terminate the run, so there is nothing for the in-run analyzer to catch while the backtest is alive.

Note: this PR is stacked on #9632 (`feature-in-run-backtest-analysis`); the diff includes its commits until it is merged. The new analysis work is contained in the last commit.

#### Related Issue

N/A

#### Motivation and Context

When a backtest dies with "Operation was canceled" or "Algorithm took longer than 10 minutes on a single time loop", the message alone doesn't tell the user (or an LLM reading the results) what caused it or how to fix it. These analyses surface the common causes — big loops in event handlers, oversized or unfiltered universes, too many securities, large history requests, untimed model training — as concrete recommendations in the results analysis findings.

N/A

- New `RuntimeErrorAnalysesTests` unit tests covering: each error message shape being flagged by exactly its own analysis and ignored by the other two, detection from the state and from the logs fallback, no false positives on non-runtime-error log lines (e.g. the Isolator's own trace messages) or unrelated runtime errors, clean runs, and Python snake_case formatting of code references in the solutions.
- Full analysis test suite (`*AnalysisTests`, `ResultsAnalyzer*`) passing locally.

#### Types of changes

#### Checklist:

### DIFF
--- a/Engine/Engine.cs
+++ b/Engine/Engine.cs
@@ -104,12 +104,14 @@ namespace QuantConnect.Lean.Engine
                 //-> Initialize messaging system
                 SystemHandlers.Notify.SetAuthentication(job);
 
+                var performanceTrackingTool = new PerformanceTrackingTool();
+
                 //-> Set the result handler type for this algorithm job, and launch the associated result thread.
-                AlgorithmHandlers.Results.Initialize(new(job, SystemHandlers.Notify, SystemHandlers.Api, AlgorithmHandlers.Transactions, AlgorithmHandlers.MapFileProvider));
+                AlgorithmHandlers.Results.Initialize(
+                    new(job, SystemHandlers.Notify, SystemHandlers.Api, AlgorithmHandlers.Transactions, AlgorithmHandlers.MapFileProvider, performanceTrackingTool));
 
                 IBrokerage brokerage = null;
                 DataManager dataManager = null;
-                var performanceTrackingTool = new PerformanceTrackingTool();
                 var synchronizer = _liveMode ? new LiveSynchronizer() : new Synchronizer();
                 try
                 {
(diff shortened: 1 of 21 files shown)


jhonabreul:
Moving these changes into #9632 directly.

PR #9651, opened 2026-08-03
Fix live currency conversion seeding after warm-up clock rewind

### LINKED ISSUE #9647: Live: StartingPortfolioValue is captured before warm-up replays, mixing deploy-time holdings prices with warm-up-start FX conversion rates
## Expected Behavior

In a live deployment, `ISetupHandler.StartingPortfolioValue` — and the initial reported
Equity derived from it — should reflect the account's real value **at deploy time**:
current holdings prices converted at current FX rates.

## Actual Behavior

When the algorithm declares a warm-up period, `BrokerageSetupHandler.Setup()` captures
`StartingPortfolioValue` from a **mixed-epoch** snapshot:

- holdings are loaded and priced from **current** data (`LoadExistingHoldingsAndOrders`
  runs before the clock is rewound), but
- the cash book's currency conversion rates are seeded **point-in-time at the warm-up
  start** (the clock has already been rewound by `PostInitialize()`), and
- the capture happens **before warm-up replays any data**, so nothing has had a chance
  to bring the rates forward yet.

The resulting `TotalPortfolioValue` mixes deploy-time holdings prices with an FX rate
that can be years old, and that value is captured once and never re-captured for the
life of the deployment.

To be clear about what is **not** the bug: seeding the conversion rates point-in-time at
the warm-up start is correct by design (during warm-up the algorithm should see
epoch-consistent rates), and the warm-up feed **does** advance them — the internal
currency subscription is created with the warm-up start date and gets a warmup
enumerator like any other subscription (`LiveTradingDataFeed.GetWarmupEnumerator`
concatenates the file-based enumerator with a history-provider tail capped by
`MaximumWarmupHistoryDaysLookBack`). In the observed deploy the rates were ~current
within seconds of warm-up starting. The defect is purely **when the starting-value
snapshot is taken**.

Observed on a live IB deployment with an SGD base currency, USD-denominated holdings and
a 700-bar daily warm-up (deployed 2026-07-30):

```
TRACE:: Cash.EnsureCurrencyDataFeed(): Adding USDSGD oanda for cash USD currency feed
TRACE:: InfluxDbHistoryProvider.GetHistory(): USDSGD 8G Forex 2023-10-12T03:55:00Z to 2023-10-12T04:00:00Z. QuoteBar. Quote. Minute
TRACE:: BaseSetupHandler.SetupCurrencyConversions():
Account Type: Margin

Symbol      Quantity    Conversion = Value in SGD
USD: $      <redacted> @     1.3623 = <redacted>
```

The seeded USDSGD rate is `1.3623` — the 2023-10-12 quote. The live rate at deploy time
was `1.2866`, so every USD-denominated term in the snapshot is overstated by 5.9% (~5%
of total portfolio value on this account). Meanwhile the *holdings* prices in the same
snapshot were current:

```
TRACE:: InfluxDbHistoryProvider.GetHistory(): <TICKER> Equity 2026-07-29T19:55:00Z to 2026-07-30T10:17:40Z. QuoteBar. Quote. Minute
```

Warm-up itself completed in ~8 seconds and the algorithm's own
`Portfolio.TotalPortfolioValue` was correct from then on — only the captured
`StartingPortfolioValue` (and everything derived from it) kept the stale-rate value.

## Root cause

`QCAlgorithm.PostInitialize()` rewinds the algorithm clock to the warm-up start:

```csharp
if (TryGetWarmupHistoryStartTime(out var result))
{
    SetDateTime(result.ConvertToUtc(TimeZone));
}
```

`BrokerageSetupHandler.Setup()` calls `PostInitialize()`, then seeds the conversion
rates, then immediately captures the starting portfolio value
(`Engine/Setup/BrokerageSetupHandler.cs`):

```csharp
algorithm.PostInitialize();                                   // clock -> warm-up start

BaseSetupHandler.SetupCurrencyConversions(algorithm, parameters.UniverseSelection);
...
StartingPortfolioValue = algorithm.Portfolio.TotalPortfolioValue;
```

`SetupCurrencyConversions` → `AlgorithmUtils.SeedCurrencyConversionRates` →
`AlgorithmUtils.SeedSecurities` → `QCAlgorithm.GetLastKnownPrices`, whose bar-count
history requests end at the algorithm's current `Time` — the warm-up start. That is
fine as a warm-up seed; it is not a valid deploy-time value, and it is captured as one.

So the further back the warm-up reaches, the staler the FX component of the captured
snapshot. With a short warm-up the error is invisible; with a multi-year warm-up it is
arbitrarily large.

## Impact

1. `StartingPortfolioValue` is wrong for any live algorithm that has a warm-up **and**
   holds cash/holdings in a currency other than the account currency. It is never
   re-captured, so `BaseResultsHandler.GetNetReturn()` —
   `(GetPortfolioValue() - StartingPortfolioValue) / StartingPortfolioValue` — reports a
   permanent phantom return (observed: -4.55% on an account that had not traded), and
   the first equity-curve sample seeds `CumulativeMaxPortfolioValue`, contaminating
   drawdown and every statistic derived from the equity curve.
2. The initial reported Equity shows the same mixed-epoch value.
3. It recurs on every restart (e.g. weekly maintenance), because setup re-runs.

A display quirk makes this hard to diagnose and easy to misread as "the rates never
warmed up": `LiveTradingResultHandler.UpdatePortfolioValues` only refreshes the reported
value when `UserExchangeIsOpen(time)` is true, and the internal currency-conversion
feeds are filtered out of `SecurityChanges.AddedSecurities`, so they never set
`_sampleChartAlways`. The displayed Equity therefore stays frozen at the bad setup-time
snapshot until the first regular-hours session begins — hours after warm-up has already
corrected the live rates — then snaps to the correct number. (Whether the reported
Equity should stay frozen pre-open at all is arguably a separate issue.)

## Proposed change

Re-capture `StartingPortfolioValue` (and the initial equity sample) **after warm-up
completes** in live mode, when both holdings prices and conversion rates are current.

Alternatives considered and why they are worse:

- Seeding the conversion rates at `DateTime.UtcNow` (either by moving
  `SetupCurrencyConversions` before `PostInitialize()` or by special-casing the history
  request end time) would fix the snapshot, but sacrifices the point-in-time consistency
  of the rates during warm-up replay, which is correct today and worth keeping.

## Reproduction

1. Live-deploy to a brokerage account whose base currency differs from the currency of
   its holdings/cash (e.g. SGD base, USD holdings).
2. Call `SetWarmUp(700, Resolution.Daily)` in `Initialize`.
3. Deploy outside the holdings' regular market hours.
4. Compare the reported Equity / net Return with `Portfolio.TotalPortfolioValue` logged
   from the algorithm, and check the seeded conversion rate in the
   `BaseSetupHandler.SetupCurrencyConversions()` trace against the live FX rate.

Reducing the warm-up to a few days makes the discrepancy disappear (the warm-up-start
seed is then near-current), which confirms the mechanism.

Internal reference: Intercom conversation 215475259372811.

Niketion:
#### Description

Fixes live currency conversion seeding when `PostInitialize()` rewinds the algorithm clock to the warm-up start time.

The change captures the deployment UTC time immediately before `PostInitialize()` and propagates that explicit reference through the initial live currency conversion seeding pipeline:

* `BrokerageSetupHandler`
* `BaseSetupHandler.SetupCurrencyConversions`
* `AlgorithmUtils.SeedCurrencyConversionRates`
* `AlgorithmUtils.SeedSecurities`
* `IAlgorithm.GetLastKnownPrices`

A backward-compatible `IAlgorithm.GetLastKnownPrices` overload accepts the explicit reference time. `QCAlgorithm` uses it for history request start and end times, retries, and endpoint filtering. The existing overload continues to use the algorithm clock, preserving normal history and backtesting behavior.

`AlgorithmPythonWrapper` explicitly forwards the reference time to its underlying `QCAlgorithm`.

A regression test covers the warm-up clock rewind, FX request endpoint, seeded conversion rate, and `StartingPortfolioValue`.

#### Related Issue

Fixes #9647

#### Motivation and Context

During live setup, `PostInitialize()` can rewind `algorithm.UtcTime` to the beginning of the configured warm-up period before currency conversion rates are seeded.

The previous flow therefore requested an FX quote at the warm-up start instead of deployment time. This could value foreign-currency balances using a stale rate and permanently distort `TotalPortfolioValue`, `StartingPortfolioValue`, net return, equity-curve statistics, and drawdown calculations.

The explicit deployment timestamp is limited to the initial live currency conversion seeding path. Normal history requests and backtesting continue to use the algorithm clock.

No.

Successfully built:

```text
dotnet build QuantConnect.Lean.sln -c Release --no-restore /p:RunAnalyzers=false
dotnet build Tests\QuantConnect.Tests.csproj -c Release --no-restore /p:RunAnalyzers=false
```

Both builds completed with zero errors.

The targeted regression test was discovered and entered its test body locally, but `BrokerageSetupHandler.Setup()` returned `false` before the timestamp and portfolio-value assertions were reached.

The local LEAN environment does not contain the complete map-file/data prerequisites required by the setup pipeline, including `Data/equity/sgx/map_files`. No fabricated data, placeholder map files, global test-initialization changes, or test bypasses were introduced.

The regression test therefore compiles but could not be completed locally. The full execution is left to CI or an environment with the complete LEAN test data.

#### Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] Refactor (non-breaking change which improves implementation)
* [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Non-functional change (xml comments/documentation/etc)

#### Checklist:

* [x] My code follows the code style of this project.
* [x] I have read the **CONTRIBUTING** document.
* [x] I have added tests to cover my changes.
* [ ] All new and existing tests passed. The solution and test project build successfully, but the targeted test could not complete locally because the required LEAN map-file/data assets are unavailable.
* [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

### DIFF
--- a/Algorithm/QCAlgorithm.History.cs
+++ b/Algorithm/QCAlgorithm.History.cs
@@ -756,13 +756,24 @@ namespace QuantConnect.Algorithm
         [DocumentationAttribute(HistoricalData)]
         public DataDictionary<IEnumerable<BaseData>> GetLastKnownPrices(IEnumerable<Symbol> symbols)
         {
+            return GetLastKnownPrices(symbols, UtcTime);
+        }
+
+        /// <summary>
+        /// Yields data to warm up multiple securities for all their subscribed data types using the specified reference time
+        /// </summary>
+        /// <param name="symbols">The symbols we want to get seed data for</param>
+        /// <param name="referenceUtcTime">The UTC time to use as the reference for the history requests</param>
+        /// <returns>Securities historical data</returns>
+        public DataDictionary<IEnumerable<BaseData>> GetLastKnownPrices(IEnumerable<Symbol> symbols, DateTime referenceUtcTime)
+        {
             if (HistoryProvider == null)
             {
                 return new DataDictionary<IEnumerable<BaseData>>();
             }
 
             var data = new Dictionary<(Symbol, Type, TickType), BaseData>();
-            GetLastKnownPricesImpl(symbols, data);
+            GetLastKnownPricesImpl(symbols, data, referenceUtcTime);
 
             return data
                 .GroupBy(kvp => kvp.Key.Item1)
@@ -808,7 +819,7 @@ namespace QuantConnect.Algorithm
         }
 
         private void GetLastKnownPricesImpl(IEnumerable<Symbol> symbols, Dictionary<(Symbol, Type, TickType), BaseData> result,
-            int attempts = 0, IEnumerable<HistoryRequest> failedRequests = null)
+            DateTime referenceUtcTime, int attempts = 0, IEnumerable<HistoryRequest> failedRequests = null)
         {
             IEnumerable<HistoryRequest> historyRequests;
             var isRetry = failedRequests != null;
@@ -818,14 +829,14 @@ namespace QuantConnect.Algorithm
             if (attempts == 0)
             {
                 historyRequests = CreateBarCountHistoryRequests(symbols, SeedLookbackPeriod,
-                    fillForward: false, useAllSubscriptions: true)
+                    fillForward: false, useAllSubscriptions: true, referenceUtcTime: referenceUtcTime)
                     .SelectMany(request =>
                     {
                         // Make open interest request daily, higher resolutions will need greater periods to return data
                         if (request.DataType == typeof(OpenInterest) && request.Resolution < Resolution.Daily)
                         {
                             return CreateBarCountHistoryRequests([request.Symbol], typeof(OpenInterest), SeedLookbackPeriod,
-                                Resolution.Daily, fillForward: false, useAllSubscriptions: true);
+                                Resolution.Daily, fillForward: false, useAllSubscriptions: true, referenceUtcTime: referenceUtcTime);
                         }
 
                         if (request.Resolution < Resolution.Minute)
@@ -837,7 +848,7 @@ namespace QuantConnect.Algorithm
                             }
 
                             return CreateBarCountHistoryRequests([request.Symbol], dataType, SeedLookbackPeriod,
-                                Resolution.Minute, fillForward: false, useAllSubscriptions: true);
+                                Resolution.Minute, fillForward: false, useAllSubscriptions: true, referenceUtcTime: referenceUtcTime);
                         }
 
                         return [request];
@@ -856,7 +867,8 @@ namespace QuantConnect.Algorithm
                         var periods = resolution == Resolution.Daily
                             ? SeedRetryDailyLookbackPeriod
                             : resolution == Resolution.Hour ? SeedRetryHourLookbackPeriod : SeedRetryMinuteLookbackPeriod;
-                        return CreateBarCountHistoryRequests([group.Key], periods, resolution, fillForward: false, useAllSubscriptions: true)
+                        return CreateBarCountHistoryRequests([group.Key], periods, resolution, fillForward: false, useAllSubscriptions: true,
+                            referenceUtcTime: referenceUtcTime)
                             .Where(request => symbolRequests.Any(x => x.DataType == request.DataType));
                     })
                     .SelectMany(x => x);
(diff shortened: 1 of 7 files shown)


Martin-Molinero:
Sorry this is a hard one being solved by the QC team at https://github.com/QuantConnect/Lean/pull/9653, thanks for the proposal though.

PR #9686, opened 2026-08-12
Clear the advance decline period watermark on reset

### LINKED ISSUE #9683: AdvanceDeclineIndicator.Reset leaves the period watermark set, so it silently drops every later bar
#### Expected Behavior

#### Actual Behavior

`AdvanceDeclineIndicator.Reset()` clears `_advanceDeclineDictionary` and `_previousPeriodValues` but leaves `_currentPeriodTime` (line 34) holding the last period it processed. `Enqueue` (lines 165 and 176) only files a bar whose `Time` reaches that watermark, so after a reset every bar earlier than the old watermark is dropped without a warning and the indicator stays unready for the rest of the replay.

Measured over 60 daily periods across 3 symbols, comparing a reset-and-replayed instance against a fresh one: **162 of 180 updates differ in value and 171 of 180 differ in `IsReady`**. Minute resolution behaves the same.

This also affects `AdvanceDeclineDifference`, `AdvanceDeclineRatio`, `AdvanceDeclineVolumeRatio`, `McClellanOscillator` and `McClellanSummationIndex`, which all build on it.

#### Potential Solution

Clear `_currentPeriodTime` in `Reset()` alongside the two dictionaries.

#### Reproducing the Problem

`ComparesAgainstExternalDataAfterReset` does not catch this. `TestHelper.AssertIndicatorHasExternalDataAfterReset` skips its assertion while the indicator is not ready, and on `master` the second pass of that test asserts **0 of 756 rows** for all three A/D indicators — the reset leaves the indicator permanently unready, so every row is skipped and the test passes vacuously. With `_currentPeriodTime` cleared it asserts 753 of 756.

A direct test: feed the indicator a set of bars, call `Reset()`, feed the same bars again, and compare each `Current.Value` against a fresh instance fed the same bars once.

#### System Information

#### Checklist

mkzung:
#### Description

`AdvanceDeclineIndicator.Reset()` clears both period dictionaries but leaves `_currentPeriodTime` holding the last period it processed. `Enqueue` only files a bar whose `Time` reaches that watermark, so after a reset every earlier bar is dropped and the indicator never becomes ready again. Clearing the watermark alongside the dictionaries is the whole fix.

The test doubles in `McClellanOscillatorTests` and `McClellanSummationIndexTests` seed the EMA values implied by the A/D history preceding their CSV. That seeding ran once, in the constructor, so it did not survive a reset. It moves into their `Reset` override, which the constructor already calls. Without this the two fixtures fail once the indicator starts producing values after a reset, which is the point of the change.

#### Related Issue

Fixes #9683.

#### Motivation and Context

Reset is meant to return an indicator to its constructed state. Here it left one field set, and the field gates whether any further data is accepted at all, so the indicator went quiet rather than wrong. 162 of 180 updates differ from a fresh instance over 60 daily periods across 3 symbols, and 171 of 180 differ in `IsReady`.

`AdvanceDeclineDifference`, `AdvanceDeclineRatio`, `AdvanceDeclineVolumeRatio`, `McClellanOscillator` and `McClellanSummationIndex` all build on this indicator.

No.

`ResetsProperlyAndReplaysTheSameValues` in `AdvanceDeclineDifferenceTests` feeds the indicator, resets it, replays the same bars, and compares each value and `IsReady` against a fresh instance. It fails on `master` and passes with the fix. Being in the shared A/D fixture, it is inherited by the ratio and volume-ratio fixtures too.

`QuantConnect.Tests.Indicators` on a clean checkout of this branch: 2768 passed, 0 failed, 5 skipped, on .NET 9 under Linux.

Worth flagging separately: `TestHelper.AssertIndicatorHasExternalDataAfterReset` skips its assertion while the indicator is not ready, so on `master` the second pass of `ComparesAgainstExternalDataAfterReset` asserts 0 of 756 rows for all three A/D indicators and passes vacuously. With this fix it asserts 753 of 756. That helper is inherited by roughly 200 fixtures, so tightening it is a separate discussion rather than part of this PR.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/AdvanceDeclineIndicator.cs
+++ b/Indicators/AdvanceDeclineIndicator.cs
@@ -150,6 +150,7 @@ namespace QuantConnect.Indicators
             {
                 _currentPeriod[key] = null;
             }
+            _currentPeriodTime = null;
 
             base.Reset();
         }
--- a/Tests/Indicators/AdvanceDeclineDifferenceTests.cs
+++ b/Tests/Indicators/AdvanceDeclineDifferenceTests.cs
@@ -84,6 +84,26 @@ namespace QuantConnect.Tests.Indicators
         }
 
         [Test]
+        public void ProducesTheSameValuesAfterReset()
+        {
+            var indicator = CreateIndicator();
+            var reference = System.DateTime.Today;
+
+            UpdateStocks(indicator, reference.AddMinutes(1), 1m, 1m, 1m);
+            UpdateStocks(indicator, reference.AddMinutes(2), 2m, 0.5m, 3m);
+            var expected = indicator.Current.Value;
+            Assert.AreNotEqual(0m, expected);
+
+            UpdateStocks(indicator, reference.AddMinutes(3), 3m, 1m, 2m);
+            indicator.Reset();
+
+            UpdateStocks(indicator, reference.AddMinutes(1), 1m, 1m, 1m);
+            UpdateStocks(indicator, reference.AddMinutes(2), 2m, 0.5m, 3m);
+
+            Assert.AreEqual(expected, indicator.Current.Value);
+        }
+
+        [Test]
         public virtual void IgnorePeriodIfAnyStockMissed()
         {
             var adDifference = (AdvanceDeclineDifference)CreateIndicator();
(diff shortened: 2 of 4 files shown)


Martin-Molinero (approved):
Thanks @mkzung!

PR #9687, opened 2026-08-12
Clear ConnorsRelativeStrengthIndex._previousInput on reset

### LINKED ISSUE #9684: ConnorsRelativeStrengthIndex.Reset leaves _previousInput set, changing the first sample after a reset
#### Expected Behavior

#### Actual Behavior

`ConnorsRelativeStrengthIndex.Reset()` does not clear `_previousInput` (line 53). Both `ComputeNextValue` (line 107) and `ComputeTrendStreak` (line 147) read it before writing it, so the first sample after a reset takes the percent-rank branch instead of the seeding branch. A real price-change ratio goes into the rolling window where a fresh instance puts zero, and the trend streak moves off zero one sample early.

Measured over 400 random 140-point series through `CRSI(3, 2, 100)`: **172 of 400 series diverge**, 176 of 56000 samples, and the first divergence always falls between sample 100 and 104 — the point where the lookback window fills.

#### Potential Solution

Set `_previousInput = null` in `Reset()`.

#### Reproducing the Problem

The window is narrow, so a single dataset does not show it: replaying `spy_crsi.csv` gives **0 divergences in 568 rows**. Random series across parameter combinations are what surface it, because the disagreement lives where `IsReady` first turns true.

Feed the indicator a series, call `Reset()`, replay it, and compare against a fresh instance.

#### System Information

#### Checklist

mkzung:
#### Description

`ConnorsRelativeStrengthIndex.Reset()` did not clear `_previousInput`. Both `ComputeNextValue` and `ComputeTrendStreak` read it before writing it, so the first sample after a reset took the percent-rank branch instead of the seeding branch.

#### Related Issue

Fixes #9684.

#### Motivation and Context

The stale input put a real price-change ratio into the rolling window where a fresh instance puts zero, and moved the trend streak off zero one sample early. The two instances then disagreed around the point where the lookback window fills.

No.

`ResetsProperlyAndReplaysTheSameValues` feeds a series, resets, replays it, and compares against a fresh instance. It fails on `master` and passes with the fix.

The window is narrow enough that one dataset misses it: replaying `spy_crsi.csv` gives 0 divergences in 568 rows. Across 400 random 140-point series through `CRSI(3, 2, 100)`, 172 of 400 series diverge, 176 of 56000 samples, and the first divergence always falls between sample 100 and 104.

`ConnorsRelativeStrengthIndexTests` on a clean checkout of this branch: 14 passed. Full `QuantConnect.Tests.Indicators` with all three reset fixes applied: 2770 passed, 0 failed, 5 skipped.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/ConnorsRelativeStrengthIndex.cs
+++ b/Indicators/ConnorsRelativeStrengthIndex.cs
@@ -175,6 +175,7 @@ namespace QuantConnect.Indicators
             _srsi.Reset();
             _priceChangeRatios.Reset();
             _trendStreak = 0;
+            _previousInput = null;
             base.Reset();
         }
     }
--- a/Tests/Indicators/ConnorsRelativeStrengthIndexTests.cs
+++ b/Tests/Indicators/ConnorsRelativeStrengthIndexTests.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using NUnit.Framework;
 using QuantConnect.Indicators;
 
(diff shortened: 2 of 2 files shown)


Martin-Molinero (approved):
Thanks @mkzung!

PR #9688, opened 2026-08-12
Clear SuperTrend._previousClose on reset

### LINKED ISSUE #9685: SuperTrend.Reset leaves _previousClose set, which changes the output when period is 1
#### Expected Behavior

#### Actual Behavior

`SuperTrend.Reset()` does not clear `_previousClose`. `ComputeNextValue` assigns it on the early return it takes while the ATR warms up, so at ordinary periods the stale value is overwritten before it can be read. At `period == 1` the ATR is ready on the first update, that early return never runs, and the trailing-band comparison reads the close carried over from before the reset.

Measured over 400 random series: **400 of 400 diverge at period 1** (2815 of 24000 samples). Period 2 and period 10 give 0 of 400.

#### Potential Solution

Set `_previousClose = 0m` in `Reset()`.

#### Reproducing the Problem

Construct `SuperTrend(1, ...)`, feed it bars, call `Reset()`, replay the same bars, and compare `Current.Value` against a fresh instance.

Related: `BasicUpperBand`, `BasicLowerBand`, `CurrentTrailingUpperBand` and `CurrentTrailingLowerBand` also survive `Reset()` holding values from the previous run. They never feed back into `Current.Value`, so they do not change the indicator's output, but a caller reading them after a reset sees stale numbers.

#### System Information

#### Checklist

mkzung:
#### Description

`SuperTrend.Reset()` did not clear `_previousClose`.

#### Related Issue

Fixes #9685.

#### Motivation and Context

`ComputeNextValue` assigns `_previousClose` on the early return it takes while the ATR warms up, so at ordinary periods the stale value is overwritten before anything reads it. A period of one leaves the ATR ready on the first update, that early return never runs, and the trailing-band comparison reads the close carried over from before the reset.

Across 400 random series: 400 of 400 diverge at period 1, 2815 of 24000 samples. Period 2 and period 10 give 0 of 400, which is why this survived until the period was varied.

No.

`ResetsProperlyWithPeriodOne` builds `SuperTrend(1, ...)`, feeds it, resets, replays, and compares against a fresh instance. It fails on `master` and passes with the fix.

`SuperTrendTests` on a clean checkout of this branch: 13 passed. Full `QuantConnect.Tests.Indicators` with all three reset fixes applied: 2770 passed, 0 failed, 5 skipped.

Not changed here: `BasicUpperBand`, `BasicLowerBand`, `CurrentTrailingUpperBand` and `CurrentTrailingLowerBand` also survive `Reset()` holding values from the previous run. They never feed back into `Current.Value`, so the indicator's output is unaffected and clearing them is a wider change than this one line. Noted on the issue.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/SuperTrend.cs
+++ b/Indicators/SuperTrend.cs
@@ -141,6 +141,7 @@ namespace QuantConnect.Indicators
             _averageTrueRange.Reset();
             _previousTrailingLowerBand = 0;
             _previousTrailingUpperBand = 0;
+            _previousClose = 0;
             _prevSuper = -1;
             base.Reset();
         }
--- a/Tests/Indicators/SuperTrendTests.cs
+++ b/Tests/Indicators/SuperTrendTests.cs
@@ -14,6 +14,7 @@
 */
 
 using System;
+using System.Collections.Generic;
 using NUnit.Framework;
 using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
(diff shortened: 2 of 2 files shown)


Martin-Molinero (approved):
Thanks @mkzung!

Issue #9622, opened 2026-07-15
Add CME 1-Ounce Gold (1OZ), Spot-Quoted Nasdaq-100 (QNDX), and Micro Ultra 10-Year Treasury Note (MWN) futures


AlexCatarino:
## Feature request

Add symbol-master support for three CME Group futures products that are currently missing from LEAN:

1. **1-Ounce Gold futures** — Globex/ClearPort code `1OZ`, COMEX. Cash-settled to the 100-oz Gold (GC) contract at 1/100 size (1 troy oz). Tick $0.25/oz. Monthly listings in the Feb/Apr/Jun/Aug/Oct/Dec cycle. Specs: https://www.cmegroup.com/markets/metals/precious/1-ounce-gold.html · FAQ: https://www.cmegroup.com/articles/faqs/faq-1-oz-gold-futures.html
2. **Spot-Quoted Nasdaq-100 futures** — code `QNDX`, CME. Part of CME's Spot-Quoted futures family (launched June 2025): smaller than Micros, quoted at the cash-index spot level, long-dated expiries with a daily financing adjustment instead of monthly/quarterly rolls. Specs: https://www.cmegroup.com/markets/equities/nasdaq/spot-quoted-nasdaq-100.html · FAQ: https://www.cmegroup.com/articles/faqs/faq-spot-quoted-futures.html
3. **Micro Ultra 10-Year U.S. Treasury Note futures** — code `MWN`, CBOT. 1/10th the size of the Ultra 10-Year (TN) contract ($10,000 face value), launched 2024 alongside the companion Micro Ultra U.S. Treasury Bond (`MWB`). Unlike TN it is cash-settled: final settlement is based on the TN daily settlement price two business days before the first delivery day of the named contract month. Quarterly listing cycle matching TN. Specs: https://www.cmegroup.com/markets/interest-rates/us-treasury/micro-ultra-10-year-us-treasury-note.html · Overview: https://www.cmegroup.com/markets/interest-rates/micro-treasury-futures.html

### Current state

None of these tickers is registered anywhere in LEAN, so `AddFuture()` cannot resolve them, which also blocks live trading and live data through the Interactive Brokers integration even when IB itself lists the products:

[...]

PR #9692, opened 2026-08-13
Add 1-Ounce Gold and Micro Ultra Treasury futures

0xpinara:
#### Description

This adds three CME futures so `AddFuture()` can find them:

- 1-Ounce Gold (1OZ, COMEX)
- Micro Ultra 10-Year U.S. Treasury Note (MTN, CBOT)
- Micro Ultra U.S. Treasury Bond (MWN, CBOT)

Each one gets a ticker constant, an expiry function, a symbol-properties row,
and a market-hours entry.

I found two things that are different from the issue:

1. 1OZ does not use the same calendar as GC/MGC. Trading stops on the third
   last business day of the month *before* the contract month. I checked 8 real
   expiry dates from CME and they all match, including the Thanksgiving one.
   I followed the same pattern already used for Henry Hub Last-day Financial (HH).

2. MWN is not the 10-Year Note micro. It is the Ultra Bond micro. The Note
   micro is MTN. The tick sizes are different too (1/64 for the Note, 1/32 for
   the Bond). I added both so it is correct either way. Happy to remove one if
   you only want one.

I left QNDX out. The issue says Spot-Quoted futures do not follow the
quarterly-roll model and need a design decision first. I can do it in a
follow-up PR.

#### Related Issue

Partially addresses #9622.

#### Motivation and Context

These tickers are not registered anywhere in LEAN, so users cannot add them at
all. This unblocks live trading and live IB data for them.

Yes, the list of supported futures should get these three new tickers.

I added new test cases in `FuturesExpiryFunctionsTests` and new rows in
`FuturesExpiryFunctionsTestData.xml`. `FuturesExpiryFunctionsTests` passes
203/203 locally.

I could not run `MarketHoursDatabaseTests` or `SymbolPropertiesDatabaseTests`
on my machine. They crash during teardown with a Python GIL error, and this
also happens on unmodified master, so I do not think it is from my change.

Some 1OZ test rows use off-cycle months on purpose, to test that it rolls
forward to the next Feb/Apr/Jun/Aug/Oct/Dec contract.

One note on the diff: the market-hours JSON looks like it changed a lot, but I
parsed the file before and after and only 3 entries were added. No existing
entry was changed. The big diff is just git lining up the wrong blocks, because
the new entries look almost the same as their neighbors.

#### Types of changes

#### Checklist:


Common/Securities/Future/FuturesExpiryFunctions.cs
```diff
@@ -58,6 +58,24 @@ public static Func<DateTime, DateTime> FuturesExpiryFunction(Symbol symbol)
...
+                    var holidays = FuturesExpiryUtilityFunctions.GetExpirationHolidays(market, symbol);
+                    // Bi-monthly contracts (Feb/2, Apr/4, Jun/6, Aug/8, Oct/10, Dec/12 cycle)
+                    while (!FutureExpirationCycles.GJMQVZ.Contains(time.Month))
+                    {
+                        time = time.AddMonths(1);
+                    }
+
+                    // This contract is cash settled, so it stops trading before the contract month even starts.
+                    // Trading terminates on the third last business day of the month prior to the contract month.
+                    var previousMonth = time.AddMonths(-1);
```
AlexCatarino, line 75: Because this returns a date in the month *before* the contract month, `1OZ` also needs registering in `FuturesExpiryUtilityFunctions.ExpiriesPriorMonth` — that's the companion half of the `HH` pattern you followed (`HH` is there at line 414).

Without it, `SymbolRepresentation.GenerateFutureTicker` renders the Feb-2026 contract as `1OZ28F26` instead of `1OZ28G26` (I confirmed this by test), and `LeanData` paths plus `FutureUniverse.ToCsv` record the wrong contract month. Add to that dictionary:

```csharp
{ Futures.Metals.OneOunceGold, 1 },
```

Worth a `[TestCase]` in `SymbolRepresentationTests.GenerateFutureTickerExpiringInPreviousMonth` alongside the existing `CL`/`HH` rows.

Common/Securities/Future/FuturesExpiryFunctions.cs
```diff
@@ -1674,6 +1692,48 @@ public static Func<DateTime, DateTime> FuturesExpiryFunction(Symbol symbol)
...
+                    while (!FutureExpirationCycles.HMUZ.Contains(time.Month))
+                    {
+                        time = time.AddMonths(1);
+                    }
+
+                    // This contract is cash settled against TN instead of delivering notes, so it stops
+                    // trading earlier than TN does.
+                    // Trading terminates 2 business days before the first delivery day of the contract
+                    // month. The first delivery day is the first business day of that month.
+                    var firstDeliveryDay = FuturesExpiryUtilityFunctions.NthBusinessDay(time, 1, holidays);
```
AlexCatarino, line 1712: Same as the 1OZ note: this lands in the previous month, so `MTN` needs an `ExpiriesPriorMonth` entry too, otherwise the Mar-2025 contract renders as `MTN27G25` instead of `MTN27H25`.

```csharp
{ Futures.Financials.MicroUltraTenYearUSTreasuryNote, 1 },
```

Common/Securities/Future/FuturesExpiryFunctions.cs
```diff
@@ -1674,6 +1692,48 @@ public static Func<DateTime, DateTime> FuturesExpiryFunction(Symbol symbol)
...
+                    return FuturesExpiryUtilityFunctions.AddBusinessDays(firstDeliveryDay, -2, holidays);
```
AlexCatarino: CBOT ch. 54102.A terminates trading **at 2:00 p.m. CT**, so the expiry should carry the time (19:00 UTC under LEAN's summer-offset convention). Returning midnight makes `IsFutureContractExpired` treat the contract as expired for the whole of its final trading day.

`F1U` is the precedent here — same 2:00 p.m. close, same 19:00 encoding, and its test case already uses `TwoPMCentralTime`.

```suggestion
                    return FuturesExpiryUtilityFunctions.AddBusinessDays(firstDeliveryDay, -2, holidays).AddHours(19);
```

Common/Securities/Future/FuturesExpiryFunctions.cs
```diff
@@ -1674,6 +1692,48 @@ public static Func<DateTime, DateTime> FuturesExpiryFunction(Symbol symbol)
...
+                    while (!FutureExpirationCycles.HMUZ.Contains(time.Month))
+                    {
+                        time = time.AddMonths(1);
+                    }
+
+                    // This contract is cash settled against UB instead of delivering bonds, so it stops
+                    // trading earlier than UB does.
+                    // Trading terminates 2 business days before the first delivery day of the contract
+                    // month. The first delivery day is the first business day of that month.
+                    var firstDeliveryDay = FuturesExpiryUtilityFunctions.NthBusinessDay(time, 1, holidays);
```
AlexCatarino, line 1733: `MWN` needs the same `ExpiriesPriorMonth` entry — currently the Mar-2025 contract renders as `MWN27G25` instead of `MWN27H25`.

```csharp
{ Futures.Financials.MicroUltraUSTreasuryBond, 1 },
```

Common/Securities/Future/FuturesExpiryFunctions.cs
```diff
@@ -1674,6 +1692,48 @@ public static Func<DateTime, DateTime> FuturesExpiryFunction(Symbol symbol)
...
+                    return FuturesExpiryUtilityFunctions.AddBusinessDays(firstDeliveryDay, -2, holidays);
```
AlexCatarino: Same 2:00 p.m. CT termination here (ch. 56102.A).

```suggestion
                    return FuturesExpiryUtilityFunctions.AddBusinessDays(firstDeliveryDay, -2, holidays).AddHours(19);
```

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -388,6 +388,8 @@ public void MicroCrudeOilExpiration(string symbol, string dateStr, string expect
         [TestCase(QuantConnect.Securities.Futures.Financials.UltraUSTreasuryBond, TwelveOne)]
         [TestCase(QuantConnect.Securities.Futures.Financials.UltraTenYearUSTreasuryNote, Zero)]
         [TestCase(QuantConnect.Securities.Futures.Financials.MicroY10TreasuryNote, Zero)]
+        [TestCase(QuantConnect.Securities.Futures.Financials.MicroUltraTenYearUSTreasuryNote, Zero)]
+        [TestCase(QuantConnect.Securities.Futures.Financials.MicroUltraUSTreasuryBond, Zero)]
```
AlexCatarino: Pairs with the 2:00 p.m. CT change above. `TwoPMCentralTime` already exists in this class (= `"19:00:00"`), so no new constant is needed.

```suggestion
        [TestCase(QuantConnect.Securities.Futures.Financials.MicroUltraTenYearUSTreasuryNote, TwoPMCentralTime)]
        [TestCase(QuantConnect.Securities.Futures.Financials.MicroUltraUSTreasuryBond, TwoPMCentralTime)]
```

AlexCatarino (commented):
Reviewed against CME's own CFTC product-certification filings (cmegroup.com returns 403 to automated fetches, but COMEX Submission 24-474 and CBOT Submission 24-090 carry the verbatim rulebook text), and built/ran the suites locally.

**The specs all check out, and the correction in the description is right.** CBOT Submission 24-090: rulebook ch. 54 = `MTN` (the 10-Year Note micro), ch. 56 = `MWN` (the Bond micro), with the two tick sizes exactly as described (MTN ½ of 1/32 = $1.5625, MWN 1/32 = $3.125) and $100/point on $10,000 face. Issue #9622 has MWN backwards. For 1OZ, COMEX ch. 131 confirms 1 troy oz, $0.25/oz, financial settlement, listings in **Feb/Apr/Jun/Aug/Oct/Dec only with no serial months**, and termination on *"the third last business day of the month prior to the contract month"* — stated three separate ways in the filing, so the departure from GC is real.

Both expiry formulas reproduce CME-published dates: the filing's own worked example for MTN Jun-2024 (30 May 2024), the listed MTN/MWN Sep-2026 (28 Aug), Dec-2026 (27 Nov) and Mar-2027 (25 Feb), and 1OZ Oct-2026 (28 Sep) and Dec-2026 (25 Nov). I also verified the market-hours claim by parsing both versions of the JSON: exactly 3 entries added, 0 existing entries changed, ordering preserved.

**Two things block.** All three contracts expire in the month *before* the contract month but are not registered in `ExpiriesPriorMonth`, which makes `Symbol.Value` one month early; and the three margin files are missing, which silently gives these contracts zero margin. Details inline.

FYI on the suites you couldn't run: `MarketHoursDatabaseTests` and `SymbolPropertiesDatabaseTests` pass here on .NET 10 (152/152, no GIL crash), so that looks local to your setup.

Tests/Common/Securities/Futures/FuturesExpiryFunctionsTests.cs
```diff
@@ -505,6 +507,7 @@ public void LumberPulpExpiryDateFunction_WithDifferentDates_ShouldFollowContract
         [TestCase(QuantConnect.Securities.Futures.Metals.MiniNYSilver, OneTwentyFivePM)]
         [TestCase(QuantConnect.Securities.Futures.Metals.Gold100Oz, OneThirtyPM)]
         [TestCase(QuantConnect.Securities.Futures.Metals.Silver5000Oz, OneTwentyFivePM)]
+        [TestCase(QuantConnect.Securities.Futures.Metals.OneOunceGold, Zero)]
```
AlexCatarino, line 534: Minor, on the XML data behind this case: two of the six 1OZ rows use contract months CME doesn't list (`2025-09-01`, `2026-11-01` — listings are Feb/Apr/Jun/Aug/Oct/Dec only). They land on the real Oct-2025 and Dec-2026 last-trade dates anyway, so changing `ContractMonth` to `2025-10-01` and `2026-12-01` asserts the same values as genuine contracts. The roll-forward behaviour could then get its own `[Test]`, the way `MCLFutures` and `Kospi200Futures` do it.

Data/symbol-properties/symbol-properties-database.csv
```diff
@@ -225,6 +225,8 @@ cbot,BWF,future,Black Sea Wheat Financially Settled (Platts) Futures,USD,50.0,0.
 cbot,EH,future,Ethanol Futures,USD,29000.0,0.001,1.0
 cbot,F1U,future,5-Year USD MAC Swap Futures,USD,1000.0,0.0078125,1.0
 cbot,KE,future,KC HRW Wheat Futures,USD,5000.0,0.0025,1.0,,1,100
+cbot,MTN,future,Micro Ultra 10-Year U.S. Treasury Note Futures,USD,100.0,0.015625,1.0
+cbot,MWN,future,Micro Ultra U.S. Treasury Bond Futures,USD,100.0,0.03125,1.0
```
AlexCatarino, line 229: `Data/future/cbot/margins/MTN.csv` and `MWN.csv` are missing. When the margin file isn't found, `FutureMarginModel.FromCsvFile` logs an error and defaults initial/maintenance to **zero**, so these contracts get effectively unlimited leverage.

`Data/future/krx/margins/KM.csv` (#9585) is the format to copy:

```
# we don't have historical information for this symbol
date,initial,maintenance
19900101,<initial>,<maintenance>
```

Data/symbol-properties/symbol-properties-database.csv
```diff
@@ -307,6 +309,7 @@ cme,RX,future,DJRE Futures,USD,100.0,0.1,1.0
 cme,SDA,future,SP500 S&P 500 Annual Dividend Index future,USD,250.0,0.05,1.0
 cme,TPD,future,TOPIX USD Futures,USD,50.0,0.5,1.0
 cme,TPY,future,TOPIX JPY Futures,JPY,5000.0,0.5,1.0
+comex,1OZ,future,1-Ounce Gold Futures,USD,1,0.25,1
```
AlexCatarino: Needs `Data/future/comex/margins/1OZ.csv` too (missing file → zero margin, as above).

Multiplier and tick are correct per ch. 131102.B/C. Cosmetic only — neighbouring rows write the multiplier and lot size in `1.0` style:

```suggestion
comex,1OZ,future,1-Ounce Gold Futures,USD,1.0,0.25,1.0
```

0xpinara: Thank you for the review, and for checking everything against the filings.
Pushed 8a0b63a59 with all the changes.

**The two blockers.** All three are now in `ExpiriesPriorMonth`, and the three
margin files exist. I got the margins from CME's Product Margins table rather
than estimating: maintenance is 220 for 1OZ (10/2026 front month), 255 for MTN
and 515 for MWN (both 09/2026-12/2026). Initial is 110% of maintenance per the
readme, rounded up like `AUP.csv` does, so 242, 281 and 567.

**The rest.** MTN and MWN return `.AddHours(19)` and their test cases use
`TwoPMCentralTime`. The three `[TestCase]` rows are in
`GenerateFutureTickerExpiringInPreviousMonth` and produce `1OZ28G26`,
`MTN27H25` and `MWN27H25`. The 1OZ contract months are now `2025-10-01` and
`2026-12-01`. MTN and MWN had the same problem with a January contract month,
so I fixed those too and sorted the rows by date. Roll-forward has its own tests
following `MCLFutures`. The `1.0` style fix is applied.

**On the test suites, you were right and it was my setup.** `PYTHONNET_PYDLL`
was not set, so pythonnet crashed on teardown. With it set everything passes:
FuturesExpiryFunctionsTests 206/206, MarketHoursDatabaseTests 110 and
SymbolPropertiesDatabaseTests 42 (152 together, same as yours),
SymbolRepresentationTests 102/102, FutureMarginBuyingPowerModelTests 72/72.

FYI, CME's margin table lists MWN as "MICRO LONG TERM US TREAS BOND FUT" and MTN as "MICRO ULTRA 10-YEAR US TREASURY FUT".

AlexCatarino (approved):
Thank you, @0xpinara.

Martin-Molinero (approved):
Thank you!

PR #9694, opened 2026-08-13
Reset the MidPrice rolling extremes

### LINKED ISSUE #9693: MidPrice does not reset its rolling maximum and minimum
MidPrice holds a Maximum and a Minimum and never overrides Reset(), so both keep the previous run's window. The first bars after a reset are computed from data that is no longer in scope.

Steps to reproduce, four bars, period 3:

```csharp
var mid = new MidPrice(3);
// High/Low: 110/100, 111/101, 112/102, 105/95
// first run:   105, 105.5, 106, 103.5
mid.Reset();
// after reset: 103.5, 103, 106, 103.5
```

Expected the second pass to repeat the first. The first period - 1 values are wrong, after which the stale window has rolled out: 1 value at period 2, 2 at period 3, 4 at period 5, 9 at period 10. MidPrice(5) is what MidPriceTests constructs.

Reset() does clear the public surface, Samples is 0, IsReady is false and Current.Value is 0, which is why TestHelper.AssertIndicatorIsInDefaultState passes. It recurses into public sub-indicator properties, and these two are private fields.

Behaviour outside a reset is unaffected. Feeding Tests/TestData/spy_midprice.txt matches all 496 expected MIDPRICE_5 values both before and after the fix.

mkzung:
#### Description

MidPrice never overrode Reset(), so its Maximum and Minimum carried the previous run.

#### Related Issue

Closes #9693

#### Motivation and Context

The first bars after a reset were computed from data outside the window. Same class as #9686, #9687 and #9688.

No.

Added ProducesTheSameValuesAfterReset to MidPriceTests, in the shape used in ConnorsRelativeStrengthIndexTests and SuperTrendTests. It fails on master and passes with the override.

Also ran the indicator against Tests/TestData/spy_midprice.txt on a master build and on this one: 496 expected MIDPRICE_5 values, no mismatches either way, so nothing outside the reset path moves.

My local NUnit host crashes on a Python.NET finalizer during test discovery, unrelated to this change, so both runs above were done from a console program linked against each build rather than through `dotnet test`.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/MidPrice.cs
+++ b/Indicators/MidPrice.cs
@@ -72,5 +72,15 @@ namespace QuantConnect.Indicators
 
             return (_maximum.Current.Value + _minimum.Current.Value) / 2;
         }
+
+        /// <summary>
+        /// Resets this indicator to its initial state
+        /// </summary>
+        public override void Reset()
+        {
+            _maximum.Reset();
+            _minimum.Reset();
+            base.Reset();
+        }
     }
 }
--- a/Tests/Indicators/MidPriceTests.cs
+++ b/Tests/Indicators/MidPriceTests.cs
@@ -13,6 +13,8 @@
  * limitations under the License.
 */
 
+using System;
+using System.Collections.Generic;
 using NUnit.Framework;
 using QuantConnect.Data.Market;
 using QuantConnect.Indicators;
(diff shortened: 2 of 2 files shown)


Martin-Molinero (approved):
Thanks @mkzung!

Issue #9697, opened 2026-08-13
PythonIndicator.Reset() does not reset the python indicator


mkzung:
PythonIndicator forwards Update, IsReady, Value and WarmUpPeriod to the python object but never overrides Reset(), so a reset defined in python is never called and the python side keeps its state.

A 3-period SMA that clears its own queue:

```python
class DuckSma(PythonIndicator):
    def __init__(self, name, period):
        self.Name = name
        self.Value = 0
        self.WarmUpPeriod = period
        self.queue = deque(maxlen=period)

    def Update(self, input):
        self.queue.appendleft(input.Value)
        self.Value = sum(self.queue) / len(self.queue)
        return len(self.queue) == self.queue.maxlen

    def Reset(self):
        self.queue.clear()
        self.Value = 0
```

```csharp
var indicator = new PythonIndicator(py);
// feed 100, 101, 102, 103, 104
indicator.Reset();
indicator.Update(new IndicatorDataPoint(t, 100m));
```

Current.Value is 102.33333333333333, the mean of the stale queue, where 100 is expected. The python Reset ran 0 times. IsReady is also True while Samples is 0, because _isReady is set in ComputeNextValue and nothing clears it.

Both wrapping paths behave the same: `new PythonIndicator(pyObject)`, and the TryConvert path QCAlgorithm.WrapPythonIndicator uses.

QCAlgorithm.IndicatorHistory calls indicator.Reset() before replaying history (QCAlgorithm.Indicators.cs:4496), and IndicatorHistory(PyObject, ...) routes python indicators into it, so indicator_history on an indicator that has already seen data returns values mixed with the old state.

PythonIndicatorTests.ResetsProperly overrides the base test and runs it against a C# SimpleMovingAverage(3) instead of the python fixture, which is why this was never caught.

PR #9698, opened 2026-08-13
Call the python reset from PythonIndicator.Reset

mkzung:
#### Description
PythonIndicator never overrode Reset(), so a reset defined in python was never called and _isReady stayed set.

#### Related Issue
#9697

#### Motivation and Context
indicator_history resets before it replays, so it returned values mixed with whatever the indicator had already seen. Same class as #9686, #9687, #9688 and #9694, one level up.

No.

Added ResetClearsTheStateHeldInPython to PythonIndicatorTests. It runs in all six python fixtures, both wrapping paths and both naming conventions, and fails on master at 100.75 against 100. The two duck-typed fixtures gained the reset they were missing.

Full suite, `--filter "TestCategory!=TravisExclude&TestCategory!=ResearchRegressionTests"`, run on pristine master and on this branch:

    master      36614 tests, 3 failed
    this branch 36620 tests, 3 failed

Same three, and they fail on master as well: CommanCallback(Python), ThreadSafety, ZipBytesReturnsByteArrayWithCorrectLength. The extra six are the new test in its six fixtures.

One note on the guard: GetMethod throws when the attribute is missing altogether and returns null only when it resolves to C#, so an unguarded call breaks every python indicator that has no reset. The first version of this did exactly that, and only the full run caught it.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/PythonIndicator.cs
+++ b/Indicators/PythonIndicator.cs
@@ -26,9 +26,12 @@ namespace QuantConnect.Indicators
     public class PythonIndicator : IndicatorBase<IBaseData>, IIndicatorWarmUpPeriodProvider
     {
         private static string _isReadyName = nameof(IsReady).ToSnakeCase();
+        private static string _resetName = nameof(Reset).ToSnakeCase();
         private PyObject _instance;
         private bool _isReady;
         private bool _pythonIsReadyProperty;
+        private PyObject _pythonResetMethod;
+        private bool _isResetting;
         private BasePythonWrapper<IIndicator> _indicatorWrapper;
 
         /// <summary>
@@ -94,6 +97,16 @@ namespace QuantConnect.Indicators
                 }
             }
 
+            using (Py.GIL())
+            {
+                // Null when the attribute is absent, is a plain value rather than a method, or resolves
+                // to the CSharp implementation. Resolving here keeps Reset off the per-call GIL round trip.
+                // SetIndicator runs again on every GetIndicatorAsManagedObject call, so release first.
+                _pythonResetMethod?.Dispose();
+                _pythonResetMethod = indicator.GetPythonMethodWithChecks(_resetName) as PyObject
+                    ?? indicator.GetPythonMethodWithChecks(nameof(Reset)) as PyObject;
+            }
+
             WarmUpPeriod = GetIndicatorWarmUpPeriod();
         }
 
@@ -133,16 +146,28 @@ namespace QuantConnect.Indicators
         /// </summary>
         public override void Reset()
         {
-            // GetMethod throws when the attribute is absent, and returns null when it is CSharp
-            if (_indicatorWrapper != null && _indicatorWrapper.HasAttr(nameof(Reset)))
+            // For an inheriting class the wrapped instance is this same object, so a python reset()
+            // calling super().reset() arrives back here through the CLR binding. Invoke the python
+            // side once per reset and let the re-entrant call fall through to the base.
+            var reentrant = _isResetting;
+            try
             {
-                using (Py.GIL())
+                if (_pythonResetMethod != null && !reentrant)
                 {
-                    _indicatorWrapper.GetMethod(nameof(Reset), pythonOnly: true)?.Invoke().Dispose();
+                    _isResetting = true;
+                    using (Py.GIL())
+                    {
+                        _pythonResetMethod.Invoke().Dispose();
+                    }
                 }
             }
-            _isReady = false;
-            base.Reset();
+            finally
+            {
+                // A python reset() that raises must still leave the CSharp side reset.
+                _isResetting = reentrant;
+                _isReady = false;
+                base.Reset();
+            }
         }
 
         /// <summary>
--- a/Indicators/PythonIndicator.cs
+++ b/Indicators/PythonIndicator.cs
@@ -129,6 +129,23 @@ namespace QuantConnect.Indicators
         public int WarmUpPeriod { get; protected set; }
 
         /// <summary>
+        /// Resets this indicator to its initial state
+        /// </summary>
+        public override void Reset()
+        {
+            // GetMethod throws when the attribute is absent, and returns null when it is CSharp
+            if (_indicatorWrapper != null && _indicatorWrapper.HasAttr(nameof(Reset)))
+            {
+                using (Py.GIL())
+                {
+                    _indicatorWrapper.GetMethod(nameof(Reset), pythonOnly: true)?.Invoke().Dispose();
+                }
+            }
+            _isReady = false;
+            base.Reset();
+        }
+
+        /// <summary>
         /// Computes the next value of this indicator from the given state
         /// </summary>
         /// <param name="input">The input given to the indicator</param>
--- a/Tests/Indicators/PythonIndicatorTests.cs
+++ b/Tests/Indicators/PythonIndicatorTests.cs
@@ -175,12 +175,19 @@ class CustomSimpleMovingAverage(PythonIndicator):
             var indicator = CreateIndicator();
             var reference = new DateTime(2024, 1, 1);
 
-            for (var i = 0; i < 3; i++)
+            // Enough points to reach IsReady, otherwise asserting it is false after the reset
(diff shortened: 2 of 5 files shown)


Indicators/PythonIndicator.cs
```diff
@@ -128,6 +128,23 @@ public override bool IsReady
...
+        /// Resets this indicator to its initial state
+        /// </summary>
+        public override void Reset()
+        {
+            // GetMethod throws when the attribute is absent, and returns null when it is CSharp
+            if (_indicatorWrapper != null && _indicatorWrapper.HasAttr(nameof(Reset)))
+            {
+                using (Py.GIL())
+                {
+                    _indicatorWrapper.GetMethod(nameof(Reset), pythonOnly: true)?.Invoke().Dispose();
```
Martin-Molinero: For inheriting classes `_indicatorWrapper` wraps the instance itself, so a subclass `reset()` calling `super().reset()` (the pre-PR correct pattern) recurses: C# `Reset()` → python `reset()` → CLR binding → C# `Reset()`... Reproduced on this branch: ~995 frames, fatal `0xC0000005`. A reentrancy guard is needed here.

Indicators/PythonIndicator.cs
```diff
@@ -128,6 +128,23 @@ public override bool IsReady
         /// </summary>
         public int WarmUpPeriod { get; protected set; }
 
+        /// <summary>
+        /// Resets this indicator to its initial state
+        /// </summary>
+        public override void Reset()
+        {
+            // GetMethod throws when the attribute is absent, and returns null when it is CSharp
+            if (_indicatorWrapper != null && _indicatorWrapper.HasAttr(nameof(Reset)))
```
Martin-Molinero: `HasAttr` is snake-case tolerant but `GetPythonMethod` falls back to PascalCase-only `GetAttr("Reset")` — so `self.reset = False` (non-method, no `Reset`) passes the guard then throws an uncaught `AttributeError`; worked before this PR. Also, a non-bound-method callable (`self.Reset = lambda: ...`) is silently skipped. Suggest resolving the reset method once in `SetIndicator`, like `_pythonIsReadyProperty` — also removes the per-call `HasAttr` GIL round-trip and avoids caching `null` under `"Reset"` in `_pythonMethods` (keyed by name only, ignores `pythonOnly`).

Indicators/PythonIndicator.cs
```diff
@@ -128,6 +128,23 @@ public override bool IsReady
...
+            {
+                using (Py.GIL())
+                {
+                    _indicatorWrapper.GetMethod(nameof(Reset), pythonOnly: true)?.Invoke().Dispose();
+                }
+            }
+            _isReady = false;
```
Martin-Molinero: If the python `reset()` raises, `_isReady = false` / `base.Reset()` are skipped → half-reset indicator. Run them before the invoke or in a `finally`.

Tests/Indicators/PythonIndicatorTests.cs
```diff
@@ -165,6 +169,23 @@ public void IsReadyAfterPeriodUpdates()
...
+
+            for (var i = 0; i < 3; i++)
+            {
+                indicator.Update(new IndicatorDataPoint(reference.AddDays(i), 100m + i));
+            }
+
+            indicator.Reset();
+            indicator.Update(new IndicatorDataPoint(reference, 100m));
+
+            Assert.AreEqual(100m, indicator.Current.Value);
```
Martin-Molinero, line 193: Also assert `IsReady` is false (or `Samples == 0`) after `Reset()` — dropping `_isReady = false` wouldn't fail this test.

Martin-Molinero (commented):
Hey @mkzung! Thanks, leaving a few comments

Fix works for the plain non-inheriting case, but two confirmed regressions: segfault via `super().reset()` in inheriting classes, and uncaught `AttributeError` from the `HasAttr`/`GetPythonMethod` name asymmetry. Resolving the reset method once in `SetIndicator` + a reentrancy guard would cover most findings — details inline.

Indicators/PythonIndicator.cs
mkzung: Confirmed, and it is on the normal path. `WrapPythonIndicator` calls `TryConvert`, which for an inheriting class hands back that object's own C# part, then points `SetIndicator` at the same object.

I counted the depth in python rather than reading the crash. The test increments a counter on entry to `reset()` and stops calling `super()` at 200, so re-entry arrives as a number instead of taking the host down. On the previous commit it reaches the cap. It is 1 now, and zero would mean the python reset was never reached, so one assertion covers both directions.

The ~995 frames you saw are that same loop without the cap.

`Reset` invokes the python side once and lets the re-entrant call fall through to the base. The flag is saved and restored rather than cleared, so a `reset()` calling `super()` twice cannot re-arm it.

Indicators/PythonIndicator.cs
mkzung: Fixed the way you suggested, and the shape is not new here: `AlgorithmPythonWrapper` resolves `OnData` and `OnMarginCall` into fields at construction.

Resolution happens once in `SetIndicator` through `GetPythonMethodWithChecks`, snake-case first and PascalCase second, both behind `HasAttr`. So `self.reset = False` now resolves to null rather than reaching `GetAttr("Reset")`, and `null` no longer lands in `_pythonMethods` under a key that ignores `pythonOnly`. The per-call round trip goes with it.

The non-bound callable I have left alone, deliberately.

That `<class 'method'>` test inside `GetPythonMethod` is doing real work: it is what separates a python override from the inherited C# binding. Accept any callable and every inheriting class starts looking like it overrides `reset`. Two wrappers call the helper directly and `BasePythonWrapper` routes every `InvokeMethod` through it, so I would rather widen it on its own.

One consequence of resolving once, since you were the one who flagged the caching: repeated `SetIndicator` now has something to release. `GetIndicatorAsManagedObject` calls it from twenty sites in `IndicatorExtensions` with no caching, unlike `WrapPythonIndicator` which keys off the handle, so the field is disposed before it is replaced. I stopped there. `AlgorithmPythonWrapper` releases its equivalents in `Dispose`, but `PythonIndicator` has no `Dispose` and does not release `_instance` or `_indicatorWrapper` either, and adding one is a lifetime contract rather than a fix.

Indicators/PythonIndicator.cs
mkzung: Confirmed. Both are in a `finally` now.

Tests/Indicators/PythonIndicatorTests.cs
mkzung: Sharper than it first looked, because the assertion would have been vacuous as well as missing. The indicator has period 14 and the test fed it three points, so it was never ready, and asserting `IsReady` is false after a reset would have passed against any implementation at all.

It feeds 20 now and asserts ready first.

Dropping `_isReady = false` turns it red in all six fixtures that inherit `PythonIndicatorTests`.

mkzung:
All four reproduced and fixed. Second commit rather than an amend, so these threads stay attached.

The three new cases build their own python classes and never touch `CreateIndicator`, so they sit in their own `PythonIndicatorResetTests` pair. Inheriting them from `PythonIndicatorTests` put a test about an inheriting class inside `PythonIndicatorNoinheritanceTests`, and ran each of them in four more fixtures without changing a thing they exercised. The strengthened `ResetClearsTheStateHeldInPython` stays where it is, since it does use `CreateIndicator`.

Five runs fail on the previous commit, none on this one. The new file adds no analyzer warning beyond the two CA1515 that every public fixture here already emits.

A caveat instead of a clean number for the wider suite. `--filter FullyQualifiedName~QuantConnect.Tests.Indicators` aborts partway on my machine, because a host wants `Data/equity/oanda/map_files` and my clone does not have it. It aborts identically on untouched master, so it is not from this change, but it does mean both runs are partial rather than a full sweep. Master gives 2771 passed and 0 failed, this branch 2783 and 0 failed, and the difference of 12 is the four reset tests: three in two fixtures, one in six.

Martin-Molinero (approved):
Nice job! Thank you

Issue #9699, opened 2026-08-14
ResetWorksAsExpected never compares the replayed bars


mkzung:
BaseConsolidatorTests.ResetWorksAsExpected collects the consolidated bars, resets, replays the same values and then compares the two lists:

```csharp
Assert.AreEqual(consolidatedBarsBefore.Count, consolidatedBarsAfter.Count);
consolidatedBarsBefore.Zip<IBaseData, IBaseData, bool>(consolidatedBarsAfter, AssertConsolidatedValues);
```

Zip is lazy and the result is discarded, so AssertConsolidatedValues never runs. Only the counts are compared, across all 16 classes that inherit the fixture.

Putting `Assert.Fail("the comparison ran")` as the first line of AssertConsolidatedValues leaves all 17 tests passing.

What that misses, on master: deleting `_workingBar = null;` from PeriodCountConsolidatorBase.Reset() leaves a genuine state leak in the base class most consolidators derive from, and `ResetWorksAsExpected` still passes 17 of 17. The bar counts match because the leak changes bar contents, not how many close.

Three of the sixteen also produce no consolidated bars at all with the values the fixture feeds, so there is nothing to compare even once the comparison runs:

- OpenInterestConsolidatorTests consolidates on TimeSpan.FromDays(1) and feeds 11 minutes
- MarketHourAwareConsolidatorTests consolidates hourly and feeds 11 minutes
- DataConsolidatorPythonWrapperTests uses a python consolidator that never sets consolidated, which is deliberate

PR #9700, opened 2026-08-14
Compare the replayed bars in ResetWorksAsExpected

mkzung:
#### Description
ResetWorksAsExpected discarded a lazy Zip, so AssertConsolidatedValues never ran and only the bar counts were compared.

#### Related Issue
#9699

#### Motivation and Context
The test is the only reset coverage the 16 consolidator fixtures share. Deleting `_workingBar = null;` from PeriodCountConsolidatorBase.Reset() leaves a real state leak in the base class most consolidators derive from, and on master the test still passes 17 of 17. With the comparison enumerated it fails.

Three fixtures also produced no consolidated bars at all, so there was nothing to compare even once the comparison runs. OpenInterest consolidates daily and fed 11 minutes; MarketHourAware consolidates hourly and fed 11 minutes. Both now cross a period boundary. The python wrapper fixture never sets consolidated on purpose, so it declares that instead.

No.

Putting `Assert.Fail` as the first line of AssertConsolidatedValues passes 17 of 17 on master, which is the bug. With this change the same planted `_workingBar` leak turns ResetWorksAsExpected red and reverting it turns it green.

All 17 fixtures pass with the comparison running.

Full suite, `--filter "TestCategory!=TravisExclude&TestCategory!=ResearchRegressionTests"`: 36614 tests, 4 failures, against 3 on master. CommanCallback(Python) and ZipBytesReturnsByteArrayWithCorrectLength fail on both. The other three, ThreadSafety, DoesNotLeak and MultiThreadReadWriteTest, alternate between runs and all pass in isolation; none of them touches a consolidator.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Common/Data/BaseConsolidatorTests.cs
+++ b/Tests/Common/Data/BaseConsolidatorTests.cs
@@ -31,6 +31,11 @@ namespace QuantConnect.Tests.Common.Data
     {
         protected abstract IDataConsolidator CreateConsolidator();
 
+        /// <summary>
+        /// Whether the test values are expected to close at least one bar
+        /// </summary>
+        protected virtual bool EmitsConsolidatedBars => true;
+
         protected virtual void AssertConsolidator(IDataConsolidator consolidator)
         {
             Assert.IsNull(consolidator.Consolidated);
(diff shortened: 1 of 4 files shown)


Martin-Molinero (approved):
Thanks @mkzung!

PR #9710, opened 2026-08-16
Report the extra bar three candlestick patterns wait for

### LINKED ISSUE #9709: Three candlestick patterns report a WarmUpPeriod one bar short
#### Expected Behavior

`WarmUpPeriod` is the number of data points an indicator needs before `IsReady` turns true.
`WarmUpIndicator` feeds exactly that many and hands back a ready indicator.

#### Actual Behavior

`AbandonedBaby`, `MatHold` and `RiseFallThreeMethods` each turn ready one bar later than they
report.

    AbandonedBaby         WarmUpPeriod 12, ready at sample 13
    MatHold               WarmUpPeriod 15, ready at sample 16
    RiseFallThreeMethods  WarmUpPeriod 15, ready at sample 16

`WindowIndicator.WarmUpPeriod => Period` holds while `IsReady` is the window's own. These three
override `IsReady` to `Samples > Period` and leave `WarmUpPeriod` alone. The other 58 candlestick
patterns use `Samples >= Period`.

#### Reproducing the Problem

```csharp
var pattern = new AbandonedBaby();
for (var i = 0; i < pattern.WarmUpPeriod; i++)
{
    pattern.Update(new TradeBar(new DateTime(2019, 1, 1).AddDays(i), Symbols.SPY,
        100m + i, 105m + i, 95m + i, 100m + i, 100m, Time.OneDay));
}
Console.WriteLine(pattern.IsReady);  // False
```

`CommonIndicatorTests.WarmsUpProperly` asserts exactly this, and `CandlestickPatternTests` does not
derive from `CommonIndicatorTests`, so it has never run for any pattern.

Same surface as #8629, different cause: there `WarmUpPeriod` is absent, here it is present and one
short.

#### System Information

master at d865a40b2.

#### Checklist

mkzung:
#### Description

`AbandonedBaby`, `MatHold` and `RiseFallThreeMethods` override `IsReady` to `Samples > Period` and
leave `WarmUpPeriod` at `WindowIndicator`'s `Period`, so each turns ready one bar after the count it
publishes. The override returns `Period + 1`.

#### Related Issue

Closes #9709

#### Motivation and Context

`WarmUpIndicator` feeds exactly `WarmUpPeriod` bars and hands back an indicator the caller treats as
ready. For these three it is not, and the first value read after warming up is 0.

Changing the comparison instead would move what the patterns report. Accumulation runs under
`if (!IsReady)`, so readiness a bar earlier drops one term from every average and the TA-Lib fixture
values move with it. `Period + 1` changes no value anywhere. If you would rather these three matched
the other 58 and used `>=`, that is a larger change and I am happy to do it that way instead.

The other 58 patterns already agree with their own `WarmUpPeriod`. `EveningDojiStar` and
`MorningDojiStar` carry AbandonedBaby's three candle settings and declare `... + 2 + 1` against its
`... + 2`; `Breakaway` and `LadderBottom` have MatHold's exact period expression with `>=`.

No.

Added `WarmsUpProperly` to `CandlestickPatternTests`, matching `CommonIndicatorTests.WarmsUpProperly`
and using the same `PatternTestParameters` source as the three tests already there.
`CandlestickPatternTests` does not derive from `CommonIndicatorTests`, so of the four shared tests
the patterns only ever ran two, which is why this went unseen.

My local NUnit host crashes in a Python.NET finalizer before any test runs, because
`QuantConnect.pythonnet` 2.0.65 wants Python 3.11 and this machine has 3.10. So the test was
replayed from a console program linked against `QuantConnect.Indicators`, running the same loop over
the same bars: 61 patterns pass, and reverting the three overrides fails exactly AbandonedBaby,
MatHold and RiseFallThreeMethods. The same program drives and resets all 177 constructible
indicators and finds no other disagreement between `WarmUpPeriod` and `IsReady`.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/CandlestickPatterns/AbandonedBaby.cs
+++ b/Indicators/CandlestickPatterns/AbandonedBaby.cs
@@ -91,6 +91,11 @@ namespace QuantConnect.Indicators.CandlestickPatterns
         }
 
         /// <summary>
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public override int WarmUpPeriod => Period + 1;
+
+        /// <summary>
         /// Computes the next value of this indicator from the given state
         /// </summary>
         /// <param name="window">The window of data held in this indicator</param>
--- a/Indicators/CandlestickPatterns/MatHold.cs
+++ b/Indicators/CandlestickPatterns/MatHold.cs
@@ -86,6 +86,11 @@ namespace QuantConnect.Indicators.CandlestickPatterns
         }
 
         /// <summary>
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public override int WarmUpPeriod => Period + 1;
+
+        /// <summary>
         /// Computes the next value of this indicator from the given state
         /// </summary>
         /// <param name="window">The window of data held in this indicator</param>
--- a/Indicators/CandlestickPatterns/RiseFallThreeMethods.cs
+++ b/Indicators/CandlestickPatterns/RiseFallThreeMethods.cs
@@ -68,6 +68,11 @@ namespace QuantConnect.Indicators.CandlestickPatterns
         }
 
         /// <summary>
+        /// Required period, in data points, for the indicator to be ready and fully initialized.
+        /// </summary>
+        public override int WarmUpPeriod => Period + 1;
+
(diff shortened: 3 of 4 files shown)


Martin-Molinero (approved):
Thank you!

Issue #8039, opened 2024-05-17
Unhandled Exception in AutoRegressiveIntegratedMovingAverage


AlexCatarino:
#### Expected Behavior
Ideally, LEAN doesn't throw unhandled exceptions.

#### Actual Behavior
```
2024-05-17T09:12:16.0912054Z ERROR:: Extensions.SetRuntimeError(): Extensions.SetRuntimeError(): RuntimeError at 05/16/2024 21:00:00 UTC. Context: Consolidators update System.ArgumentException: Matrix must be positive definite.
   at MathNet.Numerics.Providers.LinearAlgebra.ManagedLinearAlgebraProvider.CholeskyFactor(Double[] a, Int32 order)
   at MathNet.Numerics.LinearAlgebra.Double.Factorization.DenseCholesky.Create(DenseMatrix matrix)
   at MathNet.Numerics.LinearRegression.MultipleRegression.NormalEquations[T](T[][] x, T[] y, Boolean intercept)
   at QuantConnect.Indicators.AutoRegressiveIntegratedMovingAverage.MovingAverageStep(Double[][] lags, Double[] data, Double errorMa) in /Indicators/AutoRegressiveIntegratedMovingAverage.cs:line 255
   at QuantConnect.Indicators.AutoRegressiveIntegratedMovingAverage.TwoStepFit(Double[] series) in /Indicators/AutoRegressiveIntegratedMovingAverage.cs:line 230
   at QuantConnect.Indicators.AutoRegressiveIntegratedMovingAverage.ComputeNextValue(IndicatorDataPoint input) in /Indicators/AutoRegressiveIntegratedMovingAverage.cs:line 185
```

#### Potential Solution
The `AutoRegressiveIntegratedMovingAverage.MovingAverageStep` method could not throw an exception if there is a reasonable alternative (return zero?); otherwise, it could raise an exception with the matrix values. 

#### Reproducing the Problem
```python
from AlgorithmImports import *
class ARIMAException(QCAlgorithm):
    def Initialize(self):
        self.set_start_date(2024, 5, 12)
        self.equity = self.add_equity("CRKN", extended_market_hours = True)
        self.arima(self.equity.symbol, 1, 1, 1, 100)
    def on_data(self, data):
        self.log(str(data.bars.get(self.equity.symbol)))
```

Logging the data to show that the prices don't change during extended market hours may cause the issue.

#### Checklist
- [x] I have completely filled out this template
[...]

PR #9714, opened 2026-08-16
Align the AR and MA rows in AutoRegressiveIntegratedMovingAverage

### LINKED ISSUE #9712: ARIMA throws IndexOutOfRangeException when arOrder exceeds maOrder
#### Expected Behavior

`AutoRegressiveIntegratedMovingAverage` fits any order set its constructor accepts.

#### Actual Behavior

Any order set with `arOrder > maOrder` throws `IndexOutOfRangeException` on the bar that
fills the window. `ARIMA(2, 0, 1)` is the smallest. At period 50, 24 of the 80 accepted order
sets throw.

`MovingAverageStep` walks the lagged errors and indexes the AR lags with the same counter:

```csharp
var laggedErrors = LaggedSeries(_maOrder, _residuals.ToArray());
for (var i = 0; i < laggedErrors.Length; i++)
{
    var doubles = lags[i].ToList();
    doubles.AddRange(laggedErrors[i]);
```

`lags[i]` is the row for time `i + _arOrder` and `laggedErrors[j]` the row for `j + _maOrder`,
so the counters agree only when the orders do. `lags` holds `data.Length - _arOrder` rows and
`laggedErrors` holds `data.Length - _maOrder`, so `_arOrder > _maOrder` runs off the end.

The other direction does not throw. It pairs an AR row from one bar with error terms from
another, and the model is fitted on rows that never coexisted.

#### Potential Solution

Walk time rather than a row index, `t` from `Math.Max(_arOrder, _maOrder)` to
`data.Length - 1`, reading `lags[t - _arOrder]` and `laggedErrors[t - _maOrder]`, with the
target vector and the residual loop starting at the same offset.

Separately, the two `catch` blocks added for #8039 read the row width off row zero,
`lags.ToArray()[0].Length` and `appendedData.ToArray()[0].Length`, so when the fit fails
because there are no rows the handler throws instead of substituting zeros. Both widths are
known from the orders.

#### Reproducing the Problem

```csharp
var arima = new AutoRegressiveIntegratedMovingAverage(2, 0, 1, 50, true);
var reference = new DateTime(2020, 1, 1);
for (var i = 0; i < 60; i++)
{
    arima.Update(reference.AddDays(i), 100m + (decimal)Math.Sin(i / 3d) * 5m);
}
// System.IndexOutOfRangeException on the 50th update
```

Every test in `AutoregressiveIntegratedMovingAverageTests` uses
`new AutoRegressiveIntegratedMovingAverage("ARIMA", 1, 0, 1, 50)`, the diagonal where the two
orders match and the loop bound is accidentally right.

#### System Information

master at d865a40b2.

#### Checklist

mkzung:
#### Description

`MovingAverageStep` walked the lagged errors and indexed the AR lags with the same counter,
which runs off the end of `lags` whenever `arOrder > maOrder`. It now walks time and indexes
each array by its own offset.

#### Related Issue

Closes #9712

#### Motivation and Context

`ARIMA(2, 0, 1)` throws at any period. At period 50, 24 of the 80 order sets the constructor
accepts throw.

`lags[i]` is the row for time `i + _arOrder` and `laggedErrors[j]` the row for `j + _maOrder`,
so the counters agree only when the orders do.

The other direction does not throw, and that is the part worth looking hardest at. It pairs
an AR row from one bar with error terms from another, so the model gets fitted on rows that
never coexisted, and the fitted values move once they line up: `ARIMA(1, 0, 2, 50)` over a
deterministic series ends at 100.103983 before this change and 101.581063 after.

Two `catch` blocks added for #8039 also read their row width off row zero, so a fit that
failed for having no rows threw out of the handler meant to absorb it. Both widths follow
from the orders.

177 of 456 accepted order sets still throw at periods of 8 or less, so the issue is not fully
closed. `period < Math.Max(arOrder, maOrder)` is weaker than what the fit needs, by up to six
bars in the range I measured, and tightening it is a separate call.

No.

`AcceptsAnAutoRegressiveOrderAboveTheMovingAverageOrder` covers `(2, 0, 1)`, `(3, 1, 1)` and
`(4, 0, 2)`. All three throw on master. A fourth case with `maOrder > arOrder` came out of the
draft because it passes there, so it would have been a test that never fails.

Every test in the file uses `ARIMA(1, 0, 1, 50)`, the diagonal where the orders match and the
old bound is accidentally right, so the change is a no-op there. That configuration sums to
7090.1814358551 over 120 points of a deterministic series and ends at 103.9912907073, both
before and after.

My local NUnit host crashes in a Python.NET finalizer before any test runs, because
`QuantConnect.pythonnet` 2.0.65 wants Python 3.11 and this machine has 3.10, the same
limitation I noted on #9694. The figures above come from a console program linked against
`QuantConnect.Indicators`, which dumps the outcome of all 640 order-set and period
combinations on master and on this branch and diffs them: 116 newly work, 0 newly throw.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/AutoRegressiveIntegratedMovingAverage.cs
+++ b/Indicators/AutoRegressiveIntegratedMovingAverage.cs
@@ -262,10 +262,16 @@ namespace QuantConnect.Indicators
         {
             var appendedData = new List<double[]>();
             var laggedErrors = LaggedSeries(_maOrder, _residuals.ToArray());
-            for (var i = 0; i < laggedErrors.Length; i++)
+            // lags[i] is the AR row for time i + _arOrder and laggedErrors[j] the MA row for
+            // time j + _maOrder, so the two are only the same row when the orders match.
+            // Walking laggedErrors and indexing lags by the same counter ran off the end of
+            // lags whenever _arOrder > _maOrder, and paired rows from different times when
+            // it was the other way round.
+            var offset = Math.Max(_arOrder, _maOrder);
+            for (var t = offset; t < data.Length; t++)
             {
-                var doubles = lags[i].ToList();
-                doubles.AddRange(laggedErrors[i]);
+                var doubles = lags[t - _arOrder].ToList();
+                doubles.AddRange(laggedErrors[t - _maOrder]);
                 appendedData.Add(doubles.ToArray());
             }
 
@@ -274,7 +280,7 @@ namespace QuantConnect.Indicators
             {
                 try
                 {
-                    maFits = Fit.MultiDim(appendedData.ToArray(), data.Skip(_maOrder).ToArray(),
+                    maFits = Fit.MultiDim(appendedData.ToArray(), data.Skip(offset).ToArray(),
                 method: DirectRegressionMethod.NormalEquations, intercept: _intercept);
                 }
                 catch (Exception ex)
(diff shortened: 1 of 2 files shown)


mkzung:
One thing I should have put in the description: whether the new pairing fits a model, or only stops throwing.

Generated 400 points from `x[t] = 0.6 x[t-1] - 0.3 x[t-2] + noise` and fitted it with a window of 120, intercept off, reading `ArParameters` back:

    truth                 ar1  0.6      ar2 -0.3
    ARIMA(2, 0, 2, 120)   ar1  0.5882   ar2 -0.3114   residual error 0.01211
    ARIMA(2, 0, 1, 120)   ar1  0.5703   ar2 -0.3145   residual error 0.01211
    ARIMA(3, 0, 1, 120)   ar1  0.8144   ar2 -0.4298   residual error 0.01200

The first row is the diagonal this change does not touch. The second only runs with it, and recovers the same coefficients to two figures with the same residual error, so the rows are being paired in time rather than merely paired. The third is an over-specified AR(3) on AR(2) data, where the extra coefficient absorbs the difference.

Worth saying that this took two attempts. The first put the process on top of a level of 100 with no intercept, and every configuration returned coefficients summing to 1.0001: the fit was reproducing the level and the truth was invisible behind it.

Martin-Molinero (approved):
Thank you 👍

PR #9716, opened 2026-08-17
Handle null custom data source during live refresh

### LINKED ISSUE #9715: Live warmup: NullReferenceException in LiveCustomDataSubscriptionEnumeratorFactory when GetSource() returns null
## Summary

`LiveCustomDataSubscriptionEnumeratorFactory.CreateEnumerator`'s refresher lambda throws an unhandled `NullReferenceException` when the wrapped data type's `GetSource(...)` call returns `null`. The exception propagates up through `RefreshEnumerator.MoveNext()` into `BaseDataExchange.ConsumeEnumerators()`, which only logs it (`ERROR:: BaseDataExchange.ConsumeEnumerators(): ...`) and continues — so the crash is silent from the algorithm's point of view, but it fires repeatedly (observed twice per deploy, ~60s apart, matching the factory's `RefreshEnumerator` retry cadence) during live warmup.

## Where

`Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactory.cs`:

```csharp
lastSourceRefreshTime = utcNow;
var localDate = _dateAdjustment?.Invoke(...) ?? ...;
var source = sourceFactory.GetSource(config, localDate, true);          // <- can return null

// fetch the new source and enumerate the data source reader
var enumerator = EnumerateDataSourceReader(config, dataProvider, frontier, source, localDate, sourceFactory);

if (SourceRequiresFastForward(source))                                   // line 88 — NRE here
{
    ...
```

`SourceRequiresFastForward` dereferences `source` unconditionally:

```csharp
private bool SourceRequiresFastForward(SubscriptionDataSource source)
{
    return source.TransportMedium == SubscriptionTransportMedium.LocalFile   // <- NRE if source == null
        || source.TransportMedium == SubscriptionTransportMedium.RemoteFile;
}
```

Nothing between the `GetSource` call and `SourceRequiresFastForward(source)` null-checks `source`, even though `BaseData.GetSource` is a `virtual` method that user/custom data type implementations are free to return `null` from (and evidently do, under some live-warmup condition — the observed subscriptions were plain live equity minute Trade/QuoteBar configs, not user-authored custom data).

## Evidence

Two independent live deployments (same project, same LEAN build `2.5.0.0.18004`) hit the identical unhandled exception at the identical two source lines, ~1 minute apart, early in warmup:

```
ERROR:: BaseDataExchange.ConsumeEnumerators():  System.NullReferenceException: Object reference not set to an instance of an object.
   at QuantConnect.Lean.Engine.DataFeeds.Enumerators.Factories.LiveCustomDataSubscriptionEnumeratorFactory.<>c__DisplayClass5_0.<CreateEnumerator>b__0() in Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactory.cs:line 88
   at QuantConnect.Lean.Engine.DataFeeds.Enumerators.RefreshEnumerator`1.MoveNext() in Engine/DataFeeds/Enumerators/RefreshEnumerator.cs:line 56
   at QuantConnect.Lean.Engine.DataFeeds.BaseDataExchange.ConsumeEnumerators() in Engine/DataFeeds/BaseDataExchange.cs:line 191
```

- Deploy A: fires at 11:40:08.632Z and again at 11:41:08.656Z.
- Deploy B (separate deploy, ~9h earlier, same project): fires at 02:37:52.454Z and again at 02:38:52.697Z.

Both deployments went on to hit the unrelated `Isolator` 10-minute time-loop kill roughly 8-9 minutes later, while still inside warmup (large equity universe — ~5,300 subscriptions, `SetWarmUp(TimeSpan.FromDays(7))`). The isolator timeout looks like the primary/blocking issue for that ticket (tracked separately, likely a warmup-scaling/backend-latency question rather than a Lean code defect); this issue is only about the `NullReferenceException` itself, which is real and reproducible regardless of whether it contributes to the timeout.

## Proposed fix

Null-check `source` right after the `GetSource` call in the refresher lambda (and anywhere else in the class that assumes a non-null `SubscriptionDataSource`), e.g. skip the refresh cycle and return an empty enumerator when `source == null`, mirroring how a `null` source is already tolerated elsewhere in the data-feed pipeline (`BaseDataSubscriptionFactory` / `SubscriptionDataReader` treat a null/failed source as "no data this cycle", not a crash).

## Open questions for QuantConnect engineering

- Which live subscription type's `GetSource(...)` is returning `null` here — is this specific to a particular data type reachable during live equity minute-bar warmup, or can any custom/derived `BaseData.GetSource` override trigger it?
- Should `BaseDataExchange.ConsumeEnumerators()`'s catch-and-log-only behavior around enumerator advancement be tightened so a repeated NRE from the same enumerator surfaces to the algorithm/ops rather than looping silently?

## Reference

Reported via QuantConnect support conversation 215475520166104 (customer-supplied minimal repro + `[WATCHDOG]` diagnostic thread independently isolated the freeze to engine code after `Initialize` returns, during warmup subscription/data setup).

sergio12S:
#### Description

Handle a null source returned by live custom data during an enumerator refresh.

`LiveCustomDataSubscriptionEnumeratorFactory` currently passes the result of `GetSource()` into a deferred reader enumerator and then immediately dereferences it in `SourceRequiresFastForward()`. When `GetSource()` returns null, the exception escapes through `RefreshEnumerator.MoveNext()` and is repeatedly logged by the live data feed.

This change treats a null source as no data for the current refresh cycle. The refresh timestamp is still updated, so the existing retry cadence is preserved, and the data-source reader is not invoked until a valid source is available.

#### Related Issue

Closes #9715.

#### Motivation and Context

Custom data source discovery can legitimately yield no source for a given live refresh. Other data-feed paths tolerate this as an empty cycle; this factory should not turn it into an unhandled `NullReferenceException`.

No.

Added `NullSourceYieldsNoData` to `LiveCustomDataSubscriptionEnumeratorFactoryTests`. The test uses a `BaseData` implementation whose `GetSource()` returns null and verifies that:

- advancing the refresh enumerator does not throw;
- it yields no data for the cycle;
- the subscription data-source reader is never called.

The test failed with the reported `NullReferenceException` before the fix and passes after it. The full factory fixture passes: 9 tests, 0 failures.

#### Types of changes

#### Checklist

- [x] I have read the CONTRIBUTING document.
- [x] The focused factory test fixture passes.
- [x] My branch follows the `bug-<issue#>-<description>` naming convention.

### DIFF
--- a/Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactory.cs
+++ b/Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactory.cs
@@ -81,6 +81,10 @@ namespace QuantConnect.Lean.Engine.DataFeeds.Enumerators.Factories
                 lastSourceRefreshTime = utcNow;
                 var localDate = _dateAdjustment?.Invoke(utcNow.ConvertFromUtc(config.ExchangeTimeZone).Date) ?? utcNow.ConvertFromUtc(config.ExchangeTimeZone).Date;
                 var source = sourceFactory.GetSource(config, localDate, true);
+                if (source == null)
+                {
+                    return Enumerable.Empty<BaseData>().GetEnumerator();
+                }
 
                 // fetch the new source and enumerate the data source reader
                 var enumerator = EnumerateDataSourceReader(config, dataProvider, frontier, source, localDate, sourceFactory);
--- a/Tests/Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactoryTests.cs
+++ b/Tests/Engine/DataFeeds/Enumerators/Factories/LiveCustomDataSubscriptionEnumeratorFactoryTests.cs
@@ -467,6 +467,25 @@ namespace QuantConnect.Tests.Engine.DataFeeds.Enumerators.Factories
             }
         }
 
+        [Test]
+        public void NullSourceYieldsNoData()
+        {
+            var referenceLocal = new DateTime(2017, 10, 12);
+            var referenceUtc = referenceLocal.ConvertToUtc(TimeZones.NewYork);
+            var timeProvider = new ManualTimeProvider(referenceUtc);
+            var dataSourceReader = new Mock<ISubscriptionDataSourceReader>();
+            var config = new SubscriptionDataConfig(typeof(NullSourceData), Symbols.SPY, Resolution.Second,
+                TimeZones.NewYork, TimeZones.NewYork, false, false, false);
+            var request = GetSubscriptionRequest(config, referenceUtc.AddSeconds(-1), referenceUtc.AddDays(1));
+            var factory = new TestableLiveCustomDataSubscriptionEnumeratorFactory(timeProvider, dataSourceReader.Object);
+
+            using var enumerator = factory.CreateEnumerator(request, null);
+
+            Assert.IsTrue(enumerator.MoveNext());
+            Assert.IsNull(enumerator.Current);
+            dataSourceReader.Verify(reader => reader.Read(It.IsAny<SubscriptionDataSource>()), Times.Never);
+        }
+
         [TestCase(10)]
         [TestCase(60)]
         [TestCase(0)]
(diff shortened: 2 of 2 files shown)


AlexCatarino:
Thanks for the fix @sergio12S, the diagnosis and the approach here are both right.

Closing this one as deprecated by #9718, which keeps the same fix for `LiveCustomDataSubscriptionEnumeratorFactory` and extends it:

- `BaseDataCollectionSubscriptionEnumeratorFactory` has the same unguarded null source, reached in backtesting through `SubscriptionDataSourceReader.ForSource`.
- Two null/failure handling bugs in the history path the same deployment hit during warmup, in `SynchronizingHistoryProvider` and `HistoryProviderManager`.
- Coverage for a `PythonData` subclass returning `None`, which is what the reported algorithm actually does.

Nothing wrong with the work here, it is purely that the surrounding fixes are easier to review and ship as one change. Much appreciated.

PR #9727, opened 2026-08-19
Reject a ChoppinessIndex period of one

### LINKED ISSUE #9724: ChoppinessIndex throws OverflowException at a period of one
#### Expected Behavior

`ChoppinessIndex(1)` either works or is rejected by the constructor, the way `ValueAtRisk` rejects a period its formula cannot use.

#### Actual Behavior

The constructor accepts the value. The first update after warm-up throws `System.OverflowException: Value was either too large or too small for a Decimal`, so the failure lands on the data rather than on the call that is wrong.

`Indicators/ChoppinessIndex.cs:106` divides by `Math.Log10(_period)`:

```csharp
return (decimal)(100.0 * Math.Log10(((double) _trueRangeHistory.Sum()) / ((double) (maxHigh - minLow))) / Math.Log10(_period));
```

`Math.Log10(1)` is 0, so the double is infinite and the cast to decimal throws. It is only period 1: `Log10` is zero nowhere else, and a period of zero or less is already rejected by `RollingWindow`.

#### Potential Solution

Reject `period < 2` in the constructor, with the message shape `ValueAtRisk.cs:61` uses.

#### Reproducing the Problem

```csharp
var chop = new ChoppinessIndex(1);
var time = new DateTime(2020, 1, 1);
chop.Update(new TradeBar(time, Symbols.SPY, 100m, 101m, 99m, 100.5m, 1000));
```

#### System Information

master, net10.0.

mkzung:
#### Description

`ChoppinessIndex` now rejects a period below two in its constructor.

#### Related Issue

Closes #9724

#### Motivation and Context

`ComputeNextValue` divides by `Math.Log10(_period)`. At a period of one that divisor is zero, so the double is infinite and the cast to decimal throws `OverflowException` on the first update after warm-up. The constructor took the value without complaint, so the failure landed on the data rather than on the call that was wrong.

`ValueAtRisk`, `Beta`, `Correlation` and `Covariance` already reject a period their formula cannot use, with one message shape between them. This uses the same sentence.

No.

`ChoppinessIndexTests.PeriodBelowMinimumThrows`, copied from `ValueAtRiskTests.PeriodBelowMinimumThrows`.

Keeping the test and reverting only `Indicators/ChoppinessIndex.cs`:

```
Failed!  Failed: 1, Passed: 11, Total: 12
```

with the source in place:

```
Passed!  Failed: 0, Passed: 12, Total: 12
```

The red run reports no exception at all at construction, which is the defect.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/ChoppinessIndex.cs
+++ b/Indicators/ChoppinessIndex.cs
@@ -46,10 +46,15 @@ namespace QuantConnect.Indicators
         /// Creates a new ChoppinessIndex indicator using the specified period and moving average type
         /// </summary>
         /// <param name="name">The name of this indicator</param>
-        /// <param name="period">The period used for rolling windows for highs and lows</param>
+        /// <param name="period">The period used for rolling windows for highs and lows, must be greater than one</param>
         public ChoppinessIndex(string name, int period)
             : base(name)
         {
+            if (period < 2)
+            {
+                throw new ArgumentException($"Period parameter for ChoppinessIndex indicator must be greater than 1 but was {period}");
+            }
+
             _period = period;
 
             _trueRange = new TrueRange();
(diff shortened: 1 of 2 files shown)


Martin-Molinero (approved):
Thank you!

PR #9728, opened 2026-08-19
Reject the two point window Beta, Correlation and Covariance describe

### LINKED ISSUE #9725: Beta, Correlation and Covariance accept the period their message forbids
#### Expected Behavior

All three raise on a period of two, which is what their comment and their message both say.

#### Actual Behavior

`Indicators/Correlation.cs:63`:

```csharp
// Assert the period is greater than two, otherwise the correlation can not be computed
if (period < 2)
{
    throw new ArgumentException($"Period parameter for Correlation indicator must be greater than 2 but was {period}");
}
```

The comment says three, the message says three, the condition admits two. `Beta.cs:62` and `Covariance.cs:55` are the same three lines. `ValueAtRisk.cs:61` states the identical bound and tests `period < 3`.

A two point window has nothing to measure. Fed ten bars of two series that are not proportional, `Correlation` at period 2 returns 1 or -1 on every one of its nine readings; the same series at period 3 gives values between -0.40 and 0.33.

```
period  2: 1 -1 -1 -1 -1 -1 -1 -1 -1
period  3: 0.3273 -0.2402 -0.2402 -0.3974 -0.2895 -0.2895 -0.2895 -0.2895
period  4: 0.3586 0 -0.049 -0.0804 -0.0476 -0.0476 -0.0476
period 10: 0.749
```

So the value is the sign of the last move, whatever the data is.

#### Potential Solution

`period < 3` in all three, matching the message and `ValueAtRisk`. Nothing in `Tests/`, `Algorithm.CSharp/` or `Algorithm.Python/` constructs any of the three with 2; the smallest literal in use is 3.

#### Reproducing the Problem

```csharp
var correlation = new Correlation(Symbols.SPY, Symbols.AAPL, 2);
// feed any two non-proportional series; every reading is 1 or -1
```

#### System Information

master, net10.0.

mkzung:
#### Description

`Beta`, `Correlation` and `Covariance` now test `period < 3`, which is the bound their comment and their message already state.

#### Related Issue

Closes #9725

#### Motivation and Context

All three carry the comment "assert the period is greater than two", raise "must be greater than 2 but was {period}", and then test `period < 2`, so a period of two reaches the calculation.

Two points always correlate perfectly. Fed ten bars of two series that are not proportional, `Correlation` at period 2 returns 1 or -1 on every one of its nine readings:

```
period  2: 1 -1 -1 -1 -1 -1 -1 -1 -1
period  3: 0.3273 -0.2402 -0.2402 -0.3974 -0.2895 -0.2895 -0.2895 -0.2895
period  4: 0.3586 0 -0.049 -0.0804 -0.0476 -0.0476 -0.0476
period 10: 0.749
```

`Beta` and `Covariance` share that window through `DualSymbolIndicator`.

`ValueAtRisk` states the same bound and tests `period < 3`.

No.

One test per indicator, copied from `ValueAtRiskTests.PeriodBelowMinimumThrows`.

Keeping the tests and reverting the three source files:

```
Failed!  Failed: 3, Passed: 45, Total: 48
```

with them in place, including the Spearman fixture:

```
Passed!  Failed: 0, Passed: 64, Total: 64
```

Nothing in `Tests/`, `Algorithm.CSharp/` or `Algorithm.Python/` constructs any of the three with a period of 2, so no existing caller changes; the smallest literal in use is 3.

#### Types of changes

#### Checklist:

### DIFF
--- a/Indicators/Beta.cs
+++ b/Indicators/Beta.cs
@@ -59,7 +59,7 @@ namespace QuantConnect.Indicators
             : base(name, targetSymbol, referenceSymbol, 2)
         {
             // Assert the period is greater than two, otherwise the beta can not be computed
-            if (period < 2)
+            if (period < 3)
             {
                 throw new ArgumentException($"Period parameter for Beta indicator must be greater than 2 but was {period}.");
             }
--- a/Indicators/Beta.cs
+++ b/Indicators/Beta.cs
@@ -58,10 +58,10 @@ namespace QuantConnect.Indicators
         public Beta(string name, Symbol targetSymbol, Symbol referenceSymbol, int period)
             : base(name, targetSymbol, referenceSymbol, 2)
         {
-            // Assert the period is greater than two, otherwise the beta can not be computed
-            if (period < 3)
+            // Assert the period is greater than one, otherwise the beta can not be computed
+            if (period < 2)
             {
-                throw new ArgumentException($"Period parameter for Beta indicator must be greater than 2 but was {period}.");
+                throw new ArgumentException($"Period parameter for Beta indicator must be greater than 1 but was {period}.");
             }
 
             _targetReturns = new RollingWindow<double>(period);
--- a/Indicators/Correlation.cs
+++ b/Indicators/Correlation.cs
@@ -61,7 +61,7 @@ namespace QuantConnect.Indicators
             : base(name, targetSymbol, referenceSymbol, period)
         {
             // Assert the period is greater than two, otherwise the correlation can not be computed
-            if (period < 2)
+            if (period < 3)
             {
                 throw new ArgumentException($"Period parameter for Correlation indicator must be greater than 2 but was {period}");
             }
--- a/Indicators/Covariance.cs
+++ b/Indicators/Covariance.cs
@@ -52,7 +52,7 @@ namespace QuantConnect.Indicators
             : base(name, targetSymbol, referenceSymbol, 2)
         {
             // Assert the period is greater than two, otherwise the covariance can not be computed
-            if (period < 2)
+            if (period < 3)
             {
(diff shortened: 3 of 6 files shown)


Martin-Molinero (changes requested):
Thanks for the PR! Agree on `Correlation`: with a two point window Pearson is always ±1, so rejecting `period < 3` there is the right fix and matches what the message already says.

`Beta` and `Covariance` are a different case though. With `period = 2` the returns windows hold two points, and the variance / covariance of two returns are finite, well defined values, so both indicators work today for anyone using period 2. Tightening the guard to `< 3` would break those users for no functional gain.

Could you please:
- keep the `Correlation.cs` change and its test as is
- revert the guard change in `Beta.cs` and `Covariance.cs`; if you'd like to clean up the mismatch there, fix the comment and exception message to say "greater than 1", which is what the code has always enforced
- adjust the Beta / Covariance tests accordingly (e.g. assert period 1 throws with the corrected message, and period 2 is accepted). As written their comments talk about "the correlation of two points", which doesn't describe those indicators.

Thanks!

mkzung:
You are right, and the tests I wrote gave it away: they pasted the correlation argument into two indicators that do not compute a correlation.

Done as asked.

`Correlation.cs` and its test are untouched. `Beta.cs` and `Covariance.cs` are back to `period < 2`, and their message and the comment above it now say "greater than 1", which is what that guard has always enforced. The mismatch this pull request started from is gone from all three, in the direction the code supports rather than the direction I assumed.

Their tests now assert both halves: period 1 throws with the corrected message, period 2 is accepted. The comments say what each indicator actually needs, that two returns give a finite covariance, rather than repeating the correlation reasoning.

50 tests pass across `BetaIndicatorTests`, `CovarianceTests` and `CorrelationPearsonTests`, and reverting the Beta message with the tests kept fails `PeriodBelowMinimumThrows` at line 323, so it is pinned.

PR #9729, opened 2026-08-19
Run the reset contract over every indicator, not only the tested ones

### LINKED ISSUE #9726: The reset contract runs for 115 of the 202 indicators
#### Expected Behavior

The contracts in `CommonIndicatorTests<T>` hold for every indicator in the assembly.

#### Actual Behavior

They run only for an indicator that has a test class deriving from `CommonIndicatorTests<T>`. There are 115 such classes, covering 115 of the 202 concrete indicators. The other 87 are never asked.

Six reset defects were merged this month out of that gap: #9686, #9687, #9688, #9694, #9705 and #9706. `IntradayVwap` had no test class, so nothing checked that a reset returned it to its default state.

`Tests/Indicators/IndicatorTests.cs:163` already drives a test by reflection over the assembly, but it builds with `Activator.CreateInstance(indicator, new object[] { 10 })`, which reaches 52 of the 202, and passes over the rest with a bare `continue` inside one test method. Its own `Log.Trace` prints the reach.

#### Potential Solution

A fixture that constructs every indicator by reflection and asserts the reset contract on it: feed a series, reset, feed the same series, and compare the values and the ready flag; then assert the default state with `TestHelper.AssertIndicatorIsInDefaultState`. A `TestCaseSource` gives each indicator its own case, so a failure names it, and `Assert.Ignore` with a reason makes anything unreachable visible to the runner.

Worth running at more than one period. `SuperTrend` assigned its previous close on the early return it takes while the average true range warms up, so at any period above one that return fires first and hides it.

#### Reproducing the Problem

```
202  concrete public indicators in QuantConnect.Indicators
115  are named by a CommonIndicatorTests subclass
 87  are not
```

#### System Information

master, net10.0.

mkzung:
#### Description

`Tests/Indicators/IndicatorResetContractTests.cs`, one new file. It constructs every indicator in the assembly by reflection, at three periods, and asserts two things: the same series replayed around a `Reset()` gives the same values and the same ready flag, and the indicator is back in its default state afterwards, using `TestHelper.AssertIndicatorIsInDefaultState`.

#### Related Issue

Closes #9726

#### Motivation and Context

`CommonIndicatorTests<T>` states eleven contracts and runs them only for an indicator with a test class deriving from it: 115 of the 202 concrete indicators. Six reset defects were merged this month out of that gap.

The case for it is not a bug count, since it is green on master. It is what happens when the six merged fixes are reverted, keeping the fixture:

```
Failed!  Failed: 21, Passed: 1121, Skipped: 70, Total: 1212
```

| merged fix | cases | how it shows |
|---|---|---|
| #9686 advance decline watermark | 12 | `IsReady` differs at index 1, across the four indicators sharing the base |
| #9687 ConnorsRSI previous input | 3 | the value differs, first at index 4 |
| #9688 SuperTrend previous close | 1 | the value differs at index 0, at period 1 only |
| #9694 MidPrice rolling extremes | 2 | the value differs at index 0, at periods 2 and 14 |
| #9705 IntradayVwap running sums | 3 | `IsReady` true after the reset |
| #9706 NewHighsNewLows readiness | 0 | the fixture registers assets, so the untracked state never arises |

Three notes on the shape, since each was a choice:

Three periods, because of #9688. `SuperTrend` assigned its previous close on the early return it takes while the average true range warms up, so at any period above one that return fires on the first update and the stale value never shows.

The price series never revisits a level. A repeating one caught two of the six, because a value carried across a reset is one the indicator meets again a few bars later anyway.

Anything unreachable is an `Assert.Ignore` carrying its reason, so the runner counts it and names it rather than passing over it. On master that is 70 cases over 18 indicators: 36 the fixture cannot construct, 18 that throw on the series, 16 where a constructor refuses that period and the ignore quotes the indicator's own message.

Two of the five that throw are worth a look on their own: `AutoRegressiveIntegratedMovingAverage` is #9714, and `ChoppinessIndex` at a period of one is #9724.

No.

```
Passed!  Failed: 0, Passed: 1142, Skipped: 70, Total: 1212, Duration: 391 ms
```

on `b0006b29a`, plus the reverted-fix run above. The neighbouring fixtures are unaffected: `IndicatorTests`, `ChoppinessIndexTests`, `MidPriceTests`, `SuperTrendTests`, `IntradayVwapTests`, `ConnorsRelativeStrengthIndexTests`, `AdvanceDeclineDifferenceTests` and `NewHighsNewLows` together give 0 failed.

#### Types of changes

#### Checklist:

### DIFF
--- a/Tests/Indicators/IndicatorResetContractTests.cs
+++ b/Tests/Indicators/IndicatorResetContractTests.cs
@@ -29,11 +29,12 @@ namespace QuantConnect.Tests.Indicators
     /// the ones with a test class deriving from <see cref="CommonIndicatorTests{T}"/>.
     /// </summary>
     /// <remarks>
-    /// A field left set by the first pass changes the second, which is what a reset defect
-    /// looks like from the outside. Several periods, because a field assigned on an early
-    /// return during warm-up is overwritten on the first update at any larger period.
-    /// A type that cannot be constructed or fed is reported with its reason by
-    /// <see cref="Assert.Ignore(string)"/>.
+    /// A field left set by the first pass changes the second. Several periods, because a
+    /// field assigned on an early return during warm-up is overwritten at any larger one.
+    /// A two symbol indicator gets a bar per symbol stamped alike, since
+    /// <see cref="MultiSymbolIndicator{T}"/> fills its windows only on a shared time.
+    /// A type that cannot be constructed or fed is reported by
+    /// <see cref="Assert.Ignore(string)"/>, with a count of how many warmed up.
     /// </remarks>
     [TestFixture]
     public class IndicatorResetContractTests
@@ -44,11 +45,9 @@ namespace QuantConnect.Tests.Indicators
 
         private static readonly DateTime StartDate = new DateTime(2020, 1, 1);
 
-        private static readonly Symbol Target =
-            new Symbol(SecurityIdentifier.GenerateEquity("SPY", Market.USA, mapSymbol: false), "SPY");
+        private static readonly Symbol Target = Symbols.SPY;
 
-        private static readonly Symbol Reference =
-            new Symbol(SecurityIdentifier.GenerateEquity("IBM", Market.USA, mapSymbol: false), "IBM");
+        private static readonly Symbol Reference = Symbols.IBM;
 
         private static IEnumerable<TestCaseData> Cases()
         {
@@ -61,8 +60,7 @@ namespace QuantConnect.Tests.Indicators
             {
                 foreach (var period in Periods)
                 {
-                    // {m} is the test method, without which the two contracts name their
-                    // cases identically.
+                    // {m} is the test method; without it the two contracts collide.
                     yield return new TestCaseData(indicator, period)
                         .SetName($"{{m}}({indicator.Name}, period {period.ToString(CultureInfo.InvariantCulture)})");
                 }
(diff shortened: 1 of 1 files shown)


Martin-Molinero (changes requested):
Thanks for the PR, this is a nice sweep and the reverted-fix evidence is exactly the right way to justify a test-only change.

Two requests before merging:

1. Please use the existing test symbols instead of hand building them: `Symbols.SPY` / `Symbols.IBM` for `Target` / `Reference`, and the option helpers in `Tests/Symbols.cs` (e.g. `Symbols.CreateOptionSymbol`) for `OptionOn`.

2. The dual symbol indicators (`Beta`, `Correlation`, `Covariance`, ...) are only ever fed `Target` data, so they never become ready and the reset contract is checked on a permanently not-ready indicator. Since those are precisely the indicators with the most internal state, could you feed them alternating `Target` / `Reference` bars (e.g. when the constructor consumed two symbols) so they actually warm up before the reset?

Thanks!

mkzung:
You were right, and it was worse than not being ready: the two symbol indicators never started at all.

Symbols now come from `Tests/Symbols.cs`, `Symbols.SPY` and `Symbols.IBM` with `Symbols.CreateOptionSymbol` for the option ones.

Alternating bars turned out not to be enough. `MultiSymbolIndicator` fills its windows only when every symbol carries the same `CurrentInputEndTimeUtc`, so alternating across days means the two streams never share a timestamp. Each step now sends a target bar and a reference bar stamped alike, on two series that move together without being equal.

The fixture reports its own denominator, taken where `Reset` is called:

```
single stream    527 of 571 replayed cases were ready
paired streams   536 of 571
```

Nine, which is exactly the nine whose constructor took two symbols: `Beta`, `Correlation` and `Covariance` at each period. Every one read `IsReady false` before, so the contract was asserting against an object that had never left its initial state.

1187 passed and 70 skipped across the fixture and the three indicator test classes. The skips are types it still cannot construct or feed, each with its reason.

PR #9741, opened 2026-08-24
Skip paper brokerage dividends during warmup

### LINKED ISSUE #9739: PaperBrokerage.Scan() re-applies warm-up-replayed dividends to cash on every redeploy
## Summary

`PaperBrokerage.Scan()` re-applies dividends that are replayed by `SetWarmUp()`
during a live/paper redeploy, permanently inflating the paper account's cash
balance every time the algorithm is restarted with existing holdings.

## Root cause

`Brokerages/Paper/PaperBrokerage.cs`, `Scan()` (lines 72-95):

```csharp
public override void Scan()
{
    if (Algorithm.UtcTime != _lastScanTime && Algorithm.CurrentSlice != null)
    {
        _lastScanTime = Algorithm.UtcTime;

        // apply each dividend directly to the quote cash holdings of the security
        // this assumes dividends are paid out in a security's quote cash (reasonable assumption)
        foreach (var dividend in Algorithm.CurrentSlice.Dividends.Values)
        {
            Security security;
            if (Algorithm.Securities.TryGetValue(dividend.Symbol, out security))
            {
                var distribution = security.Holdings.Quantity * dividend.Distribution;
                security.QuoteCurrency.AddAmount(distribution);
            }
        }
    }

    base.Scan();
}
```

This has no `Algorithm.IsWarmingUp` guard. Compare with
`AlgorithmManager.HandleDividends` (`Engine/AlgorithmManager.cs`, ~lines 843-872),
which explicitly skips applying a dividend to the portfolio while warming up:

```csharp
if (algorithm.IsWarmingUp)
{
    // skip past dividends during warmup, the algorithms position already reflects them
    Log.Trace($"AlgorithmManager.Run(): {algorithm.Time}: Skip Dividend during warmup: {dividend}");
}
else
{
    algorithm.Portfolio.ApplyDividend(dividend, liveMode, mode);
}
```

`SecurityPortfolioManager.ApplyDividend` (`Common/Securities/SecurityPortfolioManager.cs`,
line 789) additionally early-returns whenever `liveMode` is true regardless of
warm-up state ("we currently don't properly model dividend payable dates, so in
live mode it's more accurate to rely on the brokerage cash sync"). For a real
brokerage that cash sync comes from the broker's own account snapshot.
`PaperBrokerage.Scan()` exists specifically to be that cash-sync substitute for
paper accounts, applying a dividend exactly once at its real ex-date (see the
existing test `AppliesDividendDistributionDirectlyToPortfolioCashBook` in
`Tests/Brokerages/Paper/PaperBrokerageTests.cs`). It was never updated to ignore
the historical dividends that `SetWarmUp()` streams through
`Algorithm.CurrentSlice` on every startup.

`Launcher/config.json` documents that paper brokerage deployments use
`BacktestingTransactionHandler` ("the paper brokerage requires the
BacktestingTransactionHandler"), whose `ProcessSynchronousEvents()` calls
`_brokerage.Scan()` on every loop iteration, including the rapid slice-by-slice
replay that happens while `IsWarmingUp` is true. So every dividend that falls
inside the warm-up lookback window gets credited to cash again, silently (no log
line), on every single redeploy — for a fixed warm-up length and unchanged
holdings, the exact same amount, indefinitely, for as long as those ex-dividend
dates stay inside the trailing warm-up window.

## Repro

1. Live/paper deploy an algorithm on `PaperBrokerage` holding one or more
   dividend-paying US equities, with `self.set_warm_up(N, Resolution.DAILY)`
   where `N` covers at least one past ex-dividend date for a held position.
2. Note the account's cash balance.
3. Stop and redeploy with no code change.
4. Cash balance increases by `sum(held_quantity * per_share_distribution)` for
   every dividend inside the warm-up window, with zero orders placed and the
   market closed. Repeating the redeploy repeats the same credit.

The engine log will show `AlgorithmManager.Run(): ...: Skip Dividend during
warmup: ...` for each of these events (confirming the *official* path correctly
skipped them) immediately followed by
`CloudLiveTradingResultHandler.OnWarmupFinished(): Re-captured starting
portfolio value after warm-up: ...` — the re-captured baseline already reflects
the phantom credit from `Scan()`.

## Proposed fix

Guard the dividend loop in `PaperBrokerage.Scan()` with `!Algorithm.IsWarmingUp`,
matching `AlgorithmManager.HandleDividends`:

```csharp
foreach (var dividend in Algorithm.CurrentSlice.Dividends.Values)
{
    if (Algorithm.IsWarmingUp)
    {
        continue;
    }

    Security security;
    if (Algorithm.Securities.TryGetValue(dividend.Symbol, out security))
    {
        var distribution = security.Holdings.Quantity * dividend.Distribution;
        security.QuoteCurrency.AddAmount(distribution);
    }
}
```

`Tests/Brokerages/Paper/PaperBrokerageTests.cs` has `AppliesDividendDistributionDirectlyToPortfolioCashBook`
and `AppliesDividendsOnce`, but neither sets `IsWarmingUp = true` for the
dividend slice, so this gap was never covered. Recommend a new test that feeds
a dividend slice while `Algorithm.IsWarmingUp` is true and asserts the cash
book is unchanged.

## Reference

Reported via QuantConnect support conversation 215475602810837 (evidence:
customer's own before/after cash reconciliation across 7 redeploys, syslog
confirming 220 warm-up-replayed dividends immediately preceding the post-warmup
baseline capture).

mameikagou:
#### Description
Skip paper-brokerage dividend credits while the algorithm is warming up, and add a regression test for the replayed-dividend case.

#### Related Issue
Fixes #9739

#### Motivation and Context
Warm-up replays historical dividend slices on every paper redeploy. Crediting those slices again permanently inflates the account cash balance.

No.

- Built the affected test project with .NET 10.
- Ran the two affected PaperBrokerage test methods through an isolated harness; both passed.
- Ran `dotnet format --verify-no-changes` on the changed files.
- The standard NUnit entry point could not run in the container because its global Python initialization crashed inside pythonnet.

#### Types of changes

#### Checklist:

### DIFF
--- a/Brokerages/Paper/PaperBrokerage.cs
+++ b/Brokerages/Paper/PaperBrokerage.cs
@@ -79,14 +79,17 @@ namespace QuantConnect.Brokerages.Paper
 
                 // apply each dividend directly to the quote cash holdings of the security
                 // this assumes dividends are paid out in a security's quote cash (reasonable assumption)
-                foreach (var dividend in Algorithm.CurrentSlice.Dividends.Values)
+                if (!Algorithm.IsWarmingUp)
                 {
-                    Security security;
-                    if (Algorithm.Securities.TryGetValue(dividend.Symbol, out security))
+                    foreach (var dividend in Algorithm.CurrentSlice.Dividends.Values)
                     {
-                        // compute the total distribution and apply as security's quote currency
-                        var distribution = security.Holdings.Quantity * dividend.Distribution;
-                        security.QuoteCurrency.AddAmount(distribution);
+                        Security security;
+                        if (Algorithm.Securities.TryGetValue(dividend.Symbol, out security))
+                        {
+                            // compute the total distribution and apply as security's quote currency
+                            var distribution = security.Holdings.Quantity * dividend.Distribution;
+                            security.QuoteCurrency.AddAmount(distribution);
+                        }
                     }
                 }
             }
--- a/Tests/Brokerages/Paper/PaperBrokerageTests.cs
+++ b/Tests/Brokerages/Paper/PaperBrokerageTests.cs
@@ -69,6 +69,7 @@ namespace QuantConnect.Tests.Brokerages.Paper
             var slice = new Slice(algorithm.Time, new List<BaseData>(), algorithm.Time);
             slice.Dividends.Add(new Dividend(Symbols.SPY, algorithm.Time, distributionPerShare, 100m));
             algorithm.SetCurrentSlice(slice);
+            algorithm.SetFinishedWarmingUp();
 
             // invoke brokerage
             using var brokerage = new PaperBrokerage(algorithm, null);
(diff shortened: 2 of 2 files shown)


Martin-Molinero (approved):
Thanks! 🙌

PR #9751, opened 2026-08-26
Validate option underlying subscription resolution

### LINKED ISSUE #9732: Warn when an option is subscribed at a finer resolution than its underlying
#### Expected Behavior

Subscribing an option at a finer resolution than its already-subscribed underlying should throw or warn, since IV/Greeks are computed from the underlying's last price.

#### Actual Behavior

```
equity = self.add_equity("SPY", Resolution.DAILY)
self.add_option_contract(contract, Resolution.MINUTE)
```
is accepted silently. The contract's IV/Greeks update all session against the underlying's stale daily price, so they're quietly wrong.

#### Potential Solution

When adding an option whose underlying is already subscribed at a coarser resolution: warn

#### Reproducing the Problem

Run the two lines of code above in a QC algorithm

#### System Information

QC Cloud

#### Checklist

pengpengyi92:
#### Description

Validate an option contract's effective resolution before adding its subscription. If the finest existing underlying subscription is coarser than the option contract resolution, `AddOptionContract` now throws an `ArgumentException` with guidance to add the underlying at the same or a finer resolution.

The validation uses the finest available underlying subscription, so algorithms with multiple underlying subscriptions continue to work when any subscription is sufficiently fine. It also uses `UniverseSettings.Resolution` when the option resolution argument is omitted.

#### Related Issue

Closes #9732.

#### Motivation and Context

LEAN computes option implied volatility and Greeks from the underlying's latest price. Previously, a minute option subscription could silently use a stale daily underlying price for the entire session. Failing fast prevents silently incorrect analytics without introducing a hidden subscription or changing data costs.

No. The exception explains the required subscription relationship.

- Built `Tests/QuantConnect.Tests.csproj` in Release mode with .NET 10: 0 errors.
- Ran the six new resolution-validation cases in a .NET 10 + Python runtime container: 6 passed, 0 failed.
- Ran the complete `AlgorithmAddDataTests` fixture in the same minimal container: 54 passed. The only two failures were existing Python/Pandas tests because the minimal container intentionally does not include pandas; neither test exercises this change. Upstream CI uses the full `quantconnect/lean:foundation` environment.
- Ran `git diff --check` successfully.

#### Types of changes

#### Checklist:

- [ ] All new and existing tests passed. (All new tests pass; two unrelated existing Python/Pandas tests require the upstream foundation image.)

### DIFF
--- a/Algorithm/QCAlgorithm.cs
+++ b/Algorithm/QCAlgorithm.cs
@@ -2454,6 +2454,14 @@ namespace QuantConnect.Algorithm
                 }
             }
 
+            var optionResolution = resolution ?? UniverseSettings.Resolution;
+            var underlyingResolution = underlyingConfigs.GetHighestResolution();
+            if (underlyingResolution > optionResolution)
+            {
+                throw new ArgumentException(Messages.QCAlgorithm.AddOptionContractUnderlyingResolution(
+                    symbol, optionResolution, underlying, underlyingResolution));
+            }
+
             var configs = SubscriptionManager.SubscriptionDataConfigService.Add(symbol, resolution, fillForward, extendedMarketHours,
                 dataNormalizationMode: DataNormalizationMode.Raw);
             var option = (Option)Securities.CreateSecurity(symbol, configs, leverage, underlying: underlyingSecurity);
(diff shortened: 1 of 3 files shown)


jaredbroad:
Thanks @pengpengyi92 not a bad idea -- can you please reopen as a warning instead of fatal exception, shown once. You'll find similar patterns of "show once" errors in LEAN. Minute resolution might be the best people have, especially for indexes which don't update that frequently.

pengpengyi92:
Thanks for the guidance. I replaced the fatal exception with a warning emitted once per algorithm, updated the tests, and verified the 9 related cases pass. GitHub rejected reopening this PR through the API, so I opened #9759 with the requested behavior: https://github.com/QuantConnect/Lean/pull/9759

PR #9758, opened 2026-08-27
Reject an ARIMA period too short for its own orders

### LINKED ISSUE #9757: ARIMA accepts order sets ComputeNextValue cannot serve
#### Expected Behavior

`AutoRegressiveIntegratedMovingAverage` either fits an order set its constructor accepts, or refuses it at construction.

#### Actual Behavior

168 of the 534 order sets the constructor admits throw on the bar that fills the rolling window, before the indicator ever reports ready. `ComputeNextValue` reads, over a series of length `period - diffOrder`:

```
line 201   arrayData[i]              i < arOrder      needs length >= arOrder
line 209   _residuals[maOrder+i+1]   i < maOrder      needs length >= 2*maOrder + 1
```

The constructor checks only `period < Math.Max(arOrder, maOrder)`. That never mentions `diffOrder`, although `DifferenceSeries` shortens the series it is validating, and it compares against `maOrder` where the second read needs `2*maOrder + 1`. Its message also says "must exceed" while the condition admits `period == arOrder`, the same mismatch as #9725.

`HandleExceptions` defaults to true and promises a zero value, but it wraps only `AutoRegressiveStep` and `MovingAverageStep`. Both of these reads are in `ComputeNextValue`, outside it, so the exception reaches the caller. `QCAlgorithm.ARIMA` passes its arguments straight to this constructor.

| order set | accepted from | first period that works |
|---|---:|---:|
| ARIMA(1, 0, 1) | 1 | 3 |
| ARIMA(1, 1, 1) | 1 | 4 |
| ARIMA(2, 1, 2) | 2 | 6 |

#### Potential Solution

Ask for what the reads need, `Math.Max(arOrder, 2 * maOrder + 1) + Math.Max(diffOrder, 0)`. The clamp matters because nothing validates `diffOrder` as non-negative and `DifferenceSeries` runs only when it is positive, so a negative one costs no samples. That admits 366 of the 600 order sets and none of them throws, so no configuration that works today is refused.

#### Reproducing the Problem

```csharp
var arima = new AutoRegressiveIntegratedMovingAverage(1, 1, 1, 3, true);
var start = new DateTime(2020, 1, 1);
for (var i = 0; i < 4; i++)
{
    arima.Update(start.AddDays(i), 100m + (0.37m * i));
}
```

Throws `ArgumentOutOfRangeException` on the third update. `ARIMA(1, 0, 1)` at period 2 and `ARIMA(1, 2, 0)` at period 2 are the smallest cases at each of the two reads.

#### System Information

macOS 15, .NET 10, current `master` at 07fb0182b.

#### Checklist

mkzung:
#### Description

`AutoRegressiveIntegratedMovingAverage` checks `period < Math.Max(arOrder, maOrder)`, which is weaker than what `ComputeNextValue` reads: `arrayData[arOrder - 1]` and `_residuals[2 * maOrder]`, over a series `DifferenceSeries` shortens. The guard now asks for `Math.Max(arOrder, 2 * maOrder + 1) + Math.Max(diffOrder, 0)`, and its message follows the wording `Beta`, `Correlation` and `ValueAtRisk` already use.

The clamp on `diffOrder` is there because nothing validates it as non-negative and `DifferenceSeries` runs only when it is positive, so a negative one costs no samples.

#### Related Issue

Closes #9757.

#### Motivation and Context

168 of the 534 order sets the old guard admits throw on the bar that fills the window, 127 at line 209 and 41 at line 201, before the indicator reports ready. `HandleExceptions` does not cover them: it wraps the two fitting steps, and both reads are in `ComputeNextValue`. With the change 366 of the 600 are admitted and none throws, so the same 168 are refused and nothing that works today is.

No.

`RejectsAPeriodTooShortForItsOrders` takes six order sets, two of them with a negative `diffOrder`, and asserts both sides of the boundary and the exception text. The ARIMA fixture is 22 passed; reverting the guard fails four cases and reverting the clamp fails the other two.

`IndicatorResetContractTests` gives 1140 passed and 72 skipped on this branch and on master alike, which is where a stricter constructor would show since it builds every indicator at periods 1, 2 and 14. Every ARIMA in the repository is built at period 50, where nothing changes either way.

### DIFF
--- a/Indicators/AutoRegressiveIntegratedMovingAverage.cs
+++ b/Indicators/AutoRegressiveIntegratedMovingAverage.cs
@@ -131,9 +131,14 @@ namespace QuantConnect.Indicators
                     "currently available fitting methods.");
             }
 
-            if (period < Math.Max(arOrder, maOrder))
+            // ComputeNextValue indexes arrayData and _residuals up to 2 * maOrder, over a
+            // series DifferenceSeries shortens only when diffOrder is positive.
+            var shortest = Math.Max(arOrder, maOrder > 0 ? (2 * maOrder) + 1 : 0)
+                + Math.Max(diffOrder, 0);
+            if (period < shortest)
             {
-                throw new ArgumentException("Period must exceed both arOrder and maOrder");
+                throw new ArgumentException($"Period parameter for ARIMA({arOrder}, {diffOrder}, " +
+                    $"{maOrder}) indicator must be at least {shortest} but was {period}");
             }
 
             _arOrder = arOrder;
--- a/Tests/Indicators/AutoregressiveIntegratedMovingAverageTests.cs
+++ b/Tests/Indicators/AutoregressiveIntegratedMovingAverageTests.cs
@@ -69,6 +69,31 @@ namespace QuantConnect.Tests.Indicators
             Assert.IsTrue(arima.IsReady);
         }
 
+        [TestCase(1, 0, 1, 3)]
+        [TestCase(1, 1, 1, 4)]
+        [TestCase(2, 1, 2, 6)]
+        [TestCase(1, 2, 0, 3)]
+        [TestCase(1, -2, 1, 3)]
+        [TestCase(1, -3, 2, 5)]
+        public void RejectsAPeriodTooShortForItsOrders(int arOrder, int diffOrder, int maOrder, int shortest)
+        {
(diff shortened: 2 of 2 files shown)


Martin-Molinero (approved):

Martin-Molinero:
Thanks @mkzung! Sorry for the delay reviewing

PR #9760, opened 2026-08-27
Adjust runtime statistics handling

Martin-Molinero:
Consolidates the runtime statistics update path in `BaseResultsHandler` so both the backtesting and live result handlers share a single code path.

While at it, adds a few sanity limits to the values that get stored in the runtime statistics header:

- Cap the number of runtime statistics kept (50).
- Cap key/value length (200 chars) and truncate longer entries.
- Skip entries whose key or value is not human readable text.
- Emit a one time debug message to the algorithm when an entry is skipped, so it is not silently dropped.

Adds unit tests covering the count/length limits, the readable value check and the one time warning.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018bjhpjT5npGmtVjgLWaDRH

### DIFF
--- a/Engine/Results/BacktestingResultHandler.cs
+++ b/Engine/Results/BacktestingResultHandler.cs
@@ -783,7 +783,7 @@ namespace QuantConnect.Lean.Engine.Results
         {
             lock (RuntimeStatistics)
             {
-                RuntimeStatistics[key] = value;
+                TrySetRuntimeStatistic(key, value);
             }
         }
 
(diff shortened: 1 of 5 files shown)


tommedema:
@Martin-Molinero these limits hit us in production, and without any notice, any reason they are set at 50 and not e.g. 1000?

jaredbroad:
Hi @tommedema, drop a message to support with your use case, and we can consider it. support@quantconnect.com

PR #9761, opened 2026-08-28
Extend LeastSquaresMovingAverage to accept a benchmark reference

### LINKED ISSUE #6984: Extend LeastSquaresMovingAverage to Accept a Benchmark

0xpinara:
#### Description

Adds a Least Squares Moving Average that fits the target prices against a
reference symbol instead of against time. There is a new
`LSMA(target, reference, period)` overload.

I didn't change the old `LeastSquaresMovingAverage`. It is a `WindowIndicator`,
and that puts every input into one window, so it can't hold two symbols. The new
class uses `DualSymbolIndicator`, like Beta and Correlation do. That base class
matches the two symbols by time.

The comment on the issue shows a slightly different signature. It is from May
2024, and `DualSymbolIndicator` was only added in January 2025. So I followed the
newer pattern. I can change it if you prefer.

#### Related Issue

Resolves #6984. It replaces my earlier attempt in #9371. #9426 also targets this
issue, so tell me if you'd rather use that one.

#### Motivation and Context

LSMA could only fit prices against time. There was no built in way to fit them
against a benchmark like SPY.

Yes, the new overload.

The tests use `CommonIndicatorTests`, like the other dual symbol indicators. I
added an `LSMAWithReference` column to `bi_datatest.csv`. I worked those numbers
out with the least squares formula myself, so they don't come from the indicator.
There is also a test that does the same fit by hand without MathNet, and two tests
in `AlgorithmIndicatorsTests`.

All of `QuantConnect.Tests.Indicators` passes. 2988 passed, 0 failed, 5 skipped.

Took me a while to work out why SPY and BTCUSD never paired up. The indicator just
sat at 0 and there was no error. `MultiSymbolIndicator` rounds the timestamps down
before it compares them, added in #8564, and that rounding only runs when the input
has a duration. `IndicatorDataPoint` has none, so it got skipped. Bars have one, so
the problem goes away. The test is `WorksWithDifferentTimeZones`.

#### Types of changes

#### Checklist:


Martin-Molinero (approved):
Thank you @0xpinara!

PR #9778, opened 2026-09-07
Fix stop-limit fills after trigger bar

### LINKED ISSUE #9773: EquityFillModel.StopLimitFill: bars after the trigger bar still require a close through the limit, so a triggered stop-limit never fills on an intra-bar limit touch
#### Expected Behavior

Once a stop-limit order's stop has triggered on an earlier bar, every later bar is entirely post-trigger, so the limit leg should fill under the same rule as a resting limit order (`EquityFillModel.LimitFill`): for a buy, fill when `bar.Low < LimitPrice` at `LimitPrice` (or at the open on a favourable gap); for a sell, when `bar.High > LimitPrice`.

Example, buy stop-limit with stop 100.00 / limit 100.25, minute bars:

- bar N: high 101.50, close 101.00 -> stop triggers (high > stop); close > limit, no fill. Correct: the low of this bar may predate the trigger.
- bar N+2: low 99.80, close 100.60 -> the limit was traded through after the trigger. Expected: filled at 100.25. A plain limit order at 100.25 submitted on bar N would fill here via `LimitFill`.

#### Actual Behavior

`EquityFillModel.StopLimitFill` (and the base `FillModel.StopLimitFill`) apply the close-based test on every bar, not only on the trigger bar:

```csharp
if (prices.High > order.StopPrice || order.StopTriggered)
{
    if (!order.StopTriggered) { order.StopTriggered = true; Parameters.OnOrderUpdated(order); }
    // Fill the limit order, using closing price of bar:
    // Note > Can't use minimum price, because no way to be sure minimum wasn't before the stop triggered.
    if (prices.Current < order.LimitPrice)
    {
        fill.Status = OrderStatus.Filled;
        fill.FillPrice = Math.Min(prices.High, order.LimitPrice);
        fill.FillQuantity = order.Quantity;
    }
}
```

So in the example above the order never fills: bar N+2 has `Current` (close, or ask close with a QuoteBar subscription) above the limit even though the bar traded through it. The comment's rationale ("no way to be sure minimum wasn't before the stop triggered") only holds for the bar on which `StopTriggered` flips; on subsequent bars the whole range is post-trigger.

Observed on a live paper deployment (LEAN v2.5.0.0.18054, `PaperBrokerage`, minute Trade+Quote bars): a buy stop-limit triggered on the first bar after submission, six later bars in the next 20 minutes had lows below the limit, none closed below it, and the order was still `Submitted` when the deployment was stopped 2h15m later. Eleven sibling stop-limits in the same deployment filled exactly as the current rule predicts (first later bar whose close < limit), so this is the model behaving as written, not a data or engine fault. A stop-market order, or a limit order placed once the stop is seen, would both have filled.

Related: #6974 (which data type should trigger the stop) covers the trigger step, not the post-trigger fill test. #2904 / #4363 fixed the fill *price* and the `prices.Current` source but kept the close-only condition.

#### Potential Solution

In `EquityFillModel.StopLimitFill` and `FillModel.StopLimitFill`, distinguish the trigger bar from later bars:

- read `var triggeredOnEarlierBar = order.StopTriggered;` before the trigger check;
- on the trigger bar keep the current close-based test and `Math.Min(prices.High, LimitPrice)` fill price;
- when `triggeredOnEarlierBar` is true, apply the `LimitFill` condition instead (buy: `prices.Low < LimitPrice`, sell: `prices.High > LimitPrice`) with the `LimitFill` fill price (limit price, or the open when the bar opened through the limit, with the same `FilledWithOpenDueToFavorableGap` message). `EquityFillModel` should use the TradeBar for that test the way `LimitFill` does, and `FillModel` its `Prices`.

Tests to add in `Tests/Common/Orders/Fills/EquityFillModelTests.cs` and `FillModelTests.cs`: (1) stop triggers on bar N with close above the limit, bar N+1 low below the limit and close above it -> `Filled` at the limit price on bar N+1; (2) same but bar N+1 opens below the limit -> filled at the open with the favourable-gap message; (3) the trigger bar itself with low below the limit and close above it -> no fill (unchanged); (4) the sell mirror of (1).

Support reference: Intercom conversation 215475792825594.

#### Checklist

matvt-cell:
#### Description

Updates stop-limit fill handling so that bars after the stop has already triggered use resting limit-order semantics.

The trigger bar retains the existing conservative close-based check because OHLC data does not reveal whether the bar's high/low occurred before or after the stop trigger. On subsequent bars, the entire bar is post-trigger, so:

* Buy stop-limit orders fill when the bar low penetrates the limit price.
* Sell stop-limit orders fill when the bar high penetrates the limit price.
* Favorable gaps fill at the opening price.
* The existing trigger-bar behavior remains unchanged.

The change is applied to both `EquityFillModel.StopLimitFill` and the base `FillModel.StopLimitFill`.

#### Related Issue

Fixes #9773

#### Motivation and Context

After a stop-limit order triggered, LEAN continued to apply the close-based limit test on every subsequent bar.

This could leave an already-triggered stop-limit order unfilled even when a later bar traded through its limit price, as long as that bar did not also close through the limit.

The close-based rule is appropriate on the trigger bar because the ordering of OHLC values is unknown. It is unnecessarily conservative on later bars, where the entire price range necessarily occurred after the stop had already triggered.

This change distinguishes those two cases and applies the existing resting limit-order behavior to subsequent bars.

No.

Added regression coverage for both `EquityFillModel` and the base `FillModel`, including:

* Buy stop-limit filling on a later-bar limit penetration while the close remains above the limit.
* Buy stop-limit filling at the open on a favorable gap.
* Trigger-bar low penetration not causing a fill when the close remains above the limit.
* Sell-side mirror of the later-bar penetration case.

Local test results:

* Targeted stop-limit regression corridor: **24/24 passed**
* `EquityFillModelTests` and `ImmediateFillModelTests`: **236/236 passed**

#### Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] Refactor (non-breaking change which improves implementation)
* [ ] Performance (non-breaking change which improves performance. Please add associated performance test and results)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Non-functional change (xml comments/[document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md)ation/etc)

#### Checklist:

* [x] My code follows the code style of this project.
* [ ] I have read the **CONTRIBUTING** [document](https://github.com/QuantConnect/Lean/blob/master/CONTRIBUTING.md).
* [x] I have added tests to cover my changes.
* [x] All new and relevant existing tests passed.
* [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`

### DIFF
--- a/Common/Messages/Messages.Orders.Fills.cs
+++ b/Common/Messages/Messages.Orders.Fills.cs
@@ -57,6 +57,14 @@ namespace QuantConnect
             }
 
             /// <summary>
+            /// Returns a string message saying that the order was filled using the open price due to a favorable gap
+            /// </summary>
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static string FilledWithOpenDueToFavorableGap(Securities.Security security, Prices prices)
+            {
+                return Invariant($@"Due to a favorable gap at {prices.EndTime.ToStringInvariant()} {security.Exchange.TimeZone}, order filled using the open price ({prices.Open})");
+            }
+            /// <summary>
             /// Returns a string message containing the given subscribedTypes
             /// </summary>
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/Common/Orders/Fills/EquityFillModel.cs
+++ b/Common/Orders/Fills/EquityFillModel.cs
@@ -304,25 +304,41 @@ namespace QuantConnect.Orders.Fills
             if (pricesEndTime <= order.Time) return fill;
 
             //Check if the Stop Order was filled: opposite to a limit order
+            var triggeredOnEarlierBar = order.StopTriggered;
             switch (order.Direction)
             {
                 case OrderDirection.Buy:
                     //-> 1.2 Buy Stop: If Price Above Setpoint, Buy:
-                    if (prices.High > order.StopPrice || order.StopTriggered)
+                    if (prices.High > order.StopPrice || triggeredOnEarlierBar)
                     {
-                        if (!order.StopTriggered)
+                        if (!triggeredOnEarlierBar)
                         {
                             order.StopTriggered = true;
                             Parameters.OnOrderUpdated(order);
                         }
 
-                        // Fill the limit order, using closing price of bar:
-                        // Note > Can't use minimum price, because no way to be sure minimum wasn't before the stop triggered.
-                        if (prices.Current < order.LimitPrice)
+                        if (triggeredOnEarlierBar)
                         {
+                            // The entire bar is post-trigger, so the limit leg behaves as a resting limit order.
+                            var tradeBar = GetBestEffortTradeBar(asset, order.Time);
+                            if (tradeBar != null && tradeBar.Low < order.LimitPrice)
+                            {
+                                fill.Status = OrderStatus.Filled;
+                                fill.FillPrice = order.LimitPrice;
+                                fill.FillQuantity = order.Quantity;
+
+                                if (tradeBar.Open < order.LimitPrice)
+                                {
+                                    fill.FillPrice = tradeBar.Open;
+                                    fill.Message = Messages.EquityFillModel.FilledWithOpenDueToFavorableGap(asset, tradeBar);
+                                }
+                            }
+                        }
+                        else if (prices.Current < order.LimitPrice)
+                        {
+                            // On the trigger bar, preserve the conservative close-based test.
                             fill.Status = OrderStatus.Filled;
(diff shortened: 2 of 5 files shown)


Martin-Molinero:
Hey @matvt-cell! Sorry closing the PR, think the solution will need some lean tweaks the core team will need to handle, probably adding a stop limit triggered time to be precise, else it wont behave as intended.

Issue #9777, opened 2026-09-07
Fundamentals: AccessionNumber default period is OneMonth, so HasValue is always false (and no 12M slot for 10-K accessions)


AlexCatarino:
## Expected Behavior

`EarningReports.AccessionNumber` / `FinancialStatements.AccessionNumber` should report `HasValue == true` (and `Value` should resolve without falling back) whenever Morningstar supplies an EDGAR accession number for the filing, the same way the sibling filing-metadata fields `FileDate`, `FormType`, `PeriodType` and `PeriodEndingDate` do.

## Actual Behavior

Both generated `AccessionNumber` classes use `DefaultPeriod => "OneMonth"`:

- `Common/Data/Fundamental/Generated/EarningReportsAccessionNumber.cs:34`
- `Common/Data/Fundamental/Generated/FinancialStatementsAccessionNumber.cs:34`

while every other filing-metadata field in the same classes defaults to `ThreeMonths` (`EarningReportsFileDate.cs:34`, `EarningReportsFormType.cs:34`, `EarningReportsPeriodType.cs:34`, `FinancialStatementsFileDate.cs:34`, `FinancialStatementsPeriodEndingDate.cs:34`).

Morningstar never populates the 1M slot of the accession number, so:

- `HasValue` (which only checks `EarningReports_AccessionNumber_OneMonth`, line 69) is `false` on every row;
- `Value` (line 74) misses the default and only returns a result through the `MultiPeriodField.Value` fallback (first populated period);
- the `ThreeMonths` slot carries the accession number on the large majority of rows (and `Value` does resolve to it through the fallback, so `Value` and `HasValue` disagree on every populated row).

A related gap: the two `AccessionNumber` fields have only 1M/2M/3M/6M/9M periods, while `FileDate`, `FormType`, `PeriodType` and `PeriodEndingDate` also have a `TwelveMonths` period, so the accession number of an annual report (10-K) is never exposed even though its file date and period end are.

Measured with `History<Fundamental>` over 2005-01-01..2024-01-01 in a cloud backtest (LEAN v2.5.0.0.18057):

| ticker | rows | `AccessionNumber.HasValue` | `AccessionNumber.Value` non-empty | `AccessionNumber.ThreeMonths` non-empty | `FormType.HasValue` |
| --- | --- | --- | --- | --- | --- |
| AAPL | 4781 | 0 | 4781 | 4781 | 4781 |
[...]

PR #9780, opened 2026-09-08
Use ThreeMonths as the AccessionNumber default period

0xpinara:
#### Description

Both `AccessionNumber` classes used `DefaultPeriod => "OneMonth"`. I changed them to `"ThreeMonths"` and made the `HasValue` and `Value` overrides read the 3M slot too. Same 3 line change in both files:

- `Common/Data/Fundamental/Generated/EarningReportsAccessionNumber.cs`
- `Common/Data/Fundamental/Generated/FinancialStatementsAccessionNumber.cs`

They now match the other filing metadata fields like `EarningReportsFileDate.cs`. The generator isn't in this repo, so I edited the generated files.

#### Related Issue

#9777

#### Motivation and Context

Morningstar never fills the 1M slot, so `HasValue` was false on every row while `Value` still found the number through the `MultiPeriodField.Value` fallback. The two disagreed on every populated row, and people who check `has_value` first think the accession number is missing.

`DefaultPeriod` is also used by `ConvertPeriod`, so `GetPeriodValue("")` read the wrong slot as well. I changed all three places together.

No.

Two new tests in `FundamentalTests.cs`, and I gave `TestFundamentalDataProvider` a 3M accession number for AAPL so there is something to read.

- `AccessionNumberDefaultsToThreeMonths`: `HasValue`, `Value` and `GetPeriodValue("")` all agree with the 3M slot. This one fails on master with `Expected: True But was: False`.
- `AccessionNumberHasNoValueWhenItIsMissing`: a symbol with no accession number is still `HasValue == false`, so the fix doesn't just make it always true.

Full `QuantConnect.Tests` in Release on .NET 10: 38090 passed, 54 failed, and all 54 fail without my change too. 49 are Python packages missing on my machine, and I re-ran the other 5 on a clean master to check.

#### Note about the 12M gap

I left `TwelveMonths` out because I think it is a data side thing. In #7536, the commit that made these into multi period fields, the generator created `TwelveMonths` for `FileDate`, `FormType`, `PeriodType` and `PeriodEndingDate` but skipped it for both `AccessionNumber` fields, so the period list looks like it comes from the data dictionary. The docs say `AccessionNumber.[Value 1M 2M 3M 6M 9M 12M]`, but that doesn't help, because `EquityPerShareGrowth` is written the same way and really only has 1Y, 3M, 3Y and 5Y.

If you can check that the 12M column exists, I'm happy to add it here. It would be two enum members plus one property and one tuple per class.

#### Types of changes

#### Checklist:

### DIFF
--- a/Common/Data/Fundamental/Generated/EarningReportsAccessionNumber.cs
+++ b/Common/Data/Fundamental/Generated/EarningReportsAccessionNumber.cs
@@ -31,7 +31,7 @@ namespace QuantConnect.Data.Fundamental
         /// <summary>
         /// The default period
         /// </summary>
-        protected override string DefaultPeriod => "OneMonth";
+        protected override string DefaultPeriod => "ThreeMonths";
 
         /// <summary>
         /// Gets/sets the OneMonth period value for the field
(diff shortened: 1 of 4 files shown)


AlexCatarino:
Thank you, @0xpinara .
This code is auto-generated, so we can't fix it here.

Please comment on the open issue and ask to be assigned to it before starting to work.

Issue #9269, opened 2026-02-05
Notify Missing Data Source (404 Error)


AlexCatarino:
#### Expected Behavior
The logs contain information that the data source doesn't have data:

> 20260205 17:43:50.638 ERROR:: Error downloading custom data source file, skipped: RemoteFile: Csv https://data.nasdaq.com/api/v3/datatables/WGC/GOLD_DAILY_USD.csv?api_key=<...> Error: The reader was empty for source: $https://data.nasdaq.com/api/v3/datatables/WGC/GOLD_DAILY_USD.csv?api_key=<...>

#### Actual Behavior
Nothing is displayed

#### Potential Solution
This message:

> 20260205 17:43:50.638 ERROR:: Error downloading custom data source file, skipped: RemoteFile: Csv https://data.nasdaq.com/api/v3/datatables/WGC/GOLD_DAILY_USD.csv?api_key=<...> Error: The reader was empty for source: $https://data.nasdaq.com/api/v3/datatables/WGC/GOLD_DAILY_USD.csv?api_key=<...>

is only available locally to not show API keys. We should pass this message to the user

#### Code to reproduce
```python
class GoldMarketTimingAlgorithm(QCAlgorithm):
    def initialize(self):
        self.set_start_date(2010, 1, 1)
        self._gold = self.add_data(NasdaqDataLink, "WGC/GOLD_DAILY_USD", Resolution.DAILY, TimeZones.UTC, True).symbol
```

#### Checklist
- [x] I have completely filled out this template
- [x] I have confirmed that this issue exists on the current `master` branch
- [x] I have confirmed that this is not a duplicate issue by searching [issues](https://github.com/QuantConnect/Lean/issues)

0xpinara:
Could I pick this one up?

From what I can see the message already reaches the result handler, the problem is that the URL in it still has the api_key, so it can't be shown to the user as it is. Would you want the key masked at the point the message is built, leaving the source itself alone? The exception text has the URL in it too, so there is more than one place to cover.

Also #9271 was closed in April without being merged and without a comment. Was there something wrong with that approach, or did it just get dropped?

Martin-Molinero:
Hey @0xpinara, https://github.com/QuantConnect/Lean/pull/9271 was wrong because it would cause a runtime error, would break different use cases which are working today, the solution should be more precise but generic, to be handled by a QC team member

PR #9803, opened 2026-09-16
Hide the custom data source credentials in the user facing messages

0xpinara:
#### Description
The error message for a failed custom data download had the full source url in it, and that url can hold an api key. That is why the message could not be shown to the user. Now the query string is removed from the messages the user gets.

The source itself is not touched and `ToString()` still returns the full url, so the local logs keep it for debugging like `Messages.Extensions.DownloadDataFailed` already does. The masked version is a new `ToStringWithoutQueryString()`.

Before:
`... skipped: RemoteFile: Csv https://data.nasdaq.com/.../GOLD_DAILY_USD.csv?api_key=<...> Error: ...`

After:
`... skipped: RemoteFile: Csv https://data.nasdaq.com/.../GOLD_DAILY_USD.csv Error: ...`

I also removed the stray `$` that printed `source: $https://...`.

The second commit fixes the same leak in `Api.DownloadBytes`, whose exception `QCAlgorithm.Download()` throws to the user. I kept it separate so you can drop it if it is out of scope.

#### Related Issue
#9269

#### Motivation and Context
Today nothing is shown, so when a data source is missing the algorithm just runs with no data and no reason why. The api key in the message was what blocked showing it.

This only fixes the key. It does not change where the message goes, it still goes to `Results.ErrorMessage`. #9271 wanted it sent to the runtime error queue and was closed without being merged, so let me know if you still want that.

The message also still says the reader was empty instead of naming the 404, because `HandleRemoteSourceFile` catches the download exception and returns null. I can do that in a separate PR if you want it.

No.

I added tests for the new extension method (10 cases), for `ToStringWithoutQueryString`, for the reader message, and one that runs a real `SubscriptionDataReader` with a remote source holding an api key and checks the `DownloadFailed` message. In that last one the fake download provider also throws with the url in its own message. I put the old code back for a moment to be sure that test fails without the fix.

I ran the `Extensions`, `SubscriptionDataSource`, `SubscriptionDataReader`, `TextSubscriptionDataSourceReader`, `IndexSubscriptionDataSourceReader`, `CollectionSubscriptionDataSourceReader` and `RemoteFileSubscriptionStreamReader` tests and they pass. The ones that need Python did not run on my machine.

#### Types of changes

#### Checklist:


Martin-Molinero:
Closing for now, see https://github.com/QuantConnect/Lean/issues/9269 for more details.

PR #9812, opened 2026-09-21
Fix equity dotted-ticker fallback in SymbolCache

### LINKED ISSUE #9781: AddEquity resolution becomes order-dependent when an equity permtick such as LIFE.1 is added before LIFE
Adding multiple equities should resolve each requested ticker to the correct `Symbol` / `SecurityIdentifier` regardless of the order in which `AddEquity()` is called.

For example, given three valid equity map files representing three different securities:

```text
life.1.csv
19980102,life,Q
20140203,life,Q
```

```text
atyr.csv
20190701,life,Q
20240604,life,Q
20501231,atyr,Q
```

```text
life.csv
20260129,life,Q
20501231,life,Q
```

the following securities should resolve to three different SIDs:

```text
LIFE.1  -> old Life Technologies security
ATYR    -> aTyr Pharma security, historically LIFE -> ATYR
LIFE    -> current LIFE security
```

The result should not depend on the order of the calls:

```python
self.AddEquity("LIFE.1", Resolution.Daily)
self.AddEquity("ATYR", Resolution.Daily)
self.AddEquity("LIFE", Resolution.Daily)
```

versus:

```python
self.AddEquity("LIFE", Resolution.Daily)
self.AddEquity("ATYR", Resolution.Daily)
self.AddEquity("LIFE.1", Resolution.Daily)
```

Both call orders should resolve to the same three security identities.

---

#### Actual Behavior

The result currently depends on the order of `AddEquity()` calls.

When the securities are added in this order:

```python
life1 = self.AddEquity("LIFE.1", Resolution.Daily)
atyr = self.AddEquity("ATYR", Resolution.Daily)
life = self.AddEquity("LIFE", Resolution.Daily)
```

LEAN resolves `LIFE.1` and `LIFE` to the same SID:

```text
LIFE.1 internal ID: LIFE R735QTJ8XC9X
ATYR   internal ID: LIFE X5SL0R29JU5H
LIFE   internal ID: LIFE R735QTJ8XC9X
```

As a result:

```text
SID(LIFE.1) == SID(LIFE)
```

even though `life.1.csv` and `life.csv` represent different equity securities with non-overlapping listing periods.

If the order is changed so that `LIFE` is added before `LIFE.1`:

```python
life = self.AddEquity("LIFE", Resolution.Daily)
atyr = self.AddEquity("ATYR", Resolution.Daily)
life1 = self.AddEquity("LIFE.1", Resolution.Daily)
```

the three securities resolve correctly to different SIDs.

This makes equity identity resolution dependent on the state of `SymbolCache` and on the order in which securities are registered.

---

#### Potential Solution

The behavior appears to be related to the fallback logic in `SymbolCache.TryGetSymbol()`.

When an exact ticker is not present in the cache, `SymbolCache` performs a prefix search using:

```csharp
var search = $"{ticker}.";
```

and if exactly one matching cached key exists, it returns that `Symbol`.

This fallback is documented in the source as a backwards-compatibility mechanism for custom data symbols, for example:

```text
BTC -> BTC.Bitcoin
```

However, the same logic also applies to equities.

Therefore, after:

```text
LIFE.1
```

has been added to the cache, a lookup for:

```text
LIFE
```

can find the single `LIFE.*` match and return the `LIFE.1` equity before normal equity symbol generation / map-file resolution occurs.

One possible solution would be to restrict this dotted-symbol fallback to the security types for which it was intended, such as `SecurityType.Base`.

For example, conceptually:

```csharp
var match = Symbols
    .Where(kvp =>
        kvp.Key.StartsWith(
            search,
            StringComparison.InvariantCultureIgnoreCase)
        && kvp.Value is not null
        && kvp.Value.SecurityType == SecurityType.Base)
    .ToList();
```

Another possible solution would be for `AddSecurity` / `AddEquity` to use an exact-cache lookup before creating an equity `Symbol`, instead of using the broader `SymbolCache.TryGetSymbol()` fallback behavior.

For equities, a cached permtick such as:

```text
LIFE.1
```

should not implicitly satisfy an exact request for:

```text
LIFE
```

unless the normal equity map-file / SID resolution determines that they represent the same security.

A regression test could also verify that the following two call orders produce identical SIDs:

```text
LIFE.1 -> ATYR -> LIFE
```

and:

```text
LIFE -> ATYR -> LIFE.1
```

---

#### Reproducing the Problem

The issue can be reproduced using three equity map files.

Create:

```text
Data/equity/usa/map_files/life.1.csv
```

with:

```csv
19980102,life,Q
20140203,life,Q
```

Create:

```text
Data/equity/usa/map_files/atyr.csv
```

with:

```csv
20190701,life,Q
20240604,life,Q
20501231,atyr,Q
```

Create:

```text
Data/equity/usa/map_files/life.csv
```

with:

```csv
20260129,life,Q
20501231,life,Q
```

Then run:

```python


class LifeReuseAlgorithm(QCAlgorithm):

    def Initialize(self):
        self.SetStartDate(2024, 5, 20)
        self.SetEndDate(2026, 3, 2)
        self.SetCash(100000)

        self.UniverseSettings.DataNormalizationMode = DataNormalizationMode.Raw

        life1 = self.AddEquity("LIFE.1", Resolution.Daily)
        atyr = self.AddEquity("ATYR", Resolution.Daily)
        life = self.AddEquity("LIFE", Resolution.Daily)

        self.Log("LIFE.1 internal ID: {}".format(life1.Symbol.ID))
        self.Log("ATYR internal ID: {}".format(atyr.Symbol.ID))
        self.Log("LIFE internal ID: {}".format(life.Symbol.ID))

        if life1.Symbol.ID == life.Symbol.ID:
            raise Exception(
                "LIFE.1 and LIFE incorrectly resolved to the same SID: {}".format(
                    life.Symbol.ID
                )
            )
```

Observed output:

```text
LIFE.1 internal ID: LIFE R735QTJ8XC9X
ATYR   internal ID: LIFE X5SL0R29JU5H
LIFE   internal ID: LIFE R735QTJ8XC9X
```

Changing only the registration order to:

```python
life = self.AddEquity("LIFE", Resolution.Daily)
atyr = self.AddEquity("ATYR", Resolution.Daily)
life1 = self.AddEquity("LIFE.1", Resolution.Daily)
```

avoids the collision.

The expected behavior is that symbol identity should be independent of registration order.

A smaller unit test may also be possible directly against `SymbolCache`:

```csharp
// Conceptual regression test

SymbolCache.Clear();

// Register an equity under a dotted permtick such as LIFE.1.
SymbolCache.Set("LIFE.1", oldLifeSymbol);

// An exact request for the separate equity ticker LIFE should not
// resolve to LIFE.1 through the custom-data dotted-symbol fallback.
Assert.IsFalse(SymbolCache.TryGetSymbol("LIFE", out var symbol));
```

The exact assertion may need to be adapted to the intended `SymbolCache` API behavior, but the important condition is that an equity key such as `LIFE.1` should not shadow the separate equity ticker `LIFE`.

---

#### System Information

```text
LEAN ALGORITHMIC TRADING ENGINE v2.5.0.0
Mode: DEBUG
Python: 3.11.14
Execution: LEAN CLI / local backtesting environment
```

OS:

```text
[Please add the host OS / Docker image information if available]
```

I can reproduce the problem consistently with the call order described above.

---

#### Checklist

* [x] I have completely filled out this template
* [ ] I have confirmed that this issue exists on the current `master` branch
* [x] I have confirmed that this is not a duplicate issue by searching [[issues](https://github.com/QuantConnect/Lean/issues)](https://github.com/QuantConnect/Lean/issues)
* [x] I have provided detailed steps to reproduce the issue

dawNotPoi:
#### Description

Restricts `SymbolCache`'s dotted-ticker fallback to `SecurityType.Base` symbols and adds regression coverage verifying that an Equity key such as `LIFE.1` cannot satisfy a missing exact `LIFE` lookup.

#### Related Issue

Fixes #9781

#### Motivation and Context

The dotted-ticker fallback was introduced for custom-data aliases such as `BTC.Bitcoin -> BTC`. Applying the same fallback to Equity symbols made resolution depend on registration order: caching `LIFE.1` first could cause a later lookup for `LIFE` to return the wrong security identity. Limiting the fallback to Base/custom-data symbols preserves its intended compatibility behavior without allowing an equity permtick to shadow a distinct ticker.

No. This corrects internal symbol-cache lookup behavior without changing a public API.

- Reproduced the pre-fix production behavior with a focused driver: a `LIFE` lookup incorrectly returned the cached `LIFE.1` Equity.
- Verified the post-fix production behavior with the same driver: the Equity fallback is rejected and the Base/custom-data fallback remains available.
- Compiled `Tests/QuantConnect.Tests.csproj` successfully with the pinned .NET 10 SDK container (zero compile errors).
- Verified the two changed files with `dotnet format whitespace --verify-no-changes` and `git diff --check`.
- Attempted the focused NUnit test in the lightweight SDK container. The test host was aborted before the selected assertion because the test assembly unconditionally initializes Python.NET and the container lacks LEAN's foundation Python runtime (`GIL must always be released`). Therefore, this does not claim that the focused NUnit test passed locally; the repository's official environment and complete matrix are left to GitHub CI.

#### Types of changes

#### Checklist:

- [ ] All new and existing tests passed. (Pending GitHub CI; see the local environment limitation above.)
- [x] My branch follows the naming convention `bug-<issue#>-<description>` or `feature-<issue#>-<description>`.

AI-assisted implementation and independent review were used. I reviewed the complete diff and remain responsible for the contribution.


Martin-Molinero (approved):

jhonabreul (approved):
Thank you!

